PR #436, opened 2018-06-21
Leo retries 403 error until it times out (500) if Dataproc is disabled

akarukappadath:

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/google/HttpGoogleDataprocDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/google/HttpGoogleDataprocDAO.scala
@@ -68,8 +68,8 @@ class HttpGoogleDataprocDAO(appName: String,
     retryWithRecoverWhen500orGoogleError { () =>
       executeGoogleRequest(request)
     } {
-      case e: GoogleJsonResponseException if (e.getStatusCode == 403) && (e.getMessage contains "Compute Engine API has not been used")
-        => throw DataprocDisabledException(e.getMessage)
+      case e: GoogleJsonResponseException if e.getStatusCode == 403 &&
+        (e.getDetails.getErrors.asScala.head.getDomain.equalsIgnoreCase("usageLimits")) => throw BucketObjectAccessException.apply(clusterServiceAccount, initScript, e.getMessage)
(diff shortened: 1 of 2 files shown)


akarukappadath:
jenkins retest

akarukappadath:
jenkins retest

akarukappadath:
jenkins retest

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/google/HttpGoogleDataprocDAO.scala
```diff
@@ -65,9 +65,15 @@ class HttpGoogleDataprocDAO(appName: String,
 
     val request = dataproc.projects().regions().clusters().create(googleProject.value, defaultRegion, cluster)
 
-    retryWhen500orGoogleError(() => executeGoogleRequest(request)).map { op =>
+    retryWithRecoverWhen500orGoogleError { () =>
+      executeGoogleRequest(request)
+    } {
+      case e: GoogleJsonResponseException if e.getStatusCode == 403 &&
+        (e.getDetails.getErrors.asScala.head.getDomain.equalsIgnoreCase("usageLimits")) => throw BucketObjectAccessException.apply(clusterServiceAccount, initScript, e.getMessage)
```
vkumra-broad: Why are we throwing a `BucketObjectAccessException` in this case? Seems like it should be a new/different exception relating to dataproc being disabled.

vkumra-broad: Nit: can remove the apply here.

rtitle: Ah good catch. I agree - a different exception type would be better here.

vkumra-broad: Also, do we want to prevent retry every time we get a 403 with  domain : usageLimit`, or should this be a little more specific?

akarukappadath:
jenkins retest

akarukappadath:
jenkins retest

aweng98:
jenkins retest

Issue #417, opened 2018-06-11
404 when trying to localize to new clusters in Leo


danqrodney:
I have created multiple new clusters in Leo this morning (e.g. aou-rw-staging-29/all-of-us), and they appear to initialize correctly, but subsequent attempts to localize fail with a 404... e.g.:

POST /notebooks/aou-test-f1-2116819717/all-of-us/api/localize?async=false HTTP/1.1
api_1                          | Authorization: Bearer XXXXX
api_1                          | Accept: application/json
api_1                          | Content-Type: application/json
api_1                          | User-Agent: Swagger-Codegen/1.0.0/java
api_1                          | 
api_1                          | {"~/workspaces/w0rk/.delocalize.json":"data:application/json;base64,eyJkZXN0aW5hdGlvbiI6ImdzOi8vZmMtNTVmMGUyN2UtMGRkOC00NTUyLTk2NTQtMzQ1YzYwZWIyNjU3L25vdGVib29rcyIsInBhdHRlcm4iOiJcXC5pcHluYiQifQ\u003d\u003d","~/workspaces/w0rk/.all_of_us_config.json":"data:application/json;base64,eyJXT1JLU1BBQ0VfSUQiOiJ3MHJrIiwiQVBJX0hPU1QiOiJhcGktZG90LWxvY2FsaG9zdDo4MDgxIiwiQ0RSX1ZFUlNJT05fQklHUVVFUllfREFUQVNFVCI6InRlc3RfbWVyZ2VfZGVjMjYiLCJXT1JLU1BBQ0VfTkFNRVNQQUNFIjoiYW91LXRlc3QtZjEtMjExNjgxOTcxNyIsIkJVQ0tFVF9OQU1FIjoiZmMtNTVmMGUyN2UtMGRkOC00NTUyLTk2NTQtMzQ1YzYwZWIyNjU3IiwiQ0RSX1ZFUlNJT05fQ0xPVURfUFJPSkVDVCI6ImFsbC1vZi11cy1laHItZGV2In0\u003d"}
api_1                          | --> END POST (636-byte body)
api_1                          | <-- HTTP/1.1 404 Not Found (161ms)

Note that we have *not* enabled jupyterlab in staging. So this appears to be unrelated to that functionality.

danqrodney:
We discovered this was due to the installation of our pyclient library (which brings in jupyter via ipywidgets and was messing with the path that jupyter runs in during startup.)

[...]

PR #441, opened 2018-06-21
Fix JupyterLab and related refactoring

rtitle:
See: https://github.com/DataBiosphere/leonardo/issues/417

- Fixes the localization endpoint when using JupyterLab
- Fixes Jupyter server logging to `~/jupyter.log`
- Forces python2 runtime when starting the Jupyter to server, to prevent user scripts/extensions from changing the runtime
- Organize scripts inside the docker container into subdirectories, and add a `run-jupyter.sh` start script

Tested manually. Existing tests should cover most of these; one thing we don't have is an auto-test for JupyterLab. I can create a separate ticket to add this.

TODO: update `jupyter-docker/README.md`

In all cases:


jupyter-docker/scripts/run-jupyter.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+set -e
+
+# Forces python 2
+JUPYTER_BASE="/usr/bin/python /usr/local/bin/jupyter-notebook"
+JUPYTER_CMD="$JUPYTER_BASE --NotebookApp.nbserver_extensions=\"{'jupyter_localize_extension':True}\" &> ${HOME}/jupyter.log"
```
rtitle: I had to add the `--NotebookApp.nbserver_extensions` to the jupyter command to make it load the `jupyter_localize_extension`. The same thing is also specified in `jupyter_notebook_config.py` but it wasn't getting loaded unless I explicitly added it here.

jupyter-docker/scripts/run-jupyter.sh
```diff
@@ -0,0 +1,10 @@
...
+
+echo $JUPYTER_CMD
+eval $JUPYTER_CMD
```
vkumra-broad: Nit: needs new line

vkumra-broad:
Question: We have two localize/delocalize python files -- `jupyter_localize_extension.py` and `jupyter_delocalize.py`, which one is the one that we are using?

rtitle:
Both - `jupyter_localize_extension.py` adds the `/localize` endpoint for localizing files onto the VM via API. `jupyter_delocalize.py` hooks into Jupyter `ContentsManager` to support delocalizing when the user hits "save" in the Jupyter UI.

rtitle:
jenkins retest

rtitle:
jenkins retest

rtitle:
Test failure should be addressed by https://github.com/DataBiosphere/leonardo/pull/447.

Issue #206, opened 2018-02-20
Putting laptop to sleep breaks notebook refresh token


helgridly:
Reported by Josh Mandel @ Verily. He put his laptop to sleep, came back after lunch, and got the unauthed 401:

```
{"statusCode":401,"source":"leonardo","causes":[],"exceptionClass":"org.broadinstitute.dsde.workbench.leonardo.service.AuthorizationError","stackTrace":[],"message":"Your account is unauthorized"}
```

Presumably because his browser was asleep for longer than an hour, so when it tried to use its token to refresh the token, it was expired.

rtitle:
See the wiki page here: https://github.com/DataBiosphere/leonardo/wiki/Connecting-to-a-Leo-Notebook

Proposed solution 3 should resolve this issue.

katevoss:
There are two situations where a user can get this 401 page: when a user does not have access to the Notebook, and when a user has an expired refresh token. 

Part 1: UI
- [ ] Instead of 401 page, show "Sign-in with Google" button
- [ ] Future improvement: use FC's standard login page

Part 2: Refresh token
- [ ] Check if the user has valid token
- [ ] Check if the user has access to the notebook
- [ ] If the user does have access, refresh the token and set cookie when they sign in
   - [ ] With a valid token, redirect to the notebook

Part 3: Unauthorized
- [ ] If the user does not have access, redirect to the list of Workspaces with a message of why they don't see a notebook
  - [ ] Give action they can take to request access, or information about how to get access

PR #453, opened 2018-07-02
#206 Refresh access token

katevoss:
This PR is to fix when your laptop falls asleep and your cookie expires (#206).

## To Do:
- [ ] Test on dev (create cluster, let cookie expire, use page to refresh cookie, regain access to cluster)
- [x] Replace 401 error with this page (redirect?)
- [x] Update `google-signin-client_id` for prod (currently set to DSDE Dev google project), or get it from somewhere else

**There were no API changes.**

In all cases:

- [x] Get two thumbsworth of review and PO signoff if necessary


ui/src/RefreshTokenButton.html
```diff
@@ -0,0 +1,66 @@
+<!DOCTYPE html>
+<html>
+    <head>
+        <script src="https://apis.google.com/js/platform.js?onload=renderButton" async defer></script>
+
+        <meta name="google-signin-client_id" content="806222273987-jdbngimv3abggnpggkroac2p6r98d8h6.apps.googleusercontent.com">
```
eap: Shouldn't this client ID be supplied during build or page render?

kyuksel:
@katevoss Thank you for your PR. 

Would it be possible to add a UI test for the fix? Are `To Do`s planned to be added to this PR?

kyuksel:
PS. The failed automated test is known to be flaky and should be irrelevant to this PR so `jenkins-retest`ing.

eap:
Also, is this being served as part of the React UI or is it being served as a static asset of the jupyter-docker image? Maybe the source should be in `jupyter-docker/` or `src/main/resources/jupyter` depending on how you intend to deploy it?

ui/src/RefreshTokenButton.html
katevoss: Yes the client ID should be supplied during build. How do I do that?

katevoss:
@kyuksel I'd be happy to add a UI test, are there more instructions on how to add one than there are on the [Contributions](https://github.com/DataBiosphere/leonardo/blob/develop/CONTRIBUTING.md) page? I'm new to writing tests, so if you have other resources for that let me know.

ui/src/RefreshTokenButton.html
rtitle: We can template in the correct value at page render time. For now can you make it `content="$(googleClientId)"` and we can do the substitution when we hook this into Leo (in a separate PR)?

ui/src/RefreshTokenButton.html
```diff
@@ -0,0 +1,66 @@
...
+                console.log('Logged in as: ' + googleUser.getBasicProfile().getName());
+                setCookie(access_token, expires_in);
+            }
+            function setCookie(access_token, expires_in) {
+                var expiresDate = new Date();
+                expiresDate.setSeconds(expiresDate.getSeconds() + expires_in);
+                var cookiePrefix = "LeoToken=";
+                var leoCookie = cookiePrefix + access_token + ";expires=" + expiresDate.toUTCString() + ";secure;path=/;";
+                document.cookie = leoCookie;
+                checkCookie(cookiePrefix);
```
rtitle: Does this refresh the page after the cookie is set?

rtitle (changes requested):
Looks good, a couple comments above.

Actually we're planning on serving this from the Leo proxy itself when an unauthenticated user hits a /notebook URL. Therefore I think it should be stored within the Leo jar, in https://github.com/DataBiosphere/leonardo/tree/develop/src/main/resources. Maybe make a new directory there called `static` or something.

Re testing, not sure how this can be tested currently since there are no functional changes (the page is not hooked up to anything). I propose we make a separate PR to wire up this page to Leo and add a test at that time.

katevoss:
After testing on a FiaB I got an error because the `$(googleClientId)` is not being passed through to the sign in page. 
@rtitle will take a look at what changes need to be made.

rtitle:
I rebased this PR and added a [commit](https://github.com/DataBiosphere/leonardo/pull/453/commits/49ad22ca44e809b56539c2e7caca986784731998) to template the `googleClientId` into `notebook_login.html`. Unit tests pass, however I haven't run auto tests or tested manually yet.

rtitle:
jenkins retest

rtitle:
Tests passed! This still needs to be manually tested, but should be ok to review. 

@kyuksel

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -729,21 +729,6 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
     } yield ()
   }
 
-  // Process a string using map of replacement values. Each value in the replacement map replaces its key in the string.
```
rtitle: Moved to `TemplateHelper` class

kyuksel (changes requested):

src/main/resources/static/notebook_login.html
```diff
@@ -0,0 +1,70 @@
+<!DOCTYPE html>
+<html>
+    <head>
+        <script src="https://apis.google.com/js/platform.js?onload=renderButton" async defer></script>
+
+        <meta name="google-signin-client_id" content="$(defaultClientId)">
```
rtitle: We tested this and it got substituted as:
```
<meta name="google-signin-client_id" content=""806222273987-2ntvt4hnfsikqmhhc18l64vheh4cj34q.apps.googleusercontent.com"">
```
And gapi threw an error because it interpreted the client id as an empty string: "". Need to investigate why this field is coming through with extra quotes.

kyuksel:
Let's also update [Connecting to a Leo Notebook](https://github.com/DataBiosphere/leonardo/wiki/Connecting-to-a-Leo-Notebook) which mentions this issue.

src/main/resources/static/notebook_login.html
rtitle: It's because `template` explicitly quotes the replaced string. Will change this file to not include quotes.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/ProxyRoutes.scala
```diff
@@ -31,44 +34,69 @@ trait ProxyRoutes extends UserInfoDirectives with CorsSupport with CookieHelper
...
+                      setTokenCookie(userInfo, tokenCookieName) {
+                        complete {
+                          logger.debug(s"Successfully set cookie for user $userInfo")
+                          StatusCodes.OK
+                        }
+                      }
                     }
                   }
-                }
+                case None =>
```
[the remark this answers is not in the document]
rtitle: Done for the below case which uses the clientId, not the `setCookie` case.

rtitle:
fiab-start failed because of thurloe.

jenkins retest

rtitle:
jenkins retest

rtitle:
I verified the client id substitution now works, but hit another issue. I was able to sign into Google, but when the page refreshed it just displayed the sign in page again instead of the notebook. Need to debug. Maybe the cookie isn't being set correctly?

PR #457, opened 2018-07-06
User Interface: add start button to stopped clusters.

eap:
This PR adds a button that will start any cluster whose current state is "Stopped". Rather than adding a dedicated start/stop button, I've created a button that manages cluster startup for clusters that have been stopped due to auto-pause. This button replaces the "Connect" button for eligible clusters.

In all cases:


rtitle (approved):
LGTM

PR #458, opened 2018-07-09
Allow users to configure autopause settings for clusters

akarukappadath:

In all cases:


src/main/resources/swagger/api-docs.yaml
```diff
@@ -618,6 +620,13 @@ definitions:
         description: |
           The date and time the cluster was last accessed, in ISO-8601 format.
           Date accessed is defined as the last time the cluster was created, modified, or accessed via the proxy.
+      autopause:
```
rtitle: These are in `Cluster` but I think they should be in `ClusterRequest` instead? Currently in Leo `Cluster` is the API _response_ for put/get/list cluster requests.

rtitle: Looks like `ClusterRequest` should have `autopause` and `autopauseThreshold` and `Cluster` should just have `autopauseThreshold` right?

src/main/resources/swagger/api-docs.yaml
```diff
@@ -618,6 +620,13 @@ definitions:
...
+        type: boolean
+        description: Whether autopause feature is enabled for this specific cluster.
+      autopauseThreshold:
+        type: int
+        description: The number of minutes to elapse before the cluster is autopaused. If autopause is set to false, this value is disregarded. A value of 0 is equivalent to autopause being turned off.
```
rtitle: Good text - only comment is to make it clear that the cluster only auto-pauses after it's _idle_ for the threshold. Maybe "The number of minutes of idle time to elapse..." ?

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
```diff
@@ -75,6 +80,7 @@ trait ClusterComponent extends LeoComponent {
     def serviceAccountKeyId =         column[Option[String]]    ("serviceAccountKeyId",   O.Length(254))
     def stagingBucket =               column[Option[String]]    ("stagingBucket",         O.Length(254))
     def dateAccessed =                column[Timestamp]         ("dateAccessed",          O.SqlType("TIMESTAMP(6)"))
+    def autopauseThreshold =                  column[Int]             ("autopauseThreshold")
```
rtitle: Super nitpick but it's nice to try and align the fields :)

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
```diff
@@ -252,9 +258,25 @@ trait ClusterComponent extends LeoComponent {
...
+//    def getClustersReadyToAutoFreeze(): DBIO[Seq[Cluster]] = {
+//      val now = SimpleLiteral[Timestamp]("NOW")
+//      val tsdiff = SimpleFunction.ternary[TimeUnit, Timestamp, Timestamp, Long]("TIMESTAMPDIFF")
+//      val minute = SimpleLiteral[TimeUnit]("MINUTES")
+//
+//      clusterQueryWithInstancesAndErrorsAndLabels
+//        .filter(record => tsdiff(minute, now, record._1.dateAccessed) > record._1.autopauseThreshold)
+//        .filter(_._1.status inSetBind ClusterStatus.stoppableStatuses.map(_.toString)).result map { recs => unmarshalClustersWithInstancesAndLabels(recs)}
+//    }
+    def getClustersReadyToAutoFreeze(): DBIO[Seq[Cluster]] = {
```
rtitle: Cool - this works? I'm glad :)

Remove the commented code above.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
```diff
@@ -263,7 +285,7 @@ trait ClusterComponent extends LeoComponent {
         clusterStatusQuery
       } else {
         // The trick is to find all clusters that have _at least_ all the labels in labelMap.
-        // In other words, for a given cluster, the labels provided in the query string must be
+        // In other words, for a given cluster, the labels provided in the query strincg must be
```
rtitle: typo

src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20180625_autopause_threshold.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="leonardo" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+
+    <changeSet logicalFilePath="leonardo" author="azmina" id="autopause_threshold">
+        <addColumn tableName="CLUSTER">
+            <column name="autopauseThreshold" type="INT">
+                <constraints nullable="false" />
```
rtitle: Hm, since this is a new not-null column we should probably include a migration to update all existing clusters in the DB. Something like:

1. Add the new column as nullable
2. Update all existing clusters to the current default
3. Make the new column not-null

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -53,8 +53,8 @@ case class InitializationFileException(googleProject: GoogleProject, clusterName
 case class BucketObjectException(gcsUri: String)
   extends LeoException(s"The provided GCS URI is invalid or unparseable: ${gcsUri}", StatusCodes.BadRequest)
 
-case class BucketObjectAccessException(userEmail: WorkbenchEmail, gcsUri: GcsPath)
-  extends LeoException(s"${userEmail.value} does not have access to ${gcsUri.toUri}", StatusCodes.Forbidden)
+case class BucketObjectAccessException(userEmail: Option[WorkbenchEmail], gcsUri: GcsPath, errorMessage: String)
```
rtitle: Is this relevant to this PR? I feel like this was maybe carried over from a different change? Should probably revert it.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -358,9 +359,14 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
       // Create the cluster
       machineConfig = MachineConfigOps.create(clusterRequest.machineConfig, clusterDefaultsConfig)
       initScript = GcsPath(initBucket, GcsObjectName(clusterResourcesConfig.initActionsScript.value))
+      autopauseThreshold = clusterRequest.autopause match {
+        case None => autoFreezeConfig.autoFreezeAfter
+        case Some(false) => 0
+        case _ => if (clusterRequest.autopauseThreshold == None) autoFreezeConfig.autoFreezeAfter else clusterRequest.autopauseThreshold.get
```
rtitle: Can we sanitize against other potential bad inputs? What if we get:
```
autopause = Some(true)
autopauseThreshold = Some(-2452345)
```
Currently we'd pass this directly to the DB query and bad things might happen. `Math.max(0, autopauseThreshold)` would probably resolve this particular issue.

Otherwise, the logic looks good, I think.

rtitle (changes requested):
See comments above. In general this looks great though - nice work!

rtitle:
I think something went wrong with your rebase, because this PR is showing WAY too many commits. :) Not quite sure how to fix it... can you explain how you did the rebase?

rtitle:
This can be closed in favor of https://github.com/DataBiosphere/leonardo/pull/473

PR #473, opened 2018-07-11
Allow users to configure autopause settings for cluster

akarukappadath:

In all cases:

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -570,6 +570,7 @@ definitions:
       - createdDate
       - labels
       - dateAccessed
+      - autopauseThreshold
     properties:
       id:
         type: string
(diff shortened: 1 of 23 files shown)


src/main/resources/swagger/api-docs.yaml
```diff
@@ -694,6 +698,12 @@ definitions:
           Defaults to false.
         type: boolean
         default: false
+      autopause:
+        type: boolean
+        description: Whether autopause feature is enabled for this specific cluster.
```
rtitle: Can you also say: If unset, autopause will be enabled and a system default threshold will be used.

kyuksel: should that default be mentioned somewhere, too?

rtitle: We could say "a system default of 30 minutes". If we change the default however, we need to remember to change the Swagger doc. OTOH, if the user cares about the value they should probably specify the value themselves. :)

kyuksel: fair enough

PR #475, opened 2018-07-12
Retry op when getting 401 status from Sam

aweng98:

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/sam/SamAuthProvider.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/sam/SamAuthProvider.scala
@@ -81,8 +81,10 @@ class SamAuthProvider(val config: Config, serviceAccountProvider: ServiceAccount
     * @return If the given user has permissions in this project to perform the specified action.
     */
   override def hasProjectPermission(userInfo: UserInfo, action: ProjectActions.ProjectAction, googleProject: GoogleProject)(implicit executionContext: ExecutionContext): Future[Boolean] = {
-    Future {
-      blocking(samClient.hasActionOnBillingProjectResource(userInfo, googleProject, getProjectActionString(action)))
+    retryUntilSuccessOrTimeout(shouldInvalidateSamCacheAndRetry, s"SamAuthProvider.hasProjectPermission call failed for ${userInfo.userEmail.value} check in project ${googleProject.value}")(samRetryInterval, samRetryTimeout) { () =>
+      Future {
+        blocking(samClient.hasActionOnBillingProjectResource(userInfo, googleProject, getProjectActionString(action)))
+      }
     }
   }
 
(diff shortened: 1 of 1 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/sam/SamAuthProvider.scala
```diff
@@ -127,11 +129,15 @@ class SamAuthProvider(val config: Config, serviceAccountProvider: ServiceAccount
...
+      owningProjects <- retryUntilSuccessOrTimeout(shouldInvalidateSamCacheAndRetry, s"SamAuthProvider.owningProjects.filterUserVisibleClusters call failed")(samRetryInterval, samRetryTimeout) { () =>
+        Future(blocking(samClient.listOwningProjects(userInfo)).toSet)
+      }
+      createdClusters <- retryUntilSuccessOrTimeout(shouldInvalidateSamCacheAndRetry, s"SamAuthProvider.owningProjects.filterUserVisibleClusters call failed")(samRetryInterval, samRetryTimeout) { () =>
+        Future(blocking(samClient.listCreatedClusters(userInfo)).toSet)
+      }
     } yield {
       clusters.filter { case (project, name) =>
-        owningProjects.contains(project) || createdClusters.contains((project, name))
+        owningProjects.contains(project) || createdClusters.right.get._2.contains((project, name))
```
rtitle: I don't think `.right.get._2` is correct in this case. `retryUntilSuccessOrTimeout` returns a `RetryableFuture`. What you _want_ is this implicit conversion to kick in: https://github.com/broadinstitute/workbench-libs/blob/develop/util/src/main/scala/org/broadinstitute/dsde/workbench/util/Retry.scala#L98

However I don't think it is in this case. You may need to call it explicitly. This code might work:
```
  override def filterUserVisibleClusters(userInfo: UserInfo, clusters: List[(GoogleProject, ClusterName)])(implicit executionContext: ExecutionContext): Future[List[(GoogleProject, ClusterName)]] = {
    for {
      owningProjects <- retryableFutureToFuture(retryUntilSuccessOrTimeout(shouldInvalidateSamCacheAndRetry, s"SamAuthProvider.owningProjects.filterUserVisibleClusters call failed")(samRetryInterval, samRetryTimeout) { () =>
        Future(blocking(samClient.listOwningProjects(userInfo)).toSet)
      })
      createdClusters <- retryableFutureToFuture(retryUntilSuccessOrTimeout(shouldInvalidateSamCacheAndRetry, s"SamAuthProvider.owningProjects.filterUserVisibleClusters call failed")(samRetryInterval, samRetryTimeout) { () =>
        Future(blocking(samClient.listCreatedClusters(userInfo)).toSet)
      })
    } yield {
      clusters.filter { case (project, name) =>
        owningProjects.contains(project) || createdClusters.contains((project, name))
      }
    }
  }
```

rtitle:
One comment above -- the unit test failure is probably due to the issue in `filterUserVisibleClusters`.

aweng98:
jenkins retest

aweng98:
docker images pull failed.

jenkins retest

rtitle:
I think you missed one: `SamAuthProvider.hasProjectPermission()`. Otherwise looks good.

aweng98:
eagle-eyed! added missing retry call in that method.

aweng98:
one related question. retry timeout seems to be really small at 200 millsec. `samRetryTimeout = 200 milliseconds`. Do you agree with increase it a little?

rtitle:
Hm yeah it is pretty short. I guess currently it tries up to 3 times, sleeping 100ms between each try. So:

call <fail>
sleep(100)
call <fail>
sleep(100)
call <fail>
exception is thrown

I think we didn't want it to retry so long that the actual _Leo_ API call would time out -- then again 100ms is pretty short. I bet retrying 3 times sleeping 500ms or even 1 second would be fine.

Issue #463, opened 2018-07-10
Create a user script to install GATK


rtitle:
In the absence of customizable docker containers, we should look at creating a user script which installs GATK to a cluster and allows use from within a notebook. The cluster already has Java installed, so we would need to:
   - download the GATK jar
   - install necessary python libraries (which ones)?

We can put the user script here: https://github.com/DataBiosphere/leonardo/tree/develop/jupyter-docker/user-scripts. It would also be nice to document the different user scripts we have available in a README.md in that directory.

kyuksel:
[Hard Filtering tutorial](https://drive.google.com/drive/folders/1IHApTUhdlptMW7EDGFhmUUmB_f-JBu4K) can be used as an example of the kind of GATK workflows we'd like to enable on Notebooks via the script.

abaumann:
do we want to tie this to a specific gatk version, or have it use latest?

rtitle:
Probably the latest I imagine?

kyuksel:
@tmm211 Is there a particular GATK version you need for the ASHG workshop?

tmm211:
When you are ready to test the Hard Filtering tutorial, let me know so that we can generate the data for you. 
Is there a way for us to specify the version we want? I ask because say we are a month away from a workshop- we usually stick with whatever version of GATK we have then. If a new version gets released a few days before the workshop, we will not use it because we haven't had enough time to test all the commands with it. Also others who may use the script outside of the workshop may need a specific version...they won't always use the latest.

abaumann:
I just have the GATK jars working - none of the python stuff if you need that.  right now it's a script you specify from a file in a bucket, so i can't make it dynamic enough to choose a version.  but it would be easy to cut a new version of the script right before the workshop

tmm211:
ok that sounds good. I don't think we need the python stuff, but maybe I am missing something @rtitle ? The hard filtering tutorial actually requires some R packages, which I should be able to install after running this script, right?

rtitle:
Yeah R (and python) packages should be installable in a notebook. Let's leave them out of the script unless we have a good reason to add them there. 

@abaumann once the script is in good shape can you add it to https://github.com/DataBiosphere/leonardo/tree/develop/jupyter-docker/user-scripts and make a PR? Thanks.

abaumann:
I have the jar parts all set - I'll have tiffany test that out and then make a PR

tmm211:
Testing...

[...]

PR #478, opened 2018-07-13
#463: Added user install script for gatk ...

abaumann:
...that can be used on cluster initialization.  Does not deal with any python packages that are needed for certain tools, only deals with the jars.

In all cases:

### DIFF
--- a/jupyter-docker/user-scripts/install_gatk.sh
+++ b/jupyter-docker/user-scripts/install_gatk.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+set -e
+GATK_VERSION=4.0.6.0
+GATK_ZIP_PATH=/tmp/gatk-$GATK_VERSION.zip
+
+#download the gatk zip if it doesn't already exist
+if ! [ -f $GATK_ZIP_PATH ]; then
+  # curl and follow redirects and output to a temp file
+  curl -L -o $GATK_ZIP_PATH https://github.com/broadinstitute/gatk/releases/download/$GATK_VERSION/gatk-$GATK_VERSION.zip
+fi
+
+# unzip with forced overwrite (if necessary) to /bin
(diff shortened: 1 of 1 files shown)


jupyter-docker/user-scripts/install_gatk.sh
```diff
@@ -0,0 +1,16 @@
...
+GATK_ZIP_PATH=/tmp/gatk-$GATK_VERSION.zip
+
+#download the gatk zip if it doesn't already exist
+if ! [ -f $GATK_ZIP_PATH ]; then
+  # curl and follow redirects and output to a temp file
+  curl -L -o $GATK_ZIP_PATH https://github.com/broadinstitute/gatk/releases/download/$GATK_VERSION/gatk-$GATK_VERSION.zip
+fi
+
+# unzip with forced overwrite (if necessary) to /bin
+unzip -o $GATK_ZIP_PATH -d /bin/
```
rtitle: Nit: `/bin/` might not be the best place to install the full gatk distribution. How about unzipping it to `/etc/` and then symlinking `/etc/gatk-$GATK_VERSION/gatk -> /bin/gatk`?

jupyter-docker/user-scripts/install_gatk.sh
```diff
@@ -0,0 +1,16 @@
...
+
+# make a symlink to gatk right inside bin so it's available from the existing PATH
+ln -s /bin/gatk-4.0.6.0/gatk /bin/gatk
```
rtitle: Could use the `$GATK_VERSION` variable here

abaumann:
Good catch on version - i also changed to installing into /etc and symlinking to that in /bin like suggested

rtitle (approved):
:shipit:! Pending verification from Tiffany.

rtitle:
Weird Jenkins build error - jenkins retest

PR #480, opened 2018-07-16
Synchronous portion of PUT createCluster v2

kyuksel:
This PR adds a route for `v2` of the `PUT createCluster` endpoint, which returns the response once done with the minimum number of tasks that needs to be performed. It also registers the asynchronous tasks as callbacks but isn't yet dealing with the results of those, which will come as a follow-on PR. I'll also factor out code duplications in that PR.

In all cases:


src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/LeoRoutes.scala
```diff
@@ -51,46 +51,63 @@ abstract class LeoRoutes(val leonardoService: LeonardoService, val proxyService:
             }
           }
         } ~
```
kyuksel: The route diffs look bigger than they are. I just added a top branch that checks for `v2`.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/sam/SamAuthProvider.scala
```diff
@@ -149,7 +149,8 @@ class SamAuthProvider(val config: Config, serviceAccountProvider: ServiceAccount
     * @param clusterName   The user-provided name of the Dataproc cluster
     * @return A Future that will complete when the auth provider has finished doing its business.
     */
```
kyuksel: Non-functional change below to split lines for easier viewing.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/sam/SamAuthProvider.scala
```diff
@@ -174,7 +175,8 @@ class SamAuthProvider(val config: Config, serviceAccountProvider: ServiceAccount
     * @param clusterName      The user-provided name of the Dataproc cluster
     * @return A Future that will complete when the auth provider has finished doing its business.
     */
```
kyuksel: Non-functional change below to split lines for easier viewing.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/LeoRoutes.scala
```diff
@@ -51,46 +51,63 @@ abstract class LeoRoutes(val leonardoService: LeonardoService, val proxyService:
...
-                  leonardoService.createCluster(userInfo, GoogleProject(googleProject), ClusterName(clusterName), cluster).map { cluster =>
-                    StatusCodes.OK -> cluster
+        pathPrefix("cluster") {
+          pathPrefix("v2" / Segment / Segment) { (googleProject, clusterName) =>
+            pathEndOrSingleSlash {
+              put {
+                entity(as[ClusterRequest]) { cluster =>
+                  complete {
+                    leonardoService
+                      .syncCreateCluster(userInfo, GoogleProject(googleProject), ClusterName(clusterName), cluster)
```
vkumra-broad: Shouldn't this be asyncCreateCluster?

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/LeoRoutes.scala
rtitle: `?w=1` helps with this

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/google/HttpGoogleDataprocDAO.scala
```diff
@@ -58,7 +58,12 @@ class HttpGoogleDataprocDAO(appName: String,
     new Oauth2.Builder(httpTransport, jsonFactory, null)
       .setApplicationName(appName).build()
 
-  override def createCluster(googleProject: GoogleProject, clusterName: ClusterName, machineConfig: MachineConfig, initScript: GcsPath, clusterServiceAccount: Option[WorkbenchEmail], credentialsFileName: Option[String], stagingBucket: GcsBucketName): Future[Operation] = {
+  override def createCluster(googleProject: GoogleProject,
+                             clusterName: ClusterName,
+                             machineConfig: MachineConfig,
+                             initScript: GcsPath,
+                             clusterServiceAccount: Option[WorkbenchEmail],
+                             credentialsFileName: Option[String], stagingBucket: GcsBucketName): Future[Operation] = {
```
rtitle: Put `stagingBucket` on its own line to follow the pattern?

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
```diff
@@ -111,6 +111,8 @@ trait ClusterComponent extends LeoComponent {
 
   object clusterQuery extends TableQuery(new ClusterTable(_)) {
 
+    def save(cluster: Cluster): DBIO[Cluster] = save(cluster, initBucket = None, serviceAccountKeyId = None)
```
rtitle: Any downside to just adding the default args to the existing `save` method? E.g.
```
def save(cluster: Cluster, initBucket: Option[GcsPath] = None, serviceAccountKeyId: Option[ServiceAccountKeyId] = None): DBIO[Cluster]
```

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -131,6 +131,25 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
     }
   }
 
+  // Meant to be run synchronously to the cluster creation API call
+  // i.e. We wait for the method to finish before sending the API response
```
rtitle: Pedantic, but technically it still returns a Future so it's still async in some sense. Maybe the comment could read "We complete the API response with the result of the returned Future"

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/LeoRoutes.scala
rtitle: +1, I think asyncCreateCluster would be a clearer name because this variant creates a Cluster asynchronously.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -180,6 +199,61 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
...
+          asyncClusterCreate(userEmail, serviceAccountInfo, googleProject, clusterName, augmentedClusterRequest)
+            .onComplete {
+              case Success(_) =>
+                // Finally, notify the cluster monitor that the cluster has been created
+                clusterMonitorSupervisor ! ClusterCreated(cluster, clusterRequest.stopAfterCreation.getOrElse(false))
+              case Failure(_) =>
+                // If we didn't succeed, createGoogleCluster removes resources in Google but
+                // we also need to notify our auth provider that the cluster has been deleted.
+                // We won't wait for that deletion, though.
+                authProvider.notifyClusterDeleted(userEmail, userEmail, googleProject, clusterName)
```
rtitle: If the async portion fails we should report the error to the user, probably by setting the Cluster status to Error in the DB and adding a `ClusterError` record.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -180,6 +199,61 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
...
+          clusterRequest, userEmail, clusterName, googleProject,
+          serviceAccountInfo, machineConfig, dataprocConfig.clusterUrlBase, autopauseThreshold)
+
+        val attemptToSaveClusterInDb: Future[Cluster] = dbRef.inTransaction(_.clusterQuery.save(initialCluster))
+
+        // For the success case, register the following callbacks...
+        attemptToSaveClusterInDb foreach { cluster =>
+          // Notify the auth provider and kick off the cluster creation on the Google Dataproc side
+          asyncClusterCreate(userEmail, serviceAccountInfo, googleProject, clusterName, augmentedClusterRequest)
+            .onComplete {
```
rtitle: For debuggability's sake, it'd be nice to add logging for when the async portion begins and ends for a cluster. I'd like be able to trace a full cluster creation lifecycle by looking at logs. For example we could log:
1. the API response committed (we have this)
2. the async portion starting
3. the async portion ending
4. the ClusterMonitorActor starting (we have this)
5. the ClusterMonitorActor finishing (we have this)

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -180,6 +199,61 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
...
+
+  def asyncClusterCreate(userEmail: WorkbenchEmail,
+                         serviceAccountInfo: ServiceAccountInfo,
+                         googleProject: GoogleProject,
+                         clusterName: ClusterName,
+                         clusterRequest: ClusterRequest): Future[(Cluster, GcsBucketName, Option[ServiceAccountKey])] = {
+    // Notify the AuthProvider
+    authProvider.notifyClusterCreated(userEmail, googleProject, clusterName)
+    // Proceed to creating the cluster on the Google side
+    // TODO Update this method so the stuff it returns is persisted
```
rtitle: So in the next PR it'll persist the Google stuff in the DB before kicking off the ClusterMonitorActor?

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/LeoRoutes.scala
rtitle: I think I see, `syncCreateCluster` is for the sync portion of the creation and `asyncClusterCreate` is for the async portion. It's a bit confusing because `leonardoService.syncCreateCluster` is _async_ compared with `leonardoService.createCluster`. :) Maybe there's some naming scheme for these methods so the intention is more clear. LeonardoService is not public-facing so we should be free to refactor names, etc.

src/test/scala/org/broadinstitute/dsde/workbench/leonardo/api/LeoRoutesSpec.scala
```diff
@@ -51,23 +51,32 @@ class LeoRoutesSpec extends FlatSpec with ScalatestRouteTest with CommonTestData
...
+    val clusterNames = Seq(clusterName.value, s"${clusterName.value}-v2")
+    val baseUrl = "/api/cluster"
+    val versionedBaseUrls = Seq(baseUrl, s"$baseUrl/v2")
 
-    Put(s"/api/cluster/${googleProject.value}/${clusterName.value}", newCluster.toJson) ~>
-      timedLeoRoutes.route ~> check {
-      status shouldEqual StatusCodes.OK
-
-      validateCookie { header[`Set-Cookie`] }
+    // PUT endpoints have two versions
```
rtitle: Should probably add some cases to `LeonardoServiceSpec` as well to validate the new methods (unless that'll happen in the follow-on PR).

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
kyuksel: Yes. That seems to be how it's done in the current version as well so I assume it would work. Let me know if I'm missing something.

src/test/scala/org/broadinstitute/dsde/workbench/leonardo/api/LeoRoutesSpec.scala
kyuksel: Yea, I think it'll be more meaningful to add the cases to `LeonardoServiceSpec` as part of making sure the asynchronous operations do what they are supposed to do so that the `GET` endpoint returns the same results as for clusters created via `PUT v1`.

PR #484, opened 2018-07-16
Update jupyter notebook terminal to use bash instead of dash

akarukappadath:
This has been manually tested.

In all cases:


kyuksel:
Was it not feasible to have an automated test for this? It feels like a fundamental enough feature worthy of at least smoke-testing?

rtitle:
Interacting with the terminal using Selenium turned out to be tricky because it's an HTML canvas. @akarukappadath or @aweng98 might have more details.

Issue #495, opened 2018-07-18
FISS calls return 401 after cluster pause/resume


rtitle:
Methods team was seeing this; I reproduced manually.

rtitle:
```
import firecloud.api as fc

fc.get_workspace("broad-firecloud-dsde", "Demo_2017-10-24")
<Response [401]>

fc.list_workspaces()
<Response [401]>
```

rtitle:
It looks like a scopes issue.  Here is a cluster that I restarted:
```
fc.__SESSION.credentials.__dict__
{'_scopes': ['https://www.googleapis.com/auth/cloud-platform'],
 '_service_account_email': 'pet-100182905304480635229@broad-firecloud-dsde.iam.gserviceaccount.com',
 'expiry': datetime.datetime(2018, 7, 18, 21, 4, 55, 242141),
 'token': 'ya29.c.Elr8BW7dr2WsgxTElXaS18odGRUKO-TjmfcV8HQ4xVH8KrGX7aod7sr0Vqd2aTtRPT-MqkxfjmXKQSoUMmcCndz8H2umNEY2MpL6xSL8qNVGmspLxj1bH11reUU'}
```
It has the correct SA and token, but not the right set of scopes. Compare with a cluster that I have NOT restarted:
```
fc.__SESSION.credentials.__dict__
{'_scopes': ['https://www.googleapis.com/auth/bigquery',
  'https://www.googleapis.com/auth/cloud.useraccounts.readonly',
  'https://www.googleapis.com/auth/devstorage.read_write',
  'https://www.googleapis.com/auth/logging.write',
  'https://www.googleapis.com/auth/source.read_only',
  'https://www.googleapis.com/auth/userinfo.email',
  'https://www.googleapis.com/auth/userinfo.profile'],
 '_service_account_email': 'pet-100182905304480635229@broad-firecloud-dsde.iam.gserviceaccount.com',
 'expiry': datetime.datetime(2018, 7, 18, 21, 18, 55, 486709),
 'token': 'ya29.c.EmT8BdhxK1C5jOdDvV5UgBkNHDXrju_hsUqODr2UvgoyBkCRkOkp9dkBi8XGC8fKtdvixaD0k1xI9U-qLLhbXk-fO6Ernmu1JHTKuILfATCpoPO4FnfTCL4kXShyTCwy8YUP1sGC'}
```
It has the correct set of scopes, including email/profile which is required by FC.

[...]

PR #503, opened 2018-07-21
Set service account scopes on cluster resume

rtitle:
See https://github.com/DataBiosphere/leonardo/issues/495

When we create a cluster, we set a service account and scopes which go into VM metadata. For some reason, if the instance is stopped/started, the service account persists but the scopes do not. This causes errors when using libraries which require certain scopes (e.g. FISS, AoU client) after pause/resume.

This makes Leo re-set the service account and scopes on the VM at cluster resume time. I verified manually that this works, and also added automation tests. 

In all cases:

### DIFF
--- a/automation/project/Dependencies.scala
+++ b/automation/project/Dependencies.scala
@@ -45,6 +45,11 @@ object Dependencies {
       excludeApacheHttpClient,
       excludeGoogleJsr305,
       excludeJacksonCore),
+    "com.google.apis" % "google-api-services-sourcerepo" % s"v1-rev21-1.23.0" excludeAll (
+      excludeGuavaJdk5,
+      excludeApacheHttpClient,
+      excludeGoogleJsr305,
+      excludeJacksonCore),
     "com.google.api-client" % "google-api-client"   % "1.23.0" excludeAll (
       excludeGuavaJdk5,
       excludeApacheHttpClient,
--- a/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/ClusterMonitoringSpec.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/ClusterMonitoringSpec.scala
@@ -32,7 +32,7 @@ class ClusterMonitoringSpec extends FreeSpec with LeonardoTestUtils with Paralle
           cluster.serviceAccountInfo.notebookServiceAccount shouldBe None
 
           withWebDriver { implicit driver =>
-            withNewNotebook(cluster) { notebookPage =>
+            withNewNotebook(cluster, PySpark2) { notebookPage =>
               // should not have notebook credentials because Leo is not configured to use a notebook service account
               verifyNoNotebookCredentials(notebookPage, PySpark2, petEmail)
             }
--- a/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/ClusterMonitoringSpec.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/ClusterMonitoringSpec.scala
@@ -26,7 +26,7 @@ class ClusterMonitoringSpec extends FreeSpec with LeonardoTestUtils with Paralle
         val petEmail = getAndVerifyPet(project)
 
         // Create a cluster
-        withNewCluster(project, kernel = PySpark2, apiVersion = V2) { cluster =>
+        withNewCluster(project, apiVersion = V2) { cluster =>
           // cluster should have been created with the pet service account
           cluster.serviceAccountInfo.clusterServiceAccount shouldBe Some(petEmail)
           cluster.serviceAccountInfo.notebookServiceAccount shouldBe None
@@ -34,7 +34,7 @@ class ClusterMonitoringSpec extends FreeSpec with LeonardoTestUtils with Paralle
           withWebDriver { implicit driver =>
             withNewNotebook(cluster, PySpark2) { notebookPage =>
               // should not have notebook credentials because Leo is not configured to use a notebook service account
-              verifyNoNotebookCredentials(notebookPage, PySpark2, petEmail)
+              verifyClusterServiceAccount(notebookPage, PySpark2, petEmail)
             }
           }
         }
@@ -63,7 +63,7 @@ class ClusterMonitoringSpec extends FreeSpec with LeonardoTestUtils with Paralle
           withWebDriver { implicit driver =>
             withNewNotebook(cluster) { notebookPage =>
               // should have notebook credentials
-              verifyNotebookCredentials(notebookPage, petEmail)
+              verifyNotebookServiceAccount(notebookPage, PySpark2, petEmail)
             }
           }
         }
@@ -121,7 +121,7 @@ class ClusterMonitoringSpec extends FreeSpec with LeonardoTestUtils with Paralle
             withNewNotebook(cluster, kernel = Python3) { notebookPage =>
               notebookPage.executeCell(s"""print("$printStr")""") shouldBe Some(printStr)
               // Also verify the credentials on the cluster
-              verifyNoNotebookCredentials(notebookPage, Python3, petEmail)
+              verifyClusterServiceAccount(notebookPage, Python3, petEmail)
               notebookPage.saveAndCheckpoint()
             }
 
@@ -142,7 +142,7 @@ class ClusterMonitoringSpec extends FreeSpec with LeonardoTestUtils with Paralle
               notebookPage.executeCell("sum(range(1,10))") shouldBe Some("45")
               // The credentials should be correct after pause/resume
               // See https://github.com/DataBiosphere/leonardo/issues/495.
-              verifyNoNotebookCredentials(notebookPage, Python3, petEmail)
+              verifyClusterServiceAccount(notebookPage, Python3, petEmail)
             }
           }
         }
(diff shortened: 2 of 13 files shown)


automation/project/Dependencies.scala
```diff
@@ -45,6 +45,11 @@ object Dependencies {
       excludeApacheHttpClient,
       excludeGoogleJsr305,
       excludeJacksonCore),
+    "com.google.apis" % "google-api-services-sourcerepo" % s"v1-rev21-1.23.0" excludeAll (
```
rtitle: Just added for the `CloudSourceRepositoriesScopes` class.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/google/GoogleDataprocDAO.scala
```diff
@@ -10,7 +10,14 @@ import org.broadinstitute.dsde.workbench.model.google._
...
 trait GoogleDataprocDAO {
-  def createCluster(googleProject: GoogleProject, clusterName: ClusterName, machineConfig: MachineConfig, initScript: GcsPath, clusterServiceAccount: Option[WorkbenchEmail], credentialsFileName: Option[String], stagingBucket: GcsBucketName): Future[Operation]
+  def createCluster(googleProject: GoogleProject,
+                    clusterName: ClusterName,
+                    machineConfig: MachineConfig,
+                    initScript: GcsPath,
+                    clusterServiceAccount: Option[WorkbenchEmail],
+                    credentialsFileName: Option[String],
+                    stagingBucket: GcsBucketName,
+                    serviceAccountScopes: Seq[String]): Future[Operation]
```
rtitle: Added `serviceAccountScopes`

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -331,15 +339,7 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
 
         // Now stop each instance individually
         _ <- Future.traverse(cluster.nonPreemptibleInstances) { instance =>
-          // Install a startup script on the master node so Jupyter starts back up again once the instance is restarted
```
rtitle: Moved this installation from cluster stop time to cluster start time. I think it's better to set the cluster metadata lazily, right before it's started. This ensures that recent changes are applied to clusters that have been previously auto-paused.

rtitle:
jenkins retest

rtitle:
jenkins retest

rtitle:
I think tests should pass; need to rebase with https://github.com/DataBiosphere/leonardo/pull/505 and re-run.

rtitle:
jenkins retest

rtitle:
jenkins retest

rtitle:
jenkins retest

rtitle:
jenkins retest

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/ClusterMonitoringSpec.scala
```diff
@@ -30,9 +30,9 @@ class ClusterMonitoringSpec extends FreeSpec with LeonardoTestUtils with Paralle
           cluster.serviceAccountInfo.notebookServiceAccount shouldBe None
 
           withWebDriver { implicit driver =>
-            withNewNotebook(cluster) { notebookPage =>
+            withNewNotebook(cluster, PySpark2) { notebookPage =>
               // should not have notebook credentials because Leo is not configured to use a notebook service account
-              verifyNoNotebookCredentials(notebookPage)
+              verifyNoNotebookCredentials(notebookPage, PySpark2, petEmail)
```
kyuksel: The name `verifyNoNotebookCredentials` sounds misleading to me. Given what it does, shouldn't it be something like `verifyNotebookCredentials`, or something more explanatory?

kyuksel: But I now see that there _is_ also a `verifyNotebookCredentials`, which verifies a subset of things. My suggestion would be to refactor those two methods like so:

```scala
def verifyOauth2Client()
def verifyFiss()
def verifySpark()
def verifyMetadataServer() // preferably something even more specific like verifyScopes()
```
and then if it is necessary, we can have some convenience methods that bundle some of those together as necessary.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/google/HttpGoogleComputeDAO.scala
```diff
@@ -156,6 +156,13 @@ class HttpGoogleComputeDAO(appName: String,
     }
   }
 
+  override def setServiceAccount(instanceKey: InstanceKey, serviceAccountEmail: WorkbenchEmail, serviceAccountScopes: Seq[String]): Future[Unit] = {
```
kyuksel: Nit: Can we stack the input parameters to shorten line length?

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/google/HttpGoogleDataprocDAO.scala
```diff
@@ -212,7 +207,7 @@ class HttpGoogleDataprocDAO(appName: String,
       }
   }
 
-  private def getClusterConfig(machineConfig: MachineConfig, initScript: GcsPath, clusterServiceAccount: Option[WorkbenchEmail], credentialsFileName: Option[String], stagingBucket: GcsBucketName): DataprocClusterConfig = {
+  private def getClusterConfig(machineConfig: MachineConfig, initScript: GcsPath, clusterServiceAccount: Option[WorkbenchEmail], credentialsFileName: Option[String], stagingBucket: GcsBucketName, serviceAccountScopes: Seq[String]): DataprocClusterConfig = {
```
kyuksel: Nit: Can we stack the input parameters to shorten line length?

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -100,10 +103,15 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
   //
   // The || clause is included because older clusters may not have the run-jupyter.sh script installed,
   // so we need to fall back running `jupyter notebook` directly. See https://github.com/DataBiosphere/leonardo/issues/481.
-  private lazy val masterInstanceStartupScript: immutable.Map[String, String] = {
+  private[service] lazy val masterInstanceStartupScript: immutable.Map[String, String] = {
```
kyuksel: Could remove the `immutable`s since `Map` is already an alias for it.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
kyuksel: good call 👍

kyuksel (commented):
The changes look good overall 👍 Minor comments inline. Also wondering if the scope change warrants a review by the security team?

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
rtitle: Actually... I'm seeing more frequent test failures, "should pause and resume a cluster with preemptible instances" against this PR. I'm wondering if setting the instance metadata needs to be done at stop time, as it was before. I might move that back, but keep `setServiceAccount` at cluster start time.

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/ClusterMonitoringSpec.scala
rtitle: 👍 , agreed that looks much better.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
kyuksel: Hah...it's curious why that'd cause it

automation/project/Dependencies.scala
aweng98: can you upgrade workbench-service-test version to the latest `val serviceTestV = "0.12-1a00943"`? Credential scopes is newly required parameter in `UserAuthToken`. I need `createDownloadDirectory()` in new version of workbench-service-lib.

rtitle:
Rebased, I'm going to clean up and use the PR to address https://github.com/DataBiosphere/leonardo/issues/542 as well.

rtitle:
jenkins retest

automation/project/Dependencies.scala
rtitle: I updated to the latest `workbench-service-test`.

rtitle:
Rebased this PR and fixed an [outstanding issue](https://github.com/DataBiosphere/leonardo/pull/503/commits/1abd2266373c56b51407be7932072f09074dbabc) where instance metadata was being set in the wrong place. I think this should be in good shape now -- need to verify automation tests.

rtitle:
gpalloc issues -- jenkins retest

kyuksel:
The following paragraph on the [online doc](
https://software.broadinstitute.org/firecloud/documentation/article?id=11603) will become out of date with this PR:

>FireCloud will soon be able to automatically pause idle clusters and sync your notebooks back to your
workspace. For now though, please be mindful of both running costs and the need to save your data. We don't want anyone to suffer from unhappy surprises.

rtitle:
jenkins retest

rtitle:
I am closing this for now because it's not quite ready.

Issue #492, opened 2018-07-18
Reduce the size of the Jupydocker image by cleaning the apt cache


rtitle:
I think there might be opportunities to reduce our image size. For example we have multiple layers doing `apt-get update && apt-get install ...`. These layers don't clean the apt cache when they're done, so they end up being huge. Try cleaning the apt cache at the end of every layer, and see what kind of size reduction that buys us.

vkumra-broad:
Reduces the docker image size by about 100MB

PR #507, opened 2018-07-24
Work in Progress: adding tidyverse

akarukappadath:

In all cases:

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/NotebookInstallSpec.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/NotebookInstallSpec.scala
@@ -22,7 +22,7 @@ class NotebookInstallSpec extends ClusterFixtureSpec {
               |bx.bitset.sys.copyright""".stripMargin
 
           notebookPage.executeCell("1+1") shouldBe Some("2")
-          notebookPage.executeCell(getPythonVersion) shouldBe Some("3.4.2")
+          //notebookPage.executeCell(getPythonVersion) shouldBe Some("3.4.2")
           notebookPage.executeCell(getBxPython).get should include("Copyright (c)")
         }
       }
--- a/jupyter-docker/Dockerfile
+++ b/jupyter-docker/Dockerfile
@@ -1,6 +1,6 @@
 # adapted from https://hub.docker.com/r/jupyter/base-notebook/ AKA https://github.com/jupyter/docker-stacks/tree/master/base-notebook
 
-FROM debian:jessie
+FROM debian:stretch
 
 USER root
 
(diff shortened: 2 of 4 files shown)


src/main/resources/jupyter/init-actions.sh
```diff
@@ -228,6 +228,9 @@ if [[ "${ROLE}" == 'Master' ]]; then
     log 'Starting Jupyter Notebook...'
     docker exec -d ${JUPYTER_SERVER_NAME} ${JUPYTER_SCRIPTS}/run-jupyter.sh
     log 'All done!'
+else
+    apt-get install -yq --no-install-recommends python3.5
```
rtitle: I wonder if you need to call `update_apt_get` in order for this to work. It looks like that is currently only being called on the master.

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/NotebookInstallSpec.scala
```diff
@@ -22,7 +22,7 @@ class NotebookInstallSpec extends ClusterFixtureSpec {
               |bx.bitset.sys.copyright""".stripMargin
 
           notebookPage.executeCell("1+1") shouldBe Some("2")
-          notebookPage.executeCell(getPythonVersion) shouldBe Some("3.4.2")
+          //notebookPage.executeCell(getPythonVersion) shouldBe Some("3.4.2")
```
rtitle: Would be good to fix the version here. I suspect a lot of tests need updating :)

jupyter-docker/Dockerfile
```diff
@@ -12,12 +12,14 @@ ENV DEBIAN_FRONTEND noninteractive
...
- && echo "deb $CRAN_REPO/bin/linux/debian jessie-cran34/"       >> /etc/apt/sources.list \
- && apt-key adv --keyserver keys.gnupg.net \
-    --recv-key 'E19F5F87128899B192B1A2C2AD5F960A256A04AF' \
+RUN apt-get update && apt-get install -my wget gnupg && apt-get install -my wget locales
+RUN apt-cache search locales
+RUN echo "deb $DEBIAN_REPO/debian stretch main"                   > /etc/apt/sources.list \
+ && echo "deb $DEBIAN_REPO/debian-security stretch/updates main" >> /etc/apt/sources.list \
+ && echo "deb $DEBIAN_REPO/debian stretch-backports main"        >> /etc/apt/sources.list \
+ && echo "deb $CRAN_REPO/bin/linux/debian stretch-cran34/"       >> /etc/apt/sources.list \
+ # && apt-key adv --keyserver keys.gnupg.net \
```
rtitle: Is this not needed? I admit I'm not sure what it is for :)

jupyter-docker/Dockerfile
```diff
@@ -213,7 +216,8 @@ RUN R -e 'install.packages(c( \
     "dplyr",  \
     "tidyr",  \
     "reshape2",  \
-    "bigrquery"),  \
+    "bigrquery", \
+    "tidyverse"),  \
```
rtitle: Were there any duplicated libraries now that we're installing tidyverse?

rtitle:
Looks good for early changes. A few questions above (I haven't tried the new docker image myself). Looks like some tests failed because they expect the old library versions.

While we're in the Dockerfile, there is 1 more change I wanted to try. :) It's described in https://github.com/DataBiosphere/leonardo/issues/492. Basically we can try cleaning the apt-get cache at the end of every layer that uses `apt-get`. This might reduce the image size.

rtitle:
Should this be closed?

PR #509, opened 2018-07-26
Added tidyverse to preinstalled R packages

akarukappadath:

In all cases:

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/NotebookRKernelSpec.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/NotebookRKernelSpec.scala
@@ -104,8 +104,7 @@ class NotebookRKernelSpec extends ClusterFixtureSpec {
     "should have tidyverse automatically installed" in { clusterFixture =>
       withWebDriver { implicit driver =>
         withNewNotebook(clusterFixture.cluster, RKernel) { notebookPage =>
-          notebookPage.executeCell("library(tidyverse)").get should not include("there is no package")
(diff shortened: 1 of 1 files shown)


vkumra-broad:
LGTM!
@akarukappadath , as per the ticket, what was the impact on the image size and was there any package duplication?

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/NotebookRKernelSpec.scala
```diff
@@ -101,6 +101,14 @@ class NotebookRKernelSpec extends ClusterFixtureSpec {
       }
     }
 
+    "should have tidyverse automatically installed" in { clusterFixture =>
+      withWebDriver { implicit driver =>
+        withNewNotebook(clusterFixture.cluster, RKernel) { notebookPage =>
+          notebookPage.executeCell("library(tidyverse)").get should not include("there is no package")
```
rtitle: Could we add a positive verification here? E.g. `should include X`.

The reason is that this could still throw an error meaning it's not working correctly, but still pass the test because it doesn't contain the string "there is no package".

jupyter-docker/Dockerfile
```diff
@@ -215,7 +215,8 @@ RUN R -e 'install.packages(c( \
     "dplyr",  \
     "tidyr",  \
     "reshape2",  \
-    "bigrquery"),  \
+    "bigrquery",  \
+   "tidyverse"), \
```
rtitle: Echoing @vkumra-broad question about whether there is library duplication now that we're installing tidyverse.

akarukappadath:
jenkins retest

PR #511, opened 2018-07-30
Async Cluster Creation - Season Finale!

kyuksel:
**The flow of `LeonardoService` methods called upon `PUT /api/cluster/v2/{googleProject}/{clusterName}`:**
1. `processClusterCreationRequest`: We wait for it to complete before completing the request.
2. `initiateClusterCreation`: Start cluster creation work if the cluster doesn't already exist. Any better naming suggestions are welcome.
3. `stageClusterCreation`
4. `completeClusterCreation`
5. `createGoogleCluster` & `AuthProvider.notifyClusterCreated`

**Other notes:**
- Errors encountered during the asynchronous portion (i.e. the part API request completions no longer wait for) are persisted in DB, which is now the only way to retrieve them (as opposed being part of the API response in v1)

In all cases:

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/Leonardo.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/Leonardo.scala
@@ -39,14 +39,6 @@ object Leonardo extends RestClient with LazyLogging {
   }
 
   object cluster {
-    private case class InvalidCreateClusterApiVersionException(googleProject: GoogleProject,
-                                                               clusterName: ClusterName,
-                                                               apiVersion: String)
-      extends LeoException(
-        message = s"Cluster ${googleProject.value}/${clusterName.string} cannot be created " +
-          s"because the API was called with an invalid version: $apiVersion",
-        statusCode = StatusCodes.NotFound)
-
     // TODO: custom JSON deserializer
     // the default doesn't handle some fields correctly so here they're strings
     @JsonIgnoreProperties(ignoreUnknown = true)
@@ -121,7 +113,7 @@ object Leonardo extends RestClient with LazyLogging {
 
     def listIncludingDeleted()(implicit token: AuthToken): Seq[Cluster] = {
       val path = "api/clusters?includeDeleted=true"
-
+      logger.info(s"Listing all clusters including deleted: GET /$path")
       handleClusterSeqResponse(parseResponse(getRequest(s"$url/$path")))
     }
 
@@ -140,7 +132,7 @@ object Leonardo extends RestClient with LazyLogging {
 
       // TODO Find and fix the root-cause of why we get successive 404's without this sleep, and remove the sleep
       val sleepTime = 4000
-      logger.info(s"Sleeping for ${sleepTime/2000} seconds before GET'ing /$path...")
+      logger.info(s"Sleeping for ${sleepTime/1000} seconds before GET'ing /$path...")
       Thread sleep sleepTime
 
       val cluster = handleClusterResponse(parseResponse(getRequest(url + path)))
--- a/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/Leonardo.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/Leonardo.scala
@@ -6,7 +6,7 @@ import java.util.UUID
 
 import akka.http.scaladsl.model.headers.{Authorization, Cookie, HttpCookiePair, OAuth2BearerToken}
 import akka.Done
-import akka.http.scaladsl.model.HttpHeader
+import akka.http.scaladsl.model.{HttpHeader, StatusCodes}
 import akka.http.scaladsl.server.Route
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties
 import com.fasterxml.jackson.databind.DeserializationFeature
@@ -15,6 +15,7 @@ import org.broadinstitute.dsde.workbench.ResourceFile
 import org.broadinstitute.dsde.workbench.service.RestClient
 import org.broadinstitute.dsde.workbench.auth.AuthToken
 import org.broadinstitute.dsde.workbench.config.LeoAuthToken
+import org.broadinstitute.dsde.workbench.leonardo.Leonardo.ApiVersion.V1
 import org.broadinstitute.dsde.workbench.leonardo.StringValueClass.LabelMap
 import org.broadinstitute.dsde.workbench.model.WorkbenchEmail
 import org.broadinstitute.dsde.workbench.model.google._
@@ -38,6 +39,14 @@ object Leonardo extends RestClient with LazyLogging {
   }
 
   object cluster {
+    private case class InvalidCreateClusterApiVersionException(googleProject: GoogleProject,
+                                                               clusterName: ClusterName,
+                                                               apiVersion: String)
+      extends LeoException(
+        message = s"Cluster ${googleProject.value}/${clusterName.string} cannot be created " +
+          s"because the API was called with an invalid version: $apiVersion",
+        statusCode = StatusCodes.NotFound)
+
     // TODO: custom JSON deserializer
     // the default doesn't handle some fields correctly so here they're strings
     @JsonIgnoreProperties(ignoreUnknown = true)
@@ -99,9 +108,11 @@ object Leonardo extends RestClient with LazyLogging {
       }
     }
 
-    def clusterPath(googleProject: GoogleProject, clusterName: ClusterName): String =
-      s"api/cluster/${googleProject.value}/${clusterName.string}"
-//      s"api/cluster/v2/${googleProject.value}/${clusterName.string}"
+    def clusterPath(googleProject: GoogleProject,
+                    clusterName: ClusterName,
+                    version: ApiVersion = V1): String = {
+      s"api/cluster${version.toUrlSegment}/${googleProject.value}/${clusterName.string}"
+    }
 
     def list()(implicit token: AuthToken): Seq[Cluster] = {
       logger.info(s"Listing all active clusters: GET /api/clusters")
@@ -114,8 +125,12 @@ object Leonardo extends RestClient with LazyLogging {
       handleClusterSeqResponse(parseResponse(getRequest(s"$url/$path")))
     }
 
-    def create(googleProject: GoogleProject, clusterName: ClusterName, clusterRequest: ClusterRequest)(implicit token: AuthToken): Cluster = {
-      val path = clusterPath(googleProject, clusterName)
+    def create(googleProject: GoogleProject,
+               clusterName: ClusterName,
+               clusterRequest: ClusterRequest,
+               version: ApiVersion = ApiVersion.V1)
+              (implicit token: AuthToken): Cluster = {
+      val path = clusterPath(googleProject, clusterName, version)
       logger.info(s"Create cluster: PUT /$path")
       handleClusterResponse(putRequest(url + path, clusterRequest))
     }
@@ -263,6 +278,29 @@ object Leonardo extends RestClient with LazyLogging {
       replacementMap.foldLeft(raw) { case (source, (key, replacement)) =>
         source.replaceAllLiterally("$("+key+")", s"""'$replacement'""")
       }
+    }
+  }
+
+  sealed trait ApiVersion {
+    override def toString: String
+    def toUrlSegment: String
+  }
+
+  object ApiVersion {
+    case object V1 extends ApiVersion {
+      override def toString: String = "v1"
+      def toUrlSegment: String = ""
+    }
+
+    case object V2 extends ApiVersion {
+      override def toString: String = "v2"
+      def toUrlSegment: String = "/v2"
+    }
+
+    def fromString(s: String): Option[ApiVersion] = s match {
+      case "v1" => Some(V1)
+      case "v2" => Some(V2)
+      case _ => None
     }
   }
 }
--- a/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/Leonardo.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/Leonardo.scala
@@ -66,7 +66,7 @@ object Leonardo extends RestClient with LazyLogging {
                                     jupyterExtensionUri: Option[String],
                                     jupyterUserScriptUri: Option[String],
                                     stagingBucket: String,
-                                    errors:List[ClusterError],
+                                    errors: List[ClusterError],
                                     dateAccessed: String) {
 
       def toCluster = Cluster(clusterName,
@@ -121,7 +121,7 @@ object Leonardo extends RestClient with LazyLogging {
 
     def listIncludingDeleted()(implicit token: AuthToken): Seq[Cluster] = {
       val path = "api/clusters?includeDeleted=true"
-      // logger.info(s"Listing all clusters including deleted: GET /$path")
+
       handleClusterSeqResponse(parseResponse(getRequest(s"$url/$path")))
     }
 
@@ -137,8 +137,15 @@ object Leonardo extends RestClient with LazyLogging {
 
     def get(googleProject: GoogleProject, clusterName: ClusterName)(implicit token: AuthToken): Cluster = {
       val path = clusterPath(googleProject, clusterName)
+
+      // TODO Find and fix the root-cause of why we get successive 404's without this sleep, and remove the sleep
+      val sleepTime = 7000
+      logger.info(s"Sleeping for ${sleepTime/1000} seconds before GET'ing /$path... ***")
+      Thread sleep sleepTime
+
       val cluster = handleClusterResponse(parseResponse(getRequest(url + path)))
-      logger.info(s"GET /$path. Response: $cluster")
+      logger.info(s"Get cluster: GET /$path. Response: $cluster")
+
       cluster
     }
 
--- a/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/Leonardo.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/Leonardo.scala
@@ -101,6 +101,7 @@ object Leonardo extends RestClient with LazyLogging {
 
     def clusterPath(googleProject: GoogleProject, clusterName: ClusterName): String =
       s"api/cluster/${googleProject.value}/${clusterName.string}"
+//      s"api/cluster/v2/${googleProject.value}/${clusterName.string}"
 
     def list()(implicit token: AuthToken): Seq[Cluster] = {
       logger.info(s"Listing all active clusters: GET /api/clusters")
--- a/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/Leonardo.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/Leonardo.scala
@@ -139,8 +139,8 @@ object Leonardo extends RestClient with LazyLogging {
       val path = clusterPath(googleProject, clusterName)
 
       // TODO Find and fix the root-cause of why we get successive 404's without this sleep, and remove the sleep
-      val sleepTime = 7000
-      logger.info(s"Sleeping for ${sleepTime/1000} seconds before GET'ing /$path... ***")
+      val sleepTime = 2000
+      logger.info(s"Sleeping for ${sleepTime/2000} seconds before GET'ing /$path...")
       Thread sleep sleepTime
 
       val cluster = handleClusterResponse(parseResponse(getRequest(url + path)))
--- a/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/Leonardo.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/Leonardo.scala
@@ -139,7 +139,7 @@ object Leonardo extends RestClient with LazyLogging {
       val path = clusterPath(googleProject, clusterName)
 
       // TODO Find and fix the root-cause of why we get successive 404's without this sleep, and remove the sleep
-      val sleepTime = 2000
+      val sleepTime = 4000
       logger.info(s"Sleeping for ${sleepTime/2000} seconds before GET'ing /$path...")
       Thread sleep sleepTime
 
--- a/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoTestUtils.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoTestUtils.scala
@@ -134,8 +134,7 @@ trait LeonardoTestUtils extends WebBrowserSpec with Matchers with Eventually wit
     Thread sleep Random.nextInt(30000)
 
     val clusterTimeResult = time(Leonardo.cluster.create(googleProject, clusterName, clusterRequest))
-    //TODO Uncomment the following line when the response time is truly less than 10s
-    //clusterTimeResult.duration should be < tenSeconds
+    clusterTimeResult.duration should be < tenSeconds
     logger.info("Time to get cluster create response::" + clusterTimeResult.duration)
 
     // verify the create cluster response
--- a/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoTestUtils.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoTestUtils.scala
@@ -189,7 +189,7 @@ trait LeonardoTestUtils extends WebBrowserSpec with Matchers with Eventually wit
           logger.info(s"Saved Dataproc init log file for cluster ${creatingCluster.googleProject}/${creatingCluster.clusterName} to ${initLog.getAbsolutePath}")
           logger.info(s"Saved Dataproc startup log file for cluster ${creatingCluster.googleProject}/${creatingCluster.clusterName} to ${startupLog.getAbsolutePath}")
         case None =>
-          logger.warn(s"Could not obtain Dataproc log files for cluster ${creatingCluster.googleProject}/${creatingCluster.clusterName}"
+          logger.warn(s"Could not obtain Dataproc log files for cluster ${creatingCluster.googleProject}/${creatingCluster.clusterName}")
       }
 
       runningOrErroredCluster.get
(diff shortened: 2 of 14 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
```diff
@@ -242,24 +255,19 @@ trait ClusterComponent extends LeoComponent {
...
-
-    def setToStopping(id: Long): DBIO[Int] = {
-      updateClusterStatusAndHostIp(id, ClusterStatus.Stopping, None)
+    def updateAsyncClusterCreationFields(initBucket: Option[GcsPath],
+                                         serviceAccountKey: Option[ServiceAccountKey],
+                                         cluster: Cluster): DBIO[Int] = {
+      clusterQuery.filter { _.id === cluster.id }
+        .map(c => (c.initBucket, c.serviceAccountKeyId, c.googleId, c.operationName, c.stagingBucket, c.dateAccessed))
+        .update(initBucket.map(_.toUri), serviceAccountKey.map(_.id.value), cluster.googleId,
+          cluster.operationName.map(_.value), cluster.stagingBucket.map(_.value), Timestamp.from(Instant.now))
```
kyuksel: These are the `CLUSTER` table fields that are now filled in *after* `PUT /api/cluster/{googleProject}/{clusterName}` returns.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
```diff
@@ -224,6 +221,22 @@ trait ClusterComponent extends LeoComponent {
         .map { recs => recs.headOption.flatten.map(ServiceAccountKeyId) }
     }
```
kyuksel: Moved up the `get*` methods below so `get*/set*/etc.` methods are together for easier finding.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
```diff
@@ -271,14 +279,12 @@ trait ClusterComponent extends LeoComponent {
       }
     }
```
kyuksel: Moved down the set* methods below so get*/set*/etc. methods are together for easier finding.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -182,12 +167,17 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
     } flatMap {
       case Some(existingCluster) => throw ClusterAlreadyExistsException(googleProject, clusterName, existingCluster.status)
       case None =>
-        val augmentedClusterRequest = addClusterDefaultLabels(serviceAccountInfo, googleProject, clusterName, userEmail, clusterRequest)
+        val augmentedClusterRequest = addClusterLabels(serviceAccountInfo, googleProject, clusterName, userEmail, clusterRequest)
```
kyuksel: Renamed `addClusterDefaultLabels` since it does more than just adding default labels.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -182,12 +167,17 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
...
         val clusterFuture = for {
           // Notify the auth provider that the cluster has been created
           _ <- authProvider.notifyClusterCreated(userEmail, googleProject, clusterName)
+
```
kyuksel: It was necessary to separate the bucket validation from `createGoogleCluster()` since the v2 API needs to do it earlier in the flow (before returning the API request completion) or else a URI with too long a string can be attempted to be written into MySQL, which would error out after the request completion, which is not desirable.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -207,63 +197,115 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
...
+      augmentedClusterRequest, userEmail, clusterName, googleProject,
+      serviceAccountInfo, machineConfig, dataprocConfig.clusterUrlBase, autopauseThreshold)
+
+    // Validate that the Jupyter extension URIs and Jupyter user script URI are valid URIs and reference real GCS objects
+    // and if so, save the cluster creation request parameters in DB
+    val attemptToSaveClusterInDb: Future[Cluster] = validateClusterRequestBucketObjectUri(userEmail, googleProject, augmentedClusterRequest)
+      .flatMap { _ => dbRef.inTransaction(_.clusterQuery.save(initialCluster)) }
+
+    // For the success case, register the following callbacks...
+    attemptToSaveClusterInDb foreach { cluster =>
```
kyuksel: Should the `logger.info`s below be `logger.debug`? They were certainly useful during development but I'm not sure if they may constitute too much verbosity.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -529,7 +575,8 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
     } getOrElse Future.successful(())
   }
 
-  private def validateClusterRequestBucketObjectUri(userEmail: WorkbenchEmail, googleProject: GoogleProject, clusterRequest: ClusterRequest)(implicit executionContext: ExecutionContext): Future[Unit] = {
+  private def validateClusterRequestBucketObjectUri(userEmail: WorkbenchEmail, googleProject: GoogleProject, clusterRequest: ClusterRequest)
+                                                   (implicit executionContext: ExecutionContext): Future[Unit] = {
```
kyuksel: just stacking input parameter lists above and below for less horizontal scrolling

src/test/scala/org/broadinstitute/dsde/workbench/leonardo/api/LeoRoutesSpec.scala
```diff
@@ -51,21 +51,17 @@ class LeoRoutesSpec extends FlatSpec with ScalatestRouteTest with CommonTestData
...
-    val versionedBaseUrls = Seq(baseUrl, s"$baseUrl/v2")
-
-    // PUT endpoints have two versions
-    versionedBaseUrls.zip(clusterNames) foreach { case (url, cn) =>
-      Put(s"$url/${googleProject.value}/$cn", newCluster.toJson) ~>
-        timedLeoRoutes.route ~> check {
-          status shouldEqual StatusCodes.OK
-          validateCookie { header[`Set-Cookie`] }
+
+    forallClusterCreationVersions(clusterName) { (version, clstrName, statusCode) =>
```
kyuksel: Using `forallClusterCreationVersions` allowed for keeping the testing logic the same and prevented code duplication.

src/test/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoServiceSpec.scala
```diff
@@ -65,62 +68,70 @@ class LeonardoServiceSpec extends TestKit(ActorSystem("leonardotest")) with Flat
     super.afterAll()
   }
```
kyuksel: Just split the `initFiles` build-up into three `val`s for readibility

src/test/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoServiceSpec.scala
```diff
@@ -65,62 +68,70 @@ class LeonardoServiceSpec extends TestKit(ActorSystem("leonardotest")) with Flat
...
-    }
-
-    val dbInitBucketOpt = dbFutureValue { dataAccess =>
-      dataAccess.clusterQuery.getInitBucket(project, name1)
+  "LeonardoService" should "create a single node cluster with default machine configs" in isolatedDbTest {
+    forallClusterCreationMethods { (creationMethod, clusterName) =>
+      // create the cluster
+      val clusterCreateResponse = creationMethod(userInfo, project, clusterName, testClusterRequest).futureValue
+
+      eventually {
```
kyuksel: `eventually`s were needed since `v2` API response is no longer what it used to be and the asynchronous operation may not have been performed by the time the request is completed, causing some of the subsequent checks to fail.

src/test/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoServiceSpec.scala
```diff
@@ -130,38 +141,90 @@ class LeonardoServiceSpec extends TestKit(ActorSystem("leonardotest")) with Flat
     clusterCreateResponse shouldEqual clusterGetResponse
   }
 
+  it should "create and get a cluster via v2 API" in isolatedDbTest {
+    implicit val patienceConfig = PatienceConfig(timeout = 1.second)
```
kyuksel: For some reason I had to specify the `timeout` for this one as the default one wasn't long enough...perhaps because there are more checks than usual?  🤔

src/test/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoServiceSpec.scala
```diff
@@ -65,62 +68,70 @@ class LeonardoServiceSpec extends TestKit(ActorSystem("leonardotest")) with Flat
...
-    if (notebookServiceAccount(project).isDefined) {
-      iamDAO.serviceAccountKeys should contain key(samClient.serviceAccount)
-    } else {
-      iamDAO.serviceAccountKeys should not contain key(samClient.serviceAccount)
-    }
-
-    val dbInitBucketOpt = dbFutureValue { dataAccess =>
-      dataAccess.clusterQuery.getInitBucket(project, name1)
+  "LeonardoService" should "create a single node cluster with default machine configs" in isolatedDbTest {
+    forallClusterCreationMethods { (creationMethod, clusterName) =>
```
kyuksel: Same deal as above: `forallClusterCreationMethods` helped with preventing code duplication and keeping the test logic more or less the same...for most cases but not all.

src/test/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoServiceSpec.scala
```diff
@@ -335,6 +435,41 @@ class LeonardoServiceSpec extends TestKit(ActorSystem("leonardotest")) with Flat
     // check that the instances are still in the DB (they get removed by the ClusterMonitorActor)
     val instances = dbFutureValue { _.instanceQuery.getAllForCluster(getClusterId(clusterCreateResponse)) }
     instances.toSet shouldBe Set(masterInstance, workerInstance1, workerInstance2)
+
+    // ------------------ Grossly repeating above using v2 of the cluster creation API :( ------------------
```
kyuksel: To be able to use `forallClusterCreationMethods` here, the method would have to be bloated to account for distinct `master/workerInstance` between `v1` and v2` API calls, which didn't seem worth it.

src/test/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoServiceSpec.scala
```diff
@@ -605,4 +942,54 @@ class LeonardoServiceSpec extends TestKit(ActorSystem("leonardotest")) with Flat
...
+    // instance status should still be Stopped in the DB
+    // the ClusterMonitorActor is what updates instance status
+    val instances = dbFutureValue { _.instanceQuery.getAllForCluster(getClusterId(clusterCreateResponse)) }
+    instances.size shouldBe 3
+    instances.map(_.status).toSet shouldBe Set(InstanceStatus.Stopped)
+  }
+
+  type ClusterCreationInput = (UserInfo, GoogleProject, ClusterName, ClusterRequest)
+  type ClusterCreation = ClusterCreationInput => Future[Cluster]
+
```
kyuksel: Opted for overloading of `forallClusterCreationMethods` in lieu of default parameters so most callers wouldn't have to insert empty parameter lists (unless I'm missing a way that wouldn't require that).

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -207,63 +197,115 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
...
+    for {
+      _ <- authProvider.notifyClusterCreated(userEmail, cluster.googleProject, cluster.clusterName)
+      _ = logger.info(s"Successfully notified the AuthProvider for creation of cluster ${cluster.clusterName} " +
+        s"on Google project ${cluster.googleProject.value}. Proceeding to creating the cluster on Google Dataproc...")
+
+      (googleCluster, initBucket, serviceAccountKey) <- createGoogleCluster(userEmail, cluster, clusterRequest)
+    } yield {
+      logger.info(s"Successfully finished asynchronously creating the cluster ${cluster.clusterName} " +
+        s"on Google project ${cluster.googleProject.value}. Proceeding to updating the database cluster record...")
+      dbRef.inTransaction {
```
rtitle: This seems wrong to have a `Future` inside a `yield` block. If I'm reading correctly this method would return `Future[Future[Int]]` if it weren't being coerced to `Future[Unit]`. 

Better would be doing the DB update inside the for-comprehension:
```
for {
  _ <- authProvider.notifyClusterCreated(...)
  _ = logger.info("...{)
  _ <- (googleCluster, initBucket, serviceAccountKey) <- createGoogleCluster(...)
  _ <- dbRef.inTransaction { _.clusterQuery. updateAsyncClusterCreationFields(...) }
} yield ()
```

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -207,63 +197,115 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
...
+      logger.info(s"Attempting to asynchronously create cluster '$clusterName' " +
+        s"on Google project ${googleProject.value} and notify the AuthProvider")
+      completeClusterCreation(userEmail, cluster, augmentedClusterRequest)
+        .onComplete {
+          case Success(_) =>
+            logger.info(s"Asynchronous cluster creation succeeded for cluster '$clusterName' " +
+              s"on Google project '${googleProject.value}'. Notifying ClusterMonitorSupervisor about it...")
+            clusterMonitorSupervisor ! ClusterCreated(cluster, clusterRequest.stopAfterCreation.getOrElse(false))
+          case Failure(e) =>
+            logger.info(s"Failed the asynchronous portion of the creation of cluster '$clusterName' " +
```
rtitle: Might want to make this `logger.error` and log the actual exception: 
```
logger.error(s"Failed the asynchronous portion of the creation of cluster '$clusterName' " +
  s"on Google project '${googleProject.value}'.", e)
```

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
rtitle: Good question.. I'm inclined to leave these as INFO since it feels like they will be useful.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -207,63 +197,115 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
...
+              s"on Google project '${googleProject.value}'.")
+
+            // We also want to record the error in database for future reference.
+            val errorMessage = s"Asynchronous creation of cluster '$clusterName' on Google project" +
```
rtitle: It looks like you're missing a space after `project`

rtitle:
Early feedback above! TBH I didn't review the tests as closely as the app code but I didn't see any issues. I'll do another pass of everything before merge.

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoTestUtils.scala
```diff
@@ -121,55 +121,77 @@ trait LeonardoTestUtils extends WebBrowserSpec with Matchers with Eventually wit
...
-        None
-      }.futureValue match {
-        case Some((initLog, startupLog)) =>
+      saveDataprocLogFiles(creatingCluster).onComplete {
+        case Success(Some((initLog, startupLog))) =>
           logger.info(s"Saved Dataproc init log file for cluster ${creatingCluster.googleProject}/${creatingCluster.clusterName} to ${initLog.getAbsolutePath}")
           logger.info(s"Saved Dataproc startup log file for cluster ${creatingCluster.googleProject}/${creatingCluster.clusterName} to ${startupLog.getAbsolutePath}")
-        case None => logger.warn(s"Could not obtain Dataproc log files for cluster ${creatingCluster.googleProject}/${creatingCluster.clusterName}")
+        case Success(None) =>
+          logger.warn(s"Could not obtain Dataproc log files for cluster ${creatingCluster.googleProject}/${creatingCluster.clusterName}")
```
kyuksel: Refactored  the code block to add the `Failure` `case` so if the log files can't be saved, the test fails. Previously, the test would still pass.

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoTestUtils.scala
```diff
@@ -121,55 +121,77 @@ trait LeonardoTestUtils extends WebBrowserSpec with Matchers with Eventually wit
...
+        case Failure(e) =>
+          logger.error(s"Error occurred saving Dataproc log files for cluster ${creatingCluster.googleProject}/${creatingCluster.clusterName}", e)
+          throw e
```
kyuksel: Is there a better way of accomplishing the goal of...logging the error message before failing with the exception...than re-throwing it like that?

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoTestUtils.scala
```diff
@@ -636,10 +673,10 @@ trait LeonardoTestUtils extends WebBrowserSpec with Matchers with Eventually wit
     val transformed = for {
       stagingBucketName <- OptionT.fromOption[Future](cluster.stagingBucket)
       stagingBucketObjects <- OptionT.liftF[Future, List[GcsObjectName]](googleStorageDAO.listObjectsWithPrefix(stagingBucketName, "google-cloud-dataproc-metainfo"))
-      initLogFile <- OptionT.fromOption[Future](stagingBucketObjects.filter(_.value.endsWith("dataproc-initialization-script-0_output")).headOption)
+      initLogFile <- OptionT.fromOption[Future](stagingBucketObjects.find(_.value.endsWith("dataproc-initialization-script-0_output")))
```
kyuksel: No logic change above and below...just combined `filter` + `headOption` into `find` for conciseness.

src/main/resources/swagger/api-docs.yaml
```diff
@@ -200,6 +215,41 @@ paths:
             - openid
             - email
             - profile
```
kyuksel: No change to `get` below...just moved it below `put` so the two versions of `put` would be stacked.

src/main/resources/swagger/api-docs.yaml
```diff
@@ -623,7 +673,7 @@ definitions:
           The date and time the cluster was last accessed, in ISO-8601 format.
           Date accessed is defined as the last time the cluster was created, modified, or accessed via the proxy.
       autopauseThreshold:
-        type: int
```
kyuksel: This caused an error in [Swagger editor](http://editor.swagger.io). Not sure how it hadn't before.

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoTestUtils.scala
```diff
@@ -121,55 +121,77 @@ trait LeonardoTestUtils extends WebBrowserSpec with Matchers with Eventually wit
...
     Thread sleep Random.nextInt(30000)
 
-    val clusterTimeResult = time(Leonardo.cluster.create(googleProject, clusterName, clusterRequest))
-    //TODO Uncomment the following line when the response time is truly less than 10s
-    //clusterTimeResult.duration should be < tenSeconds
-    logger.info("Time to get cluster create response::" + clusterTimeResult.duration)
+    val clusterTimeResult = time(Leonardo.cluster.create(googleProject, clusterName, clusterRequest, apiVersion))
+    logger.info("Time to get cluster create response:" + clusterTimeResult.duration)
+    // TODO Uncomment out the following when the response time is indeed less than 10 seconds
+    // clusterTimeResult.duration should be < tenSeconds
```
kyuksel: The `sleep` in `Leonardo.get()` causes the assertion above to fail even though the response time is sub-second. There may be a way to make that assertion such that the sleep doesn't cause it to fail.

kyuksel:
jenkins retest (build failure)

kyuksel:
jenkins retest

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/Leonardo.scala
```diff
@@ -38,6 +39,14 @@ object Leonardo extends RestClient with LazyLogging {
   }
 
   object cluster {
+    private case class InvalidCreateClusterApiVersionException(googleProject: GoogleProject,
```
rtitle: I don't see any usages of `InvalidCreateClusterApiVersionException` in this PR? Is it needed? Can it be removed?

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/Leonardo.scala
```diff
@@ -109,20 +121,31 @@ object Leonardo extends RestClient with LazyLogging {
 
     def listIncludingDeleted()(implicit token: AuthToken): Seq[Cluster] = {
       val path = "api/clusters?includeDeleted=true"
-      // logger.info(s"Listing all clusters including deleted: GET /$path")
```
rtitle: Should we preserve this log message? We typically log in this class when a Leo API is called.

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/Leonardo.scala
```diff
@@ -109,20 +121,31 @@ object Leonardo extends RestClient with LazyLogging {
...
       logger.info(s"Create cluster: PUT /$path")
       handleClusterResponse(putRequest(url + path, clusterRequest))
     }
 
     def get(googleProject: GoogleProject, clusterName: ClusterName)(implicit token: AuthToken): Cluster = {
       val path = clusterPath(googleProject, clusterName)
+
+      // TODO Find and fix the root-cause of why we get successive 404's without this sleep, and remove the sleep
+      val sleepTime = 4000
+      logger.info(s"Sleeping for ${sleepTime/2000} seconds before GET'ing /$path...")
```
rtitle: Should divide by 1000 :)

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/Leonardo.scala
```diff
@@ -109,20 +121,31 @@ object Leonardo extends RestClient with LazyLogging {
...
+              (implicit token: AuthToken): Cluster = {
+      val path = clusterPath(googleProject, clusterName, version)
       logger.info(s"Create cluster: PUT /$path")
       handleClusterResponse(putRequest(url + path, clusterRequest))
     }
 
     def get(googleProject: GoogleProject, clusterName: ClusterName)(implicit token: AuthToken): Cluster = {
       val path = clusterPath(googleProject, clusterName)
+
+      // TODO Find and fix the root-cause of why we get successive 404's without this sleep, and remove the sleep
```
rtitle: It's strange to me that this is needed. Will think about this more...

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoTestUtils.scala
rtitle: It seems like you got rid of the `.futureValue`. This makes the method _not_ wait for the result of the `Future`. Therefore I don't think the `throw e` would actually cause the test to fail because it's being thrown asynchronously.

In general the test code runs synchronously, so all Futures are usually attached to ta `.futureValue`. It might be better to keep the original structure, but rethrow the exception inside the `recover` block instead of returning `None`. Something like:

```
future.recover { case e =>
  log.error(...)
  throw e
}.futureValue match {
  case Some(...) => log.info("success!")
  case None => log.warn("couldn't save log files but not failing test")
}
```

src/main/resources/swagger/api-docs.yaml
rtitle: Huh, good find. I usually validate swagger here: https://bigstickcarpet.com/swagger-parser/www/index.html. Maybe there is a difference between validators?

rtitle:
Left a few new comments above, mostly in auto-test code. In general this looks good. However I want to try to understand the 404 issue a little more before thumbing. I might try to do a little debugging...

kyuksel:
jenkins retest (build failure)

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/Leonardo.scala
kyuksel: Ah yea...upon making the return type of `ApiVersion.fromString()` optional, that became unnecessary.

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/Leonardo.scala
kyuksel: Makes sense. It had been commented out and I assumed it wasn't necessary.

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/Leonardo.scala
kyuksel: ugh..thanks

src/main/resources/swagger/api-docs.yaml
kyuksel: That one complains about `int` (vs `integer`) as well.

PR #521, opened 2018-08-06
Issue  499: Update Leonardo Content Security Policy

vkumra-broad:

In all cases:


src/main/resources/jupyter/init-actions.sh
```diff
@@ -214,6 +215,11 @@ if [[ "${ROLE}" == 'Master' ]]; then
       docker cp /etc/${JUPYTER_GOOGLE_SIGN_IN_JS} ${JUPYTER_SERVER_NAME}:${JUPYTER_USER_HOME}/.jupyter/custom/
     fi
 
+    if [ ! -z ${JUPYTER_NOTEBOOK_CONFIG} ] ; then
+      log 'Installing Google sign in extension...'
+      gsutil cp ${JUPYTER_NOTEBOOK_CONFIG} /etc
+      docker cp /etc/${JUPYTER_GOOGLE_SIGN_IN_JS} ${JUPYTER_SERVER_NAME}:$/etc/jupyter/
```
rtitle: Typo, `JUPYTER_GOOGLE_SIGN_IN_JS` instead of `JUPYTER_NOTEBOOK_CONFIG`. Probably should use the pattern above with `gsutil cp`, `basename`, then `docker cp`.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/Boot.scala
```diff
@@ -51,6 +51,7 @@ object Boot extends App with LazyLogging {
     val monitorConfig = config.as[MonitorConfig]("monitor")
     val samConfig = config.as[SamConfig]("sam")
     val autoFreezeConfig = config.as[AutoFreezeConfig]("autoFreeze")
+    val contenSecurityPolicy = config.as[String]("jupyterContentSecurityPolicy.contentSecurityPolicy")
```
rtitle: Typo, `content`

src/main/resources/jupyter/init-actions.sh
```diff
@@ -214,6 +215,11 @@ if [[ "${ROLE}" == 'Master' ]]; then
       docker cp /etc/${JUPYTER_GOOGLE_SIGN_IN_JS} ${JUPYTER_SERVER_NAME}:${JUPYTER_USER_HOME}/.jupyter/custom/
     fi
 
+    if [ ! -z ${JUPYTER_NOTEBOOK_CONFIG} ] ; then
+      log 'Installing Google sign in extension...'
+      gsutil cp ${JUPYTER_NOTEBOOK_CONFIG} /etc
+      docker cp /etc/${JUPYTER_NOTEBOOK_CONFIG} ${JUPYTER_SERVER_NAME}:${JUPYTER_HOME}/
```
rtitle: Probably still want to do `basename ${JUPYTER_NOTEBOOK_CONFIG}` since right now it will be substituted with a gs:// path.

src/main/resources/jupyter/jupyter_notebook_config.py
```diff
@@ -31,6 +31,6 @@
 c.NotebookApp.tornado_settings = {
     'static_url_prefix':'/notebooks/' + fragment + '/static/',
     'headers': {
-        'Content-Security-Policy': "frame-ancestors 'self' http://localhost:3000 https://bvdp-saturn-prod.appspot.com https://bvdp-saturn-dev.appspot.com; "
+        'Content-Security-Policy': {contentSecurityPolicy}
```
rtitle: Does `{string}` get templated? I thought our pattern was `$(string)`. Unless I'm misremembering...

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/Boot.scala
rtitle: Also maybe `Option[String]` with a default as we discussed

rtitle:
Looks good after a quick review; a few comments above.

vkumra-broad:
jenkins retest

src/main/resources/reference.conf
```diff
@@ -96,3 +97,7 @@ autoFreeze {
   autoFreezeAfter = 30 minutes
   autoFreezeCheckScheduler = 1 minute
 }
+
+jupyterContentSecurityPolicy {
+  contentSecurityPolicy = "frame-ancestors 'self' http://localhost:3000 https://bvdp-saturn-prod.appspot.com https://bvdp-saturn-dev.appspot.com; script-src 'self' 'unsafe-inline'  https://apis.google.com ; style-src 'self' 'unsafe-inline'; connect-src 'self' wss://*.broadinstitute.org:* *.googleapis.com"
```
rtitle: Can you add `https://localhost:443`? I think I had added that on one of the single-cell-portal clusters. We should also find out what the prod SCP hosts are, but we can always modify this config later.

vkumra-broad: They are using Leo in an iframe?

vkumra-broad:
jenkins retest

vkumra-broad:
jenkins retest

src/main/resources/reference.conf
```diff
@@ -96,3 +97,8 @@ autoFreeze {
   autoFreezeAfter = 30 minutes
   autoFreezeCheckScheduler = 1 minute
 }
+
+jupyterContentSecurityPolicy {
```
vkumra-broad: @adrazhi could you please review the CSP here, thanks!

src/main/resources/jupyter/init-actions.sh
```diff
@@ -60,6 +60,7 @@ if [[ "${ROLE}" == 'Master' ]]; then
     JUPYTER_CUSTOM_JS_URI=$(jupyterCustomJsUri)
     JUPYTER_GOOGLE_SIGN_IN_JS_URI=$(jupyterGoogleSignInJsUri)
     JUPYTER_USER_SCRIPT_URI=$(jupyterUserScriptUri)
+    JUPYTER_NOTEBOOK_CONFIG_URI=$(jupyterNotebookConfig)
```
rtitle: Minor, the pattern for other templated fields that are substituted with URIs are like: `$(jupyterNotebookConfigUri)

src/main/resources/jupyter/init-actions.sh
```diff
@@ -214,6 +215,12 @@ if [[ "${ROLE}" == 'Master' ]]; then
       docker cp /etc/${JUPYTER_GOOGLE_SIGN_IN_JS} ${JUPYTER_SERVER_NAME}:${JUPYTER_USER_HOME}/.jupyter/custom/
     fi
 
+    if [ ! -z ${JUPYTER_NOTEBOOK_CONFIG_URI} ] ; then
+      log 'Installing Google sign in extension...'
```
rtitle: Wrong log message

src/main/resources/reference.conf
```diff
@@ -96,3 +97,8 @@ autoFreeze {
...
+  //https://*.npmjs.org and 'unsafe-eval' needed for jupyterlab
```
rtitle: I think `#` is used for comments in configs

src/main/resources/reference.conf
rtitle: I wonder if we should name the config object `jupyterConfig`. Then we could access it like `"jupyterConfig.contentSecurityPolicy"`. I could envision us adding other jupyter configs here in the future.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/Boot.scala
```diff
@@ -51,6 +51,7 @@ object Boot extends App with LazyLogging {
     val monitorConfig = config.as[MonitorConfig]("monitor")
     val samConfig = config.as[SamConfig]("sam")
     val autoFreezeConfig = config.as[AutoFreezeConfig]("autoFreeze")
+    val contentSecurityPolicy = config.as[Option[String]]("jupyterContentSecurityPolicy.contentSecurityPolicy").getOrElse("default-src: 'self'")
```
rtitle: See comment above, would be nice to access this like `"jupyterConfig.contentSecurityPolicy"`.

src/main/resources/reference.conf
vkumra-broad: Agreed, that makes more sense

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
```diff
@@ -280,5 +284,5 @@ object LeonardoJsonSupport extends SprayJsonSupport with DefaultJsonProtocol {
 
   implicit val DefaultLabelsFormat = jsonFormat6(DefaultLabels.apply)
 
-  implicit val ClusterInitValuesFormat = jsonFormat20(ClusterInitValues.apply)
+  implicit val ClusterInitValuesFormat = jsonFormat22(ClusterInitValues.apply)
```
rtitle: yikes we're at 22 fields here too? :(

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -630,7 +632,10 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
       // Upload the googleSignInJs file to the bucket
       _ <- leoGoogleStorageDAO.storeObject(initBucketName, GcsObjectName(clusterResourcesConfig.jupyterGoogleSignInJs.value), googleSignInJsContent, "text/plain")
 
-      // Upload raw files (like certs) to the bucket
+      // Update the jupytyer notebook config file
+      _ <- leoGoogleStorageDAO.storeObject(initBucketName, GcsObjectName(clusterResourcesConfig.jupyterNotebookConfig.value), jupyterNotebookConfigContent, "text/plain")
+
+      // Upload raw files (like certs) to the bucxket
```
rtitle: typo in comment

src/test/scala/org/broadinstitute/dsde/workbench/leonardo/CommonTestData.scala
```diff
@@ -54,6 +54,7 @@ trait CommonTestData { this: ScalaFutures =>
   val clusterUrlBase = dataprocConfig.clusterUrlBase
   val serviceAccountsConfig = config.getConfig("serviceAccounts.config")
   val monitorConfig = config.as[MonitorConfig]("monitor")
+  val contentSecurityPolicy = config.as[String]("jupyterContentSecurityPolicy.contentSecurityPolicy")
```
rtitle: Would be good to use `Option[String]`, the same way we do in `Boot.scala`

vkumra-broad:
jenkins retest

rtitle:
Looks like you have some intermittent test failures. There's a current issue with dockerhub, I wonder if that's related.

jenkins retest

PR #539, opened 2018-08-20
Added getCluster method and updated a couple test files

akarukappadath:
A couple files have now been updated to reflect the addition of the getCluster method. PR is being opened for feedback. 

In all cases:


src/test/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponentSpec.scala
```diff
@@ -21,258 +21,120 @@ class ClusterComponentSpec extends TestComponent with FlatSpecLike with CommonTe
...
-      jupyterUserScriptUri = None,
-      errors = List.empty,
-      instances = Set(masterInstance, workerInstance1, workerInstance2),
-      userJupyterExtensionConfig = Some(userExtConfig),
-      autopauseThreshold = if (autopause) autopauseThreshold else 0,
-      defaultClientId = None
-    )
+    val cluster1 = getCluster(1).copy(dataprocInfo = getDataprocInfo(1).copy(googleId = Some(cluster1UUID)),
+                                      auditInfo = auditInfo.copy(createdDate = createdDate, dateAccessed = dateAccessed),
+                                      instances = Set(masterInstance, workerInstance1, workerInstance2))
```
kyuksel: It'd be nice to align dataprocInfo, auditInfo and instances. Same for the copy(...) usages below.

src/test/scala/org/broadinstitute/dsde/workbench/leonardo/db/LabelComponentSpec.scala
```diff
@@ -16,7 +16,8 @@ import scala.util.Random
 class LabelComponentSpec extends TestComponent with FlatSpecLike with CommonTestData with GcsPathUtils {
 
   "LabelComponent" should "save, get,and delete" in isolatedDbTest {
-    val c1 = Cluster(
+    val cluster1 =
```
kyuksel: Typically, code blocks line the one below is right-indented since they're the continuation of the line above. Same for the other instances.

src/test/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponentSpec.scala
```diff
@@ -21,258 +21,120 @@ class ClusterComponentSpec extends TestComponent with FlatSpecLike with CommonTe
...
-      labels = Map("bam" -> "yes", "vcf" -> "no"),
-      jupyterExtensionUri = None,
-      jupyterUserScriptUri = None,
-      errors = List.empty,
-      instances = Set(masterInstance, workerInstance1, workerInstance2),
-      userJupyterExtensionConfig = Some(userExtConfig),
-      autopauseThreshold = if (autopause) autopauseThreshold else 0,
-      defaultClientId = None
-    )
+    val cluster1 = getCluster(1).copy(dataprocInfo = getDataprocInfo(1).copy(googleId = Some(cluster1UUID)),
```
rtitle: Do you need to copy `dataprocInfo` and `auditInfo`? It seems like `getCluster(1)` would already initialize them correctly.

src/test/scala/org/broadinstitute/dsde/workbench/leonardo/dns/ClusterDnsCacheSpec.scala
```diff
@@ -32,7 +32,8 @@ class ClusterDnsCacheSpec extends TestKit(ActorSystem("leonardotest")) with Flat
     super.afterAll()
   }
 
-  val c1 = Cluster(
+  val cluster1 =
+  Cluster(
```
rtitle: Indentation looks off here

rtitle: Also should this be using `getCluster`?

src/test/scala/org/broadinstitute/dsde/workbench/leonardo/CommonTestData.scala
```diff
@@ -72,8 +70,36 @@ trait CommonTestData { this: ScalaFutures =>
...
+  val clusterServiceAccount = Option(WorkbenchEmail("testClusterServiceAccount@example.com"))
+  val notebookServiceAccount = Option(WorkbenchEmail("testNotebookServiceAccount@example.com"))
+  val serviceAccountInfo = new ServiceAccountInfo(clusterServiceAccount, notebookServiceAccount)
 
+  val auditInfo = AuditInfo(userEmail, Instant.now(), None, Instant.now())
+  def getDataprocInfo(index: Int): DataprocInfo = {
+    DataprocInfo(Option(UUID.randomUUID()), Option(OperationName("operationName" + index.toString)), Option(GcsBucketName("stagingBucketName" + index.toString)), Some(IP("numbers.and.dots")))
+  }
+
+  def getCluster(index: Int): Cluster = {
```
rtitle: Nitpick, but maybe `makeCluster`, `makeDataprocInfo`, etc would be better names since they're making new instances, not returning existing ones.

rtitle:
A few comments above, looks good overall. 

It'd be nice to not carry over all the commits for `defaultClientId` in this PR (since those commits were already squashed and merged as part of a different PR). You have 3 options:
1. On this branch, `git rebase -i origin/develop` and `squash`/`fixup` all the commits. You might need to resolve a lot of conflicts.
2. Make a new branch off of develop and `git cherry-pick` the commits related to the test updates (but not the defaultClientId) to the new branch.
3. Make a new branch off of develop and manually re-make the changes to the test files in the new branch.

src/test/scala/org/broadinstitute/dsde/workbench/leonardo/db/LabelComponentSpec.scala
akarukappadath: Gotcha. I think this is a weird IntelliJ thing that happened because I didn't get to update this file yet

vkumra-broad:
Could we save the cluster to the db as part of the `makeCluster`? That might solve the issues with the `dateAccessed` and `createdDate`. Would also help with code cleanup.

src/test/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponentSpec.scala
kyuksel: They were unaligned on my IntelliJ editor without doing anything. From my experience, IntelliJ tends to shuffle things around *once* you make modifications but not otherwise so I'm not sure why you may be having that issue.

kyuksel:
> Could we save the cluster to the db as part of the makeCluster? That might solve the issues with the dateAccessed and createdDate. 

What issues specifically?

vkumra-broad:
This one

```
dataprocInfo has to be copied so that cluster1 and cluster1WithErr have the same googleId, otherwise the test on line 66 fails.
I copied 'auditInfo' since makeCluster would cause createdDate and dateAccessed to be populated by Instant.now(), and I thought that if there was something that was causing some sort of lag between the two, the test on line 66 would fail. If you don't think that sort of lag is at all likely I can remove the copy!
```

Also, since the id field is AutoInc, returning the cluster after saving it to the DB would give us that field as well.

src/test/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponentSpec.scala
rtitle: You could also make `CommonTestData.auditInfo` a `def` to ensure you get a recent timestamp every time you access it.

rtitle:
We discussed in person, but I think having `makeCluster` and `saveCluster` as separate methods makes sense. One idea is to make an implicit class to add a `save` extension method to `Cluster`. So then we can do things like this:
```
makeCluster(5).copy(...).save
```
Alternatively, with a regular method it would look like this:
```
saveCluster(makeCluster(5).copy(...))
```

kyuksel:
```
makeCluster(5).copy(...).save
```
would read very nicely 👍

rtitle:
This can be closed in favor of https://github.com/DataBiosphere/leonardo/pull/543, right?

PR #543, opened 2018-08-24
Implemented makeCluster and save functions to reduce boilerplate in test code

akarukappadath:
This is a preliminary PR that contains a few test file updates that correspond to the new functions. 

In all cases:


akarukappadath:
jenkins retest

src/test/scala/org/broadinstitute/dsde/workbench/leonardo/CommonTestData.scala
```diff
@@ -12,18 +12,19 @@ import org.broadinstitute.dsde.workbench.leonardo.auth.sam.MockPetClusterService
...
 import org.scalatest.concurrent.ScalaFutures
 
 import scala.concurrent.ExecutionContext
-import scala.concurrent.duration._
+
 
 // values common to multiple tests, to reduce boilerplate
 
-trait CommonTestData { this: ScalaFutures =>
+trait CommonTestData extends TestComponent with GcsPathUtils{ this: ScalaFutures =>
```
rtitle: I am not sure `CommonTestData` should depend on `TestComponent`. I think those concerns should be separated: e.g. it should be possible for tests to use `CommonTestData` without using a mysql database. 

Simply moving `ClusterExtensions` to `TestComponent` is probably a good solution here.

rtitle:
Looks good, just 1 comment about `CommonTestData` vs `TestComponent`.

src/test/scala/org/broadinstitute/dsde/workbench/leonardo/CommonTestData.scala
rtitle: Hm, I think it'd be better for them to not depend on one another -- it's more modular that way. Ok to redefine `serviceAccountKey` if that's the only thing needed.

PR #550, opened 2018-08-28
always push images to GCR

jacmrob:

In all cases:

### DIFF
--- a/docker/build.sh
+++ b/docker/build.sh
@@ -15,8 +15,8 @@ HELP_TEXT="$(cat <<EOF
    jar : build Leonardo jar
    -d | --docker : (default: no action) provide either "build" or "push" to
            build or push a docker image.  "push" will also perform build.
-   -r | --registry: (default: dockerhub) can be either "dockerhub" or "gcr".
-           Users of gcr should have the gcloud tool installed and configured.
+   -dr | --dockerhub-registry: The dockerhub registry to push to
+   -gr | --gcr-registry: The GCR registry to push to
    -p | --project: set the project used at either dockerhub or with gcr
            container registries.
    -n | --notebook-repo: (default: --project) the repo to push the notebooks
@@ -77,10 +77,15 @@ while [ "$1" != "" ]; do
             RUN_DOCKER=true
             DOCKER_CMD="$1"
             ;;
-        -r | --registry)
+        -dr | --dockerhub-registry)
             shift
             echo "registry == $1"
-            DOCKER_REGISTRY=$1
+            DOCKERHUB_REGISTRY=$1
+            ;;
+        -gr | --gcr-registry)
+            shift
+            echo "gcr registry == $1"
+            GCR_REGISTRY=$1
             ;;
         -t | --tag)
             shift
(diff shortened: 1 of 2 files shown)


jenkins/jenkins_build.sh
```diff
@@ -10,8 +10,8 @@ if [ "$skip_docker_build" = false ]; then
        broadinstitute/dsde-toolbox:latest vault read --format=json ${GCR_SVCACCT_VAULT} \
        | jq .data > dspci-wb-gcr-service-account.json
 
-  ./docker/build.sh jar -d push -n "us.gcr.io/broad-dsp-gcr-public" \
-      -k "dspci-wb-gcr-service-account.json"
+  ./docker/build.sh jar -d push -dr "broadinstitute" -gr "gcr.io/broad-dsp-gcr-public" \
```
vkumra-broad: Is this pushing to gcr twice? with `-gr` and with `-n`

jacmrob: nope-- the `-n` flag is for the repo to push notebooks to

docker/build.sh
```diff
@@ -214,6 +206,10 @@ function docker_cmd()
             $DOCKER_REMOTES_BINARY tag $IMAGE:${DOCKER_TAG} $IMAGE:${GIT_BRANCH}
             $DOCKER_REMOTES_BINARY push $IMAGE:${GIT_BRANCH}
 
+            echo "pushing $GCR_IMAGE docker image..."
+            $DOCKER_REMOTES_BINARY tag $IMAGE:${DOCKER_TAG} ${GCR_IMAGE}:${DOCKER_TAG}
+            $GCR_REMOTES_BINARY push ${GCR_IMAGE}:${DOCKER_TAG}
```
rtitle: For dockerhub we push `<image>:<git-sha>` _and_ `<image>:<git-branch>`. For GCR we only push `<image>:<git-sha>`. Do we need to push a tag with the branch name in GCR as well?

jenkins/jenkins_build.sh
rtitle: The notebooks image is using `us.gcr.io`, while the leo one is using `gcr.io` (global load balancer). Is 1 preferred over the other? I remember not being sure when I did the notebooks repo in GCR.

docker/build.sh
```diff
@@ -123,26 +128,13 @@ if $PRINT_HELP; then
...
-  echo "The docker registry must be either 'dockerhub' or 'gcr'"
-  echo "Provided value: ${DOCKER_REGISTRY} is not allowed."
-  exit 1
-fi
-
+DOCKERHUB_REGISTRY=${DOCKERHUB_REGISTRY:-broadinstitute}
+GCR_REGISTRY=${GCR_REGISTRY:-gcr.io/broad-dsp-gcr-public}
+DOCKER_REMOTES_BINARY="docker"
+GCR_REMOTES_BINARY="gcloud docker --"
+NOTEBOOK_REPO="${NOTEBOOK_REPO:-$REPO}"
```
rtitle: I don't think `$REPO` is a variable anymore with the refactoring. Maybe this line could be changed to:
```
NOTEBOOK_REPO="${NOTEBOOK_REPO:-$GCR_REGISTRY}"
```
Then we wouldn't even have to specify an `-n` parameter in `jenkins_build.sh`.

docker/build.sh
```diff
@@ -15,8 +15,8 @@ HELP_TEXT="$(cat <<EOF
    jar : build Leonardo jar
    -d | --docker : (default: no action) provide either "build" or "push" to
            build or push a docker image.  "push" will also perform build.
-   -r | --registry: (default: dockerhub) can be either "dockerhub" or "gcr".
-           Users of gcr should have the gcloud tool installed and configured.
+   -dr | --dockerhub-registry: The dockerhub registry to push to
+   -gr | --gcr-registry: The GCR registry to push to
    -p | --project: set the project used at either dockerhub or with gcr
```
rtitle: I don't think `-p` and `$DOCKER_PROJECT` are used anymore now that the docker/gcr repos are specified explicitly. Can we remove it?

docker/build.sh
```diff
@@ -214,6 +206,10 @@ function docker_cmd()
             $DOCKER_REMOTES_BINARY tag $IMAGE:${DOCKER_TAG} $IMAGE:${GIT_BRANCH}
             $DOCKER_REMOTES_BINARY push $IMAGE:${GIT_BRANCH}
 
+            echo "pushing $GCR_IMAGE docker image..."
```
rtitle: Would it make sense to support either dockerhub, GCR, or both? I could see some users wanting to build Leo and only push to 1 place. Maybe could have some logic like:
```
if [ -n $IMAGE ]; then
  # push to dockerhub
fi
if [ -n $GCR_IMAGE ]; then
  # push to GCR
fi
```
This would mean not having defaults for the dockerhub/GCR registries which may make sense as we specify them in `jenkins_build.sh` anyway.

rtitle:
Some inline comments above. I really like these changes because it makes things simpler. It used to be that you pass a registry (dockerhub | gcr) to the build script and it tried to infer the repo based on an optional `--project` parameter or things like `gcloud config get-value project`. Now, you specify the dockerhub/gcr repos directly using `-dr` and `-gr` flags. This is cleaner, I think.

However I'd like to get @eap's eyes on it because it's a non-backwards compatible change. @eap could you switch your stuff to pass a GCR repo to the build script via `-gr`?

rtitle:
Another comment: I see "skip tests" in the title -- it'd be good to actually run tests on this PR because it would test the build -> push images -> pull images functionality.

rtitle (changes requested):
see above

docker/build.sh
jacmrob: we shouldn't, because as of now GCR is only for deploys, which use the hash-tagged images.  branch-tagged images are for fiabs, which still pull from dockerhub only.

docker/build.sh
jacmrob: would we still want to have that option though in case we want to push notebooks image to a different registry in the future?

docker/build.sh
jacmrob: yep

docker/build.sh
jacmrob: i was wondering that as i was making this change.  I can set up that functionality, i agree that it'd be useful.

jacmrob:
jenkins retest

jenkins/jenkins_build.sh
jacmrob: i think the default when you push to GCR us `gcr.io`, which is what i went with for all the service image pushes

rtitle (approved):
The latest changes look good to me. I'm happy to see passing automation tests. I also ping @eap on slack to see if he can take a look.

PR #553, opened 2018-08-31
Remove 10s createCluster check

rtitle:
See https://docs.google.com/document/d/1qrqI4V18YuKTaX1zsJ6YugKg7lwD0W9jmZPn-jF88RQ/edit?ts=5b87f266

This check is failing pretty often in auto-tests. I think auto-tests isn't really the right place to be checking this anyway. We should be measuring performance in other ways like instrumentation on production machines. There is a lot of slowness on a "clean" fiab which doesn't necessarily reflect reality. 

In all cases:


vkumra-broad:
Should we remove the time function as well?

https://github.com/DataBiosphere/leonardo/blob/c2cc2036a264c9f340e07cbe6288b52b83feb1db/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoTestUtils.scala#L686

rtitle:
Hm, well we're still using it to log the API response time, we're just not asserting that it's <10s anymore. I thought we could keep the logging for informational purposes.

PR #554, opened 2018-09-04
Correctly handle null stagingBucket in the Cluster response

rtitle:
I think this test failed because of this issue: https://fc-jenkins.dsp-techops.broadinstitute.org/job/leonardo-fiab-test-runner/4154/

Basically `stagingBucket` was being returned as `null` in `getCluster` calls (which is possible when using the V2 createCluster API). The code was then wrapping the `null` in a `Some`, giving us: Some(GcsBucketName(null))`. This caused failures around [here](https://github.com/DataBiosphere/leonardo/blob/develop/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoTestUtils.scala#L125-L130) because the code just checks if the option is defined.

I changed the JSON deserialization code to use `Option(..)` instead of `Some(..)` which should correctly handle nulls.

In all cases:


rtitle:
Just failed due to gpalloc issues which are being addressed -- going to merge this so it can be rebased onto some other PRs.

Issue #579, opened 2018-09-17
Persist request to stop a cluster upon creation


kyuksel:
To enable horizontal scaling, we're decoupling the cluster creation from the monitoring of it. Therefore, the only way for the back-Leo to know whether to monitor a cluster at any point will be by looking it up in the database. In the case of a cluster creation with the request to stop it thereafter, the flag set by the user at the request time is not persisted in the DB. Therefore, when the back-Leo discovers such a cluster that's being created. It will create a monitor for that process but it won't know to follow up with the stopping because we're currently not persisting the information to stop a cluster upon creation. This ticket is to change that.

PR #586, opened 2018-09-19
Persist stopAfterCreation flag (#579)

kyuksel:
In the API request `stopAfterCreation` flag is optional. Other than when it's specified as `Option(true)`, I mapped it to `false` (and not `Option(false)`) in the model and in the database for simplicity downstream, assuming that we don't need to distinguish between `stopAfterCreation` being `Option(false)` vs. `None`.

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
@@ -29,7 +29,8 @@ case class ClusterRecord(id: Long,
                          stagingBucket: Option[String],
                          dateAccessed: Timestamp,
                          autopauseThreshold: Int,
-                         defaultClientId: Option[String])
+                         defaultClientId: Option[String],
+                         stopAfterCreation: Boolean)
 
 case class MachineConfigRecord(numberOfWorkers: Int,
                                masterMachineType: String,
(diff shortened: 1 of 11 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
```diff
@@ -49,33 +50,34 @@ trait ClusterComponent extends LeoComponent {
...
+    def destroyedDate =  column[Timestamp]("destroyedDate", O.SqlType("TIMESTAMP(6)"))
+    def jupyterExtensionUri = column[Option[String]]("jupyterExtensionUri", O.Length(1024))
+    def jupyterUserScriptUri = column[Option[String]]("jupyterUserScriptUri", O.Length(1024))
+    def initBucket = column[Option[String]]("initBucket", O.Length(1024))
+    def serviceAccountKeyId= column[Option[String]]("serviceAccountKeyId", O.Length(254))
+    def stagingBucket = column[Option[String]]("stagingBucket", O.Length(254))
+    def dateAccessed= column[Timestamp]("dateAccessed", O.SqlType("TIMESTAMP(6)"))
+    def autopauseThreshold = column[Int]("autopauseThreshold")
+    def defaultClientId = column[Option[String]]("defaultClientId", O.Length(1024))
+    def stopAfterCreation = column[Boolean]("stopAfterCreation")
```
kyuksel: Line#80 is the only change above. Previously, the line segments were horribly rendered on my IntelliJ so I wanted to at least standardize them. GitHub turned out to display them nicely somehow.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
```diff
@@ -343,7 +345,8 @@ object LeonardoJsonSupport extends SprayJsonSupport with DefaultJsonProtocol {
         "userJupyterExtensionConfig" -> obj.userJupyterExtensionConfig.toJson,
         "dateAccessed" -> obj.auditInfo.dateAccessed.toJson,
         "autopauseThreshold" -> obj.autopauseThreshold.toJson,
-        "defaultClientId" -> obj.defaultClientId.toJson
+        "defaultClientId" -> obj.defaultClientId.toJson,
+        "stopAfterCreation" -> Option(obj.stopAfterCreation).toJson
```
kyuksel: I think I made a mistake above and it should've been

```scala
"stopAfterCreation" -> Option(obj.stopAfterCreation).toJson
```

I will manually test the API JSON serialization/deserialization, too.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
```diff
@@ -112,7 +112,8 @@ object Cluster {
       instances = Set.empty,
       userJupyterExtensionConfig = clusterRequest.userJupyterExtensionConfig,
       autopauseThreshold = autopauseThreshold,
-      defaultClientId = clusterRequest.defaultClientId)
+      defaultClientId = clusterRequest.defaultClientId,
+      stopAfterCreation= clusterRequest.stopAfterCreation.getOrElse(false))
```
kyuksel: i.e. map `stopAfterCreation` to false unless it was an optional `true`

PS. Will fix the typo after `stopAfterCreation`

kyuksel:
Manually tested that clusters created on FIAB with
1.  stopAfterCreation = true
2.  stopAfterCreation = false
3.  stopAfterCreation not specified in the API request body

result in the expected values being populated in the newly created `CLUSTER.stopAfterCreation` column as expected
1.  true
2.  false
3.  false

PR #588, opened 2018-09-20
[EARLY] Don't allow deleting a cluster in Creating status

rtitle:
This came up when Saturn tried to delete clusters seconds after they were created. This exposed a race condition in Leo code where the cluster deletion can step on async cluster creation. It's probably safer to just disallow delete when the cluster is in Creating status. FireCloud UI already enforces this, and Saturn will be doing similar. This PR enforces it at the API level.

I haven't ran automation tests yet, the changes are just a guess.

In all cases:


rtitle:
Doesn't work yet; closing for now.

PR #594, opened 2018-09-24
Don't continually monitor zombie clusters

rtitle:
Problem: if a "zombie" cluster starts getting monitored (for example at Leo startup) the `ClusterMonitorActor` goes into an infinite error loop. It logs errors like this:

```
Caused by: com.google.api.client.googleapis.json.GoogleJsonResponseException: 403 Forbidden
{
  "code" : 403,
  "errors" : [ {
    "domain" : "usageLimits",
    "message" : "Cloud Dataproc API has not been used in project 713110123530 before or it is disabled. Enable it by visiting https://console.developers.google.com/apis/api/dataproc.googleapis.com/overview?project=713110123530 then retry. If you enabled this API recently, wait a few minutes for the action to propagate to our systems and retry.",
    "reason" : "accessNotConfigured",
    "extendedHelp" : "https://console.developers.google.com"
  } ],
  "message" : "Cloud Dataproc API has not been used in project 713110123530 before or it is disabled. Enable it by visiting https://console.developers.google.com/apis/api/dataproc.googleapis.com/overview?project=713110123530 then retry. If you enabled this API recently, wait a few minutes for the action to propagate to our systems and retry.",
  "status" : "PERMISSION_DENIED"
}
	at com.google.api.client.googleapis.json.GoogleJsonResponseException.from(GoogleJsonResponseException.java:146)
	at com.google.api.client.googleapis.services.json.AbstractGoogleJsonClientRequest.newExceptionOnError(AbstractGoogleJsonClientRequest.java:113)
	at com.google.api.client.googleapis.services.json.AbstractGoogleJsonClientRequest.newExceptionOnError(AbstractGoogleJsonClientRequest.java:40)
	at com.google.api.client.googleapis.services.AbstractGoogleClientRequest$1.interceptResponse(AbstractGoogleClientRequest.java:321)
	at com.google.api.client.http.HttpRequest.execute(HttpRequest.java:1065)
	at com.google.api.client.googleapis.services.AbstractGoogleClientRequest.executeUnparsed(AbstractGoogleClientRequest.java:419)
	at com.google.api.client.googleapis.services.AbstractGoogleClientRequest.executeUnparsed(AbstractGoogleClientRequest.java:352)
	at org.broadinstitute.dsde.workbench.google.GoogleUtilities.$anonfun$executeGoogleCall$1(GoogleUtilities.scala:70)
	at scala.util.Try$.apply(Try.scala:209)
	at org.broadinstitute.dsde.workbench.google.GoogleUtilities.executeGoogleCall(GoogleUtilities.scala:70)
	at org.broadinstitute.dsde.workbench.google.GoogleUtilities.executeGoogleCall$(GoogleUtilities.scala:67)
	at org.broadinstitute.dsde.workbench.google.AbstractHttpGoogleDAO.executeGoogleCall(AbstractHttpGoogleDAO.scala:14)
	at org.broadinstitute.dsde.workbench.google.GoogleUtilities.executeGoogleRequest(GoogleUtilities.scala:51)
	at org.broadinstitute.dsde.workbench.google.GoogleUtilities.executeGoogleRequest$(GoogleUtilities.scala:50)
	at org.broadinstitute.dsde.workbench.google.AbstractHttpGoogleDAO.executeGoogleRequest(AbstractHttpGoogleDAO.scala:14)
	at org.broadinstitute.dsde.workbench.leonardo.dao.google.HttpGoogleDataprocDAO.$anonfun$getCluster$1(HttpGoogleDataprocDAO.scala:344)
```

The loop is not broken _even after_ the zombie cluster is detected and errored by `ZombieClusterMonitor`.

Solution: Have `ClusterMonitorActor` **abort** monitoring of clusters whose DB status have been updated to a terminal state.

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
@@ -255,14 +255,21 @@ class ClusterMonitorActor(val cluster: Cluster,
     } yield ShutdownActor()
   }
 
+  private def checkCluster: Future[ClusterMonitorMessage] = {
+    getDbClusterStatus.flatMap {
+      case status if status.isMonitored => checkClusterInGoogle(status)
+      case status =>
+        logger.info(s"Stopping monitoring of cluster ${cluster.projectNameString} in status ${status}")
+        Future.successful(ShutdownActor())
+    }
+  }
+
   /**
     * Queries Google for the cluster status and takes appropriate action depending on the result.
     * @return ClusterMonitorMessage
     */
-  private def checkCluster: Future[ClusterMonitorMessage] = {
+  private def checkClusterInGoogle(leoClusterStatus: ClusterStatus): Future[ClusterMonitorMessage] = {
     for {
-      clusterStatus <- getDbClusterStatus
-
       googleStatus <- gdDAO.getClusterStatus(cluster.googleProject, cluster.clusterName)
 
       googleInstances <- getClusterInstances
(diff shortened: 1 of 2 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
```diff
@@ -274,29 +281,29 @@ class ClusterMonitorActor(val cluster: Cluster,
         case Unknown | Creating | Updating =>
           Future.successful(NotReadyCluster(googleStatus, googleInstances))
         // Take care we don't restart a Deleting or Stopping cluster if google hasn't updated their status yet
-        case Running if clusterStatus != Deleting && clusterStatus != Stopping && clusterStatus != Starting && runningInstanceCount == googleInstances.size =>
+        case Running if leoClusterStatus != Deleting && leoClusterStatus != Stopping && leoClusterStatus != Starting && runningInstanceCount == googleInstances.size =>
```
rtitle: Just renamed var to `leoClusterStatus` for clarity.

src/test/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorSpec.scala
```diff
@@ -966,4 +969,38 @@ class ClusterMonitorSpec extends TestKit(ActorSystem("leonardotest")) with FlatS
...
+  // - monitor actor shuts down
+  // - dataproc DAO should not have been called
+  it should "stop monitoring a cluster in Error status" in isolatedDbTest {
+    val savedErrorCluster = errorCluster.save()
+    savedErrorCluster shouldEqual errorCluster
+
+    val gdDAO = mock[GoogleDataprocDAO]
+    when {
+      gdDAO.getClusterStatus(mockitoEq(creatingCluster.googleProject), mockitoEq(creatingCluster.clusterName))
+    } thenReturn Future.failed(new Exception)
```
kyuksel: Do we need the `when { }` block above given we `verify(gdDAO, never())...` and we don't care about the `gdDAO` response but instead that we never called `gdDAO` at the first place? If it's necessary, shouldn't it input `errorCluster...` instead of `creatingCluster...`?

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
kyuksel: Why was `clusterStatus` not clear?

src/test/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorSpec.scala
```diff
@@ -966,4 +969,38 @@ class ClusterMonitorSpec extends TestKit(ActorSystem("leonardotest")) with FlatS
...
     }
   }
+
+  // Pre:
+  // - cluster exists in the DB with status Error
+  // - dataproc DAO throws exception
+  // Post:
+  // - cluster is not updated in the DB
+  // - monitor actor shuts down
+  // - dataproc DAO should not have been called
```
kyuksel: nice checks and I love that you documented them like that, too!

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
rtitle: Because there's a `googleStatus` which comes from Google and a `leoClusterStatus` which comes from the Leo DB (both are of type `ClusterStatus`).

src/test/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorSpec.scala
rtitle: You're right it's probably not necessary since we assert that `gdDAO.getClusterStatus` never gets called (and yes, it should be `errorCluster`, nice catch :)).

I think I included it because that's what happens currently -- `gdDAO.getClusterStatus` throws an exception because the project is not active. But with this change it should not get called.

rtitle:
Test failed with `Error was not equal to Running` which is tracked in https://github.com/DataBiosphere/leonardo/issues/597

jenkins retest

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
kyuksel: I suspected so and then couldn't see a dedicated type from a quick look. Not for this PR but it'd be clearer to me to have a separate type (or subtype) for Google statuses.

Issue #597, opened 2018-09-25
Retries in init script don't work


rtitle:
`init-actions.sh` has a couple methods to retry calls to `apt-get update` and `docker-compose pull ...`:

https://github.com/DataBiosphere/leonardo/blob/develop/src/main/resources/jupyter/init-actions.sh

However these things don't actually get retried when they fail because we use `set -e` which tells bash to exit immediately if any command exits with non-zero status.

I think this pattern should work:
```
docker-compose pull ... || DOCKER_COMPOSE_RESULT=$?
# retry if DOCKER_COMPOSE_RESULT is != 0
```

Also: I wonder if we could write a generic retry function in bash `retry <some-cmd>` instead of having separate retry methods for different operations.

PR #598, opened 2018-09-25
Add a generic bash retry function which actually works

rtitle:
See https://github.com/DataBiosphere/leonardo/issues/597

I have not run automated tests yet, opening this PR to get test results.

In all cases:

### DIFF
--- a/src/main/resources/jupyter/init-actions.sh
+++ b/src/main/resources/jupyter/init-actions.sh
@@ -21,7 +21,6 @@ function retry {
   local retries=$1
   shift
 
-  local count=0
   for ((i = 1; i <= $retries; i++)); do
     # run with an 'or' so set -e doesn't abort the bash script on errors
     exit=0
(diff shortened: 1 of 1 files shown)


src/main/resources/jupyter/init-actions.sh
```diff
@@ -4,37 +4,38 @@ set -e -x
...
+# Retry 2/5 exited 1, retrying in 4 seconds...
+# Retry 3/5 exited 1, retrying in 8 seconds...
+# Retry 4/5 exited 1, retrying in 16 seconds...
+# Retry 5/5 exited 1, no more retries left.
+#
+function retry {
+  local retries=$1
+  shift
+
+  local count=0
```
aweng98: is `count` used? I don't find usage in `retry` function.

rtitle: Oops, leftover from when I used `while` instead of `for`

Issue #563, opened 2018-09-07
Refactor DNS caching


kyuksel:
Remove adding entries as soon as clusters are detected to transition into `Running` status. 
Instead, (front-) Leo `ClusterDnsCache`s read through to the database on cache misses. This shouldn’t cause too much of a performance overhead, and is significantly easier than implementing some kind of distributed cache.
	
Also see if `ClusterDnsCache` could be opportunistically reimplemented while at it, to 
just be a guava cache instead of an actor, as we use guava caches in several places 
now and it would fit this use case. If it’s too much extra work to include as part of 
this ticket, file a separate ticket for it.

kyuksel:
On further discussion, the team has decided to go with the Guava cache. Look at `SamAuthProvider.notebookAuthCache` for an example.

Preserve but refactor the existing DNSCache test.

kyuksel:
@vkumra-broad A thought occurred to me that we may want to factor into our performance benchmarking of the cache/non-caching scenarios. 

**Short:** With the one-minute caching, the worst-case performance of an API-call-intensive browser operation scales linearly with the number of front Leos we may have.

**Long:** In the worst-case scenario, the load balancer may hit different front Leo's for API-call-intensive operations. For instance, if we have 4 front Leos and the browser makes 25 proxy calls for notebook page loading, the load balancer may distribute those 25 proxy calls across all 4 front Leos. This would result in 4 database reads as opposed to one. Not sure how likely for the load balancer to do that but something to consider...

PR #612, opened 2018-10-01
Refactor DNS Cache for horizontal scaling

vkumra-broad:
- addresses [issue #563](https://github.com/DataBiosphere/leonardo/issues/563)
- see [Leo Horizontal Scaling Implementation Plan](https://docs.google.com/document/d/1KQ4BSEQvrvdArDVvr8dupO9t9g3ksql41FCONNh3Yeg) for context


vkumra-broad:
jenkins retest

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dns/ClusterDnsCache.scala
```diff
@@ -51,41 +40,29 @@ object ClusterDnsCache {
...
-      dataAccess.clusterQuery.listActive()
-    }.map(ProcessClusters.apply)
-  }
+class ClusterDnsCache(proxyConfig: ProxyConfig, dbRef: DbReference)(implicit  executionContext: ExecutionContext) extends LazyLogging {
+
+  val ProjectNameToHost = CacheBuilder.newBuilder()
+    .expireAfterWrite(5, TimeUnit.SECONDS)
+    .maximumSize(100)
+    .build(
+      new CacheLoader[(GoogleProject, ClusterName), Future[GetClusterResponse]] {
```
kyuksel: For readability, I'd suggest refactoring the cache key into a key case class similarly to in `SamAuthProvider.notebookAuthCache`:

```scala
private case class DnsCacheKey(googleProject: GoogleProject, clusterName: ClusterName)

new CacheLoader[DnsCacheKey, Future[GetClusterResponse]] {
```

This would have an added benefit of making code with direct tuple indices more readable; e.g. `getActiveClusterByName(key.googleProject, key.clusterName)` instead of `getActiveClusterByName(key._1, key._2)`.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dns/ClusterDnsCache.scala
```diff
@@ -21,16 +20,6 @@ object ClusterDnsCache {
...
-    Props(new ClusterDnsCache(proxyConfig, dbRef))
-
-  // Actor messages:
-  sealed trait ClusterDnsCacheMessage
-  case object RefreshFromDatabase extends ClusterDnsCacheMessage
-  case class ProcessClusters(clusters: Seq[Cluster]) extends ClusterDnsCacheMessage
-  case class ProcessReadyCluster(cluster: Cluster) extends ClusterDnsCacheMessage
-  case class GetByProjectAndName(googleProject: GoogleProject, clusterName: ClusterName) extends ClusterDnsCacheMessage
-
   // Responses to GetByProjectAndName message
```
kyuksel: Update the comment above since we don't have a `GetByProjectAndName` message anymore

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dns/ClusterDnsCache.scala
```diff
@@ -21,16 +20,6 @@ object ClusterDnsCache {
...
   sealed trait GetClusterResponse
```
kyuksel: This naming throws me off. I wonder if it'd be more intuitive to rename `GetClusterResponse` as `NotebookHostStatus` or `HostStatus`, and its descendants `NotebookHostNotFound` or `HostNotFound`, etc?

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dns/ClusterDnsCache.scala
```diff
@@ -21,16 +20,6 @@ object ClusterDnsCache {
 
   @volatile var HostToIp: Map[Host, IP] = Map.empty
```
kyuksel: I wonder if we can do away with `HostToIp` and store its contents directly in the Guava cache. We could define something like

```scala
case class DnsCacheValue(host: Host, ip: IP)
```

and then update `CacheLoader.load()` accordingly?

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dns/ClusterDnsCache.scala
```diff
@@ -51,41 +40,29 @@ object ClusterDnsCache {
...
+  val ProjectNameToHost = CacheBuilder.newBuilder()
+    .expireAfterWrite(5, TimeUnit.SECONDS)
+    .maximumSize(100)
+    .build(
+      new CacheLoader[(GoogleProject, ClusterName), Future[GetClusterResponse]] {
+        def load(key: (GoogleProject, ClusterName)) = {
+          dbRef.inTransaction {dataAccess => dataAccess.clusterQuery.getActiveClusterByName(key._1, key._2)}.map{
+            c => c match {
+              case  Some(cluster) => {
+                val res = projectNameToHostEntry(cluster)._2
```
kyuksel: If this is still relevant after refactoring based on the other comments, we may want to rewrite as something like

```scala
projectNameToHostEntry(cluster) match { 
  case (_, ClusterReady(host)) => ??? 
}
```

kyuksel (changes requested):

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dns/ClusterDnsCache.scala
```diff
@@ -51,41 +40,29 @@ object ClusterDnsCache {
...
-  private def queryForClusters: Future[ProcessClusters] ={
-    dbRef.inTransaction { dataAccess =>
-      dataAccess.clusterQuery.listActive()
-    }.map(ProcessClusters.apply)
-  }
+class ClusterDnsCache(proxyConfig: ProxyConfig, dbRef: DbReference)(implicit  executionContext: ExecutionContext) extends LazyLogging {
+
+  val ProjectNameToHost = CacheBuilder.newBuilder()
+    .expireAfterWrite(5, TimeUnit.SECONDS)
+    .maximumSize(100)
```
kyuksel: We should probably make ^ configurable, too.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dns/ClusterDnsCache.scala
```diff
@@ -51,41 +40,29 @@ object ClusterDnsCache {
...
-
-  private def queryForClusters: Future[ProcessClusters] ={
-    dbRef.inTransaction { dataAccess =>
-      dataAccess.clusterQuery.listActive()
-    }.map(ProcessClusters.apply)
-  }
+class ClusterDnsCache(proxyConfig: ProxyConfig, dbRef: DbReference)(implicit  executionContext: ExecutionContext) extends LazyLogging {
+
+  val ProjectNameToHost = CacheBuilder.newBuilder()
+    .expireAfterWrite(5, TimeUnit.SECONDS)
```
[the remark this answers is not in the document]
vkumra-broad: Agreed, I came up with that number considering  that when a notebook is opened most of the call would happen within that time period or less. It would also prevent the cache from having stale data. I was planning on using load test results to tweak that duration.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dns/ClusterDnsCache.scala
vkumra-broad: I am not sure if this would work, the JupyterNameService talks directly to the ClusterDnsCache object.

vkumra-broad:
jenkins retest

vkumra-broad:
jenkins retest

vkumra-broad:
jenkins retest

vkumra-broad:
jenkins retest

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dns/ClusterDnsCache.scala
kyuksel: Ah right. @rtitle, do you know if we have a way to get around that seeming limitation where `JupyterNameService` talks to `ClusterDnsCache` object implicitly via the descriptor. It would be great to not have to have that `var`.

src/test/resources/reference.conf
```diff
@@ -12,7 +12,7 @@ mysql {
   batchSize = 5000
   host = "localhost"
   //host = ${?MYSQL_HOST}
-  port = 3311
+  port = 3306
```
kyuksel: Why did the port number change?

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dns/ClusterDnsCache.scala
```diff
@@ -97,35 +77,13 @@ class ClusterDnsCache(proxyConfig: ProxyConfig, dbRef: DbReference) extends Acto
 
   private def hostToIpEntry(c: Cluster): (Host, IP) = host(c) -> c.dataprocInfo.hostIp.get
 
-  private def projectNameToHostEntry(c: Cluster): ((GoogleProject, ClusterName), GetClusterResponse) = {
+  private def projectNameToHostEntry(c: Cluster): ((GoogleProject, ClusterName), HostStatus) = {
```
kyuksel: We can use the newly created `DnsCacheKey` instead of `(GoogleProject, ClusterName)` so

```scala
  private def projectNameToHostEntry(c: Cluster): (DnsCacheKey, HostStatus) = ???
```

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dns/ClusterDnsCache.scala
kyuksel: I'd still recommend using pattern matching like ^^.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dns/ClusterDnsCache.scala
```diff
@@ -1,18 +1,19 @@
 package org.broadinstitute.dsde.workbench.leonardo.dns
 
-import akka.actor.{Actor, Props}
+import java.util.concurrent.TimeUnit
+
 import akka.http.scaladsl.model.Uri.Host
-import akka.pattern.pipe
+import akka.remote.WireFormats.FiniteDuration
```
kyuksel: Is the above `import` used?

kyuksel (changes requested):
Some more feedback. Getting close :) I'd like to still look at the test changes.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dns/ClusterDnsCache.scala
```diff
@@ -51,41 +43,29 @@ object ClusterDnsCache {
...
+    .expireAfterWrite(dnsCacheConfig.cacheExpiryTime.toSeconds, TimeUnit.SECONDS)
+    .maximumSize(dnsCacheConfig.cacheMaxSize)
+    .build(
+      new CacheLoader[DnsCacheKey, Future[HostStatus]] {
+        def load(key: DnsCacheKey) = {
+          dbRef.inTransaction {dataAccess => dataAccess.clusterQuery.getActiveClusterByName(key.googleProject, key.clusterName)}.map {
+              case  Some(cluster) => {
+                val res = projectNameToHostEntry(cluster)._2
+                if(res.isInstanceOf[HostReady]) {
+                  ClusterDnsCache.HostToIp = Map(hostToIpEntry(cluster))
```
kyuksel: Already talked to Vikram about this but for a reminder for myself...we're gonna want to add the new entry rather than overwriting the Map content.

kyuksel: I also wonder if we can write a unit test that would fail with the code line above and pass with the change?

src/test/scala/org/broadinstitute/dsde/workbench/leonardo/dns/ClusterDnsCacheSpec.scala
```diff
@@ -1,142 +1,50 @@
...
   override def afterAll(): Unit = {
-    TestKit.shutdownActorSystem(system)
     super.afterAll()
   }
 
-  val cluster1 = makeCluster(1).copy()
-
-  val cluster2 = makeCluster(2).copy(status = ClusterStatus.Creating,
-                                     dataprocInfo = makeDataprocInfo(2).copy(hostIp = None))
+  val cluster1: Cluster = makeCluster(1).copy()
```
kyuksel: `.copy()` above looks redundant since it's not overwriting any of the `Cluster` fields.

src/test/scala/org/broadinstitute/dsde/workbench/leonardo/dns/ClusterDnsCacheSpec.scala
```diff
@@ -1,142 +1,50 @@
...
-import java.time.Instant
-import java.util.UUID
-
-import akka.actor.ActorSystem
 import akka.http.scaladsl.model.Uri.Host
-import akka.pattern.ask
-import akka.testkit.{TestActorRef, TestKit}
-import akka.util.Timeout
 import org.broadinstitute.dsde.workbench.leonardo.ClusterEnrichments.clusterEq
 import org.broadinstitute.dsde.workbench.leonardo.{CommonTestData, GcsPathUtils}
```
kyuksel: `GcsPathUtils` unused

src/test/scala/org/broadinstitute/dsde/workbench/leonardo/dns/ClusterDnsCacheSpec.scala
```diff
@@ -1,142 +1,50 @@
...
 import org.broadinstitute.dsde.workbench.leonardo.db.{DbSingleton, TestComponent}
 import org.broadinstitute.dsde.workbench.leonardo.dns.ClusterDnsCache._
 import org.broadinstitute.dsde.workbench.leonardo.model._
 import org.broadinstitute.dsde.workbench.leonardo.model.google._
-import org.broadinstitute.dsde.workbench.model.google.GcsBucketName
 import org.scalatest.concurrent.{Eventually, ScalaFutures}
 import org.scalatest.{BeforeAndAfterAll, FlatSpecLike, Matchers}
```
kyuksel: `Matchers` unused?

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dns/ClusterDnsCache.scala
```diff
@@ -51,41 +43,29 @@ object ClusterDnsCache {
...
-  private def queryForClusters: Future[ProcessClusters] ={
-    dbRef.inTransaction { dataAccess =>
-      dataAccess.clusterQuery.listActive()
-    }.map(ProcessClusters.apply)
-  }
+class ClusterDnsCache(proxyConfig: ProxyConfig, dbRef: DbReference, dnsCacheConfig: ClusterDnsCacheConfig)(implicit executionContext: ExecutionContext) extends LazyLogging {
+
+
+
+  val projectNameToHost = CacheBuilder.newBuilder()
```
kyuksel: I wonder if `projectClusterToHost` would be more self-documenting.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dns/ClusterDnsCache.scala
kyuksel: Actually, `projectNameToHostEntry` is called from only within `projectNameToHost.load()` where only the second element (`._2`) of the output tuple is used. Therefore, I'd suggest changing the method as:

```scala
private def clusterToHostStatus(c: Cluster): HostStatus = {
  if (c.status.isStartable)
    HostPaused
  else if (c.dataprocInfo.hostIp.isDefined)
    HostReady(host(c))
  else
    HostNotReady
}
```

src/test/scala/org/broadinstitute/dsde/workbench/leonardo/dns/ClusterDnsCacheSpec.scala
```diff
@@ -1,142 +1,50 @@
...
 
     // save the clusters to the db
     cluster1.save() shouldEqual cluster1
     cluster2.save() shouldEqual cluster2
-    cluster3.save() shouldEqual cluster3
 
-    // maps should be populated
+    //We test the ProjectNameToHost Cache before the HostToIp map
+    //This replicates how the proxy accesses these maps as well.
+    //ProjectNameToHost read updates the HostToIP map
```
kyuksel: Let's make sure the references to `ProjectNameToHost` and `HostToIp` are accurate at the end, in terms of naming and upper/lower case initials.

src/test/scala/org/broadinstitute/dsde/workbench/leonardo/dns/ClusterDnsCacheSpec.scala
```diff
@@ -1,142 +1,50 @@
...
-
-    eventually {
-      (actorRef ? GetByProjectAndName(project, name1)).futureValue shouldBe ClusterReady(Host(s"${cluster1.dataprocInfo.googleId.get.toString}.jupyter.firecloud.org"))
-      (actorRef ? GetByProjectAndName(project, name2)).futureValue shouldBe ClusterNotReady
-      (actorRef ? GetByProjectAndName(project, newName1)).futureValue shouldBe ClusterReady(Host(s"${newCluster1.dataprocInfo.googleId.get.toString}.jupyter.firecloud.org"))
-      (actorRef ? GetByProjectAndName(project, newName2)).futureValue shouldBe ClusterReady(Host(s"${newCluster2.dataprocInfo.googleId.get.toString}.jupyter.firecloud.org"))
-    }
+    ClusterDnsCache.HostToIp.get(Host(s"${cluster1.dataprocInfo.googleId.get.toString}.jupyter.firecloud.org")) shouldBe cluster1.dataprocInfo.hostIp
 
+    ClusterDnsCache.HostToIp.get(Host(s"${cluster2.dataprocInfo.googleId.get.toString}.jupyter.firecloud.org")) shouldBe None
```
kyuksel: Can we add another test with a cluster that `isStartable` and verify that its `HostStatus` ends up being `HostPaused` and its `ClusterDnsCache` entry is `None`?

src/test/scala/org/broadinstitute/dsde/workbench/leonardo/dns/ClusterDnsCacheSpec.scala
```diff
@@ -1,142 +1,50 @@
...
   }
```
kyuksel: Is it possible to set the Guava cache's `cacheExpiryTime` low enough to also test that the cache is flushed after that much time as expected?

kyuksel (changes requested):
I think I'm through now with the review.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpJupyterDAO.scala
```diff
@@ -28,8 +30,8 @@ class HttpJupyterDAO(val clusterDnsCache: ActorRef)(implicit system: ActorSystem
     }
   }
 
-  protected def getTargetHost(googleProject: GoogleProject, clusterName: ClusterName): Future[GetClusterResponse] = {
+  protected def getTargetHost(googleProject: GoogleProject, clusterName: ClusterName): Future[HostStatus] = {
     implicit val timeout: Timeout = Timeout(5 seconds)
-    (clusterDnsCache ? GetByProjectAndName(googleProject, clusterName)).mapTo[GetClusterResponse]
+    clusterDnsCache.projectNameToHost(DnsCacheKey(googleProject, clusterName)).mapTo[HostStatus]
```
kyuksel: Getting the compilation warning
```
[warn] /Users/kyuksel/github/leonardo/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpJupyterDAO.scala:35:21: method apply in trait LoadingCache is deprecated: see corresponding Javadoc for more information.
[warn]     clusterDnsCache.projectNameToHost(DnsCacheKey(googleProject, clusterName)).mapTo[HostStatus]
[warn]                     ^
```
which can be fixed by using `get` instead of `apply`:
```scala
clusterDnsCache.projectNameToHost.get(DnsCacheKey(googleProject, clusterName)).mapTo[HostStatus]
```

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/ProxyService.scala
```diff
@@ -249,9 +250,9 @@ class ProxyService(proxyConfig: ProxyConfig,
   /**
     * Gets the notebook server hostname from the database given a google project and cluster name.
     */
-  protected def getTargetHost(googleProject: GoogleProject, clusterName: ClusterName): Future[GetClusterResponse] = {
+  protected def getTargetHost(googleProject: GoogleProject, clusterName: ClusterName): Future[HostStatus] = {
     implicit val timeout: Timeout = Timeout(5 seconds)
-    (clusterDnsCache ? GetByProjectAndName(googleProject, clusterName)).mapTo[GetClusterResponse]
+    clusterDnsCache.projectNameToHost(DnsCacheKey(googleProject, clusterName)).mapTo[HostStatus]
```
kyuksel: Getting the compilation warning
```
[warn] /Users/kyuksel/github/leonardo/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/ProxyService.scala:255:21: method apply in trait LoadingCache is deprecated: see corresponding Javadoc for more information.
[warn]     clusterDnsCache.projectNameToHost(DnsCacheKey(googleProject, clusterName)).mapTo[HostStatus]
[warn]                     ^
```
which can be fixed by using `get` instead of `apply` (which is called implicitly with the current code):
```scala
clusterDnsCache.projectNameToHost.get(DnsCacheKey(googleProject, clusterName)).mapTo[HostStatus]
```

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dns/ClusterDnsCache.scala
kyuksel: I've made [this PR](https://github.com/DataBiosphere/leonardo/pull/627) to this PR's branch to encapsulate the `volatile` and `synchronized` stuff so the client code is cleaner. We could also utilize it for other similar situations in the codebase.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
```diff
@@ -147,8 +146,6 @@ class ClusterMonitorActor(val cluster: Cluster,
       _ <- dbRef.inTransaction { dataAccess =>
         dataAccess.clusterQuery.setToRunning(cluster.id, publicIp)
       }
-      // Ensure the cluster is ready for proxying by updating the IP -> DNS cache
-      _ <- ensureClusterReadyForProxying(publicIp, clusterStatus)
```
kyuksel: Note to self to double-check that removing ^ doesn't introduce a regression.

kyuksel:
Closing until it's ready for re-review/testing

kyuksel (commented):
Changing my reviewer status to commenter

kyuksel (approved):

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
```diff
@@ -313,15 +310,6 @@ class ClusterMonitorActor(val cluster: Cluster,
...
   }
 
-  private def isProxyAvailable(clusterStatus: ClusterStatus, ip: IP): Future[Boolean] = {
-    for {
-      _ <- ensureClusterReadyForProxying(ip, clusterStatus)
-      proxyAvailable <- jupyterProxyDAO.getStatus(cluster.googleProject, cluster.clusterName)
-    } yield {
-      proxyAvailable
-    }
-  }
```
kyuksel: Removed the method above and renamed `JupyterProxyDAO.getStatus` as `isProxyAvailable` and started calling that here directly because

1. with the call `ensureClusterReadyForProxying(ip, clusterStatus)` no longer necessary, this method would've been down to the `jupyterProxyDAO.getStatus` call
2. `JupyterProxyDAO.getStatus` was not returning us a `status` per se but whether the proxy is available or not

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpJupyterDAO.scala
```diff
@@ -8,18 +8,20 @@ import akka.pattern.ask
...
 import scala.concurrent.duration._
 import scala.concurrent.{ExecutionContext, Future}
 
-class HttpJupyterDAO(val clusterDnsCache: ActorRef)(implicit system: ActorSystem, materializer: ActorMaterializer, executionContext: ExecutionContext) extends JupyterDAO with LazyLogging {
+class HttpJupyterDAO(val clusterDnsCache: ClusterDnsCache)(implicit system: ActorSystem, materializer: ActorMaterializer, executionContext: ExecutionContext) extends JupyterDAO with LazyLogging {
 
   val http = Http(system)
 
-  override def getStatus(googleProject: GoogleProject, clusterName: ClusterName): Future[Boolean] = {
+  override def isProxyAvailable(googleProject: GoogleProject, clusterName: ClusterName): Future[Boolean] = {
```
kyuksel: Removed the `isProxyAvailable` method that used to be in `ClusterMonitorActor` and renamed `JupyterProxyDAO.getStatus` as `isProxyAvailable` and started calling that here directly because

1. with the call `ensureClusterReadyForProxying(ip, clusterStatus)` no longer necessary, this method would've been down to the `jupyterProxyDAO.getStatus` call
2. `JupyterProxyDAO.getStatus` was not returning us a `status` per se but whether the proxy is available or not

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpJupyterDAO.scala
```diff
@@ -8,18 +8,20 @@ import akka.pattern.ask
...
     getTargetHost(googleProject, clusterName) flatMap {
-      case ClusterReady(targetHost) =>
+      case HostReady(targetHost) =>
```
kyuksel: Trying to be more consistent with the usage of `Host` in the Jupyter proxying context

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
kyuksel: We're no longer trying to cache IPs as soon as hosts become available upon cluster creation so `ensureClusterReadyForProxying` (whose name is a bit misleading) is no longer necessary.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ValueBox.scala
```diff
@@ -0,0 +1,84 @@
+package org.broadinstitute.dsde.workbench.leonardo.util
+
+/**
+  * A simple container to hold a value and make it safer to access and update it
+  * Copied from https://github.com/broadinstitute/dig-loam-stream/blob/master/src/main/scala/loamstream/util/ValueBox.scala
```
kyuksel: ^^ is a project I worked on where `ValueBox` was useful so I wanted to spread the love ❤️

src/test/scala/org/broadinstitute/dsde/workbench/leonardo/dns/ClusterDnsCacheSpec.scala
```diff
@@ -1,142 +1,72 @@
...
-  implicit val timeout = Timeout(5 seconds)
+class ClusterDnsCacheSpec extends FlatSpecLike with BeforeAndAfterAll with TestComponent with ScalaFutures with Eventually with CommonTestData {
 
   override def afterAll(): Unit = {
-    TestKit.shutdownActorSystem(system)
     super.afterAll()
   }
 
-  val cluster1 = makeCluster(1).copy()
+  val clusterBeingCreated: Cluster = makeCluster(2).copy(status = ClusterStatus.Creating,
```
kyuksel: Renamed the test clusters to make their purposes more explicit and set the above's status to `Creating` from the default `Unknown`

src/test/scala/org/broadinstitute/dsde/workbench/leonardo/dns/ClusterDnsCacheSpec.scala
```diff
@@ -1,142 +1,72 @@
...
-    (actorRef ? GetByProjectAndName(project, name2)).futureValue shouldBe ClusterNotFound
-    (actorRef ? GetByProjectAndName(project, name3)).futureValue shouldBe ClusterNotFound
+    // We test the projectClusterToHostStatus cache before the hostToIp map.
+    // This replicates how the proxy accesses these maps as well.
+    // projectClusterToHostStatus read updates the HostToIP map.
+    eventually { clusterDnsCache.cache.get(cacheKeyForClusterBeingCreated).futureValue shouldEqual HostNotReady }
+    eventually { clusterDnsCache.cache.get(cacheKeyForRunningCluster).futureValue shouldEqual HostReady(runningClusterHost) }
+    eventually { clusterDnsCache.cache.get(cacheKeyForStoppedCluster).futureValue shouldEqual HostPaused }
 
-    // add one cluster to cache immediately without waiting for the cycle
```
kyuksel: Using `CacheStats` (enabled at the time of cache creation via `recordStats` in `ClusterDnsCache.scala`) to make sure the expected number of cache loadings, misses and evictions happen throughout the test. I was motivated to do this particularly because of the `500` errors we saw recently that looked like a DNS cache. 

From what I read, enabling stats recording shouldn't cause a noticeable performance penalty.

kyuksel:
Comparing some manual performance testing on my branch FIAB to the prod version using the ASHG notebook, I have
- not noticed a difference between cell executions
- notebook loading takes ~3x longer (10.5s vs 3.5s) with the new changes, culprit call being 
```
/notebooks/{project}/{cluster}/api/contents/kyuksel-dnscache-testing/Large_Notebook_File_Example.ipynb?type=notebook&_=1541784790843
```

I'm in the process of creating a second FIAB with the dev version of Leo, to eliminate potential env/network-related confounders for the performance degradation.

kyuksel:
Following up on the previous note:

The performance comparison between the two FIABs (the one with the branch vs. the one with dev) did not yield a statistically significant difference (I was nerdy enough to run a t-test, which resulted in a p-value of .33 🤓) and the branch actually had a 1s lower mean. Therefore I feel comfortable enough to merge this PR (once we're done reviewing it).

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dns/ClusterDnsCache.scala
```diff
@@ -1,90 +1,79 @@
...
-
-  private def scheduleRefresh = {
-    context.system.scheduler.scheduleOnce(proxyConfig.dnsPollPeriod, self, RefreshFromDatabase)
-  }
+class ClusterDnsCache(proxyConfig: ProxyConfig, dbRef: DbReference, dnsCacheConfig: ClusterDnsCacheConfig)
+                     (implicit executionContext: ExecutionContext)
+  extends LazyLogging {
+
+  def cache: LoadingCache[DnsCacheKey, Future[HostStatus]] = projectClusterToHostStatus
+
```
kyuksel: Made `projectClusterToHostStatus` `private` and `def`ined getter for it so the cache isn't writable outside the class.

rtitle: But isn't it still be writable via the `cache` method? External callers could do `.cache.put(...)`.

Normally we've not exposed the cache directly at all, but defined a getter like:
```
def getHostStatus(googleProject: GoogleProject, clusterName: ClusterName): Future[HostStatus] = {
  projectClusterToHostStatus.get(DnsCacheKey(googleProject, clusterName)
}
```

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
```diff
@@ -292,9 +287,11 @@ class ClusterMonitorActor(val cluster: Cluster,
         case Running if leoClusterStatus == Starting && runningInstanceCount == googleInstances.size =>
           getMasterIp.flatMap {
             case Some(ip) =>
-              isProxyAvailable(leoClusterStatus, ip).map {
-                case true =>  ReadyCluster(ip, googleInstances)
-                case false => NotReadyCluster(ClusterStatus.Running, googleInstances)
+              dbRef.inTransaction { _.clusterQuery.updateClusterHostIp(cluster.id, Some(ip)) }.flatMap { _ =>
```
rtitle: Why do we need to update the ip in the database here? I think `clusterQuery.setToRunning` already does that.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ValueBox.scala
rtitle: Cool - maybe it could go in workbench-libs?

rtitle (approved):
Looks good, just a couple comments above. Good to know that performance didn't degrade with the small cache expiry time. I'm interested whether this resolves the issue we've been seeing with the current cache, where `Stopped` cluster status does not get reflected and leads to 500 errors to the caller.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ValueBox.scala
kyuksel: I agree. Will try to factor it out in a separate PR.

kyuksel:
@rtitle Thanks for the great feedback! I think I've addressed all but the workbench-libs one, which I'll target as a separate PR. Rerunning the integration tests. Did you want to take another look before I merge?

rtitle (approved):
Latest changes look good! :shipit: if tests pass

kyuksel:
jenkins retest

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
kyuksel: It turns out, we need to update the database because later on we rely on the database to get the host information to determine `ifProxyAvailable`, or else, we think that the host isn't available and get stuck checking the status. This happens when we resume clusters and the cluster transitions from `Starting` to `Running`.

PR #617, opened 2018-10-05
Quick and dirty fix for init bucket creation failures under heavy load

kyuksel:


src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/BucketHelper.scala
```diff
@@ -29,7 +32,9 @@ class BucketHelper(dataprocConfig: DataprocConfig,
                    googleComputeDAO: GoogleComputeDAO,
                    googleStorageDAO: GoogleStorageDAO,
                    serviceAccountProvider: ServiceAccountProvider)
-                  (implicit val executionContext: ExecutionContext) extends LazyLogging {
+                  (implicit val executionContext: ExecutionContext) extends LazyLogging with Retry {
+
+  val system = ActorSystem("BucketHelperActor")
```
kyuksel: Defining ^^ feels weird. I could add it as an `implicit` to `BucketHelper` constructor and see if it'd be provided by `Boot` or such...or is there a better way?

rtitle: Yeah I don't think we should define a new `ActorSystem` -- would be better to pass one into `BucketHelper`'s constructor.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/BucketHelper.scala
```diff
@@ -42,7 +47,9 @@ class BucketHelper(dataprocConfig: DataprocConfig,
       bucketSAs <- getBucketSAs(googleProject, serviceAccountInfo)
       leoEntity = userEntity(serviceAccountProvider.getLeoServiceAccountAndKey._1)
 
-      _ <- googleStorageDAO.createBucket(dataprocConfig.leoGoogleProject, bucketName, bucketSAs, List(leoEntity))
+      _ <- retryUntilSuccessOrTimeout(failureLogMessage = s"Init bucket creation failed for Google project '$googleProject'")(30 seconds, 5 minutes) { () =>
```
rtitle: Maybe add a comment here explaining the need for extra retry-ing

kyuksel:
jenkins retest

PR #623, opened 2018-10-10
List clusters by project

rtitle:
Adds a new Leo endpoint to list clusters belonging to a given Google project.

This is to address a performance issue: when the UI calls `listClusters` it is filtering clusters to the current google project, but internally Leo is selecting ALL non-deleted clusters from the DB and testing each one against Sam. With this new endpoint, Leo includes a `WHERE googleProject = ?` in the DB query, drastically reducing the number of records we need to process.

UI PR to use this endpoint is here: TBD

Tests pass, I'd like to do manual testing on a fiab next.

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/LeoRoutes.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/LeoRoutes.scala
@@ -113,11 +113,24 @@ abstract class LeoRoutes(val leonardoService: LeonardoService, val proxyService:
             }
           }
         } ~
-        path("clusters") {
+        pathPrefix("clusters") {
           parameterMap { params =>
-            complete {
-              leonardoService.listClusters(userInfo, params).map { clusters =>
-                StatusCodes.OK -> clusters
+            path(Segment) { googleProject =>
+              get {
+                complete {
+                  leonardoService.listClustersByGoogleProject(userInfo, GoogleProject(googleProject), params).map { clusters =>
+                    StatusCodes.OK -> clusters
+                  }
+                }
+              }
+            } ~
+            pathEndOrSingleSlash {
+              get {
+                complete {
+                  leonardoService.listClusters(userInfo, params).map { clusters =>
+                    StatusCodes.OK -> clusters
+                  }
+                }
               }
             }
           }
--- a/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
@@ -454,6 +454,17 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
     }
   }
 
+  def listClustersByGoogleProject(userInfo: UserInfo, googleProject: GoogleProject, params: LabelMap): Future[Seq[Cluster]] = {
+    for {
+      paramMap <- processListClustersParameters(params)
+      clusterList <- dbRef.inTransaction { da => da.clusterQuery.listByLabelsByProject(googleProject, paramMap._1, paramMap._2) }
+      visibleClusters <- authProvider.filterUserVisibleClusters(userInfo, clusterList.map(c => (c.googleProject, c.clusterName)).toList)
+    } yield {
+      val visibleClustersSet = visibleClusters.toSet
+      clusterList.filter(c => visibleClustersSet.contains((c.googleProject, c.clusterName)))
+    }
+  }
+
   private[service] def getActiveCluster(googleProject: GoogleProject, clusterName: ClusterName, dataAccess: DataAccess): DBIO[Cluster] = {
     dataAccess.clusterQuery.getActiveClusterByName(googleProject, clusterName) flatMap {
       case None => throw ClusterNotFoundException(googleProject, clusterName)
--- a/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/api/LeoRoutesSpec.scala
+++ b/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/api/LeoRoutesSpec.scala
@@ -222,6 +222,59 @@ class LeoRoutesSpec extends FlatSpec with ScalatestRouteTest with CommonTestData
     }
   }
 
+  it should "list clusters by project" in isolatedDbTest {
+    val newCluster = ClusterRequest(Map.empty, None)
+    val newProject = GoogleProject("new-project")
+
+    for (i <- 1 to 5) {
+      Put(s"/api/cluster/${googleProject.value}/${clusterName.value}-$i", newCluster.toJson) ~> leoRoutes.route ~> check {
+        status shouldEqual StatusCodes.OK
+      }
+    }
+
+    for (i <- 6 to 10) {
+      Put(s"/api/cluster/v2/${newProject.value}/${clusterName.value}-$i", newCluster.toJson) ~> leoRoutes.route ~> check {
+        status shouldEqual StatusCodes.Accepted
+      }
+    }
+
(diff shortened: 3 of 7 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/LeoRoutes.scala
```diff
@@ -113,11 +113,24 @@ abstract class LeoRoutes(val leonardoService: LeonardoService, val proxyService:
             }
           }
         } ~
-        path("clusters") {
+        pathPrefix("clusters") {
           parameterMap { params =>
-            complete {
```
rtitle: We weren't restricting to a `GET` before -- now we are.

rtitle:
jenkins retest

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/sam/SamAuthProvider.scala
```diff
@@ -207,6 +208,8 @@ class SamAuthProvider(val config: Config, serviceAccountProvider: ServiceAccount
       case e: ApiException if e.getCode == 401 => true
       // always invalidate and retry 500 errors
       case e: ApiException if e.getCode / 100 == 5 => true
+      // retry IOExceptions
+      case _: IOException => true
```
rtitle: This was a @dvoet suggestion

src/test/scala/org/broadinstitute/dsde/workbench/leonardo/api/LeoRoutesSpec.scala
```diff
@@ -222,6 +222,59 @@ class LeoRoutesSpec extends FlatSpec with ScalatestRouteTest with CommonTestData
     }
   }
 
+  it should "list clusters by project" in isolatedDbTest {
+    val newCluster = ClusterRequest(Map.empty, None)
+    val newProject = GoogleProject("new-project")
```
vkumra-broad: Nit: use commontestdata

src/test/scala/org/broadinstitute/dsde/workbench/leonardo/api/LeoRoutesSpec.scala
```diff
@@ -222,6 +222,59 @@ class LeoRoutesSpec extends FlatSpec with ScalatestRouteTest with CommonTestData
...
+        cluster.googleProject shouldEqual newProject
+        cluster.serviceAccountInfo.clusterServiceAccount shouldEqual clusterServiceAccount(newProject)
+        cluster.serviceAccountInfo.notebookServiceAccount shouldEqual notebookServiceAccount(newProject)
+        cluster.labels shouldEqual  Map(
+          "clusterName" -> cluster.clusterName.value,
+          "creator" -> "user1@example.com",
+          "googleProject" -> newProject.value) ++ serviceAccountLabels
+      }
+
+      validateCookie { header[`Set-Cookie`] }
```
vkumra-broad: Q: why we need the `validateCookie` test in both these cases?

src/test/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponentSpec.scala
```diff
@@ -181,4 +182,33 @@ class ClusterComponentSpec extends TestComponent with FlatSpecLike with CommonTe
     autoFreezeList should not contain stoppedCluster
     autoFreezeList should not contain runningCluster2
   }
+
+  it should "list by labels and project" in isolatedDbTest {
+    val savedCluster1 = makeCluster(1)
+      .copy(labels = Map("bam" -> "yes", "vcf" -> "no", "foo" -> "bar"))
+      .save(Some(serviceAccountKey.id))
+
+    val newProject = GoogleProject("new-project")
```
vkumra-broad: nit: use commontestdata

src/test/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoServiceSpec.scala
```diff
@@ -787,6 +787,23 @@ class LeonardoServiceSpec extends TestKit(ActorSystem("leonardotest")) with Flat
     }
   }
 
+  it should "list clusters belonging to a project" in isolatedDbTest {
+    // create a couple of clusters
+    val clusterName1 = ClusterName(s"cluster-${UUID.randomUUID.toString}")
```
vkumra-broad: nit: use commontestdata

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -454,6 +454,17 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
     }
   }
 
+  def listClustersByGoogleProject(userInfo: UserInfo, googleProject: GoogleProject, params: LabelMap): Future[Seq[Cluster]] = {
+    for {
+      paramMap <- processListClustersParameters(params)
+      clusterList <- dbRef.inTransaction { da => da.clusterQuery.listByLabelsByProject(googleProject, paramMap._1, paramMap._2) }
+      visibleClusters <- authProvider.filterUserVisibleClusters(userInfo, clusterList.map(c => (c.googleProject, c.clusterName)).toList)
```
vkumra-broad: we are duplicating most of the code from `listClusters`, wondering if we can refactor to reduce duplication.

vkumra-broad:
Overall looks good. Some nits and questions.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
rtitle: Yeah maybe.. will try.

src/test/scala/org/broadinstitute/dsde/workbench/leonardo/api/LeoRoutesSpec.scala
rtitle: Copy-paste.. but each call to the endpoint should return a `Set-Cookie` so I think the test is valid.

src/test/scala/org/broadinstitute/dsde/workbench/leonardo/api/LeoRoutesSpec.scala
```diff
@@ -222,6 +222,59 @@ class LeoRoutesSpec extends FlatSpec with ScalatestRouteTest with CommonTestData
...
+    }
+  }
+
```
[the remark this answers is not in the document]
rtitle: Yeah this is tested at every level - routes, service, and db. I didn't think there was much value in test cases for labels and `includeDeleted` at the routes level since it shares the same code.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
kyuksel: One option is to modify `def listClusters` to take in an optional GoogleProject:

```scala
def listClusters(userInfo: UserInfo, params: LabelMap, googleProject: Option[GoogleProject] = None)
```

and then call `listClusters` or `listClustersByGoogleProject` based on pattern matching. I think we could further refactor this by doing something similar with `listByLabels` where it also accepts an optional project and filter or not filter based on its value. That way `LeonardoService.listClusters` would just pass down the optional parameter.

rtitle: I did this (passing the `Option[GoogleProject]` downt he `ClusterComponent`)

rtitle:
jenkins retest

rtitle:
fiab-start failed :/

jenkins retest

rtitle:
jenkins retest

PR #626, opened 2018-10-12
Fix node security vulnerabilities

rtitle:
WARNING: I don't know what I'm doing!

This is to hopefully address the yellow warning on the Leo github page: https://github.com/DataBiosphere/leonardo

What I did:
1. ```npm audit fix --force```
2. committed the changes

The UI still loads on `npm start` so hopefully it is OK. (We don't actually run this UI in production, it was contributed externally.)

In all cases:

### DIFF
--- a/ui/package-lock.json
+++ b/ui/package-lock.json
@@ -4,13 +4,759 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
+    "@babel/code-frame": {
+      "version": "7.0.0",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.0.0.tgz",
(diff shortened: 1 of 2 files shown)


kyuksel:
I wonder if the owner of the said UI could verify that it still works with the changes. 

It may also be good to get feedback from the security team.

rtitle:
Good point. @eap and @adrazhi in case you have any feedback.

rtitle:
Ah no, that happened by itself by running `npm audit fix`. But that isn't flagged as a vulnerability -- maybe instead I will try running
```
npm audit fix --package-lock-only
```

rtitle:
I ran the above and it made the same change. I think I'm just going to merge this.

Issue #625, opened 2018-10-11
Increase dataproc init script timeout


rtitle:
The current default is 10 minutes. Users may run into this, especially if they specify a user script which is long running. We should increase it to 30 minutes.

PR #628, opened 2018-10-15
#625: Make Dataproc init script timeout configurable

MatthewBemis:
will go in after https://github.com/broadinstitute/firecloud-develop/pull/1393 and tests will fail until then

In all cases:

### DIFF
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -3,6 +3,7 @@
 dataproc {
   applicationName = "firecloud:leonardo"
   dataprocDockerImage = "gcr.io/broad-dsde-prod/leonardo-notebooks:prod"
+  defaultExecutionTimeout = "1800s"
   jupyterServerName = "jupyter-server"
   firewallRuleName = "leonardo-notebooks-rule"
   networkTag = "leonardo"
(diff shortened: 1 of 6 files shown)


vkumra-broad (approved):

src/main/resources/reference.conf
```diff
@@ -3,6 +3,7 @@
 dataproc {
   applicationName = "firecloud:leonardo"
   dataprocDockerImage = "gcr.io/broad-dsde-prod/leonardo-notebooks:prod"
+  defaultExecutionTimeout = "1800s"
```
rtitle: Typesafe config can parse:
```
defaultExecutionTimeout = 30 minutes
```
We use this pattern elsewhere in the config file.

rtitle (changes requested):
One change requested to represent it as a `FiniteDuration` in Leo code, and convert to a Google duration in `HttpGoogleDataprocDAO`.

MatthewBemis:
@rtitle ready for re-review. see the note i left atop `finiteDurationToGoogleDuration`. if you think it matters that it won't support fractional seconds, i can add that support

rtitle (approved):
Looks great! cc @abaumann FYI this PR ups the cluster initialization timeout to 30 mins by default.

Can we do a quick test to verify that this works?

1. Create a bash script which sleeps for ~15 minutes and add it to a bucket
2. Create a cluster, specifying the gs:// path to that script
3. Make sure the cluster creation succeeds

I think a manual test is fine in this case, as opposed to an automation test (which would increase test execution time).

rtitle:
PS - yeah I think second-granularity is fine.

Issue #536, opened 2018-08-15
Nuke buckets in broad-dsde-prod after cluster is done


davidbernick:
And clean up existant buckets. @vkumra-broad and @kyuksel know what I'm talking about. They can elaborate.

kyuksel:
@davidbernick Would you say that, not creating the `init-bucket`s in `broad-dsde-proj` at the first place is the ideal way to go but deleting those buckets after a cluster is removed and deleting the buckets orphaned by then are an acceptable solution?

rtitle:
I think I remember some context here.

We create these buckets in broad-dsde-* because they contain Leo secrets (certs, etc) which we don't want to be viewable by the user. Originally, we were deleting these buckets after the cluster was initialized. However, this caused issues with preemptibles -- since new preemptibles may come online at any time they need access to the init bucket even after the cluster is initialized. So we stopped deleting these buckets entirely. A reasonable solution (I'm not sure why this wasn't done) would be to delete these buckets when the cluster is _deleted_.

So I propose the following:
1. Run a script or something to clean up existing buckets for any deleted clusters
2. Change Leo code to delete init buckets when a cluster is deleted

I'm still a bit hesitant to put these buckets in users' billing projects because they contain secrets/certs which we only want viewable by Leo.

rtitle:
@MatthewBemis just following up, what's the status of cleaning up the old `leoinit` buckets? cc @kcibul

MatthewBemis:
i ran it in broad-dsde-dev and broad-dsde-alpha so far. it took a while, but cleaned up buckets for ~24,000 deleted clusters. i wanted to make sure i could run it in prod in one working day so it didn't get broken up.

i will prioritize running it in broad-dsde-prod tomorrow morning. then all of the associated cleanup should be complete aside from broad-dsde-qa but we may need to handle that differently since there's no persistent db to go off of

MatthewBemis:
this is done now

PR #631, opened 2018-10-16
#536: Delete cluster init buckets after cluster is deleted

MatthewBemis:
Cleanup script to come.

In all cases:

- [x] Get two thumbsworth of review and PO signoff if necessary

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
@@ -229,10 +229,11 @@ class ClusterMonitorActor(val cluster: Cluster,
     logger.info(s"Cluster ${cluster.projectNameString} has been deleted.")
 
     for {
+      // delete the init bucket so we don't continue to accrue costs after cluster is deleted
+      _ <- deleteInitBucket
+
       // delete instances in the DB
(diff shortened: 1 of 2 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
```diff
@@ -229,6 +229,9 @@ class ClusterMonitorActor(val cluster: Cluster,
     logger.info(s"Cluster ${cluster.projectNameString} has been deleted.")
 
     for {
+      // delete the init bucket so we don't continue to accrue costs after cluster is deleted
+      _ <- deleteInitBucket
```
MatthewBemis: how convenient that this code was left behind 😄

MatthewBemis:
jenkins retest because there were some build failures that appear to be out of our control (external deps not downloading)

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
rtitle: Heh, yeah it was a complete oversight that we missed deleting this bucket at cluster deletion time

rtitle (approved):
:shipit:

kyuksel (approved):
👍 Bernick will be happy.

Issue #596, opened 2018-09-25
Return 403 instead of 401 for authentication errors


kyuksel:
This came up with a cluster creation response:

```
Sep 25 09:07:46 gce-leonardo-prod201 leonardo-proxy[4233]: 130.211.1.38 - 112193493977220814199 "ryank@broadinstitute.org" "FireCloud" [25/Sep/2018:09:07:46 -0400] 158527 + "PUT /api/cluster/v2/nih-commons-team-calcium/ryank HTTP/1.1" 401 210 "https://portal.firecloud.org/" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/69.0.3497.100 Safari/537.36"
```

Also check if it may apply to any of the other responses.

http://www.dirv.me/blog/2011/07/18/understanding-403-forbidden/index.html

rtitle:
Note this involves an API change so we'd want to communicate this to stakeholders (AoU and Saturn teams).

PR #632, opened 2018-10-16
#596: Return 401 only when it's an authentication error

MatthewBemis:
~Did my best to go through and sort things out throughout the entire service. If you're reviewing this early, just know that I am going to do another pass on this but I think I got them all.~

Note that we also need to communicate this change to comms because this does result in some minor response code changes.

In all cases:

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -253,6 +253,8 @@ paths:
           description: Bad Request
           schema:
             $ref: '#/definitions/ErrorReport'
+        '403':
+            description: User does not have permission to perform action on cluster
         '409':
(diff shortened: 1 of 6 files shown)


src/main/resources/swagger/api-docs.yaml
```diff
@@ -424,7 +424,7 @@ paths:
       responses:
         '200':
           description: Proxy connection successful
-        '401':
+        '403':
```
rtitle: It's odd that the 401/403 is only documented on these `/notebooks/...` APIs. Maybe we should include them on the `/api/cluster/...` APIs since those can throw 403 as well.

rtitle (approved):
LGTM

PR #644, opened 2018-10-23
Fix null check for unset googleClientId

calbach:
Likely this was a regression with the addition of "defaultClientId". If a default is unspecified, `googleClientId` is now set to `""`. `"" == null` fails, so we never request the client ID from the parent frame.


rtitle (approved):
good find

calbach:
@rtitle I will leave this PR in your hands to merge as I don't want to assume to much about the details of your PR review / merge process.

PR #649, opened 2018-10-24
ignore non-deterministic test

rtitle:

In all cases:

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/ClusterConcurrencySpec.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/ClusterConcurrencySpec.scala
@@ -71,8 +71,13 @@ class ClusterConcurrencySpec extends FreeSpec with LeonardoTestUtils with Parall
       }
     }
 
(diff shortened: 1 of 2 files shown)


automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/ClusterConcurrencySpec.scala
```diff
@@ -71,8 +71,13 @@ class ClusterConcurrencySpec extends FreeSpec with LeonardoTestUtils with Parall
...
     }
 
+    // TODO: ignoring this test because it's not completely deterministic.
+    // Sometimes Google will fail to start a Stopping cluster.
+    // https://github.com/DataBiosphere/leonardo/issues/648 is open to
+    // revisit this behavior in Leo.
+    //
     // create -> wait -> stop -> no wait -> start -> delete
-    "should be able to start a stopping cluster" in {
+    "should be able to start a stopping cluster" ignore {
```
rtitle: I had already noticed flakiness in this test, as evidenced by the `Thread.sleep` below. I think a safer course is to just ignore it -- this is not critical functionality, and we may end up disallowing it in Leo.

PR #662, opened 2018-10-30
Separate user script/extension testing from NotebookInstallSpec

kyuksel:
While looking into the QA triage ticket [GAWB-3957](https://broadinstitute.atlassian.net/browse/GAWB-3957), we noticed that `NotebookInstallSpec` was performing the following:

1. creating a cluster X as part of `beforeAll` clause inherited from `ClusterFixtureSpec`
2. performing some operations on cluster X
3. creating a separate cluster Y
4. performing some operations on Y to test user script installation
5. creating another cluster Z
6. performing some operations on Z to test extension installation
7. going back and performing some more operations on the initially created cluster X

## Problems
1. The tests are run serially so when steps 3-6 take more than 30 mins, cluster X is auto-paused and step 7 results in an error like

```json
{"statusCode":422,"source":"leonardo", "message":"Cluster gpalloc-qa-master-ahmkmdk/automation-test-a2groajkz is stopped. Start your cluster before proceeding."} 
```

2. The tests also take longer than necessary due to the serial execution. 
3. Time out issues manifest themselves with confusing error signatures.

## Solution
This PR creates a new test called `NotebookCustomizationSpec` that includes the two tests from `NotebookInstallSpec` that create separate clusters. It doesn't extend `ClusterFixtureSpec` since we need to create the clusters in a customized manner. These tests should also run in parallel.


kyuksel:
jenkins retest

PR #665, opened 2018-11-01
Update run-jupyter-local.sh post templating changes

calbach:
This mostly fixed things.

Note: I still had problems with jupyter_notebook_config due to the change to `ip = "localhost"`. I was able to workaround this by reverting to '*' in combination with `allow_remote_access = True`. 127.0.0.1 also worked but I wasn't able to test that since I'm working remotely currently. Leaving that alone for now.

Context: I was using this to iterate on UI extension development. Probably could have used vanilla Jupyter just as easily but figured the Leo variant is slightly more realistic.

### DIFF
--- a/docker/jupyter/run-jupyter-local.sh
+++ b/docker/jupyter/run-jupyter-local.sh
@@ -1,10 +1,10 @@
 #!/bin/bash
 #
 # To rebuild the docker image, run:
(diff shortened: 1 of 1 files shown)


rtitle:
Rebased and fixed some paths. This seems to work for me, will merge once tests pass.

PR #668, opened 2018-11-06
Retries for "Jupyter kernel is not ready" error

akarukappadath:

In all cases:

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoTestUtils.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoTestUtils.scala
@@ -21,7 +21,7 @@ import org.broadinstitute.dsde.workbench.leonardo.Leonardo.ApiVersion.{V1, V2}
 import org.broadinstitute.dsde.workbench.leonardo.StringValueClass.LabelMap
 import org.broadinstitute.dsde.workbench.model.WorkbenchEmail
 import org.broadinstitute.dsde.workbench.model.google._
-import org.broadinstitute.dsde.workbench.util.LocalFileUtil
+import org.broadinstitute.dsde.workbench.util.{LocalFileUtil, Retry}
 import org.openqa.selenium.WebDriver
 import org.scalactic.source.Position
 import org.scalatest.{Matchers, Suite}
@@ -36,7 +36,7 @@ import scala.util.{Failure, Random, Success, Try}
 
 case class TimeResult[R](result:R, duration:FiniteDuration)
 
-trait LeonardoTestUtils extends WebBrowserSpec with Matchers with Eventually with LocalFileUtil with LazyLogging with ScalaFutures {
+trait LeonardoTestUtils extends WebBrowserSpec with Matchers with Eventually with LocalFileUtil with LazyLogging with ScalaFutures with Retry {
   this: Suite with BillingFixtures =>
 
   val logDir = new File("output")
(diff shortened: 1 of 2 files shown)


automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoTestUtils.scala
```diff
@@ -481,12 +481,15 @@ trait LeonardoTestUtils extends WebBrowserSpec with Matchers with Eventually wit
 
   def withNewNotebook[T](cluster: Cluster, kernel: Kernel = Python2, timeout: FiniteDuration = 2.minutes)(testCode: NotebookPage => T)(implicit webDriver: WebDriver, token: AuthToken): T = {
     withNotebooksListPage(cluster) { notebooksListPage =>
-      notebooksListPage.withNewNotebook(kernel, timeout) { notebookPage =>
-        testCode(notebookPage)
```
kyuksel: The retry timeout below is `5 minutes`. However, the enclosing `withNewNotebook`'s `timeout` parameter defaults to `2 minutes` (which tests don't typically override) which should make the retry timeout effectively `2 minutes` as well. If we need it to be `5 minutes`, we may want to update the enclosing one. If not, let's lower it to `2 minutes` so it's not misleading.

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoTestUtils.scala
```diff
@@ -481,12 +481,15 @@ trait LeonardoTestUtils extends WebBrowserSpec with Matchers with Eventually wit
...
+      retryUntilSuccessOrTimeout(failureLogMessage = s"Cannot make new notebook")(10 seconds, 5 minutes) {() =>
+        notebooksListPage.withNewNotebook(kernel, timeout) { notebookPage =>
+          testCode(notebookPage)
+        }
       }
     }
   }
 
+
```
kyuksel: Nit: remove the redundant line above

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoTestUtils.scala
rtitle: The `retryUntilSuccessOrTimeout` method is a little misleading. If you look at the code:
```
def retryUntilSuccessOrTimeout[T](pred: Predicate[Throwable] = always, failureLogMessage: String = defaultErrorMessage)(interval: FiniteDuration, timeout: FiniteDuration)(op: () => Future[T])(implicit executionContext: ExecutionContext): RetryableFuture[T] = {
  val trialCount = Math.ceil(timeout / interval).toInt
  retryInternal(Seq.fill(trialCount)(interval), pred, failureLogMessage)(op)
}
```
it doesn't actually enforce the `timeout` passed to it, it just uses the interval/timeout to determine the number of tries. So this code as written will try up to 4 times, sleeping 30 seconds between each try. Each try will take at most 2 minutes. So at most, this test will take 4*2m + (0.5m*3) = 9.5 minutes. (Hopefully I got that right..)

I think the current values are probably appropriate, just giving more details about how this works.

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoTestUtils.scala
```diff
@@ -481,8 +481,10 @@ trait LeonardoTestUtils extends WebBrowserSpec with Matchers with Eventually wit
 
   def withNewNotebook[T](cluster: Cluster, kernel: Kernel = Python2, timeout: FiniteDuration = 2.minutes)(testCode: NotebookPage => T)(implicit webDriver: WebDriver, token: AuthToken): T = {
     withNotebooksListPage(cluster) { notebooksListPage =>
-      notebooksListPage.withNewNotebook(kernel, timeout) { notebookPage =>
-        testCode(notebookPage)
+      retryUntilSuccessOrTimeout(failureLogMessage = s"Cannot make new notebook")(30 seconds, 2 minutes) {() =>
+        notebooksListPage.withNewNotebook(kernel, timeout) { notebookPage =>
```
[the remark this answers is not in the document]
kyuksel: Oh yea...thanks for catching that!

akarukappadath:
jenkins retest

Issue #496, opened 2018-07-18
Allow changing auto-pause threshold for Running clusters


rtitle:
Currently this is exposed at cluster creation time; we'd also like to expose it for Running clusters via a PATCH endpoint.

Coordinate with Saturn team when doing this work, as they will need to update the UI.

rtitle:
@bradtaylor if we implemented this would it be something that Saturn can make use of? I'm picturing a drop-down in the UI to change the auto-pause threshold, with values like "30 minutes", "4 hours", "1 day", "disabled", or "custom" (or something like that).

Issue #288, opened 2018-03-28
Add ability to add/remove worker nodes to a running cluster


rtitle:
Users want to add/remove workers on a running cluster instead of deleting/recreating. The Dataproc API supports this:

https://cloud.google.com/dataproc/docs/reference/rest/v1beta2/projects.regions.clusters/patch
https://cloud.google.com/dataproc/docs/concepts/configuring-clusters/scaling-clusters

Expose this functionality in Leo via a new PATCH endpoint.

PR #669, opened 2018-11-06
#288 + #496: PATCH cluster endpoint: Resize cluster and update autopause threshold

MatthewBemis:

In all cases:


src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
```diff
@@ -23,7 +23,7 @@ import org.broadinstitute.dsde.workbench.model._
 import spray.json._
 
 // Create cluster API request
-case class ClusterRequest(labels: LabelMap = Map(),
+case class ClusterRequest(labels: Option[LabelMap] = Option(Map()),
```
MatthewBemis: this was the only required field. making it optional allows me to use `ClusterRequest` for the PATCH payload and an empty `LabelMap` is already equivalent to leaving the field out

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -340,6 +343,62 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
...
   }
 
+  def updateCluster(userInfo: UserInfo,
+                    googleProject: GoogleProject,
+                    clusterName: ClusterName,
+                    clusterRequest: ClusterRequest): Future[Cluster] = {
+    for {
+      cluster <- internalGetActiveClusterDetails(googleProject, clusterName) //throws 404 if nonexistent
+
+      _ <- checkClusterPermission(userInfo, DeleteCluster, cluster) //throws 404 if no auth //TODO: check a more appropriate action here
```
MatthewBemis: calling out this TODO specifically. really seems like an UpdateCluster action could be useful here

MatthewBemis:
jenkins retest because sporadic build failure?

src/main/resources/swagger/api-docs.yaml
```diff
@@ -221,6 +221,51 @@ paths:
...
+          required: true
+          type: string
+        - in: body
+          description: Request for new cluster
+          name: clusterRequest
+          required: true
+          schema:
+            $ref: '#/definitions/UpdateClusterRequest'
+      responses:
+        '200':
```
MatthewBemis: i should probably make this 202 accepted because update _can_ be async

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
MatthewBemis: added modify action here: https://github.com/broadinstitute/sam/pull/245

src/main/resources/swagger/api-docs.yaml
```diff
@@ -221,6 +221,51 @@ paths:
...
+          description: googleProject
+          required: true
+          type: string
+        - in: path
+          name: clusterName
+          description: clusterName
+          required: true
+          type: string
+        - in: body
+          description: Request for new cluster
```
rtitle: Maybe:
```
description: Request for updated cluster
name: updateClusterRequest
```

src/main/resources/swagger/api-docs.yaml
rtitle: 👍

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
```diff
@@ -356,10 +356,13 @@ class ClusterMonitorActor(val cluster: Cluster,
...
       case Some(serviceAccountEmail) =>
         // Only remove the Dataproc Worker role if there are no other clusters with the same owner
-        // in the DB with CREATING status. This prevents situations where we prematurely yank pet SA
-        // roles when the same user is creating multiple clusters.
+        // in the DB with CREATING or UPDATING status. This prevents situations where we prematurely
+        // yank pet SA roles when the same user is creating or resizing multiple clusters.
         dbRef.inTransaction { dataAccess =>
-          dataAccess.clusterQuery.countByClusterServiceAccountAndStatus(serviceAccountEmail, ClusterStatus.Creating)
+          for {
+            creating <- dataAccess.clusterQuery.countByClusterServiceAccountAndStatus(serviceAccountEmail, ClusterStatus.Creating)
```
rtitle: Instead of running 2 queries what if we made `countByClusterServiceAccountAndStatuses` which takes a `Set` of statuses and uses a SQL `IN`? Then we could achieve the same thing with 1 query.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorSupervisor.scala
```diff
@@ -63,11 +65,15 @@ class ClusterMonitorSupervisor(monitorConfig: MonitorConfig, dataprocConfig: Dat
       logger.info(s"Monitoring cluster ${cluster.projectNameString} for deletion.")
       startClusterMonitorActor(cluster, if (recreate) Some(RecreateCluster(cluster)) else None)
 
+    case ClusterUpdated(cluster) =>
+      logger.info(s"Monitor cluster ${cluster.projectNameString} for updating.")
+      startClusterMonitorActor(cluster, None)
```
rtitle: That's awesome the `ClusterMonitorActor` works as is without too many significant changes.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
rtitle: 💯

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -340,6 +343,62 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
...
+  def handleResizingCluster(existingCluster: Cluster, machineConfigOpt: Option[MachineConfig]): Future[Unit] = {
+    machineConfigOpt match {
+      case Some(machineConfig) =>
+        for {
+          // Add Dataproc Worker role to the cluster service account, if present.
+          // This is needed to be able to spin up Dataproc clusters.
+          // If the Google Compute default service account is being used, this is not necessary.
+          _ <- addDataprocWorkerRoleToServiceAccount(existingCluster.googleProject, existingCluster.serviceAccountInfo.clusterServiceAccount)
+          _ <- gdDAO.resizeCluster(existingCluster.googleProject, existingCluster.clusterName, machineConfig.numberOfWorkers, machineConfig.numberOfPreemptibleWorkers)
+          _ <- dbRef.inTransaction { dataAccess => machineConfig.numberOfWorkers.map(numWorkers => dataAccess.clusterQuery.updateNumberOfWorkers(existingCluster.id, numWorkers)) getOrElse DBIO.successful(0) }
```
rtitle: I'm wondering it make sense to invert this, to avoid even starting a DB transaction if we don't need to.
```
          _ <- machineConfig.numberOfWorkers.map { numWorkers =>
            dbRef.inTransaction { _.clusterQuery.updateNumberOfWorkers(existingCluster.id, numWorkers) }
          } getOrElse Future.successful(0)
```

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -817,15 +876,15 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
     val combinedExtension = clusterRequest.userJupyterExtensionConfig.map(_.combinedExtensions).getOrElse(Map.empty)
 
     // combine default and given labels and add labels for extensions
-    val allLabels = clusterRequest.labels ++ defaultLabels ++ nbExtensions ++ serverExtensions ++ combinedExtension
+    val allLabels = clusterRequest.labels.getOrElse(Map()) ++ defaultLabels ++ nbExtensions ++ serverExtensions ++ combinedExtension
```
rtitle: Super nitpick -- I try to prefer `Map.empty` to `Map()`

src/test/scala/org/broadinstitute/dsde/workbench/leonardo/api/LeoRoutesSpec.scala
```diff
@@ -127,6 +127,42 @@ class LeoRoutesSpec extends FlatSpec with ScalatestRouteTest with CommonTestData
     }
   }
 
+  it should "202 when updating a running cluster" in isolatedDbTest{
```
rtitle: Test says 202 but asserts 200. :) I agree that 202 makes the most sense.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -340,6 +343,62 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
...
+
+  def handleResizingCluster(existingCluster: Cluster, machineConfigOpt: Option[MachineConfig]): Future[Unit] = {
+    machineConfigOpt match {
+      case Some(machineConfig) =>
+        for {
+          // Add Dataproc Worker role to the cluster service account, if present.
+          // This is needed to be able to spin up Dataproc clusters.
+          // If the Google Compute default service account is being used, this is not necessary.
+          _ <- addDataprocWorkerRoleToServiceAccount(existingCluster.googleProject, existingCluster.serviceAccountInfo.clusterServiceAccount)
+          _ <- gdDAO.resizeCluster(existingCluster.googleProject, existingCluster.clusterName, machineConfig.numberOfWorkers, machineConfig.numberOfPreemptibleWorkers)
```
rtitle: Cool - so the existing `HttpGoogleDataprocDAO.resizeCluster` works well as is?

src/main/resources/swagger/api-docs.yaml
```diff
@@ -221,6 +221,51 @@ paths:
             - profile
 
   '/api/cluster/{googleProject}/{clusterName}':
+    patch:
```
rtitle: Maybe not for this PR but I'm wondering if we should consider retroactively versioning all our endpoints. E.g. `PATCH /api/cluster/v1/{googleProject/{clusterName}`. We have a v2 endpoint but no v1s. Maybe as a separate PR we could do a one time migration to make all endpoints versioned.

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/ClusterConcurrencySpec.scala
```diff
@@ -138,6 +138,43 @@ class ClusterConcurrencySpec extends FreeSpec with LeonardoTestUtils with Parall
       }
 
     }
+
+    // make sure adding a worker works
```
rtitle: Good test but I'm not sure it belongs in `ClusterConcurrencySpec`. Maybe we could put it in `ClusterMonitoringSpec`.

rtitle (changes requested):
Requesting changes for a few suggestions above, but it's all super minor stuff. In general this looks great, nice work!

Looks like leonardo_build failed with a docker build error... maybe transient? Probably worth a jenkins retest before digging too much.

rtitle:
Apparently jenkins retest doesn't pick up reviews, only comments

src/main/resources/swagger/api-docs.yaml
kyuksel: Wouldn't retroactive versioning break clients, which defeats the purpose of versioning for backwards-compatibility, or did you imply the non-versioned endpoints would alias to v1? 

If we'll version our endpoints going forward, it'd be start versioning this one, too while at it.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -340,6 +343,62 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
...
+
+        updatedCluster <- internalGetActiveClusterDetails(existingCluster.googleProject, existingCluster.clusterName)
+      } yield {
+        clusterMonitorSupervisor ! ClusterUpdated(updatedCluster.copy(status = ClusterStatus.Updating))
+        updatedCluster
+      }
+    } else Future.failed(ClusterCannotBeUpdatedException(existingCluster.googleProject, existingCluster.clusterName, existingCluster.status))
+  }
+
+  def handleUpdatingAutopauseThreshold(clusterId: Long, autopause: Option[Boolean], autopauseThreshold: Option[Int]): Future[Int] = {
```
kyuksel: Nit: would it be sufficient to name it just `updateAutopauseThreshold`? Same for `resizeCluster` below.

src/test/scala/org/broadinstitute/dsde/workbench/leonardo/dao/google/MockGoogleDataprocDAO.scala
```diff
@@ -93,6 +97,22 @@ class MockGoogleDataprocDAO(ok: Boolean = true) extends GoogleDataprocDAO {
...
+      val secondaryWorkerInstances = numPreemptibles.map(num => List.tabulate(num) { i => InstanceKey(googleProject, ZoneUri("my-zone"), InstanceName(s"secondary-worker-instance-$i")) }.toSet).getOrElse(Set.empty)
+      val existingWorkerInstances = instances(clusterName)(Worker)
+      val existingMasterInstance = instances(clusterName)(Master)
+
+      instances += (clusterName -> mutable.Map(Master -> existingMasterInstance, Worker -> existingWorkerInstances, SecondaryWorker -> secondaryWorkerInstances))
+    }
+
     Future.successful(())
   }
 }
```
kyuksel: Nit: add new line

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
kyuksel: Consider adding logging for the resizing process to ease debug. See v2 cluster creation methods for examples.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -340,6 +343,62 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
...
+      _ <- checkClusterPermission(userInfo, DeleteCluster, cluster) //throws 404 if no auth //TODO: check a more appropriate action here
+
+      _ <- internalUpdateCluster(cluster, clusterRequest)
+
+      updatedCluster <- internalGetActiveClusterDetails(googleProject, clusterName)
+    } yield { updatedCluster }
+  }
+
+  def internalUpdateCluster(existingCluster: Cluster, clusterRequest: ClusterRequest) = {
+    if(existingCluster.status.isUpdatable) {
```
kyuksel: We probably need another conditional for the case where we only update the auto-pause threshold since we wouldn't need to message `ClusterMonitorSupervisor` in that case.

src/test/scala/org/broadinstitute/dsde/workbench/leonardo/dao/google/MockGoogleDataprocDAO.scala
```diff
@@ -93,6 +97,22 @@ class MockGoogleDataprocDAO(ok: Boolean = true) extends GoogleDataprocDAO {
   }
 
   override def resizeCluster(googleProject: GoogleProject, clusterName: ClusterName, numWorkers: Option[Int], numPreemptibles: Option[Int]): Future[Unit] = {
```
kyuksel: Should we do any validation on the value of `numWorkers`? What happens if it's a negative number for instance?

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
kyuksel: We should also update the corresponding test to verify the status of the cluster doesn't change for that case.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -340,6 +343,62 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
...
+  }
+
+  def handleResizingCluster(existingCluster: Cluster, machineConfigOpt: Option[MachineConfig]): Future[Unit] = {
+    machineConfigOpt match {
+      case Some(machineConfig) =>
+        for {
+          // Add Dataproc Worker role to the cluster service account, if present.
+          // This is needed to be able to spin up Dataproc clusters.
+          // If the Google Compute default service account is being used, this is not necessary.
+          _ <- addDataprocWorkerRoleToServiceAccount(existingCluster.googleProject, existingCluster.serviceAccountInfo.clusterServiceAccount)
```
kyuksel: Do we have tests for when `addDataprocWorkerRoleToServiceAccount` and `gdDAO.resizeCluster` calls may fail?

src/main/resources/swagger/api-docs.yaml
MatthewBemis: oops, leftover from copy-paste :)

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
rtitle: +1 to this -- if just updating auto pause stuff we probably don't need to change the status to `Updating` and call `ClusterMonitorSupervisor`.

src/main/resources/swagger/api-docs.yaml
MatthewBemis: aliasing non-versioned routes to v1 is how we do it in sam. i agree that we should do this in a follow-up PR

kyuksel: To be clear, what I'd suggested doing now was to make the endpoint route `/api/cluster/v1/{googleProject/{clusterName}` to begin with, to save the aliasing later on.

MatthewBemis: if we do it like how it was done in sam, we can blanket-alias across all of the routes so it would be more work to version this endpoint now for little gain, IMO

kyuksel: oh okay...I agree then

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -57,6 +57,9 @@ case class ClusterCannotBeDeletedException(googleProject: GoogleProject, cluster
 case class ClusterCannotBeStartedException(googleProject: GoogleProject, clusterName: ClusterName, status: ClusterStatus)
   extends LeoException(s"Cluster ${googleProject.value}/${clusterName.value} cannot be started in ${status.toString} status", StatusCodes.Conflict)
 
+case class ClusterCannotBeUpdatedException(googleProject: GoogleProject, clusterName: ClusterName, status: ClusterStatus)
+  extends LeoException(s"Cluster ${googleProject.value}/${clusterName.value} cannot be updated in ${status.toString} status", StatusCodes.Conflict)
+
```
kyuksel: Not blocker: We could simplify it as:

```scala
case class ClusterCannotBeUpdatedException(cluster: Cluster)
   extends LeoException(s"Cluster ${cluster.projectNameString} cannot be updated when in ${cluster.status}")
```

src/test/scala/org/broadinstitute/dsde/workbench/leonardo/dao/google/MockGoogleDataprocDAO.scala
MatthewBemis: yeah that would be a good idea. there's some strange criteria for the number of workers allowed so i should capture that in a validation step so we can fail fast

MatthewBemis: i played around with this a bit and i think im going to leave the validation to google so we don't duplicate their logic. they're able to fail fast if an invalid number of workers is specified so it's essentially the same UX

MatthewBemis: the call to edit the machine config of the cluster in google isn't asynchronous, it happens and completes when the user calls the PATCH endpoint. the async part is the actual provisioning/decommissioning of the worker nodes. because google does the validation upon editing the machine config, we can fail fast and immediately return their error message to the user. their logic isn't totally straight forward which is another reason why im hesitant to reimplement it in leo; google knows best what their restrictions are. i.e. a cluster with only a master node can't be resized to have more workers, a cluster with workers can't be resized down to 0, there must be 2 or more worker nodes, there can be any number > 0 of preemptible worker nodes, etc.

the actual message itself is user-friendly and says exactly what the problem is, but it's buried underneath a giant stacktrace. i think a good middle-ground here would be to rip the message out of the GoogleJsonResponseException and throw that into a new type of LeoException. that way we still let Google do the work of validating but we return something user-friendly

for reference, here's what that error message looks like. the message itself can vary but here's an example:

```
{
  "source": "leonardo",
  "causes": [],
  "exceptionClass": "com.google.api.client.googleapis.json.GoogleJsonResponseException",
  "stackTrace": [
    {
      "className": "com.google.api.client.googleapis.json.GoogleJsonResponseException",
      "methodName": "from",
      "fileName": "GoogleJsonResponseException.java",
      "lineNumber": 146
    },
    {
      "className": "com.google.api.client.googleapis.services.json.AbstractGoogleJsonClientRequest",
      "methodName": "newExceptionOnError",
      "fileName": "AbstractGoogleJsonClientRequest.java",
      "lineNumber": 113
    },
    {
      "className": "com.google.api.client.googleapis.services.json.AbstractGoogleJsonClientRequest",
      "methodName": "newExceptionOnError",
      "fileName": "AbstractGoogleJsonClientRequest.java",
      "lineNumber": 40
    },
    {
      "className": "com.google.api.client.googleapis.services.AbstractGoogleClientRequest$1",
      "methodName": "interceptResponse",
      "fileName": "AbstractGoogleClientRequest.java",
      "lineNumber": 321
    },
    {
      "className": "com.google.api.client.http.HttpRequest",
      "methodName": "execute",
      "fileName": "HttpRequest.java",
      "lineNumber": 1065
    },
    {
      "className": "com.google.api.client.googleapis.services.AbstractGoogleClientRequest",
      "methodName": "executeUnparsed",
      "fileName": "AbstractGoogleClientRequest.java",
      "lineNumber": 419
    },
    {
      "className": "com.google.api.client.googleapis.services.AbstractGoogleClientRequest",
      "methodName": "executeUnparsed",
      "fileName": "AbstractGoogleClientRequest.java",
      "lineNumber": 352
    },
    {
      "className": "org.broadinstitute.dsde.workbench.google.GoogleUtilities",
      "methodName": "$anonfun$executeGoogleCall$1",
      "fileName": "GoogleUtilities.scala",
      "lineNumber": 70
    },
    {
      "className": "scala.util.Try$",
      "methodName": "apply",
      "fileName": "Try.scala",
      "lineNumber": 209
    },
    {
      "className": "org.broadinstitute.dsde.workbench.google.GoogleUtilities",
      "methodName": "executeGoogleCall",
      "fileName": "GoogleUtilities.scala",
      "lineNumber": 70
    },
    {
      "className": "org.broadinstitute.dsde.workbench.google.GoogleUtilities",
      "methodName": "executeGoogleCall$",
      "fileName": "GoogleUtilities.scala",
      "lineNumber": 67
    },
    {
      "className": "org.broadinstitute.dsde.workbench.google.AbstractHttpGoogleDAO",
      "methodName": "executeGoogleCall",
      "fileName": "AbstractHttpGoogleDAO.scala",
      "lineNumber": 14
    },
    {
      "className": "org.broadinstitute.dsde.workbench.google.GoogleUtilities",
      "methodName": "executeGoogleRequest",
      "fileName": "GoogleUtilities.scala",
      "lineNumber": 51
    },
    {
      "className": "org.broadinstitute.dsde.workbench.google.GoogleUtilities",
      "methodName": "executeGoogleRequest$",
      "fileName": "GoogleUtilities.scala",
      "lineNumber": 50
    },
    {
      "className": "org.broadinstitute.dsde.workbench.google.AbstractHttpGoogleDAO",
      "methodName": "executeGoogleRequest",
      "fileName": "AbstractHttpGoogleDAO.scala",
      "lineNumber": 14
    },
    {
      "className": "org.broadinstitute.dsde.workbench.leonardo.dao.google.HttpGoogleDataprocDAO",
      "methodName": "$anonfun$resizeCluster$2",
      "fileName": "HttpGoogleDataprocDAO.scala",
      "lineNumber": 197
    },
    {
      "className": "akka.dispatch.MonitorableThreadFactory$AkkaForkJoinWorkerThread$$anon$3",
      "methodName": "block",
      "fileName": "ThreadPoolBuilder.scala",
      "lineNumber": 167
    },
    {
      "className": "akka.dispatch.forkjoin.ForkJoinPool",
      "methodName": "managedBlock",
      "fileName": "ForkJoinPool.java",
      "lineNumber": 3641
    },
    {
      "className": "akka.dispatch.MonitorableThreadFactory$AkkaForkJoinWorkerThread",
      "methodName": "blockOn",
      "fileName": "ThreadPoolBuilder.scala",
      "lineNumber": 165
    },
    {
      "className": "akka.dispatch.BatchingExecutor$BlockableBatch",
      "methodName": "blockOn",
      "fileName": "BatchingExecutor.scala",
      "lineNumber": 106
    },
    {
      "className": "scala.concurrent.package$",
      "methodName": "blocking",
      "fileName": "package.scala",
      "lineNumber": 142
    },
    {
      "className": "org.broadinstitute.dsde.workbench.google.GoogleUtilities",
      "methodName": "$anonfun$retryWhen500orGoogleError$3",
      "fileName": "GoogleUtilities.scala",
      "lineNumber": 42
    },
    {
      "className": "scala.concurrent.Future$",
      "methodName": "$anonfun$apply$1",
      "fileName": "Future.scala",
      "lineNumber": 654
    },
    {
      "className": "scala.util.Success",
      "methodName": "$anonfun$map$1",
      "fileName": "Try.scala",
      "lineNumber": 251
    },
    {
      "className": "scala.util.Success",
      "methodName": "map",
      "fileName": "Try.scala",
      "lineNumber": 209
    },
    {
      "className": "scala.concurrent.Future",
      "methodName": "$anonfun$map$1",
      "fileName": "Future.scala",
      "lineNumber": 288
    },
    {
      "className": "scala.concurrent.impl.Promise",
      "methodName": "liftedTree1$1",
      "fileName": "Promise.scala",
      "lineNumber": 29
    },
    {
      "className": "scala.concurrent.impl.Promise",
      "methodName": "$anonfun$transform$1",
      "fileName": "Promise.scala",
      "lineNumber": 29
    },
    {
      "className": "scala.concurrent.impl.CallbackRunnable",
      "methodName": "run",
      "fileName": "Promise.scala",
      "lineNumber": 60
    },
    {
      "className": "akka.dispatch.BatchingExecutor$AbstractBatch",
      "methodName": "processBatch",
      "fileName": "BatchingExecutor.scala",
      "lineNumber": 55
    },
    {
      "className": "akka.dispatch.BatchingExecutor$BlockableBatch",
      "methodName": "$anonfun$run$1",
      "fileName": "BatchingExecutor.scala",
      "lineNumber": 91
    },
    {
      "className": "scala.runtime.java8.JFunction0$mcV$sp",
      "methodName": "apply",
      "fileName": "JFunction0$mcV$sp.java",
      "lineNumber": 12
    },
    {
      "className": "scala.concurrent.BlockContext$",
      "methodName": "withBlockContext",
      "fileName": "BlockContext.scala",
      "lineNumber": 81
    },
    {
      "className": "akka.dispatch.BatchingExecutor$BlockableBatch",
      "methodName": "run",
      "fileName": "BatchingExecutor.scala",
      "lineNumber": 91
    },
    {
      "className": "akka.dispatch.TaskInvocation",
      "methodName": "run",
      "fileName": "AbstractDispatcher.scala",
      "lineNumber": 40
    },
    {
      "className": "akka.dispatch.ForkJoinExecutorConfigurator$AkkaForkJoinTask",
      "methodName": "exec",
      "fileName": "ForkJoinExecutorConfigurator.scala",
      "lineNumber": 44
    },
    {
      "className": "akka.dispatch.forkjoin.ForkJoinTask",
      "methodName": "doExec",
      "fileName": "ForkJoinTask.java",
      "lineNumber": 260
    },
    {
      "className": "akka.dispatch.forkjoin.ForkJoinPool$WorkQueue",
      "methodName": "runTask",
      "fileName": "ForkJoinPool.java",
      "lineNumber": 1339
    },
    {
      "className": "akka.dispatch.forkjoin.ForkJoinPool",
      "methodName": "runWorker",
      "fileName": "ForkJoinPool.java",
      "lineNumber": 1979
    },
    {
      "className": "akka.dispatch.forkjoin.ForkJoinWorkerThread",
      "methodName": "run",
      "fileName": "ForkJoinWorkerThread.java",
      "lineNumber": 107
    }
  ],
  "message": "400 Bad Request\n{\n  \"code\" : 400,\n  \"errors\" : [ {\n    \"domain\" : \"global\",\n    \"message\" : \"Cluster configuration must have at least 2 workers.\",\n    \"reason\" : \"badRequest\"\n  } ],\n  \"message\" : \"Cluster configuration must have at least 2 workers.\",\n  \"status\" : \"INVALID_ARGUMENT\"\n}"
}```

Comment (kyuksel) [reply src/test/scala/org/broadinstitute/dsde/workbench/leonardo/dao/google/MockGoogleDataprocDAO.scala]: I see. The middle-ground plan sounds great then.

File: src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala (hunk 17 above)
Comment (kyuksel) [reply src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala]: I couldn't spot the negative tests for  ^^.

File: src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala [hunk 19]
```diff
@@ -340,6 +346,76 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
...
+          // If the Google Compute default service account is being used, this is not necessary.
+          _ <- addDataprocWorkerRoleToServiceAccount(existingCluster.googleProject, existingCluster.serviceAccountInfo.clusterServiceAccount)
+          _ <- gdDAO.resizeCluster(existingCluster.googleProject, existingCluster.clusterName, machineConfig.numberOfWorkers, machineConfig.numberOfPreemptibleWorkers) recoverWith {
+            case gjre: GoogleJsonResponseException =>
+              //typically we will revoke this role in the monitor after everything is complete, but if Google fails to resize the cluster we need to revoke it manually here
+              removeDataprocWorkerRoleFromServiceAccount(existingCluster.googleProject, existingCluster.serviceAccountInfo.clusterServiceAccount)
+              throw InvalidDataprocMachineConfigException(gjre.getMessage)
+          }
+          resizedNumWorkers <- machineConfig.numberOfWorkers.map { numWorkers =>
+            dbRef.inTransaction { _.clusterQuery.updateNumberOfWorkers(existingCluster.id, numWorkers) }.map(_ > 0)
```
kyuksel: Cool (using `map(_ > 0)` to check if the database transaction succeeded)! It'd be nice to factor it out into a utility so we can easily use in other places with DB transactions where we want to act accordingly, too. Not a blocker.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -340,6 +346,76 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
...
+          } getOrElse Future.successful(false)
+          resizedNumPreemptibles <- machineConfig.numberOfPreemptibleWorkers.map { numWorkers =>
+            dbRef.inTransaction { _.clusterQuery.updateNumberOfPreemptibleWorkers(existingCluster.id, Option(numWorkers)) }.map(_ > 0)
+          } getOrElse Future.successful(false)
+        } yield { resizedNumWorkers || resizedNumPreemptibles }
```
kyuksel: I wonder if it should be `&&` rather than `||`. I guess even when one set of workers or preemptibles are resized, semantically, it's correct to say the cluster was resized?

kyuksel (commented):
Looks good. Mainly a concern around horizontal scaling and a question around testing.

kyuksel:
P.S. Are we doing [graceful decommissioning](https://cloud.google.com/dataproc/docs/concepts/configuring-clusters/scaling-clusters)?

MatthewBemis:
re: graceful commissioning- we're not doing it now but we could totally add it to the endpoint as an option for users later on

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
MatthewBemis: yeah we'd consider it resized if either worker type was scaled

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -340,6 +346,76 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
...
+                    clusterName: ClusterName,
+                    clusterRequest: ClusterRequest): Future[Cluster] = {
+    for {
+      cluster <- internalGetActiveClusterDetails(googleProject, clusterName) //throws 404 if nonexistent
+
+      _ <- checkClusterPermission(userInfo, ModifyCluster, cluster) //throws 404 if no auth
+
+      _ <- internalUpdateCluster(cluster, clusterRequest)
+
+      updatedCluster <- internalGetActiveClusterDetails(googleProject, clusterName)
```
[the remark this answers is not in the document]
MatthewBemis: nice catch :+1:

MatthewBemis:
swatomation actually passed, it was some new downstream job that failed called "upload-swatomation-artifacts"

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -340,6 +346,74 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
...
+
+        logger.info(s"New machine config present. Resizing cluster '${existingCluster.clusterName}' " +
+          s"in Google project '${existingCluster.googleProject}'...")
+
+        for {
+          // Add Dataproc Worker role to the cluster service account, if present.
+          // This is needed to be able to spin up Dataproc clusters.
+          // If the Google Compute default service account is being used, this is not necessary.
+          _ <- addDataprocWorkerRoleToServiceAccount(existingCluster.googleProject, existingCluster.serviceAccountInfo.clusterServiceAccount)
+          _ <- gdDAO.resizeCluster(existingCluster.googleProject, existingCluster.clusterName, machineConfig.numberOfWorkers, machineConfig.numberOfPreemptibleWorkers) recoverWith {
```
rtitle: It might be good to avoid calling Google if neither `numberOfWorkers` nor `numberOfPreemptibleWorkers` changed. See comment below.

rtitle:
Just one comment above -- I'm not sure `maybeResizeCluster` is working correctly if you pass in an identical machine configuration for a given cluster. I think (because of how slick works) it would update Google and kick off monitoring. Probably not a huge deal but would be good to fix.

Rest looks good!

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -340,6 +346,81 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
...
+    }
+  }
+
+  //returns true if cluster was resized, otherwise returns false
+  def maybeResizeCluster(existingCluster: Cluster, machineConfigOpt: Option[MachineConfig]): Future[Boolean] = {
+    machineConfigOpt match {
+      case Some(machineConfig) =>
+
+        (machineConfig.numberOfWorkers, machineConfig.numberOfPreemptibleWorkers) match {
+          case (Some(_), _) | (_, Some(_))
```
MatthewBemis: i wish there was a way to bind variables when matching against multiple patterns...

MatthewBemis:
jenkins retest because google problems resolved

MatthewBemis:
jenkins retest because of build failure- will dig in more if it happens again

rtitle (approved):
Updated logic looks good

MatthewBemis:
jenkins retest because of a different build failure: ```W: Failed to fetch http://cran.mtu.edu/bin/linux/debian/jessie-cran34/InRelease  

W: Failed to fetch http://cran.mtu.edu/bin/linux/debian/jessie-cran34/Release.gpg  Unable to connect to cran.mtu.edu:http:

W: Some index files failed to download. They have been ignored, or old ones used instead.```

PR #693, opened 2018-11-27
[692] Allow users to pick their Jupyter image at cluster creation time

rtitle:
See https://github.com/DataBiosphere/leonardo/issues/692

Originally this was part of my RStudio branch but I decided to lift it into its own PR since it will be useful with the docker image upgrades as well. This change does the following:
- Adds a new optional `jupyterDockerImage` field to the `ClusterRequest` json
- If set, installs the specified Jupyter image on the cluster
- If unset, installs the latest "prod" image on the cluster (same behavior as today)
- Adds a new `CLUSTER_IMAGE` table which tracks the images installed on a cluster

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
@@ -45,7 +45,7 @@ case class ServiceAccountInfoRecord(clusterServiceAccount: Option[String],
                                     serviceAccountKeyId: Option[String])
 
 trait ClusterComponent extends LeoComponent {
-  this: LabelComponent with ClusterErrorComponent with InstanceComponent with ExtensionComponent =>
+  this: LabelComponent with ClusterErrorComponent with InstanceComponent with ExtensionComponent with ClusterImageComponent =>
 
   import profile.api._
 
@@ -92,8 +92,8 @@ trait ClusterComponent extends LeoComponent {
       (clusterServiceAccount, notebookServiceAccount, serviceAccountKeyId), stagingBucket, dateAccessed, autopauseThreshold, defaultClientId, stopAfterCreation
     ).shaped <> ({
       case (id, clusterName, googleId, googleProject, operationName, status, hostIp, creator,
-            createdDate, destroyedDate, jupyterExtensionUri, jupyterUserScriptUri, initBucket, machineConfig,
-            serviceAccountInfo, stagingBucket, dateAccessed, autopauseThreshold, defaultClientId, stopAfterCreation) =>
+      createdDate, destroyedDate, jupyterExtensionUri, jupyterUserScriptUri, initBucket, machineConfig,
+      serviceAccountInfo, stagingBucket, dateAccessed, autopauseThreshold, defaultClientId, stopAfterCreation) =>
         ClusterRecord(
           id, clusterName, googleId, googleProject, operationName, status, hostIp, creator,
           createdDate, destroyedDate, jupyterExtensionUri, jupyterUserScriptUri, initBucket,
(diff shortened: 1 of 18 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
```diff
@@ -451,31 +450,46 @@ trait ClusterComponent extends LeoComponent {
...
       )
     }
   }
 
-  // select * from cluster c left join label l on c.id = l.clusterId
-  val clusterQueryWithLabels: Query[(ClusterTable, Rep[Option[LabelTable]], Rep[Option[ExtensionTable]]), (ClusterRecord, Option[LabelRecord], Option[ExtensionRecord]), Seq] = {
+  // just clusters and labels: no instances, extensions, etc.
+  //   select * from cluster c
+  //   left join label l on c.id = l.clusterId
+  val minimalClusterQuery: Query[(ClusterTable, Rep[Option[LabelTable]]), (ClusterRecord, Option[LabelRecord]), Seq] = {
```
rtitle: I decided to rename these queries (and related `unmarshal*` methods) to `minimalClusterQuery` and `fullClusterQuery`. The former is used by "list clusters" and only joins clusters to labels. The latter is used by "get cluster" and does all the joins to retrieve cluster metadata.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
```diff
@@ -60,6 +62,16 @@ case class AuditInfo(creator: WorkbenchEmail,
                      destroyedDate: Option[Instant],
                      dateAccessed: Instant)
 
+sealed trait ClusterTool extends EnumEntry
+object ClusterTool extends Enum[ClusterTool] {
+  val values = findValues
+  case object Jupyter extends ClusterTool
+  case object RStudio extends ClusterTool
```
rtitle: RStudio not supported in the API yet, but will be soon.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -828,4 +840,12 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
       .copy(labels = allLabels)
       .copy(userJupyterExtensionConfig = updatedUserJupyterExtensionConfig)
   }
+
+  private[service] def processClusterImages(clusterRequest: ClusterRequest): Set[ClusterImage] = {
+    // Default to the configured default image if not specified in the request
+    // TODO should we validate the image somehow?
```
rtitle: I wasn't sure what kind of docker validation to do up front before we try to create the cluster. If the user passes a bogus docker image it'll fail in the init script -- is this OK or should we do some kind of validation sooner?

src/test/scala/org/broadinstitute/dsde/workbench/leonardo/ClusterEnrichments.scala
```diff
@@ -51,4 +51,12 @@ object ClusterEnrichments {
 
     fcs1 == fcs2
   }
+
+  def stripFieldsForListCluster(cluster: Cluster): Cluster = {
```
rtitle: This is to make tests that validate `list clusters` results pass.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
MatthewBemis: what type of validation can we actually do up front? in it's current state, i'm not sure what the behavior would look like if we let it fail in the init script, but i'd hate for us to get user bug reports and be on the hook for debugging what's wrong with custom docker images. as long as we can surface an error message that's useful i think this is ok

rtitle: Yeah agreed - we could parse the docker images but that seems kind of brittle. I'm ok with letting it fail at cluster init time (it should put the cluster in Error state and surface an error message with the init logs).

PR #712, opened 2018-12-06
(WIP) RStudio phase 2: actually install the image on the cluster

rtitle:
Rebased against develop branch, now that phase 1 is merged. Still needs some testing/vetting before merge, but can open for review.

In all cases:


rtitle:
@MatthewBemis maybe I was a little zealous in requesting review, this isn't done yet (tests are failing). Feel free to look but it's still WIP.

src/main/resources/jupyter/cluster-site.conf
```diff
@@ -9,12 +9,30 @@
...
+    # RStudio
+    RewriteCond %{HTTP:Upgrade} =websocket
+    RewriteCond %{REQUEST_URI} /notebooks/[^/]*/[^/]*/rstudio/.* [NC]
+    RewriteRule /notebooks/[^/]*/[^/]*/rstudio/(.*)     ws://127.0.0.1:8001/$1  [P,L]
+
+    RewriteCond %{HTTP:Upgrade} !=websocket
+    RewriteCond %{REQUEST_URI} /notebooks/[^/]*/[^/]*/rstudio/.* [NC]
+    RewriteRule /notebooks/[^/]*/[^/]*/rstudio/(.*)     http://127.0.0.1:8001/$1 [P,L]
+
+    ProxyPass /notebooks/[^/]*/[^/]*/rstudio/ http://127.0.0.1:8001/
```
rtitle: This isn't ideal because it works on `/notebooks/project/cluster/rstudio/` but not `/notebooks/project/cluster/rstudio`. But maybe phase 3 can address this.

rtitle: Also if these apache configs become too much of a headache we could consider shifting some of the rewrite logic into the Scala proxy code. I kind of like the logic localized in 1 place through, and the Scala proxy being a pass-through.

src/main/resources/jupyter/init-actions.sh
```diff
@@ -160,130 +164,142 @@ if [[ "${ROLE}" == 'Master' ]]; then
...
+
+      if [ ! -z ${JUPYTER_NOTEBOOK_CONFIG_URI} ] ; then
+        log 'Copy Jupyter notebook config...'
+        gsutil cp ${JUPYTER_NOTEBOOK_CONFIG_URI} /etc
+        JUPYTER_NOTEBOOK_CONFIG=`basename ${JUPYTER_NOTEBOOK_CONFIG_URI}`
+        docker cp /etc/${JUPYTER_NOTEBOOK_CONFIG} ${JUPYTER_SERVER_NAME}:${JUPYTER_HOME}/
+      fi
+
+      # If a Jupyter user script was specified, copy it into the jupyter docker container.
+      # TODO: not Jupyter specific
```
rtitle: Noting TODO - we should support user scripts on all images

src/main/resources/jupyter/cluster-site.conf
MatthewBemis: how will this work when we support BYOT? won't we need to move it down to the actual app in some manner at that point anyway?

this is forward-thinking and not a blocker but wanted to raise the Q since it felt relevant to your second comment

MatthewBemis (commented):
the code changes look good. i have played around with them a bit while working on phase 3, but i wouldn't say i've completely vetted/tested it. 

there are still some jupyter-specific references near these code changes. we could handle it in this phase or phase 3 but i think maybe we should just clean all of that up once the dust settles from merging everything

src/main/resources/jupyter/cluster-site.conf
rtitle: That's a good point. Yeah editing this apache conf won't be scalable for BYOT, we'd probably have to move logic in the app.

Issue #595, opened 2018-09-24
Leo should support JupyterLab extensions


zarsky-broad:
JupyterLab extensions are installed separately from Jupyter Notebook extensions. This should be supported by the API.

See here:
https://jupyterlab.readthedocs.io/en/stable/user/extensions.html#installing-extensions

PR #714, opened 2018-12-07
JupyterLab Extension Support

akarukappadath:

In all cases:


akarukappadath:
jenkins retest

akarukappadath:
jenkins retest

akarukappadath:
jenkins retest

docker/jupyter/Dockerfile
```diff
@@ -257,6 +257,15 @@ RUN R -e 'install.packages(c( \
  && echo "R_LIBS=/home/jupyter-user/.rpackages" > /home/jupyter-user/.Renviron \
  && chown -R $USER:users /home/jupyter-user/.rpackages
 
+#######################
+# JupyterLab Extension Requirements
+#######################
+
+RUN apt-get update \
+ && apt-get install -t testing -yq --no-install-recommends \
+    nodejs \
```
rtitle: Maybe we could just include `nodejs` and `npm` in the python 3 section? Adding a new layer increases the image size & build time.

jupyter-docker/scripts/extension/jupyter_install_lab_extension.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
```
rtitle: This file should be removed I think, there is no `jupyter-docker` directory anymore

src/main/resources/jupyter/init-actions.sh
```diff
@@ -247,6 +248,22 @@ if [[ "${ROLE}" == 'Master' ]]; then
...
+    if [ ! -z "${JUPYTER_LAB_EXTENSIONS}" ] ; then
+      for ext in ${JUPYTER_LAB_EXTENSIONS}
+      do
+        log 'Installing JupyterLab extension [$ext]...'
+        if [[ $ext == 'gs://'* ]]; then
+          gsutil cp $ext /etc
+          JUPYTER_EXTENSION_ARCHIVE=`basename $ext`
+          docker cp /etc/${JUPYTER_EXTENSION_ARCHIVE} ${JUPYTER_SERVER_NAME}:${JUPYTER_HOME}/${JUPYTER_EXTENSION_ARCHIVE}
+          retry 3 docker exec ${JUPYTER_SERVER_NAME} ${JUPYTER_SCRIPTS}/extension/jupyter_install_lab_extension.sh ${JUPYTER_HOME}/${JUPYTER_EXTENSION_ARCHIVE}
+        else
```
rtitle: I'm just thinking of this, but can we support `http://` or `https://` locations for lab extensions? We already do similar for nbextensions (see above). Terra likes this feature because they can host their own extensions, and not need to copy to a bucket.

rtitle (changes requested):
Looks good, a couple minor comments

docker/jupyter/Dockerfile
```diff
@@ -268,6 +271,8 @@ ADD custom/jupyter_localize_extension.py $JUPYTER_HOME/custom/
 RUN chown -R $USER:users $JUPYTER_HOME \
  && find $JUPYTER_HOME/scripts -name '*.sh' -type f | xargs chmod +x
 
+RUN chown -R $USER:users /usr/local/share/jupyter/lab
```
rtitle: Sorry.. maybe this chown could just be part of the above `RUN`:
```
RUN chown -R $USER:users $JUPYTER_HOME \
 && find $JUPYTER_HOME/scripts -name '*.sh' -type f | xargs chmod +x
 && chown -R $USER:users /usr/local/share/jupyter/lab
```

src/main/resources/swagger/api-docs.yaml
```diff
@@ -972,6 +972,10 @@ definitions:
         description: |
           Optional, map of extension name and notebook plus server extension. The extension can either be a tar.gz file on google storage or a python package.
           Example, {"ext1":"gs://bucket/extension.tar.gz", "ext2":"python-package"}
+      labExtensions:
+        type: object
+        description: |
+          Optional, map of extension name and lab extension. The extension should be a verified jupyterlab extension, a local directory containing the extension, a gzipped tarball, or a URL to a gzipped tarball.
```
rtitle: Can we validate the gzipped tarballs work?

docker/jupyter/scripts/extension/jupyter_install_lab_extension.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+set -e
+
+ if [ -n "$1" ]; then
+  JUPYTER_EXTENSION=$1
+  jupyter labextension install ${JUPYTER_EXTENSION}
```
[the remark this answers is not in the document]
akarukappadath: I updated the swagger documentation based on what the documentation for jupyterlab says : https://jupyterlab.readthedocs.io/en/stable/user/extensions.html

Do you think I should try other file types that aren't specified in the docs?

rtitle (approved):
Minor comment on the swagger docs, otherwise looks good to me

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/NotebookCustomizationSpec.scala
```diff
@@ -78,5 +78,19 @@ final class NotebookCustomizationSpec extends FreeSpec
         }
       }
     }
+
+    "should install user specified lab extensions" in {
+      withProject { project => implicit token =>
+        withNewCluster(project, request = ClusterRequest(userJupyterExtensionConfig = Some(UserJupyterExtensionConfig(labExtensions = Map("jupyterlab-toc" -> "@jupyterlab/toc"))))) { cluster =>
```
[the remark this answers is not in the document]
akarukappadath: nope, toc has to be installed by the user! I know vikram was working on making toc be installed by default for notebooks - I can make a ticket to do the same for lab if you think that's a good idea.

rtitle:
Doc update looks great 👍

mbookman:
@rtitle : is this in Terra now? I went to install the Jupyterlab TOC support yesterday.
My cluster currently has 0.35.2. Before I go rebuild it, I wanted to make sure I'd actually pick up these changes.

As a more general question - Is there somewhere I could have looked to see that this change was in fact currently available in Terra?

Thanks!

rtitle:
@mbookman This PR was to support specifying lab extensions that get installed by default on the notebook server. For example Terra UI wants to define their own extension to configure auto-save frequency and customize look and feel, which gets installed by default for all users. The consumer of this change is really Terra UI developers.

You should be able to install your own lab extensions from a notebook today, even without this PR. For example just now I did:
```
!jupyter labextension install @jupyterlab/toc
```
and I see the toc view in Lab.

Side note: we might want to just install toc by default in Lab since it's generally useful.

mbookman:
Thanks Rob. I had gone to do the jupyterlab/toc extension yesterday and ran into:

```
$ jupyter labextension install @jupyterlab/toc

Errored, use --debug for full output:
ValueError: Please install nodejs 5+ and npm before continuing installation. nodejs may be installed using conda or directly from the nodejs website.
```

Before I went and installed nodejs and npm myself, I looked to see if this infrastructure was put into the default leo docker image. I found this issue and was trying to determine whether recreating my cluster would pick up this change.

I'm wondering, as a general practice, how I might determine this before recreating the cluster.
I'm thinking about how we support researchers who might report similar problems. How can we give them insight into whether rebuilding a Terra cluster is going to pick up a particular Leo PR that was merged N days ago.

rtitle:
Oh right I forgort -- this PR started installing npm in the image so you'd need to have these changes before you can install Lab extensions. I can confirm this PR is in production today.

One way to determine that is to check the issue linked to this PR: https://github.com/DataBiosphere/leonardo/issues/595. It is labeled with a release candidate which includes the date of release (in this case 1/22/19).

mbookman:
That's very helpful. Thanks!

Issue #154, opened 2018-01-26
Remove --net=host from Leo Jupyter and expose a smaller number of ports


helgridly:
[GAWB-2626](https://broadinstitute.atlassian.net/browse/GAWB-2626)

Currently we pass --net=host when we spin up the Jupyter container, which forwards all ports from the docker container to the host. This is a bit gross. Find the list of ports required for Jupyter + Spark + Hail + god knows what else and expose only those.

There's some docker networking things to do to get the Apache proxy talking to the Jupyter docker too.

PR #718, opened 2018-12-11
RStudio phase3: genericize proxy routes and deprecate old ones

MatthewBemis:

In all cases:


src/test/scala/org/broadinstitute/dsde/workbench/leonardo/api/ProxyRoutesSpec.scala
```diff
@@ -47,208 +47,212 @@ class ProxyRoutesSpec extends FlatSpec with BeforeAndAfterAll with BeforeAndAfte
...
   }
 
-  "ProxyRoutes" should "listen on /notebooks/{project}/{name}/..." in {
-    Get(s"/notebooks/$googleProject/$clusterName").addHeader(Cookie(tokenCookie)) ~> leoRoutes.route ~> check {
-      handled shouldBe true
-      status shouldEqual StatusCodes.OK
-      validateCors()
+  val pathPrefixes = Set("notebooks", "proxy")
+
+  for(prefix <- pathPrefixes) {
```
MatthewBemis: this file is all the same, it's just nested inside of a loop to test the new routes and the deprecated ones

src/main/resources/jupyter/cluster-site.conf
```diff
@@ -12,18 +12,33 @@
...
     RewriteCond %{HTTP:Upgrade} =websocket
-    RewriteCond %{REQUEST_URI} /notebooks/[^/]*/[^/]*/rstudio/.* [NC]
-    RewriteRule /notebooks/[^/]*/[^/]*/rstudio/(.*)     ws://127.0.0.1:8001/$1  [P,L]
+    RewriteCond %{REQUEST_URI} /proxy/[^/]*/[^/]*/rstudio/.* [NC]
+    RewriteRule /proxy/[^/]*/[^/]*/rstudio/(.*)     ws://127.0.0.1:8001/$1  [P,L]
 
     RewriteCond %{HTTP:Upgrade} !=websocket
-    RewriteCond %{REQUEST_URI} /notebooks/[^/]*/[^/]*/rstudio/.* [NC]
-    RewriteRule /notebooks/[^/]*/[^/]*/rstudio/(.*)     http://127.0.0.1:8001/$1 [P,L]
+    RewriteCond %{REQUEST_URI} /proxy/[^/]*/[^/]*/rstudio/.* [NC]
```
MatthewBemis: as discussed, would like a way to not require the trailing slash. i've tried things like 

```
RewriteCond %{REQUEST_URI} /proxy/[^/]*/[^/]*/rstudio/.* [NC,OR]
RewriteCond %{REQUEST_URI} /proxy/[^/]*/[^/]*/rstudio [NC]
```
and other variations but no luck. the hammer approach may be to just duplicate these blocks entirely

src/main/resources/jupyter/init-actions.sh
```diff
@@ -160,130 +164,142 @@ if [[ "${ROLE}" == 'Master' ]]; then
...
+      # Change Python and PySpark 2 and 3 kernel specs to allow each to have its own spark
+      retry 3 docker exec -u root ${JUPYTER_SERVER_NAME} ${JUPYTER_SCRIPTS}/kernel/kernelspec.sh ${JUPYTER_SCRIPTS}/kernel ${KERNELSPEC_HOME}
+
+      log 'Installing Hail additions to Jupydocker spark.conf...'
+
+      # Install the Hail additions to Spark conf.
+      retry 3 docker exec -u root ${JUPYTER_SERVER_NAME} ${JUPYTER_SCRIPTS}/hail/spark_install_hail.sh
+
+      # Copy the actual service account JSON file into the Jupyter docker container.
+      # TODO: not Jupyter specific
```
rtitle: I think I had noted this in my PR, but would be good to lift this out if the `if [ ! -z ${JUPYTER_DOCKER_IMAGE} ] ; then` block.

src/main/resources/jupyter/init-actions.sh
```diff
@@ -160,130 +164,142 @@ if [[ "${ROLE}" == 'Master' ]]; then
...
+
+      if [ ! -z ${JUPYTER_NOTEBOOK_CONFIG_URI} ] ; then
+        log 'Copy Jupyter notebook config...'
+        gsutil cp ${JUPYTER_NOTEBOOK_CONFIG_URI} /etc
+        JUPYTER_NOTEBOOK_CONFIG=`basename ${JUPYTER_NOTEBOOK_CONFIG_URI}`
+        docker cp /etc/${JUPYTER_NOTEBOOK_CONFIG} ${JUPYTER_SERVER_NAME}:${JUPYTER_HOME}/
+      fi
+
+      # If a Jupyter user script was specified, copy it into the jupyter docker container.
+      # TODO: not Jupyter specific
```
rtitle: Same as above: move out of the `if jupyter` block.

src/main/resources/jupyter/cluster-site.conf
rtitle: There must be a solution here, I'm just not sure offhand. Can play around a little..

src/main/resources/jupyter/init-actions.sh
```diff
@@ -160,130 +164,142 @@ if [[ "${ROLE}" == 'Master' ]]; then
...
+        log 'Installing Jupyter user extension [$JUPYTER_USER_SCRIPT_URI]...'
+        gsutil cp ${JUPYTER_USER_SCRIPT_URI} /etc
+        JUPYTER_USER_SCRIPT=`basename ${JUPYTER_USER_SCRIPT_URI}`
+        docker cp /etc/${JUPYTER_USER_SCRIPT} ${JUPYTER_SERVER_NAME}:${JUPYTER_HOME}/${JUPYTER_USER_SCRIPT}
+        retry 3 docker exec -u root ${JUPYTER_SERVER_NAME} chmod +x ${JUPYTER_HOME}/${JUPYTER_USER_SCRIPT}
+        retry 3 docker exec -u root -e PIP_USER=false ${JUPYTER_SERVER_NAME} ${JUPYTER_HOME}/${JUPYTER_USER_SCRIPT}
+      fi
+
+      log 'Starting Jupyter Notebook...'
+      retry 3 docker exec -d ${JUPYTER_SERVER_NAME} ${JUPYTER_SCRIPTS}/run-jupyter.sh
```
rtitle: I just remembered, we also install a startup script on the master node to restart Jupyter when the instance is restarted:
https://github.com/DataBiosphere/leonardo/blob/develop/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala#L117-L124

We should probably change this script to test if the Jupyter image is installed.

Note: RStudio doesn't need an explicit startup command because we just use the docker `ENTRYPOINT`.

src/main/resources/jupyter/rstudio-docker-compose.yaml
```diff
@@ -0,0 +1,15 @@
+version: '2'
+services:
+  rstudio:
+    container_name: "${RSTUDIO_SERVER_NAME}"
+    image: "${RSTUDIO_DOCKER_IMAGE}"
+    network_mode: host
```
rtitle: I'm wondering if we should stop using `network_mode: host` at some point and start explicitly managing ports. We've been lucky there are no conflicts so far. Probably out of scope for this PR.

rtitle: In fact there's even a ticket: https://github.com/DataBiosphere/leonardo/issues/154

src/main/resources/reference.conf
```diff
@@ -2,9 +2,10 @@
 # Google Cloud dataproc configuration
 dataproc {
   applicationName = "firecloud:leonardo"
-  dataprocDockerImage = "gcr.io/broad-dsde-prod/leonardo-notebooks:prod"
+  dataprocDockerImage = "gcr.io/broad-dsde-prod/leonardo-jupyter:prod"
```
rtitle: Would be good to change this in firecloud-develop as well. Then we can remove the block that pushes both `leonardo-notebooks` and `leonardo-jupyter`:
https://github.com/DataBiosphere/leonardo/blob/develop/docker/build-tool.sh#L121-L128

src/main/resources/swagger/api-docs.yaml
```diff
@@ -641,6 +643,102 @@ paths:
           schema:
             $ref: '#/definitions/ErrorReport'
 
+  '/proxy/{googleProject}/{clusterName}':
```
rtitle: I'm wondering how we should document the `/rstudio`, `/jupyter` APIs. Should we put them in swagger, or treat them as more "dynamic"? Should Leo have some kind of landing page that displays the set of tools running on a cluster?

Not sure atm - just posing questions.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/ProxyRoutes.scala
```diff
@@ -49,25 +50,42 @@ trait ProxyRoutes extends UserInfoDirectives with CorsSupport with CookieHelper
...
-            extractToken { token =>
-              complete {
-                proxyService.invalidateAccessToken(token).map { _ =>
-                  logger.debug(s"Invalidated access token $token")
-                  StatusCodes.OK
+          }
+      }
+    }
+
+  protected val jupyterRoutes: Route =
```
rtitle: I don't see where `jupyterRoutes` is referenced. 

Also it would be good if the `../api/localize` path worked for both legacy and new proxy paths.

rtitle (changes requested):
A few comments above

src/main/resources/swagger/api-docs.yaml
MatthewBemis: because the field will end up being even more dynamic than just rstudio and jupyter i'd rather we not have them documented in swagger. a landing page or at least some endpoint that users can call to get info about the running tools seems like a good approach IMO

src/main/resources/reference.conf
MatthewBemis: https://github.com/broadinstitute/firecloud-develop/pull/1481

src/main/resources/jupyter/rstudio-docker-compose.yaml
MatthewBemis: i'll leave this for that ticket but agree that we should start doing this soon

MatthewBemis:
jenkins retest

docker/build-tool.sh
```diff
@@ -117,15 +117,6 @@ function push() {
     $DOCKER_BINARY push "${REPOSITORY}/leonardo-${IMAGE}:${DOCKER_TAG}"
     $DOCKER_BINARY tag "${REPOSITORY}/leonardo-${IMAGE}:${DOCKER_TAG}" "${REPOSITORY}/leonardo-${IMAGE}:${GIT_BRANCH}"
     $DOCKER_BINARY push "${REPOSITORY}/leonardo-${IMAGE}:${GIT_BRANCH}"
-
-    # For backwards compatibility; also tag and push jupyter images named 'leonardo-notebooks'.
```
rtitle: Means https://github.com/broadinstitute/firecloud-develop/pull/1481 should be merged first, right?

src/main/resources/jupyter/cluster-site.conf
rtitle: Ok, I don't think that needs to block this PR, we can have a follow-on task

src/main/resources/jupyter/init-actions.sh
rtitle: We are treating this as Jupyter specific, right? Can remove the TODO comment

src/main/resources/jupyter/rstudio-docker-compose.yaml
rtitle: sounds good

src/main/resources/swagger/api-docs.yaml
rtitle: True, but since the tools are kind of baked in today it might be nice to document them in swagger? When we have fully dynamic tools then maybe we can look into different ways of discovering them.

Maybe we could have something like the following in swagger:
- `/proxy/{googleProject}/{clusterName}/jupyter` -- access Jupyter
- `/proxy/{googleProject}/{clusterName}/jupyter/lab` -- access JupyterLab
- `/proxy/{googleProject}/{clusterName}/rstudio` -- access RStudio
- `/proxy/{googleProject}/{clusterName}/setCookie` -- sets a cookie (I don't think this is tool specific)
- `/proxy/invalidateToken` -- invalidates token (again not tool specific)

WDYT?

src/main/resources/jupyter/cluster-site.conf
```diff
@@ -9,12 +9,45 @@
...
+    RewriteRule /proxy/[^/]*/[^/]*/rstudio/(.*)     ws://127.0.0.1:8001/$1  [P,L]
+
+    RewriteCond %{HTTP:Upgrade} !=websocket
+    RewriteCond %{REQUEST_URI} /proxy/[^/]*/[^/]*/rstudio/.* [NC]
+    RewriteRule /proxy/[^/]*/[^/]*/rstudio/(.*)     http://127.0.0.1:8001/$1 [P,L]
+
+    ProxyPass /proxy/[^/]*/[^/]*/rstudio/ http://127.0.0.1:8001/
+    ProxyPassReverse /proxy/[^/]*/[^/]*/rstudio/ http://127.0.0.1:8001/
+
+    # Jupyter
```
rtitle: Could we update the automation tests to use to non-legacy endpoint(s)?

src/main/resources/jupyter/init-actions.sh
rtitle: I don't think I saw this addressed -- the above script assumes the existence of Jupyter, so probably needs to change

rtitle:
Just a few more comments, the rest looks good though

src/main/resources/jupyter/init-actions.sh
MatthewBemis: i did this in leo code instead of script https://github.com/DataBiosphere/leonardo/pull/718/commits/e72240b66372e1afc6fa0febbe1934fb557d364c#diff-9a3d2448dba603018e03e21dab912aa9R489

i think a follow-up to all of this would be to no longer hard code these things at all

docker/build-tool.sh
MatthewBemis: yes will merge that ahead of this

MatthewBemis:
jenkins retest (gpalloc was under maintenance)

MatthewBemis:
jenkins retest - hopefully gpalloc issues are cleared up

MatthewBemis:
jenkins retest

MatthewBemis:
jenkins retest

rtitle:
Sorry if my recent merge of https://github.com/DataBiosphere/leonardo/pull/787 caused some conflicts. :( Happy to re-review this when you're ready.

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/Leonardo.scala
```diff
@@ -174,6 +174,12 @@ object Leonardo extends RestClient with LazyLogging {
       mapper.readValue(response, classOf[ContentItem])
     }
 
+    def proxyBasePath(googleProject: GoogleProject, clusterName: ClusterName): String =
+      s"proxy/${googleProject.value}/${clusterName.string}"
+
+    def jupyterProxyBasePath(googleProject: GoogleProject, clusterName: ClusterName): String =
```
MatthewBemis: i've decided to not use the new paths quite yet in the tests. all of our apps still use the legacy paths so it felt wrong to leave those uncovered. the alternative was to test both paths but then we would have tests split or duplicated. 

i think in planning for the deprecation and removal of the legacy paths, we can figure out a better plan for these tests.

i'm thinking something like

1. email app teams about removal of the legacy path in X days. tell them it will become unsupported prior to removal
2. move tests to use new path
3. remove old path

...or something similar

MatthewBemis:
jenkins retest

MatthewBemis:
jenkins retest

it was using an outdated image because of reasons 😠

MatthewBemis:
jenkins retest

MatthewBemis:
jenkins retest

src/main/resources/swagger/api-docs.yaml
```diff
@@ -640,6 +642,187 @@ paths:
...
+            $ref: '#/definitions/ErrorReport'
+        '422':
+          description: Cluster is stopped
+          schema:
+            $ref: '#/definitions/ErrorReport'
+        '500':
+          description: Internal Error
+          schema:
+            $ref: '#/definitions/ErrorReport'
+#  '/proxy/{googleProject}/{clusterName}/rstudio/':
```
MatthewBemis: this is hidden for now

MatthewBemis:
jenkins retest transient build failure (hopefully)

MatthewBemis:
jenkins retest - need to look into these different kinds of build failures later as i've seen them before but they seem more frequent right now

src/main/resources/jupyter/cluster-site.conf
```diff
@@ -9,12 +9,45 @@
...
+
+    RewriteCond %{HTTP:Upgrade} =websocket
+    RewriteCond %{REQUEST_URI} /proxy/[^/]*/[^/]*/rstudio/.* [NC]
+    RewriteRule /proxy/[^/]*/[^/]*/rstudio/(.*) ws://127.0.0.1:8001/$1  [P,L]
+
+    RewriteCond %{HTTP:Upgrade} !=websocket
+    RewriteCond %{REQUEST_URI} /proxy/[^/]*/[^/]*/rstudio/.* [NC]
+    RewriteRule /proxy/[^/]*/[^/]*/rstudio/(.*) http://127.0.0.1:8001/$1 [P,L]
+
+    ProxyPass / http://127.0.0.1:8001/ retry=10
```
rtitle: I'm confused why we `ProxyPass /` here as well as below in the legacy Jupyter section. I would expect to have a more specific `ProxyPass` for RStudio, and do `/` in the catch-all. Something like:
```
# RStudio 

<rewrite rules>

ProxyPass /proxy/[^/]*/[^/]*/rstudio/ http://127.0.0.1:8001/
ProxyPassReverse /proxy/[^/]*/[^/]*/rstudio/ http://127.0.0.1:8001/

# Jupyter (legacy path)

<rewrite rules>

ProxyPass / http://127.0.0.1:8000/ retry=10
ProxyPassReverse / http://127.0.0.1:8000/
```
But if this works then maybe my understanding is incorrect..

rtitle:
I think this all looks good, I don't really have any PR comments. But I'm holding my thumb because I want to do some manual QA to verify that all the paths, etc are working. I'm spinning up a fiab now, stay tuned..

src/main/resources/swagger/api-docs.yaml
```diff
@@ -640,6 +642,187 @@ paths:
           description: Internal Error
           schema:
             $ref: '#/definitions/ErrorReport'
+  '/proxy/{googleProject}/{clusterName}/jupyter/':
```
rtitle: Minor: remove this trailing slash in Swagger

rtitle (approved):
Testing update. I tried the following paths to see if they work to access Jupyter/Lab/RStudio:
```
https://leonardo-fiab.dsde-dev.broadinstitute.org:30443/notebooks/d-proj-2019-02-27/rt-test-jupyter-only  -- no
https://leonardo-fiab.dsde-dev.broadinstitute.org:30443/notebooks/d-proj-2019-02-27/rt-test-jupyter-only/ -- yes
https://leonardo-fiab.dsde-dev.broadinstitute.org:30443/proxy/d-proj-2019-02-27/rt-test-jupyter-only/jupyter -- yes
https://leonardo-fiab.dsde-dev.broadinstitute.org:30443/proxy/d-proj-2019-02-27/rt-test-jupyter-only/jupyter/ -- yes
https://leonardo-fiab.dsde-dev.broadinstitute.org:30443/notebooks/d-proj-2019-02-27/rt-test-jupyter-only/lab -- yes
https://leonardo-fiab.dsde-dev.broadinstitute.org:30443/notebooks/d-proj-2019-02-27/rt-test-jupyter-only/lab/ -- yes
https://leonardo-fiab.dsde-dev.broadinstitute.org:30443/proxy/d-proj-2019-02-27/rt-test-jupyter-only/jupyter/lab -- yes
https://leonardo-fiab.dsde-dev.broadinstitute.org:30443/proxy/d-proj-2019-02-27/rt-test-jupyter-only/jupyter/lab/ -- yes
https://leonardo-fiab.dsde-dev.broadinstitute.org:30443/proxy/d-proj-2019-02-27/rt-test-rstudio-only/rstudio -- yes
https://leonardo-fiab.dsde-dev.broadinstitute.org:30443/proxy/d-proj-2019-02-27/rt-test-rstudio-only/rstudio/ -- yes
```
The only one that didn't work was the first one, which is a legacy Jupyter path with no trailing slash. Fix this if it's easy; but it's probably not user facing since Terra uses either get linked directly to a notebook or to the `/tree` path (which works).

I also tested creating clusters that run Jupyter only, RStudio only, and Jupyter+RStudio. All worked loading their respective tools. I also pause+resumed each cluster -- the RStudio-only one never resumed but I think this is a known issue which @akarukappadath is working on. Both Jupyter clusters came back successfully.

So pending auto tests (and that `/` issue) :shipit:!!

MatthewBemis:
see https://github.com/broadinstitute/firecloud-develop/pull/1546#issuecomment-468160803

PR #721, opened 2018-12-14
Issue 404: Install TOC extension by default

vkumra-broad:

In all cases:


docker/jupyter/scripts/extension/install_TOC.sh
```diff
@@ -0,0 +1 @@
+#!/usr/bin/env bash
```
rtitle: I think this script needs more :)

src/main/resources/jupyter/init-actions.sh
```diff
@@ -247,6 +247,9 @@ if [[ "${ROLE}" == 'Master' ]]; then
       done
     fi
 
+
+    retry 3 docker exec -u root -e PIP_USER=false ${JUPYTER_SERVER_NAME} ${JUPYTER_SCRIPTS}/extension/install_TOC.sh
```
rtitle: Probably should remove the `-u root` -- I think we install extensions as `jupyter-user`.

vkumra-broad:
I am not sure if I like approach, I would rather have this done via a user script than installing the extension by default. Thoughts?

docker/jupyter/scripts/extension/install_TOC.sh
vkumra-broad: this is what happens when you work during the plenary :)

rtitle:
I'd also lean towards a user script approach for more customizability than doing it by default. Could loop in @sharmatime too.

sharmatime:
Lets review this tomorrow (Thursday) Vikram

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/NotebookInstallSpec.scala
```diff
@@ -17,7 +17,7 @@ class NotebookInstallSpec extends ClusterFixtureSpec {
               |bx.bitset.sys.copyright""".stripMargin
 
           notebookPage.executeCell("1+1") shouldBe Some("2")
-          notebookPage.executeCell(getPythonVersion) shouldBe Some("3.6.7")
+          notebookPage.executeCell(getPythonVersion) shouldBe Some("3.6.8")
```
vkumra-broad: There was a python release recently. This change reflects that.

vkumra-broad: https://www.python.org/downloads/release/python-368/

vkumra-broad:
jenkins retest

vkumra-broad:
jenkins retest

MatthewBemis:
what is the reason for doing this by default instead of via user script like was discussed above? i really don't think that we should be fielding requests for what's installed by default. i get that this extension is fairly trivial and useful but i think in general, if somebody wants an extension installed by default while using terra then it should be terras responsibility for installing that extension. the customizability is there for terra to do so, and AoU and other apps might not want these things

vkumra-broad:
We went with by default installation since these are commonly used extensions. AoU also uses these extensions for there "getting started notebook", it's just that they install it themselves.

PR #723, opened 2018-12-18
Just compare clusters by id

rtitle:

In all cases:


src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorSupervisor.scala
```diff
@@ -173,7 +173,8 @@ class ClusterMonitorSupervisor(monitorConfig: MonitorConfig, dataprocConfig: Dat
       .inTransaction { _.clusterQuery.listMonitoredFullCluster() }
       .onComplete {
         case Success(clusters) =>
-          val clustersNotAlreadyBeingMonitored = clusters.toSet -- monitoredClusters
```
rtitle: I think this check was problematic because the cluster in the database can differ depending on what stage of monitoring it's in. Comparing `id` should be stable.

PR #726, opened 2018-12-19
Horizontal Scaling : checkCluster should not use full cluster query

vkumra-broad:
.... this causes bad things to happen to the DB. 

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorSupervisor.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorSupervisor.scala
@@ -89,21 +89,23 @@ class ClusterMonitorSupervisor(monitorConfig: MonitorConfig, dataprocConfig: Dat
         dbRef.inTransaction { dataAccess =>
           dataAccess.clusterQuery.getClusterById(cluster.id)
         }.flatMap {
-          case Some(cluster) => val clusterRequest = ClusterRequest(
-            Option(cluster.labels),
-            cluster.jupyterExtensionUri,
-            cluster.jupyterUserScriptUri,
-            Some(cluster.machineConfig),
-            None,
-            cluster.userJupyterExtensionConfig,
-            if (cluster.autopauseThreshold == 0) Some(false) else Some(true),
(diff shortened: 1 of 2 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorSupervisor.scala
```diff
@@ -86,19 +86,25 @@ class ClusterMonitorSupervisor(monitorConfig: MonitorConfig, dataprocConfig: Dat
...
+            None,
+            cluster.userJupyterExtensionConfig,
+            if (cluster.autopauseThreshold == 0) Some(false) else Some(true),
+            Some(cluster.autopauseThreshold),
+            cluster.defaultClientId,
+            cluster.clusterImages.find(_.tool == Jupyter).map(_.dockerImage))
+            leonardoService.internalCreateCluster(cluster.auditInfo.creator, cluster.serviceAccountInfo, cluster.googleProject, cluster.clusterName, clusterRequest).failed.foreach { e =>
+              logger.error(s"Error occurred recreating cluster ${cluster.projectNameString}", e)
+            }
+            Future.successful(())
```
rtitle: Indentation is a bit off. Also I don't think you need the final `Future.successful(())` -- can just return the result of `leonardoServer.internalCreateCluster`.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorSupervisor.scala
```diff
@@ -172,7 +178,7 @@ class ClusterMonitorSupervisor(monitorConfig: MonitorConfig, dataprocConfig: Dat
     val monitoredClusterIds = monitoredClusters.map(_.id)
     
     dbRef
-      .inTransaction { _.clusterQuery.listMonitoredFullCluster() }
+      .inTransaction { _.clusterQuery.listMonitored() }
```
rtitle: This doesn't do the joins, correct?

rtitle (changes requested):
1 minor comment

rtitle (approved):
👍 if tests pass

PR #759, opened 2019-01-28
[NOMERGE] Use python 3.7 on the worker nodes only

rtitle:
Experimental to see what happens. This will make the master node use the 3.6 image and the workers user 3.7. I suspect Spark/Hail won't work but it might be a temporary solution to fix cluster creation.

In all cases:


rtitle:
jenkins retest

rtitle:
https://fc-jenkins.dsp-techops.broadinstitute.org/job/leonardo-fiab-test-runner/9425/

The PySpark 3 test failed as expected, but other tests look ok (it was affected by the `us-central1-f` issue which is a separate but known issue).

rtitle:
Closing without merging

PR #766, opened 2019-02-01
Bash script for hotswapping leo jar on fiab

akarukappadath:
Wrote a bash script to perform hotswap of leo jar for a fiab. Run using 

./automation/hotswap.sh fiab-your-fiab-name

from leonardo. 

In all cases:


automation/hotswap.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+
```
rtitle: Could you add some comments at the top describe what the script is for, what args to pass, etc?

Also maybe some code that prints the usage if the arg isn't specified.

These are just good practices for bash scripts.

rtitle (approved):
Looks good!

PR #767, opened 2019-02-04
Issue 757 : Maintain state in RStudio

vkumra-broad:

In all cases:

### DIFF
--- a/docker/rstudio/Dockerfile
+++ b/docker/rstudio/Dockerfile
@@ -27,7 +27,7 @@ ENV RSTUDIO_HOME /etc/rstudio
 ADD rserver.conf $RSTUDIO_HOME/rserver.conf
 
 RUN sed 's/alwaysSaveHistory="0"/alwaysSaveHistory="1"/g' \
- && sed 's/loadRData="0"="0"/loadRData="0"="1"/g' \
+ && sed 's/loadRData="0"/loadRData="1"/g' \
  && sed 's/saveAction="0"/saveAction="1"/g'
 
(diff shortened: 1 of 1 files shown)


docker/rstudio/Dockerfile
```diff
@@ -26,4 +26,8 @@ ENV RSTUDIO_HOME /etc/rstudio
 
 ADD rserver.conf $RSTUDIO_HOME/rserver.conf
 
+RUN sed 's/alwaysSaveHistory="0"/alwaysSaveHistory="1"/g' \
```
rtitle: Don't you have to pass the filename to `sed`? I would expect something like
```
RUN sed -i 's/alwaysSaveHistory="0"/alwaysSaveHistory="1"/g' <the-filename> \
  && sed -i 's/loadRData="0"/loadRData="1"/g' <the-filename> \
  && sed -i 's/saveAction="0"/saveAction="1"/g' <the-filename>
```

MatthewBemis:
another scenario we should test is deleting a cluster and trying to restore that session on a new cluster

vkumra-broad:
Faces with @MatthewBemis :  the delete and restore on new cluster would be a good once we have gcsfuse integrated.

My manual test scenario in this case is:

1. Create cluster
2. Create a new Rstudio proj
3. Run some commands, load some data. this will populate the .Rhistory and .Rdata files. 
4. Close the session
5. Restart the cluster. 
6. Rstudio should open the previous project with data and history available from the previous session.

MatthewBemis (approved):

Issue #742, opened 2019-01-22
JupyterLab Extension From Single File


blrubenstein:
Currently, JupyterLab extensions require a full tarball with all the NPM package setup boilerplate, and that should probably live in Leo so that every app doesn't have to maintain the same boilerplate folder template.

blrubenstein:
This is the template, only the index file should be required, it should be copied and replaced. The template should be saved onto the Leo image.

https://github.com/jupyterlab/extension-cookiecutter-ts

rtitle:
While this feature makes sense, I didn't see it as a requirement for the original work. We made the choice for Leo to support JL extensions in the same format as JL itself: e.g. a public npm package or a tarball of an npm package. See: https://jupyterlab.readthedocs.io/en/stable/user/extensions.html#installing-extensions. In other words, Leo is just a pass-through to JupyterLab, and we give callers full control over their extension code.

Is it a problem for AoU or Terra to host a fully packaged extension tarball? Again, `extension-cookiecutter-ts` seems like a nice feature to add, I just don't understand why it's a blocker. Thanks

blrubenstein:
I don't think it is actually a blocker... I will remove those. I do know the Terra devs are making a choice to block their development work on this feature, because a packaged tarball can't actually sit in source control cleanly. I think it would be possible to host the structure in GCS, and just upload the typescript/javascript file on its own to that file structure. I will play around with that myself tomorrow probably, but defer to @zarsky-broad for the Terra perspective, because I know he has done a lot more research into this.

rtitle:
Would a potential architecture be for Terra/AoU to build and package their npm packages as part of their CI/CD pipeline, and pass an artifactory URL to Leo? (Or alternatively, publish the extension to npm?) I'm a little hesitant to make assumptions in Leo about extension package structure, etc. It might be beneficial for apps to have full control over their extension code & build process, instead of splitting logic between apps and Leo. Happy to talk with faces tomorrow.

blrubenstein:
Yeah, lets grab some face time today.

rtitle:
Deployed on 2/27/19

PR #778, opened 2019-02-13
Specify JupyterLab extension from single JS file 

akarukappadath:
Terra provides their extension by hosting a JS file. This ticket supports this by inserting a provided JS file into a generic JupyterLab extension template (https://github.com/jupyterlab/extension-cookiecutter-js). 

This was manually tested by creating a cluster with a test JS file placed in a Google bucket and specifying the bucket in the userJupyterExtensionConfig. Since the logic is the same for extensions specified through a URL, this test should be sufficient. 

In all cases:

### DIFF
--- a/docker/jupyter/scripts/extension/jupyter_install_lab_extension.sh
+++ b/docker/jupyter/scripts/extension/jupyter_install_lab_extension.sh
@@ -5,6 +5,15 @@ set -e
 
 if [ -n "$1" ]; then
   JUPYTER_EXTENSION=$1
-  jupyter labextension install $JUPYTER_EXTENSION
+  if [[ ${JUPYTER_EXTENSION} == *'.js' ]]; then
+    JUPYTER_EXTENSION_NAME=`basename ${JUPYTER_EXTENSION%%.*}`
+    cookiecutter --no-input https://github.com/jupyterlab/extension-cookiecutter-ts extension_name=${JUPYTER_EXTENSION_NAME}
+    cp ${JUPYTER_EXTENSION} ${JUPYTER_EXTENSION_NAME}/lib/plugin.js
+    cd ${JUPYTER_EXTENSION_NAME}
+    jlpm
+    jupyter labextension install . --no-build
+  else
+    jupyter labextension install $JUPYTER_EXTENSION
+  fi
 fi
 
--- a/docker/jupyter/scripts/extension/jupyter_install_lab_extension.sh
+++ b/docker/jupyter/scripts/extension/jupyter_install_lab_extension.sh
@@ -14,6 +14,14 @@ if [ -n "$1" ]; then
     cd ${JUPYTER_EXTENSION_NAME}
     jlpm
     jupyter labextension install .
+  elif [[ ${JUPYTER_EXTENSION} == *'.ts' ]]; then
+    # same as above but in typescript, see https://github.com/jupyterlab/extension-cookiecutter-ts
+    JUPYTER_EXTENSION_NAME=`basename ${JUPYTER_EXTENSION%%.*}`
+    cookiecutter --no-input https://github.com/jupyterlab/extension-cookiecutter-ts extension_name=${JUPYTER_EXTENSION_NAME}
+    cp -f ${JUPYTER_EXTENSION} ${JUPYTER_EXTENSION_NAME}/src/index.ts
+    cd ${JUPYTER_EXTENSION_NAME}
+    jlpm
+    jupyter labextension install .
   else
     jupyter labextension install $JUPYTER_EXTENSION
   fi
--- a/docker/jupyter/scripts/extension/jupyter_install_lab_extension.sh
+++ b/docker/jupyter/scripts/extension/jupyter_install_lab_extension.sh
@@ -6,12 +6,14 @@ set -e
 if [ -n "$1" ]; then
   JUPYTER_EXTENSION=$1
   if [[ ${JUPYTER_EXTENSION} == *'.js' ]]; then
+    # use jupyterlab extension template to create an extension using the specified JS file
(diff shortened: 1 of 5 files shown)


docker/jupyter/scripts/extension/jupyter_install_lab_extension.sh
```diff
@@ -5,6 +5,17 @@ set -e
 
 if [ -n "$1" ]; then
   JUPYTER_EXTENSION=$1
-  jupyter labextension install $JUPYTER_EXTENSION
+  if [[ ${JUPYTER_EXTENSION} == *'.js' ]]; then
```
rtitle: I had thought we want to support `*.ts` extensions (and use https://github.com/jupyterlab/extension-cookiecutter-ts). Maybe check with @zarsky-broad -- should we support `.js`, `.ts`, or both?

docker/jupyter/scripts/extension/jupyter_install_lab_extension.sh
```diff
@@ -5,6 +5,17 @@ set -e
...
+    # use jupyterlab extension template to create an extension using the specified JS file
+    # see https://github.com/jupyterlab/extension-cookiecutter-js
+    JUPYTER_EXTENSION_NAME=`basename ${JUPYTER_EXTENSION%%.*}`
+    cookiecutter --no-input https://github.com/jupyterlab/extension-cookiecutter-js extension_name=${JUPYTER_EXTENSION_NAME}
+    cp -f ${JUPYTER_EXTENSION} ${JUPYTER_EXTENSION_NAME}/lib/plugin.js
```
rtitle: If typescript I think it's `src/index.ts`

docker/jupyter/scripts/extension/jupyter_install_lab_extension.sh
```diff
@@ -5,6 +5,17 @@ set -e
...
+    cd ${JUPYTER_EXTENSION_NAME}
+    jlpm
```
rtitle: In this [tutorial](https://jupyterlab.readthedocs.io/en/stable/developer/xkcd_extension_tutorial.html) it's:
```
cd $ext
jlpm install
jupyter labextension install . --no-build
```
Is that equivalent? (I _think_ so?)

rtitle:
Looks good - confirm the `js` vs `ts` question though. In the issue Scout had specified `ts`: https://github.com/DataBiosphere/leonardo/issues/742

Also I'm curious how you've tested this -- is there a js extension you've successfully plugged into lab? Should it be an automation test?

akarukappadath:
I used an example js extension that basically writes to the console that the extension has been activated - it's the example given when you use the cookiecutter to make a extension. I had placed it in a bucket and specified it when making the cluster. Didn't think an automation test made sense since that would mean that we would have to maintain a bucket containing this js file or have the js file be part of Leo. Let me know what you think.

docker/jupyter/scripts/extension/jupyter_install_lab_extension.sh
akarukappadath: just checked with Isaac, it's js!

rtitle:
We do have tests that place items in buckets and pass gs:// paths to Leo. Here's an example: https://github.com/DataBiosphere/leonardo/blob/develop/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/NotebookCustomizationSpec.scala#L61-L80

While I'm hesitant to add more cluster creation tests I do feel like this might deserve an auto-test. It exercises a unique code path (using `cookiecutter` and `jlpm`) and is functionality that Terra will rely upon. We should have a test to make sure it doesn't regress when we update the Jupyter image, for example.

So your test could
a) place the file in a bucket
b) pass the gs:// path to Leo
c) verify the extension is installed by doing `!jupyter labextension list` from a notebook

docker/jupyter/scripts/extension/jupyter_install_lab_extension.sh
akarukappadath: I just didn't want to specify --no-build because the user would have to reload the JupyterLab page once the cluster is ready. Other than that it should be equivalent!

docker/jupyter/scripts/extension/jupyter_install_lab_extension.sh
rtitle: Sorry to change my mind -- I was thinking maybe we should just support `js` or `ts` since it's easy and will make leo more flexible. So: 
```
if [[ ${JUPYTER_EXTENSION} == *'.js' ]]; then
  <current stuff>
elif [[ ${JUPYTER_EXTENSION} == *'.ts' ]]; then
  JUPYTER_EXTENSION_NAME=`basename ${JUPYTER_EXTENSION%%.*}`
  cookiecutter --no-input https://github.com/jupyterlab/extension-cookiecutter-ts extension_name=${JUPYTER_EXTENSION_NAME}
  cp -f ${JUPYTER_EXTENSION} ${JUPYTER_EXTENSION_NAME}/src/index.ts
  cd ${JUPYTER_EXTENSION_NAME}
  jlpm
  jupyter labextension install .
else 
  <current stuff>
fi
```

I am ok with only having an auto test for `.js` if that's what Terra will be using.

akarukappadath: Sounds good - just added that!

rtitle (approved):
This looks good, except the tests aren't happy. Looks like a transient docker build thing -- I would jenkins retest.

akarukappadath:
jenkins retest

akarukappadath:
jenkins retest

akarukappadath:
jenkins retest

Issue #770, opened 2019-02-08
Token refresh not working from JupyterLab


rtitle:
See ZenDesk ticket from Nicole: https://broadinstitute.zendesk.com/agent/tickets/970

I suspect the front-end token refresh which runs as a Jupyter extension is not running lab, causing users' tokens to expire and see 401 errors. We should resolve this, either by:

- providing the same code/functionality as a labextension
- having Terra UI call /setCookie periodically in the background

schaluva:
Nicole also reached out to mention that she experiences the 401 errors when actively using JupyterLab (not just after leaving the computer for a while).

mbookman:
Cross-linking with the saturn-ui issue: https://github.com/DataBiosphere/saturn-ui/issues/1116
Presumably this is the same issue. I'm unclear whether it is a Leonardo issue or Terra.

rtitle:
@mbookman yes I think this is the same Leo issue. We're hoping to have a fix out next week.

In the meantime an (unforunate) workaround is to have a Terra notebook (not lab) open in a different tab while you use Lab.

mbookman:
Thanks! That's a simple workaround to use for now.
Probably also explains why I would see this intermittently, as I often have this very setup going for other reasons.

rtitle:
Update: we have a fix for this and are targetting a release next week (Tue).

rtitle:
This fix is deployed to production as of 2/27/19 5pm. However note you will need to rebuild your cluster in order to obtain the fix.

PR #787, opened 2019-02-22
[770] Token refresh not working from JupyterLab

rtitle:
**Issue**: users were experiencing 401 errors when using JupyterLab for extended periods of time. See issue https://github.com/DataBiosphere/leonardo/issues/770. The root cause was that the Jupyter extension which refreshes the user's credentials (as described [here](https://github.com/DataBiosphere/leonardo/wiki/Connecting-to-a-Leo-Notebook#token-refresh)) was only running in Jupyter, not JupyterLab.

**Change**: this PR refactors the existing `google_sign_in.js` code so it is usable by both Jupyter _and_ JupyterLab, and does the necessary wiring to plug it into both applications. See inline comments below for more details.

**Testing**: we don't have automation tests for this feature, but I manually validated Jupyter and Lab by looking at the browser console for calls to `https://accounts.google.com/o/oauth2/iframerpc?action=issueToken` every minute.

In all cases:


src/main/resources/jupyter/google_plugin_jupyterlab.js
```diff
@@ -0,0 +1,7 @@
+module.exports = [{
+    id: 'google_plugin_jupyterlab',
+    autoStart: true,
+    activate: function(app) {
+        require('/home/jupyter-user/.jupyter/custom/google_sign_in');
```
rtitle: This is the entrypoint for JupyterLab extensions. In this case it just loads another file.

Note this has to be an absolute path, as for relative paths it defaults to looking for node packages in some other directory.

src/main/resources/jupyter/google_sign_in.js
```diff
@@ -1,14 +1,3 @@
-require.config({
```
rtitle: No longer using [RequireJS](https://requirejs.org/) to load the Google JS client [library](https://developers.google.com/api-client-library/javascript/start/start-js). JupyterLab doesn't include RequireJS because it doesn't interact nicely with node. (There is a node [library](https://requirejs.org/docs/node.html) to support RequireJS but it broke like 10 other things when I tried to install it.)

src/main/resources/jupyter/google_sign_in.js
```diff
@@ -80,7 +69,27 @@ function set_cookie(token, expires_in) {
     document.cookie = "LeoToken="+token+";secure;expires="+expiresDate.toUTCString()+";path=/";
 }
 
+function loadGapi(google_lib, continuation) {
```
rtitle: This is the non-requireJS way to load the Google JS client library.

src/main/resources/jupyter/google_sign_in.js
```diff
@@ -42,27 +41,27 @@ function receive(event) {
...
+                    set_cookie(result.access_token, result.expires_in);
+                });
             }
+        }
 
-            // refresh token every 2 minutes
-            setInterval(doAuth, 120000);
-        });
+        // refresh token every 2 minutes
+        console.log('Starting token refresh timer');
```
rtitle: Also added some log messages so we can more easily tell if it's working.

rtitle:
jenkins retest

rtitle:
jenkins retest

PR #799, opened 2019-03-01
#801: Install python dev libs in RStudio container

MatthewBemis:

In all cases:

### DIFF
--- a/docker/rstudio/Dockerfile
+++ b/docker/rstudio/Dockerfile
@@ -21,6 +21,11 @@ RUN R -e 'install.packages(c( \
     repos="http://cran.mtu.edu")' \
  && R -e 'devtools::install_github("DataBiosphere/Ronaldo")'
 
+# install python dev libs
+RUN apt-get update
+RUN apt-get install -y libpython-dev
+RUN apt-get install -y libpython3-dev
+
 ENV RSTUDIO_PORT 8001
 ENV RSTUDIO_HOME /etc/rstudio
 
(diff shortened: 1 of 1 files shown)


docker/rstudio/Dockerfile
```diff
@@ -21,6 +21,11 @@ RUN R -e 'install.packages(c( \
     repos="http://cran.mtu.edu")' \
  && R -e 'devtools::install_github("DataBiosphere/Ronaldo")'
 
+# install python dev libs
```
rtitle: Could you put this in the above layer which starts with `RUN apt-get update`? In general fewer layers is better for docker images because it reduces image size, etc.

docker/rstudio/Dockerfile
```diff
@@ -21,6 +21,11 @@ RUN R -e 'install.packages(c( \
...
+RUN apt-get update
+RUN apt-get install -y libpython-dev
+RUN apt-get install -y libpython3-dev
```
rtitle: Does this include `pip`? Another way to install it is:
https://pip.pypa.io/en/stable/installing/

docker/rstudio/Dockerfile
MatthewBemis: 💯

docker/rstudio/Dockerfile
MatthewBemis: no, but it makes pip easily installable at the very least. i'll add it to the image

rtitle (approved):
LGTM

PR #809, opened 2019-03-06
Check that all specified cluster tools are ready before setting cluster to Running from Starting

akarukappadath:
When a cluster transitions from Starting to Running, we originally checked to see if the Jupyter proxy was up and ready. This PR goes through the cluster's clusterImages and checks the proxy of each tool before setting the cluster status to Running. 

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
@@ -146,18 +146,11 @@ trait ClusterComponent extends LeoComponent {
       }
     }
 
-//    def listMonitoredClusterOnly(): DBIO[Seq[Cluster]] = {
-//      clusterQuery.filter{_.status inSetBind ClusterStatus.monitoredStatuses.map(_.toString) }.result map { recs =>
-//        recs.map(rec => unmarshalCluster(rec, Seq.empty, List.empty, Map.empty, List.empty, List.empty, List.empty))
-//      }
-//    }
-
     def listMonitoredClusterOnly(): DBIO[Seq[Cluster]] = {
-      clusterImagesQuery.filter{_._1.status inSetBind ClusterStatus.monitoredStatuses.map(_.toString) }.result map { recs =>
-        unmarshalImageCluster(recs)
+      clusterQuery.filter{_.status inSetBind ClusterStatus.monitoredStatuses.map(_.toString) }.result map { recs =>
+        recs.map(rec => unmarshalCluster(rec, Seq.empty, List.empty, Map.empty, List.empty, List.empty, List.empty))
       }
     }
-
 
     def listMonitored(): DBIO[Seq[Cluster]] = {
       clusterLabelQuery.filter { _._1.status inSetBind ClusterStatus.monitoredStatuses.map(_.toString) }.result map { recs =>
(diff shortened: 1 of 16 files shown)


src/test/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorSpec.scala
```diff
@@ -847,69 +858,8 @@ class ClusterMonitorSpec extends TestKit(ActorSystem("leonardotest")) with FlatS
       // starting a cluster should not touch IAM roles
       verify(iamDAO, never()).removeIamRolesForUser(any[GoogleProject], any[WorkbenchEmail], mockitoEq(Set("roles/dataproc.worker")))
       verify(iamDAO, never()).removeServiceAccountKey(any[GoogleProject], any[WorkbenchEmail], any[ServiceAccountKeyId])
-    }
-  }
-
-  // Pre:
```
akarukappadath: This test wasn't actually testing anything - since the cluster started in Starting status and ended in Starting status, the eventually immediately succeeded, meaning the clusterMonitorActor didn't do anything (so the JupyterDAO and isProxyAvailable was never interacted with)

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
```diff
@@ -146,12 +146,19 @@ trait ClusterComponent extends LeoComponent {
...
+//    def listMonitoredClusterOnly(): DBIO[Seq[Cluster]] = {
+//      clusterQuery.filter{_.status inSetBind ClusterStatus.monitoredStatuses.map(_.toString) }.result map { recs =>
+//        recs.map(rec => unmarshalCluster(rec, Seq.empty, List.empty, Map.empty, List.empty, List.empty, List.empty))
+//      }
+//    }
+
     def listMonitoredClusterOnly(): DBIO[Seq[Cluster]] = {
-      clusterQuery.filter{_.status inSetBind ClusterStatus.monitoredStatuses.map(_.toString) }.result map { recs =>
-        recs.map(rec => unmarshalCluster(rec,Seq.empty, List.empty, Map.empty, List.empty, List.empty, List.empty))
+      clusterImagesQuery.filter{_._1.status inSetBind ClusterStatus.monitoredStatuses.map(_.toString) }.result map { recs =>
```
vkumra-broad: It might be easier and faster to get the to use `getAllForCluster` from `clusterImageQuery` to get the images for each cluster.

akarukappadath:
jenkins retest

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
```diff
@@ -294,9 +296,11 @@ class ClusterMonitorActor(val cluster: Cluster,
               // Otherwise, when a cluster is resumed and transitions from Starting to Running, we get stuck
               // in that state - at least with the way HttpJupyterDAO.isProxyAvailable works
               dbRef.inTransaction { _.clusterQuery.updateClusterHostIp(cluster.id, Some(ip)) }.flatMap { _ =>
-                jupyterProxyDAO.isProxyAvailable(cluster.googleProject, cluster.clusterName).map {
-                  case true => ReadyCluster(ip, googleInstances)
-                  case false => NotReadyCluster(ClusterStatus.Running, googleInstances)
+                dbRef.inTransaction { _.clusterImageQuery.getAllForCluster(cluster.id)}.flatMap { images =>
```
MatthewBemis: if you turn this `flatMap` into a `map`, you won't need to wrap your resulting values in a `Future.successful()`

for a bit more explanation: 

you can think of `flatMap` as taking the result of the previous operation, pulling it out of the box (in this case, the box is a Future), and then running the following code without putting it back inside of the box. because your result isn't getting put in the box again, you have to do it manually by wrapping the result with `Future.successful()`

`map` works similarly, but it will put the result back inside of the box. so if you switch the above to a `map`, you don't have to manually wrap it in a `Future.successful()`.

that aside, for what i believe to be a more Scala-idiomatic way of writing your if/else, I would use `Future.traverse()`. you could do something like:

```
Future.traverse(images) { image => isProxyAvailable(image.tool) }.map { isAvailable =>
    if(isAvailable.forall(_ == true)) ReadyCluster(ip, googleInstances)
    else NotReadyCluster(ClusterStatus.Running, googleInstances)
}
```

note that this is using map, so there's no need to wrap the results in `Future.successful()`

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
```diff
@@ -320,6 +324,14 @@ class ClusterMonitorActor(val cluster: Cluster,
     } yield result
   }
 
+  private def isProxyAvailable(clusterTool: ClusterTool): Future[Boolean] = {
+    val toolDAO : ToolDAO = clusterTool match {
+      case ClusterTool.Jupyter => jupyterProxyDAO
+      case _ => rstudioProxyDAO // there are only two tools and it is impossible to create a cluster with no tools
```
MatthewBemis: i would still recommend enumerating both of our supported tools _and_ throwing an exception in the case of an unrecognized tool

the benefit is that when we add a third tool, this bit of code will immediately fail and we can update it instead of it being masked because it's defaulting to RStudio. in other words, it means that the code will be more resilient to human-error

MatthewBemis (changes requested):

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
akarukappadath: good point!

akarukappadath:
jenkins retest

MatthewBemis (approved):

Issue #704, opened 2018-12-03
Have workspace bucket name available via an environment variable


deflaux:
This will allow us to programmatically write files such as images to the workspace bucket and then view them using Terra's data tab.

rtitle:
Good idea

mbookman:
+1

sharmatime:
Could be a number of approaches, so needs a tech doc

Bucket has long ID
Clusters are per project, but buckets are per workspace
Within a notebook you know what workspace you're in
Maybe when the workspace localizes the notebook to the cluster,

bradtaylor:
It would also be good to have the workspace name itself there, because you can use that to get other attributes (like the bucket) with FISS if working in python. Could the Terra team run an extension that sets this variable?

rtitle:
One idea is for Terra to localize an `.env` file to the workspace directory on the cluster when it localizes the notebooks. Then users would be able to read this `.env` file in code to access workspace-specific variables.

Another idea I've been toying with: what if Terra workspaces by default included a "skeleton" notebook which sets up the environment, and maybe includes documentation about how to get started with things like FISS, Hail, BigQuery, etc. That might be a nice experience for users.

mbookman:
In https://github.com/DataBiosphere/saturn-ui/issues/1039 I noted that if we can get the workspace name, then we can get everything else about the workspace through API calls. The 1-1 mapping of workspace and bucket seems both durable and useful, so including that as well in the same manner as the workspace name would certainly be nice.

My one question is - if there were a `.env` file, how would I reliably access it from a notebook? Is there a environment variable, path, or function which would give me that? In thinking about workoing around https://github.com/DataBiosphere/saturn-ui/issues/1039, one option was to have a notebook figure out from either the current working directory or the notebook file's directory what the path was and then extract the workspace name. But this seemed unreliable and I'm concerned that dropping down a `.env` file without additional support is just the same problem.

rtitle:
Still testing, but here's an early PR for this: https://github.com/DataBiosphere/leonardo/pull/841

This should provide `WORKSPACE_NAMESPACE`, `WORKSPACE_NAME`, and `WORKSPACE_BUCKET` EVs in a notebook (python and R kernels).

rtitle:
[...]

PR #841, opened 2019-03-29
Set workspace name/bucket in environment variable

rtitle:
My attempt at:
https://github.com/databiosphere/leonardo/issues/704
https://github.com/DataBiosphere/terra-ui/issues/986

Acceptance criteria: the user is able to access the following environment variables from a notebook (python or R):
- `WORKSPACE_NAMESPACE` -- the google project, e.g. `fc-product-demo`
- `WORKSPACE_NAME` -- the workspace name, e.g. `try-saturn`
- `WORKSPACE_BUCKET` -- the workspace bucket, e.g. `gs://fc-4205396d-2cde-4436-8903-9b374dac5bbf`

The general approach here is to parse the CWD of the notebook to get the workspace name. By convention Terra and AoU localizes notebooks to directories named by the workspace. This approach won't work if we start launching notebooks from other directories. 

The above is done at kernel startup time. It needs to be kernel startup, not cluster startup, because kernels start in the context of a workspace, while clusters start in the context of a project. I achieved this by creating a `kernel_bootstrap.sh` script which sets the appropriate EVs then starts the kernel. I also investigated doing this with a [python startup script](https://stackoverflow.com/questions/45818538/where-can-i-put-a-startup-script-in-jupyter). This worked, but (a) it was python only, so a different solution was needed for R; and (b) we don't need to modify the kernel state, just set EVs, so a bash script that can be commonly used for Python and R is appropriate for this.

A future enhancement might be to not use the notebook CWD, and instead pass information from Terra UI to the cluster somehow. This might be less brittle, but the general approach of using `kernel_bootstrap.sh` should still work. So I think this is a good first pass, and will add value to users today.

In all cases:

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookPyKernelSpec.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookPyKernelSpec.scala
@@ -142,8 +142,7 @@ class NotebookPyKernelSpec extends ClusterFixtureSpec with NotebookTestUtils {
               notebookPage.executeCell("! echo $GOOGLE_PROJECT").get shouldBe clusterFixture.billingProject.value
               notebookPage.executeCell("! echo $WORKSPACE_NAMESPACE").get shouldBe clusterFixture.billingProject.value
               notebookPage.executeCell("! echo $WORKSPACE_NAME").get shouldBe "Untitled Folder"
-              // TODO: we obtain the workspace bucket using FISS which is not yet supported in tests.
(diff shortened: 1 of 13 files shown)


rtitle:
jenkins retest

rtitle:
jenkins retest

mbookman:
Would also resolve https://github.com/DataBiosphere/terra-ui/issues/1039.

rtitle:
jenkins retest

rtitle:
jenkins retest

rtitle:
jenkins retest

rtitle:
docker build is having issues with keyserver :( jenkins retest

rtitle:
one more jenkins retest

Issue #844, opened 2019-03-29
Updating compute runtime while a cluster is running does not result in updating status


sharmatime:
When users change runtime to add start- up script and click update the compute runtime doesn’t indicate that the runtime is being re-created. I had to refresh the page to see the updated status

Desired user outcome:
 - The user is able to easily understand that their cluster is updating without the confusion of having to refresh the page 
 - UI displays the proper indicator of cluster being updated without refreshing the page.

Might be related to https://github.com/DataBiosphere/leonardo/issues/821

Issue #821, opened 2019-03-11
Newly created cluster is absent from list clusters response


MatthewBemis:
Hi Leo folks: A while back we chatted about a slight lag between requesting a status change on a cluster and seeing that change reflected back. I just observed this happening while creating a cluster. Here's a successful 'create', followed by a 'list' that shows no results. The cluster does show up in the list eventually, but it takes at least several seconds.

I don't remember exactly where we left this, but it would be great to look into the cause when you have some bandwidth. Thanks!

sharmatime:
From Rob:
**List clusters returns stale results due to Sam caching** 
 - Brett has complained about this, and Rob’s experienced this in UI
 - In UI, if you hit the play button, it doesn’t change to creating or starting for some time 
 - UI is asking Leo for status, but Leo is returning stale results
 - Rob suspects SAM caching at the root of this

User outcome:
 - The User should observe the status change without lag time, so they are not confused about the state of their cluster

rtitle:
This was released last week. Closing.

PR #849, opened 2019-04-03
List clusters stale results fix

akarukappadath:
Background: When calling the list clusters endpoint, it would fail to include recently created clusters because of a call to Sam during cluster creation that is not transactional. This fix returns clusters that were created by the user as well as any clusters returned by Sam as those that the user should be able to see. 

Testing: This was manually tested using the Terra UI. The problem presented itself in a variety of ways, including https://github.com/databiosphere/leonardo/issues/821 and https://github.com/DataBiosphere/leonardo/issues/844. To manually test this, I reproduced these situations and made sure that it had resolved. 

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
@@ -164,6 +164,12 @@ trait ClusterComponent extends LeoComponent {
       }
     }
 
+    def listClustersByCreator(creator: WorkbenchEmail): DBIO[Seq[Cluster]] = {
+      clusterQuery.filter { _.creator === creator.value }.result map { recs =>
+        recs.map(rec => unmarshalCluster(rec, Seq.empty, List.empty, Map.empty, List.empty, List.empty, List.empty))
+      }
+    }
+
     def countByClusterServiceAccountAndStatuses(clusterServiceAccount: WorkbenchEmail, statuses: Set[ClusterStatus]) = {
       clusterQuery
         .filter { _.clusterServiceAccount === Option(clusterServiceAccount.value) }
--- a/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
@@ -164,12 +164,6 @@ trait ClusterComponent extends LeoComponent {
       }
     }
 
-    def listClustersByCreator(creator: WorkbenchEmail): DBIO[Seq[Cluster]] = {
-      clusterQuery.filter { _.creator === creator.value }.result map { recs =>
-        recs.map(rec => unmarshalCluster(rec, Seq.empty, List.empty, Map.empty, List.empty, List.empty, List.empty))
-      }
-    }
-
     def countByClusterServiceAccountAndStatuses(clusterServiceAccount: WorkbenchEmail, statuses: Set[ClusterStatus]) = {
       clusterQuery
         .filter { _.clusterServiceAccount === Option(clusterServiceAccount.value) }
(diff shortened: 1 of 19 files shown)


src/test/scala/org/broadinstitute/dsde/workbench/leonardo/service/AuthProviderSpec.scala
```diff
@@ -119,10 +119,6 @@ class AuthProviderSpec extends FreeSpec with ScalatestRouteTest with Matchers wi
       clusterGetResponseException shouldBe a [ClusterNotFoundException]
       clusterGetResponseException.asInstanceOf[ClusterNotFoundException].statusCode shouldBe StatusCodes.NotFound
 
-      //list all clusters should be fine, but empty
```
akarukappadath: This section of the test will not pass, as all clusters created by a user will be returned in a list clusters call.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -527,8 +527,12 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
     for {
       paramMap <- processListClustersParameters(params)
       clusterList <- dbRef.inTransaction { da => da.clusterQuery.listByLabels(paramMap._1, paramMap._2, googleProjectOpt) }
-      visibleClusters <- authProvider.filterUserVisibleClusters(userInfo, clusterList.map(c => (c.googleProject, c.clusterName)).toList)
+      samVisibleClusters <- authProvider.filterUserVisibleClusters(userInfo, clusterList.map(c => (c.googleProject, c.clusterName)).toList)
+      // Making the assumption that users will always be able to access clusters that they create
```
MatthewBemis: i think this is a safe assumption to make. per [sam's config](https://github.com/broadinstitute/sam/blob/develop/src/main/resources/reference.conf#L286-L317), we don't even give people the permission to add/remove others from a cluster so once somebody creates a cluster they should never be removed.

at the very least this isn't driving access so even if the above weren't true, it wouldn't let anybody access data/clusters they shouldn't be able to see or didn't know about already.

MatthewBemis (approved):
:+1:

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -527,8 +527,12 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
...
+      // Fix for https://github.com/DataBiosphere/leonardo/issues/821
+      createdClusters <- dbRef.inTransaction { da => da.clusterQuery.listClustersByCreator(userInfo.userEmail).map(c => c.toList.map(cluster => (cluster.googleProject, cluster.clusterName))) }
```
rtitle: Do we need to make an additional query here? We already have the full list of clusters in the project in the `clusterList` variable. Could you just do:
```
val visibleClusters = samVisibleClusters ::: clusterList.filter(_.creator == userInfo.userEmail)
```

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
```diff
@@ -164,6 +164,12 @@ trait ClusterComponent extends LeoComponent {
       }
     }
 
+    def listClustersByCreator(creator: WorkbenchEmail): DBIO[Seq[Cluster]] = {
+      clusterQuery.filter { _.creator === creator.value }.result map { recs =>
```
rtitle: If adding this query we should also add an index on the `creator` column -- but as noted below I'm not sure it's needed.

rtitle (changes requested):
See 1 comment above about the additional DB query

akarukappadath:
jenkins retest

src/test/scala/org/broadinstitute/dsde/workbench/leonardo/service/AuthProviderSpec.scala
rtitle: Is there a way to actually test the new behavior?

rtitle:
The latest commit looks good -- I would like to see a new unit test that validates the new behavior (e.g. listClusters returns user-owned clusters even if not returned by the auth provider).

rtitle (approved):
💯

PR #865, opened 2019-04-12
Install GATK, CNN libraries, samtools, and OpenJDK in default dockerfile

akarukappadath:
A couple changes to the Dockerfile. 

GATK + CNN libraries - requested by Comms for workshops. Although the Docker image grows in size with this addition, cluster creation time is not affected. 

Samtools - requested by Brad for Terra

Java + JDK - we used to build and install these from Oracle. This meant that when Oracle updated their version, the leo build would break. This PR installs java and the jdk from debian, so it should be always pulling the most recent. Notable change: debian offers OpenJDK rather than Oracle's JDK. 

In all cases:


automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookInstallSpec.scala
```diff
@@ -50,8 +50,8 @@ class NotebookInstallSpec extends ClusterFixtureSpec with NotebookTestUtils {
       s"should be able to pip install packages using ${kernel.string}" in { clusterFixture =>
         withWebDriver { implicit driver =>
           withNewNotebook(clusterFixture.cluster, kernel) { notebookPage =>
-            // install tensorflow
```
akarukappadath: since we are now installing tensorflow by default, we have to install a package that isn't already present for this test to pass

akarukappadath:
jenkins retest

akarukappadath:
jenkins retest

akarukappadath:
jenkins retest

akarukappadath:
jenkins retest

docker/jupyter/Dockerfile
```diff
@@ -70,19 +74,17 @@ RUN echo "deb $DEBIAN_REPO/debian stretch main"                   > /etc/apt/sou
...
+ENV GATK_VERSION 4.1.0.0
+ENV GATK_ZIP_PATH /tmp/gatk-${GATK_VERSION}.zip
 
-RUN wget --header "Cookie: oraclelicense=accept-securebackup-cookie" $JAVA_URL \
- && mkdir -p /usr/lib/jdk && tar -zxf $JAVA_TGZ -C /usr/lib/jdk \
- && update-alternatives --install /usr/bin/java java $JAVA_HOME/bin/java 100 \
- && update-alternatives --install /usr/bin/javac javac $JAVA_HOME/bin/javac 100 \
- && rm $JAVA_TGZ
+ #download the gatk zip
+RUN curl -L -o $GATK_ZIP_PATH https://github.com/broadinstitute/gatk/releases/download/$GATK_VERSION/gatk-$GATK_VERSION.zip \
```
Qi77Qi: is the idea that we're installing gatk which comes with jdk instead of jdk directly?

akarukappadath: They're unrelated - we needed to add the GATK packages to our dockerfile, but also fix the jdk build problem that was causing the docker build to fail

Qi77Qi (approved):

PR #871, opened 2019-04-18
allow user to pass properties when creating cluster

Qi77Qi:
https://cloud.google.com/dataproc/docs/concepts/configuring-clusters/cluster-properties

In all cases:


project/Dependencies.scala
```diff
@@ -124,6 +125,8 @@ object Dependencies {
     workbenchGoogle,
     workbenchGoogleTest,
     workbenchMetrics,
-    sam
+    sam,
+
+    sealerate
```
Qi77Qi: just a tiny library for collecting all objects in a sealed trait. See https://github.com/mrvisser/sealerate. `enumeratum` might suffice my need, but I didn't want to extend ADT from `Enum[A]`. `sealerate` is more light weight.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
```diff
@@ -21,21 +21,23 @@ import org.broadinstitute.dsde.workbench.model.WorkbenchIdentityJsonSupport._
 import org.broadinstitute.dsde.workbench.model.google.GoogleModelJsonSupport._
 import org.broadinstitute.dsde.workbench.model.google._
 import org.broadinstitute.dsde.workbench.model._
-import spray.json._
+import spray.json.{RootJsonFormat, RootJsonReader, _}
+import ca.mrvisser.sealerate
 
 // Create cluster API request
-case class ClusterRequest(labels: Option[LabelMap] = Option(Map.empty),
```
Qi77Qi: Unless there's a reason to distinguish `None` and `Some(Map.emtpy)`, I think we should remove the `Option` wrapper here.

src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20190417_cluster_additional_dataproc_properties.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="leonardo" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+    <changeSet logicalFilePath="leonardo" author="qi" id="additional-dataproc-properties">
+        <addColumn tableName="CLUSTER">
+            <column name="properties" type="varchar(2048)">
```
MatthewBemis: so we're storing raw json in here? should we just make this `TEXT`?

src/main/resources/swagger/api-docs.yaml
```diff
@@ -1050,6 +1050,11 @@ definitions:
       machineConfig:
         description: The machine configurations for the master and worker nodes
         $ref: '#/definitions/MachineConfig'
+      properties:
+        description: Example `{"spark:spark.executor.memory": "10g"}`. See https://cloud.google.com/dataproc/docs/concepts/configuring-clusters/cluster-properties for allowed property settings
```
MatthewBemis: is the example properly escaped? yamllint didn't seem to like it. i think the colon in the link will be a problem too

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
MatthewBemis: there is, we don't want the field to be required in API requests. this would break backwards compatibility

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
```diff
@@ -21,21 +21,23 @@ import org.broadinstitute.dsde.workbench.model.WorkbenchIdentityJsonSupport._
...
+final case class ClusterRequest(labels: LabelMap,
                           jupyterExtensionUri: Option[GcsPath] = None,
                           jupyterUserScriptUri: Option[GcsPath] = None,
                           machineConfig: Option[MachineConfig] = None,
+                          properties: Map[String, String],
```
MatthewBemis: this should also be an `Option[Map[String,String]]` because existing UIs won't know to put in an empty map. in general we want to make any new fields optional if possible

MatthewBemis: spoke with Qi in person, this is handled in the json decoder now which is a better place for this logic to live. same goes for the above case

Qi77Qi:
@MatthewBemis  comments are addressed

MatthewBemis:
@Qi77Qi i don't think the Jenkins automation tests have run on this branch. Looks like it's because the base used to be `newrelic` instead of `develop`. with the most recent commit, the code looks good to me, but i'm holding my thumb until tests pass. you may have to close and re-open this PR to get them to run

Qi77Qi:
@MatthewBemis  There should've been jekins runs...https://fc-jenkins.dsp-techops.broadinstitute.org/job/leonardo-PR-fiab-test/2023/ is the latest one after I rebased on top of master

PR #882, opened 2019-04-23
Bugfix: disabling auto-pause does not work

rtitle:
I stumbled upon this when explaining to Adelaide how to use the PATCH endpoint to disable auto-pause. To disable autopause, we set a "special" value of 0 in the database. However we weren't actually honoring this value in the DB query -- so instead we were autopausing the cluster _immediately_. This hopefully fixes it.

In all cases:

### DIFF
--- a/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponentSpec.scala
+++ b/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponentSpec.scala
@@ -179,7 +179,7 @@ class ClusterComponentSpec extends TestComponent with FlatSpecLike with CommonTe
     val autopauseDisabledCluster = makeCluster(4).copy(
       auditInfo = auditInfo.copy(dateAccessed = Instant.now().minus(100, ChronoUnit.DAYS)),
       status = ClusterStatus.Running,
-      autopauseThreshold = 0)
+      autopauseThreshold = 0).save()
 
     val autoFreezeList = dbFutureValue { _.clusterQuery.getClustersReadyToAutoFreeze() }
     autoFreezeList should contain (runningCluster1)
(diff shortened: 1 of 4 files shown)


src/test/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponentSpec.scala
```diff
@@ -176,11 +176,17 @@ class ClusterComponentSpec extends TestComponent with FlatSpecLike with CommonTe
...
+      auditInfo = auditInfo.copy(dateAccessed = Instant.now().minus(100, ChronoUnit.DAYS)),
+      status = ClusterStatus.Running,
+      autopauseThreshold = 0).save()
+
     val autoFreezeList = dbFutureValue { _.clusterQuery.getClustersReadyToAutoFreeze() }
     autoFreezeList should contain (runningCluster1)
     //cluster2 is already stopped
     autoFreezeList should not contain stoppedCluster
     autoFreezeList should not contain runningCluster2
+    autoFreezeList should not contain autopauseDisabledCluster
```
rtitle: This check failed before the fix in `ClusterComponent`; now it passes.

rtitle:
jenkins retest

src/test/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponentSpec.scala
Qi77Qi: looks like the test only covered the case where `autopauseThreshold` is set to 0. Do we need to test the case where `autopause ` is `true`?

rtitle: The other clusters (`runningCluster1`, `runningCluster2`, `stoppedCluster`) all have `autopause` = `true` with a threshold of 30 mins. However it might be good to test other thresholds too. I'll add some more cases to the test.

Qi77Qi: ah I see...I'm fine with not adding more tests in that case :)

rtitle:
I also tested manually on a fiab

PR #895, opened 2019-04-29
lint cluster name before create cluster with dataproc

Qi77Qi:

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/LeoRoutes.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/LeoRoutes.scala
@@ -13,7 +13,7 @@ import akka.stream.scaladsl._
 import com.typesafe.scalalogging.LazyLogging
 import org.broadinstitute.dsde.workbench.leonardo.config.SwaggerConfig
 import org.broadinstitute.dsde.workbench.leonardo.errorReportSource
-import org.broadinstitute.dsde.workbench.leonardo.model.{ClusterRequest, LeoException}
+import org.broadinstitute.dsde.workbench.leonardo.model.{ClusterRequest, LeoException, RequestValidationError}
 import org.broadinstitute.dsde.workbench.leonardo.model.google.ClusterName
 import org.broadinstitute.dsde.workbench.leonardo.model.LeonardoJsonSupport._
 import org.broadinstitute.dsde.workbench.leonardo.service.{LeonardoService, ProxyService, StatusService}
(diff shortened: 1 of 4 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/LeoRoutes.scala
```diff
@@ -52,73 +53,78 @@ abstract class LeoRoutes(val leonardoService: LeonardoService, val proxyService:
...
-            pathEndOrSingleSlash {
-              put {
-                entity(as[ClusterRequest]) { cluster =>
-                  complete {
-                    leonardoService
-                      .processClusterCreationRequest(userInfo, GoogleProject(googleProject), ClusterName(clusterName), cluster)
-                      .map { cluster =>
-                        StatusCodes.Accepted -> cluster
+          pathPrefix("v2" / Segment / Segment) { (googleProject, clusterNameString) =>
+            validateClusterNameDirective(clusterNameString){
```
Qi77Qi: `validateClusterNameDirective` is the main addition, the rest is mostly indentation

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeoException.scala
```diff
@@ -12,3 +14,7 @@ abstract class LeoException(
     ErrorReport(Option(getMessage).getOrElse(""), Some(statusCode), Seq(), Seq(), Some(this.getClass))
   }
 }
+
+final case class RequestValidationError(message: String) extends NoStackTrace {
```
rtitle: Should this extend `LeoException`?

Qi77Qi: I didn't really want stack trace since it won't be very useful in this case (the message will be unique enough) and stack trace is the expensive part when creating exceptions.

Qi77Qi: oh I just realize there's no stack trace in leoException.

Qi77Qi:
jenkins retest

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeoException.scala
rtitle: Yeah, LeoException doesn't send a stack trace but it defines a status code and a message. Slight preference for standardizing on `LeoException` for API errors.

Qi77Qi: Talked to @rtitle offline. The idea is I want to distinct internal exceptions vs error reporting ones. I'll leave this PR as is and take another look at usage of exceptions later, and potentially remove `this.getClass` from errors we report to user.

PR #901, opened 2019-05-03
Jc creation zombie

jdcanas:

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ZombieClusterMonitor.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ZombieClusterMonitor.scala
@@ -1,6 +1,7 @@
 package org.broadinstitute.dsde.workbench.leonardo.monitor
 
 import java.time.Instant
+import java.time.Duration
 
 import akka.actor.{Actor, Props, Timers}
 import cats.implicits._
@@ -97,7 +98,16 @@ class ZombieClusterMonitor(config: ZombieClusterConfig, gdDAO: GoogleDataprocDAO
   private def isClusterActiveInGoogle(cluster: Cluster): Future[Boolean] = {
     // Clusters in Creating status may not yet exist in Google. Therefore treat all Creating clusters as active.
     if (cluster.status == ClusterStatus.Creating) {
-      Future.successful(true)
+
+        val minutesSinceImageCreation: Long = Duration.between(cluster.auditInfo.createdDate, Instant.now()).toMinutes()
+
+        if (minutesSinceImageCreation > config.creationHangTolerance.toMinutes) {
+           Future.successful(false)
+        } else {
+          logger.info("Minutes since image creation: " ++ minutesSinceImageCreation.toString)
+          Future.successful(true)
+        }
+
     } else {
       // Check if status returned by GoogleDataprocDAO is an "active" status.
       gdDAO.getClusterStatus(cluster.googleProject, cluster.clusterName) map { clusterStatus =>
--- a/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ZombieClusterMonitor.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ZombieClusterMonitor.scala
@@ -96,15 +96,15 @@ class ZombieClusterMonitor(config: ZombieClusterConfig, gdDAO: GoogleDataprocDAO
   }
 
   private def isClusterActiveInGoogle(cluster: Cluster): Future[Boolean] = {
-    // Clusters in Creating status may not yet exist in Google. Therefore treat all Creating clusters as active.
+    // Clusters in Creating status may not yet exist in Google. Clusters in creating (that are not hanging, as defined by config.creationHangTolerance) are considered active
     if (cluster.status == ClusterStatus.Creating) {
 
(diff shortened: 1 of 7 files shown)


jdcanas:
jenkins retest

jdcanas:
jenkins retest

src/test/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ZombieClusterMonitorSpec.scala
```diff
@@ -131,7 +131,7 @@ class ZombieClusterMonitorSpec extends TestKit(ActorSystem("leonardotest")) with
     }
   }
 
-  it should "should not zombify Creating clusters" in isolatedDbTest {
+  it should "should zombify creating cluster after hang period" in isolatedDbTest {
```
Qi77Qi: Do we need to keep the original test that will not zombify creating cluster if it's within hang period?

jdcanas:
jenkins retest

src/test/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ZombieClusterMonitorSpec.scala
jdcanas: I was thinking the same thing on the first pass. Just added it and cleaned up the hardcoded constants

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ZombieClusterMonitor.scala
```diff
@@ -95,9 +96,18 @@ class ZombieClusterMonitor(config: ZombieClusterConfig, gdDAO: GoogleDataprocDAO
...
-    // Clusters in Creating status may not yet exist in Google. Therefore treat all Creating clusters as active.
+    // Clusters in Creating status may not yet exist in Google. Clusters in creating (that are not hanging, as defined by config.creationHangTolerance) are considered active
     if (cluster.status == ClusterStatus.Creating) {
-      Future.successful(true)
+
+        val secondsSinceClusterCreation: Long = Duration.between(cluster.auditInfo.createdDate, Instant.now()).getSeconds
+
+        //determine if the creating cluster has been hanging past the tolerance threshold
+        if (secondsSinceClusterCreation > config.creationHangTolerance.toSeconds) {
+           Future.successful(false)
```
rtitle: Should this `if` block check the cluster status in Google? I'm thinking we want the following logic:

| status in Leo | present in Google | isClusterActiveInGoogle returns... |
| ------------- | ------------------| ---------------------------------|
| Creating, inside the hang tolerance | yes | true |
| Creating, inside the hang tolerance | no | true |
| Creating, outside the hang tolerance | yes | true |
| Creating, outside the hang tolerance | no | false |

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ZombieClusterMonitor.scala
```diff
@@ -95,17 +96,16 @@ class ZombieClusterMonitor(config: ZombieClusterConfig, gdDAO: GoogleDataprocDAO
...
-        logger.warn(s"Unable to check status of cluster ${cluster.projectNameString} for zombie cluster detection", e)
-        true
-      }
+
+    val secondsSinceClusterCreation: Long = Duration.between(cluster.auditInfo.createdDate, Instant.now()).getSeconds
+    //this or'd with the google cluster status gives creating clusters a grace period before they are marked as zombies
+    val isWithinHangTolerance =  cluster.status == ClusterStatus.Creating && secondsSinceClusterCreation > config.creationHangTolerance.toSeconds
+
+   gdDAO.getClusterStatus(cluster.googleProject, cluster.clusterName) map { clusterStatus =>
+      (ClusterStatus.activeStatuses contains clusterStatus) || isWithinHangTolerance
```
Qi77Qi: I could be misreading, but I think this is not quite right. `ClusterStatus.activeStatuses` includes `Creating`, so we won't actually be checking `isWithinHangTolerance` when status is `Creating`.

jdcanas: As I understand it based, if the cluster is active in google, it is not considered a zombie in any scenario, as the point of the zombie monitor is to remove clusters in Leo DB that aren't present in google, and the functionality is not designed to extend the other way.

Thus, the hang tolerance is only relevant in the cases in which the cluster is not active in google. That being said, I think I should split line 102 into two separate variables so it is more clear though (after I merge this maybe, so I don't have to go through the automation tests again). Something like:

`val shouldConsiderCreatingClusterActive = isWithinHangTolerance && cluster.status == ClusterStatus.Creating`

Which should maintain:

| Creating  | In google | Within hang tolerance | Desired Result |
|---|---|---|---|
| Yes | Yes | Yes | true |
| Yes | No | Yes | true |
| Yes | Yes | No | true |
| Yes | No  | No | false |
| No | Yes | Any | true |
| No | No | Any | false |

Qi77Qi: thanks for the explanation. I think that makes sense. one nit, the method name is a bit misleading after we add the hangTolerance logic. `isClusterActive` would be better. A comment like what you said here would be helpful too.

PR #916, opened 2019-05-07
Allow users to see output from a failed userscript

akarukappadath:
Leo changes in order to support Terra UI showing the output of a failed userscript. Pipes the output of the user script to a different file and copies it into the staging bucket so that users can more easily debug their userscript. 

In all cases:


src/main/resources/jupyter/jupyter_notebook_config.py
```diff
@@ -32,7 +32,5 @@
     'static_url_prefix':'/notebooks/' + fragment + '/static/',
     'headers': {
         'Content-Security-Policy': $(contentSecurityPolicy)
-    },
-    'websocket_ping_interval':30,
-    'websocket_ping_timeout':1800
+    }
```
akarukappadath: Removing this to clean up from terminal timeout experiments

src/main/resources/jupyter/init-actions.sh
```diff
@@ -302,7 +303,9 @@ if [[ "${ROLE}" == 'Master' ]]; then
         JUPYTER_USER_SCRIPT=`basename ${JUPYTER_USER_SCRIPT_URI}`
         docker cp /etc/${JUPYTER_USER_SCRIPT} ${JUPYTER_SERVER_NAME}:${JUPYTER_HOME}/${JUPYTER_USER_SCRIPT}
         retry 3 docker exec -u root ${JUPYTER_SERVER_NAME} chmod +x ${JUPYTER_HOME}/${JUPYTER_USER_SCRIPT}
-        retry 3 docker exec -u root -e PIP_USER=false ${JUPYTER_SERVER_NAME} ${JUPYTER_HOME}/${JUPYTER_USER_SCRIPT}
+        docker exec -u root -e PIP_USER=false ${JUPYTER_SERVER_NAME} ${JUPYTER_HOME}/${JUPYTER_USER_SCRIPT} &> us_output.txt || EXIT_CODE=$? && true ;
```
Qi77Qi: so we're no longer retrying?

akarukappadath: Yes. We previously added retries to all docker exec statements because some automation tests were failing transiently. Rob and I talked about it for a bit and it seems like it would make most sense to keep the retries everywhere except here because in the case of failing userscript that takes a lot of time to run, the user would have to wait three times the length of the runtime to get an error message

rtitle: Does `&>` make it run in the background? I think the init script should block on the execution of the user script.

src/main/resources/jupyter/init-actions.sh
```diff
@@ -302,7 +303,9 @@ if [[ "${ROLE}" == 'Master' ]]; then
...
+        gsutil cp us_output.txt ${JUPYTER_USER_SCRIPT_OUTPUT_URI}
+        $EXIT_CODE
```
rtitle: Not sure if this does anything..

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
```diff
@@ -269,6 +270,7 @@ object ClusterInitValues {
       proxyConfig.proxyServerName,
       clusterRequest.jupyterExtensionUri.map(_.toUri).getOrElse(""),
       clusterRequest.jupyterUserScriptUri.map(_.toUri).getOrElse(""),
+      GcsPath(stagingBucket, GcsObjectName("userscript_output.txt")).toUri,
```
rtitle: What if we named the output file based on the user script name? So if my script was named `foo.sh` the output could be `foo_output.log`?

src/main/resources/jupyter/jupyter_notebook_config.py
rtitle: Do these settings actually affect anything user facing?

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
rtitle: Then again maybe a consistent name is good so the UI is able to easily find it...

src/test/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoServiceSpec.scala
```diff
@@ -554,7 +554,7 @@ class LeonardoServiceSpec extends TestKit(ActorSystem("leonardotest")) with Flat
 
   it should "template a script using config values" in isolatedDbTest {
     // Create replacements map
-    val clusterInit = ClusterInitValues(project, name1, initBucketPath, testClusterRequestWithExtensionAndScript, dataprocConfig, clusterFilesConfig, clusterResourcesConfig, proxyConfig, Some(serviceAccountKey), userInfo.userEmail, contentSecurityPolicy, Set(jupyterImage, rstudioImage))
+    val clusterInit = ClusterInitValues(project, name1, initBucketPath, testClusterRequestWithExtensionAndScript, dataprocConfig, clusterFilesConfig, clusterResourcesConfig, proxyConfig, Some(serviceAccountKey), userInfo.userEmail, contentSecurityPolicy, Set(jupyterImage, rstudioImage), stagingBucketName)
```
rtitle: You could update this test so it asserts that `$(jupyterUserScriptOutputUri)` is substituted correctly.
1. Update https://github.com/DataBiosphere/leonardo/blob/develop/src/test/resources/jupyter/test-init-actions.sh with the variable
2. Update the `expected` val to have the expected replacement

src/main/resources/jupyter/init-actions.sh
rtitle: Oh nevermind -- `&>` means `>us_output 2>&1`. This seems reasonable then I think..

src/main/resources/jupyter/init-actions.sh
```diff
@@ -302,7 +303,9 @@ if [[ "${ROLE}" == 'Master' ]]; then
         JUPYTER_USER_SCRIPT=`basename ${JUPYTER_USER_SCRIPT_URI}`
         docker cp /etc/${JUPYTER_USER_SCRIPT} ${JUPYTER_SERVER_NAME}:${JUPYTER_HOME}/${JUPYTER_USER_SCRIPT}
         retry 3 docker exec -u root ${JUPYTER_SERVER_NAME} chmod +x ${JUPYTER_HOME}/${JUPYTER_USER_SCRIPT}
-        retry 3 docker exec -u root -e PIP_USER=false ${JUPYTER_SERVER_NAME} ${JUPYTER_HOME}/${JUPYTER_USER_SCRIPT}
+        docker exec -u root -e PIP_USER=false ${JUPYTER_SERVER_NAME} ${JUPYTER_HOME}/${JUPYTER_USER_SCRIPT} &> us_output.txt || EXIT_CODE=$? && true ;
+        gsutil cp us_output.txt ${JUPYTER_USER_SCRIPT_OUTPUT_URI}
```
rtitle: Maybe the `gsutil cp` should have a retry?

src/main/resources/jupyter/jupyter_notebook_config.py
akarukappadath: Nope, they didn't change anything

src/main/resources/jupyter/init-actions.sh
akarukappadath: I tested this by running Terra locally while pointing to a fiab of this branch. For a cluster with a failing userscript, the init script stops at this line, whereas if the script is successful it passes this line. I think because we're using `set -e` at the beginning, this will work, but I can try to see if `return $EXIT_CODE` also works if that's more readable

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
akarukappadath: Will we be allowing users to specify multiple user scripts (for example, when we introduce more tools?). I thought keeping it consistent would make it easier to follow, but if we will eventually support multiple userscripts it might be a good idea to make the name specific

src/main/resources/jupyter/init-actions.sh
akarukappadath: I just tried using return and it looks like it doesn't work, going to stick with the original!

rtitle: Yeah, I think this works because of `set -e` but can I suggest a change for readability?
```
if [ $EXIT_CODE -ne 0 ]; then
    echo "User script failed with exit code $EXIT_CODE. Output is saved to $JUPYTER_USER_SCRIPT_OUTPUT_URI."
    exit $EXIT_CODE
fi
```
This way we're not relying on `set -e` semantics and it's more clear IMO.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
rtitle: Good point. For now we just support 1 user script so maybe we cross that bridge when we come to it.

rtitle:
Looks good, just 1 more bash comment

src/main/resources/jupyter/init-actions.sh
akarukappadath: Sounds good!

akarukappadath:
jenkins retest

akarukappadath:
jenkins retest

akarukappadath:
jenkins retest

akarukappadath:
jenkins retest

akarukappadath:
jenkins retest

akarukappadath:
jenkins retest

akarukappadath:
jenkins retest

akarukappadath:
jenkins retest

akarukappadath:
jenkins retest

akarukappadath:
jenkins retest

src/test/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorSpec.scala
```diff
@@ -37,7 +40,7 @@ import scala.util.{Random, Try}
 /**
   * Created by rtitle on 9/6/17.
   */
-class ClusterMonitorSpec extends TestKit(ActorSystem("leonardotest")) with FlatSpecLike with Matchers with MockitoSugar with BeforeAndAfterAll with TestComponent with CommonTestData with GcsPathUtils with Eventually { testKit =>
+ class ClusterMonitorSpec extends TestKit(ActorSystem("leonardotest")) with FlatSpecLike with Matchers with MockitoSugar with BeforeAndAfterAll with TestComponent with CommonTestData with GcsPathUtils with Eventually { testKit =>
```
rtitle: Extra space :)

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
```diff
@@ -176,30 +179,41 @@ class ClusterMonitorActor(val cluster: Cluster,
...
-            logger.warn(s"Could not find Id for Cluster ${cluster.projectNameString}  with google cluster ID ${cluster.dataprocInfo.googleId}.")
-            DBIOAction.successful(0)
-          }
+    for {
+      _ <- cluster.dataprocInfo.stagingBucket match {
+        case Some(stagingBucketName) => {
+          for {
+            metadata <- google2StorageDAO.getObjectMetadata(stagingBucketName, GcsBlobName("userscript_output.txt"), None).compile.last.unsafeToFuture()
+            userscriptFailed = metadata match {
+              case Some(GetMetadataResponse.Metadata(_, metadataMap)) => metadataMap.exists(_ == "passed"->"false")
```
rtitle: Nit: `metadataMap.contains("passed" -> "false")` ?

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
```diff
@@ -176,30 +179,41 @@ class ClusterMonitorActor(val cluster: Cluster,
...
-      dbRef.inTransaction { dataAccess =>
-        val clusterId = dataAccess.clusterQuery.getIdByUniqueKey(cluster)
-        clusterId flatMap {
-          case Some(a) => dataAccess.clusterErrorQuery.save(a, ClusterError(errorDetails.message.getOrElse("Error not available"), errorDetails.code, Instant.now))
-          case None => {
-            logger.warn(s"Could not find Id for Cluster ${cluster.projectNameString}  with google cluster ID ${cluster.dataprocInfo.googleId}.")
-            DBIOAction.successful(0)
-          }
+    for {
+      _ <- cluster.dataprocInfo.stagingBucket match {
```
rtitle: I was wondering if this block could be factored into its own method and included in the `Future.sequence` below.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
akarukappadath: `contains` only tells you if a key exists in a map (rather than a key-value pair) so we wouldn't be able to use it here

akarukappadath:
jenkins retest

project/Dependencies.scala
```diff
@@ -83,6 +86,7 @@ object Dependencies {
   val mysql: ModuleID =     "mysql"               % "mysql-connector-java"  % "8.0.11"
   val liquibase: ModuleID = "org.liquibase"       % "liquibase-core"        % "3.5.3"
   val sealerate: ModuleID = "ca.mrvisser" %% "sealerate" % "0.0.5"
+  val googleCloudNio: ModuleID = "com.google.cloud" % "google-cloud-nio" % "0.71.0-alpha" % "test"
```
Qi77Qi: just want to point out this is only needed for google storage emulator in test

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
```diff
@@ -176,30 +179,21 @@ class ClusterMonitorActor(val cluster: Cluster,
...
-          case Some(a) => dataAccess.clusterErrorQuery.save(a, ClusterError(errorDetails.message.getOrElse("Error not available"), errorDetails.code, Instant.now))
-          case None => {
-            logger.warn(s"Could not find Id for Cluster ${cluster.projectNameString}  with google cluster ID ${cluster.dataprocInfo.googleId}.")
-            DBIOAction.successful(0)
-          }
-        }
-      }
-    ))
+    for {
+      _ <- Future.sequence(Seq(
```
Qi77Qi: minor here, we should almost never use `Seq`, it's mutable. `List` will be better

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
```diff
@@ -354,6 +348,44 @@ class ClusterMonitorActor(val cluster: Cluster,
...
+      val clusterId = dataAccess.clusterQuery.getIdByUniqueKey(cluster)
+      Future(logger.info(s"${cluster.clusterName}: clusterId in else is ${clusterId}"))
+      clusterId flatMap {
+        case Some(a) => dataAccess.clusterErrorQuery.save(a, ClusterError(errorMessage, errorCode, Instant.now))
+        case None => {
+          logger.warn(s"Could not find Id for Cluster ${cluster.projectNameString}  with google cluster ID ${cluster.dataprocInfo.googleId}.")
+          DBIOAction.successful(0)
+        }
+      }
+    }.void.handleError(e => logger.info(s"Error persisting cluster error with message '${errorMessage}' to database: ${e}"))
```
Qi77Qi: I'll use `handleError` on `persistClusterErrors` because you don't want to fail everything else because of failing in `persistClusterErrors`. With current implementation, if `google2StorageDAO.getObjectMetadata` fails, everything will fail, which isn't what you want

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
rtitle: Good point! I stand corrected

PR #935, opened 2019-05-09
Add support for running Docker inside rootless Notebook container host

dinvlad:
## Docker in Rootless Notebooks

Jupyter Notebooks can optionally run in a [rootless Docker container](https://engineering.docker.com/2019/02/experimenting-with-rootless-docker/), to provide additional protection for the host.

However, we also use *rootless mode* to provide a safer way for users
to build/run their own Docker containers *inside* notebooks.

This mode works by providing a custom binary for the Docker daemon and associated executables,
and should be considered an experimental feature.

To set it up, first build and push a custom Docker image,
```
docker build -t us.gcr.io/broad-dsp-gcr-public/leonardo-jupyter:rootless -f Dockerfile.rootless .
docker push us.gcr.io/broad-dsp-gcr-public/leonardo-jupyter:rootless
```

Next, use an **Ubuntu 18.04+** VM for the master node
(other distributions place severe limitations on filesystem performance).
Run the following as a *non-root* user on the VM:
```
DOCKER_IMG=us.gcr.io/broad-dsp-gcr-public/leonardo-jupyter:rootless \
  ROOTLESS=1 ./docker/jupyter/run-jupyter-local.sh start
```

This command downloads and starts experimental Docker daemon, as the user who calls it.
Then, the Docker executables are mounted into the notebook ("host") container to provide a way
for Jupyter users to run Docker commands from *inside* the host container.

Note that the host container starts a *separate* "rootful" Docker daemon *inside* of itself,
so that any "child" containers started by the user don't affect the host.

In summary, the flow of control is as follows:

Rootless Docker daemon → Notebook container → Rootful Docker daemon → Child/User containers

As a result, notebook users can just run *regular* docker commands, such as
`docker build` and `docker run`, with all of the usual semantics involved:

Screen Shot 2019-05-08 at 10 31 36 PM

Remaining challenges [1]:
- ~enable **slirp4netns** to improve network performance (limited testing showed that it works)~
- ~more importantly, enable `overlay`/`overlay2` storage driver inside child Docker (currently, it uses `vfs` for some reason, even with `overlay2` on the host); this is to dramatically improve filesystem performance~ [solved]

[1] https://www.slideshare.net/AkihiroSuda/dockercon-2019-hardening-docker-daemon-with-rootless-mode


dinvlad:
Resolved performance problems with network and storage:

1) The updated `run-jupyter-local.sh` script automatically installs `slirp4netns` binary to enable fast userspace networking. It doesn't come close to native speeds, but still enables comfortable ~2 Gbps throughput with Google buckets. It would be possible to achieve native speeds, but one would have to use `lxc-user-nic` driver, which is less secure: https://github.com/rootless-containers/rootlesskit/tree/v0.4.0#--netlxc-user-nic-experimental

2) Rootless container now creates and mounts a named volume `docker` from rootless Docker, using `-v docker:/var/lib/docker`. The child Docker then picks up fast `overlay2` driver from the host. This mounting trick also has a nice side-effect of preserving child container storage between reboots (so the user doesn't have to re-pull images later, for example)

jdcanas:
Closing, work captured here https://broadworkbench.atlassian.net/browse/IA-1276

PR #939, opened 2019-05-14
Jc error message change

jdcanas:
The original error message is fairly abrasive to the user; it can be received if they find themselves editting a notebook that is stopping/starting. 

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/ProxyService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/ProxyService.scala
@@ -31,7 +31,7 @@ import scala.concurrent.duration._
 import scala.concurrent.{ExecutionContext, Future}
 
 case class ClusterNotReadyException(googleProject: GoogleProject, clusterName: ClusterName)
-  extends LeoException(s"Cluster ${googleProject.value}/${clusterName.value} is not ready yet. It may be updating, try again later", StatusCodes.EnhanceYourCalm)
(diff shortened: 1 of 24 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/ProxyService.scala
```diff
@@ -31,7 +31,7 @@ import scala.concurrent.duration._
 import scala.concurrent.{ExecutionContext, Future}
 
 case class ClusterNotReadyException(googleProject: GoogleProject, clusterName: ClusterName)
-  extends LeoException(s"Cluster ${googleProject.value}/${clusterName.value} is not ready yet, chill out and try again later", StatusCodes.EnhanceYourCalm)
+  extends LeoException(s"Cluster ${googleProject.value}/${clusterName.value} is not ready yet. It may be updating, try again later", StatusCodes.UnprocessableEntity)
```
Qi77Qi: I don't think cluster not being ready is a `UnprocessableEntity`. My understanding of `UnprocessableEntity` is your request body is mal formatted

Qi77Qi: If we're changing error code, we should also check if we need to update swagger file

rtitle: Good point, +1

PR #944, opened 2019-05-20
[IA-958][risk=no]Add Safe Mode extension to repo

jmthibault79:
Have you read [CONTRIBUTING.md](https://github.com/DataBiosphere/leonardo/blob/develop/CONTRIBUTING.md) lately? If not, do that first.

I, the developer opening this PR, do solemnly pinky swear that:

- [ ] I've documented my API changes in Swagger

In all cases:

- [ ] Get a thumbsworth of review and PO signoff if necessary
- [ ] Verify all tests go green
- [ ] Squash and merge; you can delete your branch after this
- [ ] Test this change deployed correctly and works on dev environment after deployment

PR #945, opened 2019-05-21
[IA-959][risk=no]Hide more functionality in Safe Mode extension

jmthibault79:
Includes #944 for IA-958.

In all cases:

### DIFF
--- a/src/main/resources/jupyter/safe-mode.js
+++ b/src/main/resources/jupyter/safe-mode.js
@@ -1,0 +1,53 @@
+// Adapted from the All of Us Researcher Workbench "Playground Mode"
+// https://github.com/all-of-us/workbench/blob/master/api/cluster-resources/playground-extension.js
+
+// In "Safe Mode", changes are not saved back to GCS. This extension makes
+// minor UI tweaks to differentiate this mode from normal Jupyter usage, and
+// also removes/hides controls relating to persistence. Technically
+// this does not stop autosave from continuing to happen in the background, but
+// the intended use of this plugin is in a separate space from normal operation
+// which does not support localization features.
+
+define([
+    'base/js/namespace'
+], (Jupyter) => {
+  const load = () => {
+    // TODO always-on, for now
+    const enabled = true;
+    if (!enabled) {
+      return;
+    }
+
+    // Disable UI controls/notifications relating to saving.
+
+    // "notbook" is an intentional typo to match Jupyter UI HTML.
+    $('#save-notbook').remove();
+    $('#save_notebook_as').remove();
+    $('#save_checkpoint').remove();
+
+    // TODO: not yet working as intended
+    $('#menubar-container')
+        .append(
+            '<style>' +
+              '.autosave_status { display: none; }' +
+              '#safe-mode { background-color: #FFFFB2; }' +
+              '</style>');
+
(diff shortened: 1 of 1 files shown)


src/main/resources/jupyter/safe-mode.js
```diff
@@ -0,0 +1,105 @@
+// Adapted from the All of Us Researcher Workbench "Playground Mode"
+// https://github.com/all-of-us/workbench/blob/master/api/cluster-resources/playground-extension.js
+
+// TODO: the following is required until IA-979 is implemented
+// To deploy on a Leonardo cluster:
+// 1. copy to a public GCS location
+// 2. issue a PUT request to <leonardo>/api/cluster/<billing project>/<cluster name>
```
calbach: Interesting - does this get picked up immediately? Or does the cluster need to restart?

src/main/resources/jupyter/safe-mode.js
```diff
@@ -0,0 +1,105 @@
...
+    $('#restore_checkpoint').remove();
+
+    // A little cleanup: remove the first two dividers in the file menu
+    // which are no longer dividing anything
+    $('#file_menu')
+        .append(
+            '<style>' +
+              // TODO this does not remove the first divider.
+              // instead it removes the whole menu.  Why?
+              //'li:nth-of-type(1) { display: none; } ' +
```
jmthibault79: I am still a long way from understanding CSS

src/main/resources/jupyter/safe-mode.js
```diff
@@ -0,0 +1,105 @@
...
+    // TODO: convert to tooltip with this text:
+    // Safe Mode allows you to explore, change, and run the code,
+    // but your edits will not be saved.
+    // To save your work, choose Make a Copy from the File menu to
+    // make your own version.
+
+    $('#notification_area').prepend(
+        '<div id="safe-mode" style="background-color: #FFFFB2;" ' +
+          'class="notification_widget btn btn-xs navbar-btn">' +
+          '<span>Safe Mode - your edits will not be saved.' +
```
jmthibault79: Happy to change this to whatever - but note that this is only the temporary UI.

src/main/resources/jupyter/safe-mode.js
calbach: It might be easier to use a jquery selector to capture this.. I'm launching a cluster now to play around with it. The other style element approach I used because I believe those notification divs aren't rendered until they are active.

src/main/resources/jupyter/safe-mode.js
jmthibault79: This is the cluster create endpoint.  Any new notebooks on the cluster have the extension enabled.

src/main/resources/jupyter/safe-mode.js
calbach: Try this out:

```
$("#file_menu .divider")[2].remove()
```

This is a jQuery selector if you want to look it up. jQuery is definitely out of vogue so don't worry about learning too much about it - but you're likely to encounter it on older projects like this.

calbach: (but the internal piece is a valid CSS selector in this case)

jmthibault79: Thanks, that works.

PR #948, opened 2019-05-28
[IA-951][IA-952][risk=low]

jmthibault79:
Steps 2 and 3 of IA-952: Deploy Welder to all clusters, and enable file sharing between Welder and Jupyter.

In all cases:


src/main/resources/jupyter/init-actions.sh
```diff
@@ -145,6 +147,11 @@ if [[ "${ROLE}" == 'Master' ]]; then
     mkdir /certs
     chmod a+wx /work
 
+    # shared jupyter/welder dir
```
jmthibault79: Q: create a new dir like I'm doing here, or reuse an existing Jupyter dir?

src/test/resources/jupyter/test-welder-docker-compose.yaml
```diff
@@ -0,0 +1 @@
+welder
```
jmthibault79: other test docker compose files are similarly useless

src/main/resources/jupyter/welder-docker-compose.yaml
```diff
@@ -0,0 +1,19 @@
+version: '2'
+services:
+  welder:
+    container_name: "${WELDER_SERVER_NAME}"
+    image: "${WELDER_DOCKER_IMAGE}"
+    network_mode: host
+    restart: always
+    environment:
+      # needed to disable auth
```
calbach: Could you expand on this? What auth is being disabled and why?

jmthibault79: Not clear to me either - it was in the original branch I took this from.

Actually this whole section should be updated with the values Welder needs.

src/main/resources/jupyter/init-actions.sh
calbach: Per previous discussions with @Qi77Qi , the plan-of-record is to mount the shared volume directly onto the Jupyter working directory. I believe this is currently `/home/user/work` in `jupyter-docker-compose.yaml`, which appears to currently occupied by `/work`.

Perhaps just continue using the existing /work directory, but start managing it with Welder?

jmthibault79: That makes sense to me.  I'll update it to do that.

src/main/resources/jupyter/init-actions.sh
```diff
@@ -153,11 +156,15 @@ if [[ "${ROLE}" == 'Master' ]]; then
...
     gsutil cp ${RSTUDIO_DOCKER_COMPOSE} /etc
     gsutil cp ${PROXY_DOCKER_COMPOSE} /etc
+    gsutil cp ${WELDER_DOCKER_COMPOSE} /etc
 
     # Needed because docker-compose can't handle symlinks
     touch /hadoop_gcs_connector_metadata_cache
     touch auth_openidc.conf
 
+    # do not enable welder yet.  Remove flag when welder is complete.
+    WELDER_ENABLED=false
```
jmthibault79: DO NOT enable Welder's management of the Jupyter /work dir yet.  Merging this should not affect Jupyter functionality.

jmthibault79:
Confirmed that this also satisfies "step 4" of the story if you manually supply a working Welder docker image and run it via docker-compose: the Jupyter container can access Welder at http://localhost:8080/ and the Welder container can access Jupyter at http://localhost:8000/

jmthibault79:
Proxy updates enable "step 5" : access from outside.

Example URL: https://firecloud-fiab.dsde-dev.broadinstitute.org:30443/proxy/joel-billing-20190529-1/joel-welder-5/welder/status

src/main/resources/jupyter/cluster-site.conf
```diff
@@ -24,6 +24,19 @@
     ProxyPass / http://127.0.0.1:8001/ retry=10
     ProxyPassReverse / http://127.0.0.1:8001/
 
+    # Welder
```
jmthibault79: This is copy/paste and I don't understand it well.  I'd appreciate a review from Leo engineers who have worked on this previously.

src/main/resources/jupyter/cluster-site.conf
```diff
@@ -24,6 +24,19 @@
...
+
+    RewriteCond %{HTTP:Upgrade} =websocket
```
jmthibault79: I'm guessing this block can be removed.  Welder doesn't need to handle websockets.

src/main/resources/jupyter/cluster-site.conf
```diff
@@ -24,6 +24,19 @@
...
+    RewriteCond %{REQUEST_URI} /proxy/[^/]*/[^/]*/welder/.* [NC]
+    RewriteRule /proxy/[^/]*/[^/]*/welder/(.*) ws://127.0.0.1:8080/$1  [P,L]
+
+    RewriteCond %{HTTP:Upgrade} !=websocket
+    RewriteCond %{REQUEST_URI} /proxy/[^/]*/[^/]*/welder/.* [NC]
+    RewriteRule /proxy/[^/]*/[^/]*/welder/(.*) http://127.0.0.1:8080/$1 [P,L]
+
+    ProxyPass / http://127.0.0.1:8080/ retry=10
```
jmthibault79: It's not clear to me what this block does.

src/main/resources/jupyter/init-actions.sh
jmthibault79: If you want to enable it, ssh to the master node and:
```
sudo -s
gcloud docker --authorize-only
export WELDER_DOCKER_IMAGE=us.gcr.io/broad-dsp-gcr-public/welder-server:5c3d6bb
docker-compose -f /etc/welder-docker-compose.yaml up -d
```

src/main/resources/jupyter/cluster-site.conf
MatthewBemis: yeah it can be removed

src/main/resources/jupyter/cluster-site.conf
MatthewBemis: it looks correct to me. i've commented inline to provide clarity of what it all means

```
    RewriteCond %{HTTP:Upgrade} !=websocket     //if the request is NOT a websocket request
    RewriteCond %{REQUEST_URI} /proxy/[^/]*/[^/]*/welder/.* [NC]     //if the request URI matches the regex, ignoring case sensitivity (NC = not case sensitive)
    RewriteRule /proxy/[^/]*/[^/]*/welder/(.*) http://127.0.0.1:8080/$1 [P,L] //if all of the above conds are true, rewrite the request to "http://127.0.0.1:8080/$1", where $1 is the first capture group and so on. P = proxy the request, L = "last", which means to stop processing any following rulesets
```

jmthibault79:
Rebased to pick up necessary conf changes for successful auto-tests / FIAB runs

src/main/resources/jupyter/cluster-site.conf
jmthibault79: removed.

src/main/resources/jupyter/init-actions.sh
```diff
@@ -171,13 +178,17 @@ if [[ "${ROLE}" == 'Master' ]]; then
         log 'Copying SA into RStudio Docker...'
         docker cp /etc/${SERVICE_ACCOUNT_CREDENTIALS} ${RSTUDIO_SERVER_NAME}:/etc/${SERVICE_ACCOUNT_CREDENTIALS}
       fi
+      if [ ! -z ${WELDER_DOCKER_IMAGE} ] && [ "${WELDER_ENABLED}" == "true" ] ; then
```
Qi77Qi: does this mean we're copying service account file to welder container?

jmthibault79: Only if when we populate `$SERVICE_ACCOUNT_CREDENTIALS` - which is not something we normally do.  Maybe some comments would help clarify here.

Qi77Qi: hey @rtitle , can you comment on whether we need this code if welder is just using application default credential?

jmthibault79: I think yes.

In the normal case (we do not set SERVICE_ACCOUNT_CREDENTIALS) then this does nothing.

In the weird case (some kind of pluggable auth where we do in fact set SERVICE_ACCOUNT_CREDENTIALS) then this is the code which enables Google Application Default Credentials to be equal to SERVICE_ACCOUNT_CREDENTIALS, which is necessary for anything using the defaults.

Since Welder will be using ADC (like Jupyter does) then it will have the correct credentials in both cases

src/main/resources/jupyter/init-actions.sh
MatthewBemis: do you also need to `export WELDER_ENABLED=true`?

jmthibault79: no. it is only used in this script.

MatthewBemis: ah!

PR #954, opened 2019-06-05
Run the Jupyter user script as privileged

calbach:
https://broadworkbench.atlassian.net/browse/IA-1013

In all cases:


calbach:
Thanks, I don't think I can change the branch name so created https://github.com/DataBiosphere/leonardo/pull/955 ; closing this

PR #959, opened 2019-06-07
[risk=no][IA-1026]Defer the execution of Welder until after its container is created

jmthibault79:
It’s not clear why, but running Welder as an Entrypoint in its docker container does not allow it to find the [Application Default Credentials](https://cloud.google.com/docs/authentication/production) available on the host.

Welder does find the ADC if it is run inside the already-running container, so we defer execution in this way.  Interestingly, Jupyter already does this, for unrelated reasons.  Perhaps that is what enables Jupyter to use ADC as well.

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
@@ -129,7 +129,7 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
     // so we need to fall back running `jupyter notebook` directly. See https://github.com/DataBiosphere/leonardo/issues/481.
     val jupyterStart = s"docker exec -d ${dataprocConfig.jupyterServerName} /bin/bash -c '/etc/jupyter/scripts/run-jupyter.sh || /usr/local/bin/jupyter notebook'"
 
-    // TODO make this flag configurable. https://broadworkbench.atlassian.net/browse/IA-1037
+    // TODO make this flag configurable. https://broadworkbench.atlassian.net/browse/IA-1033
     val enableWelder = false
     val servicesStart = if (enableWelder) {
       val welderStart = s"docker exec -u daemon -d ${dataprocConfig.welderServerName} /opt/docker/bin/server start"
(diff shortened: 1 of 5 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -120,12 +120,13 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
...
 
-  // Startup script to install on the cluster master node. This allows Jupyter to start back up after
-  // a cluster is resumed.
+  // Startup scripts to install on the cluster master node.
+  // This allows Jupyter and Welder to start back up after a cluster is resumed.
   //
   // The || clause is included because older clusters may not have the run-jupyter.sh script installed,
   // so we need to fall back running `jupyter notebook` directly. See https://github.com/DataBiosphere/leonardo/issues/481.
   private lazy val masterInstanceStartupScript: immutable.Map[String, String] = {
+ //TODO
```
jmthibault79: Important TODO!  Until this is done, restarting the cluster after stopping it will NOT restart Welder.

Qi77Qi: wouldn't it just be something like adding `docker exec -d ${welderServerName} /bin/bash -c '/opt/bin/../server'

jmthibault79: yeah I don't think this will be difficult

jmthibault79:
To test:
1. launch a FIAB with leonardo image joel_IA-1026 or joel-test-3.
2. Log in to the Firecloud UI on the FIAB with a Billing Project Owner (such as Hermione)
3. Create a billing project, workspace, notebook, and cluster
4. Click on the notebook to open it (and set the appropriate auth cookie)
5. Navigate the same browser to 
https://leonardo-fiab.dsde-dev.broadinstitute.org:30443/proxy/<billing project>/<cluster name>/welder/status

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
jmthibault79: Confirmed: starting and stopping a cluster cause Jupyter to restart correctly but not Welder.

After doing this, ssh-ing into the Cluster Master Node and executing `docker exec -d -u daemon welder-server /opt/docker/bin/server start` started Welder again.

jmthibault79: This works now.

jmthibault79:
Successful test.  Ready for review. @Qi77Qi

jmthibault79:
Auto-Tests had 1 failure, regarding installing R packages.  I'm assuming that is not related to this change, so let's try again: jenkins retest

PR #960, opened 2019-06-11
Jc jupyter extension

jdcanas:
Initial commit for the jupyter UI extension to get some eyes on it.

Joel, I modified the extension you made so its toggleable based on the state stored in welder as opposed to an on-load event. This is relevant because users can go to playground mode after a time period of being in edit mode. Still need to actually call welder from it though. 

I still need to:
   -get product sign-off on UX
   -remove the console.logs
  ~~-verify the proper URLs to use for jupyter server and welder (im hoping localhost will work)~~
   -test in jupyter image
   -test in fiab with jupyter image and welder image running
      -payloads may change 

In all cases:


src/main/resources/jupyter/safe-mode.js
```diff
@@ -14,88 +14,43 @@
...
+        return true //TODO figure out what the welder link will be relative to the jupyter image
+    }
 
-    $('#notification_area').prepend(
-        '<div id="safe-mode" style="background-color: #FFFFB2;" ' +
-          'class="notification_widget btn btn-xs navbar-btn">' +
-          '<span>Safe Mode - your edits will not be saved.' +
-            '</span></div>');
-  };
+    function toggleUIControls(shouldHide) {
```
jmthibault79: Oh this is much nicer.  Thanks.

src/main/resources/jupyter/safe-mode.js
```diff
@@ -14,88 +14,43 @@
...
+        })
+    }
 
-    // TODO: convert to tooltip with this text:
-    // Safe Mode allows you to explore, change, and run the code,
-    // but your edits will not be saved.
-    // To save your work, choose Make a Copy from the File menu to
-    // make your own version.
+    function isSafeMode() {
+        return true //TODO figure out what the welder link will be relative to the jupyter image
```
jmthibault79: Let's make this false until it's working correctly

src/main/resources/jupyter/safe-mode.js
```diff
@@ -14,88 +14,43 @@
...
+    function checkMetaLoop() {
+        toggleUIControls(isSafeMode())
 
-    $('#notification_area')
-        .append(
-            '<style>' +
-              // e.g. Notebook Saved
-              '#notification_notebook span { display: none; } ' +
-              '</style>');
+        const interval = setInterval(() => {
```
jmthibault79: Is this going to change during the life of a notebook?  If not, we don't need to poll.

When we transition between safe mode and other modes, we will be copying the notebook file to a new directory.  Will that re-initialize this plugin, or will the current instance survive that transition?

docker/jupyter/custom/JupyterLockingExtension.js
```diff
@@ -0,0 +1,325 @@
...
+        },
+        'Reload the workspace version and discard your changes': {
+            click: () => updateLocalCopyWithRemote()
+        }
+    }
+
+    const lockIssueButtons = {
+        'Run in Playground Mode': {
+            click: () => openPlaygroundMode(),
+            'class': 'btn-primary'
```
jmthibault79: Why does this need a class but not the others?

docker/jupyter/custom/JupyterLockingExtension.js
```diff
@@ -0,0 +1,325 @@
...
+    }
+
+    const modeBannerId = "notification_mode"
+    const lockConflictTitle = "File is in use"
+    const syncIssueTitle = "File versions out of sync"
+    const syncIssueBody = "Your version of this file does not match the version in the workspace. What would you like to do?"
+    const syncIssueFatalBody = "This file was either deleted or never was stored with us."
+
+    //TODO URL resolution
+    const leoUrl = 'http://localhost:8080'
```
jmthibault79: Because this is client-side, localhost won't work.

The browser running the notebook will set an auth cookie so we will be able to use relative paths here, which will call out to the Leo proxy.  `welderUrl` is `proxy/<billing project>/<cluster name>/welder`

docker/jupyter/custom/JupyterLockingExtension.js
```diff
@@ -0,0 +1,325 @@
...
+        'Access-Control-Allow-Headers': '*'
+    }
+
+    const basePayload = {
+        // mode: "no-cors",
+        headers: headers
+    }
+
+    //TODO resolve these links
+    const jupyterBaseUrl = "http://localhost:8000/"
```
jmthibault79: Like above, we need to use relative path to Leo.  This is simply `notebooks`.

docker/jupyter/custom/JupyterLockingExtension.js
```diff
@@ -0,0 +1,325 @@
...
+    const jupyterContentsAPIUrl = jupyterBaseUrl + "api/contents/"
```
jmthibault79: What is this?

docker/jupyter/custom/JupyterLockingExtension.js
```diff
@@ -0,0 +1,325 @@
...
+
+    async function checkMeta() {
+        const payload = Object.assign(basePayload, {
+            body: JSON.stringify(getLocalPath()),
+            method: 'POST'
+        })
+
+        return fetch(checkMetaUrl, payload)
+            .then(res => {
+                console.log('in first then of checkmeta', res)
```
jmthibault79: rm console.logs

docker/jupyter/custom/JupyterLockingExtension.js
```diff
@@ -0,0 +1,325 @@
...
+    }
+
+    //this function assumes any status not included in these lists represents an in sync notebook to defend against future fields being added being auto-categorized as failures
+    function handleCheckMetaResp(res) {
+        meta = res //set meta state
+
+        const outOfSyncStatuses = ["DESYNCHRONIZED", "LOCAL_CHANGED", "REMOTE_CHANGED"]
+        const fatalSyncStatuses = ["REMOTE_NOT_FOUND"]
+
+        if (outOfSyncStatuses.includes(res.syncStatus)) {
```
jmthibault79: which of these 2 cases should have priority?

docker/jupyter/custom/JupyterLockingExtension.js
```diff
@@ -0,0 +1,325 @@
...
+                console.log('in catch for getLock')
+                console.error(err)
+            })
+    }
+
+    async function toggleMetaFailureBanner(shouldShow) {
+        const bannerId = "notification_metaFailure"
+        const bannerText = "Failed to check notebook status, changes may not be saved to workspace. Retrying..."
+
+        if ($("#" + bannerId).length > 0) {
```
jmthibault79: replace with removeModeBanner()

docker/jupyter/custom/JupyterLockingExtension.js
```diff
@@ -0,0 +1,325 @@
...
+                    const message = getLockConflictBody(body.lockedBy)
+                    promptUserWithModal(lockConflictTitle, lockIssueButtons, message)
+                })
+            }
+            //for the lock endpoint, we consider all non 'ok' statuses an error
+            throw new Error(errorText)
+        }
+    }
+
+    const getLockConflictBody = (lockHolder) => `<p>This file is currently being editted by ${lockHolder}.</p><br/><p>You can make a copy, or run it in Playground Mode to explore and execute its contents without saving any changes.`
```
jmthibault79: Split into multiple lines

docker/jupyter/custom/JupyterLockingExtension.js
```diff
@@ -0,0 +1,325 @@
...
+                keyboard_manager: Jupyter.notebook.keyboard_manager
+            })
+            .on('hidden.bs.modal', () => modalOpen = false)
+            .attr('id', 'leoUserModal')
+            .find(".close").click(() => window.history.back()) //TODO: test in docker image
+    }
+
+    async function openPlaygroundMode() {
+        const url = jupyterContentsAPIUrl + Jupyter.notebook.notebook_path
+        const newPath = meta.storageLink.localSafeModeBaseDirectory.replace('notebooks/', '') + '/' + Jupyter.notebook.notebook_name
```
jmthibault79: Why do we need to remove `notebooks/` here?

docker/jupyter/custom/JupyterLockingExtension.js
jmthibault79: I mean the debugging lines like this one.  Keep the informative ones.

src/main/resources/jupyter/safe-mode.js
jdcanas: I think it may be worth polling to design for the possibility of switching in the future. 

As its currently envisioned,  when a user wishes to transfer from playground mode to edit mode or vice versa it will result in notebook reload and thus the plugin will re-trigger, so polling is not needed  as is.

I'm open to removing it for now, but I think the performance overhead would be negligible.

docker/jupyter/custom/JupyterLockingExtension.js
jdcanas: Still pending discussion with Adrian on the styling of buttons outside of the lock modal. Didn't want to add styling when its possible for the UX to change. One of the other buttons may receive this class, but im not sure which one will be considered the 'primary'

docker/jupyter/custom/JupyterLockingExtension.js
jdcanas: Not sure I understand, they're mutually exclusive

docker/jupyter/custom/JupyterLockingExtension.js
jdcanas: The jupyter contents API requires the local path to the file, but it assumes all notebooks are within notebooks/ and wants it omitted

docker/jupyter/custom/JupyterLockingExtension.js
jdcanas: the Jupyter server-side api required to save the file, which is needed to change between edit and playground mode from within the plugin

docker/jupyter/custom/JupyterLockingExtension.js
jdcanas: Do you have any idea how to access this cookie or how I can test this? This has been something I haven't been sure how to test

jdcanas:
Still coming up with a testing strategy... the stuff hitting the Jupyter server I should be able to test by spinning up the proper docker image. For an e2e test, I will most likely need a fiab with both welder fully-functional and the jupyter image with this plugin. 

I'm hoping to merge this in this 'draft' state to allow future PRs to move forward with docker image/welder integration.

Currently, I tested all of this with a local mock server that returns whatever payloads/status codes I needed to test all use/edge cases.

docker/jupyter/custom/JupyterLockingExtension.js
```diff
@@ -0,0 +1,327 @@
+const dialog = require('base/js/dialog')
+const utils = require("base/js/utils")
+
+define(() => {
+    var modalOpen = false
```
calbach: let?

docker/jupyter/custom/JupyterLockingExtension.js
```diff
@@ -0,0 +1,327 @@
...
+            checkMeta()
+        }, lastLockedTimer)
+
+        window.onbeforeunload(function() {
+            clearInterval(syncMaintainer)
+        })
+    }
+
+    async function checkMeta() {
+        const payload = Object.assign(basePayload, {
```
calbach: This mutates basePayload - I don't think that's what you want. splay shallow copy is probably good here:


```
fetch(url, {
  ...basePayload,
  body: JSON.stringify(..
}
```

docker/jupyter/custom/JupyterLockingExtension.js
```diff
@@ -0,0 +1,327 @@
...
+            body: JSON.stringify(getLocalPath()),
+            method: 'POST'
+        })
+
+        return fetch(checkMetaUrl, payload)
+            .then(res => {
+                console.log('in first then of checkmeta', res)
+                if (!res.ok && res.status == 412) {
+                    promptUserWithModal(syncIssueTitle, noRemoteFileButtons, syncIssueFatalBody)
```
calbach: 412 here means no storage link. This is expected for a file which is not controlled by a storage link. THe extension should bail out at this point and not attempt to check metadata, lock, or do anything else.

src/main/resources/jupyter/safe-mode.js
jmthibault79: ok to keep

docker/jupyter/custom/JupyterLockingExtension.js
jmthibault79: ok

docker/jupyter/custom/JupyterLockingExtension.js
jmthibault79: ha.  that is true.  never mind.

docker/jupyter/custom/JupyterLockingExtension.js
jmthibault79: OK.  this will also have to be proxied thru Leo, since the UI won't have access to anything server-side.

jmthibault79:
Please rebase onto develop - and generally, do that instead of merging.  Merging adds extra commits to the PR that make it more confusing.

src/main/resources/jupyter/safe-mode.js
calbach: In Jupyter, it should not be possible to change notebooks within a single client-side session. Technically a notebook could change by way of /storageLinks, but this should never happen and would probably indicate deeper bugs if it did.

I would just lock in whatever mode you get back from the first checkMetadata request.

src/main/resources/jupyter/edit-mode.js
```diff
@@ -0,0 +1,337 @@
...
+    const modeBannerId = "notification_mode"
+    const lockConflictTitle = "File is in use"
+    const syncIssueTitle = "File versions out of sync"
+    const syncIssueBody = "Your version of this file does not match the version in the workspace. What would you like to do?"
+    const syncIssueFatalBody = "This file was either deleted or never was stored with us."
+
+    //TODO URL resolution
+    const leoUrl = ''
+        // const leoUrl = 'http://localhost:8081'
+    const welderUrl = leoUrl + `/proxy/${googleProject}/${clusterName}/welder`
```
calbach: Probably easier and can avoid the templating complexity: grab `Jupyter.notebook.base_url`, then replace "/notebooks/" with "/welder/"

src/main/resources/jupyter/edit-mode.js
```diff
@@ -0,0 +1,337 @@
...
+                if (!res.ok && res.status == 412) {
+                    promptUserWithModal(syncIssueTitle, noRemoteFileButtons, syncIssueFatalBody)
+                } else if (!res.ok && res.status >= 400) {
+                    throw Error(res.statusText)
+                }
+                return res.json()
+            })
+            .then(res => {
+                console.log('res in check meta after .json()', res)
+                handleCheckMetaResp(res) //sets meta state
```
calbach: optional: if lockedBy != null we could show the locked dialog at this point, rather than trying to acquire it

src/main/resources/jupyter/edit-mode.js
```diff
@@ -0,0 +1,337 @@
...
+            })
+            .catch(err => {
+                console.log('in checkMeta catch')
+                console.error(err)
+                removeElementById(modeBannerId)
+                toggleMetaFailureBanner(true)
+            })
+    }
+
+    //this function assumes any status not included in these lists represents an in sync notebook to defend against future fields being added being auto-categorized as failures
```
calbach: It would be better to autocategorize future statuses as failures. LIVE should be the only good status for the foreseeable future. This is a good thing to call out though - luckily Welder and the UI extension are deployed in tandem so we should be pretty insulated against server/client skew in this case.

src/main/resources/jupyter/edit-mode.js
```diff
@@ -0,0 +1,337 @@
...
+    function handleCheckMetaResp(res) {
+        meta = res //set meta state
+
+        const outOfSyncStatuses = ["DESYNCHRONIZED", "LOCAL_CHANGED", "REMOTE_CHANGED"]
+        const fatalSyncStatuses = ["REMOTE_NOT_FOUND"]
```
calbach: THis is actually less fatal than the other cases - I'd probably rename. The user should be able to continue and just save normally.

src/main/resources/jupyter/edit-mode.js
```diff
@@ -0,0 +1,337 @@
...
+                    promptUserWithModal(lockConflictTitle, lockIssueButtons, message)
+                })
+            }
+            //for the lock endpoint, we consider all non 'ok' statuses an error
+            throw new Error(errorText)
+        }
+    }
+
+    const getLockConflictBody = (lockHolder) => {
+        return `<p>This file is currently being editted by ${lockHolder}.</p>` +
```
calbach: edited

calbach: Unfortunately we won't have the specifics here (per the user metadata privacy doc) - will just have to say "by another user"

src/main/resources/jupyter/edit-mode.js
```diff
@@ -0,0 +1,337 @@
...
+
+    async function openPlaygroundMode() {
+        const url = jupyterContentsAPIUrl + Jupyter.notebook.notebook_path
+        const newPath = meta.storageLink.localSafeModeBaseDirectory.replace('notebooks/', '') + '/' + Jupyter.notebook.notebook_name
+
+        console.log('switching to playground path: ', newPath)
+
+        const payload = {
+            ...basePayload,
+            method: 'PATCH',
```
calbach: What are we patching here? Can you add a comment? The semantics I'd expect here are:

- copy or "save as" to playground mode directory
- redirect to new file

src/main/resources/jupyter/edit-mode.js
```diff
@@ -0,0 +1,337 @@
...
+                keyboard_manager: Jupyter.notebook.keyboard_manager
+            })
+            .on('hidden.bs.modal', () => modalOpen = false)
+            .attr('id', 'leoUserModal')
+            .find(".close").click(() => window.history.back()) //TODO: test in docker image
+    }
+
+    async function openPlaygroundMode() {
+        const url = jupyterContentsAPIUrl + Jupyter.notebook.notebook_path
+        const newPath = meta.storageLink.localSafeModeBaseDirectory.replace('notebooks/', '') + '/' + Jupyter.notebook.notebook_name
```
calbach: "notebooks" is not guaranteed to be in this string; to form the playground mode path:

- take full relative notebook path, e.g. dir/subdir/a.ipnyb
- replace localBaseDirectory with localSafeModeBaseDirectory (should both be prefixes of the path)

src/main/resources/jupyter/edit-mode.js
```diff
@@ -0,0 +1,337 @@
...
+        const payload = {
+            ...basePayload,
+            method: 'PATCH',
+            body: JSON.stringify({ path: newPath })
+        }
+
+        console.log('url: ', url, newPath)
+
+        fetch(url, payload).then(res => {
+            window.location.href = jupyterBaseUrl + "notebooks/" + newPath
```
calbach: You should be able to simplify all of this just using relative paths, e.g. `window.location.href = "/notebooks/" + newPath`

If there's a basePath set, you could even go potentially simpler: `window.location.href = newPath`

src/main/resources/jupyter/init-actions.sh
```diff
@@ -323,6 +324,15 @@ if [[ "${ROLE}" == 'Master' ]]; then
         docker cp /etc/${JUPYTER_SAFE_MODE_PLUGIN} ${JUPYTER_SERVER_NAME}:${JUPYTER_USER_HOME}/.jupyter/custom/
       fi
 
+      # If edit-mode.js was specified, copy it into the jupyter docker container.
+        if [ ! -z ${JUPYTER_EDIT_MODE_PLUGIN_URI} ] ; then
+          log 'Installing edit-mode.js extension...'
+          gsutil cp ${JUPYTER_EDIT_MODE_PLUGIN_URI} /etc
+          JUPYTER_EDIT_MODE_PLUGIN=`basename ${JUPYTER_EDIT_MODE_PLUGIN_URI}`
+          retry 3 docker exec ${JUPYTER_SERVER_NAME} mkdir -p ${JUPYTER_USER_HOME}/.jupyter/custom
```
rtitle: Just curious, is there a reason to use `~/.jupyter/custom` instead of installing this as a "normal" extension (see the block starting on L244)?

Also I'm wondering if this works out of the box alongside `google_sign_in.js` (which is also in `~/jupyter/custom`). Am wondering if this needs updating to reference `edit-mode.js` and `safe-mode.js`: https://github.com/DataBiosphere/leonardo/blob/develop/src/main/resources/jupyter/google_plugin_jupyter.js

src/main/resources/jupyter/edit-mode.js
jmthibault79: That doesn't match the Leonardo/Welder proxy URL pattern (see [cluster-site.conf](https://github.com/DataBiosphere/leonardo/blob/develop/src/main/resources/jupyter/cluster-site.conf).

Jupyter -> `leoBaseUrl/notebooks/${googleProject}/${clusterName}/jupyter`
Welder ->  `leoBaseUrl/proxy/${googleProject}/${clusterName}/welder/${pathToWelderEndpoint}`

We could make it more symmetrical as you suggest, but this would require adding a new proxy route to Leonardo, in addition to updating cluster-site.conf.  I chose the `/proxy/` pattern simply because that was easier: cluster-site.conf only.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -120,24 +120,13 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
...
 
-  // Startup scripts to install on the cluster master node.
-  // This allows Jupyter and Welder to start back up after a cluster is resumed.
+  // Startup script to install on the cluster master node. This allows Jupyter to start back up after
+  // a cluster is resumed.
+  //
+  // The || clause is included because older clusters may not have the run-jupyter.sh script installed,
+  // so we need to fall back running `jupyter notebook` directly. See https://github.com/DataBiosphere/leonardo/issues/481.
   private lazy val masterInstanceStartupScript: immutable.Map[String, String] = {
-    val googleKey = "startup-script"  // required; see https://cloud.google.com/compute/docs/startupscript
```
rtitle: Seems like you removed the welder startup script. Will that be re-added in a different PR?

src/main/resources/jupyter/edit-mode.js
calbach: OK, this seems fine then.

src/main/resources/jupyter/edit-mode.js
jdcanas: There was a discussion around this, and we determined that it would be best to leave it up to the API to determine the lock status, as its best not to trust client-side logic to determine something that could impact other users. Additionally, it allows looser coupling.

src/main/resources/jupyter/edit-mode.js
```diff
@@ -0,0 +1,337 @@
...
+                } else if (!res.ok && res.status >= 400) {
+                    throw Error(res.statusText)
+                }
+                return res.json()
+            })
+            .then(res => {
+                console.log('res in check meta after .json()', res)
+                handleCheckMetaResp(res) //sets meta state
+                toggleMetaFailureBanner(false) //sets banner for meta status
+                maintainLock(res) //gets lock
```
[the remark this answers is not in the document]
jdcanas: It does only get the lock if res.syncMode == 'Edit', but it needs to display a mode banner regardless, as we are collapsing that portion of the safe-mode extension into this extension to avoid the possibility of conflicting banners. I renamed the method to maintainLockState to clarify.

src/main/resources/jupyter/edit-mode.js
calbach: I'm not suggesting any client-side logic. The Welder API has a `lastLockedBy` field returned in the metadata response, it was supposed to be null if you can lock - non-null if locked by someone else. If that's changed, we should stop returning it in the response JSON and update the design doc.

src/main/resources/jupyter/edit-mode.js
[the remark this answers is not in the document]
calbach: OK - in my mind safe mode and locking are unrelated, so it doesn't make sense to me that `maintainLock` results in rendering of the banner (for edit or safe mode). I might suggest restructuring the tree of function calls here, but agreed it looks like the current end behavior is correct.

src/main/resources/jupyter/edit-mode.js
jdcanas: yeah, this logic was messy because the Jupyter server automatically prefixes the path to all notebooks with notebooks/ with my local config. I expected once I get it running in the docker image that this will change.

src/main/resources/jupyter/edit-mode.js
jdcanas: As I understand it, we will need to call the leo jupyter proxy endpoint (jupyterBaseUrl), which will then proxy the call to the Jupyter server at the API path specified after the leo path. I had this working locally on my Jupyter server, but this will need more testing once I have the extension in the jupyter image. 

That being said, I didn't know you could specify a relative path when assigning window.location.href, so I'll investigate simplifying this

src/main/resources/jupyter/edit-mode.js
jdcanas: added

src/main/resources/jupyter/edit-mode.js
```diff
@@ -0,0 +1,404 @@
...
+
+    const modeBannerId = "notification_mode"
+    const lockConflictTitle = "File is in use"
+    const syncIssueTitle = "File versions out of sync"
+    const syncIssueBody = "Your version of this file does not match the version in the workspace. What would you like to do?"
+    const syncIssueNotFoundBody = "This file was either deleted or never was stored with us."
+
+    //TODO URL resolution
+    // const leoUrl = ''
+    const leoUrl = 'http://localhost:8080'
```
rtitle: This is for development purposes I assume.

src/main/resources/reference.conf
```diff
@@ -25,10 +25,11 @@ clusterResources {
   proxyDockerCompose = "proxy-docker-compose.yaml"
   welderDockerCompose = "welder-docker-compose.yaml"
   proxySiteConf = "cluster-site.conf"
-  googleSignInJs = "google_sign_in.js"
-  jupyterGooglePlugin = "google_plugin_jupyter.js"
+  googleSignInJs = "google_sign_in.js" //TODO make naming convention for nbextension js files
+  jupyterGooglePlugin = "google_plugin_jupyter.js" //TODO rename this because its name is misleading as to its actual purpose?
```
rtitle: +1 on renaming this, it's really an entrypoint for ALL js extensions, not just the google one

src/main/resources/reference.conf
```diff
@@ -25,10 +25,11 @@ clusterResources {
   proxyDockerCompose = "proxy-docker-compose.yaml"
   welderDockerCompose = "welder-docker-compose.yaml"
   proxySiteConf = "cluster-site.conf"
-  googleSignInJs = "google_sign_in.js"
-  jupyterGooglePlugin = "google_plugin_jupyter.js"
+  googleSignInJs = "google_sign_in.js" //TODO make naming convention for nbextension js files
```
rtitle: +1 on refactoring this config. Instead of individual configs for each extension, could have something like:
```
# the entrypoint
jupyterPlugin = "jupyter_plugin.js"
# the actual list of extensions which get templated into jupyter_plugin.js
jupyterExtensions = [ "list", "of", "jupyter", "extensions" ]
```
I haven't thought this through deeply and could probably wait for a separate refactoring ticket outside of this PR.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -952,14 +956,19 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
...
 
     for {
       // Upload the init script to the bucket
       _ <- leoGoogleStorageDAO.storeObject(initBucketName, GcsObjectName(clusterResourcesConfig.initActionsScript.value), initScriptContent, "text/plain")
 
-      // Upload the googleSignInJs file to the bucket
+      // Upload the nb extensions to the bucket
       _ <- leoGoogleStorageDAO.storeObject(initBucketName, GcsObjectName(clusterResourcesConfig.googleSignInJs.value), googleSignInJsContent, "text/plain")
+      _ <- leoGoogleStorageDAO.storeObject(initBucketName, GcsObjectName(clusterResourcesConfig.jupyterEditModePlugin.value), editModeJsContent, "text/plain")
+      _ <- leoGoogleStorageDAO.storeObject(initBucketName, GcsObjectName(clusterResourcesConfig.jupyterSafeModePlugin.value), safeModeJsContent, "text/plain")
```
rtitle: It _seems_ like the templating should be occurring based on this code. Are you not seeing it working?

src/main/resources/reference.conf
jdcanas: done

src/main/resources/reference.conf
jdcanas: +1 on doing this in another PR

src/main/resources/jupyter/edit-mode.js
calbach: Per offline discussion: want copy rather than move, here and below (save as)

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -120,8 +120,11 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
     network = dataprocConfig.vpcNetwork.map(VPCNetworkName),
     targetTags = List(NetworkTag(dataprocConfig.networkTag)))
 
-  // Startup scripts to install on the cluster master node.
-  // This allows Jupyter and Welder to start back up after a cluster is resumed.
+  // Startup script to install on the cluster master node. This allows Jupyter to start back up after
+  // a cluster is resumed.
+  //
+  // The || clause is included because older clusters may not have the run-jupyter.sh script installed,
```
rtitle: I notice the comment `The || clause ...` is duplicated down below. Probably can remove it here.

src/main/resources/jupyter/edit-mode.js
```diff
@@ -0,0 +1,396 @@
+const dialog = require('base/js/dialog')
+const utils = require("base/js/utils")
+
+define(() => {
+    // TEMPLATED CODE
+    // Leonardo has logic to find/replace templated values in the format $(...).
+    const googleProject = $(googleProject);
```
calbach: I don't think Leo will subtitute quotes for you, should probably be `"$(googleProject)"`

src/main/resources/jupyter/edit-mode.js
```diff
@@ -0,0 +1,396 @@
...
+        mode: "no-cors",
+        headers: headers
+    }
+
+    const jupyterBaseUrl = `/notebooks/${googleProject}/${clusterName}`
+    const jupyterContentsAPIUrl = jupyterBaseUrl + "/api/contents/"
+
+    function init() {
+        console.log('here in edit mode init')
+            // checkMeta()
```
calbach: Be sure to revert this before merging - extension doesn't do anything without this

PR #963, opened 2019-06-12
Document that Leo requires Java 8

helgridly:
I tried to build Leo on my dev machine, which has Java 11 installed, and couldn't. As of Java 9, the `Nameservice` stuff [has been removed](https://bugs.openjdk.java.net/browse/JDK-8134577). This is causing [some consternation within the Java community](https://bugs.openjdk.java.net/browse/JDK-8192780) -- Google is unhappy, for one -- but at the very least, we should document the requirement.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,9 @@ $ cd leonardo
 ```
 
 ### Run Leonardo unit tests
+
+Leonardo requires Java 8 due to a dependency on Java's DNS SPI functionality. This feature is removed in Java 9 and above.
+
 Ensure docker is running. Spin up MySQL locally:
 ```
 $ ./docker/run-mysql.sh start leonardo  


README.md
```diff
@@ -41,6 +41,9 @@ $ cd leonardo
 ```
 
 ### Run Leonardo unit tests
+
+Leonardo requires Java 8 due to a dependency on Java's DNS SPI functionality. This feature is removed in Java 9 and above.
```
Qi77Qi: I think we should always assume scala apps are only reliable on java 8, which is the only targeted java version.

helgridly: you mean scala in general only targets java 8?

Qi77Qi: Yea. 2.13’s official target is also java 8.

rtitle:
Ah crap, good catch. If we ever move to >8 we'll need to figure out how to do the nameservice stuff.

Qi77Qi:
created https://broadworkbench.atlassian.net/browse/IA-1056. I don't think we need to worry about this for a while, but good to have it in backlog

PR #964, opened 2019-06-14
Change getCluster to listCluster in cluster create

vkumra-broad:

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
@@ -195,16 +195,6 @@ trait ClusterComponent extends LeoComponent {
         }
     }
 
-    def listActiveClusterByName(project: GoogleProject, name: ClusterName): DBIO[Option[Cluster]] = {
-      clusterLabelQuery
-        .filter { _._1.googleProject === project.value }
-        .filter { _._1.clusterName === name.value }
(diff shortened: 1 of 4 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
```diff
@@ -195,6 +195,16 @@ trait ClusterComponent extends LeoComponent {
         }
     }
 
+    def listActiveClusterByName(project: GoogleProject, name: ClusterName): DBIO[Option[Cluster]] = {
```
rtitle: Usually `list` methods return a `Seq`, I would name this with `get`.

Actually looking below, it seems like `getActiveClusterForDnsCache` already does what we want. Maybe rename that method to `getActiveClusterByNameMinimal` and use that in `LeonardoService.initiateClusterCreation`.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
```diff
@@ -195,6 +195,16 @@ trait ClusterComponent extends LeoComponent {
...
+      clusterLabelQuery
```
rtitle: I don't think labels are even needed here. See above comment.

rtitle (changes requested):
See above

rtitle (approved):

PR #965, opened 2019-06-14
Use high-security networks when requested by project

helgridly:
friend: https://github.com/broadinstitute/workbench-libs/pull/213

In all cases:

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -11,8 +11,8 @@ object Dependencies {
   val catsV         = "1.3.1"
 
   val workbenchUtilV    = "0.5-4c7acd5"
-  val workbenchModelV   = "0.11-2bddd5b"
-  val workbenchGoogleV  = "0.19-ad7ff9d"
+  val workbenchModelV   = "0.13-6dc016b"
+  val workbenchGoogleV  = "0.19-6dc016b"
   val workbenchGoogle2V = "0.2-2149dba"
   val workbenchMetricsV = "0.3-c5b80d2"
 
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -12,7 +12,7 @@ object Dependencies {
 
   val workbenchUtilV    = "0.5-4c7acd5"
   val workbenchModelV   = "0.11-2bddd5b"
-  val workbenchGoogleV  = "0.18-39dba9f-SNAP"
+  val workbenchGoogleV  = "0.19-ad7ff9d"
   val workbenchGoogle2V = "0.2-2149dba"
   val workbenchMetricsV = "0.3-c5b80d2"
 
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -12,7 +12,7 @@ object Dependencies {
 
   val workbenchUtilV    = "0.5-4c7acd5"
   val workbenchModelV   = "0.11-2bddd5b"
-  val workbenchGoogleV  = "0.18-6942040"
+  val workbenchGoogleV  = "0.18-39dba9f-SNAP"
   val workbenchGoogle2V = "0.2-2149dba"
   val workbenchMetricsV = "0.3-c5b80d2"
 
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -9,6 +9,8 @@ dataproc {
   welderServerName = "welder-server"
   firewallRuleName = "leonardo-notebooks-rule"
   networkTag = "leonardo"
+  projectVPCNetworkLabel = "vpc-network-name"
+  projectVPCSubnetLabel = "vpc-subnetwork-name"
   defaultScopes = ["https://www.googleapis.com/auth/userinfo.email", "https://www.googleapis.com/auth/userinfo.profile", "https://www.googleapis.com/auth/bigquery", "https://www.googleapis.com/auth/source.read_only"]
   # TODO: temp 2019-06-07
   welderDockerImage = "us.gcr.io/broad-dsp-gcr-public/welder-server:0b28747"
--- a/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/google/HttpGoogleDataprocDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/google/HttpGoogleDataprocDAO.scala
@@ -33,8 +33,6 @@ class HttpGoogleDataprocDAO(appName: String,
                             googleCredentialMode: GoogleCredentialMode,
                             override val workbenchMetricBaseName: String,
                             networkTag: NetworkTag,
-                            vpcNetwork: Option[VPCNetworkName],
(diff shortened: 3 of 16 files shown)


src/main/resources/reference.conf
```diff
@@ -9,6 +9,8 @@ dataproc {
   welderServerName = "welder-server"
   firewallRuleName = "leonardo-notebooks-rule"
   networkTag = "leonardo"
+  projectVPCNetworkLabel = "vpc-network-name"
+  projectVPCSubnetLabel = "vpc-subnetwork-name"
```
helgridly: Do you want this here, or in fc-dev leo.conf?

project/Dependencies.scala
```diff
@@ -12,7 +12,7 @@ object Dependencies {
 
   val workbenchUtilV    = "0.5-4c7acd5"
   val workbenchModelV   = "0.11-2bddd5b"
-  val workbenchGoogleV  = "0.18-6942040"
+  val workbenchGoogleV  = "0.18-39dba9f-SNAP"
```
rtitle: Is there an associated wb-libs PR? We try not to have `-SNAP` dependencies.

src/main/resources/reference.conf
rtitle: I think here is fine, unless we envision this varying by environment. 

Can you also update `leonardo-example.conf` and add some documentation?

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -731,6 +734,33 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
     }
   }
 
+
+  private def getClusterNetwork(projectLabels: Map[String, String]): Option[VPCNetworkName] = {
+    //if we specify a network label in config, we should use that to see if the project has a network specified
+    //if it does, we will use that, else we will look to see if leo config has a network specified, else we will use the default
+    dataprocConfig.projectVPCNetworkLabel match {
```
rtitle:
```
dataprocConfig.projectVPCNetworkLabel.flatMap(projectLabels.get).orElse(dataprocConfig.vpcNetwork).map(VPCNetworkName)
```
?

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -717,8 +718,10 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
       autopauseThreshold = calculateAutopauseThreshold(clusterRequest.autopause, clusterRequest.autopauseThreshold)
       clusterScopes = if(clusterRequest.scopes.isEmpty) dataprocConfig.defaultScopes else clusterRequest.scopes
       credentialsFileName = serviceAccountInfo.notebookServiceAccount.map(_ => s"/etc/${ClusterInitValues.serviceAccountCredentialsFilename}")
-
-      createClusterConfig = CreateClusterConfig(machineConfig, initScript, serviceAccountInfo.clusterServiceAccount, credentialsFileName, stagingBucket, clusterScopes, clusterRequest.properties)
+      projectLabels <- googleProjectDAO.getLabels(googleProject.value)
```
rtitle: Minor: should we only make this google DAO call if `projectVPCNetworkLabel` or `projectVPCSubnetLabel` is defined in config?

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/google/HttpGoogleDataprocDAO.scala
```diff
@@ -213,7 +211,16 @@ class HttpGoogleDataprocDAO(appName: String,
       val baseConfig = new GceClusterConfig()
         .setTags(List(networkTag.value).asJava)
 
-      (vpcNetwork, vpcSubnet) match {
+      //Dataproc only allows you to specify a subnet OR a network. Subnets will be preferred if present.
+      //High-security networks specified inside of the project will always take precedence over anything
+      //else. Thus, VPC configuration takes the following precedence:
+      // 1) High-security subnet in the project (if present)
```
rtitle: I'm wondering if these rules always hold true, since the code seems to treat subnets as highest priority. For example if I had the following in leonardo.conf:
- `vpcSubnet`
- `projectVPCNetworkLabel`
I think it would pick the subnet over the high security network, right?

rtitle (changes requested):
A few comments above

project/Dependencies.scala
helgridly: i'll have to check with matt. i think there's something somewhere

helgridly: ta-da https://github.com/broadinstitute/workbench-libs/pull/213

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
helgridly: my brain is really breaking on reading "project labels dot get or else" without it actually being `.getOrElse`, and i'm low-level anxious about which way to associate the three function calls. i'll take a poke at this and see if i can come up with something that inspires more chill in me

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/google/HttpGoogleDataprocDAO.scala
helgridly: yup, i agree. `LeoService.getClusterNetwork` and `getClusterSubnet` both have the behaviour of "lookup project label first, then fall back to leo.conf", and pass that into the `ClusterConfiguration` here, and then _this_ code picks subnet over network every time.
@MatthewBemis - is the comment the intended behaviour, or is the code?

MatthewBemis: I believe the comment was the intended behavior

helgridly:
I changed how we choose VPC subnet and network so the code matches the comment. @rtitle @MatthewBemis can you re-review? Direct link to the altered commit: https://github.com/DataBiosphere/leonardo/pull/965/commits/730199a34a67ba343e2feeca78b64c058fc6a05e

rtitle (approved):
This looks good to me 👍

rtitle:
Except... Travis and Jenkins are both unhappy. Is a unit test failing?

helgridly:
Looks like. I'll fix it

rtitle:
@helgridly I think what happened is this PR bumped the `spray-json` version and broke an assertion where we were checking a spray exception message. I don't think that assertion was useful so I just pushed a commit to remove it. Let's see how tests do.

rtitle:
Strange build failures, looks due to infrastructure. jenkins retest

PR #967, opened 2019-06-17
Add local docker compose for Jupyter + Welder

calbach:
Proof-of-concept local development setup for Jupyter + Welder.

Usage:

```
# One-time setup
gcloud auth application-default login
chmod a+r ~/.config/gcloud/application_default_credentials.json # Not ideal, should find a better workaround

# Local welder changes
welder$ sbt server/docker:publishLocal

docker/local/run.sh

# Make changes to Jupyter extension.
docker/local/run.sh
```

Limitations:
- Does not simulate the proxy; this could be added via an nginx docker service
- Does not run the Leo init-actions script
- Does not currently allow hot reload of Welder or Jupyter extensions
- Currently uses the "latest" Welder label to pickup local builds, would be better to use an explicit local tag.
- Need to expand ACLs on ADC JSON file (can probably find another workaround for this).
- Having some issues with Kernels locally - possible a websocket issue

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 .metals
 /project/boot/
 /project/plugins/project
+docker/local/etc-jupyter
 target/
 lib_managed/
 src_managed/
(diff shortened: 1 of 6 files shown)


calbach:
Better version: https://github.com/DataBiosphere/leonardo/pull/972

PR #968, opened 2019-06-18
Implement initial Welder server extension

calbach:
Still needs tests.

In all cases:

### DIFF
--- a/docker/jupyter/custom/jupyter_delocalize.py
+++ b/docker/jupyter/custom/jupyter_delocalize.py
@@ -185,7 +185,7 @@ class WelderContentsManager(FileContentsManager):
       # not; strip for Welder.
       'localPath': path.lstrip('/')
     }))
-    if not resp.ok
+    if not resp.ok:
       try:
         msg = json.dumps(resp.json())
       except:
--- a/docker/jupyter/custom/jupyter_delocalize.py
+++ b/docker/jupyter/custom/jupyter_delocalize.py
@@ -175,7 +175,7 @@ class WelderContentsManager(FileContentsManager):
     # lacking and its configuration can be complex. Having this in the server
     # logs is useful for confirming which ContentsManager is in use.
     self.log.info('initializing WelderContentsManager')
-    self.welder_base_url = 'http://localhost:8080'
+    self.welder_base_url = 'http://127.0.0.1:8080'
     super(WelderContentsManager, self).__init__(*args, **kwargs)
 
   def _welder_delocalize(self, path):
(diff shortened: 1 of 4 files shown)


docker/jupyter/custom/jupyter_delocalize.py
```diff
@@ -158,3 +159,61 @@ def delete_file(self, path):
...
+
+  def _welder_delocalize(self, path):
+    # Ignore storage link failure, throw other errors.
+    resp = requests.post('http://localhost:8080/objects', data=json.dumps({
+      'action': 'safeDelocalize',
+      # Sometimes the Jupyter UI provided "path" contains a leading /, sometimes
+      # not; strip for Welder.
+      'localPath': path.lstrip('/')
+    }))
+    if resp.status_code != requests.codes.ok:
```
Qi77Qi: does `requests.code.ok` include 204? I recently changed return code to 204...The fact it was 200 previously was an oversight

calbach: Good catch, done.

docker/jupyter/custom/jupyter_delocalize.py
```diff
@@ -158,3 +159,61 @@ def delete_file(self, path):
...
+  def __init__(self, *args, **kwargs):
+    # This log line shouldn't be necessary, but Jupyter's built-in logging is
+    # lacking and its configuration can be complex. Having this in the server
+    # logs is useful for confirming which ContentsManager is in use.
+    self.log.info('initializing WelderContentsManager')
+    super(WelderContentsManager, self).__init__(*args, **kwargs)
+
+  def _welder_delocalize(self, path):
+    # Ignore storage link failure, throw other errors.
+    resp = requests.post('http://localhost:8080/objects', data=json.dumps({
```
rtitle: Does `localhost` work from inside the notebook container? I think I remember having issues with localhost and had to use `127.0.0.1`. Could be wrong though.

rtitle:
Looks fine to me. jenkins retest for good measure.

docker/jupyter/custom/jupyter_delocalize.py
calbach: It works when I test it locally with docker-compose; I think network_mode: host should allow it. Switched to `127.0.0.1` for at least consistency with cluster-site.conf; verified this still works with docker-compose as well.

calbach:
Added some tests, which turned up a few bugs. This should be good now.

calbach:
jenkins retest (?)

PR #970, opened 2019-06-21
Flag to enable welder per cluster

akarukappadath:

In all cases:


src/main/resources/reference.conf
```diff
@@ -11,7 +11,7 @@ dataproc {
   networkTag = "leonardo"
   defaultScopes = ["https://www.googleapis.com/auth/userinfo.email", "https://www.googleapis.com/auth/userinfo.profile", "https://www.googleapis.com/auth/bigquery", "https://www.googleapis.com/auth/source.read_only"]
   # TODO: temp 2019-06-07
-  welderDockerImage = "us.gcr.io/broad-dsp-gcr-public/welder-server:0b28747"
+  welderDockerImage = "us.gcr.io/broad-dsp-gcr-public/welder-server:latest"
```
Qi77Qi: we don't want to always use latest tag because that means we can potentially be using unstable welder in prod that's not tested...Before merging this PR, we should update this with an appropriate hash (we don't have it yet, but hopefully, once https://github.com/DataBiosphere/welder/pull/15 is merged, we'll have a good tag to use

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
```diff
@@ -50,6 +48,12 @@ case class ServiceAccountInfoRecord(clusterServiceAccount: Option[String],
                                     notebookServiceAccount: Option[String],
                                     serviceAccountKeyId: Option[String])
 
+case class AuditInfoRecord(creator: String,
+                           createdDate: Timestamp,
+                           destroyedDate: Timestamp,
+                           dateAccessed: Timestamp)
```
Qi77Qi: can you add `final` for this case class...

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
```diff
@@ -37,7 +37,8 @@ final case class ClusterRequest(labels: LabelMap,
                           defaultClientId: Option[String] = None,
                           jupyterDockerImage: Option[String] = None,
                           rstudioDockerImage: Option[String] = None,
-                          scopes: Set[String] = Set.empty)
+                          scopes: Set[String] = Set.empty,
+                          enableWelder: Option[Boolean] = None)
```
Qi77Qi: I'll make this default to `false`. At call site, you'll branching out based on only `true`, `false` anyway.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -190,6 +212,29 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
     } flatMap {
       case Some(existingCluster) => throw ClusterAlreadyExistsException(googleProject, clusterName, existingCluster.status)
       case None =>
+        clusterRequest.enableWelder match {
+          case Some(true) =>
+            dbRef.inTransaction { dataAccess =>
+              dataAccess.clusterQuery.getClustersWithWelderDisabledByProject(googleProject)
+            } flatMap { clusterSeq =>
+              if (clusterSeq.isEmpty) {
+                Future.successful(())
```
Qi77Qi: fyi, there's `Future.unit`

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
```diff
@@ -295,6 +303,21 @@ trait ClusterComponent extends LeoComponent {
         .result map { recs => unmarshalFullCluster(recs)}
     }
 
+    // TODO: find how to make these not full cluster queries
+    def getClustersWithWelderEnabledByProject(project: GoogleProject): DBIO[Seq[Cluster]] = {
+      fullClusterQuery
+        .filter { _._1.googleProject === project.value }
+        .filter { _._1.welderEnabled === true }
+        .result map { recs => unmarshalFullCluster(recs)}
```
Qi77Qi: I think you only need to know whether there exists cluster. Something like `def existClustersWithWelderEnabled(project: GoogleProject): DBIO[Boolean]`

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -140,6 +146,22 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
     immutable.Map(googleKey -> servicesStart)
   }
 
+  protected def getMasterInstanceStartupScript(welderEnabled: Boolean): immutable.Map[String, String] = {
```
jmthibault79: Remove `masterInstanceStartupScript` because this is its replacement.

jmthibault79:
Looks good generally 👍

src/main/resources/reference.conf
Qi77Qi: you can use `us.gcr.io/broad-dsp-gcr-public/welder-server:fd981fe`

src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20190613_welder_enabled.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="leonardo" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+    <changeSet logicalFilePath="leonardo" author="azmina" id="welder_enabled">
+        <addColumn tableName="CLUSTER">
+            <column name="welderEnabled" type="boolean">
+                <constraints nullable="false" />
+            </column>
+        </addColumn>
+    </changeSet>
+    <changeSet id="set welderEnabled to default when welderEnabled is NULL" author="azmina">
```
rtitle: Is this extra changeSet needed? I seem to recall liquibase/mysql automatically setting new non-nullable boolean columns to `false`.

src/main/resources/swagger/api-docs.yaml
```diff
@@ -1082,6 +1082,10 @@ definitions:
         description: >
           The scopes for the cluster. Defaults (userinfo.email, userinfo.profile, bigquery, source.read_only) will be used if left blank. Important: If you choose to specify custom scopes, the defaults will be overwritten. Thus, if you need the defaults, you will need to include the default scopes in your custom list of scopes.
         default: ["https://www.googleapis.com/auth/userinfo.email", "https://www.googleapis.com/auth/userinfo.profile", "https://www.googleapis.com/auth/bigquery", "https://www.googleapis.com/auth/source.read_only"]
+      enableWelder:
+        type: boolean
+        description: If true, enables welder on the cluster if all cluster in the google project have welder enabled. If false, disables welder on the cluster if all other cluster in google project have welder disabled.
```
rtitle: I'm not sure I understand this description. Is it looking at other clusters in the project? What is the behavior if I try to enable welder on 1 cluster but not another (in the same project)?

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
```diff
@@ -295,6 +303,21 @@ trait ClusterComponent extends LeoComponent {
         .result map { recs => unmarshalFullCluster(recs)}
     }
 
+    // TODO: find how to make these not full cluster queries
```
rtitle: +1, definitely avoid `fullClusterQuery` if you can. I'm learning in perf testing that it has some performance issues...

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
```diff
@@ -295,6 +303,21 @@ trait ClusterComponent extends LeoComponent {
...
+    def getClustersWithWelderEnabledByProject(project: GoogleProject): DBIO[Seq[Cluster]] = {
+      fullClusterQuery
+        .filter { _._1.googleProject === project.value }
+        .filter { _._1.welderEnabled === true }
```
rtitle: This column should have an index if we're filtering on it

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
rtitle: It needs to be an `Option[Boolean]` to denote an optional json field. I think defaulting to None is appropriate for Option.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -190,6 +212,29 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
     } flatMap {
       case Some(existingCluster) => throw ClusterAlreadyExistsException(googleProject, clusterName, existingCluster.status)
       case None =>
+        clusterRequest.enableWelder match {
```
rtitle: This match statement doesn't look correct -- it returns a Future but that Future isn't actually chained to anything, so it probably doesn't have any effect.

rtitle: I'm also hazy on why we're checking other clusters in the project here. Did that requirement come from @sharmatime? Why should welder enablement be dependent on other clusters in the project?

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
rtitle: I'm curious why this requirement exists in the first place..

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
rtitle: +1

akarukappadath: I left it there because there was a discussion around removing the enableWelder field for cluster creation in the future when welder is released on both AoU and Terra, so we could go back to using `masterInstanceStartupScript`

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
akarukappadath: I think we're planning to remove this enableWelder field once welder has been released to both AoU and Terra. For the time between those releases, this is just a check that both welder-enabled and welder-disabled clusters don't exist in the same google project, since it would mean that people's work could get clobbered even though welder was on the cluster

rtitle: I'm not really a fan of `createCluster` querying _other_ users' clusters, which the calling user may not have permission to do. I feel this logic lies outside of the responsibility of Leo. A project owner should be able to call `listClusters` to check welder enablement across clusters. The logic to roll out welder to projects should be handled by the parent app (e.g. AoU, Terra).

akarukappadath: Hm, I thought it would be similar to how we query for any other clusters in the same project with the same clusterName. Is the concern that the query returns a list of other users' clusters? I recently made a change so that the LeoService only sees a boolean that tells if there are welder enabled/disabled clusters in the project, rather than listing them, if that makes a difference. @Qi77Qi, was this check something that the apps asked for, or just part of the original design?

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
```diff
@@ -295,6 +303,22 @@ trait ClusterComponent extends LeoComponent {
         .result map { recs => unmarshalFullCluster(recs)}
     }
 
+    def existsClustersWithWelderEnabled(project: GoogleProject): DBIO[Boolean] = {
+      clusterQuery
+        .filter { _.googleProject === project.value }
```
rtitle: By the way, you probably want to filter on _active_ clusters here. We don't care if welder is enabled/disabled on deleted clusters.

src/main/resources/jupyter/init-actions.sh
```diff
@@ -103,6 +103,7 @@ if [[ "${ROLE}" == 'Master' ]]; then
     JUPYTER_USER_SCRIPT_URI=$(jupyterUserScriptUri)
     JUPYTER_USER_SCRIPT_OUTPUT_URI=$(jupyterUserScriptOutputUri)
     JUPYTER_NOTEBOOK_CONFIG_URI=$(jupyterNotebookConfigUri)
+    WELDER_ENABLED=$(welderEnabled)
```
rtitle: One thing I noticed which could maybe be fixed in this PR. Here the variable is named `WELDER_ENABLED` whereas [here](https://github.com/DataBiosphere/leonardo/blob/develop/src/main/resources/jupyter/jupyter-docker-compose.yaml#L34), [here](https://github.com/DataBiosphere/leonardo/blob/develop/src/main/resources/jupyter/jupyter_notebook_config.py#L28), and [here](https://github.com/DataBiosphere/leonardo/blob/develop/src/main/resources/jupyter/jupyter_notebook_config.py#L38) it's named `WELDER_ENABLE`. I think we need to make the variable names consistent.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
rtitle: I would still remove it -- it's confusing to have duplicated code. If we need the old version we have git history. I would say leaving it as a `def` going forward sounds fine though.

src/main/resources/jupyter/init-actions.sh
akarukappadath: thanks for catching that!

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
MatthewBemis: There has been some face-talking about this so I don't know where it landed but IMO this is added complexity for no benefit. AoU is going to recreate all clusters to use welder and from that point on they will only be creating welder enabled clusters. Once this is enabled in Terra, we will only be creating welderEnabled clusters across the entire platform and the flag will be taken out. AoU and Terra projects do not mix so there should be no scenario where this is a legitimate concern.

If there is a benefit and I am missing it (or if my assumptions above no longer hold true), please fill me in.

There was a discussion about it in the leo-data-syncing channel a few weeks ago and we never came to an agreement. See https://bvdp.slack.com/archives/CJMDJUJ06/p1560169892067400

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -163,6 +167,29 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
...
+            throw CannotEnableWelderException(googleProject, clusterName)
+          }
+        }
+      case _ =>
+        dbRef.inTransaction { dataAccess =>
+          dataAccess.clusterQuery.existsClustersWithWelderEnabled(googleProject)
+        } map { welderEnabledClustersExist =>
+          if (welderEnabledClustersExist) {
+            throw CannotDisableWelderException(googleProject, clusterName)
+          }
```
Qi77Qi: minor, I would do
```
    enableWelder match {
      case Some(true) =>
          dataAccess.clusterQuery.existsClustersWithWelderDisabled(googleProject)
        } flatMap { welderDisabledClustersExist =>
          if (welderDisabledClustersExist) {
            Future.failed(CannotEnableWelderException(googleProject, clusterName))
          } else Future.unit
        }
      case _ =>
          dataAccess.clusterQuery.existsClustersWithWelderEnabled(googleProject)
        } flatMap { welderEnabledClustersExist =>
          if (welderEnabledClustersExist) {
            Future.failed(CannotDisableWelderException(googleProject, clusterName))
          } else Future.unit
        }
    }
```

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
```diff
@@ -295,6 +303,24 @@ trait ClusterComponent extends LeoComponent {
         .result map { recs => unmarshalFullCluster(recs)}
     }
 
+    def existsClustersWithWelderEnabled(project: GoogleProject): DBIO[Boolean] = {
```
MatthewBemis: these two queries are no longer used

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -51,6 +51,12 @@ case class ClusterNotFoundException(googleProject: GoogleProject, clusterName: C
 case class ClusterAlreadyExistsException(googleProject: GoogleProject, clusterName: ClusterName, status: ClusterStatus)
   extends LeoException(s"Cluster ${googleProject.value}/${clusterName.value} already exists in ${status.toString} status", StatusCodes.Conflict)
 
+case class CannotDisableWelderException(googleProject: GoogleProject, clusterName: ClusterName)
```
MatthewBemis: looks like these two exceptions can be removed too

akarukappadath:
jenkins retest

src/main/resources/jupyter/jupyter_notebook_config.py
```diff
@@ -25,7 +25,7 @@
 # directory (which contains several other files on the image). We don't want
 # to mount there as it effectively deletes existing files on the image.
 # See https://docs.google.com/document/d/1b8wIydzC4D7Sbb6h2zWe-jCvoNq-bbD02CT1cnKLbGk
-if os.environ['WELDER_ENABLE'] == 'true':
+if os.environ['WELDER_ENABLED'] == 'true':
```
calbach: Oops - thanks.

rtitle:
Rebase this branch to get the fix for automation tests (https://github.com/DataBiosphere/leonardo/pull/973)

PR #971, opened 2019-06-24
Welder ext: handle deletes and errors

calbach:

In all cases:

### DIFF
--- a/docker/jupyter/custom/jupyter_delocalize.py
+++ b/docker/jupyter/custom/jupyter_delocalize.py
@@ -178,21 +178,30 @@ class WelderContentsManager(FileContentsManager):
     self.welder_base_url = 'http://127.0.0.1:8080'
     super(WelderContentsManager, self).__init__(*args, **kwargs)
 
-  def _welder_delocalize(self, path):
+  def _post_welder(self, action, path):
     # Ignore storage link failure, throw other errors.
     resp = requests.post(self.welder_base_url + '/objects', data=json.dumps({
-      'action': 'safeDelocalize',
+      'action': action,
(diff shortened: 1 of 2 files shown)


docker/jupyter/custom/jupyter_delocalize.py
```diff
@@ -178,21 +178,30 @@ def __init__(self, *args, **kwargs):
...
         pass
-      raise IOError("safeDelocalize failed: " + msg)
+
+      if resp.status_code == 412 and error_json.get('errorCode', -1) == 1:
+        # 1: Storage Link not found; expected for unmanaged files.
+        return
+
+      msg = resp.reason or 'unknown Welder error'
+      if error_json != {}:
+        msg = json.dumps(error_json)
```
Qi77Qi: just curious, do error messages get logged somewhere?

calbach: I see them getting logged by Tornado / Jupyter locally, but worth confirming once we're able to deploy this fully.

docker/jupyter/custom/jupyter_delocalize.py
```diff
@@ -231,5 +239,5 @@ def rename_file(self, old_path, new_path):
     super(WelderContentsManager, self).rename_file(old_path, new_path)
 
   def delete_file(self, path):
-    # TODO(IA-1049): Integrate Welder delete.
+    self._post_welder('delete', path)
```
Qi77Qi: if welder delete fails, will we still delete the local file?

calbach: No - this line with throw an exception. I have a test case to cover this situation.

PR #972, opened 2019-06-25
Add local docker-compose for Welder+Jupyter

calbach:
Proof-of-concept local development setup for Jupyter + Welder.

Usage:

```
# One-time setup
gcloud auth application-default login
chmod a+r ~/.config/gcloud/application_default_credentials.json # Not ideal, should find a better workaround

# Modify this script with your own GCS bucket (accessible to your ADC login)
cat docker/local/setup.sh

# Local welder changes
welder$ sbt server/docker:publishLocal

docker/local/run.sh

# Make changes to Jupyter extension.
docker/local/run.sh

# Open browser to http://localhost/notebooks/test-project/test-cluster
```

Limitations:
- Does not run the Leo init-actions script
- Currently uses the "latest" Welder label to pickup local builds, would be better to use an explicit local tag.
- Need to expand ACLs on ADC JSON file (can probably find another workaround for this).


Qi77Qi (approved):

calbach:
There are still some rough edges on this PR. @Qi77Qi up to you whether you want to merge this or drop it. As long as the branch isn't deleted I can continue to use this - so if no one else on the team would use this script then feel free to drop.

Qi77Qi:
@calbach  we're actually going to remove the whole docker directory pretty soon in favor of https://github.com/databiosphere/terra-docker, do you mind moving this over there? I don't mind merging it there even if it has rough edges as long as it's documented somewhere (basically move what you have in the description to say a readme file)

calbach:
Is the Leo-specific docker-compose setup moving as well? If not, I think it may more sense to live in the Leo repo.

Namely these files: https://github.com/DataBiosphere/leonardo/tree/develop/src/main/resources/jupyter

Qi77Qi:
@calbach no...I guess we can move this to top level then?

calbach:
Probably.. I don't really have bandwidth to mess with this PR at the moment. Feel free to drop it or take it over if you need it.

jdcanas:
This has been sitting for a while. I'm going to close it now since welder+jupyter have been pretty stable, and @calbach mentioned he does not have capacity for this. We can always reference this in the future.

PR #975, opened 2019-06-27
Export WELDER_ENABLED for docker-compose

calbach:
I tried out the new flag, and see that this variable is not being exported to the Jupyter image. I believe this is the issue.

In all cases:


rtitle (approved):
Ah good catch - yeah it needs to be exported

akarukappadath:
thank you for catching this!

PR #976, opened 2019-06-27
Apply basic linting to Leo extension JS files

calbach:
Used `eslint --fix` with default Google JS Style rules (with 4sp indent to minimize the diff). My editor was complaining about most lines when I opened this file, and semicolons are generally a good idea in JS.

Mostly this changed the indentation, added semicolons, changed quotes to `'`.

In all cases:

### DIFF
--- a/docker/jupyter/custom/.eslintrc.js
+++ b/docker/jupyter/custom/.eslintrc.js
@@ -1,0 +1,20 @@
+module.exports = {
+  'env': {
+    'browser': true,
+    'es6': true,
+  },
+  'extends': [
+    'google',
+  ],
+  'globals': {
+    'Atomics': 'readonly',
+    'SharedArrayBuffer': 'readonly',
+  },
+  'parserOptions': {
+    'ecmaVersion': 2018,
+    'sourceType': 'module',
+  },
+  'rules': {
+   "comma-dangle": ["error", "never"],  
+},
+};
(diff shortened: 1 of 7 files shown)


src/main/resources/jupyter/safe-mode.js
```diff
@@ -16,87 +16,87 @@ define(() => {
...
-    }
+        'Access-Control-Allow-Headers': '*',
+    };
 
     const basePayload = {
-        mode: "no-cors",
-        headers: headers
-    }
+        mode: 'no-cors',
+        headers: headers,
```
jdcanas: Trailing commas are not allowed in the JSON spec http://www.json.org/ (from the discussion here: https://stackoverflow.com/questions/201782/can-you-use-a-trailing-comma-in-a-json-object).

Apparently, some browsers will play nice but others won't, so I think it's best to remove it.

jdcanas:
jenkins retest

jdcanas (changes requested):
Looks good overall, though I'd prefer to get rid of the trailing commas.

Not sure why the tests failed, I triggered a retest.

src/main/resources/jupyter/safe-mode.js
calbach: I agree with you stylistically - updated.

To clarify though - JSON != Javascript. JSON is stricter than object definition in Javascript, though the naming might lead one to think they are identical. Trailing commas are valid in Javascript, though incidentally they might cause problems in very old browsers like IE8 (already not supported by AoU or Terra). The SO thread you link to is talking about JSON, not Javascript.

jdcanas:
jenkins retest

jdcanas:
closed for this https://github.com/DataBiosphere/leonardo/pull/1035, didn't want to deal with rebase

PR #978, opened 2019-06-28
enabling extensions

jdcanas:
Merging this now that the enableWelder flag is functioning. 

In all cases:

### DIFF
--- a/src/main/resources/jupyter/edit-mode.js
+++ b/src/main/resources/jupyter/edit-mode.js
@@ -77,9 +77,9 @@ define(() => {
     }
 
     function init() {
-        // console.info('edit mode plugin initialized')
-        // checkMeta()
-        // initSyncMaintainer()
+        console.info('edit mode plugin initialized')
+        checkMeta()
+        initSyncMaintainer()
     }
 
     function initSyncMaintainer() {
@@ -401,7 +401,5 @@ define(() => {
         location.reload(true)
     }
 
-    return {
-        load_ipython_extension: init
-    }
+    init()
 })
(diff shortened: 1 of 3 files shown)


rtitle (approved):

Qi77Qi (approved):
:shipit:

src/main/resources/jupyter/edit-mode.js
```diff
@@ -401,7 +401,5 @@ define(() => {
         location.reload(true)
     }
 
-    return {
-        load_ipython_extension: init
-    }
+    init()
```
calbach, line 404: Interesting  - can anyone explain how these files are executed? I tried to trace this in `init-actions.sh` but couldn't figure it out. These don't seem to have the normal Jupyter JS extension setup.

calbach (approved):

jmthibault79 (commented):
Looks OK but I have the same question as CH

src/main/resources/jupyter/edit-mode.js
jdcanas: ```extension_entry.js``` imports the extensions. Where it is invoked, I'm not sure. In ```init-actions.sh``` it indicates the fact it is renamed to custom.js is important, which tells me something downstream imports it, but I haven't had any luck finding what beyond that. I followed how ```google_sign_in.js``` is handled. @rtitle may know more

rtitle: Yeah https://github.com/DataBiosphere/leonardo/blob/develop/src/main/resources/jupyter/extension_entry.js is the entrypoint to these internal extensions. It is injected into Jupyter via the `custom.js` hook.

Note these are different than user-supplied extensions, which are installed by `jupyter nbextension install ...`.

PR #979, opened 2019-06-28
Parameterize Welder image, store used image in CLUSTER_IMAGE

MatthewBemis:
\<your comments for this PR go here\>

Have you read [CONTRIBUTING.md](https://github.com/DataBiosphere/leonardo/blob/develop/CONTRIBUTING.md) lately? If not, do that first.

I, the developer opening this PR, do solemnly pinky swear that:

- [ ] I've documented my API changes in Swagger

In all cases:

- [ ] Get a thumbsworth of review and PO signoff if necessary
- [ ] Verify all tests go green
- [ ] Squash and merge; you can delete your branch after this
- [ ] Test this change deployed correctly and works on dev environment after deployment

PR #980, opened 2019-06-28
Fix bug whereby enabling welder caused the Jupyer image to not run

rtitle:
@calbach hit this when he was testing welder. This is related to: https://github.com/DataBiosphere/leonardo/pull/979#discussion_r298772833

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
@@ -13,7 +13,7 @@ import enumeratum.{Enum, EnumEntry}
 import net.ceedubs.ficus.Ficus._
 import org.broadinstitute.dsde.workbench.leonardo.config.{ClusterDefaultsConfig, ClusterFilesConfig, ClusterResourcesConfig, DataprocConfig, ProxyConfig}
 import org.broadinstitute.dsde.workbench.leonardo.model.Cluster._
-import org.broadinstitute.dsde.workbench.leonardo.model.ClusterTool.{Jupyter, RStudio}
+import org.broadinstitute.dsde.workbench.leonardo.model.ClusterTool.{Jupyter, RStudio, Welder}
 import org.broadinstitute.dsde.workbench.leonardo.model.google.DataprocRole.SecondaryWorker
 import org.broadinstitute.dsde.workbench.leonardo.model.google.GoogleJsonSupport._
 import org.broadinstitute.dsde.workbench.leonardo.model.google._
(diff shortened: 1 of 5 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
```diff
@@ -277,7 +277,7 @@ object ClusterInitValues {
       clusterImages.find(_.tool == Jupyter).map(_.dockerImage).getOrElse(""),
       clusterImages.find(_.tool == RStudio).map(_.dockerImage).getOrElse(""),
       proxyConfig.jupyterProxyDockerImage,
-      dataprocConfig.welderDockerImage,
+      clusterImages.find(_.tool == Welder).map(_.dockerImage).getOrElse(""),
```
rtitle, line 280: I think this was also missed in https://github.com/DataBiosphere/leonardo/pull/979 -- the user-specified welder image needs to be templated to `init-actions.sh`.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -1111,25 +1111,19 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
     //We will use the client-supplied image, if present, otherwise we will use a default.
     //If welder is not enabled, we won't use any image.
     //Eventually welder will be enabled for all clusters and this will be way cleaner.
-    val welderImage = if(clusterRequest.enableWelder.getOrElse(false)) {
+    val welderImageOpt: Option[ClusterImage] = if(clusterRequest.enableWelder.getOrElse(false)) {
       val i = clusterRequest.welderDockerImage.getOrElse(dataprocConfig.welderDockerImage)
       Some(ClusterImage(Welder, i, now))
     } else None
 
-    val images = Set(
```
rtitle, line 1119: This is the meat of the fix. If `enableWelder` is set to true, and the user doesn't provide a Jupyter image, this method will return a `Set(welderImage)` and not install Jupyter!

src/test/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoServiceSpec.scala
```diff
@@ -208,7 +209,10 @@ class LeonardoServiceSpec extends TestKit(ActorSystem("leonardotest")) with Flat
     eventually {
       val createdCluster = leo.getActiveClusterDetails(userInfo, project, name1).futureValue
 
-      createdCluster.clusterImages.map(_.dockerImage) should contain (dataprocConfig.welderDockerImage)
+      // cluster images should contain welder and Jupyter
```
rtitle, line 212: Updated tests to verify presence of the Jupyter image. I verified that these tests _failed_ without the fix in `LeonardoService`.

rtitle:
I tested in a fiab and I think this is working correctly. I created a cluster with:

a) enableWelder=false
b) enableWedler=true
c) enableWelder=true, welderDockerImage=some-custom-image

In all cases the correct images were in the Leo DB and on the cluster.

Qi77Qi (approved):

MatthewBemis:
Ugh. Thanks for fixing this.

PR #981, opened 2019-07-01
Test fix to call more efficient 'list cluster'

rtitle:
See https://broadworkbench.atlassian.net/browse/QA-687

The tests were calling the global `/api/clusters?includeDeleted=true` while monitoring clusters. When run in alpha, the response hit the akka-http 8MB limit and failed the test.

Changed it to call list by project (which is what the app calls): `/api/clusters/{googleProject}?includeDeleted=true`.

In all cases:

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/ClusterFixtureSpec.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/ClusterFixtureSpec.scala
@@ -10,6 +10,7 @@ import org.broadinstitute.dsde.workbench.util.addJitter
 import org.scalatest.{BeforeAndAfterAll, Outcome, fixture}
 
 import scala.concurrent.duration._
+import scala.language.postfixOps
 import scala.util.{Failure, Success, Try}
 
 
--- a/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/ClusterFixtureSpec.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/ClusterFixtureSpec.scala
@@ -10,7 +10,6 @@ import org.broadinstitute.dsde.workbench.util.addJitter
 import org.scalatest.{BeforeAndAfterAll, Outcome, fixture}
 
 import scala.concurrent.duration._
-import scala.language.postfixOps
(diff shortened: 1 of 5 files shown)


automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/ClusterFixtureSpec.scala
```diff
@@ -10,6 +10,7 @@ import org.broadinstitute.dsde.workbench.util.addJitter
 import org.scalatest.{BeforeAndAfterAll, Outcome, fixture}
 
 import scala.concurrent.duration._
+import scala.language.postfixOps
```
rtitle: I was sick of compiler warnings so added this in a couple places

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/Leonardo.scala
```diff
@@ -112,14 +112,14 @@ object Leonardo extends RestClient with LazyLogging {
       s"api/cluster${version.toUrlSegment}/${googleProject.value}/${clusterName.string}"
     }
 
-    def list()(implicit token: AuthToken): Seq[Cluster] = {
```
rtitle, line 115: I removed the old `list()` and `listIncludingDeleted()` because I think we should always be calling the per-project version.

MatthewBemis (approved):
Great timing, reviewing this from QA triage 👍

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/ClusterFixtureSpec.scala
Qi77Qi: I think you can just add https://github.com/DataBiosphere/leonardo/blob/develop/project/Settings.scala#L72 to automation's setting.scala file

Qi77Qi (approved):

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/ClusterFixtureSpec.scala
rtitle: Ah cool

PR #983, opened 2019-07-01
Retry for Cluster Creation

akarukappadath:
Added retry to cluster creation in the case of a rateLimitExceeded Google Exception. Also added new relic metrics to note whenever we receive this exception from Google. Moved new relic object so that it doesn't need to be passed around from Boot. 

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
@@ -6,7 +6,7 @@ import java.time.Instant
 import akka.actor.ActorSystem
 import akka.http.scaladsl.model.StatusCodes
 import cats.Monoid
-import cats.data.{Ior, NonEmptyList, OptionT}
+import cats.data.{Ior, OptionT}
 import cats.implicits._
 import com.google.api.client.googleapis.json.GoogleJsonResponseException
 import com.google.api.client.http.HttpResponseException
@@ -728,7 +728,7 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
         gdDAO.createCluster(googleProject, clusterName, createClusterConfig)
       }
       operation <- retryResult match {
-        case Right((List.empty, op)) => Future.successful(op)
+        case Right((errors, op)) if errors == List.empty => Future.successful(op)
         case Right((errors, op)) =>
           Metrics.newRelic.incrementCounterIO("zoneCapacityClusterCreationFailure", errors.length).unsafeRunAsync(_ => ())
           Future.successful(op)
--- a/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
@@ -28,7 +28,6 @@ import org.broadinstitute.dsde.workbench.leonardo.model.google._
 import org.broadinstitute.dsde.workbench.leonardo.util.BucketHelper
 import org.broadinstitute.dsde.workbench.model.google._
 import org.broadinstitute.dsde.workbench.model.{ErrorReport, UserInfo, WorkbenchEmail}
-import org.broadinstitute.dsde.workbench.newrelic.NewRelicMetrics
 import org.broadinstitute.dsde.workbench.util.Retry
 import slick.dbio.DBIO
 import spray.json._
@@ -109,8 +108,7 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
                       protected val authProvider: LeoAuthProvider,
                       protected val serviceAccountProvider: ServiceAccountProvider,
                       protected val bucketHelper: BucketHelper,
-                      protected val contentSecurityPolicy: String,
-                      protected val newRelic: NewRelicMetrics)
+                      protected val contentSecurityPolicy: String)
                      (implicit val executionContext: ExecutionContext,
                       implicit override val system: ActorSystem) extends LazyLogging with Retry {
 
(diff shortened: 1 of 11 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -735,6 +745,13 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
     }
   }
 
+  private def whenGoogleZoneCapacityIssue(throwable: Throwable): Boolean = {
+    throwable match {
+      case t: GoogleJsonResponseException => t.getStatusCode == 429 && t.getDetails.getErrors.asScala.head.getDomain.equalsIgnoreCase("rateLimitExceeded")
```
MatthewBemis: is "rateLimitExceeded" as specific as it gets in the error message? my minor worry is that 429 with "rateLimitExceeded" isn't unique to just google zone issues and we may end up retrying google failures when we shouldn't be. i do acknowledge that google error messages are usually bad and there might not be a better alternative here...

it also seems weird that they're returning a rate limit error when they're the ones out of compute. we're not exceeding any rate limit AFAICT.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -723,9 +724,18 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
 
       // Create the cluster
       createClusterConfig = CreateClusterConfig(machineConfig, initScript, serviceAccountInfo.clusterServiceAccount, credentialsFileName, stagingBucket, clusterScopes, clusterVPCSettings, clusterRequest.properties)
-      operation <- gdDAO.createCluster(googleProject, clusterName, createClusterConfig)
+      retryResult <- retryExponentially(whenGoogleZoneCapacityIssue, "Cluster creation failed because zone with adequate resources was not found") { () =>
+        gdDAO.createCluster(googleProject, clusterName, createClusterConfig)
+      }
+      operation <- retryResult match {
+        case Right(value) => Future.successful(value._2)
```
MatthewBemis: readability nit: you can do `case Right((_, op)) => Future.successful(op)` here to name the value in the tuple that you're using

akarukappadath:
jenkins retest

MatthewBemis (approved):
Left a couple of comments, none are blocking

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
rtitle: There are 3 possible cases here:
1. The request succeeds the first time, in which case we'll have a `Right((List.empty, result))`
2. The request gets retried but eventually succeeds. In which we'll have a `Right((errors, result))`.
3. The request gets retried and never succeeds. In which case we'll have a `Left(errors)`.

We should probably increment the counter for cases 2 and 3, but the current code is only incrementing for case 3.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -723,9 +724,18 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
...
       // Create the cluster
       createClusterConfig = CreateClusterConfig(machineConfig, initScript, serviceAccountInfo.clusterServiceAccount, credentialsFileName, stagingBucket, clusterScopes, clusterVPCSettings, clusterRequest.properties)
-      operation <- gdDAO.createCluster(googleProject, clusterName, createClusterConfig)
+      retryResult <- retryExponentially(whenGoogleZoneCapacityIssue, "Cluster creation failed because zone with adequate resources was not found") { () =>
+        gdDAO.createCluster(googleProject, clusterName, createClusterConfig)
+      }
+      operation <- retryResult match {
+        case Right((_, op)) => Future.successful(op)
+        case Left(errors) =>
+          Metrics.newRelic.incrementCounterIO("zoneCapacityClusterCreationFailure").unsafeRunAsync(_ => ())
```
rtitle: (my comment went away, trying again)

There are 3 possible cases here:

1. The request succeeds the first time, in which case we'll have a Right((List.empty, result))
2. The request gets retried but eventually succeeds. In which we'll have a Right((errors, result)).
3. The request gets retried and never succeeds. In which case we'll have a Left(errors).

We should probably increment the counter for cases 2 and 3, but the current code is only incrementing for case 3.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -723,9 +724,18 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
...
+          val exceptionMessage = errors.toList.map(_.toString).mkString(", ")
```
rtitle: Is it useful to return _all_ the messages? Probably just the last one is fine.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
akarukappadath: Does it matter then that a cluster that fails once with this error and then succeeds has the same impact on the counter as a cluster that gets recreated and fails multiple times with this error?

rtitle: That's a good question. I'm inclined to increment the counter anytime this error occurs, even if it's part of a retry. We should add separate instrumentation for Error'd clusters, broken out by cause.

akarukappadath:
jenkins retest

rtitle (approved):
Latest changes look 👍 to me

PR #984, opened 2019-07-02
Minimal cluster query for clearing kernelFoundBusyDate

akarukappadath:
Small PR to use a minimal cluster query to avoid unnecessary left joins. 

In all cases:


src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
```diff
@@ -362,7 +362,7 @@ trait ClusterComponent extends LeoComponent {
     }
 
     def clearKernelFoundBusyDateByProjectAndName(googleProject: GoogleProject, clusterName: ClusterName): DBIO[Int] = {
-      clusterQuery.getActiveClusterByName(googleProject, clusterName) flatMap {
+      clusterQuery.getActiveClusterByNameMinimal(googleProject, clusterName) flatMap {
```
rtitle, line 365: While we're in here can you also update `updateDateAccessedByProjectAndName`?

rtitle (approved):

akarukappadath:
jenkins retest

PR #985, opened 2019-07-02
Optimize fullClusterQuery in some areas

MatthewBemis:

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
@@ -194,14 +194,14 @@ trait ClusterComponent extends LeoComponent {
     // find* and get* methods do query the INSTANCE table
 
     def getActiveClusterByName(project: GoogleProject, name: ClusterName): DBIO[Option[Cluster]] = {
-      fullClusterQueryByUniqueKey(project, name)
+      fullClusterQueryByUniqueKey(project, name, Some(dummyDate))
         .result map { recs =>
           unmarshalFullCluster(recs).headOption
         }
     }
 
     def getDeletingClusterByName(project: GoogleProject, name: ClusterName): DBIO[Option[Cluster]] = {
-      fullClusterQueryByUniqueKey(project, name)
+      fullClusterQueryByUniqueKey(project, name, Some(dummyDate))
         .filter { _._1.status === ClusterStatus.Deleting.toString }
         .result map { recs =>
           unmarshalFullCluster(recs).headOption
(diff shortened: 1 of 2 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
```diff
@@ -194,19 +194,14 @@ trait ClusterComponent extends LeoComponent {
...
         .result map { recs =>
           unmarshalFullCluster(recs).headOption
         }
     }
 
     def getDeletingClusterByName(project: GoogleProject, name: ClusterName): DBIO[Option[Cluster]] = {
-      fullClusterQuery
-        .filter { _._1.googleProject === project.value }
-        .filter { _._1.clusterName === name.value }
+      fullClusterQueryByUniqueKey(project, name)
```
MatthewBemis: still deciding if this is right to use here

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
```diff
@@ -577,16 +574,21 @@ trait ClusterComponent extends LeoComponent {
...
     val destroyedDate = destroyedDateOpt.getOrElse(dummyDate)
 
-    fullClusterQuery
-      .filter { _._1.googleProject === googleProject.value }
-      .filter { _._1.clusterName === clusterName.value }
-      .filter { _._1.destroyedDate === Timestamp.from(destroyedDate) }
-      .result
+    for {
+      ((((((cluster, instance), error), label), extension), image), scopes) <-
+        clusterQuery filter (_.googleProject === googleProject.value) filter (_.clusterName === clusterName.value) filter (_.destroyedDate === Timestamp.from(destroyedDate)) joinLeft
```
MatthewBemis, line 584: as far as i can tell, the key change here is applying the filters on the initial `clusterQuery` rather than on the entire result of the `fullClusterQuery`. doing it afterwards resulted in large derived tables which slowed down the query considerably. the sql that this generates is equally horrific but the explain plan is much better

MatthewBemis: Before:
Screen Shot 2019-07-02 at 11 13 07 AM

After:
Screen Shot 2019-07-02 at 11 12 47 AM

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
MatthewBemis: ok this appears to be correct. anything in `Deleting` state doesn't have a `destroyedDate` yet so it's okay if we use the default dummy date.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
```diff
@@ -194,19 +194,14 @@ trait ClusterComponent extends LeoComponent {
...
           unmarshalFullCluster(recs).headOption
         }
     }
 
     def getDeletingClusterByName(project: GoogleProject, name: ClusterName): DBIO[Option[Cluster]] = {
-      fullClusterQuery
-        .filter { _._1.googleProject === project.value }
-        .filter { _._1.clusterName === name.value }
+      fullClusterQueryByUniqueKey(project, name, Some(dummyDate))
         .filter { _._1.status === ClusterStatus.Deleting.toString }
```
MatthewBemis, line 205: this does a filter after the fact....but the data has already been pared down so it should be fine

MatthewBemis: (this also wasn't the target of this perf improvement but it should still benefit from the change)

rtitle (approved):
Looks good. That's definitely counter-intuitive of slick. But I'm glad we can make it do what we want relatively easily -- I think this should have a big impact.

Another place to potentially optimize is "list clusters by project" (which the UI calls pretty often). Looking at the code it does a single left join (to labels) then a series of filters. I wonder if it would perform better if doing the filtering first. 

I'm inclined to merge this then re-run performance tests, and see how they look. Based on results we could then decide whether to optimize "list clusters".

MatthewBemis:
@rtitle agreed, I will merge this now and then we can evaluate other opportunities based on test results. 

Upon quick inspection, the query that Slick generates for listing clusters by project doesn't seem unreasonable, and it's hitting the indexes. Maybe Slick only does terrible things for queries that are extremely complex?

PR #987, opened 2019-07-03
Change ZombieClusterMonitor to mark clusters Deleted instead of Error

rtitle:
This came from a conversation with @bradtaylor, @tmm211, and @sharmatime.

Leo has functionality to detect clusters that are active in Leo (meaning running or paused) but do not exist in Google. We call these "zombie" clusters. Our current behavior when we detect a zombie cluster is to set it to `Error` status and persist a specific error message.

However now that I'm thinking about it this situation is not really an error. It might be better to just set the clusters to `Deleted` in Leo to reconcile the state between Google <--> Leo.

This came up because we've run a cleanup script in the past to delete Google clusters that were mistakenly autogenerated by Terra. Now users are seeing error messages for old deleted clusters. See JIRA: https://broadworkbench.atlassian.net/browse/SUP-20. As part of this change I'd also like to run a DB migration to delete old clusters in this zombie state.

Let me know what you think.

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ZombieClusterMonitor.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ZombieClusterMonitor.scala
@@ -7,6 +7,7 @@ import akka.actor.{Actor, Props, Timers}
 import cats.implicits._
 import com.typesafe.scalalogging.LazyLogging
 import org.broadinstitute.dsde.workbench.google.GoogleProjectDAO
+import org.broadinstitute.dsde.workbench.leonardo.Metrics
 import org.broadinstitute.dsde.workbench.leonardo.config.ZombieClusterConfig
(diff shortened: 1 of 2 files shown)


kyuksel:
Semantically, a cluster being deleted without Leonardo's knowledge sounds like an error state to me (at least from Leonardo's perspective) but if this provides a pragmatic solution to the issue the user reported, I understand that. 

With the change, the way to differentiate a cluster deleted through Leonardo vs. not through Leonardo, would be to check the corresponding error record, if any?

rtitle:
Yes, clusters in this state would have an error record. This PR also adds a NewRelic metric for clusters that have been deleted this way.

I agree it seems like an Error state, but from the user's perspective the cluster is deleted. 

Going to rebase this PR and re-run tests.

Qi77Qi:
just to double check, we'll be able to tell if a cluster is deleted from unexpected error vs normal delete right?

Qi77Qi (approved):

rtitle:
@Qi77Qi yes via the CLUSTER_ERROR table in the database

PR #989, opened 2019-07-03
IA-1107 flush welder cache before stopping cluster

Qi77Qi:
Tested with in fiab.
Note flush is `/cache/flush` is called before the server restart
```
2019-07-05 20:56:41,054 INFO [ioapp-compute-2] o.h.s.m.Logger [Logger.scala:29] HTTP/1.1 POST /cache/flush Headers(Host: 127.0.0.1:8080, User-Agent: akka-http/10.1.1, X-Forwarded-For: 35.188.148.107, X-Forwarded-Host: fd60a817-0e4d-48c8-b12f-3941333fb016.jupyter-dev.firecloud.org, X-Forwarded-Server: automation-test-akyaymuuz-m.c.gpalloc-dev-master-7d9mobu.internal, Connection: close, Content-Length: 0) body=""
2019-07-05 20:56:41,057 INFO [ioapp-compute-7] o.h.s.m.Logger [Logger.scala:29] HTTP/1.1 204 No Content Headers() body=""
2019-07-05 21:02:35,698 INFO [ioapp-compute-4] o.h.b.c.n.NIO1SocketServerGroup [NIO1SocketServerGroup.scala:239] Service bound to address /0:0:0:0:0:0:0:0:8080
2019-07-05 21:02:35,715 INFO [ioapp-compute-3] o.h.s.b.BlazeServerBuilder [BlazeServerBuilder.scala:339]
  _   _   _        _ _
 | |_| |_| |_ _ __| | | ___
 | ' \  _|  _| '_ \_  _(_-<
 |_||_\__|\__| .__/ |_|/__/
             |_|
```

In all cases:


src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -92,6 +93,7 @@ case class InvalidDataprocMachineConfigException(errorMsg: String)
   extends LeoException(s"${errorMsg}", StatusCodes.BadRequest)
 
 class LeonardoService(protected val dataprocConfig: DataprocConfig,
+                      protected val welderDao: WelderDAO,
```
rtitle, line 96: I think there are lots of compilation errors due to this new field (I'm looking at travis results).

Qi77Qi: ah yea, I haven't fix tests yet

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -590,6 +592,7 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
                gdDAO.resizeCluster(cluster.googleProject, cluster.clusterName, numPreemptibles = Some(0))
              else Future.unit
 
+        _ <- welderDao.flushCache(cluster.googleProject, cluster.clusterName).handleError(e => logger.error(s"fail to flush welder cache for ${cluster}"))
```
rtitle: Does the flush API call block until the cache is written to disk? Just thinking about race conditions: we don't want to stop the instance before the cache is actually flushed.

rtitle (changes requested):
Requesting changes to fix compilation errors. Also would be good to update the Leo `stopCluster` tests to verify that welder is called.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
Qi77Qi: yep. it blocks

Qi77Qi:
@rtitle  fixed tests...For stopCluster unit test, I don't think the tests are set up in a way I can easily mock out the leo service being used. I'll verify manually in fiab.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
rtitle: Also: should we only do this if welder is enabled?

rtitle (approved):
Looks good

Qi77Qi:
@rtitle thanks for review!

PR #1000, opened 2019-07-12
Support renames with Welder

calbach:
https://broadworkbench.atlassian.net/browse/IA-1028

In all cases:

### DIFF
--- a/docker/jupyter/custom/jupyter_delocalize.py
+++ b/docker/jupyter/custom/jupyter_delocalize.py
@@ -178,6 +178,27 @@ class WelderContentsManager(FileContentsManager):
     self.welder_base_url = 'http://127.0.0.1:8080'
     super(WelderContentsManager, self).__init__(*args, **kwargs)
 
+  def _extract_welder_error(self, resp):
+      try:
+        return json.dumps(resp.json())
+      except:
+        return resp.reason or 'unknown Welder error'
+
+  def _check_welder_edit_mode(self, path):
+    resp = requests.post(self.welder_base_url + '/objects/metadata', data=json.dumps({
+      # Sometimes the Jupyter UI provided "path" contains a leading /, sometimes
+      # not; strip for Welder.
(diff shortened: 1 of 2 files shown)


docker/jupyter/custom/jupyter_delocalize.py
```diff
@@ -239,8 +256,13 @@ def save(self, model, path=''):
...
   def rename_file(self, old_path, new_path):
-    # TODO(IA-1028): Integrate Welder renaming.
-    super(WelderContentsManager, self).rename_file(old_path, new_path)
+    if not self._check_welder_edit_mode(old_path) and not self._check_welder_edit_mode(new_path):
+      # If we're not touching any edit mode files, just do a normal move.
+      return super(WelderContentsManager, self).rename_file(old_path, new_path)
+
+    # These methods already properly handle edit mode semantics.
+    self.save(self.get(old_path), new_path)
+    self.delete(old_path)
```
Qi77Qi: I think we'll need to revert the save if `delete` fails

calbach: Good catch. These two-phase actions make me a bit nervous - but we'll need to shift this logic into Welder to fix it.

calbach: (added)

Qi77Qi: I feel like a rename api in welder will be tricky to implement as well. How about welder provides /checkLock api?

calbach: `/checkLock` would not help here.

Actually on further research of the GCS APIs - looks like move is not a real operation: https://cloud.google.com/storage/docs/renaming-copying-moving-objects#storage-move-object-java  so there is not as much of an advantage as I'd thought to moving this into Welder - it would still need to do a two-phase approach.

Qi77Qi (approved):

PR #1003, opened 2019-07-12
Move notebooks directory to config

akarukappadath:

In all cases:

### DIFF
--- a/docker/jupyter/scripts/run-jupyter.sh
+++ b/docker/jupyter/scripts/run-jupyter.sh
@@ -5,10 +5,15 @@ set -e
 # Change the default umask to support R/W access to the shared volume with welder
 umask 002
 
-# Forces python 3
-JUPYTER_BASE="/usr/local/bin/python3 /usr/local/bin/jupyter-notebook"
-JUPYTER_CMD="$JUPYTER_BASE --NotebookApp.nbserver_extensions=\"{'jupyter_localize_extension':True}\" &> ${HOME}/jupyter.log"
+if [ -n "$1" ]; then
+  NOTEBOOKS_DIR=$1
 
-echo $JUPYTER_CMD
+  # Forces python 3
+  JUPYTER_BASE="/usr/local/bin/python3 /usr/local/bin/jupyter-notebook"
+  JUPYTER_CMD="$JUPYTER_BASE --NotebookApp.nbserver_extensions=\"{'jupyter_localize_extension':True}\" &> ${NOTEBOOKS_DIR}"
 
-eval $JUPYTER_CMD
+  echo $JUPYTER_CMD
+
+  eval $JUPYTER_CMD
+
+fi
--- a/docker/jupyter/scripts/run-jupyter.sh
+++ b/docker/jupyter/scripts/run-jupyter.sh
@@ -5,7 +5,7 @@ set -e
 # Change the default umask to support R/W access to the shared volume with welder
 umask 002
 
-NOTEBOOKS_DIR=$1
+NOTEBOOKS_DIR=${1:-${HOME}/notebooks}
 
 # Forces python 3
 JUPYTER_BASE="/usr/local/bin/python3 /usr/local/bin/jupyter-notebook"
--- a/docker/jupyter/scripts/run-jupyter.sh
+++ b/docker/jupyter/scripts/run-jupyter.sh
@@ -5,15 +5,12 @@ set -e
 # Change the default umask to support R/W access to the shared volume with welder
 umask 002
 
-if [ -n "$1" ]; then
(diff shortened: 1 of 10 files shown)


docker/jupyter/scripts/run-jupyter.sh
```diff
@@ -5,9 +5,11 @@ set -e
 # Change the default umask to support R/W access to the shared volume with welder
 umask 002
 
+NOTEBOOKS_DIR=$1
```
rtitle: [This](https://github.com/DataBiosphere/leonardo/blob/develop/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala#L134) also needs to be updated if `run-jupyter.sh` now takes an arg.

rtitle: Also it might be nice if this had a default:
```
NOTEBOOKS_DIR=${1:-${HOME}/notebooks}
```
What do you think?

src/main/resources/jupyter/jupyter-docker-compose.yaml
```diff
@@ -10,7 +10,7 @@ services:
     network_mode: host
     volumes:
       # shared with welder
-      - /work:/home/jupyter-user/notebooks
+      - /work:$(notebooksDir)
```
rtitle: This can be the EV that was exported by `init-actions.sh`: `${NOTEBOOKS_DIR}`.

This would also let you avoid needing to template `jupyter-docker-compose.yaml`.

src/main/resources/jupyter/jupyter_notebook_config.py
```diff
@@ -28,7 +28,7 @@
 if os.environ['WELDER_ENABLED'] == 'true':
   # Only enable this along with Welder, as this change is backwards incompatible
   # for older localization users.
-  c.NotebookApp.notebook_dir = '/home/jupyter-user/notebooks'
+  c.NotebookApp.notebook_dir = $(notebooksDir)
```
rtitle: This could also be the EV: `${NOTEBOOKS_DIR}`. 

Doesn't really matter in this case because this file is already templated. But it might be useful anyway to set this EV inside the Jupyter container.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -1006,6 +1005,7 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
     val editModeJsContent = templateResource(clusterResourcesConfig.editModeJs, replacements)
     val safeModeJsContent = templateResource(clusterResourcesConfig.safeModeJs, replacements)
     val jupyterNotebookConfigContent = templateResource(clusterResourcesConfig.jupyterNotebookConfigUri, replacements)
+    val jupyterDockerComposeContent = templateResourceWithNoQuotes(clusterResourcesConfig.jupyterDockerCompose, replacements)
```
rtitle: Again this is probably not needed if you just use the EV in `jupyter-docker-compose.yaml`

rtitle (changes requested):
A few comments above.

src/main/resources/jupyter/jupyter_notebook_config.py
```diff
@@ -28,7 +28,7 @@
 if os.environ['WELDER_ENABLED'] == 'true':
   # Only enable this along with Welder, as this change is backwards incompatible
   # for older localization users.
-  c.NotebookApp.notebook_dir = '/home/jupyter-user/notebooks'
+  c.NotebookApp.notebook_dir = ${NOTEBOOKS_DIR}
```
rtitle: In order to use it here I think `NOTEBOOKS_DIR` needs to be in the `environment` section of `jupyter-docker-compose.yaml`'. Something like:
```
NOTEBOOKS_DIR: "${NOTEBOOKS_DIR}"
```

rtitle: Also this may need to be
```
os.environ['NOTEBOOKS_DIR']
```
similar to the above evs.

rtitle (changes requested):
A few more comments, almost there though :)

rtitle (approved):

rtitle:
Hm hold on... this sets the notebooks dir to `/home/jupyter-user/notebooks`, which is different from right now. Do we want to _only_ do this if welder is enabled for a cluster?

akarukappadath:
jenkins retest

akarukappadath:
jenkins retest

akarukappadath:
jenkins retest

rtitle (approved):

PR #1007, opened 2019-07-17
Add metrics for welder enabled cluster requests

akarukappadath:

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
@@ -197,8 +197,8 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
         val clusterImages = processClusterImages(clusterRequest)
         val clusterFuture = for {
           // Metrics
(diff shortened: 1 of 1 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -196,6 +196,10 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
         val augmentedClusterRequest = augmentClusterRequest(serviceAccountInfo, googleProject, clusterName, userEmail, clusterRequest)
         val clusterImages = processClusterImages(clusterRequest)
         val clusterFuture = for {
+          // Metrics
+          _ <- Metrics.newRelic.incrementCounterFuture("clusterRequests")
```
Qi77Qi: Shall we rename this to something like `numberOfCreateClusterRequest`. It's not obvious what `clusterRequests` means by itself.

Same comment for the other metrics, something like ``numberOfCreateClusterRequestWithWelder`?

Qi77Qi (approved):

PR #1009, opened 2019-07-18
IA-1157 - Use UUIDs for Sam resource IDs

MatthewBemis:
I'd like to touch a few things up. The general approach will stay the same, though.

In all cases:


src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20190715_add_internal_resource_id.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="leonardo" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+    <changeSet logicalFilePath="leonardo" author="mbemis" id="add-cluster-internal-resource-id">
+        <addColumn tableName="CLUSTER">
+            <column name="internalId" type="VARCHAR(254)">
```
MatthewBemis, line 5: i would have liked to enforce that this is unique at the database-level but given that the IDs generated in the below sql command won't be unique, that's a no-go without a more thorough migration

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/sam/SwaggerSamClient.scala
```diff
@@ -115,18 +115,18 @@ class SwaggerSamClient(samBasePath: String, cacheEnabled: Boolean, cacheExpiryTi
     credential.refreshAccessToken.getTokenValue
   }
 
-  def createNotebookClusterResource(userEmail: WorkbenchEmail, googleProject: GoogleProject, clusterName: ClusterName) = {
+  def createNotebookClusterResource(internalId: String, userEmail: WorkbenchEmail, googleProject: GoogleProject, clusterName: ClusterName) = {
     logger.debug(s"Calling Sam createNotebookClusterResource as PET for user [${userEmail.value}] with project [${googleProject.value}] and clusterName [${clusterName.value}]")
```
MatthewBemis, line 120: these log messages are only `debug` so i think i'll take out the unused project and cluster name params and just log the resource ID here instead

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/IamProxyAuthProvider.scala
```diff
@@ -120,24 +120,26 @@ class IamProxyAuthProvider(config: Config, serviceAccountProvider: ServiceAccoun
...
     * Leo will wait, so be timely!
     *
+    * @param internalId     The internal ID for the cluster (i.e. used for Sam resources)
     * @param creatorEmail     The email address of the user in question
     * @param googleProject The Google project the cluster was created in
     * @param clusterName   The user-provided name of the Dataproc cluster
     * @return A Future that will complete when the auth provider has finished doing its business.
     */
-  def notifyClusterCreated(creatorEmail: WorkbenchEmail, googleProject: GoogleProject, clusterName: ClusterName)(implicit executionContext: ExecutionContext): Future[Unit] = Future.successful(())
+  def notifyClusterCreated(internalId: String, creatorEmail: WorkbenchEmail, googleProject: GoogleProject, clusterName: ClusterName)(implicit executionContext: ExecutionContext): Future[Unit] = Future.successful(())
```
Qi77Qi: if this is supposed to be UUID, I would use `java.util.UUID` for its type

MatthewBemis: would love to, but that would require migrating existing clusters to have UUIDs as their sam resource IDs which is non-trivial

see https://github.com/DataBiosphere/leonardo/pull/1009/files#diff-b924c27cbfe0ca9ea81a34fca450e6f3R92

MatthewBemis: but a new type, like `InternalId` might make sense here

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/sam/SwaggerSamClient.scala
MatthewBemis: i changed my mind and i'm keeping it as-is

MatthewBemis:
jenkins retest

the failures are a bit suspect but i just want another run to confirm

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
```diff
@@ -17,6 +17,7 @@ import org.broadinstitute.dsde.workbench.model.WorkbenchEmail
 import org.broadinstitute.dsde.workbench.model.google.{GcsBucketName, GcsPath, GcsPathSupport, GoogleProject, ServiceAccountKey, ServiceAccountKeyId, parseGcsPath}
 
 final case class ClusterRecord(id: Long,
+                               internalId: String,
```
MatthewBemis, line 20: i'm not really happy with this. i would have liked for there to no longer be an `id` field and to just have `internalId` with type `UUID` (although in the DB it would actually be a `VARCHAR(36)` because querying binary fields is horrible).

in order to accomplish that, i'd need to:

1. remove ID field
2. generate UUIDs for every existing cluster in the database and make this the primary key
3. for all existing _active_ clusters, generate a new `notebook-cluster` resource in sam. migration for this wouldn't actually be _too_ hairy, i could just use the rawls or agora SAs since they have permission to get access tokens on behalf of users. an alternative would be to insert the records directly into openDJ but i don't actually think that's a good idea going forward. note that all of this works because of the assumption that no user has modified their cluster's policies in sam

since this will fix the problems as-is, i'm not going to go down the rabbit-hole but maybe when i have some free time in the future i will.

MatthewBemis:
I think the failures on this are legitimate since they're reproducible. Digging in...

src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20190715_add_internal_resource_id.xml
rtitle: Yeah, I agree with this reasoning.

src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20190715_add_internal_resource_id.xml
```diff
@@ -0,0 +1,18 @@
...
+                <constraints nullable="true"/>
+            </column>
+        </addColumn>
+    </changeSet>
+    <changeSet logicalFilePath="leonardo" author="mbemis" id="backpopulate-internal-resource-id">
+        <sql stripComments="true">
+            update CLUSTER c set internalId = CONCAT(c.googleProject, '_', c.clusterName);
+        </sql>
+    </changeSet>
+    <changeSet logicalFilePath="leonardo" author="mbemis" id="internal-id-non-nullable">
```
rtitle, line 15: Even though it's non-unique I was wondering if we should add an index for `internalId` anyway. However we're never actually using it in a `WHERE` clause (and I don't see why we would), so I think it's ok without one.

rtitle (approved):
Sorry for the late review. This all looks good. Thanks for the description of alternative approaches -- I agree this is the best pragmatic approach. 👍

rtitle:
jenkins retest

rtitle:
fiab-start failed, `no space left on device` :(

jenkins retest, rolling the dice again

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/IamProxyAuthProvider.scala
rtitle: I ended up making a new type `ClusterInternalId`

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/sam/SamAuthProvider.scala
```diff
@@ -46,9 +46,9 @@ class SamAuthProvider(val config: Config, serviceAccountProvider: ServiceAccount
       new CacheLoader[SamCacheKey, Future[Boolean]] {
         def load(key: SamCacheKey) = {
           key match {
-            case NotebookAuthCacheKey(userInfo, action, googleProject, clusterName, executionContext) =>
+            case NotebookAuthCacheKey(internalId, userInfo, action, googleProject, clusterName, executionContext) =>
               // tokenExpiresIn should not taken into account when comparing cache keys
-              hasNotebookClusterPermissionInternal(userInfo.copy(tokenExpiresIn = 0), action, googleProject, clusterName)(executionContext)
+              hasNotebookClusterPermissionInternal(internalId, userInfo.copy(tokenExpiresIn = 0), action, googleProject, clusterName)(executionContext)
```
Qi77Qi, line 51: wonder why the executionContext needs to be passed expliclity

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
```diff
@@ -86,9 +86,12 @@ case class ClusterImage(tool: ClusterTool,
                         dockerImage: String,
                         timestamp: Instant)
 
+case class ClusterInternalId(value: String) extends ValueObject
+
 // The cluster itself
 // Also the API response for "list clusters" and "get active cluster"
 final case class Cluster(id: Long = 0, // DB AutoInc
+                   internalId: ClusterInternalId,
```
rtitle, line 94: I just realized this adds a field to our API json response; will update Swagger.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/ProxyService.scala
```diff
@@ -75,13 +75,35 @@ class ProxyService(proxyConfig: ProxyConfig,
...
+          } map {
+            case Some(id) => id
+            case None => throw ProxyException(googleProject, clusterName)
+          }
+        }
+      }
+    )
+
+  def getCachedClusterInternalId(googleProject: GoogleProject, clusterName: ClusterName): Future[ClusterInternalId] = {
+    clusterInternalIdCache.get((googleProject, clusterName))
```
Qi77Qi: should this return `Future[Option[ClusterInterald]]`?

Qi77Qi (approved):

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/sam/SamAuthProvider.scala
rtitle: This is kind of a hack to pass the implicit EC to the guava cache loading function. Implicits aren't threaded through into guava, so the only way for the cache loader to get it is via the cache key. I tested once upon a time and the cache still works correctly with the EC in the cache key.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/ProxyService.scala
rtitle: You know, that might be better -- we can deal with the `None` case outside of the cache, in `authCheck`. Will make that change.

PR #1008, opened 2019-07-18
Hack and slash automation tests

rtitle:
See [this doc](https://docs.google.com/document/d/1YaoOQFCGg49UEIJoOBAMi8Np1lvb4bpSiIM_PkskENQ/edit?ts=5d35fb56#) which attempts to summarize the changes here.

This was originally going to be a smaller change but turned into a larger refactor once I saw the state of some of the auto-tests.

This PR does the following:
- Reorganizes the specs into more appropriate categories
- Removes tests that are obsolete, duplicated, or not useful
- Changes the tests to claim a gpalloc project only once, at the beginning of all tests
- Removes JVM forking, and instead uses `ParallelTestExecution` for parallelism

This should hopefully result in fewer, more reliable, better organized tests. 

Have you read [CONTRIBUTING.md](https://github.com/DataBiosphere/leonardo/blob/develop/CONTRIBUTING.md) lately? If not, do that first.

I, the developer opening this PR, do solemnly pinky swear that:

- [ ] I've documented my API changes in Swagger

In all cases:

- [ ] Get a thumbsworth of review and PO signoff if necessary
- [ ] Verify all tests go green
- [ ] Squash and merge; you can delete your branch after this
- [ ] Test this change deployed correctly and works on dev environment after deployment

PR #1010, opened 2019-07-19
Jc automation more tests

jdcanas:
More tests!

In all cases:


automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookExtensionSpec.scala
```diff
@@ -87,13 +88,14 @@ class NotebookExtensionSpec extends ClusterFixtureSpec with NotebookTestUtils {
               notebookPage.addCodeAndExecute("1+1")
 
               notebookPage.saveNotebook()
-              //sleep 4 minutes. We do this to ensure the assertions are true after a certain about of time
+
+              //sleep 4 minutes. We do this to ensure the assertions are true after a certain about of time.
+              //eventually is not applicable as we want to ensure its true after the time period, not at some point until timeout
               logger.info("Waiting 4 minutes as lock takes time to be reflected in metadata")
               Thread.sleep(240000)
```
Qi77Qi, line 95: can we remove this sleep now?

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/Notebook.scala
```diff
@@ -45,6 +48,16 @@ object Notebook extends RestClient with LazyLogging {
     new NotebooksListPage(url + path)
   }
 
+  def createFileAtRootDir(googleProject: GoogleProject, clusterName: ClusterName, fileName: String)(implicit token: AuthToken): File = {
```
Qi77Qi: maybe rename this to something like `createFileInJupyter`?
It was weird to me that creating a file at Root dir requires a http call

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookExtensionSpec.scala
jdcanas: I don't think we can. I added an additional comment to explain why it's needed. The goal is to check that after 4minutes, the condition holds true. I want to make sure no subsequent periodic ticks mess up the assertions. Eventually exits as soon as it is true. I looked through the ```org.scalatest.concurrent``` documentation and couldn't find any cleaner ways to do this.

rtitle (approved):
Looks good. Once this is in I can merge it into https://github.com/DataBiosphere/leonardo/pull/1008, which is nearly ready.

Qi77Qi (approved):

PR #1016, opened 2019-07-29
[IA-1183] Jc monitor workers

jdcanas:

In all cases:

### DIFF
--- a/src/main/resources/jupyter/init-actions.sh
+++ b/src/main/resources/jupyter/init-actions.sh
@@ -20,10 +20,21 @@ set -e -x
 # Array for new relic instrumentation
 # UPDATE THIS IF YOU ADD MORE STEPS:
 # currently the steps are:
-# (START) init, after env setup, after installs, after copying files from google
-# after docker_compose, after welder docker start, after jupyter docker start, after nbextension install
-# after server extension install, after combined extension install, after static files copied to docker, after jupyter user script
-# after lab extension install, after jupyter notebook start, after jupyter docker start, after python install (END)
+# START init,
+# .. after env setup
+# .. after installs
+# .. after copying files from google and into docker
+# .. after docker compose
+# .. after welder start
+# .. after hail and spark
+# .. after nbextension install
+# .. after server extension install
+# .. after combined extension install
+# .. after static files copied to docker
+# .. after jupyter user script
+# .. after lab extension install
+# .. after jupyter notebook start
+# after python install (END)
 STEP_TIMINGS=($(date +%s))
 
 function retry {
@@ -389,9 +400,7 @@ if [[ "${ROLE}" == 'Master' ]]; then
        STEP_TIMINGS+=($(date +%s))
 
        retry 5 docker exec -u root ${JUPYTER_SERVER_NAME} chown -R jupyter-user:users ${JUPYTER_HOME}
-       retry 5 docker exec -u root ${JUPYTER_SERVER_NAME} chown -R jupyter-user:users /usr/local/share/jupyter/lab
-
-       STEP_TIMINGS+=($(date +%s))
+       retry 5 docker exec -u root ${JUPYTER_SERVER_NAME} chown -R jupyter-user:users /usr/local/share/jupyter/la
 
       #Install lab extensions
       #Note: lab extensions need to installed as jupyter user, not root
--- a/src/main/resources/jupyter/init-actions.sh
+++ b/src/main/resources/jupyter/init-actions.sh
@@ -458,4 +458,4 @@ log "Finished installing Python $PYTHON_VERSION"
 
 STEP_TIMINGS+=($(date +%s))
 
-log "Timings: $STEP_TIMINGS"
+log "Timings: ${STEP_TIMINGS[@]}"
--- a/src/main/resources/jupyter/init-actions.sh
+++ b/src/main/resources/jupyter/init-actions.sh
@@ -344,7 +344,7 @@ if [[ "${ROLE}" == 'Master' ]]; then
       if [ ! -z ${SAFE_MODE_JS_URI} ] && [ "${WELDER_ENABLED}" == "true" ]; then
         log 'Installing safe-mode.js extension...'
         gsutil cp ${SAFE_MODE_JS_URI} /etc
-        SAFE_MODE_JS=`basename ${SAFE_MODE_JS_URI}`\
+        SAFE_MODE_JS=`basename ${SAFE_MODE_JS_URI}`
         docker cp /etc/${SAFE_MODE_JS} ${JUPYTER_SERVER_NAME}:${JUPYTER_USER_HOME}/.jupyter/custom/
       fi
 
@@ -433,7 +433,7 @@ if [[ "${ROLE}" == 'Master' ]]; then
 
        STEP_TIMINGS+=($(date +%s))
     fi
-fiSTE
+fi
 
 # Install Python 3.6 on the master and worker VMs
 export PYTHON_VERSION=3.6.8
--- a/src/main/resources/jupyter/init-actions.sh
+++ b/src/main/resources/jupyter/init-actions.sh
@@ -391,6 +391,8 @@ if [[ "${ROLE}" == 'Master' ]]; then
        retry 5 docker exec -u root ${JUPYTER_SERVER_NAME} chown -R jupyter-user:users ${JUPYTER_HOME}
        retry 5 docker exec -u root ${JUPYTER_SERVER_NAME} chown -R jupyter-user:users /usr/local/share/jupyter/lab
 
+       STEP_TIMINGS+=($(date +%s))
+
       #Install lab extensions
       #Note: lab extensions need to installed as jupyter user, not root
       if [ ! -z "${JUPYTER_LAB_EXTENSIONS}" ] ; then
@@ -431,7 +433,7 @@ if [[ "${ROLE}" == 'Master' ]]; then
 
        STEP_TIMINGS+=($(date +%s))
     fi
-fi
+fiSTE
 
 # Install Python 3.6 on the master and worker VMs
 export PYTHON_VERSION=3.6.8
--- a/src/main/resources/jupyter/init-actions.sh
+++ b/src/main/resources/jupyter/init-actions.sh
@@ -17,6 +17,15 @@ set -e -x
 # Retry 4/5 exited 1, retrying in 16 seconds...
 # Retry 5/5 exited 1, no more retries left.
 #
+# Array for new relic instrumentation
+# UPDATE THIS IF YOU ADD MORE STEPS:
+# currently the steps are:
+# (START) init, after env setup, after installs, after copying files from google
+# after docker_compose, after welder docker start, after jupyter docker start, after nbextension install
+# after server extension install, after combined extension install, after static files copied to docker, after jupyter user script
+# after lab extension install, after jupyter notebook start, after jupyter docker start, after python install (END)
+STEP_TIMINGS=($(date +%s))
+
 function retry {
   local retries=$1
   shift
@@ -97,10 +106,16 @@ if [[ "${ROLE}" == 'Master' ]]; then
     JUPYTER_NB_EXTENSIONS=$(jupyterNbExtensions)
     JUPYTER_COMBINED_EXTENSIONS=$(jupyterCombinedExtensions)
     JUPYTER_LAB_EXTENSIONS=$(jupyterLabExtensions)
+    GOOGLE_SIGN_IN_JS_URI=$(googleSignInJsUri)
+    EXTENSION_ENTRY_URI=$(extensionEntryUri)
+    JUPYTER_LAB_GOOGLE_PLUGIN_URI=$(jupyterLabGooglePluginUri)
+    SAFE_MODE_JS_URI=$(safeModeJsUri)
+    EDIT_MODE_JS_URI=$(editModeJsUri)
     JUPYTER_USER_SCRIPT_URI=$(jupyterUserScriptUri)
     JUPYTER_USER_SCRIPT_OUTPUT_URI=$(jupyterUserScriptOutputUri)
     JUPYTER_NOTEBOOK_CONFIG_URI=$(jupyterNotebookConfigUri)
-    JUPYTER_NOTEBOOK_FRONTEND_CONFIG_URI=$(jupyterNotebookFrontendConfigUri)
+
+    STEP_TIMINGS+=($(date +%s))
 
     log 'Installing prerequisites...'
 
@@ -141,6 +156,8 @@ if [[ "${ROLE}" == 'Master' ]]; then
     # https://docs.docker.com/compose/install/#install-compose
     retry 5 curl -L "https://github.com/docker/compose/releases/download/1.22.0/docker-compose-$(uname -s)-$(uname -m)" -o /usr/local/bin/docker-compose
     chmod +x /usr/local/bin/docker-compose
+
+    STEP_TIMINGS+=($(date +%s))
 
     log 'Copying secrets from GCS...'
 
@@ -190,6 +207,8 @@ if [[ "${ROLE}" == 'Master' ]]; then
       gcloud docker --authorize-only
     fi
 
+    STEP_TIMINGS+=($(date +%s))
+
     log 'Starting up the Jupydocker...'
 
     # Run docker-compose for each specified compose file.
@@ -215,12 +234,16 @@ if [[ "${ROLE}" == 'Master' ]]; then
     retry 5 docker-compose "${COMPOSE_FILES[@]}" pull
     retry 5 docker-compose "${COMPOSE_FILES[@]}" up -d
 
+    STEP_TIMINGS+=($(date +%s))
+
     # if Welder is installed, start the service.
     # See https://broadworkbench.atlassian.net/browse/IA-1026
     if [ ! -z ${WELDER_DOCKER_IMAGE} ] && [ "${WELDER_ENABLED}" == "true" ] ; then
       log 'Starting Welder file synchronization service...'
       retry 3 docker exec -d ${WELDER_SERVER_NAME} /opt/docker/bin/entrypoint.sh
     fi
+
+    STEP_TIMINGS+=($(date +%s))
 
     # Jupyter-specific setup, only do if Jupyter is installed
     if [ ! -z ${JUPYTER_DOCKER_IMAGE} ] ; then
@@ -234,22 +257,7 @@ if [[ "${ROLE}" == 'Master' ]]; then
       # Install the Hail additions to Spark conf.
       retry 3 docker exec -u root ${JUPYTER_SERVER_NAME} ${JUPYTER_SCRIPTS}/hail/spark_install_hail.sh
 
-      # Install jupyter_notebook_config.py
-      if [ ! -z ${JUPYTER_NOTEBOOK_CONFIG_URI} ] ; then
-        log 'Copy Jupyter notebook config...'
-        gsutil cp ${JUPYTER_NOTEBOOK_CONFIG_URI} /etc
-        JUPYTER_NOTEBOOK_CONFIG=`basename ${JUPYTER_NOTEBOOK_CONFIG_URI}`
-        docker cp /etc/${JUPYTER_NOTEBOOK_CONFIG} ${JUPYTER_SERVER_NAME}:${JUPYTER_HOME}/
-      fi
-
-      # Install notebook.json
-      if [ ! -z ${JUPYTER_NOTEBOOK_FRONTEND_CONFIG_URI} ] ; then
-        log 'Copy Jupyter frontend notebook config...'
-        gsutil cp ${JUPYTER_NOTEBOOK_FRONTEND_CONFIG_URI} /etc
-        JUPYTER_NOTEBOOK_FRONTEND_CONFIG=`basename ${JUPYTER_NOTEBOOK_FRONTEND_CONFIG_URI}`
-        docker cp /etc/${JUPYTER_NOTEBOOK_FRONTEND_CONFIG} ${JUPYTER_SERVER_NAME}:${JUPYTER_HOME}/nbconfig/
-      fi
-
+      STEP_TIMINGS+=($(date +%s))
 
       #Install NbExtensions
       if [ ! -z "${JUPYTER_NB_EXTENSIONS}" ] ; then
@@ -272,6 +280,8 @@ if [[ "${ROLE}" == 'Master' ]]; then
         done
       fi
 
+      STEP_TIMINGS+=($(date +%s))
+
       #Install serverExtensions
       if [ ! -z "${JUPYTER_SERVER_EXTENSIONS}" ] ; then
         for ext in ${JUPYTER_SERVER_EXTENSIONS}
@@ -287,6 +297,8 @@ if [[ "${ROLE}" == 'Master' ]]; then
           fi
         done
       fi
+
+      STEP_TIMINGS+=($(date +%s))
 
       #Install combined extensions
       if [ ! -z "${JUPYTER_COMBINED_EXTENSIONS}"  ] ; then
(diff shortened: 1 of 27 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterServiceMonitor.scala
```diff
@@ -0,0 +1,105 @@
...
+import org.broadinstitute.dsde.workbench.leonardo.monitor.ClusterServiceMonitor.{DetectClusterStatus, JupyterStatus, Status, TimerKey, WelderStatus}
+import org.broadinstitute.dsde.workbench.model.google.GoogleProject
+import org.broadinstitute.dsde.workbench.newrelic.NewRelicMetrics
+
+import scala.concurrent.Future
+
+object ClusterServiceMonitor {
+
+
+  val newRelic: NewRelicMetrics = Metrics.newRelic
```
rtitle: Is this `val` used? It looks like `ClusterServiceMonitor` uses the `NewRelicMetrics` passed to its constructor.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterServiceMonitor.scala
```diff
@@ -0,0 +1,105 @@
...
+  def props(config: ClusterServiceConfig, gdDAO: GoogleDataprocDAO, googleProjectDAO: GoogleProjectDAO, dbRef: DbReference, welderDAO: WelderDAO, jupyterDAO: JupyterDAO, newRelic: NewRelicMetrics): Props = {
+    Props(new ClusterServiceMonitor(config, gdDAO, googleProjectDAO, dbRef, welderDAO, jupyterDAO, newRelic: NewRelicMetrics))
+  }
+
+  sealed trait ClusterServiceMonitorMessage
+  case object DetectClusterStatus extends ClusterServiceMonitorMessage
+  case object TimerKey extends ClusterServiceMonitorMessage
+
+  sealed trait ServiceStatus extends Product with Serializable
+  sealed trait JupyterStatus extends ServiceStatus with Product with Serializable
```
rtitle: I wonder if the status checks could be generalized to other tools besides Jupyter/Welder. For example `ClusterMontitorActor` takes a `Map[ClusterTool, () => Future[Boolean]]` for status checking tools. There may be opportunities for code reuse in `ClusterServiceMonitor`.

See https://github.com/DataBiosphere/leonardo/blob/develop/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorSupervisor.scala#L157

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterServiceMonitor.scala
```diff
@@ -0,0 +1,105 @@
...
+
+          ()
+          }
+        }
+      }
+  }
+
+  private def getActiveClustersFromDatabase: Future[Map[GoogleProject, Seq[Cluster]]] = {
+    dbRef.inTransaction {
+      _.clusterQuery.listActive
```
rtitle: I think this query joins to `CLUSTER_LABEL` which is not needed here - could we make a `listActiveClusterOnly` method which avoids the join?

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterServiceMonitor.scala
```diff
@@ -0,0 +1,105 @@
...
+  override def preStart(): Unit = {
+    super.preStart()
+    timers.startPeriodicTimer(TimerKey, DetectClusterStatus, config.pollPeriod)
+  }
+
+  override def receive: Receive = {
+    case DetectClusterStatus =>
+      // Get active clusters from the Leo DB, grouped by project
+      val activeClusters: Future[List[Cluster]] = getActiveClustersFromDatabase.flatMap(clusterMap => {
+        Future.successful(clusterMap.values.flatten.toList)
```
rtitle: Why group by project in `getActiveClustersFromDatabase` if you're just flattening anyway?

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterServiceMonitor.scala
```diff
@@ -0,0 +1,105 @@
...
+      })
+
+      //check the status of the workers in each active cluster and log instances of a down worker to new relic
+      activeClusters.foreach { cs =>
+        cs.foreach { cluster =>
+          checkClusterStatus(cluster).foreach { status =>
+
+            if (status.jupyterStatus.equals(JupyterStatus.JupyterDown)) {
+              logger.info("jupyter down")
```
rtitle: Probably would be useful to log the cluster in these messages

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterServiceMonitor.scala
```diff
@@ -0,0 +1,105 @@
...
+
+  override def receive: Receive = {
+    case DetectClusterStatus =>
+      // Get active clusters from the Leo DB, grouped by project
+      val activeClusters: Future[List[Cluster]] = getActiveClustersFromDatabase.flatMap(clusterMap => {
+        Future.successful(clusterMap.values.flatten.toList)
+      })
+
+      //check the status of the workers in each active cluster and log instances of a down worker to new relic
+      activeClusters.foreach { cs =>
```
rtitle: It doesn't really matter here because nothing is consuming the results of these futures -- but in general it's not great practice to use `forEach` to spawn a bunch of fire-and-forget futures. I'd recommend chaining the futures together such that a single future is returned.

Could the code be structured like the following?

```
      val future: Future[Unit] = for {
        activeClusters <- getActiveClustersFromDatabase()
        statuses <- Future.traverse(activeClusters)(checkClusterStatus)
        _ <- Future.traverse(statuses)(handleStatus)
      } yield ()

      future.failed.foreach { e =>
        logger.error("Error occurred monitoring active clusters", e)
      }
```

src/test/scala/org/broadinstitute/dsde/workbench/leonardo/api/StatusRoutesSpec.scala
```diff
@@ -42,7 +42,7 @@ class StatusRoutesSpec extends FlatSpec with Matchers with ScalatestRouteTest wi
     }
   }
 
-  it should "give 500 for not ok" in {
+  it should "give 500 fornot ok" in {
```
rtitle: typo

src/test/scala/org/broadinstitute/dsde/workbench/leonardo/dao/MockJupyterDAO.scala
```diff
@@ -5,8 +5,13 @@ import org.broadinstitute.dsde.workbench.model.google.GoogleProject
 import scala.concurrent.Future
 
 object MockJupyterDAO extends JupyterDAO {
-
   override def isProxyAvailable(googleProject: GoogleProject, clusterName: ClusterName): Future[Boolean] = Future.successful(true)
 
   override def isAllKernalsIdle(googleProject: GoogleProject, clusterName: ClusterName): Future[Boolean] = Future.successful(true)
 }
+
+class MockJupyterDAO(isUp: Boolean) extends JupyterDAO {
```
rtitle: It feels weird to have an object `MockJupyterDAO` and class `MockJupyterDAO` both extending `JupyterDAO`. Could we maybe just have the class with a default parameter for `isUp`?

src/test/scala/org/broadinstitute/dsde/workbench/leonardo/dao/MockWelderDAO.scala
```diff
@@ -10,3 +10,9 @@ object MockWelderDAO extends WelderDAO {
 
   override def isProxyAvailable(googleProject: GoogleProject, clusterName: ClusterName): Future[Boolean] = Future.successful(true)
 }
+
+class MockWelderDAO(isUp: Boolean) extends WelderDAO {
```
rtitle: Same as above

rtitle (changes requested):
In general looks good and it does what it's supposed to. Left some comments/suggestions above.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterServiceMonitor.scala
```diff
@@ -0,0 +1,100 @@
...
+      logger.info(s"jupyter down for cluster ${status.cluster.clusterName.value} in project ${status.cluster.googleProject.value}")
+      newRelic.incrementCounterIO("jupyterDown").unsafeRunAsync(_ => ())
+    }
+
+    if (status.welderStatus.equals(WelderStatus.WelderDown)) {
+      logger.info(s"welder enabled and down for cluster ${status.cluster.clusterName.value} in project ${status.cluster.googleProject.value}")
+      newRelic.incrementCounterIO("welderDown").unsafeRunAsync(_ => ())
+    }
+
+    Future.unit
```
Qi77Qi: the two `if` statements return `Unit`. So there's no point returning a `Future[Unit]` if you're doing it this way.

I would do the following instead:
```
    for {
      _ <-     if (status.jupyterStatus.equals(JupyterStatus.JupyterDown)) {
        logger.info(s"jupyter down for cluster ${status.cluster.clusterName.value} in project ${status.cluster.googleProject.value}")
        newRelic.incrementCounterIO("jupyterDown")
      } else IO.unit
      _ <- if (status.welderStatus.equals(WelderStatus.WelderDown)) {
        logger.info(s"welder enabled and down for cluster ${status.cluster.clusterName.value} in project ${status.cluster.googleProject.value}")
        newRelic.incrementCounterIO("welderDown")
      } else IO.unit
    } yield ()
```

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterServiceMonitor.scala
```diff
@@ -0,0 +1,100 @@
...
+    }
+  }
+
+  def checkClusterStatus(cluster: Cluster): Future[Status] = {
+    for {
+      isWelderUp <- welderDAO.isProxyAvailable(cluster.googleProject, cluster.clusterName)
+      isJupyterUp <- jupyterDAO.isProxyAvailable(cluster.googleProject, cluster.clusterName)
+
+      //if welder isn't enabled, the status is will be OK
+      welderStatus: WelderStatus = if (!isWelderUp && cluster.welderEnabled) WelderStatus.WelderDown else WelderStatus.WelderOK
```
Qi77Qi: remove `: WelderStatus` and `: JupyterStatus` since they introduce an extra `withFilter` call...or move them to the yield block
```
yield {
      val welderStatus: WelderStatus = if (!isWelderUp && cluster.welderEnabled) WelderStatus.WelderDown else WelderStatus.WelderOK
      val jupyterStatus: JupyterStatus = if (isJupyterUp) JupyterStatus.JupyterOK else JupyterStatus.JupyterDown
      Status(welderStatus, jupyterStatus, cluster)
    }
```

src/test/resources/reference.conf
```diff
@@ -219,6 +219,10 @@ zombieClusterMonitor {
   creationHangTolerance = 9 second
 }
 
+clusterServiceMonitor {
+  pollPeriod = 1 second
```
Qi77Qi, line 219: maybe a bit less frequent?

jdcanas: this is the test config, its what the zombie monitor does and it makes the unit tests faster. I typically wait for 3 periods to ensure proper behavior

jdcanas:
@rtitle @Qi77Qi Addressed all feedback and rebased. The code is significantly changed so it can probably use some eyes again

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterToolMonitor.scala
```diff
@@ -0,0 +1,82 @@
...
+
+  private def handleClusterStatus(statuses: List[ToolStatus]): Future[Unit] = {
+    statuses.foreach { status =>
+      if (!status.isUp) {
+        val toolName = status.tool.toString
+        logger.info(s"The tool ${toolName} is down")
+        newRelic.incrementCounterIO(toolName + "Down")
+      }
+    }
+    Future.unit
```
Qi77Qi: this will not do anything. You're just constructing a list of `IO`s and they'll never run

src/main/resources/jupyter/init-actions.sh
```diff
@@ -106,6 +127,8 @@ if [[ "${ROLE}" == 'Master' ]]; then
     JUPYTER_USER_SCRIPT_OUTPUT_URI=$(jupyterUserScriptOutputUri)
     JUPYTER_NOTEBOOK_CONFIG_URI=$(jupyterNotebookConfigUri)
 
+    STEP_TIMINGS+=($(date +%s))
```
Qi77Qi, line 125: how are these metrics being sent to newrelic?

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterToolMonitor.scala
jdcanas: Ah I was experimenting with incrementCounterFuture but forgot to add the runSync when I changed it back, good catch. Unfortunately, I don't think the tests can verify this because it seems mockito only verifies the last call in the chain (i.e., if I specify  ```verify(mockNewRelic, times(3)).incrementCounterIO("JupyterDown").unsafeRunSync()``` in the unit tests, the test will ignore the string passed to incrementCounterIO in the verify() call and only checks that unsafeRunSync() was called, which defeats the purpose)

src/main/resources/jupyter/init-actions.sh
jdcanas: they are not. Since this is run on the cluster, we run into the same issue sending them to new relic as we do with welder metrics. You have to inspect them in the log on the cluster

src/main/resources/jupyter/init-actions.sh
```diff
@@ -17,6 +17,26 @@ set -e -x
 # Retry 4/5 exited 1, retrying in 16 seconds...
 # Retry 5/5 exited 1, no more retries left.
 #
+# Array for new relic instrumentation
```
Qi77Qi: can you remove the mention of newrelic since the metrics are not sent to newrelic

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterToolMonitor.scala
Qi77Qi: why not just use https://github.com/DataBiosphere/leonardo/pull/1016#discussion_r312968160 ?

src/main/resources/jupyter/init-actions.sh
```diff
@@ -17,6 +17,26 @@ set -e -x
 # Retry 4/5 exited 1, retrying in 16 seconds...
 # Retry 5/5 exited 1, no more retries left.
 #
+# Array for instrumentation
+# UPDATE THIS IF YOU ADD MORE STEPS:
+# currently the steps are:
+# START init,
```
rtitle, line 23: Ah sorry, this will conflict a bit with https://github.com/DataBiosphere/leonardo/pull/1034, where I split the init script into 2 to support custom Dataproc images. Hopefully it shouldn't be too bad to merge.

src/main/resources/jupyter/init-actions.sh
```diff
@@ -97,10 +117,16 @@ if [[ "${ROLE}" == 'Master' ]]; then
     JUPYTER_NB_EXTENSIONS=$(jupyterNbExtensions)
     JUPYTER_COMBINED_EXTENSIONS=$(jupyterCombinedExtensions)
     JUPYTER_LAB_EXTENSIONS=$(jupyterLabExtensions)
+    GOOGLE_SIGN_IN_JS_URI=$(googleSignInJsUri)
```
rtitle: Sorry again - these vars look wrong post-rebase.

These should be removed (because they're now in the docker image):
GOOGLE_SIGN_IN_JS_URI
EXTENSION_ENTRY_URI
JUPYTER_LAB_GOOGLE_PLUGIN_URI
SAFE_MODE_JS_URI
EDIT_MODE_JS_URI

This should be added (because it configures the JS code):
JUPYTER_NOTEBOOK_FRONTEND_CONFIG_URI

src/main/resources/jupyter/init-actions.sh
rtitle: Seems like they're just logged, not sent to NR. We shouldn't put a NR key on users' clusters. This is a good first step; wonder if there is a solution for getting this data to NR.

src/main/resources/jupyter/init-actions.sh
```diff
@@ -234,22 +268,7 @@ if [[ "${ROLE}" == 'Master' ]]; then
       # Install the Hail additions to Spark conf.
       retry 3 docker exec -u root ${JUPYTER_SERVER_NAME} ${JUPYTER_SCRIPTS}/hail/spark_install_hail.sh
 
-      # Install jupyter_notebook_config.py
```
rtitle, line 237: Needs to be added back

src/main/resources/jupyter/init-actions.sh
```diff
@@ -305,6 +328,54 @@ if [[ "${ROLE}" == 'Master' ]]; then
         done
       fi
 
+      STEP_TIMINGS+=($(date +%s))
+
+      retry 3 docker exec -u root -e PIP_USER=false ${JUPYTER_SERVER_NAME} ${JUPYTER_SCRIPTS}/extension/install_jupyter_contrib_nbextensions.sh
```
rtitle: Needs to be removed

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpJupyterDAO.scala
```diff
@@ -57,7 +57,7 @@ object HttpJupyterDAO {
   implicit val sessionDecoder: Decoder[Session] = Decoder.forProduct1("kernel")(Session)
 }
 
-trait JupyterDAO {
+trait JupyterDAO extends ToolDAO {
   def isAllKernalsIdle(googleProject: GoogleProject, clusterName: ClusterName): Future[Boolean]
   def isProxyAvailable(googleProject: GoogleProject, clusterName: ClusterName): Future[Boolean]
```
rtitle: Don't need to redefine `isProxyAvailable` if it's defined in the base trait

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpRStudioDAO.scala
```diff
@@ -27,6 +27,6 @@ class HttpRStudioDAO(val clusterDnsCache: ClusterDnsCache)(implicit system: Acto
   }
 }
 
-trait RStudioDAO {
+trait RStudioDAO extends ToolDAO {
   def isProxyAvailable(googleProject: GoogleProject, clusterName: ClusterName): Future[Boolean]
```
rtitle: Same as above - can remove `isProxyAvailable`

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpWelderDAO.scala
```diff
@@ -45,7 +45,7 @@ class HttpWelderDAO(val clusterDnsCache: ClusterDnsCache)
   }
 }
 
-trait WelderDAO {
+trait WelderDAO extends ToolDAO {
   def flushCache(googleProject: GoogleProject, clusterName: ClusterName): Future[Unit]
   def isProxyAvailable(googleProject: GoogleProject, clusterName: ClusterName): Future[Boolean]
```
rtitle: Same as above - can remove `isProxyAvailable`

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpJupyterDAO.scala
```diff
@@ -57,7 +57,7 @@ object HttpJupyterDAO {
   implicit val sessionDecoder: Decoder[Session] = Decoder.forProduct1("kernel")(Session)
 }
 
-trait JupyterDAO {
+trait JupyterDAO extends ToolDAO {
   def isAllKernalsIdle(googleProject: GoogleProject, clusterName: ClusterName): Future[Boolean]
```
rtitle, line 61: Not your code, but "kernels" is misspelled :)

src/main/resources/reference.conf
```diff
@@ -122,6 +122,10 @@ zombieClusterMonitor {
   creationHangTolerance = 1 hour
 }
 
+clusterServiceMonitor {
```
rtitle: Minor: maybe name this `clusterToolMonitor` to match the class name it corresponds with

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/ClusterServiceConfig.scala
```diff
@@ -0,0 +1,5 @@
+package org.broadinstitute.dsde.workbench.leonardo.config
+
+import scala.concurrent.duration.FiniteDuration
+
+case class ClusterServiceConfig(pollPeriod: FiniteDuration)
```
rtitle: `ClusterToolMonitorConfig`?

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterToolMonitor.scala
```diff
@@ -0,0 +1,82 @@
...
+
+  def props(config: ClusterServiceConfig, gdDAO: GoogleDataprocDAO, googleProjectDAO: GoogleProjectDAO, dbRef: DbReference, toolDAOs: Map[ClusterTool, ToolDAO], newRelic: NewRelicMetrics): Props = {
+    Props(new ClusterToolMonitor(config, gdDAO, googleProjectDAO, dbRef, toolDAOs, newRelic))
+  }
+
+  sealed trait ClusterServiceMonitorMessage
+  case object DetectClusterStatus extends ClusterServiceMonitorMessage
+  case object TimerKey extends ClusterServiceMonitorMessage
+
+  case class ToolStatus(val isUp: Boolean, val tool: ClusterTool)
```
rtitle: It might be good to record the cluster project/name in `ToolStatus`. Otherwise we'll just log "welder is down" without knowing which cluster, which is probably useful information. :)

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterToolMonitor.scala
```diff
@@ -0,0 +1,82 @@
...
+  override def receive: Receive = {
+    case DetectClusterStatus =>
+      for {
+        activeClusters <- getActiveClustersFromDatabase
+        statuses <- Future.traverse(activeClusters)(checkClusterStatus)
+        _ <- Future.traverse(statuses)(handleClusterStatus)
+      } yield ()
+  }
+
+  private def handleClusterStatus(statuses: List[ToolStatus]): Future[Unit] = {
```
rtitle: Based on `Future.traverse(statuses)(handleClusterStatus)` I would expect `handleClusterStatus` to take a single `ToolStatus`, not a `List[ToolStatus]`. Unless I'm misunderstanding something..

rtitle (changes requested):
A few more comments above. Sorry for the rebase pain due to some of my PRs -- if you want we could work through it together.

Qi77Qi (approved):

rtitle (approved):
Looks good!

PR #1020, opened 2019-08-01
[IA-1228] Dismiss 'notebook changed' modal

rtitle:
Note: I don't love this solution. I'm trying to address this intermittent issue in automation tests:

image

It's happening because the filesystem last_modified timestamp is sometimes a few seconds _after_ the notebook creation timestamp. Jupyter displays this modal if there is a diff of more than 500ms.

I am not sure why the timestamps are different. I can't seem to reproduce it when manually going through the same steps as the tests. It could be a clock sync issue -- or it could be the docker shared volume is messing with file mtimes somehow. Not sure.

This PR makes the tests resilient to this situation by dismissing the modal. If this were a true product issue I'd be hesitant to fixing the tests like this -- but given it's not reproducible (and we've had no user complaints) I'm ok with this change.

I did verify manually that the new selenium logic does the right thing.

In all cases:

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookPage.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookPage.scala
@@ -94,6 +94,10 @@ class NotebookPage(override val url: String)(override implicit val authToken: Au
   lazy val notebookNotification: Query = cssSelector("[id='notification_notebook']")
 
   lazy val saveButton: Query = cssSelector(s"[id='${saveButtonId}']")
+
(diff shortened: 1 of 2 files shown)


jdcanas (approved):
It feels like this could hide bugs from the automation test, as this showing up was a legitimate bug at one point in welder's life time: https://broadworkbench.atlassian.net/browse/IA-1111

I don't see how any options other than possibly 1 from the email chain would alleviate this concern though.

 Perhaps we add a comment above the method linking the ticket of the bug Gary posted and call it a day.

rtitle:
Thanks @jdcanas. I agree this could hide bugs, that's why I don't love this solution. But I'm not able to find or reproduce the bug, so I'm kinda stuck. This is my "big hammer" solution. :)

Good idea about adding a comment, I'll do that.

PR #1024, opened 2019-08-05
First cut at passing region to cluster creation, etc.

gbggrant:

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/google/HttpGoogleDataprocDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/google/HttpGoogleDataprocDAO.scala
@@ -16,6 +16,7 @@ import com.google.api.services.dataproc.model.{Cluster => DataprocCluster, Clust
 import com.google.api.services.oauth2.Oauth2
 import org.broadinstitute.dsde.workbench.google.AbstractHttpGoogleDAO
 import org.broadinstitute.dsde.workbench.google.GoogleCredentialModes._
+import org.broadinstitute.dsde.workbench.google.GoogleUtilities.RetryPredicates._
 import org.broadinstitute.dsde.workbench.leonardo.model.google.DataprocRole.{Master, SecondaryWorker, Worker}
 import org.broadinstitute.dsde.workbench.leonardo.model.google._
 import org.broadinstitute.dsde.workbench.leonardo.service.{AuthenticationError, DataprocDisabledException}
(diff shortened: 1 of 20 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/google/HttpGoogleDataprocDAO.scala
```diff
@@ -73,7 +74,7 @@ class HttpGoogleDataprocDAO(appName: String,
 
   }
 
-  override def deleteCluster(googleProject: GoogleProject, clusterName: ClusterName): Future[Unit] = {
+  override def deleteCluster(googleProject: GoogleProject, clusterName: ClusterName, region: Option[String]): Future[Unit] = {
     val request = dataproc.projects().regions().clusters().delete(googleProject.value, defaultRegion, clusterName.value)
```
MatthewBemis: did you mean to pass `region` into here?

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
```diff
@@ -212,6 +213,10 @@ class ClusterMonitorActor(val cluster: Cluster,
     } yield res
   }
 
+  private def getRegion: Option[String] = {
+    cluster.labels.get(dataprocConfig.terraRegionLabel.get) flatMap WorkbenchProjectLocation.fromName map(_.storageLocation)
```
MatthewBemis: it's probably all the same in the end but we should probably tweak `WorkbenchProjectLocation` so we can easily use a `computeRegion` instead of `storageLocation`

kcibul: you should be able to call computeRegionAndZones which gives you a map of regions -> zones.  It's a map because I could see in regions where we use a multi-region bucket we might want to support multiple regions.  In this specific case, I would just take the first one since (a) we don't have multiple in the hackathon and (b) even if we did, we don't have anything more intelligent to do at this point!

kcibul: ...and they are different b/c for multi-region buckets, storage location isn't the same as a compute region (e.g. 'us' vs 'us-central1')

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/google/HttpGoogleDataprocDAO.scala
gbggrant: Yes thanks.

rtitle:
Going to close this in favor of https://github.com/DataBiosphere/leonardo/pull/1933

PR #1034, opened 2019-08-15
[IA-1264] Leo changes to support custom dataproc images

rtitle:
There are a few different changes here, let me give the highlights.

- Adds a new config parameter: `dataprocConfig.customDataprocImage`. If set, Leo will create clusters using the provided Dataproc image. If unset, the behavior is the same as today.
- To facilitate this, I broke up our current init script into `init-vm.sh` (which installs VM prerequisites) and `init-actions.sh` (which does the Docker container setup). `init-vm.sh` is _only_ provided if `dataprocConfig.customDataprocImage` is _unset_. `init-actions.sh` is provided no matter what.
- Added a new IAM call to set `roles/compute.imageUser` which is needed to pull the custom image. This is only done if `dataprocConfig.customDataprocImage` is set.
- I couldn't resist doing some refactoring. I created a new `ClusterHelper` class, and centralized all cluster-related IAM calls there. In the future I want to use this class to centralize more cluster-related logic. In fact this is what I started doing in [IA-1156](https://broadworkbench.atlassian.net/browse/IA-1156), so this is a nice step in that direction.

I tested custom images in a fiab, and existing auto-tests seem to work. Again this PR should be mergeable with no change in Leo behavior -- we'll only start using a custom image once we do the automation work and set it in leo.conf.

In all cases:


jenkins/prepare-custom-leonardo-jupyter-dataproc-image.sh
```diff
@@ -0,0 +1,173 @@
...
+
+#
+# This script sets up a custom Dataproc image for Leonardo clusters.
+# See: https://cloud.google.com/dataproc/docs/guides/dataproc-images
+#
+
+#
+# Constants and Global Vars
+#
+leonardo_jupyter_docker_image_tag="dev"
```
rtitle: This script is used to actually create the custom Dataproc image. It is not currently hooked up to anything; but we intend to wire it to Jenkins once we do https://broadworkbench.atlassian.net/browse/IA-1265. I'm just committing here so we can iterate on it once we do the Jenkins work.

src/main/resources/jupyter/init-actions.sh
```diff
@@ -102,46 +112,6 @@ if [[ "${ROLE}" == 'Master' ]]; then
     JUPYTER_NOTEBOOK_CONFIG_URI=$(jupyterNotebookConfigUri)
     JUPYTER_NOTEBOOK_FRONTEND_CONFIG_URI=$(jupyterNotebookFrontendConfigUri)
 
-    log 'Installing prerequisites...'
```
rtitle, line 127: The "Installing prerequisites" stuff now lives in `init-vm.sh`. See PR description.

src/main/resources/jupyter/init-actions.sh
```diff
@@ -162,23 +132,8 @@ if [[ "${ROLE}" == 'Master' ]]; then
...
     touch auth_openidc.conf
 
-    # If we have a service account JSON file, create an .env file to set GOOGLE_APPLICATION_CREDENTIALS
-    # in the docker container. Otherwise, we should _not_ set this environment variable so it uses the
-    # credentials on the metadata server.
     if [ ! -z ${SERVICE_ACCOUNT_CREDENTIALS} ] ; then
       echo "GOOGLE_APPLICATION_CREDENTIALS=/etc/${SERVICE_ACCOUNT_CREDENTIALS}" > /etc/google_application_credentials.env
-      if [ ! -z ${JUPYTER_DOCKER_IMAGE} ] ; then
-        log 'Copying SA into Jupyter Docker...'
-        docker cp /etc/${SERVICE_ACCOUNT_CREDENTIALS} ${JUPYTER_SERVER_NAME}:/etc/${SERVICE_ACCOUNT_CREDENTIALS}
```
rtitle, line 196: This is an unrelated bug which I noticed during the hackathon: we're doing `docker cp` here before we actually start the docker containers (oops). We didn't notice this because we don't currently use `SERVICE_ACCOUNT_CREDENTIALS` in prod. Anyway, this change fixes it.

src/main/resources/jupyter/init-actions.sh
```diff
@@ -353,27 +324,4 @@ if [[ "${ROLE}" == 'Master' ]]; then
       retry 3 docker exec -d ${JUPYTER_SERVER_NAME} ${JUPYTER_SCRIPTS}/run-jupyter.sh ${NOTEBOOKS_DIR}
       log 'All done!'
     fi
-fi
-
-# Install Python 3.6 on the master and worker VMs
```
rtitle, line 401: Also moved to `init-vm.sh`.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
```diff
@@ -424,25 +425,18 @@ class ClusterMonitorActor(val cluster: Cluster,
           if (count > 0) {
             Future.successful(())
           } else {
-            // Retry 409s with exponential backoff. This can happen if concurrent policy updates are made in the same project.
```
rtitle, line 444: IAM logic lifted out of here and moved to `ClusterHelper`

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorSupervisor.scala
```diff
@@ -176,46 +177,41 @@ class ClusterMonitorSupervisor(monitorConfig: MonitorConfig, dataprocConfig: Dat
...
-                    Metrics.newRelic.incrementCounterIO("maxKernelExceededClusters").unsafeRunAsync(_ => ())
-                    logger.info(s"Auto pausing ${cluster.googleProject}/${cluster.clusterName} due to exceeded max kernel active time")
-                  } else {
-                    Metrics.newRelic.incrementCounterIO("activeKernelClusters").unsafeRunAsync(_ => ())
-                    logger.info(s"Not going to auto pause cluster ${cluster.googleProject}/${cluster.clusterName} due to active kernels")
-                  }
-                  maxKernelActiveTimeExceeded
-
-                } else isIdle
+    pauseableClusters <- clusters.toList.filterA[Future] { cluster =>
```
rtitle: There are no functional changes here -- just minor formatting cleanup, and fixes the red squiggles in Intellij.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorSupervisor.scala
```diff
@@ -176,46 +177,41 @@ class ClusterMonitorSupervisor(monitorConfig: MonitorConfig, dataprocConfig: Dat
...
         }
     }
     _ <- Metrics.newRelic.gauge("autopausedClusters", pauseableClusters.length).unsafeToFuture()
-    _ <- pauseableClusters.traverse{
-      cl =>
-        logger.info(s"Auto freezing cluster ${cl.clusterName} in project ${cl.googleProject}")
-        leonardoService.internalStopCluster(cl).attempt.map { e =>
-          e.fold(t => logger.warn(s"Error occurred auto freezing cluster ${cl.projectNameString}", e), identity)
-        }
+    _ <- pauseableClusters.traverse[Future, Unit] { cl =>
```
rtitle: Same here -- makes Intellij happy.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
```diff
@@ -0,0 +1,89 @@
...
+      }))
+
+    // Add the Dataproc Worker role in the user's project to the cluster service account, if present.
+    // This is needed to be able to spin up Dataproc clusters using a custom service account.
+    // If the Google Compute default service account is being used, this is not necessary.
+    val dataprocWorkerIO = serviceAccountInfo.clusterServiceAccount.map { email =>
+      retryIam(googleProject, email, Set("roles/dataproc.worker"))
+    } getOrElse IO.unit
+
+    // Add the Compute Image User role in Leo's project to the Google API service account.
```
rtitle: This threw me for a while -- there's a service account called the "Google API service account" which needs permission to pull the custom image. This is distinct from the Leo service account (which is used to make the Dataproc call) and the pet service account (whose credentials are put on the cluster).

rtitle:
jenkins retest

rtitle:
jenkins retest

jenkins/prepare-custom-leonardo-jupyter-dataproc-image.sh
```diff
@@ -0,0 +1,173 @@
...
+
+retry 5 curl -L "${docker_compose_binary_download_url:?}" -o "${docker_compose_binary_download_target_filename:?}"
+chmod +x "${docker_compose_binary_download_target_filename:?}"
+
+# Pull docker image versions as of the time this script ran; this caches them in the
+# dataproc custom instance image.
+if [[ -n ${docker_image_var_names:?} ]]; then
+    for _docker_image_var_name in ${docker_image_var_names:?}
+    do
+        _docker_image="${!_docker_image_var_name:?}"
```
jdcanas: what does the exclamation point do here?

jenkins/prepare-custom-leonardo-jupyter-dataproc-image.sh
```diff
@@ -0,0 +1,173 @@
...
+        retry 5 docker pull "${_docker_image:?}"
+    done
+else
```
jdcanas: does this else ever actually get run? I haven't seen the ${:?} syntax before, but from playing around with it locally and from some documentation I read, it seems that in [[ -n ${docker_image_var_names:?} ]] if docker_image_var_names isn't defined or is empty it will exit without executing the else. Am I misunderstanding this?

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/google/HttpGoogleDataprocDAO.scala
```diff
@@ -279,7 +287,7 @@ class HttpGoogleDataprocDAO(appName: String,
         )
     }
 
-    val dataprocProps: Map[String, String] = if (numWorkers.get == 0) {
+    val dataprocProps: Map[String, String] = if (createClusterConfig.machineConfig.numberOfWorkers.get == 0) {
```
jdcanas, line 290: Seems like it could be useful to add getters and abstract some of the guts of createClusterConfig in the future

jdcanas (approved):
Let's get those time savings :)

jenkins/prepare-custom-leonardo-jupyter-dataproc-image.sh
rtitle: Uh.. I'm not sure. Rob M wrote this. :) I feel like this stuff could be simplified if we start reading the docker imgs from an env variable.

jenkins/prepare-custom-leonardo-jupyter-dataproc-image.sh
rtitle: I think you're right -- `:?` probably shouldn't be there in this case. I think it was intended for the script to continue even if there aren't docker images set.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorSupervisor.scala
Qi77Qi: new IJ should be happy without the explicit types

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
```diff
@@ -0,0 +1,89 @@
...
+      for {
+        emailOpt <- IO.fromFuture(IO(googleComputeDAO.getGoogleApiServiceAccount(googleProject)))
+        _ <- emailOpt match {
+          case Some(email) => retryIam(dataprocConfig.leoGoogleProject, email, Set("roles/compute.imageUser"))
+          case None => IO.raiseError(ClusterIamSetupException(googleProject))
+        }
+      } yield ()
+    } else IO.unit
+
+    List(dataprocWorkerIO, computeImageUserIO).sequence[IO, Unit].void
```
Qi77Qi: minor, you can do `sequence_` to avoid the extra `void` call..you can also `parSequence_` to make them run in paralell

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
```diff
@@ -0,0 +1,89 @@
...
+  def generateServiceAccountKey(googleProject: GoogleProject, serviceAccountEmailOpt: Option[WorkbenchEmail]): IO[Option[ServiceAccountKey]] = {
+    // TODO: implement google2 version of GoogleIamDAO
+    serviceAccountEmailOpt.traverse { email =>
+      IO.fromFuture(IO(googleIamDAO.createServiceAccountKey(googleProject, email)))
+    }
+  }
+
+  def removeServiceAccountKey(googleProject: GoogleProject, serviceAccountEmailOpt: Option[WorkbenchEmail], serviceAccountKeyIdOpt: Option[ServiceAccountKeyId]): IO[Unit] = {
+    // TODO: implement google2 version of GoogleIamDAO
+    (serviceAccountEmailOpt, serviceAccountKeyIdOpt).mapN { case (email, keyId) =>
```
Qi77Qi, line 85: use `parMapN`?

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
```diff
@@ -0,0 +1,89 @@
...
+case class ClusterIamSetupException(googleProject: GoogleProject)
+  extends LeoException(s"Error occurred setting up IAM roles in project ${googleProject.value}")
+
+class ClusterHelper(dataprocConfig: DataprocConfig,
+                    gdDAO: GoogleDataprocDAO,
+                    googleComputeDAO: GoogleComputeDAO,
+                    googleIamDAO: GoogleIamDAO)
+                   (implicit val executionContext: ExecutionContext, val system: ActorSystem) extends LazyLogging with Retry {
+
+  implicit val cs = IO.contextShift(executionContext)
```
Qi77Qi: cs should just be passed in implicitly...there's a global one defined in IOApp

jenkins/prepare-custom-leonardo-jupyter-dataproc-image.sh
```diff
@@ -0,0 +1,173 @@
...
+# See: https://cloud.google.com/dataproc/docs/guides/dataproc-images
+#
+
+#
+# Constants and Global Vars
+#
+leonardo_jupyter_docker_image_tag="dev"
+leonardo_jupyter_docker_image="us.gcr.io/broad-dsp-gcr-public/leonardo-jupyter:${leonardo_jupyter_docker_image_tag:?}"
+welder_server_docker_image_tag="latest"
+welder_server_docker_image="us.gcr.io/broad-dsp-gcr-public/welder-server:${welder_server_docker_image_tag:?}"
```
Qi77Qi: there's no welder dev image tag

leonardo-example.conf
```diff
@@ -39,6 +39,11 @@ dataproc {
   # If the subnet is in a shared VPC, this field can accept a partial URI (ex.
   # "projects/[projectId]/regions/us-east1/sub0") in addition to the short-name form "sub0".
   #vpcSubnet = "default"
+
+  # Optional. The custom dataproc image to use. If not specified Leo will choose a public Dataproc
+  # image appropriate for the docker image being installed.
+  # See: https://cloud.google.com/dataproc/docs/guides/dataproc-images
+  #customDataprocImage = "projects/[project_id]/global/images/[image_name]"
```
Qi77Qi, line 46: should this be in reference.conf with a manually generated image?

rtitle: Yeah... I'm also wondering if this should be a required param. It could default to `1.2-deb9` in reference.conf, and be overridden with a custom-image in leonardo.conf.

rtitle:
@jdcanas @Qi77Qi I rebased this and re-tested in a fiab with the images in `broad-dsp-gcr-public`. It works! I think this PR is ready to go, if you want to take another look (most code is unchanged from before). 

Also note this PR will have no effect until we make a fc-develop PR to specify the custom images to use, which we can make once we've generated them in Jenkins.

leonardo-example.conf
Qi77Qi: yea, I was wondering if there's anything else we're waiting on. Like can we use the custom image in reference.conf today? Even if not, can we make the parameter `dataprocImage` and use the default image for now (as apposed to use an optional parameter)?

rtitle: I think there's a reason I didn't want to do this.. let's chat in person. :)

rtitle: Facetalked - i specified the custom image in reference.conf, will watch the tests!

rtitle:
jenkins retest

PR #1036, opened 2019-08-16
[IA-846] Record NR metrics in ClusterMonitorActor

rtitle:
Depends on https://github.com/broadinstitute/workbench-libs/pull/269

This adds a counter and a responseTime metric for each status transition in `ClusterMonitorActor` -- e.g. Creating->Running, Stopped->Stopped, Starting->Running, etc.

Haven't actually tested it with real NR yet, which is why the PR is marked [WIP].

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
@@ -511,7 +511,7 @@ class ClusterMonitorActor(val cluster: Cluster,
   private def recordMetrics(origStatus: ClusterStatus, finalStatus: ClusterStatus): IO[Unit] = {
     for {
       endTime <- IO(System.currentTimeMillis)
-      name = s"ClusterMonitor/${origStatus}_to_${finalStatus}"
+      name = s"ClusterMonitor/${origStatus}->${finalStatus}"
       duration = (endTime - startTime).millis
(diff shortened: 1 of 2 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
```diff
@@ -489,4 +506,14 @@ class ClusterMonitorActor(val cluster: Cluster,
   private def getDbClusterStatus: Future[ClusterStatus] = {
     dbRef.inTransaction { _.clusterQuery.getClusterStatus(cluster.id) } map { _.getOrElse(ClusterStatus.Unknown) }
   }
+
+  private def recordMetrics(origStatus: ClusterStatus, finalStatus: ClusterStatus): IO[Unit] = {
+    for {
+      endTime <- IO(System.currentTimeMillis)
+      name = s"ClusterMonitor/${origStatus}->${finalStatus}"
```
Qi77Qi: not sure if `->` is valid in a metrics name

Qi77Qi (approved):

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
rtitle: Heh maybe not -- will try it out, and maybe replace with `_to_`

PR #1038, opened 2019-08-21
[no ticket] Only monitor tools on Running clusters

rtitle:
This was just something I noticed while testing a different PR. The `ClusterToolMonitor` was flagging clusters in `Creating`, `Stopped`, etc status. It is expected that tools aren't running on those clusters, so we probably shouldn't alert. I changed `ClusterToolMonitor` to only look at `Running` clusters.

In all cases:


Qi77Qi (approved):
is listActiveOnly still used elsewhere? If not, shall we delete it?

rtitle:
I guess it isn't.. I'll delete it.

PR #1044, opened 2019-08-26
IA-1302 IA-1308 test new R kernel image

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-1302

superseded by https://github.com/DataBiosphere/leonardo/pull/1045
\<your comments for this PR go here\>

Have you read [CONTRIBUTING.md](https://github.com/DataBiosphere/leonardo/blob/develop/CONTRIBUTING.md) lately? If not, do that first.

I, the developer opening this PR, do solemnly pinky swear that:

- [ ] I've documented my API changes in Swagger

In all cases:

- [ ] Get a thumbsworth of review and PO signoff if necessary
- [ ] Verify all tests go green
- [ ] Squash and merge; you can delete your branch after this
- [ ] Test this change deployed correctly and works on dev environment after deployment

PR #1045, opened 2019-08-29
[IA-1302] [IA-1308] Cluster IAM fixes affecting preemtibles

rtitle:
This contains the changes in https://github.com/DataBiosphere/leonardo/pull/1044, plus some additional commits.

Basically I changed `ClusterHelper.removeIamRoles` to check for other clusters in the DB owned by the same user/project before removing roles. I think removal of some roles can cause other clusters to break. 

We previously had similar logic [here](https://github.com/DataBiosphere/leonardo/blob/develop/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala#L431) -- I lifted this into `ClusterHelper`.

Not tested yet..

In all cases:

### DIFF
--- a/automation/src/test/resources/reference.conf
+++ b/automation/src/test/resources/reference.conf
@@ -1,0 +1,1 @@
+leonardo.rImageUrl = "us.gcr.io/broad-dsp-gcr-public/terra-jupyter-r:0.0.2"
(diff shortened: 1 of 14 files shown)


rtitle:
@Qi77Qi tests passed on this - what do you think?

Qi77Qi:
just saw what you changed...sounds reasonable to me...thanks for fixing this! 👍

Qi77Qi (approved):

PR #1046, opened 2019-08-29
IA-937 attempt to add automation as submodule

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-937

Note after this PR, you don't need to open another IJ for working on automation tests

In all cases:


Qi77Qi:
jenkins retest

Dockerfile-tests
```diff
@@ -0,0 +1,14 @@
+FROM bigtruedata/sbt
```
rtitle, line 1: Do we need a `Dockerfile-tests` if it's all in the same sbt project? :)

automation-tests-docker-entrypoint.sh
```diff
@@ -5,7 +5,9 @@ echo $SBT_CMD
 
 set -o pipefail
 
-sbt -batch -Djsse.enableSNIExtension=false -Dheadless=true "${SBT_CMD}"
+cp -rf src/test/resources/* automation/src/test/resources
```
rtitle, line 8: I'm curious why this `cp` is needed... though I'm just starting reviewing this PR.

Dockerfile-tests
Qi77Qi: I think we do unless we revamp how we run automation tests in ci, which I think might not be a bad idea, but I think that's another ticket

automation-tests-docker-entrypoint.sh
Qi77Qi: the render config script in firecloud-automation-tests render them under `src/test/resources/*`, but automation being a sub project, we need these config files under `automation/src/test/resources` now

docker/build.sh
```diff
@@ -224,7 +222,7 @@ function docker_cmd()
 #            # pushes the juptyer notebooks and rstudio docker images that go on dataproc clusters
 #            bash ./docker/build-tool.sh --push -i jupyter -r "${NOTEBOOK_REPO}" -t "${DOCKER_TAG}" -b "${GIT_BRANCH}"
 #            bash ./docker/build-tool.sh --push -i rstudio -r "${NOTEBOOK_REPO}" -t "${DOCKER_TAG}" -b "${GIT_BRANCH}"
-
```
rtitle, line 227: Wonder if we should delete the above lines and the `build-tool.sh` script. Maybe in a separate PR.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/ProxyDAO.scala
```diff
@@ -1,4 +1,5 @@
-package org.broadinstitute.dsde.workbench.leonardo.dao
+package org.broadinstitute.dsde.workbench.leonardo
+package dao
```
rtitle, line 2: Is this correct? Several instances below too.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
```diff
@@ -378,9 +378,11 @@ class ClusterMonitorActor(val cluster: Cluster,
     val result = cluster.dataprocInfo.stagingBucket match {
       case Some(stagingBucketName) => {
         for {
+          _ <- IO(logger.info("persisting errors")).unsafeToFuture()
```
rtitle: I like to put `${cluster.projectNameString}` in every log message, if I can -- helps with traceability

rtitle:
I think this looks good - I don't really see any functional changes, just refactoring.

I think I like the `http` and `automation` module names.

Sorry I merged a PR and you now have some conflicts. :( Hopefully not to hard to resolve. This PR will be annoying to leave open for an extended time.

I've just scanned GitHub so far but I'll also check out this branch and poke around running unit tests, etc. I'll thumb if all seems good.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
Qi77Qi: these log lines weren't there before...I'll remove them for now....regarding your comment, I think it's a great point, and I think we should do https://broadworkbench.atlassian.net/browse/IA-906 for adding logging in json format (with clusterName, and googleProject) which can be indexed in kibana

rtitle (approved):
This seems to work for me. :shipit: once conflicts are resolved, etc.

Might want to give the team a heads up too, may need to re-import in Intellij and change sbt commands :)

PR #1048, opened 2019-08-30
[IA-1297] Welder deploy script

rtitle:
JIRA: https://broadworkbench.atlassian.net/browse/IA-1297

This PR:
- Adds `startup.sh` which is deployed as a [GCE startup script](https://cloud.google.com/compute/docs/startupscript) on all clusters
   - Previously this was defined in code but I made it its own file
- Adds a config `deployWelderLabel` which will install welder on all clusters with the given label (we can set it to `saturnAutoCreated` to deploy to Terra clusters)
- Changed startup script installation from cluster _stop_ time to cluster _start_ time, so we can install scripts "just in time"

Unit tests pass, haven't run automation yet. However I've been manually testing in a fiab and it seems to do the right thing. We should do more exhaustive testing of welder functionality.

In all cases:

### DIFF
--- a/src/main/resources/jupyter/startup.sh
+++ b/src/main/resources/jupyter/startup.sh
@@ -18,11 +18,11 @@ export WELDER_SERVER_NAME=$(welderServerName)
 export NOTEBOOKS_DIR=$(notebooksDir)
 export WELDER_ENABLED=$(welderEnabled)
 export DEPLOY_WELDER=$(deployWelder)
+export UPDATE_WELDER=$(updateWelder)
 export WELDER_DOCKER_IMAGE=$(welderDockerImage)
 
 if [ "$DEPLOY_WELDER" == "true" ] ; then
     echo "Deploying Welder on cluster $GOOGLE_PROJECT / $CLUSTER_NAME..."
-    export WELDER_ENABLED="true"
 
     # Run welder-docker-compose
     gcloud auth configure-docker
@@ -33,6 +33,14 @@ if [ "$DEPLOY_WELDER" == "true" ] ; then
 
     # Move existing notebooks to new notebooks dir
     docker exec -it $JUPYTER_SERVER_NAME bash -c "ls -I jupyter.log -I localization.log -I notebooks /home/jupyter-user | xargs -d '\n'  -I file mv file $NOTEBOOKS_DIR"
+fi
+
+if [ "$UPDATE_WELDER" == "true" ] ; then
+    # Run welder-docker-compose
+    gcloud auth configure-docker
+    docker-compose -f /etc/welder-docker-compose.yaml stop
+    docker-compose -f /etc/welder-docker-compose.yaml rm -f
+    docker-compose -f /etc/welder-docker-compose.yaml up -d
 fi
 
 # Start Jupyter
--- a/src/main/resources/jupyter/startup.sh
+++ b/src/main/resources/jupyter/startup.sh
@@ -9,6 +9,7 @@ set -e -x
 # cluster if not already installed.
 ##
 
+# Templated values
 export GOOGLE_PROJECT=$(googleProject)
 export CLUSTER_NAME=$(clusterName)
 export OWNER_EMAIL=$(loginHint)
@@ -21,6 +22,7 @@ export WELDER_DOCKER_IMAGE=$(welderDockerImage)
 
 if [ "$DEPLOY_WELDER" == "true" ] ; then
     echo "Deploying Welder on cluster $GOOGLE_PROJECT / $CLUSTER_NAME..."
+    export WELDER_ENABLED="true"
 
     # Run welder-docker-compose
     gcloud auth configure-docker
--- a/src/main/resources/jupyter/startup.sh
+++ b/src/main/resources/jupyter/startup.sh
@@ -5,7 +5,7 @@ set -e -x
 ##
 # This is a startup script designed to run on Leo-created Dataproc clusters.
 #
-# It starts Jupyter and Welder processes. It also optionally deploys welder on a
+# It starts up Jupyter and Welder processes. It also optionally deploys welder on a
 # cluster if not already installed.
 ##
 
@@ -89,9 +89,11 @@ if [ "$deploy_welder" = true ] ; then
 fi
 
 # Start Jupyter
+echo "Starting Jupyter on cluster $google_project / $cluster_name..."
 docker exec -d jupyter-server /bin/bash -c "/etc/jupyter/scripts/run-jupyter.sh $notebooks_dir || /usr/local/bin/jupyter notebook"
 
 # Start welder, if enabled
 if [ "$welder_enabled" = true ] ; then
+    echo "Starting Welder on cluster $google_project / $cluster_name..."
     docker exec -d welder-server /opt/docker/bin/entrypoint.sh
 fi
--- a/src/main/resources/jupyter/startup.sh
+++ b/src/main/resources/jupyter/startup.sh
@@ -9,91 +9,36 @@ set -e -x
 # cluster if not already installed.
 ##
 
-# Detaults. Assumes welder is enabled and does not need to be deployed.
-welder_enabled=true
-deploy_welder=false
+export GOOGLE_PROJECT=$(googleProject)
+export CLUSTER_NAME=$(clusterName)
+export OWNER_EMAIL=$(loginHint)
+export JUPYTER_SERVER_NAME=$(jupyterServerName)
+export WELDER_SERVER_NAME=$(welderServerName)
+export NOTEBOOKS_DIR=$(notebooksDir)
+export WELDER_ENABLED=$(welderEnabled)
+export DEPLOY_WELDER=$(deployWelder)
+export WELDER_DOCKER_IMAGE=$(welderDockerImage)
 
-while [ "$1" != "" ]; do
-    case $1 in
-        --project)
-            shift
-            google_project=$1
-            ;;
-        --cluster-name)
-            shift
-            cluster_name=$1
-            ;;
-        --owner-email)
-            shift
-            owner_email=$1
-            ;;
-        --enable-welder)
-            welder_enabled=true
-            ;;
-        --deploy-welder)
-            deploy_welder=true
-            welder_enabled=true
-            ;;
-        --welder-docker-image)
-            shift
-            welder_docker_image=$1
-            ;;
-    esac
-    shift
-done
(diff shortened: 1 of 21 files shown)


src/main/resources/jupyter/startup.sh
```diff
@@ -0,0 +1,46 @@
...
+set -e -x
+
+##
+# This is a startup script designed to run on Leo-created Dataproc clusters.
+#
+# It starts up Jupyter and Welder processes. It also optionally deploys welder on a
+# cluster if not already installed.
+##
+
+# Templated values
```
rtitle, line 12: I wasn't a big fan of templating but GCE doesn't give a good way to pass args to startup scripts. They suggest reading parameters from instance metadata but that seems unwieldy.

src/main/resources/jupyter/startup.sh
```diff
@@ -0,0 +1,46 @@
...
+
+    # Run welder-docker-compose
+    gcloud auth configure-docker
+    docker-compose -f /etc/welder-docker-compose.yaml up -d
+
+    # Set EVs inside Jupyter container necessary for welder
+    docker exec -it $JUPYTER_SERVER_NAME bash -c "echo $'export WELDER_ENABLED=true\nexport NOTEBOOKS_DIR=$NOTEBOOKS_DIR' >> /home/jupyter-user/.bashrc"
+
+    # Move existing notebooks to new notebooks dir
+    docker exec -it $JUPYTER_SERVER_NAME bash -c "ls -I jupyter.log -I localization.log -I notebooks /home/jupyter-user | xargs -d '\n'  -I file mv file $NOTEBOOKS_DIR"
```
rtitle: This `mv`s everything under `~jupyter-user/` to `~jupyter-user/notebooks/`. I didn't want to `cp` because we could potentially run out of disk (depending what data files the user has on their cluster). It's a little scary `mv`ing users' files -- we should definitely test to make sure there are no edge cases. This could also break notebook code that uses absolute paths, but I'm not sure we can avoid that.

src/main/resources/reference.conf
```diff
@@ -28,12 +28,15 @@ dataproc {
   #vpcSubnet = "default"
   customDataprocImage = "projects/broad-dsp-gcr-public/global/images/custom-leo-image-08a4ad8"   # updated 8/26
   dataprocDockerImage = "us.gcr.io/broad-dsp-gcr-public/leonardo-jupyter:5c51ce6935da"           # updated 8/26
+  # Set to deploy welder to clusters with the given label
+  #deployWelderLabel = "saturnVersion"
```
rtitle: Commented for now, so we should be able to deploy this prior to deploying the Terra UI changes, and just enable this config when we're ready.

Qi77Qi: what does this label mean?

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -1116,4 +1101,39 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
...
+    !cluster.welderEnabled && dataprocConfig.deployWelderLabel.exists(cluster.labels.contains)
+  }
+
+  private def enableWelder(cluster: Cluster): IO[Cluster] = {
+    for {
+      _ <- IO(logger.info(s"Deploying welder on cluster ${cluster.projectNameString}"))
+      _ <- Metrics.newRelic.incrementCounterIO("welderDeployed")
+      now <- IO(Instant.now)
+      welderImage = ClusterImage(Welder, dataprocConfig.welderDockerImage, now)
+      _ <- dbRef.inTransactionIO { _.clusterQuery.enableWedler(cluster.id, ClusterImage(Welder, dataprocConfig.welderDockerImage, Instant.now)) }
```
Qi77Qi: should `now` be reused here instead of calling `Instant.now` directly?

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -1116,4 +1101,39 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
...
+      newCluster = cluster.copy(welderEnabled = true, clusterImages = cluster.clusterImages + welderImage)
```
Qi77Qi: will we accidentally add 2 welder images?

src/main/resources/jupyter/startup.sh
```diff
@@ -0,0 +1,46 @@
...
+
+# Start Jupyter
+echo "Starting Jupyter on cluster $GOOGLE_PROJECT / $CLUSTER_NAME..."
+docker exec -d $JUPYTER_SERVER_NAME /bin/bash -c "/etc/jupyter/scripts/run-jupyter.sh $NOTEBOOKS_DIR || /usr/local/bin/jupyter notebook"
+
+# Start welder, if enabled
+if [ "$WELDER_ENABLED" == "true" ] ; then
+    echo "Starting Welder on cluster $GOOGLE_PROJECT / $CLUSTER_NAME..."
+    docker exec -d $WELDER_SERVER_NAME /opt/docker/bin/entrypoint.sh
+fi
```
Qi77Qi, line 51: is it guaranteed that this will only happen one time or it'd be okay to run this script multiple times on a cluster?

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -1116,4 +1101,39 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
...
 
+  private def shouldDeployWelder(cluster: Cluster): Boolean = {
+    !cluster.welderEnabled && dataprocConfig.deployWelderLabel.exists(cluster.labels.contains)
+  }
+
+  private def enableWelder(cluster: Cluster): IO[Cluster] = {
+    for {
+      _ <- IO(logger.info(s"Deploying welder on cluster ${cluster.projectNameString}"))
+      _ <- Metrics.newRelic.incrementCounterIO("welderDeployed")
+      now <- IO(Instant.now)
```
Qi77Qi: I might add `timer: Timer[IO]` as implicit val for this class, and use `timer` here to get `now`

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -1116,4 +1101,39 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
     Set(welderImageOpt, Some(jupyterImage), rstudioImageOpt).flatten
   }
 
+  private def shouldDeployWelder(cluster: Cluster): Boolean = {
+    !cluster.welderEnabled && dataprocConfig.deployWelderLabel.exists(cluster.labels.contains)
+  }
+
+  private def enableWelder(cluster: Cluster): IO[Cluster] = {
+    for {
+      _ <- IO(logger.info(s"Deploying welder on cluster ${cluster.projectNameString}"))
```
Qi77Qi: shall we log cluster name as well

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/DbReference.scala
```diff
@@ -64,6 +65,10 @@ case class DbReference(private val dbConfig: DatabaseConfig[JdbcProfile])(implic
     import dataAccess.profile.api._
     database.run(f(dataAccess).transactionally.withTransactionIsolation(isolationLevel))
   }
+
+  def inTransactionIO[T](f: (DataAccess) => DBIO[T], isolationLevel: TransactionIsolation = TransactionIsolation.RepeatableRead): IO[T] = {
```
Qi77Qi, line 69: 👍 for using `IO` here :)

src/main/resources/jupyter/startup.sh
rtitle: This script runs on every cluster startup. It's intended for `DEPLOY_WELDER` to only be enabled one time -- after which `WELDER_ENABLED` should be true. We should test this (maybe I will add a unit test).

src/main/resources/reference.conf
rtitle: Terra UI sets this label on all clusters (I think the current version is `6`). I'm not sure when they increase it. I thought this was a good way to distinguish Terra clusters.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
rtitle: `projectNameString` logs both project name and cluster name.

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
rtitle: 👍 good catch

src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
rtitle: It's a `Set` so I don't think so... plus this method is only called for clusters where welder is disabled.

rtitle: Actually changed this to do a DB `upsert` and `filter` on the current set to ensure we don't duplicate it.

src/main/resources/reference.conf
Qi77Qi: just want to understand how the label works. How do we get the label on our end?

rtitle: Ah - label is a Leo concept. They're stored in the DB in the `CLUSTER_LABEL` table. Clients can optionally set labels on cluster creation, and use labels to filter `listClusters` results. I happen to know Terra-UI sets this label. :)

rtitle:
Made a couple more changes based on PR feedback; redeploying to a FIAB for further testing.

src/main/resources/reference.conf
Qi77Qi: Gotcha! That makes sense

Qi77Qi (approved):

rtitle:
jenkins retest

rtitle:
jenkins retest

PR #1056, opened 2019-09-10
[IA-1304] Automation test fix for lockedBy hash 

akarukappadath:

In all cases:


automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookDataSyncingSpec.scala
```diff
@@ -59,13 +61,16 @@ class NotebookDataSyncingSpec extends ClusterFixtureSpec with NotebookTestUtils
                 remoteContentSize shouldBe localContentSize
               }
 
+              val hashedUser = Some(String.format("%064x", new BigInteger(1, MessageDigest.getInstance("SHA-256").digest((gcsPath.bucketName + ":" + clusterFixture.cluster.creator).getBytes("UTF-8")))))
```
jdcanas: I'd like to see this split into two lines with the `MessageDigest....` as its own val with a meaningful name. I find it a bit hard to read as is.

jdcanas (approved):
:shipit:

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookDataSyncingSpec.scala
```diff
@@ -59,13 +61,16 @@ class NotebookDataSyncingSpec extends ClusterFixtureSpec with NotebookTestUtils
...
+
               eventually(timeout(Span(4, Minutes)), interval(Span(30, Seconds))) {
                 val gcsLockedBy: Option[String] = getLockedBy(gcsPath.bucketName, GcsBlobName(gcsPath.objectName.value)).unsafeRunSync()
                 val welderLockedBy: Option[String] = Welder.getMetadata(clusterFixture.cluster, gcsPath, isEditMode).lastLockedBy
 
                 gcsLockedBy should not be None
                 welderLockedBy should not be None
                 gcsLockedBy shouldBe welderLockedBy
+                gcsLockedBy shouldBe hashedUser
```
Qi77Qi: what is `welderLockedBy` in the line before this?

akarukappadath: After welder updates the `lockedBy` metadata, we're reading the value from the bucket twice - once through welder (`welderLockedBy`) and once directly through gcs (`gcsLockedBy`). This basically checks to make sure that welder can correctly read the `lockedBy` metadata of a notebook, but we weren't checking to see if welder is correctly locking it in the first place, which is what the new check does.

jdcanas (approved):
👍

project/Dependencies.scala
```diff
@@ -162,6 +162,7 @@ object Dependencies {
     "com.fasterxml.jackson.core" % "jackson-core"         % jacksonV,
     "com.fasterxml.jackson.module" % ("jackson-module-scala_" + scalaV) % jacksonV,
     "ch.qos.logback"  % "logback-classic" % "1.2.3"  % "test",
+    "com.google.guava" % "guava" % "28.0-jre",
```
rtitle, line 165: Do we need this? I don't see guava imported.

rtitle (approved):
LGTM

project/Dependencies.scala
akarukappadath: It looks like guava was being brought in by another package - this fixes the error we were seeing when trying to run autotests locally

PR #1061, opened 2019-09-16
[IA-1361] Add trace ID to cluster creation logs

kyuksel:
This will help with performance testing errors and more where we may see an error response on Jenkins output log and it's difficult to see the relevant Leo logs on Kibana, etc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,11 +48,21 @@ Ensure docker is running. Spin up MySQL locally:
 ```
 $ ./docker/run-mysql.sh start leonardo  
 ```
-Build Leonardo and run tests.
+Build Leonardo and run all unit tests.
 ```
 export SBT_OPTS="-Xmx2G -Xms1G -Dmysql.host=localhost -Dmysql.port=3311"
 sbt clean compile "project http" test
 ```
+You can also run a particular test suite, e.g.
+```
+sbt "project http" "testOnly *LeoAuthProviderHelperSpec"
+```
+or a particular test within a suite, e.g.
+```
+sbt "project http" "testOnly *LeoAuthProviderHelperSpec -- -z map"
+```
+where `map` is a substring within the test name.
+
 Once you're done, tear down MySQL.
 ```
 ./docker/run-mysql.sh stop leonardo
--- a/README.md
+++ b/README.md
@@ -55,11 +55,11 @@ sbt clean compile "project http" test
 ```
 You can also run a particular test suite, e.g.
 ```
-sbt "project http" "testOnly *LeoAuthProviderHelperSpec"
+sbt "testOnly *LeoAuthProviderHelperSpec"
 ```
 or a particular test within a suite, e.g.
 ```
-sbt "project http" "testOnly *LeoAuthProviderHelperSpec -- -z map"
+sbt "testOnly *LeoAuthProviderHelperSpec -- -z map"
 ```
 where `map` is a substring within the test name.
 
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/LeoAuthProviderHelper.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/LeoAuthProviderHelper.scala
@@ -8,7 +8,7 @@ import java.util.UUID.randomUUID
 import net.ceedubs.ficus.Ficus._
 import org.broadinstitute.dsde.workbench.leonardo.model._
 import org.broadinstitute.dsde.workbench.leonardo.model.google.ClusterName
-import org.broadinstitute.dsde.workbench.model.{TraceId, UserInfo, WorkbenchEmail}
+import org.broadinstitute.dsde.workbench.model.{UserInfo, WorkbenchEmail}
 import org.broadinstitute.dsde.workbench.model.google.GoogleProject
 import org.broadinstitute.dsde.workbench.util.FutureSupport
 
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/LeoAuthProviderHelper.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/LeoAuthProviderHelper.scala
@@ -52,7 +52,7 @@ class LeoAuthProviderHelper(wrappedAuthProvider: LeoAuthProvider, authConfig: Co
         Future.failed(AuthProviderException(wrappedClassName, isTimeout = true))
       case NonFatal(e) =>
         val wrappedClassName = wrappedAuthProvider.getClass.getSimpleName
-        logger.error(s"Auth provider $wrappedClassName throw an exception", e)
+        logger.error(s"Auth provider $wrappedClassName threw an exception", e)
         Future.failed(AuthProviderException(wrappedClassName))
     }
 
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/LeoAuthProviderHelper.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/LeoAuthProviderHelper.scala
@@ -45,19 +45,17 @@ class LeoAuthProviderHelper(wrappedAuthProvider: LeoAuthProvider, authConfig: Co
   private lazy val providerTimeout = authConfig.as[FiniteDuration]("providerTimeout")
   private implicit val scheduler = system.scheduler
 
-  private def safeCall[T](future: => Future[T])(implicit executionContext: ExecutionContext): Future[T] = {
+  private def safeCall[T](traceId: TraceId = TraceId(java.util.UUID.randomUUID))(future: => Future[T])(implicit executionContext: ExecutionContext): Future[T] = {
     val exceptionHandler: PartialFunction[Throwable, Future[Nothing]] = {
       case e: LeoException => Future.failed(e)
       case te: TimeoutException =>
         val wrappedClassName = wrappedAuthProvider.getClass.getSimpleName
-        val randomTraceId = TraceId(java.util.UUID.randomUUID)
-        logger.error(s"[$randomTraceId] Auth provider $wrappedClassName timed out after $providerTimeout", te)
-        Future.failed(AuthProviderException(wrappedClassName, isTimeout = true, traceId = randomTraceId))
+        logger.error(s"[$traceId] Auth provider $wrappedClassName timed out after $providerTimeout", te)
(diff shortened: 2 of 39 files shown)


http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/service/AuthProviderSpec.scala
```diff
@@ -107,7 +107,7 @@ class AuthProviderSpec extends FreeSpec with ScalatestRouteTest with Matchers wi
       leo.deleteCluster(userInfo, project, cluster1Name).futureValue
 
       //verify we correctly notified the auth provider
-      verify(spyProvider).notifyClusterCreated(any[ClusterInternalId], any[WorkbenchEmail], any[GoogleProject], any[ClusterName])(any[ExecutionContext])
+      verify(spyProvider).notifyClusterCreated(any[ClusterInternalId], any[WorkbenchEmail], any[GoogleProject], any[ClusterName], any[TraceId])(any[ExecutionContext])
```
kyuksel: The line above is where `NullPointerException` is coming from.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -219,6 +222,7 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
                                     googleProject: GoogleProject,
                                     clusterName: ClusterName,
                                     clusterRequest: ClusterRequest): Future[Cluster] = {
+    val traceId = TraceId(randomUUID)
```
Qi77Qi, line 225: I'd prefer this is being passed in so that one `traceId` can correlate as many logs as possible

project/Settings.scala
```diff
@@ -35,7 +35,7 @@ object Settings {
     "-language:implicitConversions",     // Allow definition of implicit functions called views
     "-unchecked",                        // Enable additional warnings where generated code depends on assumptions.
     "-Xcheckinit",                       // Wrap field accessors to throw an exception on uninitialized access.
-    "-Xfatal-warnings",                  // Fail the compilation if there are any warnings.
+//    "-Xfatal-warnings",                  // Fail the compilation if there are any warnings.
```
Qi77Qi: this needs to be turned on before this can be merged

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/LeoAuthProviderHelper.scala
```diff
@@ -4,19 +4,22 @@ import akka.actor.ActorSystem
...
 import scala.concurrent.duration._
 import scala.concurrent.{ExecutionContext, Future, TimeoutException}
 import scala.util.control.NonFatal
 
-case class AuthProviderException(authProviderClassName: String, isTimeout: Boolean = false)
-  extends LeoException(s"Call to $authProviderClassName auth provider ${if (isTimeout) "timed out" else "failed"}", StatusCodes.InternalServerError)
+case class AuthProviderException(authProviderClassName: String,
+                                 isTimeout: Boolean = false,
+                                 traceId: TraceId)
+  extends LeoException(s"[$traceId] Call to $authProviderClassName auth provider ${if (isTimeout) "timed out" else "failed"}", StatusCodes.InternalServerError)
```
rtitle: Do we want to make the status code in `AuthProviderException` match the status code from Sam (instead of `InternalServerError`)?

README.md
```diff
@@ -48,11 +48,21 @@ Ensure docker is running. Spin up MySQL locally:
...
 ```
-Build Leonardo and run tests.
+Build Leonardo and run all unit tests.
 ```
 export SBT_OPTS="-Xmx2G -Xms1G -Dmysql.host=localhost -Dmysql.port=3311"
 sbt clean compile "project http" test
 ```
+You can also run a particular test suite, e.g.
+```
+sbt "project http" "testOnly *LeoAuthProviderHelperSpec"
```
Qi77Qi: `project http` is not necessary needed. Doesn't hurt to have it either

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
rtitle: I think `processClusterCreationRequest` is the entry point for the `v2` createCluster request. `createCluster` and `internalCreateCluster` are for `v1` and deprecated. I'm actually removing them in my branch. :)

project/Settings.scala
rtitle: +1

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/service/AuthProviderSpec.scala
rtitle: I have a theory about what's going on. The key is in the error message.
```
[info] This message may appear after an NullPointerException if the last matcher is returning an object 
[info] like any() but the stubbed method signature expect a primitive argument, in this case,
[info] use primitive alternatives.
[info]     when(mock.get(any())); // bad use, will raise NPE
[info]     when(mock.get(anyInt())); // correct usage use
```

In fact, the last argument is `any[TraceId]`, an object. [TraceId](https://github.com/broadinstitute/workbench-libs/blob/develop/model/src/main/scala/org/broadinstitute/dsde/workbench/model/models.scala#L6) is a Scala value object, meaning at runtime it is a primitive. So this message applies -- the method expects a primitive but our matcher returns an object.

This seems to be an example where Mockito (a Java library) doesn't play nice with Scala language  features (values objects in this case).

I tried a local change where I made my own `TraceId` in `LeonardoModel.scala` as:
```
case class TraceId(uuid: UUID)
```
E.g. it's _not_ a value object. When I use this class, mockito is happy and the test passes. I'm happy to PR that change into this branch.

rtitle: Made a PR, feel free to merge or not: https://github.com/DataBiosphere/leonardo/pull/1062

Qi77Qi: ah...I would never thought this is the issue since the code executes just fine even though matcher actually failed.

It seems like `TraceId(any[UUID])` would work fine as matcher, so I think we should use this instead of creating a separate `TraceId`

rtitle: ^ I didn't try that, it's worth a try. Agreed that would be preferable if it works.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/LeoAuthProviderHelper.scala
kyuksel: @Qi77Qi and I had talked faces about this and we thought it's still appropriate to return `500` because the error is not due to the user's (in the perf test case, cluster creation) request and there isn't an action they can take. If anything, it feels that the error message isn't helpful to the user and I wonder if we should change it to something more high level? 🤔

Qi77Qi: I think the other reason is swagger generated client doesn't really give you the status code from sam other than throwing an exception. (If this is not true, it might still be useful to give user 404 etc if it's 404 from sam)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
kyuksel: The names are confusing indeed. Perhaps we can rename `processClusterCreationRequest` as `createCluster` once you remove the older version.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/LeoAuthProviderHelper.scala
kyuksel: Yea, I think there's that too.

README.md
kyuksel: Ah...it's interesting that I get 
```
[error] (automation / Test / test) sbt.TestsFailedException: Tests unsuccessful
```
with 
```
sbt clean compile test
```
i.e. with `"project http"` omitted, even though all tests pass. But I don't get that with the `testOnly` command line.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/WhitelistAuthProvider.scala
```diff
@@ -72,7 +73,7 @@ class WhitelistAuthProvider(config: Config, serviceAccountProvider: ServiceAccou
     * @param clusterName   The user-provided name of the Dataproc cluster
     * @return A Future that will complete when the auth provider has finished doing its business.
     */
-  def notifyClusterCreated(internalId: ClusterInternalId, creatorEmail: WorkbenchEmail, googleProject: GoogleProject, clusterName: ClusterName)(implicit executionContext: ExecutionContext): Future[Unit] = Future.successful(())
+  def notifyClusterCreated(internalId: ClusterInternalId, creatorEmail: WorkbenchEmail, googleProject: GoogleProject, clusterName: ClusterName, traceId: TraceId = TraceId(randomUUID))(implicit executionContext: ExecutionContext): Future[Unit] = Future.successful(())
```
Qi77Qi: fyi you can do `Future.unit`

Qi77Qi (approved):

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/LeoAuthProviderHelper.scala
rtitle: OK, that makes sense.

That's a good point about the error message - could we change it to something more user friendly than `SamAuthProvider failed`?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
rtitle: Good idea

rtitle (approved):
This looks good - my one comment/question is around changing the LeoAuthException error message.

PR #1062, opened 2019-09-16
Use Leo TraceId

rtitle:
Note the target of this PR is https://github.com/DataBiosphere/leonardo/pull/1061/. See https://github.com/DataBiosphere/leonardo/pull/1061/#discussion_r324776435 for context. This makes Mockito happy by matching a non-value object. Feel free to use or not.

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
@@ -384,6 +384,8 @@ object ExtensionType extends Enum[ExtensionType] {
   case object LabExtension extends ExtensionType
 }
 
+case class TraceId(uuid: UUID)
+
 object LeonardoJsonSupport extends SprayJsonSupport with DefaultJsonProtocol {
   implicit object URLFormat extends JsonFormat[URL] {
     def write(obj: URL) = JsString(obj.toString)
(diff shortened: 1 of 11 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
```diff
@@ -384,6 +384,8 @@ object ExtensionType extends Enum[ExtensionType] {
   case object LabExtension extends ExtensionType
 }
 
+case class TraceId(uuid: UUID)
```
Qi77Qi, line 387: just curious, why does using a separate data type fixes the null pointer exception?

rtitle: I think because it's not a value object

rtitle: Compare with https://github.com/broadinstitute/workbench-libs/blob/develop/model/src/main/scala/org/broadinstitute/dsde/workbench/model/models.scala#L6

Qi77Qi: sorry forgot to comment back here...I saw the other comment :)

rtitle:
I see this was merged... @Qi77Qi came up with a better solution, probably want to back this out :)

PR #1064, opened 2019-09-17
[IA-1363] Cluster creation failing for new projects

rtitle:
See https://broadworkbench.atlassian.net/browse/IA-1363.

This makes it try to use `service-[project-number]@dataproc-accounts.iam.gserviceaccount.com` if it exists; falling back to `[project-number]@cloudservices.gserviceaccount.com` for granting IAM roles.

I will test this on a fiab.

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/google/GoogleComputeDAO.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/google/GoogleComputeDAO.scala
@@ -23,8 +23,6 @@ trait GoogleComputeDAO {
 
   def getComputeEngineDefaultServiceAccount(googleProject: GoogleProject): Future[Option[WorkbenchEmail]]
 
-  def getGoogleApiServiceAccount(googleProject: GoogleProject): Future[Option[WorkbenchEmail]]
-
   def getProjectNumber(googleProject: GoogleProject): Future[Option[Long]]
 
   def setMachineType(instanceKey: InstanceKey, newMachineType: MachineType): Future[Unit]
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
@@ -61,7 +61,7 @@ class ClusterHelper(dbRef: DbReference,
     } getOrElse IO.unit
 
     // TODO: replace this logic with a group based approach so we don't have to manipulate IAM directly in the image project.
-    // See <jira>
+    // See https://broadworkbench.atlassian.net/browse/IA-1364
     //
     // Add the Compute Image User role in the image project to the Google API service account.
     // This is needed in order to use a custom dataproc VM image.
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
@@ -89,30 +89,21 @@ class ClusterHelper(dbRef: DbReference,
     List(dataprocWorkerIO, computeImageUserIO).parSequence_
   }
 
-  // Service account email format documented in:
-  // https://cloud.google.com/iam/docs/service-accounts#google-managed_service_accounts
-  private[leonardo] def googleApiServiceAccount: Long => WorkbenchEmail = projectNumber =>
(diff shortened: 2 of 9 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/google/GoogleComputeDAO.scala
```diff
@@ -23,8 +23,6 @@ trait GoogleComputeDAO {
 
   def getComputeEngineDefaultServiceAccount(googleProject: GoogleProject): Future[Option[WorkbenchEmail]]
 
-  def getGoogleApiServiceAccount(googleProject: GoogleProject): Future[Option[WorkbenchEmail]]
```
rtitle, line 26: I figured this didn't really belong in `GoogleComputeDAO` -- I moved this logic to `ClusterHelper`

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/CommonTestData.scala
```diff
@@ -198,6 +198,12 @@ trait CommonTestData{ this: ScalaFutures =>
   protected def modifyInstanceKey(instanceKey: InstanceKey): InstanceKey = {
     instanceKey.copy(name = InstanceName(instanceKey.name.value + "_2"))
   }
+
+  protected class LeoMockGoogleIamDAO extends MockGoogleIamDAO {
```
rtitle: The wb-libs `MockGoogleIamDAO` didn't have quite the right behavior for `findServiceAccount`, so I extended it here.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
```diff
@@ -88,6 +89,32 @@ class ClusterHelper(dbRef: DbReference,
...
+    val findDataprocServiceAccount = findSA compose dataprocServiceAccount
+
+    (for {
+      number <- OptionT(IO.fromFuture(IO(googleComputeDAO.getProjectNumber(googleProject))))
+      // use dataproc service account if it exists; otherwise fall back to the API service account
+      precedence = (findDataprocServiceAccount, findGoogleApiServiceAccount).mapN(_ orElse _)
+      sa <- precedence(number)
+    } yield sa).value
+  }
+
```
Qi77Qi, line 104: this code is a bit hard to read without IDE...I kinda prefer something like 
```
    for {
      projectNumber <- OptionT.liftF(IO.fromFuture(IO(googleComputeDAO.getProjectNumber(googleProject))))
      apiSA = WorkbenchEmail(s"$projectNumber@dataproc-accounts.iam.gserviceaccount.com")
      dataprocSA = WorkbenchEmail(s"service-$projectNumber@dataproc-accounts.iam.gserviceaccount.com")
      sa <- findSA(dataprocSA) orElse findSA(apiSA)
    } yield sa
```

rtitle: updated

kyuksel (approved):
Changes look good 👍 I couldn't be sure though, where we're testing whether the piece of the logic

```scala
val transformed = for {
      projectNumber <- OptionT(IO.fromFuture(IO(googleComputeDAO.getProjectNumber(googleProject))))
      apiSA = WorkbenchEmail(s"$projectNumber@cloudservices.gserviceaccount.com")
      dataprocSA = WorkbenchEmail(s"service-$projectNumber@dataproc-accounts.iam.gserviceaccount.com")
      sa <- findSA(dataprocSA) orElse findSA(apiSA)
    } yield sa
```
that picks `apiSA` vs. `dataprocSA` works the right way.

rtitle:
Thanks @kyuksel. Good point, we're probably not verifying that in a unit test because we're using mock IAM DAOs which always return `Some(sa)`. I could see about adding a unit test for this (may be in a separate PR).

Right now I am verifying the functionality on a fiab.

Qi77Qi (approved):

rtitle:
@Qi77Qi @kyuksel the logic with `findSA(..) orElse findSA(..)` was not really working. I changed the logic to just add IAM entries for _both_ service accounts. It's more IAM we need to manage but this should hopefully go away when we switch to a group-based approach.

I'm retesting in a fiab.

rtitle:
jenkins retest

rtitle:
jenkins retest

rtitle:
This seems to work based on fiab testing

rtitle:
`java.lang.OutOfMemoryError: Metaspace` :(

jenkins retest

Issue #167, opened 2018-01-29
Store deleted-by user in Clusters table and pass that value to authProvider.notifyClusterDeleted()


jmthibault79:
I got lazy in #139 and didn't pass the actual user deleting the cluster to the AuthProvider because it wasn't available to the ClusterMonitorActor.  This isn't an immediate concern because the only extant implementation (SamAuthProvider) doesn't use this information, but we should come back to this because it means fulfilling the contract we're making in the interface.

We do know the user deleting the cluster at the time of the API call, so we can store that in the Clusters table.  The ClusterMonitorActor can then query for that information when it needs to call AuthProvider.notifyClusterDeleted()

helgridly:
I also think that keeping a record of both creator and deleter (deletor?) in the db is independently helpful, so 👍 on this

PR #139, opened 2018-01-25
GAWB-3108 Don't Delete Sam Cluster Resources

jmthibault79:
\<your comments for this PR go here\>

Have you read [CONTRIBUTING.md](../CONTRIBUTING.md) lately? If not, do that first.

I, the developer opening this PR, do solemnly pinky swear that:

- [ ] I've documented my API changes in Swagger

In all cases:

- [ ] Get two thumbsworth of review and PO signoff if necessary
- [ ] Verify all tests go green
- [ ] Squash and merge; you can delete your branch after this
- [ ] Test this change deployed correctly and works on dev environment after deployment

Issue #137, opened 2018-01-25
Leo/Sam auth: SamAuthProvider should not delete the cluster resource in Sam


rtitle:
Currently `LeoService.internalDeleteCluster()` calls `SamAuthProvider.notifyClusterDeleted()` calls `SwaggerSamClient.deleteNotebookClusterResource()`.  However when a cluster is deleted it is NOT gone yet -- the ClusterMonitorActor needs to monitor its state from Deleting to Deleted, and it should still be retrievable via Leo get/list APIs. Therefore I don't think Leo should call Sam `deleteNotebookClusterResource()` until the `ClusterMonitorActor` confirms deletion.

I found this while using the UI: when I delete a cluster it disappears from the UI immediately, because the "list clusters" endpoint no longer returns it because it is no longer permissioned in Sam (though it is still Deleting in Leo).

jmthibault79:
When Leonardo transitions a cluster from `Deleting` to `Deleted` then we *do* want to call `SwaggerSamClient.deleteNotebookClusterResource()`.

The current PR does not cover this, so I will add it.

PR #1069, opened 2019-09-19
[IA-1385] Don't prematurely remove Sam resources

rtitle:
https://broadworkbench.atlassian.net/browse/IA-1385

Hoping this fixes the issue where `Error`'d clusters are un-gettable and un-deletable.

Also snuck in another fix to the IAM stuff based on recent revelations from Google.

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
@@ -185,7 +185,7 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
         val internalId = ClusterInternalId(UUID.randomUUID().toString)
         val augmentedClusterRequest = augmentClusterRequest(serviceAccountInfo, googleProject, clusterName, userEmail, clusterRequest)
         val clusterImages = processClusterImages(clusterRequest)
-        val clusterFuture = for {
+        for {
           // Metrics
           _ <- Metrics.newRelic.incrementCounterFuture("numberOfCreateClusterRequests")
           _ <- if (clusterRequest.enableWelder.getOrElse(false)) Metrics.newRelic.incrementCounterFuture("numberOfWelderEnabledCreateClusterRequests") else Future.unit
@@ -204,15 +204,6 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
         } yield {
           savedCluster
         }
-
-        // If cluster creation failed on the Google side, createGoogleCluster removes resources in Google.
-        // We also need to notify our auth provider that the cluster has been deleted.
-        clusterFuture.andThen {
-          // Don't wait for this future
-          case Failure(_) => authProvider.notifyClusterDeleted(internalId, userEmail, userEmail, googleProject, clusterName)
-        }
-
-        clusterFuture
     }
   }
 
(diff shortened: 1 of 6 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
```diff
@@ -78,9 +78,17 @@ class ClusterHelper(dbRef: DbReference,
             IO.unit
           } else {
             for {
-              emails <- getDataprocServiceAccounts(googleProject)
-              _ <- if (emails.isEmpty) IO.raiseError(ClusterIamSetupException(imageProject))
-                   else emails.traverse(e => retryIam(imageProject, e, Set("roles/compute.imageUser")))
+              projectNumber <- IO.fromFuture(IO(googleComputeDAO.getProjectNumber(googleProject))).map(_.getOrElse(throw ClusterIamSetupException(imageProject)))
+              roles = Set("roles/compute.imageUser")
+
+              // The Dataproc SA is used to retrieve the image. However projects created prior to 2016
```
rtitle, line 85: Last change to this code, I promise. Instead of adding _both_ service accounts, try to add the dataproc one first, and add the API one if that fails. Google confirmed that only projects created prior to 2016 do _not_ have a dataproc SA, which matches what I have seen.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -204,15 +204,6 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
         } yield {
           savedCluster
         }
-
-        // If cluster creation failed on the Google side, createGoogleCluster removes resources in Google.
```
rtitle, line 208: I think this logic was just a mistake. We should treat an `Error` cluster as an active Sam resource, since you can still perform operations on it. The only time we call `notifyClusterDeleted` now is when the cluster transitions to `Deleted` state in `ClusterMonitorActor`.

jmthibault79: I did a little spelunking to help figure out why we had this logic.  I wasn't successful, but maybe these links will be educational: #137 #139 #167

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
```diff
@@ -78,9 +78,17 @@ class ClusterHelper(dbRef: DbReference,
             IO.unit
           } else {
             for {
-              emails <- getDataprocServiceAccounts(googleProject)
-              _ <- if (emails.isEmpty) IO.raiseError(ClusterIamSetupException(imageProject))
-                   else emails.traverse(e => retryIam(imageProject, e, Set("roles/compute.imageUser")))
+              projectNumber <- IO.fromFuture(IO(googleComputeDAO.getProjectNumber(googleProject))).map(_.getOrElse(throw ClusterIamSetupException(imageProject)))
```
Qi77Qi: minor, I think it's better to `.flatMap(_.fold(IO.pure)(_ => IO.raiseError(ClusterIamSetupException(imageProject))))` to avoid throwing the exception and only to be caught somewhere else in the app

Qi77Qi (approved):

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
rtitle: Thanks, that looks better

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
rtitle: Thanks, that's useful. I think we were on the right track but missed this subtlety.

rtitle:
jenkins retest

rtitle:
Still getting 
```
java.lang.OutOfMemoryError: Metaspace
```
jenkins retest

kyuksel (commented):
Sorry for the late review. LGTM 👍

PR #1070, opened 2019-09-20
[IA-1323] Jc creation timeout

jdcanas:
Of particular interest is the unit test - should I be verifying more?

In all cases:

### DIFF
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -104,8 +104,8 @@ monitor {
   recreateCluster = true
   creatingTimeLimit = 1 hour
   startingTimeLimit = 20 minutes
-  stoppingTimeLimit = 10 minutes
-  deletingTimeLimit = 20 minutes
+  stoppingTimeLimit = 1 hour
+  deletingTimeLimit = 1 hour
   updatingTimeLimit = 20 minutes
 }
 
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -2,7 +2,6 @@
 # Google Cloud dataproc configuration
 dataproc {
   applicationName = "firecloud:leonardo"
-  defaultExecutionTimeout = 30 minutes
   jupyterServerName = "jupyter-server"
   rstudioServerName = "rstudio-server"
   welderServerName = "welder-server"
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -103,6 +103,7 @@ monitor {
   pollPeriod = 15 seconds
   maxRetries = -1  # means retry forever
   recreateCluster = true
+  creationTimeLimit = 20 minutes
 }
 
 # akka values are not specified here because they are only picked up in the leonardo.conf
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -103,7 +103,11 @@ monitor {
   pollPeriod = 15 seconds
   maxRetries = -1  # means retry forever
   recreateCluster = true
-  creationTimeLimit = 20 minutes
+  creatingTimeLimit = 1 hour
+  startingTimeLimit = 20 minutes
+  stoppingTimeLimit = 10 minutes
+  deletingTimeLimit = 20 minutes
+  updatingTimeLimit = 20 minutes
 }
 
 # akka values are not specified here because they are only picked up in the leonardo.conf
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/MonitorConfig.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/MonitorConfig.scala
@@ -7,4 +7,7 @@ import scala.concurrent.duration.FiniteDuration
 /**
   * Created by rtitle on 9/6/17.
   */
-case class MonitorConfig(pollPeriod: FiniteDuration, maxRetries: Int = -1, recreateCluster: Boolean = true, monitorStatusTimeouts: Map[ClusterStatus, FiniteDuration])
+case class MonitorConfig(pollPeriod: FiniteDuration,
+                         maxRetries: Int = -1,
+                         recreateCluster: Boolean = true,
+                         monitorStatusTimeouts: Map[ClusterStatus, FiniteDuration])
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/MonitorConfig.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/MonitorConfig.scala
@@ -5,6 +5,4 @@ import scala.concurrent.duration.FiniteDuration
 /**
   * Created by rtitle on 9/6/17.
   */
-case class MonitorConfig(pollPeriod: FiniteDuration,
-                         maxRetries: Int = -1,
-                         recreateCluster: Boolean = true)
+case class MonitorConfig(pollPeriod: FiniteDuration, maxRetries: Int = -1, recreateCluster: Boolean = true, creationTimeLimit: FiniteDuration)
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/MonitorConfig.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/MonitorConfig.scala
@@ -1,8 +1,10 @@
 package org.broadinstitute.dsde.workbench.leonardo.config
 
+import org.broadinstitute.dsde.workbench.leonardo.model.google.ClusterStatus
+
 import scala.concurrent.duration.FiniteDuration
 
 /**
   * Created by rtitle on 9/6/17.
   */
-case class MonitorConfig(pollPeriod: FiniteDuration, maxRetries: Int = -1, recreateCluster: Boolean = true, creationTimeLimit: FiniteDuration)
+case class MonitorConfig(pollPeriod: FiniteDuration, maxRetries: Int = -1, recreateCluster: Boolean = true, monitorStatusTimeouts: Map[ClusterStatus, FiniteDuration])
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/google/HttpGoogleDataprocDAO.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/google/HttpGoogleDataprocDAO.scala
@@ -35,8 +35,7 @@ class HttpGoogleDataprocDAO(appName: String,
                             override val workbenchMetricBaseName: String,
                             networkTag: NetworkTag,
                             defaultRegion: String,
-                            zoneOpt: Option[String],
-                            defaultExecutionTimeout: FiniteDuration)
+                            zoneOpt: Option[String])
                            (implicit override val system: ActorSystem, override val executionContext: ExecutionContext)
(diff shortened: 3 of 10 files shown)


http/src/main/resources/reference.conf
```diff
@@ -100,6 +100,7 @@ monitor {
   pollPeriod = 15 seconds
   maxRetries = -1  # means retry forever
   recreateCluster = true
+  creationTimeLimit = 20 minutes
```
jdcanas: generous window given to not flag hanging clusters prematurely. Custom images may take longer, so this may need to be bumped up in the future, but this should give a good tolerance

http/src/test/resources/reference.conf
```diff
@@ -101,6 +101,7 @@ monitor {
   pollPeriod = 1 second
   maxRetries = -1  # means retry forever
   recreateCluster = true
+  creationTimeLimit = 15 seconds
```
jdcanas: very short window for testing so test doesn't take too long. This didn't interfere with any other tests locally, and shouldn't with how our tests are designed, but it's worth noting that this may need to be tuned in the future.

jdcanas:
jenkins retest

http/src/main/resources/reference.conf
rtitle: We may want to bump this higher by default - maybe 1 hour? A lot of users specify user scripts which can take a long time to execute.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
```diff
@@ -137,9 +140,19 @@ class ClusterMonitorActor(val cluster: Cluster,
     * @return ScheduleMonitorPass
     */
   private def handleNotReadyCluster(status: ClusterStatus, instances: Set[Instance]): Future[ClusterMonitorMessage] = {
-    logger.info(s"Cluster ${cluster.projectNameString} is not ready yet (cluster status = $status, instance statuses = ${instances.groupBy(_.status).mapValues(_.size)}). Checking again in ${monitorConfig.pollPeriod.toString}.")
-    persistInstances(instances).map { _ =>
-      ScheduleMonitorPass
+    val currTimeElapsed: FiniteDuration = startupTimings.getOrElse(cluster.id, 0 seconds)
+    logger.info(s"Cluster ${cluster.projectNameString} has taken ${currTimeElapsed.toString} so far, continuing to monitor.")
```
rtitle: I feel like there's some duplicate logging here -- can we remove this line and just log in the `if..else` cases?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
```diff
@@ -137,9 +140,19 @@ class ClusterMonitorActor(val cluster: Cluster,
...
+
+    if (currTimeElapsed > monitorConfig.creationTimeLimit && status == ClusterStatus.Creating) {
+      logger.info(s"Detected that ${cluster.projectNameString} has been creating too long, time limit is ${monitorConfig.creationTimeLimit}. Failing it.")
+      Metrics.newRelic.incrementCounterIO("createClusterTimeout").unsafeRunSync()
+      handleFailedCluster(ClusterErrorDetails(Code.DEADLINE_EXCEEDED.value,Some(s"Failed to create cluster ${cluster.projectNameString} within ${monitorConfig.creationTimeLimit.toMinutes.toString()} minutes")),instances)
+    } else {
+      startupTimings += (cluster.id -> startupTimings.getOrElse(cluster.id, 0 seconds).+(monitorConfig.pollPeriod))
```
rtitle: We have the monitor start time already - couldn't we get the elapsed time by:
```
val elapsedTime = System.currentTimeMillis - startTime
```
?

I'm not sure we need the `startupTimings` Map.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
```diff
@@ -137,9 +140,19 @@ class ClusterMonitorActor(val cluster: Cluster,
     * @return ScheduleMonitorPass
     */
   private def handleNotReadyCluster(status: ClusterStatus, instances: Set[Instance]): Future[ClusterMonitorMessage] = {
-    logger.info(s"Cluster ${cluster.projectNameString} is not ready yet (cluster status = $status, instance statuses = ${instances.groupBy(_.status).mapValues(_.size)}). Checking again in ${monitorConfig.pollPeriod.toString}.")
-    persistInstances(instances).map { _ =>
-      ScheduleMonitorPass
+    val currTimeElapsed: FiniteDuration = startupTimings.getOrElse(cluster.id, 0 seconds)
+    logger.info(s"Cluster ${cluster.projectNameString} has taken ${currTimeElapsed.toString} so far, continuing to monitor.")
+
+    if (currTimeElapsed > monitorConfig.creationTimeLimit && status == ClusterStatus.Creating) {
```
rtitle: Sorry if this is scope creep - I think we should enforce a timeout for the other "monitored" statuses too:
- Creating
- Starting
- Stopping
- Deleting
- Updating

We have definitely seen clusters stuck in these states.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
```diff
@@ -219,6 +232,7 @@ class ClusterMonitorActor(val cluster: Cluster,
 
   private def shouldRecreateCluster(code: Int, message: Option[String]): Boolean = {
     // TODO: potentially add more checks here as we learn which errors are recoverable
+    logger.info("determining if we should re-create cluster")
```
rtitle: Log `${cluster.projectNameString}` -- it helps with debugging

rtitle (changes requested):
A few comments above. Also: Dataproc supports a timeout param, we should set it based on the `creationTimeLimit` param in Leo. I think the code is here:

https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/google/HttpGoogleDataprocDAO.scala#L234

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/MonitorConfig.scala
```diff
@@ -1,10 +1,10 @@
...
+
 import scala.concurrent.duration.FiniteDuration
 
 /**
   * Created by rtitle on 9/6/17.
   */
-case class MonitorConfig(pollPeriod: FiniteDuration,
-                         maxRetries: Int = -1,
-                         recreateCluster: Boolean = true)
+case class MonitorConfig(pollPeriod: FiniteDuration, maxRetries: Int = -1, recreateCluster: Boolean = true, monitorStatusTimeouts: Map[ClusterStatus, FiniteDuration])
```
kyuksel: Nit: I'd love for lines to be 120 or less char long so we can split screen on IntelliJ without scrolling horizontally. I also find it easier to read when args are stacked vertically.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
```diff
@@ -137,9 +137,18 @@ class ClusterMonitorActor(val cluster: Cluster,
     * @return ScheduleMonitorPass
     */
   private def handleNotReadyCluster(status: ClusterStatus, instances: Set[Instance]): Future[ClusterMonitorMessage] = {
-    logger.info(s"Cluster ${cluster.projectNameString} is not ready yet (cluster status = $status, instance statuses = ${instances.groupBy(_.status).mapValues(_.size)}). Checking again in ${monitorConfig.pollPeriod.toString}.")
-    persistInstances(instances).map { _ =>
-      ScheduleMonitorPass
+    val currTimeElapsed: Long = (System.currentTimeMillis() - startTime)
+
+    if (monitorConfig.monitorStatusTimeouts.keySet.contains(status) &&
+      currTimeElapsed > monitorConfig.monitorStatusTimeouts.getOrElse(status, throw new Exception(s"monitor config is not properly configured to handle status: $status")).toMillis) {
```
kyuksel: Nit:

```suggestion
      currTimeElapsed > monitorConfig.monitorStatusTimeouts.getOrElse(status, throw new Exception(s"Monitor config is not properly configured to handle status: $status")).toMillis) {
```

http/src/test/resources/reference.conf
```diff
@@ -101,6 +100,11 @@ monitor {
   pollPeriod = 1 second
   maxRetries = -1  # means retry forever
   recreateCluster = true
+  creatingTimeLimit = 15 seconds
+  startingTimeLimit = 15 seconds
+  stoppingTimeLimit = 15 seconds
+  deletingTimeLimit = 15 seconds
+  updatingTimeLimit = 15 seconds
```
kyuksel: Any downside to lowering these limits so tests can finish more quickly?

kyuksel (commented):
Looks good overall. Some comments & questions inline.

http/src/test/resources/reference.conf
jdcanas: when it is too low it has the potential to conflict with tests that use the cluster monitor in a hard-to-track-down way

http/src/main/resources/reference.conf
```diff
@@ -103,6 +102,11 @@ monitor {
   pollPeriod = 15 seconds
   maxRetries = -1  # means retry forever
   recreateCluster = true
+  creatingTimeLimit = 1 hour
+  startingTimeLimit = 20 minutes
+  stoppingTimeLimit = 10 minutes
```
rtitle: Stopping a cluster can take a weirdly long time. I'd advocate for increasing this limit -- especially since most users don't explicitly stop but rely on auto-pause. Maybe 1 hour?

Maybe also 1 hour for deleting since that's not really user facing.

The other limits look good.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/google/HttpGoogleDataprocDAO.scala
```diff
@@ -231,7 +230,7 @@ class HttpGoogleDataprocDAO(appName: String,
     // Create a NodeInitializationAction, which specifies the executable to run on a node.
     // This executable is our init-actions.sh, which will stand up our jupyter server and proxy.
     val initActions = config.initScripts.map { script =>
-      new NodeInitializationAction().setExecutableFile(script.toUri).setExecutionTimeout(finiteDurationToGoogleDuration(defaultExecutionTimeout))
+      new NodeInitializationAction().setExecutableFile(script.toUri)
```
rtitle: If this is not set does it default to infinite, or is there a dataproc default timeout?

rtitle (approved):
Nice! A couple nits above but overall 👍

jdcanas:
jenkins retest

jdcanas:
failed with `ERROR: for bond-app  sandbox 3ba27453366fe5691bb083878e45d6fb6a385a2ca6b4ab0c8c2b008d395e5e84 not found`
Jenkins retest

jdcanas:
jenkins retest

PR #1071, opened 2019-09-20
[IA-1393] Bump heap and metaspace in build

rtitle:

In all cases:


jdcanas (approved):
Worth a try, https://docs.oracle.com/javase/8/docs/technotes/guides/troubleshoot/memleaks002.html seems to think it will do something.

Qi77Qi (approved):
thanks for looking into this! I was thinking maybe somewhere in test we're leaking memory..but hard to tell

PR #1072, opened 2019-09-20
[IA-1393] fixing memory leak

jdcanas:
Seems to work locally™.

### DIFF
--- a/http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/service/TestProxy.scala
+++ b/http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/service/TestProxy.scala
@@ -2,11 +2,11 @@ package org.broadinstitute.dsde.workbench.leonardo.service
 
 import java.io.InputStream
 import java.security.{KeyStore, SecureRandom}
-import javax.net.ssl.{KeyManagerFactory, SSLContext, TrustManagerFactory}
 
+import javax.net.ssl.{KeyManagerFactory, SSLContext, TrustManagerFactory}
 import akka.http.scaladsl.Http.ServerBinding
 import akka.http.scaladsl.marshallers.sprayjson.SprayJsonSupport._
-import akka.http.scaladsl.model.headers.{`Access-Control-Allow-Origin`, ContentDispositionTypes, `Content-Disposition`}
+import akka.http.scaladsl.model.headers.{ContentDispositionTypes, `Access-Control-Allow-Origin`, `Content-Disposition`}
 import akka.http.scaladsl.model.HttpResponse
 import akka.http.scaladsl.model.ws.{BinaryMessage, Message, TextMessage}
 import akka.http.scaladsl.server.Directives._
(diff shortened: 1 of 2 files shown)


rtitle (commented):
Not sure about having these methods return Futures. If `beforeAll` in AuthProviderSpec doesn’t wait for the server to start then could we have race conditions? E.g. test failures because the server hasn’t started yet.

Maybe we could slap on a `startProxyServer.futureValue` in `AuthProviderSpec.beforeAll` to ensure the server is started before tests start running. Does that make sense?

jdcanas:
> Not sure about having these methods return Futures. If `beforeAll` in AuthProviderSpec doesn’t wait for the server to start then could we have race conditions? E.g. test failures because the server hasn’t started yet.
> 
> Maybe we could slap on a `startProxyServer.futureValue` in `AuthProviderSpec.beforeAll` to ensure the server is started before tests start running. Does that make sense?

I think the change I just made accomplishes this, but its possible it doesn't as the mistake is due to unfamiliarity with futures. I originally thought the .futureValue was extracting from the future in before due to the type requirement of how it was being used in afterAll. 

So lifting the server out of the future with .futureValue ensures the future is complete prior to execution passing the line it is on? Is it an antipattern to await on isCompleted or onComplete, which strikes me as more intuitive? 

If futureValue behaves as we're describing it, I think I can change the afterAll from `Await.result(serverBinding, 10.seconds)
        .terminate(hardDeadline = 3.seconds)` to `serverBinding.terminate(hardDeadline = 3.seconds)` where `serverBinding` is the return of `serverBinding.futureValue`

rtitle (approved):

rtitle:
Yup `.futureValue` is something from ScalaTest. It waits for the future to complete and adds some extra assertions. See http://doc.scalatest.org/3.0.1-2.12/org/scalatest/concurrent/ScalaFutures.html

I think the current code addresses the problem I mentioned above. 👍

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/service/TestProxy.scala
```diff
@@ -52,11 +56,21 @@ trait TestProxy { this: ScalaFutures =>
...
   def shutdownProxyServer() = {
-    serverBinding.unbind().futureValue
+    val onceAllConnectionsTerminated: Future[Http.HttpTerminated] =
+      Await.result(serverBinding, 10.seconds)
+        .terminate(hardDeadline = 3.seconds)
+
+    // once all connections are terminated,
+    // - you can invoke coordinated shutdown to tear down the rest of the system:
+    onceAllConnectionsTerminated.flatMap { _ =>
+      system.terminate()
```
Qi77Qi, line 68: 👍 on terminating `system`...but is `Future` really needed?

jdcanas: this is how the documentation recommended doing it, not sure the advantage though

Qi77Qi (approved):

Qi77Qi:
:shipit:  thanks for taking this on!

PR #1073, opened 2019-09-23
IA-184 swap out swagger client 

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-184

Mostly opening this PR to gather thoughts on the direction I'm going.

1. Move all config read into `Config` object. We shouldn't need to read and parse configuration in various places.
2. Replace swaggerclient with plain rest API call with http4s
- Retry and logging are handled by `Client[IO]`. And we'll see all requests to sam (turn logging response body off because we retrieve pet key from sam)
3. More traceId throughout call stack
4. Fix a bug swaggerClient was still expecting resourceId in sam for cluster's in the format of "<project>_<clusterName>", but we actually no longer store cluster's resourceId that way since https://github.com/DataBiosphere/leonardo/pull/1009 
5. Add `core` sub project for sharing code between `automation` and `http`
6. Fix automation tests decoder for `Cluster` response from `leonardo`
7. Fix hotswap script

In all cases:


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/sam/SamAuthProvider.scala
```diff
@@ -102,30 +90,36 @@ class SamAuthProvider(val config: Config, serviceAccountProvider: ServiceAccount
...
+    for {
+      hasNotebookAction <- samhasNotebookClusterPermission(internalId, action, authorization)
+      res <- if(action == ConnectToCluster) IO.pure(hasNotebookAction) else {
+        samhasProjectPermission(googleProject, action, authorization).map(r => hasNotebookAction || r)
       }
-    }
+    } yield res
   }
 
+  private def samhasProjectPermission(googleProject: GoogleProject, action: LeoAuthAction, authHeader: Authorization): IO[Boolean] = for {
```
kyuksel: I was a bit confused when I first saw this method's name because there is also a `hasProjectPermission` base method overridden above. I understand that the goal of having this additional method is factoring out some common functionality. I wonder if something like `hasProjectPermissionInternal` would be more consistent with other occurrences of this case in the code base...or some other name...just a thought...

Same for the other `private def samhas...` methods. If you decide to keep that name, I'd capitalize `h` (`samHas...` instead of `samhas...`).

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/Config.scala
```diff
@@ -1,16 +1,22 @@
...
+import org.broadinstitute.dsde.workbench.model.WorkbenchEmail
 import org.broadinstitute.dsde.workbench.model.google.GoogleProject
 import org.broadinstitute.dsde.workbench.util.toScalaDuration
+import org.http4s.Uri
 
 import scala.collection.JavaConverters._
+import scala.concurrent.duration._
 
-package object config {
+object Config {
```
kyuksel, line 20: 👍

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpSamDAO.scala
```diff
@@ -1,74 +1,220 @@
...
+    * Returning a failed Future will prevent the cluster from being created, and will call notifyClusterDeleted for the same cluster.
+    * Leo will wait, so be timely!
+    *
+    * @param internalId     The internal ID for the cluster (i.e. used for Sam resources)
+    * @param creatorEmail     The email address of the user in question
+    * @param googleProject The Google project the cluster was created in
+    * @param clusterName   The user-provided name of the Dataproc cluster
+    * @return A Future that will complete when the auth provider has finished doing its business.
+    */
+  def createCluster(internalId: ClusterInternalId, creatorEmail: WorkbenchEmail, googleProject: GoogleProject, clusterName: ClusterName, traceId: TraceId = TraceId(randomUUID)): IO[Unit] = {
```
kyuksel: The method name throws me off a bit since we're not actually creating a cluster with it but creating the cluster's associated auth resources within the auth provider...so I wonder if something like `createClusterAuthResources` may be clearer. Similarly for `deleteCluster()` below.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeoAuthProvider.scala
```diff
@@ -89,9 +83,5 @@ abstract class LeoAuthProvider(authConfig: Config, serviceAccountProvider: Servi
...
     * @return A Future that will complete when the auth provider has finished doing its business.
     */
-  def notifyClusterDeleted(internalId: ClusterInternalId, userEmail: WorkbenchEmail, creatorEmail: WorkbenchEmail, googleProject: GoogleProject, clusterName: ClusterName)(implicit executionContext: ExecutionContext): Future[Unit]
-
-  protected def getLeoServiceAccountAndKey: (WorkbenchEmail, File) = {
-    serviceAccountProvider.getLeoServiceAccountAndKey
-  }
-}
+  def notifyClusterDeleted(internalId: ClusterInternalId, userEmail: WorkbenchEmail, creatorEmail: WorkbenchEmail, googleProject: GoogleProject, clusterName: ClusterName): F[Unit]
+}
```
kyuksel, line 92:
```suggestion
}

```

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/auth/sam/SamAuthProviderSpec.scala
```diff
@@ -1,291 +1,184 @@
...
-    // request should not have been retried
-    verify(samClientThrowing404, never).invalidatePetAccessToken(any[WorkbenchEmail], any[GoogleProject])
-  }
-
 }
+
+object SamAuthProviderSpec {
+  def testClient(app: HttpApp[IO]): Client[IO] =
+    Client.fromHttpApp[IO](app)
+}
```
kyuksel, line 186:
```suggestion
}

```

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/auth/LeoAuthProviderHelperSpec.scala
```diff
@@ -1,89 +0,0 @@
...
-  * Created by rtitle on 1/25/18.
-  */
-class LeoAuthProviderHelperSpec extends TestKit(ActorSystem("leonardotest")) with FlatSpecLike with Matchers with CommonTestData with ScalaFutures with BeforeAndAfterAll {
-  import system.dispatcher
-
-  override def afterAll(): Unit = {
-    TestKit.shutdownActorSystem(system)
-    super.afterAll()
-  }
-
```
kyuksel, line 28: How are the exceptions (that used to be tested below) handled now? Do we no longer need map Sam exceptions to Leo exceptions?

kyuksel (commented):
Made a first pass with some tinkering and questions inline. Thanks for simplifying the code 👍

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/LeoRoutes.scala
```diff
@@ -42,6 +46,8 @@ abstract class LeoRoutes(val leonardoService: LeonardoService, val proxyService:
 
   def leoRoutes: Route =
     requireUserInfo { userInfo =>
+      implicit val traceId = ApplicativeAsk.const[IO, TraceId](TraceId(UUID.randomUUID()))
```
rtitle, line 49: Ah so it's essentially `ReaderT[IO, ?]`? Nice :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/LeoAuthProviderHelper.scala
```diff
@@ -1,95 +0,0 @@
...
-  extends LeoAuthProvider(authConfig, serviceAccountProvider) with FutureSupport with LazyLogging {
-
-  // Default timeout is specified in reference.conf
-  private lazy val providerTimeout = authConfig.as[FiniteDuration]("providerTimeout")
-  private implicit val scheduler = system.scheduler
-
-  private def safeCall[T](traceId: TraceId = TraceId(randomUUID))(future: => Future[T])(implicit executionContext: ExecutionContext): Future[T] = {
-    val exceptionHandler: PartialFunction[Throwable, Future[Nothing]] = {
-      case e: LeoException => Future.failed(e)
-      case te: TimeoutException =>
```
rtitle, line 52: It should be ok to remove this `Helper` layer... but it did 2 things:
1. Map internal Sam errors to a more  user-friendly `AuthProviderException`
2. Enforce a client side timeout for Sam calls (see `future.withTimeout` below). If Sam is slow or down I didn't want timeouts to propagate to the Leo API layer.

Ideally the `http4s` solution would handle these 2 things as well.

rtitle:
At a high level this looks great, I plan to do a more detailed review later.

I'm going to steal the `TraceId` stuff in my PR

Qi77Qi:
jenkins retest

Qi77Qi:
jenkins restart

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/LeoAuthProviderHelper.scala
Qi77Qi: timeout is enforced slightly different. After this PR, it'll be a max retry reached error. For translating errors to `AuthProviderException`, I kinda prefer it handled in route level and leo internals should worry much about how these exceptions are translated to API errors (something like https://github.com/DataBiosphere/welder/blob/master/server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/WelderApp.scala#L23)...Caller of some of these methods of auth provider might translate errors to other errors.

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/auth/LeoAuthProviderHelperSpec.scala
Qi77Qi: I think I prefer exception mapping to API errors done at leo route level (something like https://github.com/DataBiosphere/leonardo/pull/1073#discussion_r329683236) instead of throughout leo internals. Cuz if different layers in leo all try to translate errors, it's pretty confusing what the final error will be that we show to users

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/Boot.scala
```diff
@@ -96,23 +70,29 @@ object Boot extends IOApp with LazyLogging {
...
-        val statusService = new StatusService(gdDAO, samDAO, dbRef, dataprocConfig)
+        val statusService = new StatusService(gdDAO, appDependencies.samDAO, dbRef, dataprocConfig)
         val leoRoutes = new LeoRoutes(leonardoService, proxyService, statusService, swaggerConfig) with StandardUserInfoDirectives
         IO.fromFuture(IO(Http().bindAndHandle(leoRoutes.route, "0.0.0.0", 8080)
           .recover {
             case t: Throwable =>
               logger.error("FATAL - failure starting http server", t)
               throw t
-          }.void))
+          }.void)) >> IO.never
```
Qi77Qi, line 82: this is needed so that the `use` block doesn't return immediately

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpSamDAO.scala
Qi77Qi: updated

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoTestUtils.scala
```diff
@@ -59,7 +59,7 @@ trait LeonardoTestUtils extends WebBrowserSpec with Matchers with Eventually wit
   lazy val voldyAuthToken = UserAuthToken(voldyCreds, AuthTokenScopes.userLoginScopes)
   lazy val ronEmail = ronCreds.email
 
-  val clusterPatience = PatienceConfig(timeout = scaled(Span(30, Minutes)), interval = scaled(Span(20, Seconds)))
+  val clusterPatience = PatienceConfig(timeout = scaled(Span(10, Minutes)), interval = scaled(Span(20, Seconds)))
```
rtitle, line 61: Cool.. maybe this helps tests execute faster. Should be ok with the shorter creation time.

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoTestUtils.scala
```diff
@@ -344,7 +342,7 @@ trait LeonardoTestUtils extends WebBrowserSpec with Matchers with Eventually wit
 
   def randomClusterName: ClusterName = ClusterName(s"automation-test-a${makeRandomId().toLowerCase}z")
 
-  def defaultClusterRequest: ClusterRequest = ClusterRequest(Map("foo" -> makeRandomId()), enableWelder = Some(enableWelder))
+  val defaultClusterRequest: ClusterRequest = ClusterRequest(Map("foo" -> makeRandomId()), enableWelder = Some(enableWelder))
```
rtitle: `makeRandomId()` is a side effect.. is this ok being a `val`? I'm wondering if we'll get naming conflicts if tests execute concurrently.

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoTestUtils.scala
Qi77Qi: yea, the idea is if the cluster hasn't been able to start within 10 min. Something is wrong, and we should just abort

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoTestUtils.scala
Qi77Qi: oh, yea. Good call

core/src/main/scala/org.broadinstitute.dsde.workbench.leonardo/models.scala
```diff
@@ -0,0 +1,23 @@
+package org.broadinstitute.dsde.workbench.leonardo
```
rtitle, line 1: So this currently contains just the stuff that's identical between `http` and `automation`, right? We can have a tech debt ticket to move all models here.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/DefaultServiceAccountProvider.scala
```diff
@@ -1,36 +1,20 @@
...
 import org.broadinstitute.dsde.workbench.leonardo.model.ServiceAccountProvider
-import org.broadinstitute.dsde.workbench.model.{UserInfo, WorkbenchEmail}
 import org.broadinstitute.dsde.workbench.model.google.GoogleProject
-
-import scala.concurrent.{ExecutionContext, Future}
+import org.broadinstitute.dsde.workbench.model.{TraceId, UserInfo, WorkbenchEmail}
 
 /**
   * A ServiceAccountProvider which uses all defaults.
+  * TOOD: is this still needed? Or move this to test code?
```
rtitle, line 12: We definitely don't use it in production.. not sure about test code.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/sam/PetClusterServiceAccountProvider.scala
```diff
@@ -1,34 +1,28 @@
...
-    Future {
-      Option(blocking(samClient.getPetServiceAccount(userInfo, googleProject)))
-    }
+import org.broadinstitute.dsde.workbench.model.{TraceId, UserInfo, WorkbenchEmail}
+import org.http4s.headers.Authorization
+import org.http4s.{AuthScheme, Credentials}
+
+class PetClusterServiceAccountProvider[F[_]: Monad](sam: SamDAO[F]) extends ServiceAccountProvider[F] {
+  override def getClusterServiceAccount(userInfo: UserInfo, googleProject: GoogleProject)(implicit ev: ApplicativeAsk[F, TraceId]): F[Option[WorkbenchEmail]] = {
+    val authHeader = Authorization(Credentials.Token(AuthScheme.Bearer, userInfo.accessToken.token))
```
rtitle, line 15: Speculative question. We call some Sam APIs as the user, some as the pet service account, and others as the Leo service account. Should `HttpSamDAO` have knowledge about which user should call which API -- or should each method on `SamDAO` take an `Authorization` header, leaving it up to the caller? Asking as right now it seems like a mix of both, which is a little confusing (to me).

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/sam/PetNotebookServiceAccountProvider.scala
```diff
@@ -1,45 +1,45 @@
...
-  *
   * This is not the default ServiceAccountProvider.
   *
   * To enable, change the configuration value serviceAccounts.providerClass to
   * org.broadinstitute.dsde.workbench.leonardo.auth.PetNotebookServiceAccountProvider
   *
   * See the README for more information.
   */
-class PetNotebookServiceAccountProvider(val config: Config) extends ServiceAccountProvider(config) with SamProvider {
+class PetNotebookServiceAccountProvider[F[_]: Monad](samDAO2: SamDAO[F]) extends ServiceAccountProvider[F] {
```
rtitle: We may want to just delete this class.. it's not used in prod and if we _were_ to use it I think we would have a security vulnerability (I can give more details offline if you're curious :)).

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/sam/SamAuthProvider.scala
```diff
@@ -102,30 +89,40 @@ class SamAuthProvider(val config: Config, serviceAccountProvider: ServiceAccount
...
-    }
+    } yield res
   }
 
+  private def hasProjectPermissionInternal(googleProject: GoogleProject, action: LeoAuthAction, authHeader: Authorization)(implicit ev: ApplicativeAsk[F, TraceId]): F[Boolean] = for {
+    actionString <- Effect[F].fromEither(getProjectActionString(action))
+    res <- samDao.hasResourcePermission(googleProject, actionString, ResourceTypeName.BillingProject, authHeader)
+  } yield res
+
+  private def hasNotebookClusterPermissionInternal(clusterInternalId: ClusterInternalId, action: LeoAuthAction, authHeader: Authorization)(implicit ev: ApplicativeAsk[F, TraceId]): F[Boolean] = for {
```
rtitle, line 125: Do we need this overloaded method or can it just be folded into the other `hasNotebookClusterPermissionInternal`?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/sam/SamAuthProvider.scala
```diff
@@ -102,30 +89,40 @@ class SamAuthProvider(val config: Config, serviceAccountProvider: ServiceAccount
...
-        if (action == ConnectToCluster) {
-          hasNotebookAction
-        } else {
-          hasNotebookAction || blocking(samClient.hasActionOnBillingProjectResource(userInfo, googleProject, getProjectActionString(action)))
-        }
+  private def hasNotebookClusterPermissionInternal(internalId: ClusterInternalId, authorization: Authorization, action: NotebookClusterActions.NotebookClusterAction, googleProject: GoogleProject, clusterName: ClusterName)(implicit ev: ApplicativeAsk[F, TraceId]): F[Boolean] = {
+    for {
+      traceId <- ev.ask
+      hasNotebookAction <- hasNotebookClusterPermissionInternal(internalId, action, authorization)
+      res <- if(action == ConnectToCluster) Sync[F].pure(hasNotebookAction) else {
```
rtitle, line 108: I think this logic is correct, but it's a bit confusing so would be good to retain the above comment.
```
// if action is connect, check only cluster resource. If action is anything else, either cluster or project must be true
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/sam/SamAuthProvider.scala
```diff
@@ -134,23 +131,20 @@ class SamAuthProvider(val config: Config, serviceAccountProvider: ServiceAccount
...
-      })
-      createdClusters <- retryableFutureToFuture(retryUntilSuccessOrTimeout(shouldInvalidateSamCacheAndRetry, s"SamAuthProvider.filterUserVisibleClusters.createdClusters call failed")(samRetryInterval, samRetryTimeout) { () =>
-        Future(blocking(samClient.listCreatedClusters(userInfo)).toSet)
-      })
-    } yield {
-      clusters.filter { case (project, name) =>
-        owningProjects.contains(project) || createdClusters.contains((project, name))
-      }
+      projectPolicies <- samDao.getResourcePolicies[SamProjectPolicy](authHeader, ResourceTypeName.BillingProject)
+      owningProjects = projectPolicies.collect {case x if(x.accessPolicyName == AccessPolicyName.Owner) => x.googleProject}
```
rtitle, line 142: Maybe add `.toSet` since we're doing a `contains` check below

rtitle: Also intellij says "unnecessary parentheses" after the `if`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/sam/SamAuthProvider.scala
```diff
@@ -134,23 +131,20 @@ class SamAuthProvider(val config: Config, serviceAccountProvider: ServiceAccount
...
+      clusterPolicies <- samDao.getResourcePolicies[SamNotebookClusterPolicy](authHeader, ResourceTypeName.NotebookCluster)
+      createdPolicies = clusterPolicies.filter(_.accessPolicyName == AccessPolicyName.Creator)
```
rtitle, line 144: Should we `.collect` here instead of `.filter` so we can do another `.contains` below instead of `.exists`?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/sam/SamAuthProvider.scala
```diff
@@ -190,33 +174,10 @@ class SamAuthProvider(val config: Config, serviceAccountProvider: ServiceAccount
...
-        case e if shouldInvalidateSamCacheAndRetry(e) =>
-          // invalidate the pet token cache between retries in case it contains stale entries
-          // See https://github.com/DataBiosphere/leonardo/issues/290
-          samClient.invalidatePetAccessToken(creatorEmail, googleProject)
-          throw e
-      }
-    }
-  }
-
-  private def shouldInvalidateSamCacheAndRetry(t: Throwable): Boolean = {
```
rtitle, line 209: We decided to remove this since we weren't sure if it was still needed right? I'm 👍 on this.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/sam/SwaggerSamClient.scala
```diff
@@ -1,188 +0,0 @@
...
-import org.broadinstitute.dsde.workbench.leonardo.model.google.ClusterName
-import org.broadinstitute.dsde.workbench.model.{UserInfo, WorkbenchEmail}
-import org.broadinstitute.dsde.workbench.model.google.GoogleProject
-
-import scala.collection.JavaConverters._
-import scala.concurrent.duration.FiniteDuration
-
-case class UserEmailAndProject(userEmail: WorkbenchEmail, googleProject: GoogleProject)
-
-class SwaggerSamClient(samBasePath: String, cacheEnabled: Boolean, cacheExpiryTime: FiniteDuration, cacheMaxSize: Int, leoEmail: WorkbenchEmail, leoPem: File) extends LazyLogging {
```
rtitle, line 26: 👋

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpSamDAO.scala
```diff
@@ -1,74 +1,280 @@
...
+
+import scala.collection.JavaConverters._
+import scala.concurrent.duration._
+import scala.util.control.NoStackTrace
+import org.broadinstitute.dsde.workbench.leonardo.JsonCodec._
+
+class HttpSamDAO[F[_]: Effect](httpClient: Client[F], config: HttpSamDaoConfig)(implicit logger: Logger[F]) extends SamDAO[F] with Http4sClientDsl[F] {
+  private val saScopes = Seq(PlusScopes.USERINFO_EMAIL, PlusScopes.USERINFO_PROFILE, StorageScopes.DEVSTORAGE_READ_ONLY)
+
+  private[leonardo] def petTokenCache(implicit ev: ApplicativeAsk[F, TraceId]) = CacheBuilder.newBuilder()
```
rtitle: Similar question as above.. should `HttpSamDAO` be responsible for the `petTokenCache`, or should that live in `SamAuthProvider` and `HttpSamDAO` takes arbitrary `Authorization` headers.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpSamDAO.scala
```diff
@@ -1,74 +1,280 @@
...
+    override type ResourcePolicy = SamNotebookClusterPolicy
+  }
+  final case object BillingProject extends ResourceTypeName {
+    override def toString: String = "billing-project"
+    override type ResourcePolicy = SamProjectPolicy
+  }
+}
+
+final case object NotFoundException extends NoStackTrace
+final case class SamAuthProviderException(msg: String) extends LeoException(message = s"SamAuthProvider error: $msg")
```
rtitle: I think we landed on a friendlier error message in a separate PR. See https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/LeoAuthProviderHelper.scala#L19

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -221,18 +223,18 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
...
                                     clusterName: ClusterName,
-                                    clusterRequest: ClusterRequest): Future[Cluster] = {
-    val traceId = TraceId(randomUUID)
+                                    clusterRequest: ClusterRequest)(implicit ev: ApplicativeAsk[IO, TraceId]): Future[Cluster] = {
     for {
-      _ <- checkProjectPermission(userInfo, CreateClusters, googleProject)
-
+      _ <- checkProjectPermission(userInfo, CreateClusters, googleProject).unsafeToFuture()
+      traceId <- ev.ask.unsafeToFuture()
+_ = logger.info(s"[$traceId] successfully checked project permission")
```
rtitle: Indentation.. or maybe this statement should be removed?

rtitle (commented):
Lots of good stuff here, nice work. There's a lot of small refactoring (which is good), and the more significant changes look well done to me. I left some comments/questions inline, interested to hear your opinions.

I think/hope this will improve performance by calling Sam via an async client, rather than Java swagger codegen wrapped in a `Future`.

I'm pro getting this merged soon, it's cool that automation tests are passing.

core/src/main/scala/org.broadinstitute.dsde.workbench.leonardo/models.scala
Qi77Qi: yea, I didn't even move everything that's shared. I mostly moved only things that are related for this PR. I'm thinking we can slowly moving things over time when we see opportunities. But I think a tech debt ticket is fine too.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/DefaultServiceAccountProvider.scala
Qi77Qi: yea,  I saw it it's referenced in https://github.com/DataBiosphere/leonardo/tree/develop/automation/gce-deploy...

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/sam/PetClusterServiceAccountProvider.scala
Qi77Qi: yea, it was pretty confusing to me as well...but I'm not sure if we should differentiate them in `HttpSamDao` because the dao itself probably only cares that given a token, it calls sam, and sam can reject the call if token is invalid or token doesn't have enough permission.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/sam/SamAuthProvider.scala
Qi77Qi: nice catch! I definitely didn't mean to overload the function. They can't really be folded into one another because they're reused in different places, so I just renamed one of them

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpSamDAO.scala
Qi77Qi: 👍 ...I added traceId, but didn't add the `isTimeOut` boolean (with current setup, when max retry reaches, http4s will throw the same exception that caused it to fail. We'll see something like `[INFO] [22:42:39.401] [scala-execution-context-global-2185] org.http4s.client.middleware.Retry - Request threw an exception on attempt #4. Retrying after 4 milliseconds` in log, so hopeful it should be obvious enough. If not, we can revisit)...Added a test for testing retry https://github.com/DataBiosphere/leonardo/pull/1073/files#diff-76a78c30139db6f5b1c6cf011343f9f1R120

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
Qi77Qi: should be removed :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/sam/PetClusterServiceAccountProvider.scala
rtitle: Agreed.. but in that case we should probably move `petTokenCache`, `getCachedPetAccessToken`, and `getAccessTokenUsingLeoPem` out of `HttpSamDAO`. Each method in Sam DAO should just take a token; callers are responsible for deciding which token to pass.

rtitle (approved):
Overall this looks good, approving. I think I have a few minor unaddressed comments, but I'd also be fine with those coming in a follow on PR.

rtitle:
By they way, I opened this interesting UL bug related to this. Cluster creation doesn't work when the incoming token is a pet token, not a user token. See: https://broadworkbench.atlassian.net/browse/IA-1431

That fix can be a separate PR, just mentioning it here as it relates to this code.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpSamDAO.scala
Qi77Qi: I think the refactoring as is is big enough that I'll feel more comfortable for more refactoring in a separate PR.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/sam/PetClusterServiceAccountProvider.scala
Qi77Qi: I think https://broadworkbench.atlassian.net/browse/IA-1431 is a good opportunity for refactoring like this.

Qi77Qi:
jenkins retest

Qi77Qi:
jenkins retest

Qi77Qi:
jenkins retest

Qi77Qi:
jenkins retest

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpSamDAO.scala
```diff
@@ -39,12 +40,13 @@ import org.broadinstitute.dsde.workbench.leonardo.JsonCodec._
 class HttpSamDAO[F[_]: Effect](httpClient: Client[F], config: HttpSamDaoConfig)(implicit logger: Logger[F]) extends SamDAO[F] with Http4sClientDsl[F] {
   private val saScopes = Seq(PlusScopes.USERINFO_EMAIL, PlusScopes.USERINFO_PROFILE, StorageScopes.DEVSTORAGE_READ_ONLY)
 
-  private[leonardo] def petTokenCache(implicit ev: ApplicativeAsk[F, TraceId]) = CacheBuilder.newBuilder()
+  private[leonardo] val petTokenCache: LoadingCache[UserEmailAndProject, Option[String]] = CacheBuilder.newBuilder()
```
rtitle, line 43: Ahh.. nice find

PR #1074, opened 2019-09-24
[WIP] [IA-1156] Move async cluster creation to back Leo

rtitle:
_Note: this is a work-in-progress, opening for early feedback. Unit tests do not work yet!_

Please read this first. There is a lot of refactoring here. I will try to describe what the refactoring means in terms of front/back Leo.

## Front Leo
* Relevant classes: `LeonardoService`
* No more async cluster creation. In fact, `LeonardoService` does not do anything async to the request.
* `LeonardoService` no longer talks to Google for createCluster requests. (It only talks to Sam and the Leo DB.)
* `LeonardoService` _still_ talks to Google for deleteCluster, stopCluster, startCluster, and updateCluster. We could potentially move those Google calls to back Leo as well if there is a need for it (but in a separate PR).
* As a result of this, `LeonardoService` is a bit smaller: 667 lines instead of 1147. It also hopefully performs better in the case of many simultaneous cluster creation requests.

## Back Leo
* Relevant classes: `ClusterMonitorSupervisor`, `ClusterMonitorActor`
* Logic is mostly the same, although `ClusterMonitorActor` now performs cluster creation as part of the `Creating` phase.
* Removed weird callback to `LeonardoService` from `ClusterMonitorSupervisor`. It is now able to use `ClusterHelper` (a shared component) instead.

## Shared
* Relevant classes: `ClusterHelper`, `BucketHelper`, `TemplateHelper`
   * Note: if you have feedback on names I'm all ears; I don't love the word "helper".
* The intent of these classes is to contain useful Leo functionality that is shared between front and back Leo. While the DAOs are a thin layer on top of Google APIs, think of the Helper classes as orchestrating calls to the Google DAOs to fulfill Leo business logic.
* `ClusterHelper` has functionality around managing Dataproc clusters, including create, stop, start, etc. A lot of the `LeonardoService` code moved here. 
* `BucketHelper` has functionality around GCS, including creating buckets, populating them, permissioning them, etc. This was a pre-existing class but I updated it to use the `google2` storage interpreter.
* `TemplateHelper` contains our templating engine. I rewrote it to use `fs2.Stream` instead of `scala.io.Source`.

Any early feedback welcome; next steps are code cleanup, fixing unit tests, and doing a lot of testing to hopefully catch any regressions.

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
@@ -172,7 +172,6 @@ trait ClusterComponent extends LeoComponent {
     def listMonitoredClusterOnly(): DBIO[Seq[Cluster]] = {
       clusterQuery
         .filter { _.status inSetBind ClusterStatus.monitoredStatuses.map(_.toString) }
-        .filter { _.googleId.isDefined }
         .result map { recs =>
           recs.map(rec => unmarshalCluster(rec, Seq.empty, List.empty, Map.empty, List.empty, List.empty, List.empty))
         }
@@ -253,6 +252,14 @@ trait ClusterComponent extends LeoComponent {
     def getClusterById(id: Long): DBIO[Option[Cluster]] = {
       fullClusterQuery.filter { _._1.id === id }.result map { recs =>
         unmarshalFullCluster(recs).headOption
+      }
+    }
+
+    def getMinimalClusterById(id: Long): DBIO[Option[Cluster]] = {
+      findByIdQuery(id).result.headOption.map { recOpt =>
+        recOpt map { rec =>
+          unmarshalCluster(rec, Seq.empty, List.empty, Map.empty, List.empty, List.empty, List.empty)
+        }
       }
     }
 
@@ -373,6 +380,12 @@ trait ClusterComponent extends LeoComponent {
         .map(c => (c.initBucket, c.serviceAccountKeyId, c.googleId, c.operationName, c.stagingBucket, c.dateAccessed))
         .update(initBucket.map(_.toUri), serviceAccountKey.map(_.id.value), cluster.dataprocInfo.googleId,
           cluster.dataprocInfo.operationName.map(_.value), cluster.dataprocInfo.stagingBucket.map(_.value), Timestamp.from(Instant.now))
+    }
+
+    def clearAsyncClusterCreationFields(cluster: Cluster): DBIO[Int] = {
+      findByIdQuery(cluster.id)
+        .map(c => (c.initBucket, c.serviceAccountKeyId, c.googleId, c.operationName, c.stagingBucket, c.dateAccessed))
+        .update(None, None, None, None, None, Timestamp.from(Instant.now))
     }
 
     def updateClusterStatus(id: Long, newStatus: ClusterStatus): DBIO[Int] = {
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
@@ -313,17 +313,16 @@ case class ClusterInitValues private (googleProject: String,
 object ClusterInitValues {
   val serviceAccountCredentialsFilename = "service-account-credentials.json"
 
-  def apply(googleProject: GoogleProject, clusterName: ClusterName, initBucketName: GcsBucketName, clusterRequest: ClusterRequest, dataprocConfig: DataprocConfig,
-            clusterFilesConfig: ClusterFilesConfig, clusterResourcesConfig: ClusterResourcesConfig, proxyConfig: ProxyConfig,
-            serviceAccountKey: Option[ServiceAccountKey], userEmailLoginHint: WorkbenchEmail, contentSecurityPolicy: String,
-            clusterImages: Set[ClusterImage], stagingBucket: GcsBucketName, welderEnabled: Boolean): ClusterInitValues =
+  def apply(cluster: Cluster, initBucketName: GcsBucketName, serviceAccountKey: Option[ServiceAccountKey],
+            dataprocConfig: DataprocConfig, proxyConfig: ProxyConfig, clusterFilesConfig: ClusterFilesConfig,
+            clusterResourcesConfig: ClusterResourcesConfig, contentSecurityPolicy: String): ClusterInitValues =
     ClusterInitValues(
-      googleProject.value,
-      clusterName.value,
-      clusterImages.find(_.tool == Jupyter).map(_.dockerImage).getOrElse(""),
-      clusterImages.find(_.tool == RStudio).map(_.dockerImage).getOrElse(""),
+      cluster.googleProject.value,
+      cluster.clusterName.value,
+      cluster.clusterImages.find(_.tool == Jupyter).map(_.dockerImage).getOrElse(""),
+      cluster.clusterImages.find(_.tool == RStudio).map(_.dockerImage).getOrElse(""),
       proxyConfig.jupyterProxyDockerImage,
-      clusterImages.find(_.tool == Welder).map(_.dockerImage).getOrElse(""),
+      cluster.clusterImages.find(_.tool == Welder).map(_.dockerImage).getOrElse(""),
       GcsPath(initBucketName, GcsObjectName(clusterFilesConfig.jupyterServerCrt.getName)).toUri,
       GcsPath(initBucketName, GcsObjectName(clusterFilesConfig.jupyterServerKey.getName)).toUri,
       GcsPath(initBucketName, GcsObjectName(clusterFilesConfig.jupyterRootCaPem.getName)).toUri,
@@ -336,20 +335,20 @@ object ClusterInitValues {
       dataprocConfig.rstudioServerName,
       dataprocConfig.welderServerName,
       proxyConfig.proxyServerName,
-      clusterRequest.jupyterUserScriptUri.map(_.toUri).getOrElse(""),
-      GcsPath(stagingBucket, GcsObjectName("userscript_output.txt")).toUri,
+      cluster.jupyterUserScriptUri.map(_.toUri).getOrElse(""),
+      cluster.dataprocInfo.stagingBucket.map(x => GcsPath(x, GcsObjectName("userscript_output.txt")).toUri).getOrElse(""),
       serviceAccountKey.map(_ => GcsPath(initBucketName, GcsObjectName(serviceAccountCredentialsFilename)).toUri).getOrElse(""),
-      userEmailLoginHint.value,
+      cluster.auditInfo.creator.value,
       contentSecurityPolicy,
-      clusterRequest.userJupyterExtensionConfig.map(x => x.serverExtensions.values.mkString(" ")).getOrElse(""),
-      clusterRequest.userJupyterExtensionConfig.map(x => x.nbExtensions.values.mkString(" ")).getOrElse(""),
-      clusterRequest.userJupyterExtensionConfig.map(x => x.combinedExtensions.values.mkString(" ")).getOrElse(""),
-      clusterRequest.userJupyterExtensionConfig.map(x => x.labExtensions.values.mkString(" ")).getOrElse(""),
+      cluster.userJupyterExtensionConfig.map(x => x.serverExtensions.values.mkString(" ")).getOrElse(""),
+      cluster.userJupyterExtensionConfig.map(x => x.nbExtensions.values.mkString(" ")).getOrElse(""),
+      cluster.userJupyterExtensionConfig.map(x => x.combinedExtensions.values.mkString(" ")).getOrElse(""),
+      cluster.userJupyterExtensionConfig.map(x => x.labExtensions.values.mkString(" ")).getOrElse(""),
       GcsPath(initBucketName, GcsObjectName(clusterResourcesConfig.jupyterNotebookConfigUri.value)).toUri,
       GcsPath(initBucketName, GcsObjectName(clusterResourcesConfig.jupyterNotebookFrontendConfigUri.value)).toUri,
-      clusterRequest.defaultClientId.getOrElse(""),
-      welderEnabled.toString,  // TODO: remove this and conditional below when welder is rolled out to all clusters
-      if (welderEnabled) dataprocConfig.welderEnabledNotebooksDir else dataprocConfig.welderDisabledNotebooksDir
+      cluster.defaultClientId.getOrElse(""),
+      cluster.welderEnabled.toString,  // TODO: remove this and conditional below when welder is rolled out to all clusters
+      if (cluster.welderEnabled) dataprocConfig.welderEnabledNotebooksDir else dataprocConfig.welderDisabledNotebooksDir
     )
 }
 
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
@@ -108,10 +108,10 @@ case class ClusterError(errorMessage: String,
                         errorCode: Int,
                         timestamp: Instant)
 
-case class DataprocInfo(googleId: Option[UUID],
-                        operationName: Option[OperationName],
-                        stagingBucket: Option[GcsBucketName],
-                        hostIp: Option[IP])
+case class DataprocInfo(googleId: Option[UUID] = None,
+                        operationName: Option[OperationName] = None,
+                        stagingBucket: Option[GcsBucketName] = None,
+                        hostIp: Option[IP] = None)
 
 case class AuditInfo(creator: WorkbenchEmail,
                      createdDate: Instant,
@@ -163,25 +163,23 @@ final case class Cluster(id: Long = 0, // DB AutoInc
 object Cluster {
   type LabelMap = Map[String, String]
 
-  def create(clusterRequest: ClusterRequest,
-             internalId: ClusterInternalId,
-             userEmail: WorkbenchEmail,
-             clusterName: ClusterName,
-             googleProject: GoogleProject,
-             serviceAccountInfo: ServiceAccountInfo,
-             machineConfig: MachineConfig,
-             clusterUrlBase: String,
-             autopauseThreshold: Int,
-             clusterScopes: Set[String],
-             operation: Option[Operation] = None,
-             stagingBucket: Option[GcsBucketName] = None,
-             clusterImages: Set[ClusterImage] = Set.empty): Cluster = {
+  def createInitial(clusterRequest: ClusterRequest,
+                    internalId: ClusterInternalId,
+                    userEmail: WorkbenchEmail,
+                    clusterName: ClusterName,
+                    googleProject: GoogleProject,
+                    serviceAccountInfo: ServiceAccountInfo,
+                    machineConfig: MachineConfig,
+                    clusterUrlBase: String,
+                    autopauseThreshold: Int,
+                    clusterScopes: Set[String],
+                    clusterImages: Set[ClusterImage] = Set.empty): Cluster = {
     Cluster(
       internalId = internalId,
       clusterName = clusterName,
       googleProject = googleProject,
       serviceAccountInfo = serviceAccountInfo,
-      dataprocInfo = DataprocInfo(operation.map(_.uuid), operation.map(_.name), stagingBucket, None),
+      dataprocInfo = DataprocInfo(),
       auditInfo = AuditInfo(userEmail, Instant.now(), None, Instant.now(), None),
       machineConfig = machineConfig,
       properties = clusterRequest.properties,
@@ -199,6 +197,12 @@ object Cluster {
       clusterImages = clusterImages,
       scopes = clusterScopes,
       welderEnabled = clusterRequest.enableWelder.getOrElse(false))
+  }
+
+  def createFinal(cluster: Cluster, operation: Operation, stagingBucket: GcsBucketName): Cluster = {
+    cluster.copy(
+      dataprocInfo = DataprocInfo(Option(operation.uuid), Option(operation.name), Option(stagingBucket), None)
+    )
   }
 
   // TODO it's hacky to re-parse the Leo config in the model object.
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
@@ -351,7 +351,7 @@ object ClusterInitValues {
       GcsPath(initBucketName, GcsObjectName(clusterResourcesConfig.jupyterNotebookConfigUri.value)).toUri,
       GcsPath(initBucketName, GcsObjectName(clusterResourcesConfig.jupyterNotebookFrontendConfigUri.value)).toUri,
       cluster.defaultClientId.getOrElse(""),
-      cluster.welderEnabled.toString,  // TODO: remove this and conditional below when welder is rolled out to all clusters
+      cluster.welderEnabled.toString,  // TODO: remove this and conditional below when we sunset FireCloud, which does not have welder enabled
       if (cluster.welderEnabled) dataprocConfig.welderEnabledNotebooksDir else dataprocConfig.welderDisabledNotebooksDir
     )
 }
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
@@ -25,7 +25,8 @@ import org.broadinstitute.dsde.workbench.leonardo.model.google.{ClusterStatus, I
 import org.broadinstitute.dsde.workbench.leonardo.monitor.ClusterMonitorActor._
 import org.broadinstitute.dsde.workbench.leonardo.monitor.ClusterMonitorSupervisor.{ClusterDeleted, ClusterSupervisorMessage, RemoveFromList}
 import org.broadinstitute.dsde.workbench.leonardo.util.ClusterHelper
-import org.broadinstitute.dsde.workbench.model.google.{GcsLifecycleTypes, GoogleProject}
+import org.broadinstitute.dsde.workbench.model.ErrorReport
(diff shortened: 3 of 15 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/TemplateHelper.scala
```diff
@@ -0,0 +1,53 @@
...
+import cats.effect._
+import fs2._
+import org.broadinstitute.dsde.workbench.leonardo.config.ClusterResourcesConfig
+import org.broadinstitute.dsde.workbench.leonardo.model.ClusterResource
+
+import scala.concurrent.ExecutionContext
+
+object TemplateHelper {
+
+  def readFile[F[_]: ContextShift: Sync](file: File, blockingEc: ExecutionContext): F[Array[Byte]] = {
```
Qi77Qi: ah nice 👍 I was hesitating if using `F` is a bit too much in the other PR :)

Does the file need to be all read into memory? Just wonder if this can return `Stream[F, Byte]`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
```diff
@@ -98,25 +113,204 @@ class ClusterHelper(dbRef: DbReference,
...
+          // Note: we don't support changing the machine type for worker instances. While this is possible
+          // in GCP, Spark settings are auto-tuned to machine size. Dataproc recommends adding or removing nodes,
+          // and rebuilding the cluster if new worker machine/disk sizes are needed.
+          IO.unit
+      }
+    }.void
+  }
+
+  def updateMasterDiskSize(cluster: Cluster, diskSize: Int): IO[Unit] = {
+    cluster.instances.toList.traverse { instance =>
```
Qi77Qi: fyi, you can use `traverse_` so that you don't need to `void` later

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
```diff
@@ -98,25 +113,204 @@ class ClusterHelper(dbRef: DbReference,
...
+    for {
+      metadata <- getMasterInstanceStartupScript(cluster, welderDeploy, welderUpdate)
+
+      // Add back the preemptible instances, if any
+      _ <- if (cluster.machineConfig.numberOfPreemptibleWorkers.exists(_ > 0))
+        IO.fromFuture(IO(gdDAO.resizeCluster(cluster.googleProject, cluster.clusterName, numPreemptibles = cluster.machineConfig.numberOfPreemptibleWorkers)))
+      else IO.unit
+
+      // Start each instance individually
+      _ <- cluster.nonPreemptibleInstances.toList.traverse { instance =>
```
Qi77Qi: if we can start them in parallel, shall we do `parTraverse`?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
```diff
@@ -98,25 +113,204 @@ class ClusterHelper(dbRef: DbReference,
...
+  }
+
+  private def cleanUpGoogleResourcesOnError(googleProject: GoogleProject, clusterName: ClusterName, initBucketName: GcsBucketName, serviceAccountInfo: ServiceAccountInfo, serviceAccountKeyOpt: Option[ServiceAccountKey]): IO[Unit] = {
+    logger.error(s"Cluster creation failed in Google for $googleProject / ${clusterName.value}. Cleaning up resources in Google...")
+
+    // Clean up resources in Google
+    val deleteBucket = bucketHelper.deleteInitBucket(initBucketName) map { _ =>
+      logger.info(s"Successfully deleted init bucket ${initBucketName.value} for ${googleProject.value} / ${clusterName.value}")
+    } recover { case e =>
+      logger.error(s"Failed to delete init bucket ${initBucketName.value} for ${googleProject.value} / ${clusterName.value}", e)
```
Qi77Qi: logging is side effecting. It's better to do `recoverWith{case e => IO(logger.error(xxx))}`  or require `implicit log: Logger[IO]` implicit, and do `recoverWith{case e => log.error(xxx)}`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
```diff
@@ -98,25 +113,204 @@ class ClusterHelper(dbRef: DbReference,
...
+    val removeIamRoles = removeClusterIamRoles(googleProject, serviceAccountInfo).map { _ =>
+      logger.info(s"Successfully removed IAM roles for ${googleProject.value} / ${clusterName.value}")
+    } recover { case e =>
+      logger.error(s"Failed to remove IAM roles for ${googleProject.value} / ${clusterName.value}", e)
+    }
+
+    List(deleteBucket, deleteCluster, deleteServiceAccountKey, removeIamRoles)
+      .map(Stream.eval)
+      .reduce(_ ++ _)
+      //.parJoin(4)  TODO: why does this not compile?
```
Qi77Qi: I think you need Stream of Streams for `parJoin` to work. Here you have `List[Stream[IO, ???]`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
```diff
@@ -126,4 +320,94 @@ class ClusterHelper(dbRef: DbReference,
...
+        d <- k.privateKeyData.decode
+      } yield GcsResource(GcsBlobName(ClusterInitValues.serviceAccountCredentialsFilename), d.getBytes(Charset.defaultCharset))
+    ).unNone
+
+    (rawFilesToUpload ++ rawResourcesToUpload ++ templatedResourcesToUpload ++ privateKey).flatMap { resource =>
+      bucketHelper.storeObject(initBucketName, resource.gcsBlobName, resource.content, "text/plain")
+    }
+  }
+
+  private def getClusterVPCSettings(projectLabels: Map[String, String]): Option[Either[VPCNetworkName, VPCSubnetName]] = {
```
Qi77Qi: shall we create a sum type for `VPCNetworkName` and `VPCSubnetName` instead of using `Either`?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
```diff
@@ -336,20 +339,20 @@ object ClusterInitValues {
...
+      cluster.userJupyterExtensionConfig.map(x => x.nbExtensions.values.mkString(" ")).getOrElse(""),
+      cluster.userJupyterExtensionConfig.map(x => x.combinedExtensions.values.mkString(" ")).getOrElse(""),
+      cluster.userJupyterExtensionConfig.map(x => x.labExtensions.values.mkString(" ")).getOrElse(""),
       GcsPath(initBucketName, GcsObjectName(clusterResourcesConfig.jupyterNotebookConfigUri.value)).toUri,
       GcsPath(initBucketName, GcsObjectName(clusterResourcesConfig.jupyterNotebookFrontendConfigUri.value)).toUri,
-      clusterRequest.defaultClientId.getOrElse(""),
-      welderEnabled.toString,  // TODO: remove this and conditional below when welder is rolled out to all clusters
-      if (welderEnabled) dataprocConfig.welderEnabledNotebooksDir else dataprocConfig.welderDisabledNotebooksDir
+      cluster.defaultClientId.getOrElse(""),
+      cluster.welderEnabled.toString,  // TODO: remove this and conditional below when welder is rolled out to all clusters
```
Qi77Qi: shall we change the comment to when we sunset firecloud...just so we don't forget about firecloud :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
```diff
@@ -266,33 +266,61 @@ class ClusterMonitorActor(val cluster: Cluster,
...
+
+    future.recover { case e =>
+      logger.error(s"Failed to create cluster ${cluster.projectNameString} in Google", e)
+      val errorMessage = e match {
+        case leoEx: LeoException =>
+          ErrorReport.loggableString(leoEx.toErrorReport)
+        case _ =>
+          s"Failed to create cluster ${cluster.projectNameString} due to ${e.toString}"
+      }
+      FailedCluster(ClusterErrorDetails(-1, Option(errorMessage)), Set.empty)
```
Qi77Qi: any reason why `Option` is used in many places? Looks to me it can just be `Some` when the variable is never going to be `null`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
```diff
@@ -151,16 +151,16 @@ class ClusterMonitorActor(val cluster: Cluster,
     */
   private def handleReadyCluster(publicIp: IP, instances: Set[Instance]): Future[ClusterMonitorMessage] = {
     for {
-      clusterStatus <- getDbClusterStatus
+      cluster <- getDbCluster
```
Qi77Qi, line 154: original code seems to just getting `Status`, any reason changing it to getting the whole `Cluster`?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
```diff
@@ -266,33 +266,61 @@ class ClusterMonitorActor(val cluster: Cluster,
...
+      }
+    } yield next
+  }
+
   /**
     * Queries Google for the cluster status and takes appropriate action depending on the result.
     * @return ClusterMonitorMessage
     */
-  private def checkClusterInGoogle(leoClusterStatus: ClusterStatus): Future[ClusterMonitorMessage] = {
+  private def checkClusterInGoogle(cluster: Cluster): Future[ClusterMonitorMessage] = {
```
Qi77Qi, line 323: same question as before

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -166,176 +137,62 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
...
+    } yield cluster
   }
 
-  def internalGetActiveClusterDetails(googleProject: GoogleProject, clusterName: ClusterName): Future[Cluster] = {
+  private def internalGetActiveClusterDetails(googleProject: GoogleProject, clusterName: ClusterName): Future[Cluster] = {
     dbRef.inTransaction { dataAccess =>
-      getActiveCluster(googleProject, clusterName, dataAccess)
+      dataAccess.clusterQuery.getActiveClusterByName(googleProject, clusterName)
+    } map {
+      case None => throw ClusterNotFoundException(googleProject, clusterName)
```
Qi77Qi, line 203: prefer to use `flatMap` and do `Future.failed(e)`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/TemplateHelper.scala
rtitle: :)

These methods could return `Stream[F, Byte]` but I think we'll still need to read it into memory at some point before we persist to GCS (`createBlob` takes `Array[Byte]`).

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
rtitle: Yeah that should be fine I think

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
rtitle: Yeah that makes sense... think I misunderstood `parJoin`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
rtitle: That probably makes sense.. will look into it.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
rtitle: Habit - I agree to use `Some` when not dealing with potential nulls

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
rtitle: Because now at the `Creating` phase we're populating fields of the cluster which are needed for later stages. Previously we could assume everything except for `status` was relatively immutable through the life of `ClusterMonitorActor` -- this is no longer the case. 

This query isn't doing any joins so my hope is it's not too much more expensive than `getDbClusterStatus`.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
```diff
@@ -256,6 +255,14 @@ trait ClusterComponent extends LeoComponent {
       }
     }
 
+    def getMinimalClusterById(id: Long): DBIO[Option[Cluster]] = {
```
kyuksel, line 258: Is what's _minimal_ the cluster or the cluster _fields_? If the latter, would it be better to rename it to `getMinimalClusterFieldsById` or something?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
```diff
@@ -163,25 +163,23 @@ final case class Cluster(id: Long = 0, // DB AutoInc
...
-             clusterName: ClusterName,
-             googleProject: GoogleProject,
-             serviceAccountInfo: ServiceAccountInfo,
-             machineConfig: MachineConfig,
-             clusterUrlBase: String,
-             autopauseThreshold: Int,
-             clusterScopes: Set[String],
-             operation: Option[Operation] = None,
-             stagingBucket: Option[GcsBucketName] = None,
-             clusterImages: Set[ClusterImage] = Set.empty): Cluster = {
```
kyuksel, line 178: Can we add a comment explaining the distinction between _initial_ and _final_ `create`s?

kyuksel: On a second look, it looks like _clusterInitial_ is instance-ing a `Cluster` and _clusterFinal_ is adding `DataprocInfo` so perhaps the names could reflect that? My main hope is to reserve `createCluster`y names for the actual Dataproc cluster creation code to reduce overloading of the term.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/TemplateHelper.scala
```diff
@@ -0,0 +1,53 @@
...
+  private def fileStream[F[_]: ContextShift: Sync](file: File, blockingEc: ExecutionContext): Stream[F, Byte] = {
+    io.file.readAll[F](file.toPath, blockingEc, 4096)
+  }
+
+  private def resourceStream[F[_]: ContextShift: Sync](clusterResource: ClusterResource, blockingEc: ExecutionContext): Stream[F, Byte] = {
+    val inputStream = Sync[F].delay(getClass().getResourceAsStream(s"${ClusterResourcesConfig.basePath}/${clusterResource.value}"))
+    io.readInputStream[F](inputStream, 4096, blockingEc)
+  }
+
+}
```
kyuksel, line 41:
```suggestion
}

```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorSupervisor.scala
```diff
@@ -95,27 +92,8 @@ class ClusterMonitorSupervisor(monitorConfig: MonitorConfig, dataprocConfig: Dat
...
-              Some(cluster.autopauseThreshold),
-              cluster.defaultClientId,
-              cluster.clusterImages.find(_.tool == Jupyter).map(_.dockerImage).flatMap(s => ContainerImage.stringToJupyterDockerImage(s)))
-            val createFuture = leonardoService.internalCreateCluster(cluster.auditInfo.creator, cluster.serviceAccountInfo, cluster.googleProject, cluster.clusterName, clusterRequest, traceId)
-            createFuture.failed.foreach { e =>
-              logger.error(s"[$traceId] Error occurred recreating cluster ${cluster.projectNameString}", e)
-            }
-            createFuture
-          case None => Future.failed(new WorkbenchException(s"[$traceId] Cluster ${cluster.projectNameString} not found in the database"))
+          dataAccess.clusterQuery.clearAsyncClusterCreationFields(cluster) >>
```
kyuksel, line 95: If `>>` is the same as `andThen`, the latter might be more readable.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorSupervisor.scala
```diff
@@ -58,7 +55,7 @@ object ClusterMonitorSupervisor {
   private case object CheckForClusters extends ClusterSupervisorMessage
 }
 
-class ClusterMonitorSupervisor(monitorConfig: MonitorConfig, dataprocConfig: DataprocConfig, clusterBucketConfig: ClusterBucketConfig, gdDAO: GoogleDataprocDAO, googleComputeDAO: GoogleComputeDAO, googleStorageDAO: GoogleStorageDAO, google2StorageDAO: GoogleStorageService[IO], dbRef: DbReference, authProvider: LeoAuthProvider, autoFreezeConfig: AutoFreezeConfig, jupyterProxyDAO: JupyterDAO, rstudioProxyDAO: RStudioDAO, welderProxyDAO: WelderDAO, leonardoService: LeonardoService, clusterHelper: ClusterHelper)
+class ClusterMonitorSupervisor(monitorConfig: MonitorConfig, dataprocConfig: DataprocConfig, clusterBucketConfig: ClusterBucketConfig, gdDAO: GoogleDataprocDAO, googleComputeDAO: GoogleComputeDAO, googleStorageDAO: GoogleStorageDAO, google2StorageDAO: GoogleStorageService[IO], dbRef: DbReference, authProvider: LeoAuthProvider, autoFreezeConfig: AutoFreezeConfig, jupyterProxyDAO: JupyterDAO, rstudioProxyDAO: RStudioDAO, welderProxyDAO: WelderDAO, clusterHelper: ClusterHelper)
   extends Actor with Timers with LazyLogging {
   import context.dispatcher
```
kyuksel, line 61: Now that more of the cluster creation will be handled by the back leo, I wonder if we should try to use `TraceId` here as much as possible, too. Ideally, we could persist a `traceId` in the DB when the first record is created by the front Leo so monitors could retrieve and use it throughout the lifecycle of a cluster. I know it's work but it feels like it'd pay off when it comes to debugging. Alternatively, we could avoid the DB persistence work and create a new traceId for the back-Leo ops (perhaps by having `createChildActor` accept a `TraceId`). Even though we wouldn't be able to have a consistent `traceId` across the front- and back-Leos this way, it would be better than nothing.

kyuksel (commented):
Made a first pass with some comments inline. Will also add some offline.

kyuksel:
It'd be great to performance-test these changes before merging.

kyuksel:
I'm tinkering through potential performance implications. 

For instance, if 200 users kick off cluster creation and all are rapidly registered to the DB by the front-Leos, when the back-Leo discovers them, since we're using actors, processing of them (i.e. submitting cluster creation requests to Google) will be performed sequentially, correct? If so, could this mean some of the latter ones in the queue may have to wait for quite a while. I wonder how this might increase the variance in cluster creation duration overall. 

Also, is it possible that when we have a large amount of clusters that are discovered to be created at a scheduled polling, the corresponding actions may not be taken in time before the next scheduled polling happens, and if so, would that cause any issues?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
rtitle: It's minimal in that it queries the `CLUSTER` table but doesn't do any joins. The result is a `Cluster` object that is partially populated (e.g. it has fields directly on the cluster table but doesn't have `instances` for example). 

I'd love to represent this better in the type system somehow, rather then relying on method naming conventions. I'll think about this, but may be better for a separate PR.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
rtitle: Yes you're right -- good feedback, I'll make the names clearer.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/TemplateHelper.scala
rtitle: ?

(is this a new github feature or just something I've never used?)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorSupervisor.scala
Qi77Qi: `>>` or `*>` is pretty easy once you're used to it..and it's pretty commonly in libraries like fs2..so I think it's worth the effort to get familiar with it

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorSupervisor.scala
Qi77Qi: I have a different pattern for introducing `TraceId` in my other PR...so I'll leave it out here for now

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorSupervisor.scala
rtitle: It's not, there's an important distinction. :) `>>` is defined for any monad and is essentially a `flatMap` where you throw away the input. `a >> b` is equivalent to `a.flatMap { _ => b }`.

`andThen` is implemented on scala Futures and is importantly _not_ a `flatMap` -- it is used to create a fire-and-forget callback (akin to `onComplete`).

I've been using `>>` as a more concise way of writing `flatMap { _ => ...}`, but I agree symbolic operators are not always readable. I'm open to other opinions here.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorSupervisor.scala
rtitle: I agree, this is a good idea. When there are errors and back-leo persists to the `CLUSTER_ERROR` table, we could include a trace id to correlate with back-Leo logs. I'll try to make this happen -- thanks!

rtitle (commented):
Good questions. Agreed this needs performance testing.

> when the back-Leo discovers them, since we're using actors, processing of them (i.e. submitting cluster creation requests to Google) will be performed sequentially, correct? 

I think they will likely still happen in parallel -- we create a child actor for each cluster, and separate actors can work in parallel. It's possible to impose limits to parallelism at the supervisor level (perhaps by providing a custom dispatcher). This might be an interesting knob to tune.

> is it possible that when we have a large amount of clusters that are discovered to be created at a scheduled polling, the corresponding actions may not be taken in time before the next scheduled polling happens, and if so, would that cause any issues?

I _think_ the stateful-ness of actors handles this, as we track monitored clusters in a `var`. It's a good thing to stress test though.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorSupervisor.scala
rtitle: Ohh thanks @Qi77Qi, I'll look at that.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/TemplateHelper.scala
kyuksel: Yea, I think it's pretty neat for simple things. You can use it by hitting the +- button between Preview and aA buttons in edit mode.

kyuksel:
> I think they will likely still happen in parallel -- we create a child actor for each cluster, and separate actors can work in parallel.

Then I'd guess that as long as child actor creations can keep up with simultaneous cluster creation requests, clients shouldn't be backed up much 🤞

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorSupervisor.scala
kyuksel: huh...I see this in `DBIOAction.scala`:

```scala
  /** A shortcut for `andThen`. */
  final def >> [R2, S2 <: NoStream, E2 <: Effect](a: DBIOAction[R2, S2, E2]): DBIOAction[R2, S2, E with E2] =
    andThen[R2, S2, E2](a)
```

Either way, I don't feel strongly and your comments were informative :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorSupervisor.scala
kyuksel: > I have a different pattern for introducing TraceId in my other PR...so I'll leave it out here for now

@Qi77Qi The pattern you mention is for how to pass around `TraceId`, correct? We would still need to persist a `TraceId` in DB to be able to use it across the lifecycle.

rtitle:
I'm working on rebasing this and cleaning it up. Closing this PR for now.

PR #1077, opened 2019-09-25
[IA-1391] Fail gracefully if welder-deploy doesn't work

rtitle:
See https://broadworkbench.atlassian.net/browse/IA-1391

Makes Leo `startCluster` fail with a friendly error if trying to start a cluster that's too old.

In all cases:

### DIFF
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -39,8 +39,8 @@ dataproc {
   # The current value is only used for integration testing.
   updateWelderLabel = "TEST_ONLY_UPDATE_WELDER"
 
-  # Leo will prevent starting any clusters that were created before this date.
-  clusterCutoffDate = "2019-06-04"
+  # Leo will prevent deploying welder to any clusters that were created before this date.
+  deployWelderCutoffDate = "2019-08-01"
 }
 
 # cluster scripts and config
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -38,6 +38,9 @@ dataproc {
   # TODO: change this to 'saturnVersion' once ready to enable in production for Terra clusters.
   # The current value is only used for integration testing.
   updateWelderLabel = "TEST_ONLY_UPDATE_WELDER"
+
+  # Leo will prevent starting any clusters that were created before this date.
+  clusterCutoffDate = "2019-06-04"
 }
 
 # cluster scripts and config
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
@@ -613,38 +613,38 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
 
   def internalStartCluster(userEmail: WorkbenchEmail, cluster: Cluster): Future[Unit] = {
     if (cluster.status.isStartable) {
-      if (isClusterBeforeCutoffDate(cluster)) {
-        Future.failed(ClusterOutOfDateException())
-      } else {
-        val welderDeploy = shouldDeployWelder(cluster)
-        val welderUpdate = shouldUpdateWelder(cluster)
-        for {
-          // Check if welder should be deployed
-          updatedCluster <- if (welderDeploy || welderUpdate) updateWelder(cluster).unsafeToFuture() else Future.successful(cluster)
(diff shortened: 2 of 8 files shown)


http/src/main/resources/reference.conf
```diff
@@ -38,6 +38,9 @@ dataproc {
   # TODO: change this to 'saturnVersion' once ready to enable in production for Terra clusters.
   # The current value is only used for integration testing.
   updateWelderLabel = "TEST_ONLY_UPDATE_WELDER"
+
+  # Leo will prevent starting any clusters that were created before this date.
+  clusterCutoffDate = "2019-06-04"
```
rtitle: I'm still verifying this date

Qi77Qi: 6/6 is when there were a prod release

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -66,6 +68,12 @@ case class ClusterCannotBeDeletedException(googleProject: GoogleProject, cluster
 case class ClusterCannotBeStartedException(googleProject: GoogleProject, clusterName: ClusterName, status: ClusterStatus)
   extends LeoException(s"Cluster ${googleProject.value}/${clusterName.value} cannot be started in ${status.toString} status", StatusCodes.Conflict)
 
+case class ClusterOutOfDateException()
+  extends LeoException(
+    "Your notebook runtime is out of date, and cannot be started due to recent updates in Terra. If you generated " +
+    "data or copied external files to the runtime that you want to keep please contact support. Otherwise, simply " +
```
Qi77Qi: is there an email or link we can put in this message? just so user won't need to figure out where to go to contact support themselves

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -1144,4 +1155,12 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
     immutable.Map(googleKey -> startupScriptContent)
   }
 
+  private def isClusterBeforeCutoffDate(cluster: Cluster): Boolean = {
+    (for {
+      dateStr <- dataprocConfig.clusterCutoffDate
+      date <- Try(new SimpleDateFormat("yyyy-MM-dd").parse(dateStr)).toOption
+      isClusterBeforeCutoffDate = cluster.auditInfo.createdDate.isBefore(date.toInstant)
+    } yield isClusterBeforeCutoffDate) getOrElse false
```
Qi77Qi: if we can't get cluster's created date, it's probably in bad state? if that's the case, shall we default to `true`?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
rtitle: Yes good call, will add a link.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
rtitle: If the config is not present or it's not parseable we probably shouldn't block cluster start for all clusters... hence defaulting to `false`.

Qi77Qi (approved):

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -605,35 +613,38 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
...
-             else Future.unit
-
-        // Start each instance individually
-        _ <- Future.traverse(updatedCluster.nonPreemptibleInstances) { instance =>
-          // Install a startup script on the master node so Jupyter starts back up again once the instance is restarted
-          instance.dataprocRole match {
-            case Some(Master) =>
-              googleComputeDAO.addInstanceMetadata(instance.key, getMasterInstanceStartupScript(updatedCluster, welderDeploy, welderUpdate)) >>
+      if (isClusterBeforeCutoffDate(cluster)) {
+        Future.failed(ClusterOutOfDateException())
```
rtitle: I just realized this would block old clusters starting in _firecloud_, which is not what we want. Probably needs to only block start if we're trying to deploy welder on the cluster.

PR #1081, opened 2019-10-01
[IA-1326] adding metric for async cluster creation failure

jdcanas:
Will add alert after this is in.

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
@@ -220,7 +220,7 @@ class ClusterMonitorActor(val cluster: Cluster,
         for {
           // update the cluster status to Error
           _ <- dbRef.inTransaction { _.clusterQuery.updateClusterStatus(cluster.id, ClusterStatus.Error) }
-          _ <- Metrics.newRelic.incrementCounterFuture("asyncClusterCreationFailure")
+          _ <- Metrics.newRelic.incrementCounterFuture(s"asyncClusterCreationFailure-${errorDetails.code}")
 
           // Remove the Dataproc Worker IAM role for the pet service account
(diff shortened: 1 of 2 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
```diff
@@ -210,6 +210,8 @@ class ClusterMonitorActor(val cluster: Cluster,
         for {
           // update the cluster status to Error
           _ <- dbRef.inTransaction { _.clusterQuery.updateClusterStatus(cluster.id, ClusterStatus.Error) }
+          _ <- Metrics.newRelic.incrementCounterFuture("asyncClusterCreationFailure")
```
Qi77Qi: would be nice to have this count broken down into different error cases (some general error cases say `Initialization action failed`, `Google error`, `Internal server error` etc. I `select * from CLUSTER_ERROR limit 100` to gather some ideas about what some of those categories can be)

jdcanas: While I agree this would be useful and could be easily accomplished by including the error code in the metric name, it would make alerting it a challenge, as newrelic does not support wildcards in this fashion https://discuss.newrelic.com/t/feature-idea-custom-metric-wildcards/23751

Qi77Qi: can we have a few major categories like the ones I pointed out (@rtitle might have a better idea what better categories we should monitor), which shouldn't be many. And create alerts for each of them

jdcanas: so I added the code to the metric. There are only 17 distinct ones. We can choose to alert on which ones we find relevant and change that at will.

Qi77Qi (approved):

PR #1088, opened 2019-10-09
IA-1437 disallow highcpu machine type for free credit projects

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-1437

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
@@ -81,6 +81,8 @@ case class ClusterCannotBeUpdatedException(cluster: Cluster)
 case class ClusterMachineTypeCannotBeChangedException(cluster: Cluster)
   extends LeoException(s"Cluster ${cluster.projectNameString} in ${cluster.status} status must be stopped in order to change machine type", StatusCodes.Conflict)
 
+final case class InvalidClusterMachineType(msg: String) extends LeoException(msg, StatusCodes.BadRequest)
+
 case class ClusterDiskSizeCannotBeDecreasedException(cluster: Cluster)
   extends LeoException(s"Cluster ${cluster.projectNameString}: decreasing master disk size is not allowed", StatusCodes.PreconditionFailed)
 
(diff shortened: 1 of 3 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -168,6 +170,10 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
                     clusterRequest: ClusterRequest): Future[Cluster] = {
     val traceId = TraceId(randomUUID)
     for {
+    // Disallow users creating `highcpu` machine type with free credit projects
+      _ <- if(googleProject.value.startsWith("fccredits") && clusterRequest.machineConfig.exists(_.masterMachineType.exists(_.contains("highcpu"))))
```
rtitle, line 174: Geraldine has some concerns about this; some workshops require `highcpu` clusters for legitimate purposes. It might be better to put specific disallowed machine types in leo.conf, which we could change as needed.

rtitle (changes requested):
see comment above

Comms also asked to be informed about any changes around free credits projects, as it impacts workshops. So let's hold off on merging until we coordinate with them (it's also encouraging that we haven't had any new bitcoin users since yesterday).

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -81,6 +81,8 @@ case class ClusterCannotBeUpdatedException(cluster: Cluster)
 case class ClusterMachineTypeCannotBeChangedException(cluster: Cluster)
   extends LeoException(s"Cluster ${cluster.projectNameString} in ${cluster.status} status must be stopped in order to change machine type", StatusCodes.Conflict)
 
+final case class InvalidClusterMachineType(msg: String) extends LeoException(msg, StatusCodes.BadRequest)
```
kyuksel, line 84: `Unallowed` instead of `Invalid` may be more correct

Qi77Qi:
security team has turned on alerts, and will potentially shutdown services if they detect malicious behavior. Hence closing this PR.

PR #1094, opened 2019-10-11
Delete delocalization.json and label old clusters 

akarukappadath:
moved from branch `rt-fix-startup-script-again` to avoid rebase

In all cases:

### DIFF
--- a/http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoServiceSpec.scala
+++ b/http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoServiceSpec.scala
@@ -1352,7 +1352,7 @@ class LeonardoServiceSpec extends TestKit(ActorSystem("leonardotest")) with Flat
     computeDAO.instanceMetadata.values.map(_.keys).flatten.toSet shouldBe Set("startup-script")
   }
 
-  it should "outdated cluster should successfully start and be labelled" in isolatedDbTest {
+  it should "label and start an outdated cluster" in isolatedDbTest {
     // check that the cluster does not exist
     gdDAO.clusters should not contain key(name1)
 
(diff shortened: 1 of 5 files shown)


http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoServiceSpec.scala
```diff
@@ -1370,13 +1372,16 @@ class LeonardoServiceSpec extends TestKit(ActorSystem("leonardotest")) with Flat
...
+    leo.startCluster(userInfo, project, name1).futureValue
 
-    // cluster status should still be Stopped in the DB
-    dbFutureValue { _.clusterQuery.getClusterByUniqueKey(clusterCreateResponse) }.get
-      .status shouldBe ClusterStatus.Stopped
+    // cluster status should Running and have new label
+    val cluster = dbFutureValue { _.clusterQuery.getClusterByUniqueKey(clusterCreateResponse) }.get
 
+    eventually {
+      cluster.status shouldBe ClusterStatus.Running
```
rtitle: I think it should be `Starting` at this point

rtitle (approved):
Changes look good; one minor comment which I think is causing a test failure

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoServiceSpec.scala
akarukappadath: I was hoping to test that the cluster doesn't get stuck in Starting status, since they did when the script wasn't correctly working

rtitle: Yeah.. but the `ClusterMonitorActor` is what transitions it to `Running`, and this test is only verifying `LeonardoService`. The unit tests don't start a real cluster, so the startup script isn't even run.

We have an [automation test](https://github.com/DataBiosphere/leonardo/blob/develop/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookClusterMonitoringSpec.scala#L184-L221) to verify that the startup script deploys welder. I'm not sure it's worth adding another one for disabling delocalization -- it sounds tricky to set up and we can probably by confident with manual testing.

PR #1100, opened 2019-10-15
[RW-3586] Enable setting custom environment variables within a cluster

als364:
https://precisionmedicineinitiative.atlassian.net/browse/RW-3586 actual ticket, which probably no one has access to

--

In all cases:


automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoModelCopy.scala
```diff
@@ -40,7 +40,8 @@ case class ClusterRequest(labels: LabelMap = Map(),
                           rstudioDockerImage: Option[String] = None,
                           welderDockerImage: Option[String] = None,
                           scopes: Set[String] = Set.empty,
-                          enableWelder: Option[Boolean] = None)
+                          enableWelder: Option[Boolean] = None,
+                          customClusterEnvironmentVariables: Option[Map[String, String]] = None)
```
Qi77Qi: 1. this is probably not the place you want to make the change. https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala#L87 might be the place you need

2. Use `Map[String, String]` instead of wrapping in `Option` since an empty `Map` already has represents the idea of no custom environment variable

Qi77Qi:
what is the motivation for custom environment variable for cluster creation? is this going to be exposed in UI?

als364:
I work on All of Us and we want to expose the workspace CDR version as an environment variable in our notebooks by default.

Qi77Qi:
fyi, https://broadworkbench.atlassian.net/wiki/spaces/IA/pages/100401153/Testing+notebook+functionality+with+Fiab has some documentation around how you can test the change in a fiab

als364:
I don't think I have permissions to view that page.

http/src/main/resources/jupyter/init-actions.sh
```diff
@@ -92,6 +92,8 @@ if [ ! -z ${SERVICE_ACCOUNT_CREDENTIALS} ] ; then
   export GOOGLE_APPLICATION_CREDENTIALS=/etc/${SERVICE_ACCOUNT_CREDENTIALS}
 fi
 
+gsutil cp 'custom-env-vars.json' /etc
```
rtitle: Think this needs to be a bucket path. Also it looks like `jupyter-docker-compose.yaml` is expecting a file named `/etc/custom_env_vars.env`, not `custom-env-vars.json`.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
```diff
@@ -84,7 +84,8 @@ final case class ClusterRequest(labels: LabelMap = Map.empty,
                                 rstudioDockerImage: Option[String] = None,
                                 welderDockerImage: Option[String] = None,
                                 scopes: Set[String] = Set.empty,
-                                enableWelder: Option[Boolean] = None)
+                                enableWelder: Option[Boolean] = None,
+                                customClusterEnvironmentVariables: Map[String, String] = Map.empty[String, String])
```
rtitle: You probably also need to update the json serializer `ClusterRequestFormat`, also in this file. This field can be handled the same way as `labels`.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
```diff
@@ -305,6 +306,7 @@ case class ClusterInitValues private (googleProject: String,
 
 object ClusterInitValues {
   val serviceAccountCredentialsFilename = "service-account-credentials.json"
+  val customEnvVarFilename = "custom-env-vars.env"
```
rtitle: In `jupyter-docker-compose` it expects a file named `custom_env_vars.env` -- this should probably be consistent.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -934,11 +934,19 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
       clusterRequest.enableWelder.getOrElse(false))
     val replacements: Map[String, String] = clusterInit.toMap
 
+    val tmpCustomEnvVarFile = File.createTempFile("custom-env-vars", "env")
```
rtitle: It's probably possible to build this content as a `String` and upload to GCS directly, instead of creating a tmp file. It's already in memory in the `clusterRequest`.

http/src/main/resources/jupyter/init-actions.sh
als364: Will unify filenames.

Using a bucket path would make more sense. I'm having trouble figuring out whether the bucket we upload things to in `initializeBucketOptions` is passed to this script and if so how.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
als364: Want me to make an `EnvironmentVariableMap` type for this too or just leave it as a `Map[String, String]`?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
als364: a la the templated resources at line ~970?

http/src/main/resources/jupyter/init-actions.sh
rtitle: There's a few things to do. You can probably use `jupyterNotebookConfigUri` as a model.

1. Add a `customEnvVarsUri` field to [ClusterResourcesConfig](https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/ClusterResourcesConfig.scala)
2. Add a `customEnvVars` field to [ClusterInitValues](https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala#L271)
3. In `ClusterInitValues.apply`, populate it like [this](https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala#L341)
4. In [LeonardoService.initializeBucketObjects](https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala#L922), build the content and upload to GCS by adding a call like [this](https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala#L971).
5. In [init-actions.sh](https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/resources/jupyter/init-actions.sh), add a line like [this](https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/resources/jupyter/init-actions.sh#L131)
6. In [init-actions.sh](https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/resources/jupyter/init-actions.sh), add a `gsutil cp` like [this](https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/resources/jupyter/init-actions.sh#L239-L242)
7. Set the `env_file` in jupyter-docker-compose as you have currently

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
rtitle: Nah, leave it as `Map[String, String]`. I don't think the type alias adds any value.

Btw, you can probably just write `Map.empty` - it should infer the types.

als364: Scala magic!

http/src/main/resources/jupyter/init-actions.sh
als364: For #5, what actually populates the values on the RHS?

rtitle: Leo will replace templated variables in the format `$(..)`. The templated variable corresponds to the variable names in `ClusterInitValues` (steps 2-3).

rtitle:
Btw, you probably need to merge in https://github.com/DataBiosphere/leonardo/pull/1105, or rebase your branch once that PR is merged to fix a compilation error.

als364:
retest this please

als364:
bah

als364:
jenkins retest

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookCustomizationSpec.scala
```diff
@@ -106,5 +106,29 @@ final class NotebookCustomizationSpec extends GPAllocFixtureSpec with ParallelTe
...
+        withWebDriver { implicit driver =>
+          withNewNotebook(cluster, Python3) { notebookPage =>
+            notebookPage.executeCell("import os")
+            val envVar = notebookPage.executeCell("os.getenv('KEY')")
+            envVar shouldBe Some("'value'")
+          }
+        }
+      }
+
+      withNewCluster(billingProject, request = defaultClusterRequest.copy(customClusterEnvironmentVariables = Map("KEY" -> "value"))) { cluster =>
```
rtitle: Could this use the same cluster as above? It's costly to create a lot of clusters in tests; better to reuse them if possible.
```
      withNewCluster(billingProject, request = defaultClusterRequest.copy(customClusterEnvironmentVariables = Map("KEY" -> "value"))) { cluster =>
          withNewNotebook(cluster, Python3) { notebookPage =>
            notebookPage.executeCell("import os")
            val envVar = notebookPage.executeCell("os.getenv('KEY')")
            envVar shouldBe Some("'value'")
          }
          withNewNotebook(cluster, RKernel) { notebookPage =>
            val envVar = notebookPage.executeCell("Sys.getenv('KEY')")
            envVar shouldBe Some("'value'")
          }
        }
      }
```

rtitle (changes requested):
Looks good! Just 1 minor comment in the test. Other than that 💯

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookCustomizationSpec.scala
als364: Yes, can do

rtitle (approved):
:shipit: once tests pass

Qi77Qi (approved):

als364:
jenkins retest

als364:
@rtitle @Qi77Qi The swatomation test failure happened twice in a row which makes me think it's not just some flakey thing. I have no idea what [this failure](https://fc-jenkins.dsp-techops.broadinstitute.org/job/leonardo-fiab-test-runner/17659/testReport/junit/org.broadinstitute.dsde.workbench.leonardo.notebooks/NotebookCustomizationSpec/NotebookCustomizationSpec_should_populate_user_specified_environment_variables/) means. Have you ever seen something like this before?

rtitle:
I have an idea, can I add a commit to this PR?

rtitle:
I did, let's see how this run does :)

rtitle:
It finally passed! I'll leave it to you @als364 to merge when ready. Remember to Squash and merge.

PR #1122, opened 2019-10-28
[IA-1156] Move async cluster creation to back Leo

rtitle:
Rebased branch with the latest changes, including scalafmt. It all compiles, some unit tests still failing though. Working through it…

In all cases:


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/LeoRoutes.scala
```diff
@@ -113,31 +121,41 @@ abstract class LeoRoutes(
...
+                          .unsafeToFuture()
                       }
                     } ~
                     delete {
                       complete {
-                        leonardoService.deleteCluster(userInfo, GoogleProject(googleProject), clusterName).map { _ =>
-                          StatusCodes.Accepted
-                        }
+                        leonardoService
+                          .deleteCluster(userInfo, GoogleProject(googleProject), clusterName)
```
Qi77Qi, line 127: fyi, this can be `.deleteCluster(userInfo, GoogleProject(googleProject), clusterName).as(StatusCodes.Accepted)`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
```diff
@@ -333,6 +328,13 @@ trait ClusterComponent extends LeoComponent {
         unmarshalFullCluster(recs).headOption
       }
 
+    def getMinimalClusterById(id: Long): DBIO[Option[Cluster]] =
+      findByIdQuery(id).result.headOption.map { recOpt =>
+        recOpt map { rec =>
+          unmarshalCluster(rec, Seq.empty, List.empty, Map.empty, List.empty, List.empty, List.empty)
```
Qi77Qi: shall we create a new data type say `MinimalCluster` for this? Stubbing those fields out feels pretty weird to me. As caller of this function, I would expect all the data is from DB instead of some values being stubbed out

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
```diff
@@ -106,10 +105,10 @@ case class UserJupyterExtensionConfig(nbExtensions: Map[String, String] = Map(),
...
 case class ClusterResource(value: String) extends ValueObject
 
-case class DataprocInfo(googleId: Option[UUID],
-                        operationName: Option[OperationName],
-                        stagingBucket: Option[GcsBucketName],
-                        hostIp: Option[IP])
+case class DataprocInfo(googleId: Option[UUID] = None,
+                        operationName: Option[OperationName] = None,
+                        stagingBucket: Option[GcsBucketName] = None,
+                        hostIp: Option[IP] = None)
```
Qi77Qi: are these fields really can be any combination of `Some` and `None`? Or shouldn't `googleId` at least exist?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
```diff
@@ -169,19 +168,17 @@ object Cluster {
...
-             operation: Option[Operation] = None,
-             stagingBucket: Option[GcsBucketName] = None,
              clusterImages: Set[ClusterImage] = Set.empty): Cluster =
     Cluster(
       internalId = internalId,
       clusterName = clusterName,
       googleProject = googleProject,
       serviceAccountInfo = serviceAccountInfo,
-      dataprocInfo = DataprocInfo(operation.map(_.uuid), operation.map(_.name), stagingBucket, None),
+      dataprocInfo = DataprocInfo(),
```
Qi77Qi: It feels to be Cluster should have `dataprocInfo` field as an `Option`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/BucketHelper.scala
```diff
@@ -2,71 +2,56 @@ package org.broadinstitute.dsde.workbench.leonardo
...
-    } yield bucketName
+
+      readerAcl = NonEmptyList
+        .fromList(bucketSAs)
+        .map(readers => Map(StorageRole.ObjectViewer -> readers))
+        .getOrElse(Map.empty)
+      ownerAcl = Map(StorageRole.ObjectAdmin -> NonEmptyList.one(leoEntity))
+
+      _ <- google2StorageDAO.insertBucket(googleProject, bucketName)
+      _ <- google2StorageDAO.setIamPolicy(bucketName, readerAcl ++ ownerAcl)
```
Qi77Qi, line 53: 👍 🎉

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
```diff
@@ -24,19 +39,390 @@ case class ClusterIamSetupException(googleProject: GoogleProject)
...
+
+      // Resize the cluster in Google
+      _ <- IO.fromFuture(
+        IO(gdDAO.resizeCluster(cluster.googleProject, cluster.clusterName, numWorkers, numPreemptibles))
+      )
+    } yield ()
+
+  def setMasterMachineType(cluster: Cluster, machineType: MachineType): IO[Unit] =
+    cluster.instances.toList.traverse_ { instance =>
+      instance.dataprocRole match {
```
Qi77Qi: use `.traverse` instead of pattern matching?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
```diff
@@ -24,19 +39,390 @@ case class ClusterIamSetupException(googleProject: GoogleProject)
...
+  def generateServiceAccountKey(googleProject: GoogleProject,
+                                serviceAccountEmailOpt: Option[WorkbenchEmail]): IO[Option[ServiceAccountKey]] =
+    serviceAccountEmailOpt.traverse { email =>
+      IO.fromFuture(IO(googleIamDAO.createServiceAccountKey(googleProject, email)))
+    }
+
+  def removeServiceAccountKey(googleProject: GoogleProject,
+                              serviceAccountEmailOpt: Option[WorkbenchEmail],
+                              serviceAccountKeyIdOpt: Option[ServiceAccountKeyId]): IO[Unit] =
+    (serviceAccountEmailOpt, serviceAccountKeyIdOpt).mapN {
```
Qi77Qi, line 265: maybe `parMapN`? since the two requests can run at the same time?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
```diff
@@ -24,19 +39,390 @@ case class ClusterIamSetupException(googleProject: GoogleProject)
...
+                                            initBucketName: GcsBucketName,
+                                            serviceAccountInfo: ServiceAccountInfo,
+                                            serviceAccountKeyOpt: Option[ServiceAccountKey]): IO[Unit] = {
+    // Clean up resources in Google
+    val deleteBucket = (for {
+      _ <- bucketHelper.deleteInitBucket(initBucketName)
+      _ <- log.info(
+        s"Successfully deleted init bucket ${initBucketName.value} for ${googleProject.value} / ${clusterName.value}"
+      )
+    } yield ()).recoverWith {
```
Qi77Qi: would it be a bit cleaner if you do` .attempt`, and then pattern match on the result?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/TemplateHelper.scala
```diff
@@ -0,0 +1,40 @@
+package org.broadinstitute.dsde.workbench.leonardo.util
+
+import java.io.File
+
+import cats.effect._
+import fs2._
+import org.broadinstitute.dsde.workbench.leonardo.config.ClusterResourcesConfig
+import org.broadinstitute.dsde.workbench.leonardo.model.ClusterResource
+
+object TemplateHelper {
```
Qi77Qi, line 10: nice 👍

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/service/AuthProviderSpec.scala
```diff
@@ -121,15 +123,15 @@ class AuthProviderSpec
...
 
       // get status
-      val clusterStatus = leo.getActiveClusterDetails(userInfo, project, cluster1Name).futureValue
+      val clusterStatus = leo.getActiveClusterDetails(userInfo, project, cluster1Name).unsafeToFuture.futureValue
 
       cluster1 shouldEqual clusterStatus
 
       // list
-      val listResponse = leo.listClusters(userInfo, Map()).futureValue
+      val listResponse = leo.listClusters(userInfo, Map()).unsafeToFuture.futureValue
```
Qi77Qi, line 155: might as well just do `unsafeRunSync` in these places?

Qi77Qi (commented):
Did a first pass 👍  Looks great

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
rtitle: Yeah, that's probably a good idea. Will look into it.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/LeoRoutes.scala
```diff
@@ -113,31 +121,41 @@ abstract class LeoRoutes(
...
+                          .map { _ =>
+                            StatusCodes.Accepted
+                          }
+                          .unsafeToFuture()
```
Qi77Qi: shall we add something like https://github.com/broadinstitute/sam/blob/79417ba82faf5030ac98847940771fd941a801fb/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/api.scala to get rid of those explicit `unsafeToFuture` calls? I don't feel strongly either way..but just fyi

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
rtitle: `googleId` is populated asynchronously so doesn't need to exist.

Yeah it probably makes more sense for `dataprocInfo` to be optional instead.

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/service/AuthProviderSpec.scala
rtitle: I did `unsafeToFuture.futureValue` in all the tests. `.futureValue` is a ScalaTest thing which uses implicit `PatienceConfig`s, etc. Maybe we could make the tests use `IO` at some point but I didn't want to change them too much.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/LeoRoutes.scala
rtitle: Ah cool.

rtitle:
Thanks for the early feedback @Qi77Qi!

Unit tests finally passing, will check automation tests. Coveralls says:
```
coverage/coveralls — Coverage decreased (-1.4%) to 71.871%
```
which I find a bit suspicious because I removed a bunch of tests :)

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/service/AuthProviderSpec.scala
Qi77Qi: ah make sense...I didn't think of `PatienceConfig`...

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
rtitle: Actually there's only 1 `IO` here -- the `mapN` is unwrapping 2 `Option`s

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
rtitle: I'm thinking of maybe doing this in a separate PR…

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
```diff
@@ -324,29 +303,22 @@ object ClusterInitValues {
...
+            contentSecurityPolicy: String): ClusterInitValues =
     ClusterInitValues(
-      googleProject.value,
-      clusterName.value,
-      stagingBucketName.value,
-      clusterImages.find(_.tool == Jupyter).map(_.dockerImage).getOrElse(""),
-      clusterImages.find(_.tool == RStudio).map(_.dockerImage).getOrElse(""),
+      cluster.googleProject.value,
+      cluster.clusterName.value,
+      cluster.dataprocInfo.map(_.stagingBucket.value).getOrElse(""),
```
Qi77Qi: not necessarily this PR, but I feel like this function should return an `Either[String, ClusterInitValues]` and in the case where these required fields (such as stagingBucket) is `None` for some reason, we can error out

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
```diff
@@ -565,12 +540,15 @@ object LeonardoJsonSupport extends SprayJsonSupport with DefaultJsonProtocol {
...
-              fields.getOrElse("hostIp", JsNull).convertTo[Option[IP]]
-            ),
+            (fields.get("googleId"), fields.get("operationName"), fields.get("stagingBucket")).mapN {
+              (googleId, operationName, stagingBucket) =>
+                DataprocInfo(
+                  googleId.convertTo[UUID],
+                  operationName.convertTo[OperationName],
+                  stagingBucket.convertTo[GcsBucketName],
+                  fields.getOrElse("hostIp", JsNull).convertTo[Option[IP]]
+                )
```
Qi77Qi, line 551: semantic is a bit different between this and the original code...This code will return `None` if one of `googleId`, `operationName`, `stagingBucket` is `None`, vs the original logical seems to indicate these fields are not `None` or `Some` at the same time

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
rtitle: Yeah.. we currently just substitute with "" which may cause errors downstream. We could distinguish between required/optional fields and do some error checking up front.

Not going to do that in this PR, but good idea :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
rtitle: Yeah this is a semantic change.. `DataprocInfo` is optional now, so I was saying to only construct it if `googleId`, `operationName`, and `stagingBucket` are present in the input json (which are required fields in `DataprocInfo`). I think this is valid based on the data model.. and actually, this json reader is never actually used since we never serialize input JSON to `Cluster` objects.

rtitle:
fiab-start failed, looks unrelated to this PR. jenkins retest

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
Qi77Qi: cool cool...That's way better than what it was before...Just wanted to call out the semantic change explicitly :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
```diff
@@ -202,7 +202,7 @@ class ClusterMonitorActor(
         )
     } else {
       logger.info(
-        s"Cluster ${cluster.projectNameString} is not ready yet and has taken ${currTimeElapsed.toString} so far (Dataproc cluster status = $status, GCE instance statuses = ${instances
+        s"Cluster ${cluster.projectNameString} is not ready yet and has taken ${currTimeElapsed.millis.toSeconds} so far (Dataproc cluster status = $status, GCE instance statuses = ${instances
```
Qi77Qi: shall we add `seconds` after the variable? just so it's clear in the log what the unit is

Qi77Qi (approved):
:shipit:

Qi77Qi:
some automation tests failed

rtitle:
Thanks, yeah I see the bugs, will make a fix.

rtitle:
Note: I also fixed https://broadworkbench.atlassian.net/browse/IA-1443 as part of this PR.

rtitle:
fiab-start failed. jenkins retest

rtitle:
fiab-start failed again? It looks unrelated to Leo. 

jenkins retest please

rtitle:
All auto tests passed 🎉

http/src/main/resources/reference.conf
```diff
@@ -106,11 +106,12 @@ monitor {
   pollPeriod = 15 seconds
   maxRetries = -1  # means retry forever
   recreateCluster = true
-  creatingTimeLimit = 1 hour
-  startingTimeLimit = 20 minutes
-  stoppingTimeLimit = 1 hour
-  deletingTimeLimit = 1 hour
-  updatingTimeLimit = 20 minutes
+  # Defines timeouts for cluster status transitions. If a status is not listed there is no timeout.
```
kyuksel, line 109: Do we have documented somewhere what action is taken when a timeout is detected in each of the states below?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
```diff
@@ -175,40 +180,47 @@ class ClusterMonitorActor(
...
-
-      logger.info(
-        s"Detected that ${cluster.projectNameString} has been stuck in status $status too long. Failing it. Current timeout config: ${monitorConfig.monitorStatusTimeouts.toString}"
-      )
+    val currTimeElapsed: FiniteDuration = (System.currentTimeMillis() - startTime).millis
+    monitorConfig.monitorStatusTimeouts.get(cluster.status) match {
+      case Some(timeLimit) if currTimeElapsed > timeLimit =>
+        logger.info(s"Detected that ${cluster.projectNameString} has been stuck in status ${cluster.status} too long.")
+
+        // Take care not to Error out a cluster if it timed out in Starting status
```
kyuksel, line 197: Why are we no longer erroring out a cluster in this case?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorSupervisor.scala
```diff
@@ -115,15 +105,14 @@ class ClusterMonitorSupervisor(
...
-  leonardoService: LeonardoService,
   clusterHelper: ClusterHelper
 )(implicit metrics: NewRelicMetrics[IO], clusterToolToToolDao: ClusterTool => ToolDAO[ClusterTool])
     extends Actor
     with Timers
     with LazyLogging {
   import context.dispatcher
 
-  var monitoredClusters: Set[Cluster] = Set.empty
+  var monitoredClusterIds: Set[Long] = Set.empty
```
kyuksel, line 115: I know this PR didn't introduce it but not having any implicit or explicit protection on this `var` against race conditions is a bit concerning, no? Doing something like in `ClusterDnsCache.hostToIpMapping` may be an easy remedy.

```scala
private val monitoredClusterIds: ValueBox[Set[Long]] = ValueBox(Set.empty)
```

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -986,6 +986,9 @@ definitions:
         items:
           type: string
         description: The scopes for the cluster.
+      customClusterEnvironmentVariables:
```
Qi77Qi, line 989: is this change related to this PR?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -190,134 +159,48 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
                                                                  cluster.clusterName)
       _ <- hasPermission match {
         case false =>
-          IO(
-            logger.warn(
+          log
```
kyuksel, line 161: wow..scalafmt took it to another level here 🙈

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
```diff
@@ -446,12 +454,17 @@ trait ClusterComponent extends LeoComponent {
...
+          cluster.dataprocInfo.map(_.googleId),
+          cluster.dataprocInfo.map(_.operationName.value),
+          cluster.dataprocInfo.map(_.stagingBucket.value),
           Timestamp.from(Instant.now)
         )
 
+    def clearAsyncClusterCreationFields(cluster: Cluster): DBIO[Int] =
+      findByIdQuery(cluster.id)
+        .map(c => (c.initBucket, c.serviceAccountKeyId, c.googleId, c.operationName, c.stagingBucket, c.dateAccessed))
+        .update(None, None, None, None, None, Timestamp.from(Instant.now))
```
Qi77Qi: `Instant.now` is side effecting..So ideally it should be passed in as parameter

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorSupervisor.scala
Qi77Qi, line 115: should this be `ClusterInternalId`? since we've changed clusterId to uuid recently

http/src/main/resources/swagger/api-docs.yaml
rtitle: This field was added in a previous PR, but I noticed it was not added to swagger. I also had to persist it in the database in this PR, since back Leo is doing the cluster creation.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
rtitle: Yeah.. or use a DB function like `current_timestamp`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorSupervisor.scala
rtitle: `ClusterInternalId` is a UUID, but we still have `id` which is a `Long`. Either should work here..

http/src/main/resources/reference.conf
rtitle: No, but I can create a wiki page..

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
rtitle: Because `Error` is an unrecoverable status. We don't want a user to start their cluster; have it time out; then have Leo error out (and delete) their cluster, potentially losing data.

This code transitions the cluster back to `Stopped`, so the user can try again or create a ticket -- either way, we won't delete their data.

http/src/main/resources/reference.conf
kyuksel: Either that (with a link to it here) or we could add that info directly here

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
kyuksel: got it

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorSupervisor.scala
rtitle: I think this is safe -- this is an actor, and changes to `monitoredClusterIds` is only done in the actor context (not by a future). Actors provide a safe encapsulation of state, so mutable `var`s are allowed. I agree it's probably easy to access this `var` in an unsafe way, which is one of the things that bothers me about akka..

http/src/main/resources/reference.conf
rtitle: Good idea, will add here.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
rtitle: Slick apparently doesn't support updates with non-"pure" values. So you can't do:
```
update TABLE set col = current_timestamp
```
I instead changed these methods to take an `Instant` as a parameter, and had callers do stuff like:
```
for {
  now <- IO(Instant.now)
  _ <- dbRef.inTransactionIO { _.clusterQuery. clearAsyncClusterCreationFields(cluster, now) }
} yield ()
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -1121,183 +645,54 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
...
+            log.error(e)(
               s"User ${userEmail.value} does not have access to ${gcsPath.bucketName} / ${gcsPath.objectName}"
-            )
-            throw BucketObjectAccessException(userEmail, gcsPath)
-          case e if whenGoogle401(e) =>
-            logger.warn(s"Could not validate object [${gcsUri}] as user [${userEmail.value}]", e)
-            ()
+            ) >> IO.raiseError(BucketObjectAccessException(userEmail, gcsPath))
+          case e if when401(e) =>
+            log.warn(e)(s"Could not validate object [${gcsUri}] as user [${userEmail.value}]") >> IO.unit
```
Qi77Qi: `IO.unit` is not needed since `log.warn` returns `IO.unit` I think

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/BucketHelper.scala
```diff
@@ -76,110 +61,55 @@ class BucketHelper(
...
-   */
-  def updateUserBucket(bucketName: GcsBucketName,
-                       googleProject: GoogleProject,
-                       serviceAccountInfo: ServiceAccountInfo): Future[Unit] =
-    for {
-      // available service accounts ((cluster or default SA) and notebook SA, if they exist) -> Reader
-      bucketSAs <- getBucketSAs(googleProject, serviceAccountInfo)
-
-      _ <- setBucketAcls(bucketName, bucketSAs, List.empty)
+      providerReaders <- Stream.eval(
```
Qi77Qi, line 71: just want to point out, that if `providerReaders` is empty, then everything after this line won't happen. Is this intended with `insertBucket` and `setIamPolicy`?

rtitle: Hmm… good catch, this is probably not intended. We should be `++`ing the readers and groups. Will fix.

rtitle: Wait.. am I missing something?
```
providerReaders <- Stream.eval(serviceAccountProvider.listUsersStagingBucketReaders(userEmail).map(_.map(userIdentity)))
providerGroups <- Stream.eval(serviceAccountProvider.listGroupsStagingBucketReaders(userEmail).map(_.map(groupIdentity)))
```
These would still evaluate if `providerReaders` and `providerGroups` are empty right? The stream always contains 1 element, even if that element is `List.empty`.

rtitle:
```
scala> for {
     | a <- Stream.eval(IO(List.empty))
     | b <- Stream.eval(IO(List(1,2,3)))
     | } yield a ++ b
res9: fs2.Stream[[x]cats.effect.IO[x],List[Int]] = Stream(..)

scala> res9.compile.last.unsafeRunSync
res10: Option[List[Int]] = Some(List(1, 2, 3))
```

Qi77Qi: oh my bad, I think this is fine...I thought `Stream.eval(serviceAccountProvider.listUsersStagingBucketReaders(userEmail).map(_.map(userIdentity)))` is creating a `Stream[IO, someReader]`, but it's actually `Stream[IO, List[someReader]]`

I was thinking something like
```
for {
    a <- Stream.emits(List.empty[Int]).covary[IO]
    b <- Stream.eval(IO(List(1,2,3)))
  } yield a :: b
res4: Stream[IO[x], List[Int]] = Stream(..)

@ res4.compile.toList.unsafeRunSync
res5: List[List[Int]] = List()
```

rtitle: Right.. that makes sense

PR #1123, opened 2019-10-28
[IA-1325] Support building multiple custom leo images with different dataproc versions

akarukappadath:

In all cases:

### DIFF
--- a/jenkins/Jenkinsfile
+++ b/jenkins/Jenkinsfile
@@ -128,9 +128,9 @@ pipeline {
           description: "The google project to use")
     string(name: "BUCKET_NAME", defaultValue: "gs://leo-dataproc-image",
           description: "The bucket where the image will be stored")
+    string(name: "dataproc_version", defaultValue: ["1.2.79-debian9", "1.4.15-debian9"],
+          description: "A custom image will be build for each of these dataproc versions")
     string(name: "terra_jupyter_base", defaultValue: "0.0.1")
-    string(name: "spark_version", defaultValue: "0.0.1",
-              description: "Is not currently baked into the image, but is here for future functionality")
     string(name: "terra_jupyter_hail", defaultValue: "0.0.1",
           description: "Is not currently baked into the image, but is here for future functionality")
     string(name: "terra_jupyter_r", defaultValue: "0.0.3")
@@ -181,7 +181,7 @@ pipeline {
             ))
           }
 
-          currentBuild.displayName = "#${BUILD_NUMBER}: $IMAGE_BASE_NAME-$imageID"
+          currentBuild.displayName = "#${BUILD_NUMBER}: $IMAGE_BASE_NAME-$dataproc_version-$imageID"
         }
       }
     }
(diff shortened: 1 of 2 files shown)


jenkins/Jenkinsfile
```diff
@@ -268,7 +268,7 @@ pipeline {
             cd jenkins/dataproc-custom-images/dataproc-custom-images
             python generate_custom_image.py \
               --image-name "$IMAGE_BASE_NAME-$imageID" \
-              --dataproc-version "1.2.79-debian9" \
+              --dataproc-version "1.4.15-debian9" \
```
Qi77Qi: shall we upgrade python version as well?

@rtitle shall we make the script to support both old and new versions?

rtitle: Yeah.. we'll have to maintain both old and new versions for a little while. Maybe this Jenkins job could take as an input parameter which dataproc version to use.

rtitle: Btw, it would be good to revisit this code when upgrading Hail: https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/google/HttpGoogleDataprocDAO.scala#L329-L350

I bet we could remove the `"spark:spark.yarn.am.memory"` setting. Also we currently set the image to `1.2-deb9` if the custom image is empty. Not sure how to handle this… should we just require that a custom image is used going forward?

Qi77Qi: I think it makes sense to always use custom dataproc image going forward

rtitle (approved):
Looks good to me. I see some test images here: https://console.cloud.google.com/compute/images?project=broad-dsp-gcr-public&tab=images&imagessize=50

jenkins/Jenkinsfile
```diff
@@ -264,32 +265,43 @@ pipeline {
...
-              --customization-script ../prepare-custom-leonardo-jupyter-dataproc-image.sh \
-              --zone us-central1-a \
-              --gcs-bucket $BUCKET_NAME \
-              --project-id=$GOOGLE_PROJECT \
-              --disk-size=30
+              git submodule init && git submodule update
           """
+          script {
+            for(String dp_version: dataproc_versions.trim().split(' ')) {
+              def dp_version_formatted = dp_version.replace(".","-")
```
Qi77Qi, line 272: what does the transforming for? 🤔

akarukappadath: We can't put periods in the image name, or else you get this: 
```ERROR: (gcloud.compute.disks.create) Could not fetch resource:
 - Invalid value for field 'resource.name': 'custom-leo-image-1.2.79-debian9-eafb4cb-install'. Must be a match of regex '(?:[a-z](?:[-a-z0-9]{0,61}[a-z0-9])?)'```

Qi77Qi (commented):
:shipit:

Qi77Qi:
curious, does automation tests passing mean anything for this change? Just wonder if we're testing the newly built image somewhere

akarukappadath:
No, we're not currently testing because I think we would need https://broadworkbench.atlassian.net/browse/IA-1427 in place first for all the leo changes

PR #1124, opened 2019-10-28
IA-1427 Use both legacyDataproc image and new dataproc image

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-1427

1. Previously listCluster returns an object that’s kinda shared in leo internal implementation, which means whenever we make changes to this `Cluster` case class, it’s very easy to accidentally introduce new fields or rename fields in `listCluster`’s response…I want to move away from that.

For **listCluster** API:
* userJupyterExtensionConfig
* scopes
* errors

These fields are removed because they were previously stubbed with dummy data, and shouldn't be used

For both **listCluster** and **getCluster** API:
`tool` and `dockerImage` fields under `clusterImages` field are renamed to `imageType` and `imageUrl`

Confirmed with AOU that the change is okay

2. Support 2 custom dataproc images
3. Reenable hailspec

TODO:
1. Confirm with terra-ui the api change is okay

---------

In all cases:


automation/src/test/resources/reference.conf
```diff
@@ -1,6 +1,6 @@
 leonardo.rImageUrl = "us.gcr.io/broad-dsp-gcr-public/terra-jupyter-r:0.0.4"
 leonardo.pythonImageUrl = "us.gcr.io/broad-dsp-gcr-public/terra-jupyter-python:0.0.3"
-leonardo.hailImageUrl = "us.gcr.io/broad-dsp-gcr-public/terra-jupyter-hail:0.0.1"
+leonardo.hailImageUrl = "us.gcr.io/broad-dsp-gcr-public/terra-jupyter-hail:qi-test-7"
```
Qi77Qi: will update once we have a new hail image

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -987,6 +987,82 @@ definitions:
           type: string
         description: The scopes for the cluster.
 
+  ListClusterResponse:
```
Qi77Qi: the only difference is removing `errors` field which isn't used by anyone. And it's value was misleading

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookClusterMonitoringSpec.scala
```diff
@@ -159,7 +159,7 @@ class NotebookClusterMonitoringSpec extends GPAllocFixtureSpec with ParallelTest
           .futureValue
 
         val request = defaultClusterRequest.copy(
-          jupyterDockerImage = None,
+          jupyterDockerImage = Some(LeonardoConfig.Leonardo.hailImageUrl),
```
akarukappadath: why are we using the hail image instead of base?

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20191028_rename_cluster_image.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="leonardo" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+    <changeSet logicalFilePath="leonardo" author="qi" id="rename-tool">
+        <renameColumn newColumnName="name" oldColumnName="tool" tableName="CLUSTER_IMAGE"/>
+    </changeSet>
+    <changeSet logicalFilePath="leonardo" author="qi" id="rename-docker-image">
+        <renameColumn newColumnName="image" oldColumnName="dockerImage" tableName="CLUSTER_IMAGE"/>
```
akarukappadath: would it be more clear to call this `image_url`?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -912,6 +941,10 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
       else Future.successful(Map.empty[String, String])
       clusterVPCSettings = getClusterVPCSettings(projectLabels)
 
+      // If user is using https://github.com/DataBiosphere/terra-docker/tree/master#terra-base-images for jupyter image, then
+      // we will use the new custom dataproc image
+      dataprocImage = if(clusterImages.exists(_.dockerImage.contains("terra-jupyter"))) dataprocConfig.customDataprocImage else dataprocConfig.legacyCustomDataprocImage
```
akarukappadath: don't the terra-docker hail images 0.1 and 0.2 depend on old spark?

Qi77Qi: I don’t think hail 0.1 and 0.2 actually worked. So i think we can assume 0.1 and 0.2 should never be used

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookClusterMonitoringSpec.scala
Qi77Qi: This test is testing some hail method. We can probably just use base and check something else

rtitle: Yeah, I feel like this test should not use hail. Reading below, it looks like Hail verification is disabled anyway.

http/src/main/resources/jupyter/init-vm.sh
```diff
@@ -1,170 +0,0 @@
-#!/usr/bin/env bash
```
rtitle, line 1: 👍 . I guess this logic now just lives here: https://github.com/DataBiosphere/terra-docker/blob/master/dataproc-custom-images/prepare-custom-leonardo-jupyter-dataproc-image.sh

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20191120_rename_cluster_image.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="leonardo" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+    <changeSet logicalFilePath="leonardo" author="qi" id="rename-tool-to-imageType">
```
rtitle: Minor; could both `renameColumn`s be in the same `changeSet`?

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -119,7 +119,7 @@ paths:
           schema:
             type: array
             items:
-              $ref: '#/definitions/Cluster'
+              $ref: '#/definitions/ListClusterResponse'
```
rtitle, line 122: There's another `listCluster` API to change -- the one that's like `/api/clusters/{googleProject}`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/Boot.scala
```diff
@@ -13,13 +13,7 @@ import com.typesafe.sslconfig.ssl.ConfigSSLContextBuilder
...
 import io.chrisdavenport.log4cats.slf4j.Slf4jLogger
 import org.broadinstitute.dsde.workbench.google.GoogleCredentialModes.{Pem, Token}
-import org.broadinstitute.dsde.workbench.google.{
-  GoogleStorageDAO,
-  HttpGoogleDirectoryDAO,
-  HttpGoogleIamDAO,
-  HttpGoogleProjectDAO,
-  HttpGoogleStorageDAO
-}
+import org.broadinstitute.dsde.workbench.google.{GoogleStorageDAO, HttpGoogleDirectoryDAO, HttpGoogleIamDAO, HttpGoogleProjectDAO, HttpGoogleStorageDAO}
```
rtitle: scalafmt might change this back :)

Qi77Qi: yea...I haven't run scalafmt yet

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20191120_rename_cluster_image.xml
Qi77Qi: will give it a try

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookClusterMonitoringSpec.scala
Qi77Qi: yea, I wanted to check with you before I remove the hail part..is it okay to change this to base image, and what would be a good validation command? if nothing in particular, I'll just use something random

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/LeoRoutes.scala
```diff
@@ -188,21 +190,22 @@ abstract class LeoRoutes(
     ExceptionHandler {
       case requestValidationError: RequestValidationError =>
         complete(StatusCodes.BadRequest, requestValidationError.getMessage)
+      case e: ClusterNotFoundException =>
```
rtitle: Just curious, why add this? I think `ClusterNotFoundException` is an instance of `LeoException`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
```diff
@@ -106,19 +106,37 @@ case class AuditInfo(creator: WorkbenchEmail,
                      dateAccessed: Instant,
                      kernelFoundBusyDate: Option[Instant])
 
-sealed trait ClusterTool extends EnumEntry with Serializable with Product
-object ClusterTool extends Enum[ClusterTool] {
+sealed trait ClusterContainerServiceType extends EnumEntry with Serializable with Product {
+  def imageType: ClusterImageType
+}
+object ClusterContainerServiceType extends Enum[ClusterContainerServiceType] {
```
rtitle, line 112: I'm curious why we need both `ClusterContainerServiceType` and `ClusterImageType` enums -- they seem duplicative of each other. When do we use one v. the other?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
```diff
@@ -525,62 +564,13 @@ object LeonardoJsonSupport extends SprayJsonSupport with DefaultJsonProtocol {
...
 
-  implicit val ClusterToolFormat = EnumEntryFormat(ClusterTool.withName)
+  implicit val ClusterToolFormat = EnumEntryFormat(ClusterImageType.withName)
 
   implicit val ClusterImageFormat = jsonFormat3(ClusterImage.apply)
 
   implicit val ClusterInternalIdFormat = ValueObjectFormat(ClusterInternalId)
 
-  implicit object ClusterFormat extends RootJsonFormat[Cluster] {
-    override def read(json: JsValue): Cluster =
```
rtitle, line 535: +1 we don't need `read` for `ClusterFormat`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
```diff
@@ -386,29 +382,36 @@ class ClusterMonitorActor(
...
+                                                        clusterResult.serviceAccountKey,
+                                                        clusterResult.cluster,
                                                         now)
       }
+      clusterImage = ClusterImage(ClusterImageType.CustomDataProc, clusterResult.customDataprocImage.asString, now)
+      // Save dataproc image in the database
+      _ <- dbRef.inTransactionIO(
+        _.clusterImageQuery.save(cluster.id, clusterImage)
+      ).handleErrorWith {
+        case e: java.sql.SQLIntegrityConstraintViolationException if(e.getMessage.contains("Duplicate entry")) => // This could happen when we recreate a cluster
```
rtitle: Catching the exception and pattern matching the message string may not be the best approach.. I feel like a SQL SELECT then INSERT if not present would be better.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/leonardoServiceContractModel.scala
```diff
@@ -0,0 +1,71 @@
+package org.broadinstitute.dsde.workbench.leonardo.service
```
rtitle, line 1: I like this class -- should it go in `model` instead of `service`?

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20191028_rename_cluster_image.xml
Qi77Qi: renamed to `imageUrl`

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/api/RoutesTestJsonSupport.scala
```diff
@@ -0,0 +1,107 @@
...
+import cats.implicits._
+import spray.json._
+import org.broadinstitute.dsde.workbench.leonardo.model.LeonardoJsonSupport._
+import org.broadinstitute.dsde.workbench.model.google.GoogleModelJsonSupport.{GcsPathFormat => _, _}
+import org.broadinstitute.dsde.workbench.model.WorkbenchIdentityJsonSupport._
+import org.broadinstitute.dsde.workbench.leonardo.model.google.GoogleJsonSupport._
+import org.broadinstitute.dsde.workbench.leonardo.service.ListClusterResponse
+import spray.json.DefaultJsonProtocol
+
+object RoutesTestJsonSupport extends DefaultJsonProtocol{
```
rtitle, line 21: This is needed for tests that deserialize json to a `Cluster` object, right?

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookClusterMonitoringSpec.scala
rtitle: Hm.. no strong opinion, something random is fine. This test is verifying that a cluster w/ preemptibles can be paused/resumed successfully, not any particular hail functionality.

rtitle:
Done with first pass, a few comments above

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/LeoRoutes.scala
Qi77Qi: I was suspicious that this error was causing 500, so tried making it explicit this is returning 400...But it might've been a misunderstanding..I can remove this for now

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
Qi77Qi: the main difference is `ClusterImageType` can include `customDataproc`, vs `ClusterContainerServiceType` doesn't include that

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/leonardoServiceContractModel.scala
Qi77Qi: These contract specific models don't need to be shared outside service, so I kinda prefer they stay in `service` and maybe we can even make them package private 🤔

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/api/RoutesTestJsonSupport.scala
Qi77Qi: yea..these are only needed in the test. I didn't want to pollute main app code with this.

akarukappadath:
It looks like Terra-UI does use the `id` field for clusters (ex: https://github.com/DataBiosphere/terra-ui/blob/6997a831d94796ec0f75be616c8ee75832e732bd/src/pages/Clusters.js#L99). We should change Terra-UI to use something else if we want to keep `id` from users, or keep it in the `listClusters` response

Qi77Qi:
@akarukappadath thanks for checking! I added `id` back for now

http/src/main/resources/reference.conf
```diff
@@ -25,7 +25,9 @@ dataproc {
   # Unset to use project-level defaults. Note a VPC/subnet with the literal name "default" may not exist.
   #vpcNetwork = "default"
   #vpcSubnet = "default"
-  customDataprocImage = "projects/broad-dsp-gcr-public/global/images/custom-leo-image-dbb0929"
+  # This image is used for legacy jupyter image where hail is compatible with earlier version of spark
+  legacyCustomDataprocImage = "projects/broad-dsp-gcr-public/global/images/custom-leo-image-dbb0929"
+  customDataprocImage = "projects/broad-dsp-gcr-public/global/images/custom-leo-image-1-4-15-debian9-test3" # TODO: update this with a valid image
```
rtitle, line 30: Just calling out this TODO - should be addressed before merge, right?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
```diff
@@ -112,13 +112,17 @@ class ClusterHelper(
...
-            List(clusterResourcesConfig.initVmScript, clusterResourcesConfig.initActionsScript)
-          else List(clusterResourcesConfig.initActionsScript)
+          initScriptResources = List(clusterResourcesConfig.initActionsScript)
           initScripts = initScriptResources.map(resource => GcsPath(initBucketName, GcsObjectName(resource.value)))
           credentialsFileName = cluster.serviceAccountInfo.notebookServiceAccount
             .map(_ => s"/etc/${ClusterInitValues.serviceAccountCredentialsFilename}")
 
+          // If user is using https://github.com/DataBiosphere/terra-docker/tree/master#terra-base-images for jupyter image, then
+          // we will use the new custom dataproc image
+          dataprocImage = if (cluster.clusterImages.exists(_.imageUrl.contains("terra-jupyter")))
```
rtitle: Wondering if we should invert this?
```
if (cluster.clusterImages.exists(_.imageUrl == dataprocConfig.jupyterImage))
  dataprocConfig.legacyCustomDataprocImage
else dataprocConfig.customDataprocImage
```
That way the new one is the default, and we only use the legacy for the legacy image.

rtitle:
Just a couple comments above… otherwise looks good 👍.

Also automation tests seem a bit unhappy..

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
Qi77Qi: I think that wouldn't work if somehow user uses an older legacy image than what's in config. But I agree it's probably better than what i have here.

http/src/main/resources/reference.conf
Qi77Qi: @akarukappadath , can we use this image in production? Otherwise, I can try to build another usable image outside this PR.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
rtitle: True.. that's probably not very likely though

rtitle (approved):
:shipit: if tests pass

Qi77Qi:
jenkins retest

Qi77Qi:
jenkins retest

Qi77Qi:
jenkins retest

PR #1128, opened 2019-10-30
[PROD-292] listClusters returns empty list for AoU service account

rtitle:
See https://broadworkbench.atlassian.net/browse/PROD-292

I think this was introduced in the Sam client upgrade. The bug is in `listClusters`. The intended behavior is for `listClusters` to return clusters the caller created; and clusters in projects which the caller is an owner of. As a result of this bug we were only returning the former, not the latter.

The fix is pretty trivial: we essentially got the types wrong on a `.contains` call. See inline.

Also added a unit test for this case.

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
@@ -800,12 +800,12 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
     } yield {
       // Making the assumption that users will always be able to access clusters that they create
       // Fix for https://github.com/DataBiosphere/leonardo/issues/821
-      val visibleClusters = samVisibleClusters ::: clusterList
+      val userCreatedClusters: List[(GoogleProject, ClusterInternalId)] = clusterList
         .filter(_.auditInfo.creator == userInfo.userEmail)
-        .map(c => (c.googleProject, c.clusterName))
+        .map(c => (c.googleProject, c.internalId))
         .toList
-      val visibleClustersSet = visibleClusters.toSet
-      clusterList.filter(c => visibleClustersSet.contains((c.googleProject, c.clusterName)))
+      val visibleClustersSet = (samVisibleClusters ::: userCreatedClusters).toSet
+      clusterList.filter(c => visibleClustersSet.contains((c.googleProject, c.internalId)))
     }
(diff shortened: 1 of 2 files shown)


Qi77Qi (approved):

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -800,12 +800,12 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
     } yield {
       // Making the assumption that users will always be able to access clusters that they create
       // Fix for https://github.com/DataBiosphere/leonardo/issues/821
-      val visibleClusters = samVisibleClusters ::: clusterList
```
rtitle, line 803: Here's the bug. We were essentially `:::`ing a `List[(GoogleProject, ClusterInternalId)]` with a `List[(GoogleProject, ClusterName)]`. This compiled, and the inferred type was `List[(GoogleProject, Product with Serializable)]`. We were then doing a `.contains` check on that, which was failing.

I reorganized the code a bit and tried to make the types explicit. The compiler now complains if we pass the wrong type to `.contains`.

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/service/AuthProviderSpec.scala
```diff
@@ -295,6 +296,20 @@ class AuthProviderSpec
       listResponse shouldBe Seq(cluster1).map(stripFieldsForListCluster)
     }
 
+    "should return clusters the user didn't create if the auth provider says yes" in isolatedDbTest {
```
rtitle, line 299: I validated this test fails before the fix, and passes after the fix.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
Qi77Qi: not enough type safety :)

rtitle: Yup

rtitle:
selenium timeout `from unknown error: web view not found`

jenkins retest

Qi77Qi:
thanks for fixing this!

PR #1130, opened 2019-10-30
[IA-1364] Use Google Groups for enabling users to pull the custom Dataproc image

kyuksel:
See [IA-1364](https://broadworkbench.atlassian.net/browse/IA-1364) for the motivation and the implementation plan.

**Remaining work:**

- [x] template the Google group email and add it to the config
- [x] make sure the Leo SA was delegated domain-wide authority in `quality` and `prod` domains (in addition to `test`) and grant it the right IAM roles
- [x] figure out why [one of the unit tests](https://github.com/DataBiosphere/leonardo/pull/1130/files#r340811150) is failing sporadically
- [x] manually fiab-test boot and cluster creation/deletion
- [x] make sure the integration tests pass
- [x] perf-test it
- [ ] visually verify that [this ticket](https://broadworkbench.atlassian.net/browse/IA-1486) was fulfilled, to make sure the changes will work in prod
- [x] create a ticket to remove the IAM roles previously added by the old code (which should no longer be necessary when this ticket is done) from the project `broad-dsp-gcr-public` once we make sure the changes are working as expected


http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorSpec.scala
```diff
@@ -704,7 +730,10 @@ class ClusterMonitorSpec
         mockitoEq(deletingCluster.googleProject),
         mockitoEq(deletingCluster.clusterName)
       )(any[ApplicativeAsk[IO, TraceId]])
-      verify(iamDAO, times(1)).removeIamRolesForUser(any[GoogleProject], any[WorkbenchEmail], any[Set[String]])
+      verify(iamDAO, times(1)).removeIamRoles(any[GoogleProject],
+                                              any[WorkbenchEmail],
+                                              mockitoEq(MemberType.ServiceAccount),
+                                              any[Set[String]])
```
kyuksel, line 745: ^^ fails sporadically with

```
[info] - should monitor until DELETED state *** FAILED ***
[info]   org.mockito.exceptions.verification.WantedButNotInvoked: Wanted but not invoked:
[info] googleIamDAO.removeIamRoles(
[info]     <any>,
[info]     <any>,
[info]     serviceAccount,
[info]     <any>
[info] );
[info] -> at org.broadinstitute.dsde.workbench.leonardo.monitor.ClusterMonitorSpec.$anonfun$new$45(ClusterMonitorSpec.scala:699)
[info] Actually, there were zero interactions with this mock.
```

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/api/TestLeoRoutes.scala
```diff
@@ -31,6 +37,14 @@ trait TestLeoRoutes { this: ScalatestRouteTest with Matchers with CommonTestData
   implicit private val nr = FakeNewRelicMetricsInterpreter
   implicit def unsafeLogger = Slf4jLogger.getLogger[IO]
 
+  val mockGoogleDirectoryDAO = new MockGoogleDirectoryDAO()
+  Await.result(
+    mockGoogleDirectoryDAO.createGroup(dataprocImageProjectGroupName,
+                                       dataprocImageProjectGroupEmail,
+                                       Option(mockGoogleDirectoryDAO.lockedDownGroupSettings)),
+    Duration.Inf
+  )
```
kyuksel: This is to create the Google group ahead of testing. 

Does it make sense to factor it out a bit to a `def` like

```scala
def runAndWaitFor[T](f: Awaitable[T]): T = Await.result(f, Duration.Inf)
```
so it can be called as 
```scala
runAndWaitFor(mockGoogleDirectoryDAO.createGroup(
                   dataprocImageProjectGroupName,
                   dataprocImageProjectGroupEmail,
                   Option(mockGoogleDirectoryDAO.lockedDownGroupSettings)))
```
and put the `def` in a new `TestSupport` trait or something? Sam does this.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/Boot.scala
```diff
@@ -47,119 +50,142 @@ import scala.concurrent.duration._
...
-                throw t
+                                     appDependencies.dbReference,
+                                     appDependencies.metrics)
+          )
+
+          // TODO Template and get the group and domain values from config
+          val dpImageUserGoogleGroupName = "kyuksel-test-dataproc-image-group-11"
+          val dpImageUserGoogleGroupEmail = WorkbenchEmail(s"$dpImageUserGoogleGroupName@test.firecloud.org")
+          val ch = appDependencies.clusterHelper
+          ch.createDataprocImageUserGoogleGroupIfItDoesntExist(appDependencies.googleDirectoryDAO,
```
rtitle: Could we pass `googleDirectoryDAO` in the constructor for `ClusterHelper` instead?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/Boot.scala
```diff
@@ -47,119 +50,142 @@ import scala.concurrent.duration._
...
+                                                               dpImageUserGoogleGroupName,
+                                                               dpImageUserGoogleGroupEmail)
+            .flatMap { _ =>
+              ch.addDataprocImageUserIamRole(ch.googleIamDAO,
```
rtitle: Same here -- no need to pass `googleIamDAO` in the method, `ClusterHelper` already has one via the constructor.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/Boot.scala
```diff
@@ -220,7 +256,14 @@ final case class AppDependencies[F[_]](google2StorageDao: GoogleStorageService[F
                                        clusterDnsCache: ClusterDnsCache,
                                        googleComputeDAO: HttpGoogleComputeDAO,
                                        googleDataprocDAO: HttpGoogleDataprocDAO,
+                                       googleDirectoryDAO: HttpGoogleDirectoryDAO,
                                        samDAO: HttpSamDAO[F],
                                        welderDAO: HttpWelderDAO[F],
                                        clusterHelper: ClusterHelper,
                                        metrics: NewRelicMetrics[F])
+
+final case class GoogleGroupCreationException(googleGroup: WorkbenchEmail)
```
rtitle: Maybe these exceptions should live in `ClusterHelper` instead since that's where they're thrown from

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/Boot.scala
rtitle: Also it looks like this future is fire-and-forget… we should probably wait for it to complete before starting the app (and fail to start the app if the future fails), right?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
```diff
@@ -26,27 +28,28 @@ class ClusterHelper(
   dataprocConfig: DataprocConfig,
   gdDAO: GoogleDataprocDAO,
   googleComputeDAO: GoogleComputeDAO,
-  googleIamDAO: GoogleIamDAO
+  googleDirectoryDAO: GoogleDirectoryDAO,
+  val googleIamDAO: GoogleIamDAO
```
rtitle: Does it need to be a `val`?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
```diff
@@ -59,45 +62,39 @@ class ClusterHelper(
...
-    //
-    // Add the Compute Image User role in the image project to the Google API service account.
-    // This is needed in order to use a custom dataproc VM image.
-    // If a custom image is not being used, this is not necessary.
+    // TODO Template and get the group and domain values from config
+    val dataprocImageUserGoogleGroupName = "kyuksel-test-dataproc-image-group-11"
+    val dataprocImageUserGoogleGroupEmail = WorkbenchEmail(s"$dataprocImageUserGoogleGroupName@test.firecloud.org")
+
+    // Add the user's service account to the Google group that has compute.imageUser role on the custom Dataproc image project.
     val computeImageUserIO = dataprocConfig.customDataprocImage.flatMap(parseImageProject) match {
```
rtitle: Maybe we should split the computeImageUser group stuff to a new method because it's not really manipulating IAM anymore (so doesn't feel like it belongs in `updateClusterIamRoles`).

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
```diff
@@ -127,6 +119,74 @@ class ClusterHelper(
         IO.fromFuture(IO(googleIamDAO.removeServiceAccountKey(googleProject, email, keyId)))
     } getOrElse IO.unit
 
+  def createDataprocImageUserGoogleGroupIfItDoesntExist(googleDirectoryDAO: GoogleDirectoryDAO,
+                                                        dpImageUserGoogleGroupName: String,
+                                                        dpImageUserGoogleGroupEmail: WorkbenchEmail): Future[Unit] = {
```
rtitle: I've been trying to standardize on `IO` in this class (and in my other PR, `LeonardoService`). Let's try to use IO everywhere instead of Future going forward.

http/src/test/resources/reference.conf
```diff
@@ -93,6 +93,11 @@ sam {
 
 google {
   leoServiceAccountJsonFile = "fakePath"
+  groups {
+    // TODO Change the default values
```
rtitle: This will eventually need to be in firecloud-develop I think because we need to template environment-specific values

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/api/TestLeoRoutes.scala
```diff
@@ -31,6 +37,14 @@ trait TestLeoRoutes { this: ScalatestRouteTest with Matchers with CommonTestData
   implicit private val nr = FakeNewRelicMetricsInterpreter
   implicit def unsafeLogger = Slf4jLogger.getLogger[IO]
 
+  val mockGoogleDirectoryDAO = new MockGoogleDirectoryDAO()
+  Await.result(
```
rtitle: Btw, I try to prefer `future.futureValue` over `Await.result(future, duration)` in unit tests. `futureValue` comes from ScalaTest and has some improvements around polling, timeouts, error messages, etc.

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/api/TestLeoRoutes.scala
rtitle: Ah see above comment -- use ScalaTest `.futureValue`

rtitle:
Done with first pass, some comments/suggestions inline.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/Boot.scala
kyuksel: Ah...these and a few others you pointed out below are left-overs from some previous method relocations. Thanks for the catches!

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/Boot.scala
kyuksel: 👍

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
kyuksel: not anymore 😄

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/Boot.scala
kyuksel: 👍

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
kyuksel: Totally!

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/Boot.scala
kyuksel: Absolutely!

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
kyuksel: IO for the win 👍

http/src/test/resources/reference.conf
kyuksel: Will keep one here for unit testing purposes but get the env-specific ones for the non-test code form firecloud-develop as you mentioned.

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/api/TestLeoRoutes.scala
kyuksel: Perfect!

kyuksel (commented):
@rtitle I've addressed your comments. Debugging a few `LeonardoServiceSpec` unit tests that started failing after the changes. It looks like cluster creation doesn't error out when it should:

```
[info] - should cluster creation should end in Error state if adding IAM roles fails *** FAILED ***
[info]   The code passed to eventually never returned normally. Attempted 8799 times over 5.000078537283333 minutes. Last failure message: Some(Creating) was not equal to Some(Error). (LeonardoServiceSpec.scala:1450)
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
```diff
@@ -2,130 +2,222 @@ package org.broadinstitute.dsde.workbench.leonardo
...
-      case (email, keyId) =>
-        IO.fromFuture(IO(googleIamDAO.removeServiceAccountKey(googleProject, email, keyId)))
-    } getOrElse IO.unit
+  private def updateGroupMembership(groupEmail: WorkbenchEmail,
+                                    memberEmail: WorkbenchEmail,
+                                    addToGroup: Boolean): IO[Unit] =
+    IO.fromFuture[Unit] {
+      IO {
+        retryExponentially(when409, s"Could not update group '$groupEmail' for member '$memberEmail'") { () =>
+          logger.info(s"Checking if '$memberEmail' is part of group '$groupEmail'...")
```
rtitle: Should these be `debug` statements?

rtitle:
This overall looks good.. eager to see the firecloud-develop changes. Also there are a few failing unit tests.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
kyuksel: Yes, on my clean-up list :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
```diff
@@ -2,130 +2,222 @@ package org.broadinstitute.dsde.workbench.leonardo
...
+      }
+
+  /**
+   *  Add the user's service account to the Google group.
+   *  This group has compute.imageUser role on the custom Dataproc image project,
+   *  which allows the user's cluster to pull the image.
+   */
+  def updateDataprocImageGroupMembership(googleProject: GoogleProject, createCluster: Boolean): IO[Unit] = {
+    // TODO Template and get the group and domain values from config
+    val dataprocImageUserGoogleGroupName = "kyuksel-test-dataproc-image-group-11"
```
Qi77Qi: this TODO needs to be addressed before this can be merged

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
```diff
@@ -2,130 +2,222 @@ package org.broadinstitute.dsde.workbench.leonardo
...
-              dataprocSA = WorkbenchEmail(s"service-$projectNumber@dataproc-accounts.iam.gserviceaccount.com")
-              apiSA = WorkbenchEmail(s"$projectNumber@cloudservices.gserviceaccount.com")
-              _ <- retryIam(imageProject, dataprocSA, roles).recoverWith {
-                case e if when400(e) => retryIam(imageProject, apiSA, roles)
+  private def createDataprocImageUserGoogleGroupIfItDoesntExist(
+    dpImageUserGoogleGroupName: String,
+    dpImageUserGoogleGroupEmail: WorkbenchEmail
+  ): IO[Unit] = {
+    logger.info(s"Checking if Dataproc image user Google group '${dpImageUserGoogleGroupEmail}' already exists...")
+    IO.fromFuture[Unit] {
```
Qi77Qi, line 160: I'd prefer instead of doing something like `IO.fromFuture(IO{a whole block of code})`, we do 
```
for {
  grouEmailOpt <- IO.fromFuture(IO(googleDirectoryDAO.getGoogleGroup(dpImageUserGoogleGroupEmail)))
 _ <- grouEmailOpt match {
case None =>
???
case Some(x) =>
???
}
} yield ???
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
```diff
@@ -137,3 +229,14 @@ class ClusterHelper(
   }
 
 }
+
+object ClusterHelper {
```
Qi77Qi, line 233: is there any benefit wrapping these case classes in `ClusterHelper` object?

kyuksel: Nothing functional but I find the code easier to read that way.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
kyuksel: What'd be the benefit of doing that?

Qi77Qi: it defeats the purpose of using `IO` if you're just doing `IO.fromFuture{IO(my whole `Future` based app)}.

also see https://typelevel.org/cats-effect/datatypes/io.html#keep-granularity

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
kyuksel: yes, there'll be a firecloud-develop PR for that

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
kyuksel: Thanks for the reference. Just pushed a commit that refactors it.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
```diff
@@ -40,24 +44,87 @@ class ClusterHelper(
...
         () =>
-          if (createCluster) {
+          val iamUpdate = if (createCluster) {
             googleIamDAO.addIamRoles(project, email, MemberType.ServiceAccount, roles)
           } else {
             googleIamDAO.removeIamRoles(project, email, MemberType.ServiceAccount, roles)
           }
+          iamUpdate map { _ =>
+            ()
+          }
```
kyuksel, line 117: This ugliness is for the method to yield an `IO[Unit]` (from `IO[Boolean]` which `add/removeIamRoles()` returns) 🙈 This type difference existed in the dev version as well but there was another IO construction there, which somehow makes the types fit.

project/Dependencies.scala
```diff
@@ -3,102 +3,125 @@ import sbt._
 object Dependencies {
   val scalaV = "2.12"
```
kyuksel, line 5: Not sure what happened below. I just ran the formatter  🤷‍♂

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/Boot.scala
```diff
@@ -47,119 +48,133 @@ import scala.concurrent.duration._
...
-                                          clusterDateAccessedActor)
-      val statusService = new StatusService(appDependencies.googleDataprocDAO,
-                                            appDependencies.samDAO,
+                                     appDependencies.dbReference,
+                                     appDependencies.metrics)
+          )
+
+          appDependencies.clusterHelper
+            .setupDataprocImageGoogleGroup()
+            .unsafeRunSync()
```
Qi77Qi: I'd prefer to put this in a for comprehension in https://github.com/DataBiosphere/leonardo/pull/1130/files#diff-fad211c102fcc3c8dc9b99504c200f08R140

Something like:
```
        for {
          _ <- appDependencies.clusterHelper.setupDataprocImageGoogleGroup()
          _ <- IO.fromFuture(IO(Http().bindAndHandle(leoRoutes.route, "0.0.0.0", 8080))).void
            .onError {
            case t: Throwable =>
              unsafeLogger.error(t)("FATAL - failure starting http server")
          }
          _ <- IO.never
        } yield ()
```

This way, it's very clear where effects are happening(everything in for comp), where all declarations are(everything outside for comp).

project/Dependencies.scala
```diff
@@ -3,102 +3,125 @@ import sbt._
...
+  val guavaV = "28.0-jre"
 
-  val workbenchUtilV    = "0.5-4c7acd5"
-  val workbenchModelV   = "0.13-6dc016b"
-  val workbenchGoogleV  = "0.21-58c913d"
-  val workbenchGoogle2V = "0.6-f846fad"
+  val workbenchUtilV = "0.5-4c7acd5"
+  val workbenchModelV = "0.13-6dc016b"
+  val workbenchGoogleV = "0.21-890a74f"
+  val workbenchGoogle2V = "0.6-8bae8e8"
```
Qi77Qi: this hash is older than `f846fad`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
```diff
@@ -2,131 +2,229 @@ package org.broadinstitute.dsde.workbench.leonardo
...
+      }
 
+  /**
+   * Add the user's service account to the Google group.
+   * This group has compute.imageUser role on the custom Dataproc image project,
+   * which allows the user's cluster to pull the image.
+   */
+  def updateDataprocImageGroupMembership(googleProject: GoogleProject, createCluster: Boolean): IO[Unit] =
+    dataprocConfig.customDataprocImage.flatMap(parseImageProject) match {
+      case None => IO.unit
```
Qi77Qi: 1. you can use `.traverse`
2. I'd get rid of 
```
              val projectNumberOptIO = IO.fromFuture(IO(googleComputeDAO.getProjectNumber(googleProject)))
              val clusterIamSetupExceptionIO = IO.raiseError[Long](ClusterIamSetupException(imageProject))
```

Something like
```
    dataprocConfig.customDataprocImage.flatMap(parseImageProject).traverse {
      imageProject =>
        for {
          count <- IO.fromFuture(IO(dbRef.inTransaction {
            _.clusterQuery.countActiveByProject(googleProject)
          }))
          _ <- // Note: Don't remove the account if there are existing active clusters in the same project,
            // because it could potentially break other clusters. We only check this for the 'remove' case.
            if (count > 0 && !createCluster) {
              IO.unit
            } else {
              for {
                projectNumberOptIO <- IO.fromFuture(IO(googleComputeDAO.getProjectNumber(googleProject)))
                projectNumber <- IO.fromEither(projectNumberOptIO.toRight(ClusterIamSetupException(imageProject)))
                // Note that the Dataproc service account is used to retrieve the image, and not the user's
                // pet service account. There is one Dataproc service account per Google project. For more details:
                // https://cloud.google.com/dataproc/docs/concepts/iam/iam#service_accounts
                dataprocServiceAccountEmail = WorkbenchEmail(
                  s"service-${projectNumber}@dataproc-accounts.iam.gserviceaccount.com"
                )
                _ <- updateGroupMembership(googleGroupsConfig.dataprocImageProjectGroupEmail,
                  dataprocServiceAccountEmail,
                  createCluster)
              } yield ()
            }
        } yield ()
    }.void
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
```diff
@@ -2,131 +2,229 @@ package org.broadinstitute.dsde.workbench.leonardo
...
-          if (create) {
-            googleIamDAO.addIamRolesForUser(project, email, roles)
+          val iamUpdate = if (createCluster) {
+            googleIamDAO.addIamRoles(project, email, MemberType.ServiceAccount, roles)
           } else {
-            googleIamDAO.removeIamRolesForUser(project, email, roles)
+            googleIamDAO.removeIamRoles(project, email, MemberType.ServiceAccount, roles)
+          }
+          iamUpdate map { _ =>
+            ()
```
Qi77Qi: I would do
```
if (createCluster) {
    googleIamDAO.addIamRoles(project, email, MemberType.ServiceAccount, roles).void
} else {
      googleIamDAO.removeIamRoles(project, email, MemberType.ServiceAccount, roles).void
 }
```

And get rid of 
```
iamUpdate map { _ =>
            ()
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
```diff
@@ -2,131 +2,229 @@ package org.broadinstitute.dsde.workbench.leonardo
...
           }
       }))
 
-    // Add the Dataproc Worker role in the user's project to the cluster service account, if present.
-    // This is needed to be able to spin up Dataproc clusters using a custom service account.
-    // If the Google Compute default service account is being used, this is not necessary.
-    val dataprocWorkerIO = serviceAccountInfo.clusterServiceAccount.map { email =>
-      // Note: don't remove the role if there are existing active clusters owned by the same user,
+    serviceAccountInfo.clusterServiceAccount.map { email =>
```
Qi77Qi: same, use `.traverse`

```
    serviceAccountInfo.clusterServiceAccount.traverse { email =>
      // Note: Don't remove the role if there are existing active clusters owned by the same user,
      // because it could potentially break other clusters. We only check this for the 'remove' case,
      // it's ok to re-add the roles.
      IO.fromFuture(IO(dbRef.inTransaction { _.clusterQuery.countActiveByClusterServiceAccount(email) })).flatMap {
        count =>
          if (count > 0 && !createCluster) {
            IO.unit
          } else {
            retryIam(googleProject, email, Set("roles/dataproc.worker"))
          }
      }
    }.void
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
```diff
@@ -137,3 +235,14 @@ class ClusterHelper(
...
 }
+
+object ClusterHelper {
+  final case class ClusterIamSetupException(googleProject: GoogleProject)
+      extends LeoException(s"Error occurred setting up IAM roles in project ${googleProject.value}")
+
+  final case class GoogleGroupCreationException(googleGroup: WorkbenchEmail)
+      extends LeoException(s"Failed to create the Google group '${googleGroup}'", StatusCodes.InternalServerError)
+
+  final case class ImageProjectNotFoundException()
```
Qi77Qi: use `final case object` instead

http/src/test/resources/reference.conf
Qi77Qi: Is `TODO` still valid?

http/src/test/resources/reference.conf
```diff
@@ -93,6 +93,12 @@ sam {
 
 google {
   leoServiceAccountJsonFile = "fakePath"
+  groups {
+    subEmail = "google@dev.test.firecloud.org"
+    // TODO Change the default values
+    dataprocImageProjectGroupName = "dataproc-image-project-group"
+    dataprocImageProjectGroupEmail = ${google.groups.dataprocImageProjectGroupName}"@test.firecloud.org"
```
Qi77Qi, line 99: where does `google.groups.dataprocImageProjectGroupName` come from?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
```diff
@@ -203,7 +203,7 @@ object Cluster {
   // A possible future solution might be to separate Cluster into an internal representation (backed by the database)
   // and an API-response representation (which may contain additional metadata/fields).
   private lazy val cachedClusterUrlBase: String = {
-    val config = ConfigFactory.parseResources("leonardo.conf").withFallback(ConfigFactory.load())
+    val config = ConfigFactory.parseResources("leonardo.conf").withFallback(ConfigFactory.load()).resolve()
```
rtitle, line 206: Fyi I got rid of `cachedClusterUrlBase` in my PR (but we can leave it here).

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -537,16 +537,25 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
...
                                    updatedNumWorkersAndPreemptibles.right) recoverWith {
             case gjre: GoogleJsonResponseException =>
-              // Typically we will revoke this role in the monitor after everything is complete, but if Google fails to resize the cluster we need to revoke it manually here
+              // Typically we will revoke this role in the monitor after everything is complete, but if Google fails to
+              // resize the cluster we need to revoke it manually here
               clusterHelper
                 .removeClusterIamRoles(existingCluster.googleProject, existingCluster.serviceAccountInfo)
                 .unsafeToFuture()
+              // Remove member from the Google Group that has the IAM role to pull the Dataproc image
+              clusterHelper
```
rtitle: This future should probably be chained with the one above it. Maybe something like:
```
val cleanupIO = for {
  _ <- clusterHelper.removeClusterIamRoles(existingCluster.googleProject, existingCluster.serviceAccountInfo)
  _ <- clusterHelper.updateDataprocImageGroupMembership(existingCluster.googleProject, createCluster = false)
  _ <- IO(logger.error(s"Could not successfully update cluster ${existingCluster.projectNameString}", gjre))
  err <- IO.raiseError(InvalidDataprocMachineConfigException(gjre.getMessage))
} yield err

cleanupIO.unsafeToFuture
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
rtitle: +1

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
rtitle: +1

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
rtitle: +1, you also use `.traverse_` and git rid of the trailing `.void`, right?

http/src/test/resources/reference.conf
rtitle: I think it's a reference to the previous line in the config

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorSpec.scala
rtitle: I've seen this too in other PRs and dev. Maybe we should have a ticket for it.

project/Dependencies.scala
rtitle: Yeah.. double check these workbench-libs hashes. I don't think `google2` should have changed?

project/Dependencies.scala
```diff
@@ -3,102 +3,125 @@ import sbt._
...
-  val excludeFindbugsJsr        = ExclusionRule(organization = "com.google.code.findbugs", name = "jsr305")
-  val excludeGson               = ExclusionRule(organization = "com.google.code.gson", name = "gson")
-  val excludeGoogleApiClient    = ExclusionRule(organization = "com.google.api-client", name = "google-api-client")
-  val excludeGoogleApiClientJackson2  = ExclusionRule(organization = "com.google.http-client", name = "google-http-client-jackson2")
-  val excludeGoogleHttpClient   = ExclusionRule(organization = "com.google.http-client", name = "google-http-client")
-  val excludeJacksonCore        = ExclusionRule(organization = "com.fasterxml.jackson.core", name =  "jackson-core")
-  val excludeJacksonAnnotation  = ExclusionRule(organization = "com.fasterxml.jackson.core", name = "jackson-annotations")
-  val excludeSlf4j              = ExclusionRule(organization = "org.slf4j", name = "slf4j-api")
-  val excludeLogbackCore        = ExclusionRule(organization = "ch.qos.logback", name = "logback-core")
+  val excludeAkkaActor = ExclusionRule(organization = "com.typesafe.akka", name = "akka-actor_2.12")
```
rtitle, line 25: Oof.. should we avoid formatting `Dependencies.scala`? I kinda liked it better before. :)

rtitle:
This looks good, mostly stylistic comments above. I'm happy automation tests passed! Interested to see what performance tests uncover.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/Boot.scala
kyuksel: `setupDataprocImageGoogleGroup()` is called only if we're in `backLeo` mode and I'd rather keep all operations performed exclusively to `backLeo` in a single block so I converted that call to 

```scala
for {
            _ <- appDependencies.clusterHelper.setupDataprocImageGoogleGroup()
          } yield ()
```

As for `IO(Http().bindAndHandle`, putting `IO.never` inside `for` gives a dead code error so I ended up with 

```scala
(
          for {
            _ <- IO.fromFuture {
              IO {
                Http()
                  .bindAndHandle(leoRoutes.route, "0.0.0.0", 8080)
                  .onError {
                    case t: Throwable =>
                      unsafeLogger.error(t)("FATAL - failure starting http server").unsafeToFuture()
                  }
              }
            }
          } yield ()
        ) >> IO.never
```
🙈

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
kyuksel: 👍

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
kyuksel: 👍

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
kyuksel: so many pearls 😄

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
kyuksel: ah...yesss!

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
kyuksel: 👍 
Right.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
kyuksel: 👍

http/src/test/resources/reference.conf
kyuksel: Nope. Thanks for the catch!

http/src/test/resources/reference.conf
kyuksel: Exactly. Note that we need to `resolve()` the config upon loading for the referencing to work; e.g.

```scala
val config = ConfigFactory
    .parseResources("leonardo.conf")
    .withFallback(ConfigFactory.load())
    .resolve()
```

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorSpec.scala
kyuksel: This stopped happening at some point while I was addressing the first round of PR comments but I didn't realize it then and couldn't pinpoint what may have fixed it  😞

project/Dependencies.scala
kyuksel: Where do you see `f846fad`? 

I see
`890a74f` in 
https://github.com/broadinstitute/workbench-libs/blob/develop/google/CHANGELOG.md#021 
and `8bae8e8` in 
https://github.com/broadinstitute/workbench-libs/blob/develop/google2/CHANGELOG.md#06

project/Dependencies.scala
kyuksel: I hear you. I manually reverted most, if not all, of them and excluded this file in scalafmt config.

project/Dependencies.scala
kyuksel: (Talked faces with Qi)

`google2` shouldn't have changed. I probably messed it up while rebasing. While at it, I updated it to the latest (`0.6-4df42ca`)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
Qi77Qi: minor, `err` is not needed.
```
val cleanupIO = for {
  _ <- clusterHelper.removeClusterIamRoles(existingCluster.googleProject, existingCluster.serviceAccountInfo)
  _ <- clusterHelper.updateDataprocImageGroupMembership(existingCluster.googleProject, createCluster = false)
  _ <- IO(logger.error(s"Could not successfully update cluster ${existingCluster.projectNameString}", gjre))
  _ <- IO.raiseError(InvalidDataprocMachineConfigException(gjre.getMessage))
} yield ()
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/Boot.scala
```diff
@@ -47,119 +48,137 @@ import scala.concurrent.duration._
...
-      val statusService = new StatusService(appDependencies.googleDataprocDAO,
-                                            appDependencies.samDAO,
+                                     appDependencies.dbReference,
+                                     appDependencies.metrics)
+          )
+
+          for {
+            _ <- appDependencies.clusterHelper.setupDataprocImageGoogleGroup()
+          } yield ()
+        }
```
Qi77Qi, line 139: this IO is not chained to the main one...so I don't think this is actually doing anything.

So I would move this down to the main for comp, something like
```
  for {
          _ <- if(leoExecutionModeConfig.backLeo) appDependencies.clusterHelper.setupDataprocImageGoogleGroup() else IO.unit
          _ <- IO.fromFuture {
            IO {
              Http()
                .bindAndHandle(leoRoutes.route, "0.0.0.0", 8080)
                .onError {
                  case t: Throwable =>
                    unsafeLogger.error(t)("FATAL - failure starting http server").unsafeToFuture()
                }
            }
          } >> IO.never
        } yield ()
```

kyuksel: I'm not sure I understand how moving it makes a difference. Aren't we `flatMap`ing on `appDependencies.clusterHelper.setupDataprocImageGoogleGroup()` either way?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
kyuksel: 👍

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/Boot.scala
kyuksel: P.S. My motivation for keeping it up inside the `if (backLeo)` block was to be able to see all backLeo-specific operations in one place.

rtitle: I think Qi's right. The expression:
```
for {
   _ <- appDependencies.clusterHelper.setupDataprocImageGoogleGroup()
} yield ()
```
itself returns an `IO`, which is never run or chained to any other IO. The `startup` method must return a single `IO`, which is run under the hood by `IOApp`. Therefore I agree with Qi's suggestion to chain together all the `IO`s in a single for-comp.

project/Dependencies.scala
rtitle: Still a lot of changes in the diff but I can live with it if it's just indentation.

rtitle:
Looks good, my only remaining comment is the same as Qi's: there's an orphan `IO` in `Boot.scala`.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/Boot.scala
kyuksel: Ah right...I forgot it's the `IOApp` that runs the returned `IO`.

kyuksel: I get the compilation error

```
[error] Boot.scala:159:13: dead code following this construct
[error]           _ <- IO.fromFuture {
[error]             ^
[error] one error found
```

```scala
_ <- IO.fromFuture {
            IO {
              Http()
                .bindAndHandle(leoRoutes.route, "0.0.0.0", 8080)
                .onError {
                  case t: Throwable =>
                    unsafeLogger.error(t)("FATAL - failure starting http server").unsafeToFuture()
                }
            }
          } >> IO.never
```

I think it's because `IO.never` returns an `IO[Nothing]` but not sure. I have to look into this more.

Qi77Qi: okay...in that case you probably need to move `IO.never` outside the for comp...
```
val res =   for {
          _ <- if(leoExecutionModeConfig.backLeo) appDependencies.clusterHelper.setupDataprocImageGoogleGroup() else IO.unit
          _ <- IO.fromFuture {
            IO {
              Http()
                .bindAndHandle(leoRoutes.route, "0.0.0.0", 8080)
                .onError {
                  case t: Throwable =>
                    unsafeLogger.error(t)("FATAL - failure starting http server").unsafeToFuture()
                }
            }
          }
        } yield ()

res >> IO.never
```

Qi77Qi: now that I think about it...I think this kinda of one off environment specific setup should probably live in terraform just like leo's SA is created by terraform.

kyuksel: Hmm...I can see a rationale for that. My reasons for not doing that would be:

1. From the ephemeral environment presentation, it sounded like account creation type of stuff is not automated in terraform. It'd be nice to not have to add to the 34 manual steps Greg mentioned during the presentation.
2. Even if we created the group elsewhere, we still need to do checks at boot time to make sure the group is there, etc. so we wouldn't be able to remove most of the code I added to Boot.
3. I think the code in Boot.scala is well factored and doesn't clobber the file.
4. We'd still need the config templating and make sure it matches the externally created group.
5. We already have the code in place that works.

Anyway...I've addressed the last code comments.

Qi77Qi: I wasn't suggesting reverting the code in this PR since we really want this PR to be out..but more of thinking addressing it in a separate ticket.

> From the ephemeral environment presentation, it sounded like account creation type of stuff is not automated in terraform. It'd be nice to not have to add to the 34 manual steps Greg mentioned during the presentation.

I'm 90% sure we can automate this with terraform

> Even if we created the group elsewhere, we still need to do checks at boot time to make sure the group is there, etc. so we wouldn't be able to remove most of the code I added to Boot.

I'll buy the check argument.

rtitle (approved):
👍

Qi77Qi (approved):

PR #1131, opened 2019-10-30
[IA-1475] Improve regex for custom docker image

akarukappadath:

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
@@ -46,7 +46,7 @@ object ContainerRegistry {
 // Repo format: https://docs.docker.com/docker-hub/repos/
 // Docker hub doesn't have documentation about length limitation on repo name and image name, but 100 max seems reasonable
   final case object DockerHub extends ContainerRegistry {
-    def regex: Regex = """^([a-z0-9-_\/\S]){1,100}(:?[a-z0-9-_]{1,100}\S)$""".r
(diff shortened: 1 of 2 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
```diff
@@ -46,7 +46,7 @@ object ContainerRegistry {
 // Repo format: https://docs.docker.com/docker-hub/repos/
 // Docker hub doesn't have documentation about length limitation on repo name and image name, but 100 max seems reasonable
   final case object DockerHub extends ContainerRegistry {
-    def regex: Regex = """^([a-z0-9-_\/\S]){1,100}(:?[a-z0-9-_]{1,100}\S)$""".r
+    def regex: Regex = """^[A-Za-z0-9]+[\w./-]+(?::\w[\w.-]+)?(?:@[\w+.-]+:[A-Fa-f0-9]{32,})?$""".r
```
Qi77Qi, line 49: can you add a test to https://github.com/DataBiosphere/leonardo/blob/develop/http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModelSpec.scala#L217 in terms of what the previous regex is failing on

Qi77Qi (approved):

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
rtitle: I'm not sure this new regex is correct, it matches things like `"a//////////"`. I may clean it up (it relates to image autodetection).

PR #1134, opened 2019-11-05
[IA-1490] Remove hardcoded 5s proxy timeout

rtitle:
https://broadworkbench.atlassian.net/browse/IA-1490

cc @calbach 

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/ProxyService.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/ProxyService.scala
@@ -26,6 +26,7 @@ import org.broadinstitute.dsde.workbench.leonardo.model.NotebookClusterActions._
 import org.broadinstitute.dsde.workbench.leonardo.model._
 import org.broadinstitute.dsde.workbench.leonardo.model.google.ClusterName
 import org.broadinstitute.dsde.workbench.leonardo.monitor.ClusterDateAccessedActor.UpdateDateAccessed
+import org.broadinstitute.dsde.workbench.util.toScalaDuration
 import org.broadinstitute.dsde.workbench.model.{TraceId, UserInfo}
 import org.broadinstitute.dsde.workbench.model.google.GoogleProject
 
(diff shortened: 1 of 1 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/ProxyService.scala
```diff
@@ -65,6 +66,9 @@ class ProxyService(
 )(implicit val system: ActorSystem, materializer: ActorMaterializer, executionContext: ExecutionContext)
     extends LazyLogging {
 
+  final val requestTimeout = toScalaDuration(system.settings.config.getDuration("akka.http.server.request-timeout"))
```
kyuksel, line 69: so it'll be `60s`, correct?

rtitle: Yes: https://github.com/broadinstitute/firecloud-develop/blob/dev/base-configs/leonardo/leonardo.conf.ctmpl#L71

I was planning to confirm that after the auto-tests run by looking at Leo app logs in `fiab-stop`.

kyuksel: I first wondered if 60s for Jupyter to Proxy may be too much, in which case having a smaller timeout could alert us to an slowness issue but that's probably not the best way to detect such issues. We may want to have metrics for it if possible.

> I was planning to confirm that after the auto-tests run by looking at Leo app logs in fiab-stop.

Good idea.

kyuksel (approved):

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/ProxyService.scala
rtitle:
```
[INFO] [15:49:33.281] [ioapp-compute-4] o.b.d.w.l.service.ProxyService - Leo proxy request timeout is 1 minute
```

PR #1136, opened 2019-11-06
[IA-1401] Lower expiry time for clusterInternalIdCache

rtitle:
https://broadworkbench.atlassian.net/browse/IA-1401

AoU is occasionally seeing 404s on proxy endpoints after clusters are deleted and recreated with the same name.

All proxy endpoints go through [authCheck](https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/ProxyService.scala#L132) which looks up the clusterInternalId (cached by `(project, name)`) and asks Sam for permission (which is also cached). If Sam says the user doesn't have `GetClusterStatus` permission, the proxy endpoint returns 404 (e.g. pretend the cluster doesn't exist).

This PR lowers the TTL for the `clusterInternalIdCache` to 2 minutes. Previously it was 1 hour; so if a cluster was deleted/recreated within that window it was likely an old internalId was still cached which no longer exists in Sam. It is unlikely/impossible for a cluster to be deleted/recreated within 2 minutes.

This does mean the proxy will be making more frequent DB queries. I think this is probably OK -- the query to retrieve a cluster's internalId should be pretty light, since it looks up a single record entirely using indexes.

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/ProxyService.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/ProxyService.scala
@@ -72,8 +72,8 @@ class ProxyService(
   /* Cache for the bearer token and corresponding google user email */
   private[leonardo] val googleTokenCache = CacheBuilder
     .newBuilder()
-    .expireAfterWrite(proxyConfig.cacheExpiryTime.toMinutes, TimeUnit.MINUTES)
-    .maximumSize(proxyConfig.cacheMaxSize)
+    .expireAfterWrite(proxyConfig.tokenCacheExpiryTime.toSeconds, TimeUnit.SECONDS)
+    .maximumSize(proxyConfig.tokenCacheMaxSize)
     .build(
       new CacheLoader[String, Future[(UserInfo, Instant)]] {
         def load(key: String) =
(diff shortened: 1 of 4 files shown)


rtitle:
@calbach if you have a good test case for the 404s I'd be interested to try it against this change.

Qi77Qi (approved):

rtitle:
jenkins retest

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/ProxyService.scala
```diff
@@ -94,8 +94,8 @@ class ProxyService(
   /* Cache for the cluster internal id from the database */
   private[leonardo] val clusterInternalIdCache = CacheBuilder
     .newBuilder()
-    .expireAfterWrite(proxyConfig.cacheExpiryTime.toMinutes, TimeUnit.MINUTES)
-    .maximumSize(proxyConfig.cacheMaxSize)
+    .expireAfterWrite(proxyConfig.internalIdCacheExpiryTime.toSeconds, TimeUnit.SECONDS)
+    .maximumSize(proxyConfig.internalIdCacheMaxSize)
```
calbach, line 98: Confirming: Would this fix affect the proxy APIs but not the `getCluster()` API?

calbach (commented):
Thanks! Unfortunately I don't have a 100% reproducible test case, but in general: create new cluster, use cluster a bit, delete cluster, recreate cluster with the same name.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/ProxyService.scala
rtitle: Correct. The `getCluster` API does not do any caching.

PR #1140, opened 2019-11-07
IA-1495 enable monitoring

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-1495

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/google/HttpGoogleDataprocDAO.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/google/HttpGoogleDataprocDAO.scala
@@ -342,8 +342,10 @@ class HttpGoogleDataprocDAO(
       "spark:spark.yarn.am.memory" -> "640m"
     )
 
+    val stackdriverProps = Map("dataproc:dataproc.monitoring.stackdriver.enable" -> "true")
+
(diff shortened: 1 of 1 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/google/HttpGoogleDataprocDAO.scala
```diff
@@ -342,8 +342,10 @@ class HttpGoogleDataprocDAO(
       "spark:spark.yarn.am.memory" -> "640m"
     )
 
+    val stackdriverProps = Map("dataproc:dataproc.monitoring.stackdriver.enable" -> "true")
+
     val swConfig = new SoftwareConfig()
-      .setProperties((authProps ++ dataprocProps ++ yarnProps ++ createClusterConfig.properties).asJava)
+      .setProperties((authProps ++ dataprocProps ++ yarnProps ++ createClusterConfig.properties ++ stackdriverProps).asJava)
```
rtitle: `createClusterConfig.properties` is user-specified, I think we had put it last so it can override other configs we set by default. May still be good to have it last so users could turn off monitoring if need be.

rtitle (changes requested):
So simple! 1 minor comment

rtitle (approved):

Qi77Qi:
jenkins retest

PR #1141, opened 2019-11-07
[IA-1420] Update Leonardo APIs to launch an RStudio image on a cluster

rtitle:
Pretty small changes here - this essentially adds `rstudioDockerImage` to Leo swagger and adds a basic automation test verifying that it still works.

I will make some follow-up tickets for RStudio phase 1:
1. Integrate with Terra UI. This is a bit tricky. We don't want to put "RStudio" in the UI -- but at the same time the UI needs to somehow specify the `rstudioDockerImage` in the `createCluster` request, and let users actually _open_ it somehow using the rstudio proxy URL. I'll brainstorm a bit and run by @sharmatime.
2. Implement token refresh somehow in RStudio to prevent users from being signed out (for Jupyter we do this with an [extension](https://github.com/DataBiosphere/leonardo/blob/develop/docker/jupyter/custom/google_sign_in.js)).
3. Make an `rstudio-bioconductor` image in `anvil-docker` (maybe Nitesh could do this?)

In all cases:


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
```diff
@@ -88,8 +88,8 @@ final case class ClusterRequest(labels: LabelMap = Map.empty,
                                 autopauseThreshold: Option[Int] = None,
                                 defaultClientId: Option[String] = None,
                                 jupyterDockerImage: Option[ContainerImage] = None,
-                                rstudioDockerImage: Option[String] = None,
-                                welderDockerImage: Option[String] = None,
+                                rstudioDockerImage: Option[ContainerImage] = None,
```
Qi77Qi, line 91: I'd prefer to make `jupyterDockerImage` and `rstudioDockerImage` into a sum type. Just so it's obvious that it's either or

rtitle: From a Leo perspective you can actually have both running at the same time (they will run in different containers and listen at different proxy URLs). Terra UI may want to enforce one or the other though. We need to think about the exact Terra UI behavior..

Qi77Qi: gotcha. And we're not running welder when jupyter is not up right?

rtitle: Well right now it always launches Jupyter. If you don't specify a `jupyterDockerImage` in the request it uses the one in config. We probably want to retain that behavior for backwards compatibility. So the welder logic is unchanged as well.

`rstudioDockerImage` is truly optional -- there is no default and it doesn't influence Jupyter/welder.

rtitle: Actually I might change this behavior.. need to discuss with Adrian.

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/rstudio/RStudioSpec.scala
```diff
@@ -1,30 +1,24 @@
...
-                       rstudioDockerImage = Some("us.gcr.io/broad-dsp-gcr-public/leonardo-rstudio:860d5862f3f5")
-                     )) { cluster =>
-        withWebDriver { _ =>
-          val getResult = Try(RStudio.getApi(billingProject, cluster.clusterName))
-          getResult.isSuccess shouldBe true
-          getResult.get should not include "ProxyException"
-        }
+    "should launch RStudio" in { clusterFixture =>
+      withWebDriver { _ =>
+        // TODO: look into adding some selenium-based RStudio tests?
```
kyuksel, line 18: Should we create a ticket for ^^ ?

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoServiceSpec.scala
```diff
@@ -265,13 +266,13 @@ class LeonardoServiceSpec
...
-    val customWelderImage = Some("my-custom-welder-image-link")
+    val customWelderImage = GCR("my-custom-welder-image-link")
 
     // create the cluster
     val clusterRequest = testClusterRequest.copy(
       machineConfig = Some(singleNodeDefaultMachineConfig),
       stopAfterCreation = Some(true),
-      welderDockerImage = customWelderImage,
+      welderDockerImage = Some(customWelderImage),
       enableWelder = Some(true)
```
kyuksel, line 276: Since the [corresponding logic](https://github.com/DataBiosphere/leonardo/blob/be64e89cb2637796039d22953fb2945e66396e5c/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala#L1379) is a bit complicated, I wonder if we should test the client-supplied Welder image case with `enableWelder = Some(false)` or if it is too obscure to care about.

kyuksel (approved):
👍 with minor comment and question

rtitle:
I'm going to close this and re-open

PR #1144, opened 2019-11-15
[IA-1511] Check images for OK status when cluster creates

rtitle:
https://broadworkbench.atlassian.net/browse/IA-1511

Hopefully address a few common test failures. See comments inline.

In all cases:


automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoSuite.scala
```diff
@@ -74,7 +74,6 @@ final class LeonardoSuite
       new NotebookDataSyncingSpec,
       new NotebookGATKSpec,
 //      new NotebookHailSpec, TODO: revisit and enable this once hail image https://broadworkbench.atlassian.net/browse/IA-1237
-      new NotebookLocalizeFileSpec,
```
rtitle, line 77: Let's disable this test. My rationale:
1. It's failing frequently in Jenkins, lowering our pass percentage and affecting our velocity
2. It's legacy code, meaning we are not maintaining it even if it is flaky
3. Legacy `/localize` may be running on some old clusters, but we are not deploying it to any _new_ clusters
4. It's used in FireCloud, which is also legacy and IMO doesn't need to be tested

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookClusterMonitoringSpec.scala
```diff
@@ -257,9 +257,10 @@ class NotebookClusterMonitoringSpec extends GPAllocFixtureSpec with ParallelTest
                                              welderDockerImage = Some(LeonardoConfig.Leonardo.oldWelderDockerImage))
       ) { cluster =>
         // Verify welder is running with old version
-        val statusResponse = Welder.getWelderStatus(cluster).unsafeRunSync()
+        val statusResponse = Welder.getWelderStatus(cluster).attempt.unsafeRunSync()
```
rtitle, line 260: Using `attempt` for better error reporting. Otherwise it throws a hard-to-understand `circe` exception.

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookTestUtils.scala
```diff
@@ -177,114 +174,6 @@ trait NotebookTestUtils extends LeonardoTestUtils {
     testResult.get
   }
 
-  def withLocalizeDelocalizeFiles[T](cluster: Cluster,
```
rtitle, line 180: This is now dead code

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
```diff
@@ -390,37 +390,13 @@ class ClusterMonitorActor(
...
         case Running
-            if leoClusterStatus != Deleting && leoClusterStatus != Stopping && leoClusterStatus != Starting && runningInstanceCount == googleInstances.size =>
-          getMasterIp.map {
-            case Some(ip) => ReadyCluster(ip, googleInstances)
-            case None     => NotReadyCluster(ClusterStatus.Running, googleInstances)
-          }
-        case Running if leoClusterStatus == Starting && runningInstanceCount == googleInstances.size =>
+            if leoClusterStatus != Deleting && leoClusterStatus != Stopping && runningInstanceCount == googleInstances.size =>
           getMasterIp.flatMap {
-            case Some(ip) =>
```
rtitle, line 400: The meat of the fix. Previously we were only checking `ClusterTool` status in the transition from `Starting -> Running`. I lifted that code into its own method and now call it for both `Starting -> Running` and `Creating -> Running` status transitions. I also collapsed some redundant `case` statements.

Qi77Qi (approved):

PR #1153, opened 2019-11-21
[IA-1420] [IA-1502] RStudio backend work

rtitle:
This PR _should_ contain all the backend work required for RStudio in Q4. **knocks on wood**

Friends with https://github.com/broadinstitute/firecloud-develop/pull/1935

High level changes in this PR:
- Added an optional `toolDockerImage` field to the `createCluster` API. Deprecated `jupyterDockerImage` in favor of `toolDockerImage`.
- Added `HttpDockerDAO` which calls Docker APIs to autodetect what tool a given imageUrl contains.
   - I hope I didn't go too overboard with this! It seems to work. See comments inline.
- Set `clusterUrl` appropriately depending on what tool is being used. (Terra UI depends on this field.)
- Updated automation tests to use new `toolDockerImage` field and enabled `RStudioSpec`

In all cases:


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpDockerDAO.scala
```diff
@@ -0,0 +1,203 @@
...
+import org.http4s.client.Client
+import org.http4s.client.dsl.Http4sClientDsl
+import org.http4s.client.middleware.FollowRedirect
+import org.http4s.headers.Authorization
+
+/**
+ * Talks to Docker remote APIs to retrieve manifest information in order to try and figure out
+ * what tool it's running.
+ *
+ * This article was used as a guide: https://ops.tips/blog/inspecting-docker-image-without-pull/
```
rtitle, line 31: Again: hope I didn't go too overboard with this but it seems to work :)

If there are any issues in practice we could fall back to having clients explicitly specify images.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
```diff
@@ -30,15 +30,13 @@ sealed trait ContainerRegistry extends Product with Serializable {
 }
 object ContainerRegistry {
   final case object GCR extends ContainerRegistry {
-    def regex: Regex = """(.*gcr.io)\/(.+)\/(.+)(\:?.*)$""".r
-
+    val regex: Regex = """^(.*gcr.io)/([\w.-]+/[\w.-]+)(?::(\w[\w.-]+))?(?:@([\w+.-]+:[A-Fa-f0-9]{32,}))?$""".r
```
rtitle: Tweaked this regex to be more like the GCR one

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
```diff
@@ -30,15 +30,13 @@ sealed trait ContainerRegistry extends Product with Serializable {
...
     override def toString: String = "gcr"
   }
 
-// Repo format: https://docs.docker.com/docker-hub/repos/
-// Docker hub doesn't have documentation about length limitation on repo name and image name, but 100 max seems reasonable
+  // Repo format: https://docs.docker.com/docker-hub/repos/
   final case object DockerHub extends ContainerRegistry {
-    def regex: Regex = """^[A-Za-z0-9]+[\w./-]+(?::\w[\w.-]+)?(?:@[\w+.-]+:[A-Fa-f0-9]{32,})?$""".r
+    val regex: Regex = """^([\w.-]+/[\w.-]+)(?::(\w[\w.-]+))?(?:@([\w+.-]+:[A-Fa-f0-9]{32,}))?$""".r
```
rtitle, line 41: Added grouping and fixed some mistakes in the original regex: it would match things like `a////////`.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
```diff
@@ -106,13 +104,22 @@ case class AuditInfo(creator: WorkbenchEmail,
                      dateAccessed: Instant,
                      kernelFoundBusyDate: Option[Instant])
 
-sealed trait ClusterTool extends EnumEntry with Serializable with Product
+sealed trait ClusterTool extends EnumEntry with Serializable with Product {
```
rtitle: This probably conflicts with https://github.com/DataBiosphere/leonardo/pull/1124.. will cross that bridge when I come to it.

rtitle:
jenkins retest

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/rstudio/RStudioSpec.scala
```diff
@@ -1,30 +1,24 @@
...
-                       rstudioDockerImage = Some("us.gcr.io/broad-dsp-gcr-public/leonardo-rstudio:860d5862f3f5")
-                     )) { cluster =>
-        withWebDriver { _ =>
-          val getResult = Try(RStudio.getApi(billingProject, cluster.clusterName))
-          getResult.isSuccess shouldBe true
-          getResult.get should not include "ProxyException"
-        }
+    "should launch RStudio" in { clusterFixture =>
+      withWebDriver { _ =>
+        // TODO: look into adding some selenium-based RStudio tests?
```
kyuksel: Do we need this TODO? I believe we have a JIRA ticket for it.

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/rstudio/RStudioSpec.scala
```diff
@@ -1,30 +1,24 @@
...
+        val getResult = Try(RStudio.getApi(clusterFixture.cluster.googleProject, clusterFixture.cluster.clusterName))
+        getResult.isSuccess shouldBe true
+        getResult.get should not include "ProxyException"
```
kyuksel, line 23: Are there things we can expect in the GET response to make the check stricter or just not seeing `ProxyException` good enough?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
```diff
@@ -30,15 +30,13 @@ sealed trait ContainerRegistry extends Product with Serializable {
 }
```
kyuksel, line 31: `LeonardoModel` has gotten really long. I wonder if it's time we moved some pieces to their own - perhaps all but Cluster stuff and then this file could be renamed as ClusterModel or something. Could start with moving Container/Image related constructs.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpDockerDAO.scala
```diff
@@ -0,0 +1,203 @@
+package org.broadinstitute.dsde.workbench.leonardo
+package dao
```
kyuksel, line 2: What's the benefit of having the second `package` line above instead of a single line (`package org.broadinstitute.dsde.workbench.leonardo`)?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
kyuksel: Should we be more strict to ensure expressions would only match one of the four possible `gcr.io`s?

```suggestion
    val regex: Regex = """^((us\.|eu\.|asia\.)?gcr.io)/([\w.-]+/[\w.-]+)(?::(\w[\w.-]+))?(?:@([\w+.-]+:[A-Fa-f0-9]{32,}))?$""".r
```

kyuksel: It could be nice to add negative tests against paths starting with `.gcr.io`, `foogcr.io`, `foo.gcr.io`, `foo.bar.gcr.io`, `foo.gcr.io.bar`, etc.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
kyuksel: Cool. Can we add a test against that case?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -782,31 +791,34 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
...
+      autodetectedImageOpt <- clusterRequest.toolDockerImage.traverse { image =>
+        dockerDAO.detectTool(image).flatMap {
+          case None       => IO.raiseError(ImageAutoDetectionException(traceId, image))
+          case Some(tool) => IO.pure(ClusterImage(tool, image.imageUrl, now))
+        }
+      }
+      // Figure out the tool image. Rules:
+      // - if we were able to autodetect an image, use that
+      // - else if a legacy jupyterDockerImage param was sent, use that
+      // - ese use the default jupyter image
```
kyuksel:
```suggestion
      // - else use the default jupyter image
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -782,31 +791,34 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
...
+    for {
+      now <- IO(Instant.now)
+      traceId <- ev.ask
+      // Try to autodetect the image
+      autodetectedImageOpt <- clusterRequest.toolDockerImage.traverse { image =>
+        dockerDAO.detectTool(image).flatMap {
+          case None       => IO.raiseError(ImageAutoDetectionException(traceId, image))
+          case Some(tool) => IO.pure(ClusterImage(tool, image.imageUrl, now))
+        }
+      }
```
kyuksel, line 809: Are we testing the image figuring out logic somewhere?

kyuksel (commented):
Looks good overall - comments inline

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/rstudio/RStudioSpec.scala
rtitle: Yeah we do have a JIRA - I can remove the TODO, or make it reference the JIRA

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/rstudio/RStudioSpec.scala
rtitle: Probably - will add. Yeah I don't like just having negative checks -- too easy to have false passes.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpDockerDAO.scala
rtitle: @Qi77Qi taught me this… trying to remember. :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
rtitle: I think @Qi77Qi started this by moving API-contract stuff to a different file. I'd probably be in favor of splitting it up more, but I tend to think we should have a dedicated PR to do that..

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
rtitle: Yeah that's reasonable, although I think the suggestion is not quite right (there's an unmatched paren). I'll try this and improve the regex unit tests.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
rtitle: Will do

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
rtitle: I think there are unit tests but they could be improved.. will look, thanks.

rtitle:
Thanks for the great feedback @kyuksel! I'll apply the review comments -- also `RStudioSpec` seems to be failing, will look into it.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpDockerDAO.scala
Qi77Qi: with this, everything under `org.broadinstitute.dsde.workbench.leonardo` namespace will be in scope automatically, so you don't have to `import org.broadinstitute.dsde.workbench.leonardo._`

kyuksel: Oh interesting. I can see the appeal in terms of convenience. At the same time I fear this may make it even easier to set up an implicit hell situation 🙈

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpDockerDAO.scala
kyuksel: I wouldn't say 'overboard' but 'fancy'  😊

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
kyuksel: I think LeonardoModel is one of our upcoming walk-through topics. Perhaps we can come up with a plan then.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpDockerDAO.scala
Qi77Qi: yea, there needs to be a good reason if we're putting implicits under `org.broadinstitute.dsde.workbench.leonardo`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
kyuksel: Ah yea...I've updated it. Hope it works now.

rtitle: It does! added in latest commit

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
rtitle: Added

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookClusterMonitoringSpec.scala
```diff
@@ -160,7 +160,7 @@ class NotebookClusterMonitoringSpec extends GPAllocFixtureSpec with ParallelTest
           .futureValue
 
         val request = defaultClusterRequest.copy(
-          jupyterDockerImage = Some(LeonardoConfig.Leonardo.baseImageUrl),
+          toolDockerImage = None,
```
Qi77Qi: is changing `Some` to `None` intentional?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpDockerDAO.scala
```diff
@@ -0,0 +1,208 @@
...
+
+  def apply[F[_]: Concurrent](httpClient: Client[F])(implicit logger: Logger[F]): HttpDockerDAO[F] =
+    new HttpDockerDAO[F](httpClient)
+
+  // Decoders
+  implicit val tokenDecoder: Decoder[Token] = Decoder.instance { d =>
+    for {
+      token <- d.downField("token").as[String]
+    } yield Token(token)
+  }
```
Qi77Qi, line 158: fyi, you can do `Decoder.decodeString.map(Token)`

rtitle:
@kyuksel this is ready for another look

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookClusterMonitoringSpec.scala
rtitle: I don't think so… good catch. Will put it back.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpDockerDAO.scala
rtitle: Actually that doesn't seem to work for json like:
```
{
  …,
  "token": "something",
  …
}
```
Is there a better way?

kyuksel (approved):

PR #1158, opened 2019-11-26
QA-995 fix flaky test

Qi77Qi:
There were a few failed tests yesterday since https://github.com/DataBiosphere/leonardo/pull/1124.

This should fix the flakiness

cc @gdlex4015 

---
Have you read [CONTRIBUTING.md](https://github.com/DataBiosphere/leonardo/blob/develop/CONTRIBUTING.md) lately? If not, do that first.

I, the developer opening this PR, do solemnly pinky swear that:

- [ ] I've documented my API changes in Swagger

In all cases:

- [ ] Get a thumbsworth of review and PO signoff if necessary
- [ ] Verify all tests go green
- [ ] Squash and merge; Delete your branch after this
- [ ] Test this change deployed correctly and works on dev environment after deployment

PR #1159, opened 2019-11-26
QA-974 fix flaky tests

Qi77Qi:
https://broadworkbench.atlassian.net/browse/QA-974

When we don't do `saveAndCheckpoint()`, chrome will pop up a window asking "do you want save the changes before quitting", which I think is timing out tests from time to time

---

In all cases:

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookClusterMonitoringSpec.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookClusterMonitoringSpec.scala
@@ -37,7 +37,6 @@ class NotebookClusterMonitoringSpec extends GPAllocFixtureSpec with ParallelTest
           // Create a notebook and execute a cell
           withNewNotebook(cluster, kernel = Python3) { notebookPage =>
             notebookPage.executeCell(s"""print("$printStr")""") shouldBe Some(printStr)
-            notebookPage.saveAndCheckpoint()
           }
 
           // Stop the cluster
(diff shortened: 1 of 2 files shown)


kyuksel (commented):
Good theory 🤞 

If this is something we want to do with every notebook, I wonder if we should build it into `withNewNotebook` or something to prevent us from forgetting it  🤔

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookBioconductorKernelSpec.scala
```diff
@@ -40,6 +40,7 @@ class NotebookBioconductorKernelSpec extends ClusterFixtureSpec with NotebookTes
         withNewNotebook(clusterFixture.cluster, RKernel) { notebookPage =>
           // Make sure BiocManager has the correct version of Bioconductor
           notebookPage.executeCell("""BiocManager::version() == "3.9"""")
+          notebookPage.saveAndCheckpoint()
```
kyuksel: In your [other flaky test fix PR](https://github.com/DataBiosphere/leonardo/pull/1158/files), you also added the line with `shutdownKernel()`. Should we do that here as well?

Qi77Qi: I don't think `shutdownKernel` is relevant...I don't want to blindly add `shutdownKernel ` everywhere. I had it in the other test because I saw improvement after adding `shutdownKernel`, and that test was a bit more complicated

kyuksel (approved):

Qi77Qi:
jenkins retest

PR #1162, opened 2019-12-03
Jc update transition

jdcanas:
This PR

-Has quite a few lines of commented out code, that may be needed based on a product decision
-enables updates to master machine type on a running cluster
-this is accomplished by a flag which allows the update method to stop the cluster, and queue the update + restart for back leo to handle
-adds 2 fields to the DB, the flag and the config to update to
-I have NOT tested this extensively with a FIAB yet, nor written as many test cases I would like,  hence the draft

In all cases:


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
```diff
@@ -41,9 +33,9 @@ final case class ClusterRecord(id: Long,
                                stagingBucket: Option[String],
                                autopauseThreshold: Int,
                                defaultClientId: Option[String],
-                               stopAfterCreation: Boolean,
-                               welderEnabled: Boolean,
-                               customClusterEnvironmentVariables: Map[String, String])
+                               updatedMachineConfig: MachineConfigRecord,
+                               customClusterEnvironmentVariables: Map[String, String],
+                               clusterFlags: ClusterFlagsRecord)
```
jdcanas: we have reached the tuple size limit in scala, so I had to group these. We should talk about normalizing our DB schema...

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
```diff
@@ -385,9 +381,52 @@ class ClusterMonitorActor(
       _ <- dbRef.inTransaction { _.clusterQuery.clearKernelFoundBusyDate(cluster.id, now) }
       // Record metrics in NewRelic
       _ <- recordMetrics(cluster.status, ClusterStatus.Stopped).unsafeToFuture()
+//      result <- if (cluster.stopAndUpdate) handleStoppedClusterUpdate(cluster, googleInstances) else Future.successful(ShutdownActor(RemoveFromList(cluster)))
     } yield ShutdownActor(RemoveFromList(cluster))
   }
 
+  private def handleStoppedClusterUpdate(cluster: Cluster, googleInstances: Set[Instance]): Future[ClusterMonitorMessage] = {
```
jdcanas: this isn't currently used atm, but will be if product wants to enable the case where numWorkers and machineType are updated at the same time on a running cluster, so leaving it for now

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorSupervisor.scala
```diff
@@ -187,10 +187,51 @@ class ClusterMonitorSupervisor(
           logger.error(s"Error occurred stopping cluster ${cluster.projectNameString} after creation", e)
         }
 
-    case ClusterStopped(cluster) =>
+//    case ClusterStopQueued(cluster) =>
```
jdcanas, line 195: possibly needed with product decision

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -305,6 +321,18 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
           dbRef.inTransactionIO { _.clusterQuery.updateClusterStatus(existingCluster.id, ClusterStatus.Updating, now) }.void
         } else IO.unit
 
+        //check update follow-up
```
jdcanas, line 343: there is currently a bug in the below block, where if both clusterResized and shouldFollowup are true, an exception will occur as the cluster status will be 'Updating', and then the code will attempt to send it to "Stopping". This needs product resolution. The two solutions available are:

1. Disallow changes to numWorkers and masterMachineType in the same update
2. Defer the stopping to back leo

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -557,33 +608,9 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
       //if you've got to here you at least have GetClusterDetails permissions so a 403 is appropriate if you can't actually stop it
       _ <- checkClusterPermission(userInfo, StopStartCluster, cluster, throw403 = true)
 
-      _ <- internalStartCluster(userInfo.userEmail, cluster)
+      _ <- clusterHelper.internalStartCluster(cluster)
     } yield ()
 
-  def internalStartCluster(userEmail: WorkbenchEmail, cluster: Cluster): IO[Unit] =
```
jdcanas, line 572: moved to cluster helper, needed in backleo

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -811,45 +838,4 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
     Set(welderImageOpt, Some(jupyterImage), rstudioImageOpt).flatten
   }
 
-  private def getWelderAction(cluster: Cluster): WelderAction =
```
jdcanas, line 826: moved to cluster helper

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorSpec.scala
```diff
@@ -1324,6 +1324,110 @@ class ClusterMonitorSpec
...
+        updatedCluster shouldBe 'defined
+
+        updatedCluster.map(_.machineConfig.masterMachineType) shouldBe Some(newMachineType)
+        updatedCluster.map(_.stopAndUpdate) shouldBe Some(false)
+
+        updatedCluster.map(_.status) shouldBe Some(ClusterStatus.Starting)
+
+      }
+
+      verify(storageDAO, never()).deleteBucket(any[GcsBucketName], any[Boolean])
```
jdcanas, line 1431: this may be a bit unecessary

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorSupervisorSpec.scala
```diff
@@ -3,7 +3,7 @@ package org.broadinstitute.dsde.workbench.leonardo.monitor
 import java.time.Instant
```
jdcanas, line 3: I didn't actually change anything functionally in this file, but I happened to open it by accident before ClusterMonitorActorSpec, and had to refactor because I had no idea what it was trying to test with all the duplication

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20191125_stop_and_update.xml
```diff
@@ -0,0 +1,42 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
jdcanas, line 1: so here, in addition to a flag, I add a full updated MachineType to the db, as more or less a temporary variable. The need to do this was the motivation for this https://broadworkbench.atlassian.net/secure/RapidBoard.jspa?rapidView=35&projectKey=IA&view=planning&selectedIssue=IA-1530.  It may be a bit excessive to include all fields in MachineType, because only the updatedMasterMachineType is leveraged at the moment, but I wanted to ensure we could support other fields being updated for the Stop/Delete case if that became a priority suddenly.

PR #1165, opened 2019-12-04
IA-1600 switching db's execution context

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-1600

Most changes are mechanical Future to IO change. http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/DbReference.scala has the most important changes

1. Change `inTransaction` to return `IO` everywhere. Hence lots of `Future` to `IO` conversion. Recommended way to convert `DBIO` to `IO` after this PR is to do `dbio.transaction` instead of `dbRef.inTransaction(dbio)`, but `inTransaction` is still available.
2. Refactor `CommonTestData` into an `Object` because mixing multiple traits causes `unspecified fields` easily and it can be tricky to get the order of mixing traits right.
3. Use `Blocker` for db transaction.

Perf test done for the PR. Latest run https://fc-jenkins.dsp-techops.broadinstitute.org/view/Leonardo/job/LeoPerfTestRunner/282/

---

In all cases:


Qi77Qi:
jenkins retest

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -197,7 +198,7 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
         .getNotebookServiceAccount(userInfo, googleProject)
       serviceAccountInfo = ServiceAccountInfo(clusterServiceAccountOpt, notebookServiceAccountOpt)
 
-      clusterOpt <- dbRef.inTransactionIO { _.clusterQuery.getActiveClusterByNameMinimal(googleProject, clusterName) }
+      clusterOpt <- dbRef.dataAccess.clusterQuery.getActiveClusterByNameMinimal(googleProject, clusterName).transaction
```
Qi77Qi: this will be preferred way of converting `DBIO` to `IO`...most other places are doing 
```
dbRef.inTransaction {dbRef.dataAccess.clusterQuery.getActiveClusterByNameMinimal(googleProject, clusterName)}
```
which is a bit more verbose, they are that way for now because I introduced the syntax after I made all the other changes. Manual change to all the other places is a bit labor intense, hence leaving them as is for now.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/Boot.scala
```diff
@@ -97,17 +98,15 @@ object Boot extends IOApp with LazyLogging {
...
                                                 petGoogleStorageDAO,
-                                                appDependencies.dbReference,
                                                 appDependencies.authProvider,
                                                 appDependencies.serviceAccountProvider,
                                                 bucketHelper,
                                                 clusterHelper,
                                                 appDependencies.dockerDAO)
       if (leoExecutionModeConfig.backLeo) {
-        val jupyterDAO = new HttpJupyterDAO(appDependencies.clusterDnsCache)
         val rstudioDAO = new HttpRStudioDAO(appDependencies.clusterDnsCache)
```
rtitle: Maybe put `rstudioDAO` in `AppDependencies` too?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/LeoRoutes.scala
```diff
@@ -25,15 +25,9 @@ import org.broadinstitute.dsde.workbench.leonardo.service.{LeonardoService, Prox
...
 import org.broadinstitute.dsde.workbench.model.ErrorReportJsonSupport._
 import org.broadinstitute.dsde.workbench.model.google.GoogleProject
-import org.broadinstitute.dsde.workbench.model.{
-  ErrorReport,
-  TraceId,
-  WorkbenchEmail,
-  WorkbenchException,
-  WorkbenchExceptionWithErrorReport
-}
+import org.broadinstitute.dsde.workbench.model.{ErrorReport, TraceId, WorkbenchEmail, WorkbenchException, WorkbenchExceptionWithErrorReport}
```
rtitle: May need to run `scalafmt` :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpJupyterDAO.scala
```diff
@@ -1,54 +1,52 @@
...
-
-class HttpJupyterDAO(val clusterDnsCache: ClusterDnsCache)(implicit system: ActorSystem,
-                                                           executionContext: ExecutionContext,
-                                                           materializer: ActorMaterializer)
-    extends JupyterDAO
+class HttpJupyterDAO[F[_]: Timer: ContextShift: Concurrent](val clusterDnsCache: ClusterDnsCache[F], client: Client[F])
+    extends JupyterDAO[F]
     with LazyLogging {
-
-  val http = Http(system)
```
rtitle, line 26: Cool, no more akka-http client here..

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpRStudioDAO.scala
```diff
@@ -11,15 +12,17 @@ import org.broadinstitute.dsde.workbench.model.google.GoogleProject
...
-class HttpRStudioDAO(val clusterDnsCache: ClusterDnsCache)(implicit system: ActorSystem,
-                                                           executionContext: ExecutionContext)
+class HttpRStudioDAO(val clusterDnsCache: ClusterDnsCache[IO])(implicit system: ActorSystem,
+                                                           executionContext: ExecutionContext,
+                                                               cs: ContextShift[IO],
+                                                               timer: Timer[IO])
     extends RStudioDAO
     with LazyLogging {
 
   val http = Http(system)
```
rtitle: Do we want to convert `HttpRStudioDAO` to http4s too? :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/DbReference.scala
```diff
@@ -100,3 +105,7 @@ class DataAccess(val profile: JdbcProfile)(implicit val executionContext: Execut
     sql"select version()".as[String]
   }
 }
+
+final class DBIOOps[A](private val dbio: DBIO[A]) extends AnyVal {
+  def transaction(implicit dbRef: DbReference[IO]): IO[A] = dbRef.inTransaction(dbio)
```
rtitle, line 117: I wonder if this `DBIOOps` class should have another method that takes a `TransactionIsolation` param and passes it to `inTransaction`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/Boot.scala
```diff
@@ -192,12 +191,12 @@ object Boot extends IOApp with LazyLogging {
       clientWithRetryAndLogging = Http4sLogger[F](logHeaders = true, logBody = false)(clientWithRetryWithCustomSSL)
 
       samDao = HttpSamDAO[F](clientWithRetryAndLogging, httpSamDap2Config, blocker)
-      dbRef <- Resource.make(ConcurrentEffect[F].delay(DbReference.init(liquibaseConfig)))(
-        db => ConcurrentEffect[F].delay(db.database.close)
-      )
-      clusterDnsCache = new ClusterDnsCache(proxyConfig, dbRef, clusterDnsCacheConfig)
+      concurrentDbAccessPermits <- Resource.liftF(Semaphore[F](1000L))
```
rtitle: Can we put this in config?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/DbReference.scala
```diff
@@ -43,42 +43,47 @@ object DbReference extends LazyLogging {
...
   }
 
-  def inTransactionIO[T](
-    f: (DataAccess) => DBIO[T],
+  def inTransaction[T](
+    dbio: DBIO[T],
     isolationLevel: TransactionIsolation = TransactionIsolation.RepeatableRead
-  )(implicit cs: ContextShift[IO]): IO[T] =
-    IO.fromFuture(IO(inTransaction(f, isolationLevel)))
+  ): F[T] = concurrentDbAccessPermits.withPermit(blocker.blockOn(Async.fromFuture(Async[F].delay(inTransactionFuture(dbio, isolationLevel)))))
```
rtitle: So if I'm understanding this right this runs DB operations on a Blocker backed by a `cachedThreadPool` -- but with a semaphore enforcing a max concurrent access of 1000.  This seems reasonable to me.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/DbReference.scala
Qi77Qi: I don't see `TransactionIsolation` being used anywhere..but I can definitely add it

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/DbReference.scala
Qi77Qi: yep

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
```diff
@@ -201,9 +201,9 @@ class ClusterMonitorActor(
         if (cluster.status == Starting) {
           for {
             _ <- persistInstances(cluster, googleInstances)
-            _ <- clusterHelper.stopCluster(cluster).unsafeToFuture()
-            now <- IO(Instant.now).unsafeToFuture()
-            _ <- dbRef.inTransaction { _.clusterQuery.setToStopping(cluster.id, now) }
+            _ <- clusterHelper.stopCluster(cluster)
+            now <- IO(Instant.now)
+            _ <- dbRef.inTransaction { dbRef.dataAccess.clusterQuery.setToStopping(cluster.id, now) }
```
rtitle: For classes that do a lot of DB stuff we could store the clusterQuery up front:
```
val clusterQuery = dbRef.dataAccess.clusterQuery
```
then later:
```
dbRef.inTransaction { clusterQuery.setToStopping(…) }
```
Just thinking out loud about ways to reduce syntax..

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ZombieClusterMonitor.scala
```diff
@@ -51,19 +51,20 @@ class ZombieClusterMonitor(config: ZombieClusterConfig,
...
-      val zombieClusters = getActiveClustersFromDatabase.flatMap { clusterMap =>
+      val zombieClusters = getActiveClustersFromDatabase.unsafeToFuture().flatMap { clusterMap =>
         clusterMap.toList.flatTraverse {
           case (project, clusters) =>
             // Check if the project is active
             isProjectActiveInGoogle(project).flatMap {
               case true =>
                 // If the project is active, check each individual cluster
-                logger.debug(s"Project ${project.value} containing ${clusters.size} clusters is active in Google")
+                logger.debug(s"Project ${project.value} containing ${clusters.size} clusters that're active in Google")
```
rtitle: I think "is" is correct - it's referring to the project

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/ProxyService.scala
```diff
@@ -97,26 +98,25 @@ class ProxyService(
...
       }
     )
 
   def getCachedClusterInternalId(googleProject: GoogleProject, clusterName: ClusterName)(
     implicit ev: ApplicativeAsk[IO, TraceId]
-  ): Future[ClusterInternalId] =
-    clusterInternalIdCache.get((googleProject, clusterName)).flatMap {
-      case Some(clusterInternalId) => Future.successful(clusterInternalId)
+  ): IO[ClusterInternalId] =
+    IO(clusterInternalIdCache.get((googleProject, clusterName))).flatMap {
```
rtitle: This might need `blocker.blockOn(IO(clusterInternalIdCache.get(…)))`. I remember being bitten in the past by this.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/ProxyService.scala
```diff
@@ -169,56 +165,47 @@ class ProxyService(
...
+  def getTargetHost(googleProject: GoogleProject, clusterName: ClusterName): IO[HostStatus] =
+    Proxy.getTargetHost[IO](clusterDnsCache, googleProject, clusterName)
+
   private def proxyInternal(userInfo: UserInfo,
                             googleProject: GoogleProject,
                             clusterName: ClusterName,
-                            request: HttpRequest): Future[HttpResponse] = {
-    logger.debug(s"Received proxy request for user $userInfo")
+                            request: HttpRequest): IO[HttpResponse] = {
+    logger.info(s"Received proxy request for user $userInfo: ${clusterDnsCache.stats} / ${clusterDnsCache.size}")
```
rtitle: I think we should leave this at `debug` -- proxy requests can be pretty high volume.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/ProxyService.scala
```diff
@@ -169,56 +165,47 @@ class ProxyService(
...
+            handleWebSocketRequest(targetHost, request, upgrade)
+          case None          =>
+            handleHttpRequest(targetHost, request)
         }
-        responseFuture recoverWith {
+
+        IO.fromFuture(IO(responseFuture)).flatMap {
+          r =>
+            if (r.status.isFailure())
+              IO(logger.info(s"Error response for proxied request: ${r.status}")).as(r)
```
rtitle: Not sure about the volume of logs here either, but this may be nice to have. We could also log the request path here.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/ProxyService.scala
```diff
@@ -248,6 +235,8 @@ class ProxyService(
     // 3. build a new HttpRequest
     val newRequest = request.copy(headers = newHeaders, uri = newUri)
 
+    logger.info(s"proxying request ${newRequest}")
```
rtitle: Remove this

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/ProxyService.scala
```diff
@@ -359,3 +340,29 @@ class ProxyService(
     "Connection"
   ).map(_.toLowerCase)
 }
+
+object ProxyService {
+  // From the outside, we are going to support TWO paths to access Jupyter:
+  // 1)   /notebooks/{googleProject}/{clusterName}/
+  // 2)   /proxy/{googleProject}/{clusterName}/jupyter/
+  // 3)   /notebooks/{googleProject}/{clusterName}/jupyter/
```
rtitle, line 364: I think we can remove case 3 here. `/notebooks/{googleProject}/{clusterName}/jupyter/` was never a supported proxy path. We previously only had `/notebooks/{googleProject}/{clusterName}/`, and we are trying to migrate to `/proxy/{googleProject}/{clusterName}/jupyter/`.

I think FireCloud is probably the only system on the legacy path at this point..

rtitle:
I did a pass (only skimmed the tests). In general it looks good, I like the approach taken with `DbRef`. Some comments/cleanup above. I'd like to do another pass before merge.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/ProxyService.scala
Qi77Qi: yea, this was when I was debugging the weird path issue....will revert

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/ProxyService.scala
Qi77Qi: I added 3 because somehow in automation tests, all proxied requests have this path

Qi77Qi:
https://fc-jenkins.dsp-techops.broadinstitute.org/view/Leonardo/job/LeoPerfTestRunner/277/gatling/report/leocreatemonitordeletesimulation-20200114045348493/ looks better than dev runs in terms of response time except `/lock`, which is due to the following error
```
Caused by: com.google.api.client.googleapis.json.GoogleJsonResponseException: 403 Forbidden
{
  "code" : 403,
  "errors" : [ {
    "domain" : "global",
    "message" : "pet-110530393451290928813@billing-leo-env-perf-228.iam.gserviceaccount.com does not have storage.objects.list access to leo-perf-test-do-not-delete.",
    "reason" : "forbidden"
  } ],
  "message" : "pet-110530393451290928813@billing-leo-env-perf-228.iam.gserviceaccount.com does not have storage.objects.list access to leo-perf-test-do-not-delete."
}
```

Still investigating why the error happened

http/src/main/resources/reference.conf
```diff
@@ -91,6 +91,7 @@ mysql {
     connectionTimeout = 5000
     numThreads = 50
   }
+  concurrency = 120
```
rtitle, line 94: Trying to understand how `concurrency` relates to `numThreads`. As I understand it slick uses a thread pool internally. So do we have this situation?
```
global ec --> (cached thread pool with 120-limit semaphore) --> (slick thread pool of 50)
```
I am wondering if `concurrency` and `numThreads` should be the same value… what do you think?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/Boot.scala
```diff
@@ -192,14 +191,20 @@ object Boot extends IOApp with LazyLogging {
       httpClientWithCustomSSL <- blaze.BlazeClientBuilder[F](blockingEc, Some(sslContext)).resource
       clientWithRetryWithCustomSSL = Retry(retryPolicy)(httpClientWithCustomSSL)
       clientWithRetryAndLogging = Http4sLogger[F](logHeaders = true, logBody = false)(clientWithRetryWithCustomSSL)
+      clientWithNoRetryNoHeaderLogging = Http4sLogger[F](logHeaders = false, logBody = false)(httpClientWithCustomSSL)
+      // TODO: Use this client without retry for RStudioDAO because `checkClusterStatus` in `ClusterToolMonitor` will check
```
rtitle, line 195: Ah good catch, yeah we should fix this (can be in a separate PR).

rtitle: I guess RStudio-enabled clusters will have the same problem, but with Jupyter failing. Maybe I'll put in a PR for this.

http/src/main/resources/reference.conf
Qi77Qi: I think `numThreads` is probably physical threads, which is different from `IO`s. See https://typelevel.org/cats-effect/concurrency/basics.html#green-threads...so I think we can have more `IO`s than jvm threads.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/ProxyRoutes.scala
```diff
@@ -60,14 +64,16 @@ trait ProxyRoutes extends UserInfoDirectives with CorsSupport with CookieHelper
...
                 complete {
-                  val proxyFuture = proxyService.proxyRequest(userInfo, googleProject, clusterName, request)
                   // we are discarding the request entity here. we have noticed that PUT requests caused by
                   // saving a notebook when a cluster is stopped correlate perfectly with CPU spikes.
                   // in that scenario, the requests appear to pile up, causing apache to hog CPU.
-                  proxyFuture.failed.foreach { _ =>
-                    request.entity.discardBytes().future
-                  }
-                  proxyFuture
+                  for {
```
rtitle: Do you need the for-comp? Could this just be:
```
complete {
  // we are discarding the request entity here. we have noticed that PUT requests caused by
  // saving a notebook when a cluster is stopped correlate perfectly with CPU spikes.
  // in that scenario, the requests appear to pile up, causing apache to hog CPU.
  proxyService.proxyRequest(userInfo, googleProject, clusterName, request).onError {
    case e =>
      IO(logger.warn(s"proxy request failed for ${userInfo} ${googleProject} ${clusterName}", e)) <* IO
        .fromFuture(IO(request.entity.discardBytes().future))
}
```
?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/Boot.scala
Qi77Qi: that would be great...I tried to do it...but it wasn't trivial..so just putting a todo here for now

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/ProxyRoutes.scala
Qi77Qi: ah probably not...will update

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpJupyterDAO.scala
```diff
@@ -1,54 +1,52 @@
...
+            uri = Uri.unsafeFromString(
+              s"https://${targetHost.toString}/notebooks/$googleProject/$clusterName/api/status"
+            )
+          )
+        )
+      case _ => Concurrent[F].pure(false)
     }
 
-  def isAllKernalsIdle(googleProject: GoogleProject, clusterName: ClusterName): Future[Boolean] =
+  def isAllKernalsIdle(googleProject: GoogleProject, clusterName: ClusterName): F[Boolean] =
```
rtitle: This was here before, but `kernel` is misspelled :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
```diff
@@ -195,31 +208,14 @@ trait ClusterComponent extends LeoComponent {
...
             MachineConfigRecord.tupled.apply(machineConfig),
-            properties
-              .map(
-                x =>
-                  x.as[Map[String, String]]
-                    .fold(e => throw new RuntimeException(s"fail to read `properties` field due to ${e.getMessage}"),
-                          identity)
-              )
-              .getOrElse(Map.empty), //in theory, throw should never happen
+            properties.getOrElse(Map.empty), //in theory, throw should never happen
```
rtitle: Can remove comment I think

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
```diff
@@ -195,31 +208,14 @@ trait ClusterComponent extends LeoComponent {
...
-                x =>
-                  x.as[Map[String, String]]
-                    .fold(e =>
-                            throw new RuntimeException(
-                              s"fail to read `customClusterEnvironmentVariables` field due to ${e.getMessage}"
-                            ),
-                          identity)
-              )
-              .getOrElse(Map.empty) //in theory, throw should never happen
+            customClusterEnvironmentVariables.getOrElse(Map.empty) //in theory, throw should never happen
```
rtitle: Same as above

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpJupyterDAO.scala
Qi77Qi: it was probably me who introduced the typo :) will update

rtitle (approved):
I think this looks good.. I especially like the DB class refactor, I think it makes things more readable. Giving my tentative 👍, but please do make sure perf/automation tests pass on the latest changes.

Qi77Qi:
Did a few perf tests and they look good..going to merge this

PR #1166, opened 2019-12-04
IA-1535 reduce concurrent cluster creation requests

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-1535

I'm hoping reducing the number of concurrent cluster creation requests will make tests more reliable

---

In all cases:

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoTestUtils.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoTestUtils.scala
@@ -5,6 +5,7 @@ import java.nio.charset.StandardCharsets
 import java.nio.file.Files
 
 import akka.actor.ActorSystem
+import cats.effect.concurrent.Semaphore
 import cats.effect.{Blocker, IO}
 import cats.implicits._
 import com.typesafe.scalalogging.LazyLogging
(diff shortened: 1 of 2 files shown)


rtitle (approved):
We can try it but I'm not sure it'll have much effect since the bucket creation is async to the API request. Back Leo sweeps the database every 15 seconds and starts the monitor actors. I just checked and we do have some jitter when starting the `ClusterMonitorActor`: https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala#L179 to try and spread things out.

One thing to confirm: is google2 `insertBucket` _actually_ retrying when it gets this rate limit exception? Also Is there a way to add some jitter between retries, so all retries don't just happen in sync?

Qi77Qi:
Yea, it's retried https://github.com/broadinstitute/workbench-libs/blob/develop/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala#L200


https://github.com/broadinstitute/workbench-libs/blob/8bae8e8e57442987d3f4afc25e8be63ee3b47d99/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleDataprocInterpreter.scala#L141 is where jitter happens

Qi77Qi:
I'm going to merge this for now...and continue monitor test runs

PR #1168, opened 2019-12-05
IA-1504 Implement jupyterStartUserScriptUri feature

calbach:
https://broadworkbench.atlassian.net/browse/IA-1504

Design: https://docs.google.com/document/d/11IP3zflxzjN5QlX5eXc8OpzvO1g6XgjHpAt8arBnkwQ/edit#
Ticket: https://broadworkbench.atlassian.net/browse/IA-1504

Notes:
- Output log convention is slightly different, as the start script will have multiple executions (one per resume).
- Only real logic for this PR is in the shell scripts
- Have not manually tested this yet
- Also fixed EXIT_CODE logic on initialization script

@rtitle FYI

---

In all cases:


automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoModelCopy.scala
```diff
@@ -54,7 +55,8 @@ case class DefaultLabels(clusterName: ClusterName,
                          clusterServiceAccount: Option[WorkbenchEmail],
                          notebookServiceAccount: Option[WorkbenchEmail],
                          notebookExtension: Option[String],
-                         notebookUserScript: Option[String]) {
+                         notebookUserScript: Option[String],
+                         jupyterUserScriptUri: Option[String]) {
```
jmthibault79: [drive by comment without real understanding]

Do you mean `jupyterStartUserScriptUri` here?

calbach: Good catch - this was wrong. I think to match the convention here I want `notebookStartUserScript`; unclear to me why this is different than the rest of the naming.

Qi77Qi:
you need to use liquid base to tell mysql that there's schema change. Add a similar file to https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20191107_add_cluster_custom_environment_variables.xml, and then append it to https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changelog.xml

Use `sbt compile` to make sure everything compiles.
Try run `sbt http/test` to reproduce error.

`sbt` is a bit slow. I recommend run `sbt`, and then you'll be in sbt shell, from there, do `project http`, and then you can do `~compile` to have it compiling continuously, or `test`, or `testOnly xxx` for running one single test

calbach:
Added automation test, `sbt automation/compile` worked

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookClusterMonitoringSpec.scala
```diff
@@ -7,7 +7,7 @@ import org.broadinstitute.dsde.workbench.dao.Google.googleStorageDAO
 import org.broadinstitute.dsde.workbench.leonardo._
 import org.broadinstitute.dsde.workbench.model.google.GcsEntityTypes.Group
 import org.broadinstitute.dsde.workbench.model.google.GcsRoles.Reader
-import org.broadinstitute.dsde.workbench.model.google.{EmailGcsEntity, GcsObjectName, GcsPath, parseGcsPath}
+import org.broadinstitute.dsde.workbench.model.google.{EmailGcsEntity, GcsEntityTypes, GcsObjectName, GcsPath, GcsRoles, parseGcsPath}
```
rtitle: Seems like these could be unused imports?

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookCustomizationSpec.scala
```diff
@@ -136,5 +136,52 @@ final class NotebookCustomizationSpec extends GPAllocFixtureSpec with ParallelTe
...
+          googleStorageDAO.setObjectAccessControl(bucketName,
+            objectName,
+            EmailGcsEntity(GcsEntityTypes.User, ronPetServiceAccount),
+            GcsRoles.Owner)
+
+          withNewCluster(billingProject,
+            request = defaultClusterRequest.copy(jupyterStartUserScriptUri = Some(startScriptUri))) {
+          cluster =>
+            withWebDriver { implicit driver =>
+              withNewNotebookInSubfolder(cluster, Python3) { notebookPage =>
```
rtitle: Minor: maybe these tests could use `withNewNotebook` instead of `withNewNotebookInSubfolder` since I don't think we necessarily need subfolders, and subfolder creation adds selenium clicks and complexity.

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookCustomizationSpec.scala
```diff
@@ -136,5 +136,52 @@ final class NotebookCustomizationSpec extends GPAllocFixtureSpec with ParallelTe
...
+        val ronPetServiceAccount = Sam.user.petServiceAccountEmail(billingProject.value)(ronAuthToken)
+        googleStorageDAO.setBucketAccessControl(bucketName,
+          EmailGcsEntity(GcsEntityTypes.User, ronPetServiceAccount),
+          GcsRoles.Owner)
+
+        // Add the user script to the bucket. This script increments and writes a count to file,
+        // tracking the number of times it has been invoked.
+        val startScriptString = "#!/usr/bin/env bash\n\n" +
+          "count=$(cat ~/leo_test_start_count.txt || echo 0)\n" +
+          "echo $(($count + 1)) > ~/leo_test_start_count.txt"
```
rtitle: Haven't finished the PR yet but does the start script run as `root` or `jupyter-user`? If `root` then `~` won't go to the expected place.

http/src/main/resources/jupyter/init-actions.sh
```diff
@@ -348,17 +348,18 @@ END
...
         gsutil cp ${JUPYTER_USER_SCRIPT_URI} /etc
         JUPYTER_USER_SCRIPT=`basename ${JUPYTER_USER_SCRIPT_URI}`
         docker cp /etc/${JUPYTER_USER_SCRIPT} ${JUPYTER_SERVER_NAME}:${JUPYTER_HOME}/${JUPYTER_USER_SCRIPT}
         retry 3 docker exec -u root ${JUPYTER_SERVER_NAME} chmod +x ${JUPYTER_HOME}/${JUPYTER_USER_SCRIPT}
         # Execute the user script as privileged to allow for deeper customization of VM behavior, e.g. installing
         # network egress throttling. As docker is not a security layer, it is assumed that a determined attacker
         # can gain full access to the VM already, so using this flag is not a significant escalation.
-        docker exec --privileged -u root -e PIP_USER=false ${JUPYTER_SERVER_NAME} ${JUPYTER_HOME}/${JUPYTER_USER_SCRIPT} &> us_output.txt || EXIT_CODE=$? && true ;
+        EXIT_CODE=0
+        docker exec --privileged -u root -e PIP_USER=false ${JUPYTER_SERVER_NAME} ${JUPYTER_HOME}/${JUPYTER_USER_SCRIPT} &> us_output.txt || EXIT_CODE=$?
```
rtitle, line 377: This seems correct, but I'm not sure if that `&& true` was added for any particular reason...

http/src/main/resources/jupyter/init-actions.sh
```diff
@@ -368,6 +369,15 @@ END
         fi
       fi
 
+      # If a Jupyter start user script was specified, copy it into the jupyter docker container for consumption by startup.sh.
+      if [ ! -z ${JUPYTER_START_USER_SCRIPT_URI} ] ; then
```
rtitle, line 388: I don't see where `$JUPYTER_START_USER_SCRIPT_URI` is being set… it may need to be exported in the top of `init-actions.sh` and templated in by Leo.

http/src/main/resources/jupyter/startup.sh
```diff
@@ -54,6 +57,22 @@ if [ "$UPDATE_WELDER" == "true" ] ; then
     docker-compose -f /etc/welder-docker-compose.yaml up -d
 fi
 
+# If a Jupyter start user script was specified, execute it now. It should already be in the docker container
+# via initialization in init-actions.sh (we explicitly do not want to recopy it from GCS on every cluster resume).
+if [ ! -z ${JUPYTER_START_USER_SCRIPT_URI} ] ; then
+  JUPYTER_START_USER_SCRIPT=`basename ${JUPYTER_START_USER_SCRIPT_URI}`
+  log 'Executing Jupyter user start script [$JUPYTER_START_USER_SCRIPT]...'
```
rtitle, line 101: I don't think the `log` function is in this file (should we add it?)

http/src/main/resources/jupyter/startup.sh
```diff
@@ -54,6 +57,22 @@ if [ "$UPDATE_WELDER" == "true" ] ; then
...
+  EXIT_CODE=0
+  docker exec --privileged -u root -e PIP_USER=false ${JUPYTER_SERVER_NAME} ${JUPYTER_HOME}/${JUPYTER_START_USER_SCRIPT} &> start_output.txt || EXIT_CODE=$?
+  if [ $EXIT_CODE -ne 0 ]; then
+    echo "User start script failed with exit code ${EXIT_CODE}. Output is saved to ${JUPYTER_START_USER_SCRIPT_OUTPUT_URI}"
+    retry 3 gsutil -h "x-goog-meta-passed":"false" cp start_output.txt ${JUPYTER_START_USER_SCRIPT_OUTPUT_URI}
```
rtitle, line 106: I'm wondering what will happen if the start user script has a bug. This call will fail, and the startup script will exit. The cluster will remain in `Starting` until Leo times it out and it moves back to `Stopped`. Is this the behavior we want?
* Should we allow clients to modify the user startup script for `Stopped` clusters so they can iterate without a delete/recreate?
* Should we still startup Jupyter in this case _without_ running the user startup script?

Not sure the answer yet, thinking out loud..

http/src/main/resources/jupyter/startup.sh
```diff
@@ -54,6 +57,22 @@ if [ "$UPDATE_WELDER" == "true" ] ; then
...
+    exit $EXIT_CODE
+  else
+    retry 3 gsutil -h "x-goog-meta-passed":"true" cp start_output.txt ${JUPYTER_START_USER_SCRIPT_OUTPUT_URI}
```
rtitle, line 109: Similarly I don't think `retry` is in here

http/src/main/resources/jupyter/init-actions.sh
rtitle: Based on further reading it appears it is templated by Leo, so just needs to be exported above

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookClusterMonitoringSpec.scala
calbach: Oops, reverted this file

rtitle (changes requested):
Left some comments above.. in general though I think you hit the right major pieces. :)

http/src/main/resources/jupyter/init-actions.sh
calbach: Oops thanks, forgot to add that. Done

http/src/main/resources/jupyter/init-actions.sh
calbach: The whole script is on `set -e`; presumably this was added to allow cluster start to succeed even if the init script fails

http/src/main/resources/jupyter/startup.sh
calbach: Oops, copied here

http/src/main/resources/jupyter/startup.sh
calbach: Good question. How about on the initial cluster creation? Will it enter the error state if this script fails to execute?

The user script itself is actually not a huge deal in my view. The user already has control over the return code from that script, they can choose whether or not to include `set -e`, for example. So having a non-zero exit code crash things seems good there.

I think it's a general question to apply to this entire script, which could fail transiently for other reasons though. In my opinion the cluster should enter the error state if this happens. Possibly following N stop/start retries by the Leo monitor.

http/src/main/resources/jupyter/startup.sh
calbach: Added

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookCustomizationSpec.scala
calbach: Yeah, it's root. Will use $JUPYTER_HOME in that case - should be safe I think

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookCustomizationSpec.scala
calbach: Done, thanks

calbach (commented):
Thanks! feedback applied. I think it's safe to say that original Jenkins run should fail

http/src/main/resources/jupyter/init-actions.sh
calbach: Sorry - correction: it exits anyway if the script fails, but it wants to copy the log before exiting `init-actions.sh`, hence the `|| true`

rtitle:
The test failure is:
```
 org.scalatest.exceptions.TestFailedException: "[cat: /etc/jupyter/leo_test_start_count.txt: No such file or directory]" was not equal to "[1]"
      at org.scalatest.MatchersHelper$.indicateFailure(MatchersHelper.scala:343)
      at org.scalatest.Matchers$AnyShouldWrapper.shouldBe(Matchers.scala:6919)
      at org.broadinstitute.dsde.workbench.leonardo.notebooks.NotebookCustomizationSpec.$anonfun$new$26(NotebookCustomizationSpec.scala:168)
      at org.broadinstitute.dsde.workbench.leonardo.notebooks.NotebookTestUtils.$anonfun$withNewNotebook$5(NotebookTestUtils.scala:114)
      at org.broadinstitute.dsde.workbench.leonardo.notebooks.NotebooksListPage.$anonfun$withNewNotebook$2(NotebooksListPage.scala:90)
      at scala.util.Try$.apply(Try.scala:213)
      at org.broadinstitute.dsde.workbench.leonardo.notebooks.NotebooksListPage.withNewNotebook(NotebooksListPage.scala:90)
      at org.broadinstitute.dsde.workbench.leonardo.notebooks.NotebookTestUtils.$anonfun$withNewNotebook$4(NotebookTestUtils.scala:113)
      at scala.concurrent.Future$.$anonfun$apply$1(Future.scala:659)
      at scala.util.Success.$anonfun$map$1(Try.scala:255)
      at scala.util.Success.map(Try.scala:213)
      at scala.concurrent.Future.$anonfun$map$1(Future.scala:292)
      at scala.concurrent.impl.Promise.liftedTree1$1(Promise.scala:33)
      at scala.concurrent.impl.Promise.$anonfun$transform$1(Promise.scala:33)
      at scala.concurrent.impl.CallbackRunnable.run(Promise.scala:64)
      at java.util.concurrent.ForkJoinTask$RunnableExecuteAction.exec(ForkJoinTask.java:1402)
      at java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:289)
      at java.util.concurrent.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:1056)
      at java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1692)
      at java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:157)
```
It seems like the file the startup script was supposed to create doesn't exist. So the startup script didn't run correctly for whatever reason. I'll look around a bit..

http/src/main/resources/jupyter/init-actions.sh
rtitle: Yeah - it the current version will it copy the log even if the user script fails? I think so -- which is why I don't see why `&& true` was necessary.

rtitle:
@Qi77Qi's `add fluentd for user startup script` commit should help debug cases like this :)

http/src/main/resources/jupyter/init-actions.sh
calbach: Ah yes, agreed. Sorry, I thought we were talking about the `|| EXIT_CODE=` piece; forgot `&& true` was even there. AFAICT the `&& true` was a no-op.

calbach (commented):
Thanks, I am at you all's mercy to debug further since I cannot see Jenkins logs (VPN access soon...)

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookCustomizationSpec.scala
```diff
@@ -161,7 +161,7 @@ final class NotebookCustomizationSpec extends GPAllocFixtureSpec with ParallelTe
             GcsRoles.Owner)
 
           withNewCluster(billingProject,
-            request = defaultClusterRequest.copy(jupyterStartUserScriptUri = Some(startScriptUri))) {
+            request = defaultClusterRequest.copy(jupyterStartUserScriptUri = Some(startScriptUri), jupyterUserScriptUri = Some(startScriptUri))) {
```
calbach: Why add this? Now I expect this script to be run 3x, so the test should hopefully fail

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
```diff
@@ -101,6 +102,7 @@ trait ClusterComponent extends LeoComponent {
     def destroyedDate = column[Timestamp]("destroyedDate", O.SqlType("TIMESTAMP(6)"))
     def jupyterExtensionUri = column[Option[String]]("jupyterExtensionUri", O.Length(1024))
     def jupyterUserScriptUri = column[Option[String]]("jupyterUserScriptUri", O.Length(1024))
+    def jupyterStartUserScriptUri = column[Option[String]]("jupyterStartUserScriptUri", O.Length(1024))
```
Qi77Qi: minor, if you add
```
  implicit private val gsPathMappedColumnType: BaseColumnType[GcsPath] =
    MappedColumnType
      .base[GcsPath, String](_.toUri, s => parseGcsPath(s).fold(e => throw e, identity))
```
then you can do  `def jupyterStartUserScriptUri = column[Option[GcsPath]]("jupyterStartUserScriptUri", O.Length(1024))` here and you don't need `_.toUri` and `parseGcsPath` in different places.

The `MappedColumnType` defined above will throw exception if the value we read from DB doesn't conform to `gs://xx/xx`, vs your current implementation won't throw exception and instead just resolve invalid `jupyterStartUserScriptUri` to `None`...

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/leonardoServiceContractModel.scala
```diff
@@ -27,6 +27,7 @@ final case class ListClusterResponse(id: Long,
                                      labels: LabelMap,
                                      jupyterExtensionUri: Option[GcsPath],
                                      jupyterUserScriptUri: Option[GcsPath],
+                                     jupyterStartUserScriptUri: Option[GcsPath],
```
Qi77Qi: do we really intend to expose this in `ListCluster` API as well? If so, this needs to be reflected in swagger doc

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/CommonTestData.scala
```diff
@@ -177,6 +181,7 @@ trait CommonTestData { this: ScalaFutures =>
     labels = Map(),
     jupyterExtensionUri = Option(GcsPath(GcsBucketName("bucketName"), GcsObjectName("extension"))),
     jupyterUserScriptUri = Option(GcsPath(GcsBucketName("bucketName"), GcsObjectName("userScript"))),
+    jupyterStartUserScriptUri = Option(GcsPath(GcsBucketName("bucketName"), GcsObjectName("startScript"))),
```
Qi77Qi: minor, Use `Option(xxx)` does an additional `null` check, so here you can just use `Some` now that there's no way `GcsPath(GcsBucketName("bucketName"), GcsObjectName("startScript"))` is null

http/src/main/resources/jupyter/init-actions.sh
```diff
@@ -177,6 +180,18 @@ END
  read_from_head true
  tag jupyter
 </source>
+END
+
+    # Add stack driver configuration for user start up script
+    tee /etc/google-fluentd/config.d/userStartupScirpt.conf << END
```
Qi77Qi: Rob added similar config for this in a PR just merged....so we can remove this now.

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/CommonTestData.scala
calbach: Done

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/leonardoServiceContractModel.scala
calbach: Added. I don't really care whether these get returned on list, but it would be inconsistent to return one of these script URIs and not the other

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
calbach: Done, seems to have worked. GcsParseError is not a throwable though, let me know if you have a better exception type for me to wrap this in.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/leonardoServiceContractModel.scala
Qi77Qi: If you don't care about the field for now, let's leave it out. I prefer exposing fields only when needed. `ListCluster` and getCluster API do differ anyway. This won't be the only difference. `GetCluster` joins a few tables, gives back more fields and hence less performant. Even though in this case, returning  `jupyterStartUserScriptUri` doesn't require joining a different table for `ListCluster` today, we might rework our DB schema.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
Qi77Qi: it is internal server error in this case, so Exception should be fine...

Qi77Qi (approved):

rtitle (approved):

PR #1160, opened 2019-11-27
IA-1427 deprecate image and tool fields instead of removing

Qi77Qi:
Per https://broadinstitute.slack.com/archives/CA3HY2Q3S/p1574875370231800

---
Have you read [CONTRIBUTING.md](https://github.com/DataBiosphere/leonardo/blob/develop/CONTRIBUTING.md) lately? If not, do that first.

I, the developer opening this PR, do solemnly pinky swear that:

- [ ] I've documented my API changes in Swagger

In all cases:

- [ ] Get a thumbsworth of review and PO signoff if necessary
- [ ] Verify all tests go green
- [ ] Squash and merge; Delete your branch after this
- [ ] Test this change deployed correctly and works on dev environment after deployment

PR #1169, opened 2019-12-05
IA-1532 Revert "IA-1427 deprecate image and tool fields instead of removing"

Qi77Qi:
Reverts DataBiosphere/leonardo#1160

https://broadworkbench.atlassian.net/browse/IA-1532

https://github.com/DataBiosphere/terra-ui/pull/1956 is merged, so this should be good to go


rtitle (approved):
Is AoU ok with this change?

Qi77Qi:
@rtitle yea, we had a convo with AOU, and the change was in dev for a few days before we realize it broke terra-ui and made #1160 to accommodate terra-ui

PR #1170, opened 2019-12-06
[IA-1544 IA-1494] Persist RStudio session across restarts

rtitle:
https://broadworkbench.atlassian.net/browse/IA-1544

I've tested this manually on a fiab and it seems to work. Automated testing depends on https://broadworkbench.atlassian.net/browse/IA-697.

---

In all cases:

### DIFF
--- a/automation/src/test/resources/reference.conf
+++ b/automation/src/test/resources/reference.conf
@@ -6,4 +6,4 @@ leonardo.baseImageUrl = "us.gcr.io/broad-dsp-gcr-public/terra-jupyter-base:0.0.6
 leonardo.oldWelderImage = "us.gcr.io/broad-dsp-gcr-public/welder-server:6a783a5"
 leonardo.currentWelderImage = "us.gcr.io/broad-dsp-gcr-public/welder-server:60e28bc" # This needs to be exactly the same in http/src/main/resources/reference.conf
 leonardo.bioconductorImageUrl = "us.gcr.io/broad-dsp-gcr-public/terra-jupyter-bioconductor:0.0.7"
-leonardo.rstudioBaseImageUrl = "us.gcr.io/anvil-gcr-public/anvil-rstudio-base:0.0.1"
+leonardo.rstudioBaseImageUrl = "us.gcr.io/anvil-gcr-public/anvil-rstudio-base:0.0.2"
--- a/http/src/main/resources/jupyter/init-actions.sh
+++ b/http/src/main/resources/jupyter/init-actions.sh
@@ -179,6 +179,30 @@ END
 </source>
 END
 
+    # Add stack driver configuration for user start up script
+    tee /etc/google-fluentd/config.d/userStartupScript.conf << END
+<source>
+ @type tail
+ format none
+ path /var/log/user.log
+ pos_file /var/tmp/fluentd.google.user.startup.pos
+ read_from_head true
+ tag user_startup
+</source>
+END
+
+    # Add stack driver configuration for user shut down script
+    tee /etc/google-fluentd/config.d/userShutdownScript.conf << END
+<source>
+ @type tail
+ format none
+ path /var/log/user.log
+ pos_file /var/tmp/fluentd.google.user.shutdown.pos
+ read_from_head true
+ tag user_shutdown
+</source>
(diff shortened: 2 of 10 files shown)


http/src/main/resources/jupyter/init-actions.sh
```diff
@@ -177,6 +177,30 @@ END
...
+ @type tail
+ format none
+ path /var/log/user.log
+ pos_file /var/tmp/fluentd.google.user.startup.pos
+ read_from_head true
+ tag user_startup
+</source>
+END
+
+    # Add stack driver configuration for user shut down script
```
rtitle: I didn't actually test this yet… cc @Qi77Qi

automation/src/test/resources/reference.conf
```diff
@@ -6,4 +6,4 @@ leonardo.baseImageUrl = "us.gcr.io/broad-dsp-gcr-public/terra-jupyter-base:0.0.6
 leonardo.oldWelderImage = "us.gcr.io/broad-dsp-gcr-public/welder-server:6a783a5"
 leonardo.currentWelderImage = "us.gcr.io/broad-dsp-gcr-public/welder-server:60e28bc" # This needs to be exactly the same in http/src/main/resources/reference.conf
 leonardo.bioconductorImageUrl = "us.gcr.io/broad-dsp-gcr-public/terra-jupyter-bioconductor:0.0.7"
-leonardo.rstudioBaseImageUrl = "us.gcr.io/anvil-gcr-public/anvil-rstudio-base:0.0.1"
+leonardo.rstudioBaseImageUrl = "us.gcr.io/anvil-gcr-public/anvil-rstudio-base:0.0.2"
```
rtitle, line 9: I manually bumped a couple rstudio images in the AnVIL GCR.. see https://github.com/anvilproject/anvil-docker/pull/3

Also the bioconductor one that Nitesh made seems to work ok, based on manual testing.

http/src/main/resources/jupyter/init-actions.sh
```diff
@@ -177,6 +177,30 @@ END
...
+    tee /etc/google-fluentd/config.d/userShutdownScript.conf << END
+<source>
+ @type tail
+ format none
+ path /var/log/user.log
```
Qi77Qi: this path should be different?

rtitle: fixed

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
```diff
@@ -619,6 +627,34 @@ class ClusterHelper(
     }
   }
 
+  private def getMasterInstanceShutdownScript(cluster: Cluster): IO[Map[String, String]] = {
+    val googleKey = "shutdown-script" // required; see https://cloud.google.com/compute/docs/shutdownscript
+
+    // These things need to be provided to ClusterInitValues, but aren't actually needed for the shutdown script
+    val dummyInitBucket = GcsBucketName("")
+    val dummyStagingBucket = GcsBucketName("")
```
kyuksel: Would it be better to not overload `ClusterInitValues` for this purpose? If we must, should the `dummy` fields be optional?

kyuksel (commented):
Looks good to me overall. Questions inline.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
rtitle: Good point.. I like having `ClusterInitValues` used for init, start, and shutdown scripts because it means each script has access to all templated variables. Totally agree about making the dummy fields optional.. will try and fix.

rtitle: Did this in latest commit: https://github.com/DataBiosphere/leonardo/pull/1170/commits/45b7f1231b093f8cd42377f24b9dfff28c8e0e51

kyuksel (approved):

PR #1172, opened 2019-12-09
[IA-1548] fixing creation timeout

jdcanas:
Testing in fiab currently, want the auto tests to run
---

In all cases:


rtitle (approved):
LGTM

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
```diff
@@ -47,6 +48,7 @@ class ClusterHelper(
   proxyConfig: ProxyConfig,
   clusterResourcesConfig: ClusterResourcesConfig,
   clusterFilesConfig: ClusterFilesConfig,
+  monitorConfig: MonitorConfig,
```
Qi77Qi, line 51: do we need the whole `MonitorConfig` for `ClusterHelper`? It looks like here it just need the `clusterCreationTimeout`?

jdcanas: it currently just needs that timeout, but I wanted to keep the ClusterHelper constructor consistent and pass the entire monitorconfig since the helper used heavily in back leo. There's not really much harm in passing a bigger object than needed since its not like it will be instantiated multiple times, and the clusterHelper constructor is used in so many places its a pain to modify

Qi77Qi: I think we should give it only what clusterHelper need. It makes it look like `ClusterHelper` cares about the other configs, so if you refactor code at some point, you may think the other fields will affect `ClusterHelper`, but in fact they don't. It also makes it confusing in tests since you'll need to stub about the whole `MachineConfig` object vs only what `ClusterHelper` cares.

jdcanas:
jenkins retest

PR #1173, opened 2019-12-16
[IA-1556] listClusters should reflect installed tool

rtitle:
https://broadworkbench.atlassian.net/browse/IA-1556

This PR does the following:
1. Sets a `tool` label on the cluster when it is created, which can either be `Jupyter` or `RStudio`. The label is returned in either `getCluster` or `listCluster` responses.
2. Ensures the `clusterUrl` field is correct in both `getCluster` and `listCluster` responses.

@zarsky-broad does the above work for you? (I'm working on the other issues you reported as well.) 

---

In all cases:


zarsky-broad:
Excellent.

akarukappadath:
Terra goes by '1 tool per cluster', but does AoU also follow this? Is this something we want to bake into Leo?

rtitle:
@akarukappadath that's a good question. I think it's ok to bake "1 tool per cluster" into Leo. There's been no product ask for multi-tools per cluster, and in some sense we've _already_ baked this into Leo with the `toolDockerImage` and `clusterUrl` parameters. I think we'll probably want to start supporting >1 tools per cluster when we do k8s, but that will involve a lot of other work too.

Qi77Qi (approved):
My only concern is it's not clear to me if clusterImage belong in listCluster. In general, it's not clear to me which fields belong in listCluster API vs getCluster...it seems we just add fields as demand arises, which doesn't seem desirable to me.

rtitle:
Thanks.. yeah we could use requirements about what belongs in listCluster or not. Currently my rule of thumb is that `listCluster` does not join to _any_ tables, except labels. But that's not a firm requirement.

Qi77Qi:
@rtitle I think that rule sounds good to me

PR #1174, opened 2019-12-16
[IA-1557] RStudio needs to set CSP in order to work in an iframe

rtitle:
https://broadworkbench.atlassian.net/browse/IA-1557

Previously we relied on Jupyter to set `Content-Security-Policy`, meaning we plumbed the value from Leo `reference.conf` to `jupyter_notebook_config.py` on the cluster. Unfortunately this approach is Jupyter-specific, and RStudio server doesn't seem to have an equivalent way to set CSP.

However, I can't think of a reason that the _tool_ needs to set it -- I think the Leo proxy could just set it instead. That's what I did in this PR. I verified in a local Terra that Jupyter and RStudio loads in an iframe.

image

@zarsky-broad fyi

---

In all cases:

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/Notebook.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/Notebook.scala
@@ -26,7 +26,7 @@ object Notebook extends RestClient with LazyLogging {
     mapper.readValue(response, classOf[NotebookContentItem])
 
   def notebooksBasePath(googleProject: GoogleProject, clusterName: ClusterName): String =
-    s"notebooks/${googleProject.value}/${clusterName.string}"
+    s"proxy/${googleProject.value}/${clusterName.string}/jupyter"
 
   def notebooksTreePath(googleProject: GoogleProject, clusterName: ClusterName): String =
     s"${notebooksBasePath(googleProject, clusterName)}/tree"
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
@@ -353,7 +353,6 @@ final case class ClusterTemplateValues private (googleProject: String,
                                                 jupyterStartUserScriptOutputBaseUri: String,
                                                 jupyterServiceAccountCredentials: String,
                                                 loginHint: String,
-                                                contentSecurityPolicy: String,
                                                 jupyterServerExtensions: String,
                                                 jupyterNbExtensions: String,
                                                 jupyterCombinedExtensions: String,
@@ -380,8 +379,7 @@ object ClusterTemplateValues {
             dataprocConfig: DataprocConfig,
             proxyConfig: ProxyConfig,
             clusterFilesConfig: ClusterFilesConfig,
-            clusterResourcesConfig: ClusterResourcesConfig,
-            contentSecurityPolicy: String): ClusterTemplateValues =
+            clusterResourcesConfig: ClusterResourcesConfig): ClusterTemplateValues =
     ClusterTemplateValues(
       cluster.googleProject.value,
       cluster.clusterName.value,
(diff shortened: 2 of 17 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
```diff
@@ -428,7 +426,8 @@ class ClusterHelper(
           clusterResourcesConfig.rstudioDockerCompose,
           clusterResourcesConfig.proxyDockerCompose,
           clusterResourcesConfig.proxySiteConf,
-          clusterResourcesConfig.welderDockerCompose
+          clusterResourcesConfig.welderDockerCompose,
+          clusterResourcesConfig.jupyterNotebookConfigUri
```
rtitle, line 430: `jupyterNotebookConfigUri` is a non-templated resource now. Technically we can/should probably put it in the base Jupyter image itself. However Leo still needs to upload it to support older image versions.

rtitle:
jenkins retest

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
```diff
@@ -353,7 +353,6 @@ final case class ClusterTemplateValues private (googleProject: String,
                                                 jupyterStartUserScriptOutputBaseUri: String,
                                                 jupyterServiceAccountCredentials: String,
                                                 loginHint: String,
-                                                contentSecurityPolicy: String,
```
rtitle, line 356: `contentSecurityPolicy` doesn't need to be a templated field anymore

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/api/ProxyRoutesSpec.scala
```diff
@@ -312,5 +312,6 @@ class ProxyRoutesSpec
       if (optionsRequest) Some(`Access-Control-Allow-Methods`(OPTIONS, POST, PUT, GET, DELETE, HEAD, PATCH))
       else None
     )
+    header("Content-Security-Policy") shouldBe Some(RawHeader("Content-Security-Policy", contentSecurityPolicy))
```
rtitle, line 315: This check gets run on every test case in `ProxyRoutesSpec`

jdcanas (approved):
LGTM

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/Notebook.scala
```diff
@@ -26,7 +26,7 @@ object Notebook extends RestClient with LazyLogging {
     mapper.readValue(response, classOf[NotebookContentItem])
 
   def notebooksBasePath(googleProject: GoogleProject, clusterName: ClusterName): String =
-    s"notebooks/${googleProject.value}/${clusterName.string}"
+    s"proxy/${googleProject.value}/${clusterName.string}/jupyter"
```
Qi77Qi, line 29: does this change how ui access notebook?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
Qi77Qi: add you add a comment about this in source code?

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/Notebook.scala
rtitle: UI is already using `/proxy` for everything as we recommend.. this is just making tests use the correct path.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
rtitle: Yeah will do in a separate PR

PR #1197, opened 2020-01-02
[no ticket] Fix 500 error for malformed proxy requests

rtitle:
Fix an issue @dmohs reported: we will now throw 404 instead of 500 for proxy requests where the cluster cannot be found.

---

In all cases:


Qi77Qi (approved):

Qi77Qi:
is there any swagger update needed?

rtitle:
Thanks.. Swagger already has:
```
        '404':
          description: Cluster not found
          schema:
            $ref: '#/definitions/ErrorReport'
```
for proxy requests so I think it's fine. Good catch though.

PR #1201, opened 2020-01-08
[IA-1576] [IA-1578] Recover gracefully from kernel out-of-memory errors

rtitle:
JIRAs:
https://broadworkbench.atlassian.net/browse/IA-1576
https://broadworkbench.atlassian.net/browse/IA-1578

Tech Doc:
https://docs.google.com/document/d/1KaZgfLxFIK9vLuFZ0uThgWqjQd0HJS9ACDdJv8xZ8xg/edit#

This PR corresponds to Solution A of the above tech doc, plus an automation test. We now dynamically determine a memory limit for the Jupyter container, and specify it in the docker-compose.yaml. The memory calculation is configurable. See comments inline.

---

In all cases:


automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/ClusterFixtureSpec.scala
```diff
@@ -63,7 +63,7 @@ abstract class ClusterFixtureSpec extends fixture.FreeSpec with BeforeAndAfterAl
       ronCluster = mockedCluster
     }
 
-    if(clusterCreationFailureMsg.nonEmpty)
+    if (clusterCreationFailureMsg.nonEmpty)
```
rtitle, line 66: `scalafmt` picked up some unrelated things in the `automation` project.

http/src/main/resources/jupyter/jupyter-docker-compose.yaml
```diff
@@ -37,3 +41,6 @@ services:
     env_file:
       - /etc/google_application_credentials.env
       - /etc/custom_env_vars.env
+    # See https://docs.docker.com/engine/reference/run/#user-memory-constraints
+    mem_limit: ${MEM_LIMIT}   # hard limit on memory consumption by the container
+    memswap_limit: -1         # no constraints on swap usage - use as much as allowed by the host
```
rtitle: We're not actually enabling swap on the VM yet, so this has no effect.

http/src/main/resources/reference.conf
```diff
@@ -17,39 +17,40 @@ dataproc {
...
-  welderEnabledNotebooksDir = "/home/jupyter-user/notebooks"
+  ]
   # Unset to use project-level defaults. Note a VPC/subnet with the literal name "default" may not exist.
   #vpcNetwork = "default"
   #vpcSubnet = "default"
   # This image is used for legacy jupyter image where hail is compatible with earlier version of spark
   legacyCustomDataprocImage = "projects/broad-dsp-gcr-public/global/images/custom-leo-image-1-2-79-debian9-5e0248a"
   customDataprocImage = "projects/broad-dsp-gcr-public/global/images/custom-leo-image-1-4-15-debian9-5e0248a"
 
+  dataprocReservedMemory = 4.5g
```
rtitle, line 28: Experimentally this seems to work for n1-standard-2, n1-standard-4, n1-standard-8 machine types.

http/src/main/resources/reference.conf
```diff
@@ -17,39 +17,40 @@ dataproc {
...
+image {
+  welderDockerImage = "us.gcr.io/broad-dsp-gcr-public/welder-server:60e28bc"
+  jupyterImage = "us.gcr.io/broad-dsp-gcr-public/leonardo-jupyter:5c51ce6935da"                  # updated 8/26/19
+
+  jupyterImageRegex = "us.gcr.io/broad-dsp-gcr-public/([a-z0-9-_]+):(.*)"
+  rstudioImageRegex = "us.gcr.io/anvil-gcr-public/([a-z0-9-_]+):(.*)"
+  dockerhubImageRegex = "broadinstitute/([a-z0-9-_]+):(.*)"
+}
+
+welder {
```
rtitle, line 40: I decided to break some things into a new `welder` structure (the diff looks a little odd).

http/src/main/resources/reference.conf
```diff
@@ -17,39 +17,40 @@ dataproc {
...
+  dataprocReservedMemory = 4.5g
+}
+
+image {
+  welderDockerImage = "us.gcr.io/broad-dsp-gcr-public/welder-server:60e28bc"
+  jupyterImage = "us.gcr.io/broad-dsp-gcr-public/leonardo-jupyter:5c51ce6935da"                  # updated 8/26/19
+
+  jupyterImageRegex = "us.gcr.io/broad-dsp-gcr-public/([a-z0-9-_]+):(.*)"
+  rstudioImageRegex = "us.gcr.io/anvil-gcr-public/([a-z0-9-_]+):(.*)"
+  dockerhubImageRegex = "broadinstitute/([a-z0-9-_]+):(.*)"
```
Qi77Qi: shall we name this to `broadDockerhubImageRegex`?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
```diff
@@ -362,7 +362,8 @@ final case class ClusterTemplateValues private (googleProject: String,
                                                 googleClientId: String,
                                                 welderEnabled: String,
                                                 notebooksDir: String,
-                                                customEnvVarsConfigUri: String) {
+                                                customEnvVarsConfigUri: String,
+                                                memLimit: String) {
```
Qi77Qi, line 366: shall we have a new type for `MemorySize` or something instead of using `String`?

rtitle:
fiab-start failed on something unrelated.

jenkins retest

http/src/main/resources/jupyter/jupyter-docker-compose.yaml
Qi77Qi: can you document unit for this mem_limit variable

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/WelderConfig.scala
```diff
@@ -0,0 +1,10 @@
+package org.broadinstitute.dsde.workbench.leonardo.config
+
+final case class WelderConfig(
+  welderEnabledNotebooksDir: String,
```
Qi77Qi: does it make sense to use `java.nio.file.Path`?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
```diff
@@ -83,6 +90,8 @@ class ClusterHelper(
           } else IO.pure(Map.empty[String, String])
           clusterVPCSettings = getClusterVPCSettings(projectLabels)
 
+          resourceConstraints <- getClusterResourceContraints(cluster)
```
Qi77Qi, line 93: if we fail to calculate constrain, we will fail cluster creation...maybe worth doing some perf test for this PR for the extra google calls?

http/src/main/resources/jupyter/jupyter-docker-compose.yaml
rtitle: Yeah… actually I may just remove it here because we're not enabling swap space anyway

http/src/main/resources/reference.conf
rtitle: This wasn't my addition, but yeah that makes sense. Will do.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/WelderConfig.scala
rtitle: 👍

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
rtitle: 👍

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
rtitle: It's true this adds Google calls to back Leo. Hopefully they are cheap because it's just looking up machine types, not mutating anything. But good idea about perf testing.

Another way to do this is to have a static Map for memory for all Google machine types… but I didn't like that because we'd need to remember to keep it up to date.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
rtitle: I think this field should be `String` by design because this is for templating values… this class is populated with the literal `String`s to be templated.

However I did make a `MemorySize` class and am using it instead of `Int` in other places.

rtitle:
@Qi77Qi I think I addressed PR feedback.. will run performance tests as you suggested.

Qi77Qi (approved):

http/src/main/resources/reference.conf
```diff
@@ -17,39 +17,40 @@ dataproc {
...
   #vpcSubnet = "default"
   # This image is used for legacy jupyter image where hail is compatible with earlier version of spark
   legacyCustomDataprocImage = "projects/broad-dsp-gcr-public/global/images/custom-leo-image-1-2-79-debian9-251b76c"
   customDataprocImage = "projects/broad-dsp-gcr-public/global/images/custom-leo-image-1-4-15-debian9-251b76c"
 
+  dataprocReservedMemory = 4.5g
+}
+
+image {
+  welderDockerImage = "us.gcr.io/broad-dsp-gcr-public/welder-server:60e28bc"
```
Qi77Qi, line 32: I just realize this refactoring seems to break the regex in https://github.com/broadinstitute/dsp-jenkins/blob/master/jobs/fc-jenkins/WelderUpdateHash.groovy#L32 ...it was probably broken from a previous PR...

rtitle: Oh, I didn't realize that existed

rtitle: I was testing the regex and it _seems_ to work -- the line `welderDockerImage = …` is unchanged, it's just in a different structure.

rtitle:
I ran the 200 user createCluster test against this PR, two times.
1. 200/200 successful cluster creations. `listCluster` had some Sam errors, likely due to known Sam issues on perf.
2. 200/200 successful cluster creations. No `listCluster` errors.

In either case I didn't see any issues with the new Google APIs, so I think this PR is safe. Will merge and monitor the nightly runs.

PR #1202, opened 2020-01-09
[no ticket] Fix generated `clusterUrl` for getCluster response

rtitle:
I noticed that `clusterUrl` contained the wrong URL (e.g. `https://notebooks.firecloud.org/proxy/broad-firecloud-dsde/rt-rstudio-test/jupyter`) for an RStudio image. This fixes it and adds a unit test.

Also ran scalafmt which picked up some unrelated changes (in a separate commit).

---

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
@@ -13,7 +13,7 @@ import enumeratum.{Enum, EnumEntry}
 import org.broadinstitute.dsde.workbench.leonardo.config._
 import org.broadinstitute.dsde.workbench.leonardo.model.Cluster._
 import org.broadinstitute.dsde.workbench.leonardo.model.ClusterContainerServiceType.JupyterService
-import org.broadinstitute.dsde.workbench.leonardo.model.ClusterImageType.{Jupyter, RStudio, Welder}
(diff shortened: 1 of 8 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
```diff
@@ -245,7 +245,7 @@ object Cluster {
                     labels: Map[String, String]): URL = {
     val tool = clusterImages
       .map(_.imageType)
-      .filterNot(_ == Welder)
+      .filterNot(Set(Welder, CustomDataProc).contains)
```
rtitle, line 248: This is the actual fix

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModelSpec.scala
```diff
@@ -249,4 +249,38 @@ class LeonardoModelSpec extends TestComponent with FlatSpecLike with Matchers wi
     ))
     ContainerImage.stringToJupyterDockerImage("asd/asdf") shouldBe (Some(ContainerImage.DockerHub("asd/asdf")))
   }
+
+  "Cluster" should "generate a correct cluster URL" in {
```
rtitle, line 253: This is the unit test for the fix.

Qi77Qi (approved):

PR #1226, opened 2020-01-17
[IA-1623] Fix ClusterToolMonitor spam

rtitle:
https://broadworkbench.atlassian.net/browse/IA-1623

Basically updated `ClusterToolMonitor` to query the CLUSTER_IMAGE table and check status only for the enabled tools.

Unit tests pass, haven't tested in a fiab yet.

---

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterToolMonitor.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterToolMonitor.scala
@@ -10,10 +10,8 @@ import org.broadinstitute.dsde.workbench.leonardo.config.ClusterToolConfig
 import org.broadinstitute.dsde.workbench.leonardo.dao.ToolDAO
 import org.broadinstitute.dsde.workbench.leonardo.dao.google.GoogleDataprocDAO
 import org.broadinstitute.dsde.workbench.leonardo.db.{clusterQuery, DbReference}
-import org.broadinstitute.dsde.workbench.leonardo.model.google.ClusterName
-import org.broadinstitute.dsde.workbench.leonardo.model.ClusterContainerServiceType
+import org.broadinstitute.dsde.workbench.leonardo.model.{ClusterContainerServiceType, RunningCluster}
 import org.broadinstitute.dsde.workbench.leonardo.monitor.ClusterToolMonitor._
-import org.broadinstitute.dsde.workbench.model.google.GoogleProject
 import org.broadinstitute.dsde.workbench.newrelic.NewRelicMetrics
 
 import scala.concurrent.ExecutionContext
@@ -95,7 +93,3 @@ class ClusterToolMonitor(
         }
     }
 }
-
-final case class RunningCluster(googleProject: GoogleProject,
-                                clusterName: ClusterName,
-                                containers: List[ClusterContainerServiceType])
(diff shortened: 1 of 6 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterToolMonitor.scala
```diff
@@ -88,21 +87,15 @@ class ClusterToolMonitor(
...
+        .isProxyAvailable(cluster.googleProject, cluster.clusterName)
+        .map { status =>
+          ToolStatus(status, tool, cluster)
+        }
     }
-
 }
 
-final case class RunningCluster(googleProject: GoogleProject, clusterName: ClusterName, welderEnabled: Boolean)
+final case class RunningCluster(googleProject: GoogleProject,
```
rtitle: I was wondering if we want to move `RunningCluster` to `LeonardoModel` or something, since it's used outside of this class

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterToolMonitorSpec.scala
```diff
@@ -42,34 +44,65 @@ class ClusterToolMonitorSpec
...
     welderEnabledCluster.save()
+    welderDisabledCluster.save()
+    notRunningCluster.save()
+    rstudioCluster.save()
 
     withServiceActor() { (_, mockNewRelic) =>
       Thread.sleep(clusterToolConfig.pollPeriod.toMillis * 3)
-      verify(mockNewRelic, never()).incrementCounterFuture(ArgumentMatchers.startsWith("JupyterDown"),
-                                                           ArgumentMatchers.anyInt())(ArgumentMatchers.any())
-      verify(mockNewRelic, never()).incrementCounterFuture(ArgumentMatchers.startsWith("WelderDown"),
```
rtitle, line 56: These tests had some problems… for one it was verifying `incrementCounterFuture`, which is not actually called by `ClusterToolMonitor`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterToolMonitor.scala
Qi77Qi: maybe to https://github.com/DataBiosphere/leonardo/blob/develop/core/src/main/scala/org.broadinstitute.dsde.workbench.leonardo/models.scala?

rtitle: Yeah that sounds good

Qi77Qi (approved):

Qi77Qi:
thanks for working on this! 👍

Qi77Qi:
oh can you also revert https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/Boot.scala#L194? we can use the same client for all tools with your PR

rtitle:
yes! thanks for reminder

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterToolMonitor.scala
rtitle: Actually that would require pulling more things into `models.scala` unrelated to this change. Putting in `LeonardoModel` for now.. might be good to have a tech debt ticket to populate models.scala.

PR #1233, opened 2020-01-23
[IA-1512] Patch endpoint allowStop enhancement

jdcanas:
Patch endpoint allowStop feature is in a reviewable state, DB changes and pubsub implemented, with all tests.

See this diagram for the system interactions implemented: https://www.draw.io/#G1hZK-hLNV-nXQPQvKXf3OLlS8sRUTd4JY

See this document for high-level design details: https://docs.google.com/document/d/1EAkTfXEZ2uzbK46kftbJcAwB3VRSYfof2OovHYRs484/edit


automation/hotswap.sh
```diff
@@ -5,17 +5,20 @@
 
 set -eu
 
-if [ -z "$1" ]
-    then
-        echo "No arguments supplied. Please provide FIAB name as an argument."
-        exit 1
+# see https://stackoverflow.com/questions/3601515/how-to-check-if-a-variable-is-set-in-bash
+if [ -z ${1+x} ]; then
```
jdcanas, line 9: this if was not working

automation/hotswap.sh
```diff
@@ -5,17 +5,20 @@
...
 fi
 
 FIAB=$1
 
-printf "Generating the Leo jar...\n\n"
+printf "Generating the Leo jar. This will fail silently here if you have any compilation errors, or print a success message otherwise\n\n"
 # Example: /Users/qi/workspace/leonardo/http/target/scala-2.12/http-assembly-0.1-437ee4a9-SNAPSHOT.jar
 LEO_JAR_PATH=$(sbt -Dsbt.log.noformat=true "project http" assembly | tail -3 | head -2 | grep -o '/Users[^ ]*')
+
+printf "Jar successfully generated."
```
jdcanas, line 20: the build could fail silently before

automation/src/test/resources/reference.conf
```diff
@@ -8,3 +8,5 @@ leonardo.currentWelderImage = "us.gcr.io/broad-dsp-gcr-public/welder-server:ed93
 leonardo.bioconductorImageUrl = "us.gcr.io/broad-dsp-gcr-public/terra-jupyter-bioconductor:0.0.9"
 leonardo.rstudioBaseImageUrl = "us.gcr.io/anvil-gcr-public/anvil-rstudio-base:0.0.2"
 leonardo.rstudioBioconductorImageUrl = "us.gcr.io/anvil-gcr-public/anvil-rstudio-bioconductor:0.0.3"
+leonardo.pubsubGoogleProject = "broad-dsde-dev"
+leonardo.topicName = "leonardo-pubsub-test" //each fiab will have a unique topic name sourced from leonardo.conf, this is largely irrelevant but needed to initializie some objects
```
jdcanas: this comment is very important to note - all fiabs will have a unique topic based on the timestamp at which they are created, to prevent interference. The tests will also clean-up the topics after themselves.

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeoPubsubSpec.scala
```diff
@@ -0,0 +1,73 @@
...
+
+      eventually(timeout(Span(3, Minutes))) {
+        val getCluster: Cluster = Leonardo.cluster.get(clusterFixture.cluster.googleProject, clusterFixture.cluster.clusterName)
+        getCluster.status shouldBe ClusterStatus.Starting
+        getCluster.machineConfig shouldBe machineConfig
+      }
+    }
+
+  override def afterAll(): Unit = {
+    cleanupTopic().unsafeRunSync()
```
jdcanas: the topic is created automatically by our google2 pub/sub implementation, so we just need to clean-up

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoModelCopy.scala
```diff
@@ -27,22 +27,7 @@ case class Cluster(clusterName: ClusterName,
...
-                          defaultClientId: Option[String] = None,
-                          toolDockerImage: Option[String] = None,
-                          welderDockerImage: Option[String] = None,
-                          scopes: Set[String] = Set.empty,
-                          enableWelder: Option[Boolean] = None,
-                          customClusterEnvironmentVariables: Map[String, String] = Map.empty)
+case class ClusterRequest(labels: LabelMap = Map(), jupyterExtensionUri: Option[String] = None, jupyterUserScriptUri: Option[String] = None, jupyterStartUserScriptUri: Option[String] = None, machineConfig: Option[MachineConfig] = None, properties: Map[String, String] = Map(), stopAfterCreation: Option[Boolean] = None, userJupyterExtensionConfig: Option[UserJupyterExtensionConfig] = None, autopause: Option[Boolean] = None, autopauseThreshold: Option[Int] = None, defaultClientId: Option[String] = None, toolDockerImage: Option[String] = None, welderDockerImage: Option[String] = None, scopes: Set[String] = Set.empty, enableWelder: Option[Boolean] = None, customClusterEnvironmentVariables: Map[String, String] = Map.empty, allowStop: Option[Boolean] = Some(false))
```
jdcanas: i am but a slave to scalafmt at this point...

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/FollowupComponent.scala
```diff
@@ -0,0 +1,44 @@
...
+object followupQuery extends TableQuery(new FollowupTable(_)) {
+  def save(followupDetails: ClusterFollowupDetails, masterMachineType: Option[String]): DBIO[Int] =
+    followupQuery += FollowupRecord(followupDetails.clusterId,
+                                    followupDetails.clusterStatus.entryName,
+                                    masterMachineType)
+
+  def delete(followupDetails: ClusterFollowupDetails): DBIO[Int] =
+    baseFollowupQuery(followupDetails).delete
+
+  //TODO: should this differentiate itself between no cluster with that ID found, and no machine config found?
```
jdcanas: I plan to clean this up before the PR is merged, but this TODO is more of a question to reviewers than myself. I think the answer is no, because to the pubsub reader which retrieves this, it makes no difference why there is no follow-up details, both result in the same no-op.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -0,0 +1,205 @@
...
+    )
+
+    response
+  }
+
+  def messageHandler: Pipe[IO, Event[LeoPubsubMessage], Unit] = in => {
+    in.flatMap { event =>
+      for {
+        _ <- Stream
+          .eval(IO.pure(event.consumer.ack())) //we always ack first, as it could cause an endless loop of exceptions to do it after
```
jdcanas: I decide to ack first here, otherwise Leo could concievably get stuck in a state where it keeps trying to process the same message and throwing errors. If it somehow gets into an undesirable state, I think it is better to simply flush the message out of the system, as manual intervention would be required otherwise.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -317,6 +326,22 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
           clusterQuery.updateClusterStatus(existingCluster.id, ClusterStatus.Updating, now).transaction.void
         } else IO.unit
 
+        //maybeMasterMachineTypeChanged will throw an error if the appropriate flag hasn't been set to allow special update transitions
+        shouldFollowup = masterMachineTypeChanged.map(result => result.shouldFollowup).getOrElse(false)
+        _ <- if (shouldFollowup) {
+          //we do not support follow-up transitions when the cluster is set to an updating status
+          if (!shouldUpdate.combineAll) {
```
jdcanas, line 337: this is nasty, but necessary afaik. It is needed because, if the cluster is resized (aka `shouldUpdate.combineAll` is true), then this function sets the cluster status to updating while it processes.

Instead of further complicating this complicated transition, we choose to only respect the resize, and ignore the request to update the machineType. The idea is that this shouldn't be a common use case in the wild when you slice into what users are doing. A metric is included in the else clause to track if it is more common than we think.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -529,29 +575,9 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
       //if you've got to here you at least have GetClusterDetails permissions so a 403 is appropriate if you can't actually stop it
       _ <- checkClusterPermission(userInfo, StopStartCluster, cluster, throw403 = true)
 
-      _ <- internalStopCluster(cluster)
+      _ <- clusterHelper.stopCluster(cluster)
     } yield ()
 
-  def internalStopCluster(cluster: Cluster): IO[Unit] =
```
jdcanas, line 534: I moved a lot of functions to the clusterHelper as they are needed in backLeo now too.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -563,31 +589,9 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
       _ <- checkClusterPermission(userInfo, StopStartCluster, cluster, throw403 = true)
 
       now <- timer.clock.realTime(TimeUnit.MILLISECONDS)
-      _ <- internalStartCluster(userInfo.userEmail, cluster, Instant.ofEpochMilli(now))
```
jdcanas, line 565: this email is never used in downstream functions, removing it from the method signature

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -563,31 +589,9 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
...
+      _ <- clusterHelper.internalStartCluster(cluster)
     } yield ()
 
-  def internalStartCluster(userEmail: WorkbenchEmail, cluster: Cluster, now: Instant): IO[Unit] =
```
jdcanas, line 568: moved to ClusterHelper

automation/src/test/resources/reference.conf
Qi77Qi: What's the error if this doesn't exist?

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoModelCopy.scala
```diff
@@ -27,22 +27,7 @@ case class Cluster(clusterName: ClusterName,
...
```
Qi77Qi: make sure to run scalafmt before merging

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/Boot.scala
```diff
@@ -55,6 +60,8 @@ object Boot extends IOApp with LazyLogging {
     implicit val materializer = ActorMaterializer()
     import system.dispatcher
 
+    logger.info("beginning startUp function")
```
Qi77Qi: can we remove this log

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/PubsubConfig.scala
```diff
@@ -0,0 +1,9 @@
+package org.broadinstitute.dsde.workbench.leonardo.config
+
+import org.broadinstitute.dsde.workbench.model.google.GoogleProject
+
+final case class PubsubConfig(
```
Qi77Qi, line 5: think I had a previous comment about removing this datatype and just use the `PublisherConfig` directly

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/FollowupComponent.scala
```diff
@@ -0,0 +1,44 @@
...
+import org.broadinstitute.dsde.workbench.leonardo.monitor.ClusterFollowupDetails
+
+import scala.concurrent.ExecutionContext
+import LeoProfile.api._
+
+case class FollowupRecord(clusterId: Long, status: String, masterMachineType: Option[String])
+
+class FollowupTable(tag: Tag) extends Table[FollowupRecord](tag, "CLUSTER_FOLLOWUP") {
+  def clusterId = column[Long]("clusterId", O.PrimaryKey)
+  def status = column[String]("status", O.Length(254))
```
Qi77Qi: can we do `column[ClusterStatus]` here, and then in `mappedColumnImplicits`, and something similar to 
```
    implicit val gsPathMappedColumnType: BaseColumnType[GcsPath] =
      MappedColumnType
        .base[GcsPath, String](_.toUri, s => parseGcsPath(s).fold(e => throw new Exception(e.toString()), identity))
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/FollowupComponent.scala
```diff
@@ -0,0 +1,44 @@
...
+
+import org.broadinstitute.dsde.workbench.leonardo.monitor.ClusterFollowupDetails
+
+import scala.concurrent.ExecutionContext
+import LeoProfile.api._
+
+case class FollowupRecord(clusterId: Long, status: String, masterMachineType: Option[String])
+
+class FollowupTable(tag: Tag) extends Table[FollowupRecord](tag, "CLUSTER_FOLLOWUP") {
+  def clusterId = column[Long]("clusterId", O.PrimaryKey)
```
Qi77Qi, line 13: which column this maps to in `CLUSTER`?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -0,0 +1,205 @@
...
+      case msg @ StopUpdateMessage(_, _) =>
+        handleStopUpdateMessage(msg)
+      case msg @ ClusterTransitionFinishedMessage(_) =>
+        handleClusterTransitionFinished(msg)
+      case _ => IO.unit
+    }
+
+    //ensure we don't crash if the handler throws an exception
+    response.handleErrorWith(
+      e =>
```
Qi77Qi: I would actually do `onError` here.
```
response.onError {
  case e => IO(logger.error(xxx))
}
```

And get rid of `response` in line 52

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
Qi77Qi: is this still true now that we're recovering from the errors? for some intermittent google errors, we probably should replay the messages right?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -0,0 +1,205 @@
...
+        _ <- Stream
+          .eval(IO.pure(event.consumer.ack())) //we always ack first, as it could cause an endless loop of exceptions to do it after
+        _ <- Stream.eval(messageResponder(event.msg))
+      } yield ()
+    }
+  }
+
+  val process: Stream[IO, Unit] = (subscriber.messages through messageHandler).repeat
+
+  def handleStopUpdateMessage(message: StopUpdateMessage): IO[Unit] =
```
Qi77Qi: I think all methods other than `process` can be marked as `private`?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -0,0 +1,205 @@
...
+      case _ => IO.unit
+    }
+
+}
+
+sealed trait LeoPubsubMessage {
+  def messageType: String
+}
+
+final case class StopUpdateMessage(updatedMachineConfig: MachineConfig, clusterId: Long) extends LeoPubsubMessage {
```
Qi77Qi: can you move those sub classes of `LeoPubsubMessage` into an object `LeoPubsubMessage`?

like 
```
object LeoPubsubMessage {
 ...
}

File: http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala [hunk 20]
```diff
@@ -0,0 +1,205 @@
...
+      messageType <- message.downField("messageType").as[String]
+      value <- messageType match {
+        case "stopUpdate"         => message.as[StopUpdateMessage]
+        case "transitionFinished" => message.as[ClusterTransitionFinishedMessage]
+        case other                => Left(DecodingFailure(s"found a message with an unknown type when decoding: ${other}", List.empty))
+      }
+    } yield value
+  }
+
+  implicit val machineConfigEncoder: Encoder[MachineConfig] =
```
Qi77Qi: can you move this encoder to `org.broadinstitute.dsde.workbench.leonardo.JsonCodec`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -119,26 +118,36 @@ case class ImageAutoDetectionException(traceId: TraceId, image: ContainerImage)
...
-                                                              implicit override val system: ActorSystem,
-                                                              log: Logger[IO],
-                                                              cs: ContextShift[IO],
-                                                              metrics: NewRelicMetrics[IO],
-                                                              dbRef: DbReference[IO],
-                                                              timer: Timer[IO])
+case class UpdateResult(hasUpdateSucceded: Boolean, shouldFollowup: Boolean, followupAction: UpdateTransition)
+
+sealed trait UpdateTransition
+case class StopStartTransition(updateConfig: MachineConfig) extends UpdateTransition
```
Qi77Qi: move all sub classes to an object `UpdateTransition`...that way it's clearer what all the sub classes are

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/PubsubConfig.scala
jdcanas: yeah, I was going to, but i need the queueSize, which isn't in the PublisherConfig

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/FollowupComponent.scala
jdcanas: this maps directly to the id field in the cluster table via the foreign key constraint on line 16

automation/src/test/resources/reference.conf
jdcanas: I updated the comment to be more clear, this topic is used in an auto test to ensure the subscriber can appropriately connect, but it is never published/read from. The unique topic (not this) is the one used in the Leonardo instance the tests are run against. 

So an error will be thrown if it is not present.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
jdcanas: Hmm.... I'm having trouble thinking of a scenario in which the order matters now, even if there are intermittent google errors, we would still ack as we handle these errors in the messageResponder

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/FollowupComponent.scala
jdcanas: I did this ` MappedColumnType .base[ClusterStatus, String](_.entryName, s => ClusterStatus.withName(s))`. Not sure if I need to make a helper function to lift it to an either, or if the fact withName throws an exception is sufficient?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
jdcanas: I'm not sure what you mean, onError isn't a method on either IO or fs2.Stream

Qi77Qi: oh..it'll be there once you do `import cats.implicits._`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/FollowupComponent.scala
Qi77Qi: yea, here `ClusterStatus.withName(s))` would just throw an exception

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
jdcanas: I did it for all but `messageResponder`, which is used heavily in the unit tests. It seems mockito is incapable of mocking an fs2 Stream via

```
queue.enqueue1(Event(message.asInstanceOf[LeoPubsubMessage], mock[AckReplyConsumer]))
    when {
      googleSubscriber.messages
    } thenReturn queue.dequeue
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -317,6 +331,24 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
...
+        shouldFollowup = masterMachineTypeChanged.map(result => result.shouldFollowup).getOrElse(false)
+        _ <- if (shouldFollowup) {
+          //we do not support follow-up transitions when the cluster is set to an updating status
+          if (!shouldUpdate.combineAll) {
+            val action = masterMachineTypeChanged.map(result => result.followupAction).getOrElse(Noop(None))
+            logger.info(s"detected follow-up action necessary: ${action}")
+            handleClusterTransition(existingCluster, action)
+          } else {
+            logger.warn(
+              "A user tried to resize the cluster and master machine type at the same time, which is not supported. Only the cluster will be resized."
```
jdcanas: I'm beginning to think this should be an exception, as it could be really confusing for a user if we implement  option #2 below. Thoughts @sharmatime @Qi77Qi @rtitle ? 

To summarize, if the user tries to update either `numberOfPreemptibleWorkers` or `numberOfWorkers` in conjunction with `masterMachineType`, we disallow it because it causes complications. There are a few options in this case. I'm leading towards #1, even though #2 is what we originally discussed in person. 

    1) disallow this in the UI and throw an exception if this happens
    2) provide a warning in the UI, and only respect the change in `numberOfPreemptibleWorkers` or `numberOfWorkers`, ignoring the change in machine type.

automation/hotswap.sh
rtitle: TIL

automation/src/test/resources/reference.conf
```diff
@@ -8,3 +8,6 @@ leonardo.currentWelderImage = "us.gcr.io/broad-dsp-gcr-public/welder-server:ed93
 leonardo.bioconductorImageUrl = "us.gcr.io/broad-dsp-gcr-public/terra-jupyter-bioconductor:0.0.9"
 leonardo.rstudioBaseImageUrl = "us.gcr.io/anvil-gcr-public/anvil-rstudio-base:0.0.2"
 leonardo.rstudioBioconductorImageUrl = "us.gcr.io/anvil-gcr-public/anvil-rstudio-bioconductor:0.0.3"
+leonardo.pubsubGoogleProject = "broad-dsde-dev"
```
rtitle: Does this work for tests run in broad-dsde-qa? I wonder if these pubsub configs should live here: https://github.com/broadinstitute/firecloud-automated-testing/blob/master/configs/leonardo/application.conf.ctmpl

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
Qi77Qi: Would be nice to have the validation in UI.

but I think in API, we should also notify user. Not "throwing" an exception though, just do `IO.raiseError(new someBadRequestException)`

automation/src/test/resources/reference.conf
Qi77Qi: currently the pubsub topic is only created in broad-dsde-dev via terraform. All other environments need to be run with tf change as well... @jdcanas I think we can do that now since things are working now?

http/src/main/resources/reference.conf
```diff
@@ -149,6 +149,12 @@ serviceAccounts {
   }
 }
 
+pubsub {
+  pubsubGoogleProject = "broad-dsde-dev"
```
rtitle: Not sure we should have `broad-dsde-dev` in Leo reference.conf.. this should probably just be in firecloud-develop.

jdcanas: Oops,  it is so I think I can remove this, it was just needed to run leo locally before that work was done

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/DataprocConfig.scala
```diff
@@ -7,6 +7,7 @@ final case class DataprocConfig(
   applicationName: String,
   dataprocDefaultRegion: String,
   dataprocZone: Option[String],
+  //TODO: refactor a common config with "realm" in swagger config and the field in pubsub config
```
rtitle, line 10: What field(s) is this comment referring to?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/FollowupComponent.scala
rtitle: I think returning an `Option` is fine, we don't need to differentiate… and the database should maintain referential integrity for the foreign key, so there should always be a corresponding CLUSTER defined.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -0,0 +1,194 @@
...
+import org.broadinstitute.dsde.workbench.model.WorkbenchException
+import cats.implicits._
+
+import scala.concurrent.ExecutionContext
+
+case class ClusterNotFoundException(
+  clusterId: Long,
+  override val message: String =
+    "Could not process ClusterTransitionFinishedMessage because it was not found in the database"
+) extends LeoException
```
rtitle: When you extend `LeoException` you should provide a status otherwise it'll default to 500. For example: https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala#L51

rtitle: …although these errors are not returned directly to users, right? How do errors manifest? I'll keep reviewing & try to figure it out :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -0,0 +1,194 @@
...
+      case msg @ StopUpdateMessage(_, _) =>
+        handleStopUpdateMessage(msg)
+      case msg @ ClusterTransitionFinishedMessage(_) =>
+        handleClusterTransitionFinished(msg)
+      case _ => IO.unit
+    }
+
+    //ensure we don't crash if the handler throws an exception
+    response.onError {
+      case e => IO(logger.error(s"Unable to process a message in received from pub/sub subscription in messageResponder: ${message}, errorMessage: ${e.getMessage}", e))
```
rtitle: So it seems like errors are just logged, correct? I wonder if we should do any other action, like put the cluster in Error state, or maybe have a timeout for the entire stop-update transition. I'm not really sure yet, will think about it..

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -0,0 +1,194 @@
...
+        }
+        case Some(resolvedCluster) =>
+          IO.raiseError(
+            new WorkbenchException(
+              s"Failed to process StopUpdateMessage for Cluster ${resolvedCluster.projectNameString}. This is likely due to a mismatch in state between the db and the message, or an improperly formatted machineConfig in the message. Cluster details: ${resolvedCluster}"
+            )
+          )
+        case None =>
+          IO.raiseError(
+            new WorkbenchException(
```
rtitle, line 89: Minor comment, but I see you use `ClusterNotFoundException` in another place for this, but `WorkbenchException` here. Would be good to be consistent, probably just using `WorkbenchException` is fine.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -0,0 +1,194 @@
...
+      .inTransaction { clusterQuery.getClusterById(message.clusterId) }
+      .flatMap {
+        case Some(resolvedCluster)
+            if ClusterStatus.stoppableStatuses.contains(resolvedCluster.status) && !message.updatedMachineConfig.masterMachineType.isEmpty => {
+          val followupDetails = ClusterFollowupDetails(message.clusterId, ClusterStatus.Stopped)
+          logger.info(
+            s"stopping cluster ${resolvedCluster.projectNameString} in messageResponder, and saving a record for ${resolvedCluster.id}"
+          )
+          for {
+            _ <- dbRef.inTransaction(
```
rtitle, line 75: What if there is already a followup record in the DB for this cluster, does this fail?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
rtitle: Hm… I tend to think that returning an error (409 Conflict?) is cleaner than silently ignoring the request to change the machine type.

I do agree that this is not expected to be a common transition, so not worth the complexity to unify the `Updating` status with the stop-start transition.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -327,6 +359,21 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
     } else IO.raiseError(ClusterCannotBeUpdatedException(existingCluster))
   }
 
+  private def handleClusterTransition(existingCluster: Cluster, transition: UpdateTransition): IO[Unit] =
+    transition match {
+      case StopStartTransition(machineConfig) =>
+        for {
+          _ <- metrics.incrementCounter(s"pubsub/LeonardoService/StopStartTransition ")
```
rtitle: Any reason for the trailing space? :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -405,43 +452,49 @@ class LeonardoService(protected val dataprocConfig: DataprocConfig,
...
-        } yield true
-
-      case Some(_) =>
-        IO.raiseError(ClusterMachineTypeCannotBeChangedException(existingCluster))
+          _ <- clusterHelper.updateMasterMachineType(existingCluster, MachineType(updatedMasterMachineType))
+        } yield UpdateResult(true, false, Noop(None))
+
+      case Some(updatedMasterMachineType) =>
+        logger.info("in stop and update case of maybeChangeMasterMachineType")
+        //        Future.failed(ClusterMachineTypeCannotBeChangedException(existingCluster))
```
rtitle: Remove commented line.
Also should some of these `info` statements be `debug`?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -0,0 +1,194 @@
...
+      }
+
+      //No actions for other statuses yet. There is some logic that will be needed for all other cases (i.e. the 'None' case where no cluster is found in the db and possibly the case that checks for the data in the DB)
+      // TODO: Refactor once there is more than one case
+      case _ => IO.unit
+    }
+
+}
+
+sealed trait LeoPubsubMessage {
```
rtitle, line 139: A couple questions about the data model here:
- should the base trait have `clusterId` and `messageType`?
- it's not clear to me why `ClusterFollowupDetails` needs a `ClusterStatus`, since the status is on the cluster record itself

My proposal in code:
```
sealed trait LeoPubsubMessage {
  def messageType: String
  def clusterId: Long
}
object LeoPubsubMessage {
  final case class StopUpdateMessage(updatedMachineConfig: MachineConfig, clusterId: Long) extends LeoPubsubMessage {
    val messageType = "stopUpdate"
  }
  final case class ClusterTransitionFinishedMessage(clusterId: Long) extends LeoPubsubMessage {
    val messageType = "transitionFinished"
  }
}
```

Let me know what you think..

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/FollowupComponent.scala
```diff
@@ -0,0 +1,43 @@
...
+import LeoProfile.api._
+import LeoProfile.mappedColumnImplicits.statusMappedColumnType
+import org.broadinstitute.dsde.workbench.leonardo.model.google.ClusterStatus
+import org.broadinstitute.dsde.workbench.leonardo.monitor.LeoPubsubMessage.ClusterFollowupDetails
+
+case class FollowupRecord(clusterId: Long, status: ClusterStatus, masterMachineType: Option[String])
+
+class FollowupTable(tag: Tag) extends Table[FollowupRecord](tag, "CLUSTER_FOLLOWUP") {
+  def clusterId = column[Long]("clusterId", O.PrimaryKey)
+  def status = column[ClusterStatus]("status", O.Length(254))
```
rtitle, line 14: I asked this in another place, but does this table need a `status` column? Seems it's duplicating `CLUSTER.status`, to which it has a 1-1 relationship.

rtitle:
I made some comments inline, but in general this looks good! Thanks for including the design doc and diagram. The queue implementation looks good, and I'm excited to start using it for other functionality as well.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
rtitle: I'm wondering if we need to delete the record from the followup table in case any error occurs, maybe in `response.onError` in `messageResponder`. Otherwise 1 error might cause DB conflicts for subsequent updates to the same cluster.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/DataprocConfig.scala
jdcanas: the leoGoogleProject field. We have googleProject being read into 3 different config objects currently.

Search `"{{$googleProj}}"` here to confirm it is indeed the same exact value, templated in 3 different times in 3 different places.

https://github.com/broadinstitute/firecloud-develop/blob/dev/base-configs/leonardo/leonardo.conf.ctmpl

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
jdcanas: they shouldn't be presented to the user, they are printed in line 46 of the messageResponder via

```
response.onError {
      case e => IO(logger.error(s"Unable to process a message in received from pub/sub subscription in messageResponder: ${message}, errorMessage: ${e.getMessage}", e))
    }
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
jdcanas: As of right now, I think logging is the appropriate response, as the cluster won't be in a bad state and we will acknowledge the message and flush it from the system. From the users perspective, if this happens, their cluster will not go through the desired transition, but it should still display the correct state in the UI throughout. It could be confusing as their update won't finish, but it should recover naturally.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
jdcanas: yes, it will, since its tied to the primary key the DB will enforce 1 entry in the follow-up table per cluster, added on-error clean-up logic.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
jdcanas: fixed them up and added more helpful cluster info to them. I think it will be useful for the initial phases to have as much logging as possible in case anything goes wrong

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
jdcanas: My thought behind this was, The transition finished message needs the status due to the asynchronous nature of the message queue and the need to ensure that the status in the DB matches the status in the message before proceeding with the action. 

In other words, if it was a simple 'transition done' message, the user could quickly delete the cluster, or some other unexpected status change could happen (it errors somehow), causing a failure without a clear cause in the log. 

It is true that the check is a bit duplicative with the methods that are called though. I like the idea of simplifying here as I think it may be a bit over-engineered

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/FollowupComponent.scala
jdcanas: this saved status is needed, as otherwise subscriber has no idea at what cluster status it should perform the associated follow-up action. This field isn't reflective of the true cluster status, rather a static status. When the cluster status equals the status in this table, the record should be removed.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/DataprocConfig.scala
rtitle: Ah I see, yeah

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/FollowupComponent.scala
rtitle: I see - I didn't realize that nuance, I think that makes sense then

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
rtitle: That makes sense.. the message is published from the ClusterMonitorActor when the cluster is Stopped, but something could have happened in the meantime

rtitle:
I think all my PR comments are addressed. This is nearly ready to go, but I think it should be rebased first… specifically I don't think this branch has commit https://github.com/DataBiosphere/leonardo/commit/be0822085833bbc11b43fdda473ab461fc465480, so the automation tests are not actually running against it.

Once it's rebased, automation tests pass, and the terraform+manual IAM steps are done in all envs, I say :shipit:

automation/src/test/resources/reference.conf
jdcanas: Might be worth jumping on a quick call for this. I'm working on the terraform subtask. As I understand it though, the tests in qa run in a fiab, which will get their topic name from https://github.com/broadinstitute/firecloud-develop/blob/dev/base-configs/leonardo/leonardo.conf.ctmpl#L88 , and thus the only thing that needs to be done is add the appropriate role to the qa svc account (which we've done). If the service account has the appropriate permissions, our google2 pubsub impl automatically creates whatever topic you tell it to use. 

I do think I need to move leonardo.pubsubGoogleProject to the config you mention though rob, based on this understanding. However, it should be sufficient to keep a static project here, as the auto test code only write to pubsub, never read to avoid conflicts. The topic in leonardo (which is randomly generated from the linked config above) is the one that goes full cycle and is unique.

rtitle: My point was that the pubSub project might be `broad-dsde-dev` or `broad-dsde-qa` for automation tests -- so the project configured here needs to match the project configured in Leo. So I don't think we can have a static project in this automation/reference.conf, we should template it in https://github.com/broadinstitute/firecloud-automated-testing/blob/master/configs/leonardo/application.conf.ctmpl.

So the google2 impl creates the topic if it doesn't exist? Do we need to terraform it then?

rtitle (changes requested):
Requesting changes just to rebase this PR, and make sure automation tests pass.

automation/src/test/resources/reference.conf
rtitle: Happy to jump on a call too, i could be misunderstanding something.

Qi77Qi: > My point was that the pubSub project might be broad-dsde-dev or broad-dsde-qa for automation tests -- so the project configured here needs to match the project configured in Leo. 

Agree...

> So the google2 impl creates the topic if it doesn't exist? Do we need to terraform it then?

I think terraform change is still needed since it sets up some permission https://github.com/broadinstitute/terraform-terra/pull/111/files#diff-145fc03d6d7602a93c883b635fb3b01eR9

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeoPubsubSpec.scala
```diff
@@ -0,0 +1,72 @@
...
+
+    queue.enqueue1("automation-test-message").unsafeRunSync()
+
+    eventually(timeout(Span(2, Minutes))) {
+      val size = queue.getSize.unsafeRunSync()
+      size shouldBe 0
+    }
+  }
+
+  "Google subscriber should receive a published message" in { clusterFixture =>
```
Qi77Qi: rename the test properly and move it to its own spec potentially

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoModelCopy.scala
```diff
@@ -42,7 +42,8 @@ case class ClusterRequest(labels: LabelMap = Map(),
                           welderDockerImage: Option[String] = None,
                           scopes: Set[String] = Set.empty,
                           enableWelder: Option[Boolean] = None,
-                          customClusterEnvironmentVariables: Map[String, String] = Map.empty)
+                          customClusterEnvironmentVariables: Map[String, String] = Map.empty,
+                          allowStop: Option[Boolean] = Some(false))
```
Qi77Qi: can we do `allowStop: Boolean = false`

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20200122_cluster_followup_table.xml
```diff
@@ -0,0 +1,20 @@
...
+
+    <changeSet logicalFilePath="leonardo" author="jdcanas" id="cluster-followup-table">
+        <createTable tableName="CLUSTER_FOLLOWUP">
+            <column name="clusterId" type="BIGINT">
+                <constraints primaryKey="true"/>
+            </column>
+            <column name="status" type="VARCHAR(254)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="masterMachineType" type="varchar(256)"></column>
```
Qi77Qi, line 12: is `masterMachineType` the only thing we need a stop in order to update?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
```diff
@@ -242,6 +279,72 @@ class ClusterHelper(
...
+          case ClusterOutOfDate                 => IO.raiseError(ClusterOutOfDateException())
+        }
+        _ <- if (welderAction == DisableDelocalization && !cluster.labels.contains("welderInstallFailed"))
+          dbRef.inTransaction { labelQuery.save(cluster.id, "welderInstallFailed", "true") } else IO.unit
+
+        // Start the cluster in Google
+        _ <- startCluster(updatedCluster, welderAction)
+
+        // Update the cluster status to Starting
+        now <- IO(Instant.now)
```
Qi77Qi, line 300: this is duplicated in line 287

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
```diff
@@ -255,7 +358,22 @@ class ClusterHelper(
...
     } yield ()
 
-  def setMasterMachineType(cluster: Cluster, machineType: MachineType): IO[Unit] =
+  def updateMasterMachineType(existingCluster: Cluster, machineType: MachineType): IO[Unit] =
+    for {
+      _ <- log.info(
+        s"New machine config present. Changing machine type to ${machineType} for cluster ${existingCluster.projectNameString}..."
+      )
+      // Update the machine type in Google
+      _ <- setMasterMachineTypeInGoogle(existingCluster, MachineType(machineType.value))
```
Qi77Qi: `machineType` is the right type. You don't need to unwrap it and re-wrap it

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
```diff
@@ -255,7 +358,22 @@ class ClusterHelper(
...
+      // Update the DB
+      now <- IO(Instant.now)
+      _ <- dbRef.inTransaction {
+        clusterQuery.updateMasterMachineType(existingCluster.id, MachineType(machineType.value), now)
```
Qi77Qi: same as previous comment

Qi77Qi:
I'd also like to see a documentation somewhere that list clearly what fields require `stop`, what fields require `delete`, and what fields require no stop nor delete..something like
```
noo stop, no delete required: field 1, field2..
needs stop: field 3, field 4...
needs delete: field 5, field 6...
```

Just to double check, this PR only supports updating `machineType`?

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20200122_cluster_followup_table.xml
jdcanas: yes

Qi77Qi (approved):

Qi77Qi:
oh...don't forget about tf alpha

jdcanas:
> oh...don't forget about tf alpha

https://github.com/broadinstitute/terraform-terra/pull/121

rtitle (approved):
🚀

PR #1239, opened 2020-01-28
[IA-1637] Hail jobs fail with "mismatched python version"

rtitle:
Corresponding terra-docker PR: https://github.com/DataBiosphere/terra-docker/pull/99

---

In all cases:


automation/src/test/resources/bucket-tests/hail_samples.csv
```diff
@@ -0,0 +1,2505 @@
+,sample,age,ancestry,bmi,fg,fi,hdl,height,ldl,population,sex,tc,tg,whr
```
rtitle, line 1: This is synthetic data, and the same as in this workspace: https://app.terra.bio/#workspaces/fc-product-demo/2019_ASHG_Reproducible_GWAS

For some reason I couldn't reproduce the Hail error on a small (5x5) dataframe. But this data does reproduce it. Maybe the dataset needs to be larger in order to be distributed to the workers -- just guessing.

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookHailSpec.scala
```diff
@@ -116,5 +116,43 @@ class NotebookHailSpec extends ClusterFixtureSpec with NotebookTestUtils {
...
+                  |hl.init()
+                """.stripMargin
+              notebookPage.executeCell(importHail)
+
+              // Import the DataFrame into a Hail table
+              val result = notebookPage.executeCell(s"samples = hl.Table.from_pandas(df, key = 'sample')")
+              result shouldBe 'defined
+              result.get should not include ("FatalError")
+              result.get should not include ("PythonException")
+              // TODO what's a positive verification?
```
rtitle: I'll add some more assertions here

jenkins/Jenkinsfile
```diff
@@ -279,9 +279,13 @@ pipeline {
           script {
             for(String dp_version: dataproc_versions.trim().split(' ')) {
               def dp_version_formatted = dp_version.replace(".","-")
+              // Dataproc 1.2 is used for the legacy image and must have python 3.6 installed
```
rtitle, line 282: This is a little hacky, but couldn't think of a cleaner way. Basically we want:
1. legacy image -- dataproc 1.2 -- python 3.6.8 on the workers
2. terra-docker image -- dataproc 1.4 -- python 3.7.4 on the workers

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookHailSpec.scala
kyuksel: Is there a way to get a slice of the frame to use as expected value in positive verification?

kyuksel: It looks like you're doing that below already so it should be good?

kyuksel (commented):
Did you get to verify the Jenkins job with both the legacy and the non-legacy images to make sure they result in the right version of Python installed on workers?

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookHailSpec.scala
rtitle: Ah, the below assertion was copy-pasted and probably not correct (it should have more than 4 rows). This is WIP, still playing around with it.

rtitle:
> Did you get to verify the Jenkins job with both the legacy and the non-legacy images to make sure they result in the right version of Python installed on workers?

Yup running here: https://fc-jenkins.dsp-techops.broadinstitute.org/job/leonardo-dataproc-image/268/console, when it's done I'll try out the generated images.

Qi77Qi (approved):

jenkins/Jenkinsfile
```diff
@@ -173,7 +173,7 @@ pipeline {
 
     stage('Leonardo Git') {
       steps {
-        git credentialsId: 'jenkins-ssh-github', url: 'git@github.com:DataBiosphere/leonardo.git', branch: 'develop'
+        git credentialsId: 'jenkins-ssh-github', url: 'git@github.com:DataBiosphere/leonardo.git', branch: 'rt-ia-1637'
```
rtitle: TODO revert me!

rtitle:

PR #1248, opened 2020-01-30
Jc add user message

jdcanas:
what do you think about this enhancement, to help provide context to a user?


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -86,9 +86,11 @@ case class ClusterOutOfDateException()
       StatusCodes.Conflict
     )
 
-case class ClusterCannotBeUpdatedException(cluster: Cluster)
-    extends LeoException(s"Cluster ${cluster.projectNameString} cannot be updated in ${cluster.status} status",
-                         StatusCodes.Conflict)
+case class ClusterCannotBeUpdatedException(cluster: Cluster, userHint: String = "")
```
Qi77Qi, line 89: can we take `projectNameString` and `status` for the exception instead of the whole `Cluster`? Just to give refactoring more flexibility

Qi77Qi (approved):
👍 lgtm..one minor comment

rtitle (approved):

PR #1255, opened 2020-02-04
IA-1632 Don't publish image to dockerhub

Qi77Qi:
part of https://broadworkbench.atlassian.net/browse/IA-1632

---

In all cases:

### DIFF
--- a/docker/build.sh
+++ b/docker/build.sh
@@ -201,11 +201,6 @@ function docker_cmd()
                 $DOCKER_REMOTES_BINARY push $DOCKERHUB_IMAGE:${DOCKER_TAG}
                 $DOCKER_REMOTES_BINARY tag $DEFAULT_IMAGE:${DOCKER_TAG} $DOCKERHUB_IMAGE:${GIT_BRANCH}
                 $DOCKER_REMOTES_BINARY push $DOCKERHUB_IMAGE:${GIT_BRANCH}
-
-                echo "pushing $TESTS_IMAGE docker image..."
-                $DOCKER_REMOTES_BINARY push $TESTS_IMAGE:${DOCKER_TAG_TESTS}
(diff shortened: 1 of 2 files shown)


rtitle:
I seem to recall there is some piece of our CI which still checks dockerhub - maybe something in fiab deployment? Let me look…

rtitle:
Here it is. 
- fiab docker-compose reads from Dockerhub :facepalm:
   - https://github.com/broadinstitute/firecloud-develop/blob/dev/run-context/fiab/configs/leonardo/docker-compose.yaml.ctmpl#L22
- live docker-compose reads from GCR:
   - https://github.com/broadinstitute/firecloud-develop/blob/dev/run-context/live/configs/leonardo/docker-compose.yaml.ctmpl#L4

There may be other things lurking too. We should make sure we review all CI/CD before removing the dockerhub push.

rtitle (changes requested):
Check CI/CD scripts

Qi77Qi:
@rtitle  https://github.com/broadinstitute/firecloud-develop/pull/2001

rtitle:
The fc-develop test run failed… not sure we can merge this until https://github.com/broadinstitute/firecloud-develop/pull/2001 is merged, right? Otherwise fiabs will fail to find the right images in dockerhub.

Actually, I'm not sure how this PR test run passed 🤔

rtitle:
OK, _now_ I think this PR along with https://github.com/broadinstitute/firecloud-develop/pull/2008 should be good. Will merge when tests pass

rtitle (approved):

rtitle:
Failed with `"e6562d07b6d53eff1bc645d231d7d9cc9316430b" did not start with substring "ed93873"`. So we still have some issue.

It seems different though.. the correct dev welder hash is `e6562d0`. This seems like the Leo version is correct on the fiab, but it's using an old _test_ image. Hmm..

rtitle:
Ah, it's because this PR skips building `Dockerfile-tests` which is still in dockerhub only. Gonna look at 1 more change

PR #1260, opened 2020-02-05
[PROD-340] OOM issues in AoU notebooks

rtitle:
See https://broadworkbench.atlassian.net/browse/PROD-340

For an `n1-highmem-4` machine (which AoU uses), the previous `mem_limit` was calculated to be `20.5G`. I observed only 20G available memory on the VM when I ssh'd in (soo close). Indeed I was able to crash the VM by using all available memory in the kernel.

With this change, the OOM graceful exit works as expected for `n1-highmem-4` machine type. I re-verified `n1-standard-2`, `n1-standard-4`, and `n1-standard-8` machine types as well.

Note - this is just a band-aid fix until we move to GCE. Not only will GCE free up a ton of memory compared to Dataproc, but we'll have a lot more control over the environment and shouldn't have to make estimations like this.

---

In all cases:


Qi77Qi (approved):
sometime later, I think we should make machine type an ADT as well, just it's clear what machine types we support (instead of supporting all machine type which can randomly fail)..

PR #1262, opened 2020-02-05
[IA-1588] GCE Init Script

kyuksel:
[IA-1588](https://broadworkbench.atlassian.net/browse/IA-1588)

`gce-init.sh` is a fusion of `init-actions.sh` and `startup.sh` used to spin up Dataproc clusters with some GCE-specific adjustments. When `GCE_OPERATION` is set to `creating`, it behaves like `init-actions` and when set to `restarting`, like `startup`.

Tested via the following set-up:
- Create a cluster on dev Terra UI while logged in as dev account
- Create a GCE VM on the GCP console while logged in as test.firecloud.org account with service account set to the pet dev account
-  Add the gce-init script to the manually created GCE VM on the console with templated values taken from the Terra cluster's master VM's startup and init-actions scripts
- Ssh into the GCE VM and iterate running the new script via `sudo google_metadata_script_runner --script-type startup --debug`


http/src/main/resources/jupyter/gce-init.sh
```diff
@@ -0,0 +1,537 @@
...
+ @type tail
+ format none
+ path /var/log/daemon.log
+ pos_file /var/tmp/fluentd.google.user.daemon.pos
+ read_from_head true
+ tag daemon
+</source>
+END
+
+    service google-fluentd reload
```
kyuksel, line 203: ^^ didn't work in my VM manually created off `Ubuntu 1804`. We don't seem to be installing it explicitly for our Dataproc clusters. I think Dataproc has a property that one can set at cluster creation that installs it. GCE doesn't seem to have that. I'll double-check this when creating the GCE image and install it there if necessary.

http/src/main/resources/jupyter/gce-init.sh
```diff
@@ -0,0 +1,537 @@
...
+  if [ ! -z "$JUPYTER_DOCKER_IMAGE" ] ; then
+      echo "Starting Jupyter on cluster $GOOGLE_PROJECT / $CLUSTER_NAME..."
+      docker exec -d $JUPYTER_SERVER_NAME /bin/bash -c "export WELDER_ENABLED=$WELDER_ENABLED && export NOTEBOOKS_DIR=$NOTEBOOKS_DIR && (/etc/jupyter/scripts/run-jupyter.sh $NOTEBOOKS_DIR || /usr/local/bin/jupyter notebook)"
+
+      # TODO: Do we still need this for GCE?
+      if [ "$WELDER_ENABLED" == "true" ] ; then
+          # fix for https://broadworkbench.atlassian.net/browse/IA-1453
+          # TODO: remove this when we stop supporting the legacy docker image
+          docker exec -u root jupyter-server sed -i -e 's/export WORKSPACE_NAME=.*/export WORKSPACE_NAME="$(basename "$(dirname "$(pwd)")")"/' /etc/jupyter/scripts/kernel/kernel_bootstrap.sh
+      fi
```
kyuksel, line 453: I ended up leaving the support for the legacy image in to be safe.

http/src/main/resources/jupyter/gce-init.sh
```diff
@@ -0,0 +1,537 @@
+#!/usr/bin/env bash
+
+# This init script instantiates the tool (e.g. Jupyter) docker images on the Dataproc cluster master node.
```
rtitle: Let's maybe adjust this comment to not say Dataproc. Can remove the `Adapted from` line too i think.

http/src/main/resources/jupyter/gce-init.sh
```diff
@@ -0,0 +1,537 @@
...
+    sleep $wait
+  done
+  return 1
+}
+
+function log() {
+  echo "[$(date +'%Y-%m-%dT%H:%M:%S%z')]: $@"
+}
+
+function betterAptGet() {
```
rtitle: I don't believe this function is used by the init script, maybe we should remove it.

http/src/main/resources/jupyter/gce-init.sh
kyuksel: Thanks for the catch.

http/src/main/resources/jupyter/gce-init.sh
kyuksel: 👍

http/src/main/resources/jupyter/gce-init.sh
```diff
@@ -0,0 +1,528 @@
...
+export JUPYTER_START_USER_SCRIPT_URI=$(jupyterStartUserScriptUri)
+# Include a timestamp suffix to differentiate different startup logs across restarts.
+export JUPYTER_START_USER_SCRIPT_OUTPUT_URI="$(jupyterStartUserScriptOutputBaseUri)-$(date -u "+%Y.%m.%d-%H.%M.%S").txt"
+export NOTEBOOKS_DIR=$(notebooksDir)
+export WELDER_SERVER_NAME=$(welderServerName)
+export WELDER_DOCKER_IMAGE=$(welderDockerImage)
+export WELDER_ENABLED=$(welderEnabled)
+
+# If a Google credentials file was specified, grab the service account json file
+# and set the GOOGLE_APPLICATION_CREDENTIALS EV.
```
jdcanas: nit: use variable or ENV VAR over EV, creates a bit of cognitive load to use infrequent abbreviations, I like comments holding my hand :)

http/src/main/resources/jupyter/gce-init.sh
```diff
@@ -0,0 +1,528 @@
+#!/usr/bin/env bash
```
Qi77Qi, line 1: can we move this file outside `jupyter` dir?

http/src/main/resources/jupyter/gce-init.sh
```diff
@@ -0,0 +1,528 @@
...
+export NOTEBOOKS_DIR=$(notebooksDir)
+export WELDER_SERVER_NAME=$(welderServerName)
+export WELDER_DOCKER_IMAGE=$(welderDockerImage)
+export WELDER_ENABLED=$(welderEnabled)
+
+# If a Google credentials file was specified, grab the service account json file
+# and set the GOOGLE_APPLICATION_CREDENTIALS EV.
+# This overrides the credentials on the metadata server.
+# SERVICE_ACCOUNT_CREDENTIALS=$(jupyterServiceAccountCredentials)
+SERVICE_ACCOUNT_CREDENTIALS=""
```
jdcanas: is line 101 meant to be uncommented? It seems like line 102 always causes the below if to be skipped

http/src/main/resources/jupyter/gce-init.sh
```diff
@@ -0,0 +1,537 @@
...
+export JUPYTER_START_USER_SCRIPT_OUTPUT_URI="$(jupyterStartUserScriptOutputBaseUri)-$(date -u "+%Y.%m.%d-%H.%M.%S").txt"
+export NOTEBOOKS_DIR=$(notebooksDir)
+export WELDER_SERVER_NAME=$(welderServerName)
+export WELDER_DOCKER_IMAGE=$(welderDockerImage)
+export WELDER_ENABLED=$(welderEnabled)
+
+# If a Google credentials file was specified, grab the service account json file
+# and set the GOOGLE_APPLICATION_CREDENTIALS EV.
+# This overrides the credentials on the metadata server.
+# SERVICE_ACCOUNT_CREDENTIALS=$(jupyterServiceAccountCredentials)
```
akarukappadath: When will this be uncommented? Is there a ticket?

http/src/main/resources/jupyter/gce-init.sh
```diff
@@ -0,0 +1,528 @@
...
+SERVICE_ACCOUNT_CREDENTIALS=""
+if [ ! -z ${SERVICE_ACCOUNT_CREDENTIALS} ] ; then
```
jdcanas: nit: I was curious about the ${var} syntax, vs $var and "$var", so I did a bit of digging. Apparently, it's good practice to surround this in quotes, and I think the ${} is unecessary

i.e. `"$SERVICE_ACCOUNT_CREDENTIALS"`

https://stackoverflow.com/questions/18135451/what-is-the-difference-between-var-var-and-var-in-the-bash-shell

tl;dr ${} is only needed if its part of a larger string, "$var" is good to use when invoking control statements on the variable to avoid unintended effects

http/src/main/resources/jupyter/gce-init.sh
```diff
@@ -0,0 +1,528 @@
...
+# .. after user script
+# .. after lab extension install
+# .. after jupyter notebook start
+# END
+
+
+#####################################################################################################
+# Set up that is NOT specific to GCE_OPERATION
+#####################################################################################################
+STEP_TIMINGS=($(date +%s))
```
jdcanas: this is an array of milliseconds since epoch to track the time between the points it is invoked. I'd say its worth either removing them, or thinking about where it is relevant to take a snapshot of how long the script has taken so far. I think accompanying each with a log of what 'step' it is currently on, as a lot of them have, would be sufficient if you want to keep them

http/src/main/resources/jupyter/gce-init.sh
```diff
@@ -0,0 +1,528 @@
...
+    gsutil cp ${WELDER_DOCKER_COMPOSE} /etc
+
+    # Needed because docker-compose can't handle symlinks
+    touch /hadoop_gcs_connector_metadata_cache
+
+    # Not all images have the directory used for Stackdriver configs. If so, create it
+    mkdir -p /etc/google-fluentd/config.d
+
+    # Add stack driver configuration for welder
+    tee /etc/google-fluentd/config.d/welder.conf << END
```
jdcanas, line 168: can these 3 stackdriver configurations be bundled into a function? It seems like the format and 'tool' are the only parameters needed

http/src/main/resources/jupyter/gce-init.sh
```diff
@@ -0,0 +1,528 @@
...
+      gcloud --quiet auth configure-docker
+    fi
+
+    STEP_TIMINGS+=($(date +%s))
+
+    log 'Starting up the Jupydocker...'
+
+    # Run docker-compose for each specified compose file.
+    # Note the `docker-compose pull` is retried to avoid intermittent network errors, but
+    # `docker-compose up` is not retried.
```
jdcanas: is this true? it looks like docker-compose up is retried on line 251. If it isn't, why not?

jdcanas (commented):
Mostly small stylistic/bash stuff, didn't dive too deep into business logic yet

http/src/main/resources/jupyter/gce-init.sh
kyuksel: Great catch! It was left over from testing with hardcoded values.

http/src/main/resources/jupyter/gce-init.sh
kyuksel: Agreed. When I see `EV`, my mind goes to electric vehicle  🤷‍♂

http/src/main/resources/jupyter/gce-init.sh
kyuksel: No. See https://github.com/DataBiosphere/leonardo/pull/1262#discussion_r377183676.

http/src/main/resources/jupyter/gce-init.sh
kyuksel: Cool - thanks for the reference!

http/src/main/resources/jupyter/gce-init.sh
rtitle: Why? Our current pattern is to use the `jupyter` dir for resources that need to be uploaded to the init bucket.

http/src/main/resources/jupyter/gce-init.sh
rtitle: I think we decided we would remove the code around `SERVICE_ACCOUNT_CREDENTIALS` and `GOOGLE_APPLICATION_CREDENTIALS` since it's never used.

http/src/main/resources/jupyter/gce-init.sh
rtitle: Cool; I can never remember this stuff

http/src/main/resources/jupyter/gce-init.sh
rtitle: +1

http/src/main/resources/jupyter/gce-init.sh
rtitle: It might be a debian thing (systemd?). May need an ubuntu analog if going that route.

http/src/main/resources/jupyter/gce-init.sh
rtitle: Yeah there's a discrepancy between code and comment. I think we wanted to retry `pull` because it could fail intermittently, but failures on `up` means something is broken and isn't really helped by a retry.

http/src/main/resources/jupyter/gce-init.sh
```diff
@@ -0,0 +1,528 @@
...
+      cat /etc/`basename ${WELDER_DOCKER_COMPOSE}`
+    fi
+
+    retry 5 docker-compose "${COMPOSE_FILES[@]}" config
+    retry 5 docker-compose "${COMPOSE_FILES[@]}" pull
+    retry 5 docker-compose "${COMPOSE_FILES[@]}" up -d
+
+    STEP_TIMINGS+=($(date +%s))
+
+    # If we have a service account JSON file, create an .env file to set GOOGLE_APPLICATION_CREDENTIALS
```
rtitle: We can remove this block

http/src/main/resources/jupyter/gce-init.sh
```diff
@@ -0,0 +1,528 @@
...
+
+    # Jupyter-specific setup, only do if Jupyter is installed
+    if [ ! -z ${JUPYTER_DOCKER_IMAGE} ] ; then
+      log 'Installing Jupydocker kernelspecs...'
+
+      # Change Python and PySpark 2 and 3 kernel specs to allow each to have its own spark
+      # TODO This is baked into terra-jupyter-base as of version 0.0.6. Keeping it here for now to support prior image versions.
+      retry 3 docker exec -u root ${JUPYTER_SERVER_NAME} ${JUPYTER_SCRIPTS}/kernel/kernelspec.sh ${JUPYTER_SCRIPTS}/kernel ${KERNELSPEC_HOME}
+
+      # Install hail addition if the image is old leonardo jupyter image or it's a hail specific image
```
rtitle: Can probably remove this, we won't support hail on GCE

http/src/main/resources/jupyter/gce-init.sh
```diff
@@ -0,0 +1,528 @@
...
+            retry 3 docker exec ${JUPYTER_SERVER_NAME} ${JUPYTER_SCRIPTS}/extension/jupyter_install_lab_extension.sh $ext
+          fi
+        done
+      fi
+
+      STEP_TIMINGS+=($(date +%s))
+
+      # fix for https://broadworkbench.atlassian.net/browse/IA-1453
+      # TODO: remove this when we stop supporting the legacy docker image
+      if [ ! -z ${WELDER_DOCKER_IMAGE} ] && [ "${WELDER_ENABLED}" == "true" ] ; then
```
rtitle: I think remove this too, since we won't support the legacy image on GCE

http/src/main/resources/jupyter/gce-init.sh
rtitle: I think remove it, the legacy image can be Dataproc-only

http/src/main/resources/jupyter/gce-init.sh
```diff
@@ -0,0 +1,528 @@
...
+    log 'Starting up the Jupydocker...'
+
+    # Run docker-compose for each specified compose file.
+    # Note the `docker-compose pull` is retried to avoid intermittent network errors, but
+    # `docker-compose up` is not retried.
+    COMPOSE_FILES=(-f /etc/`basename ${PROXY_DOCKER_COMPOSE}`)
+    cat /etc/`basename ${PROXY_DOCKER_COMPOSE}`
+    if [ ! -z ${JUPYTER_DOCKER_IMAGE} ] ; then
+      COMPOSE_FILES+=(-f /etc/`basename ${JUPYTER_DOCKER_COMPOSE}`)
+      cat /etc/`basename ${JUPYTER_DOCKER_COMPOSE}`
```
rtitle: Do we need a GCE version of `jupyter-docker-compose.yml`? I think the existing one tries to mount in some Hadoop stuff. Or does it work as is?

http/src/main/resources/jupyter/gce-init.sh
```diff
@@ -0,0 +1,528 @@
...
+  export DEPLOY_WELDER=$(deployWelder)
+  export UPDATE_WELDER=$(updateWelder)
+  export DISABLE_DELOCALIZATION=$(disableDelocalization)
+
+  # Sometimes we want to update Welder without having to delete and recreate a cluster
+  if [ "$UPDATE_WELDER" == "true" ] ; then
+      gcloud auth configure-docker
+      docker-compose -f /etc/welder-docker-compose.yaml stop
+      docker-compose -f /etc/welder-docker-compose.yaml rm -f
+      docker-compose -f /etc/welder-docker-compose.yaml up -d
```
rtitle, line 422: Maybe this should also have a `docker-compose pull` with a retry

rtitle:
A few comments above, overall looks good! I noticed some CI issues on this PR, you may want to rebase this branch to get a clean test run.

http/src/main/resources/jupyter/gce-init.sh
kyuksel: I don't think the current format is very user-friendly. I'll make it report the entire execution time for the time being and if it becomes necessary, we can add more granular execution times later.

http/src/main/resources/jupyter/gce-init.sh
kyuksel: Good idea. These will probably move to the GCE image, though. Will keep in mind during that work.

http/src/main/resources/jupyter/gce-init.sh
kyuksel: Good catch! Probably a left-over comment.

http/src/main/resources/jupyter/gce-init.sh
kyuksel: What was the background on this?

http/src/main/resources/jupyter/gce-init.sh
kyuksel: 👍

http/src/main/resources/jupyter/gce-init.sh
kyuksel: Good question. It worked as is since the directories seem to exist on GCE too but it probably makes sense to create a GCE-specific `yml` since we don't need any of those volumes there.

http/src/main/resources/jupyter/gce-init.sh
kyuksel: Yea, I was wondering how this worked in the other script without a `pull`  🤔

http/src/main/resources/jupyter/gce-init.sh
kyuksel: My intuition would be to expect only jupyter-specific files inside the jupyter directory, no? There are other files that'd need to be moved out with that logic, though so perhaps it'd be better to do the reorganization for all later?

http/src/main/resources/jupyter/gce-init.sh
kyuksel: I forgot this. Thank you.

http/src/main/resources/jupyter/gce-init.sh
Qi77Qi: if these are jupyter specific resources, it makes sense to be under `jupyter`. If not, shouldn't we move away from having these files in jupyter?...`jupyter` looks pretty overloaded.

http/src/main/resources/jupyter/gce-init.sh
kyuksel: Yea, let's revisit this once we know what image we'll go with for GCE.

http/src/main/resources/jupyter/gce-init.sh
kyuksel: If the common denominator for the files in that directory is uploading of them to the init bucket, we could also rename the directory as `init-bucket` or something?

http/src/main/resources/jupyter/jupyter-docker-compose-gce.yaml
```diff
@@ -0,0 +1,31 @@
...
+services:
+  jupyter:
+    container_name: "${JUPYTER_SERVER_NAME}"
+    image: "${JUPYTER_DOCKER_IMAGE}"
+    # Override entrypoint with a placeholder to keep the container running indefinitely.
+    # The cluster init script will run some scripts as root and then start pyspark as
+    # jupyter-user via docker exec.
+    entrypoint: "tail -f /dev/null"
+    network_mode: host
+    volumes:
```
kyuksel, line 15: Created this spin off with Hadoop/Spark related volumes removed. Didn't make the Scala changes since we'll refactor ClusterHelper, etc.

kyuksel:
jenkins retest

http/src/main/resources/jupyter/gce-init.sh
rtitle: The `jupyter` dir today is for all resources that are updated to the init bucket. We could rename it - might make sense as it's used for rstudio as well. Maybe name it `init-bucket-resources`?

Also need to change this line: https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/ClusterResourcesConfig.scala#L18

http/src/main/resources/jupyter/gce-init.sh
```diff
@@ -0,0 +1,474 @@
...
+    sleep $wait
+  done
+  return 1
+}
+
+function log() {
+  echo "[$(date +'%Y-%m-%dT%H:%M:%S%z')]: $@"
+}
+
+display_time() {
```
rtitle, line 49: fancy

http/src/main/resources/jupyter/gce-init.sh
```diff
@@ -0,0 +1,474 @@
...
+    gsutil cp ${SERVER_KEY} /certs
+    gsutil cp ${ROOT_CA} /certs
+    gsutil cp ${PROXY_SITE_CONF} /etc
+    gsutil cp ${JUPYTER_DOCKER_COMPOSE} /etc
+    gsutil cp ${RSTUDIO_DOCKER_COMPOSE} /etc
+    gsutil cp ${PROXY_DOCKER_COMPOSE} /etc
+    gsutil cp ${WELDER_DOCKER_COMPOSE} /etc
+
+    # Needed because docker-compose can't handle symlinks
+    touch /hadoop_gcs_connector_metadata_cache
```
rtitle: This may not exist on GCE

http/src/main/resources/jupyter/gce-init.sh
rtitle: Yeah I'd probably lean towards a GCE-specific docker-compose yaml

http/src/main/resources/jupyter/gce-init.sh
```diff
@@ -0,0 +1,474 @@
...
+    if [ ! -z "$WELDER_DOCKER_IMAGE" ] && [ "$WELDER_ENABLED" == "true" ] ; then
+      log 'Starting Welder (file synchronization service)...'
+      retry 3 docker exec -d ${WELDER_SERVER_NAME} /opt/docker/bin/entrypoint.sh
+    fi
+
+    # Jupyter-specific setup, only do if Jupyter is installed
+    if [ ! -z "$JUPYTER_DOCKER_IMAGE" ] ; then
+      log 'Installing Jupydocker kernelspecs...'
+
+      # Change Python and PySpark 2 and 3 kernel specs to allow each to have its own spark
```
rtitle: I'd probably change this comment to `Install kernelspecs inside the Jupyter container`. The TODO is still valid.

http/src/main/resources/jupyter/gce-init.sh
rtitle: If there's no `pull` then `up` does the pull. But I think it's better to have an explicit `pull` with retries.

rtitle (approved):
Looks good - I think this is mergeable if tests are passing. I'll then rebase it into my branch and start to test it out with the Leo code changes.

kyuksel:
seemingly unrelated leonardo-build failure

jenkins retest

PR #1271, opened 2020-02-07
[no ticket] more fix

Qi77Qi:
Since the processor stream only runs in backleo, we shouldn't run subscriber.start in front leo...subscriber.start and processor should run on the same instance

I think this is the real fix

---

In all cases:


jdcanas (approved):

rtitle (approved):
:shipit:

PR #1276, opened 2020-02-12
Clean up listClusters joining and filtering

rtitle:
@Qi77Qi what do you think of this? Note the 1st commit is just `scalafmt`, so look at the second.

---

In all cases:


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/LeonardoServiceDbQueries.scala
```diff
@@ -14,13 +14,20 @@ import org.broadinstitute.dsde.workbench.leonardo.http.service.{
...
 import org.broadinstitute.dsde.workbench.model.WorkbenchEmail
 import org.broadinstitute.dsde.workbench.model.google.{GcsBucketName, GoogleProject}
 
 import scala.concurrent.ExecutionContext
 
 object LeonardoServiceDbQueries {
+
+  type ClusterJoinLabel = Query[(ClusterTable, Rep[Option[LabelTable]]), (ClusterRecord, Option[LabelRecord]), Seq]
+
+  def clusterLabelQuery(baseQuery: Query[ClusterTable, ClusterRecord, Seq]): ClusterJoinLabel =
```
rtitle, line 26: The general pattern is to do filtering before joining… hence these `join` methods take a base query which should be pre-filtered before join.

rtitle: I admit this is not an intuitive way to use slick… but I have been burned by this.

Qi77Qi (approved):

PR #1279, opened 2020-02-14
[IA-982] Jc test autopause

jdcanas:
- Add a test for autopause
- Modifies the save cluster log function to only attempt the save if the cluster is running, as it throws an exception otherwise

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/cluster/ClusterAutopauseSpec.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/cluster/ClusterAutopauseSpec.scala
@@ -1,12 +1,11 @@
 package org.broadinstitute.dsde.workbench.leonardo.cluster
 
 import org.broadinstitute.dsde.workbench.auth.AuthToken
-import org.broadinstitute.dsde.workbench.leonardo.{ClusterFixtureSpec, ClusterStatus, GPAllocFixtureSpec, Leonardo, LeonardoTestUtils}
-import org.scalatest.time.{Minutes, Seconds, Span}
+import org.broadinstitute.dsde.workbench.leonardo.{ClusterStatus, GPAllocFixtureSpec, Leonardo, LeonardoTestUtils}
+import org.scalatest.time.{Seconds, Span}
 import org.scalatest.{DoNotDiscover, ParallelTestExecution}
 
 @DoNotDiscover
-//class ClusterAutopauseSpec extends ClusterFixtureSpec with LeonardoTestUtils {
 class ClusterAutopauseSpec extends GPAllocFixtureSpec with ParallelTestExecution with LeonardoTestUtils {
 
   implicit val ronToken: AuthToken = ronAuthToken
--- a/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/cluster/ClusterAutopauseSpec.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/cluster/ClusterAutopauseSpec.scala
@@ -18,7 +18,7 @@ class ClusterAutopauseSpec extends ClusterFixtureSpec with LeonardoTestUtils {
       clusterRequest = defaultClusterRequest.copy(autopause = Some(true), autopauseThreshold = Some(1))
     )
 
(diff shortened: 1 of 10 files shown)


jdcanas:
failing due to `NotebookClusterMonitoringSpec`, jenkins retest

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/cluster/ClusterAutopauseSpec.scala
```diff
@@ -0,0 +1,28 @@
...
+import org.scalatest.time.{Minutes, Seconds, Span}
+import org.scalatest.DoNotDiscover
+
+@DoNotDiscover
+class ClusterAutopauseSpec extends ClusterFixtureSpec with LeonardoTestUtils {
+
+  "patch autopause then pause properly" in { clusterFixture =>
+    Leonardo.cluster
+      .get(clusterFixture.cluster.googleProject, clusterFixture.cluster.clusterName)
+      .autopauseThreshold shouldBe 0
```
Qi77Qi: what's the behavior when autopause is 0? 🤔 
Can we create a cluster with short autopause threshold directly instead of patching it afterwards?

kyuksel: The logic is 

```scala
private def calculateAutopauseThreshold(autopause: Option[Boolean], autopauseThreshold: Option[Int]): Int =
    autopause match {
      case None =>
        autoFreezeConfig.autoFreezeAfter.toMinutes.toInt
      case Some(false) =>
        autoPauseOffValue
      case _ =>
        if (autopauseThreshold.isEmpty) autoFreezeConfig.autoFreezeAfter.toMinutes.toInt
        else Math.max(autoPauseOffValue, autopauseThreshold.get)
    }
```

where `autoPauseOffValue` is set to `0`.

It's sort of complicated and I couldn't find unit or integration tests that check all of the possibilities 🤷‍♂

jdcanas: it does nothing when its 0 https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/resources/swagger/api-docs.yaml#L980

We can but the patch takes negligible time. I figure since each test is so expensive, its good to test that patching it works as well in the same test.

jdcanas: ah, I found a test in unit tests for the patch. I'll update this, and add a test for the calculateAutopauseThreshold logic

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/cluster/ClusterAutopauseSpec.scala
```diff
@@ -1,27 +1,31 @@
...
-      clusterFixture.cluster.googleProject,
-      clusterFixture.cluster.clusterName,
-      clusterRequest = defaultClusterRequest.copy(autopause = Some(true), autopauseThreshold = Some(1))
-    )
-
-    eventually(timeout(Span(90, Seconds)), interval(Span(10, Seconds))) {
-      val cluster = Leonardo.cluster.get(clusterFixture.cluster.googleProject, clusterFixture.cluster.clusterName)
-      cluster.autopauseThreshold shouldBe 1
-      cluster.status shouldBe ClusterStatus.Stopping
+//class ClusterAutopauseSpec extends ClusterFixtureSpec with LeonardoTestUtils {
```
Qi77Qi: commented out code should be removed?

Qi77Qi (approved):
👍 once commented out code is cleaned up

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/cluster/ClusterAutopauseSpec.scala
jdcanas: I'm not sure how to run non-cluster fixture spec tests locally, this isn't necessarily done yet.

rtitle:
👍

PR #1280, opened 2020-02-14
IA-1669 IA-1672 Use pubsub for cluster creation

Qi77Qi:

---

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
@@ -16,6 +16,7 @@ import org.broadinstitute.dsde.workbench.leonardo.model.ClusterContainerServiceT
 import org.broadinstitute.dsde.workbench.leonardo.model.google.DataprocRole.SecondaryWorker
 import org.broadinstitute.dsde.workbench.leonardo.model.google.GoogleJsonSupport._
 import org.broadinstitute.dsde.workbench.leonardo.model.google._
+import org.broadinstitute.dsde.workbench.leonardo.monitor.LeoPubsubMessage.CreateCluster
 import org.broadinstitute.dsde.workbench.model.WorkbenchIdentityJsonSupport._
 import org.broadinstitute.dsde.workbench.model._
 import org.broadinstitute.dsde.workbench.model.google.GoogleModelJsonSupport.{GcsPathFormat => _, _}
@@ -94,12 +95,6 @@ case class ClusterResource(value: String) extends ValueObject
 
 case class DataprocInfo(googleId: UUID, operationName: OperationName, stagingBucket: GcsBucketName, hostIp: Option[IP])
 
-case class AuditInfo(creator: WorkbenchEmail,
-                     createdDate: Instant,
-                     destroyedDate: Option[Instant],
-                     dateAccessed: Instant,
-                     kernelFoundBusyDate: Option[Instant])
-
 sealed trait ClusterContainerServiceType extends EnumEntry with Serializable with Product {
   def imageType: ClusterImageType
   def proxySegment: String
@@ -141,8 +136,6 @@ object ClusterUI {
     else ClusterUI.Other
 }
 
-case class ClusterImage(imageType: ClusterImageType, imageUrl: String, timestamp: Instant)
-
 final case class RunningCluster(googleProject: GoogleProject,
                                 clusterName: ClusterName,
                                 containers: List[ClusterContainerServiceType])
@@ -178,6 +171,25 @@ final case class Cluster(
 ) {
   def projectNameString: String = s"${googleProject.value}/${clusterName.value}"
   def nonPreemptibleInstances: Set[Instance] = instances.filterNot(_.dataprocRole.contains(SecondaryWorker))
+  def toCreateCluster(runtimeConfig: RuntimeConfig, traceId: Option[TraceId]): CreateCluster = CreateCluster(
+    id,
+    ClusterProjectAndName(googleProject, clusterName),
+    serviceAccountInfo,
+    dataprocInfo,
+    auditInfo,
+    properties,
+    jupyterExtensionUri,
+    jupyterUserScriptUri,
+    jupyterStartUserScriptUri,
+    userJupyterExtensionConfig,
+    defaultClientId,
+    clusterImages,
+    scopes,
+    welderEnabled,
+    customClusterEnvironmentVariables,
+    runtimeConfig,
+    traceId
+  )
   def toListClusterResp(runtimeConfig: RuntimeConfig): ListClusterResponse =
     ListClusterResponse(
       id,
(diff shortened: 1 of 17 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
```diff
@@ -415,6 +427,81 @@ object ClusterTemplateValues {
...
+    ClusterTemplateValues(
+      createCluster.clusterProjectAndName.googleProject.value,
+      createCluster.clusterProjectAndName.clusterName.value,
+      stagingBucketName.map(_.value).getOrElse(""),
+      createCluster.clusterImages.find(_.imageType == Jupyter).map(_.imageUrl).getOrElse(""),
+      createCluster.clusterImages.find(_.imageType == RStudio).map(_.imageUrl).getOrElse(""),
+      proxyConfig.jupyterProxyDockerImage,
+      createCluster.clusterImages.find(_.imageType == Welder).map(_.imageUrl).getOrElse(""),
+      initBucketName
+        .map(n => GcsPath(n, GcsObjectName(clusterFilesConfig.jupyterServerCrt.getName)).toUri)
```
jdcanas, line 450: can this be refactored into a function that takes the relevant name from the config?

i.e. 
```
 initBucketName
        .map(n => GcsPath(n, GcsObjectName(clusterResourcesConfig.jupyterDockerCompose.value)).toUri)
        .getOrElse("")
```

turns to 

```
bucketNameToUri(initBucketName, clusterResourcesConfig.jupyterDockerCompose.value)
```

It would look cleaner as one line instead of 3 in my opinion.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriberModels.scala
```diff
@@ -0,0 +1,215 @@
...
+      extends LeoPubsubMessage {
+    val messageType = "stopUpdate"
+  }
+
+  case class ClusterTransitionFinishedMessage(clusterFollowupDetails: ClusterFollowupDetails, traceId: Option[TraceId])
+      extends LeoPubsubMessage {
+    val messageType = "transitionFinished"
+  }
+
+  case class CreateCluster(id: Long,
```
jdcanas, line 31: Do we want to suffix this with message to stay consistent with the other two? Im fine with dropping it from the other two if we want to be less verbose since its already a part of LeoPubsubMessage (which it wasn't when they were originally named), but I think we should stay consistent.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriberModels.scala
```diff
@@ -0,0 +1,215 @@
...
+  }
+
+  final case class ClusterFollowupDetails(clusterId: Long, clusterStatus: ClusterStatus)
+      extends Product
+      with Serializable
+}
+
+final case class PubsubException(message: String) extends Exception
+
+object LeoPubsubCodec {
```
jdcanas, line 59: Maybe we should move the Codec to another file. I can see this file getting unwieldy pretty fast

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriberModels.scala
```diff
@@ -0,0 +1,215 @@
...
+  implicit val leoPubsubMessageEncoder: Encoder[LeoPubsubMessage] = Encoder.instance { message =>
+    message match {
+      case m: StopUpdateMessage                => m.asJson
+      case m: ClusterTransitionFinishedMessage => m.asJson
+      case m: CreateCluster => m.asJson
+    }
+  }
+}
+
+sealed trait PubsubHandleMessageError extends NoStackTrace {
```
jdcanas: same thought as with the codec, it may be worth moving the error handling object and trait to another file to keep the single responsibility principle for these files intact

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
Qi77Qi: this is direct copy from existing code, which Rob will have some refactoring so I'd like to keep it as is so that later when we merge his PR, it's more obvious that this code can be refactored the exact same way

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriberModels.scala
Qi77Qi: I'll get rid of the suffix for the other ones.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriberModels.scala
Qi77Qi: I think it's okay to keep the codec in the same file...models for this file alone isn't that big...if it gets bigger, we can easily move it then.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriberModels.scala
Qi77Qi: I'll move these back to the other file

jdcanas: Why do we want to suppress the stack trace here with NoStackTrace? Users shouldn't see any of these messages as is, they are all back-leo logs.

Qi77Qi: having stacktrace is more costly and it won't be helpful much since all the error cases are uniquely defined, so there won't be ambiguity where the error comes from

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/ClusterHelper.scala
```diff
@@ -87,54 +84,53 @@ class ClusterHelper(
   import dbRef._
 
   def createCluster(
-    cluster: Cluster,
-    runtimeConfig: RuntimeConfig
+    params: CreateCluster
```
jdcanas, line 87: I remember you talking about this approach at our mobbing, I like how it looks

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoServiceSpec.scala
```diff
@@ -1242,11 +1238,12 @@ class LeonardoServiceSpec
...
 
-    finalQueueSize shouldBe 1
+    finalQueueSize shouldBe 2 //one message for creating cluster, one message for updating cluster
 
-    val message = queue.dequeue1.unsafeRunSync().asInstanceOf[StopUpdateMessage]
+    queue.dequeue1.unsafeRunSync() //discard createCluster message
+    val message = queue.dequeue1.unsafeRunSync().asInstanceOf[StopUpdate]
 
-    val castMessage = message.asInstanceOf[StopUpdateMessage]
+    val castMessage = message.asInstanceOf[StopUpdate]
```
jdcanas, line 1246: I know it was there already, but isn't this `asInstanceOf` redundant with the one on line 1244?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
```diff
@@ -415,6 +427,81 @@ object ClusterTemplateValues {
         .getOrElse(""),
       clusterResourceConstraints.map(_.memoryLimit.toString).getOrElse("")
     )
+
+  def fromCreateCluster(createCluster: CreateCluster,
```
jdcanas, line 431: I think this method name is a bit misleading. It leads me to believe it is the equivalent of an `unapply` for `CreateCluster` with respect to `toCreateCluster`, where as it is really wrapping `CreateCluster` and other information needed for templating into another datatype

jdcanas (approved):
Looks good, a few nits.

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoServiceSpec.scala
Qi77Qi: yea...this is an oversight..

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
Qi77Qi: what would you recommend as a better naming? I think over time, these methods should go away as we slowly deprecating generic `Cluster` being used everywhere

Qi77Qi:
I'm going to merge this as is and fix https://github.com/DataBiosphere/leonardo/pull/1280#discussion_r381262327 in the next PR

Qi77Qi:
@rtitle would appreciate your review once you're back from vacation :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModel.scala
rtitle: Seems reasonable to me, I am refactoring `ClusterTemplateValues` to its own file

rtitle:
This looks fine.. one question about the flow. 

1. `LeonardoService` (front Leo) creates the DB record and enqueues a message
2. `LeoPubsubMessageSubscriber` (back Leo) receives the message and creates the cluster in Google.
3. `ClusterMonitorActor` (back Leo) starts polling Google for the cluster to be ready

Is it possible for (3) to start before (2) finishes? And does that cause any issues? I'm wondering if we want (3) to only query for clusters that have already been created in Google.

Qi77Qi:
>  I'm wondering if we want (3) to only query for clusters that have already been created in Google

I think that's a good idea...I'll look into that

PR #1281, opened 2020-02-14
IA-1587 Refactoring and GCE

rtitle:
## Overview
This PR implements GCE in Leo, but also does a lot of refactoring to make the integration cleaner. I'll try to explain the high level changes below, and annotate the PR with specific comments where it makes sense.

## Models
The majority of this PR is model refactoring. Highlights:
* Removed `LeonardoModel` and `GoogleModel` files, moved most of the models to `core`
* core models:
   * Split up roughly by functionality into `runtimeModels`, `dataprocModels`, and `containerModels` files
   * I tried to avoid using the word "Cluster" in non-dataproc models, which led to a ton of changes
   * Added circe encoders/decoders for all models
* http models:
   * `LeonardoServiceContractModel` contains our contracts for API requests/responses
   * `LeoPubsubMessageSubscriberModels` contains models for pubsub messaging
   * The http models use a mixture of circe and spray-json codecs. I decided to keep our spray-json encoders for API responses to not risk API-breaking changes. The spray-json stuff is localized to `LeoRoutesSprayJsonCodec`.

## Configs
I reorganized `reference.conf` a bit:
* It now lists all available Leo configs. Templated fields are either commented or annotated with "replace_me". The only configs in fc-develop are templated configs. 
   * fc-develop PR: https://github.com/broadinstitute/firecloud-develop/pull/2034
* Moved some things out of `DataprocConfig` to more appropriate homes

## Business Logic
* Removed `GoogleComputeDAO` and `HttpGoogleComputeDAO` entirely in favor of wb-libs `GoogleComputeService`
   * wb-libs PR: https://github.com/broadinstitute/workbench-libs/pull/284
* Moved `ClusterHelper.initializeBucketObjects` to `BucketHelper`
* Added a `VPCHelper` class for VPC/subnet/firewall-related stuff
* Moved `ClusterTemplateValues` out of `LeonardoModel` to a dedicated file in `util`
* Added a `ComputeHelper` class for GCE support (still has some todo's)
* Renamed `http/src/main/resources/jupyter` to `http/src/main/resources/init-resources`

## Testing
* Made unit tests work with all the above changes (they all pass, with some TODOs)
* Made minimal changes to automation tests. I left the `*Copy` classes, I want to confirm all the old contracts still work. Auto tests not passing yet.

## TODO
* Get auto tests passing
* Finish GCE integration in ClusterMonitorActor, ComputeHelper, ZombieMonitor
* Finish VPCHelper
* Add new routes 
* Integrate new init script
[...]

PR #1284, opened 2020-02-19
IA-1669 add gce api

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-1669

1. Refactoring around routes
2. Add route that supports both GCE and dataproc
3. Remove some type aliases introduced from last PR.
4. [**Attention**] Drop "properties" from top level `createClusterRequest`. It's now under `machineConfig`. This should be okay since there's no cluster that's using this field in prod.

TODO (separate PR):
1. Add swagger doc https://broadworkbench.atlassian.net/browse/IA-1731

---

In all cases:


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeService.scala
```diff
@@ -0,0 +1,277 @@
...
+//            ) >> IO.raiseError(BucketObjectAccessException(userEmail, gcsPath))
+//          case e if when401(e) =>
+//            log.warn(e)(s"Could not validate object [${gcsUri}] as user [${userEmail.value}]")
+//        }
+    }
+  }
+}
+
+object RuntimeService {
+  case class CreateRuntimeRequest(labels: LabelMap,
```
kyuksel: Sorry for the nit but names like `CreateRuntimeRequest` throws me off everytime with my mind first thinking as though there were something called `RuntimeRequest` and this is to create it  🤷‍♂ Not sure if this sort of naming is in too many places now to change but `RuntimeCreationRequest` would read a lot better to me.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/RuntimeRoutes.scala
```diff
@@ -0,0 +1,65 @@
...
+        }
+      }
+    }
+  }
+
+  private[api] def createRuntimeHandler(userInfo: UserInfo,
+                                   googleProject: GoogleProject,
+                                   runtimeName: ClusterName, //TODO: rename this to RuntimeName once Rob's PR is in
+                                   req: CreateRuntimeRequest): IO[ToResponseMarshallable] = for {
+    traceId <- IO(UUID.randomUUID())
```
kyuksel: It looks like a lot happens in the line below which would be good to unpack in a walk-through or something, and potentially add a comment here.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/RuntimeRoutes.scala
```diff
@@ -0,0 +1,65 @@
...
+import scala.concurrent.ExecutionContext
+
+class RuntimeRoutes(runtimeService: RuntimeService[IO])(implicit timer: Timer[IO], ec: ExecutionContext) extends UserInfoDirectives {
+  def runtimeRoutes: server.Route = requireUserInfo { userInfo =>
+    implicit val traceId = ApplicativeAsk.const[IO, TraceId](TraceId(UUID.randomUUID()))
+    pathPrefix("google/v1/runtime") {
+      pathPrefix(googleProjectSegment / Segment) { (googleProject, runtimeNameString) =>
+        validateRuntimeNameDirective(runtimeNameString) { runtimeName =>
+          pathEndOrSingleSlash {
+            post {
```
rtitle, line 36: create runtime should be `PUT`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/RuntimeRoutes.scala
```diff
@@ -0,0 +1,65 @@
...
+import org.broadinstitute.dsde.workbench.leonardo.model.google.ClusterName
+import org.broadinstitute.dsde.workbench.model.{TraceId, UserInfo}
+import org.broadinstitute.dsde.workbench.model.google.GoogleProject
+
+import scala.concurrent.ExecutionContext
+
+class RuntimeRoutes(runtimeService: RuntimeService[IO])(implicit timer: Timer[IO], ec: ExecutionContext) extends UserInfoDirectives {
+  def runtimeRoutes: server.Route = requireUserInfo { userInfo =>
+    implicit val traceId = ApplicativeAsk.const[IO, TraceId](TraceId(UUID.randomUUID()))
+    pathPrefix("google/v1/runtime") {
```
rtitle: We should have equivalent paths in the new scheme for get runtimes, list runtimes, stop/start runtime, and delete runtime. Could be done in a follow-up PR if this one is just scoped to `create runtime`.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -951,4 +881,74 @@ object LeonardoService {
...
+    val defaultLabels = DefaultLabels(
+      clusterName,
+      googleProject,
+      creator,
+      serviceAccountInfo.clusterServiceAccount,
+      serviceAccountInfo.notebookServiceAccount,
+      clusterRequest.jupyterUserScriptUri,
+      clusterRequest.jupyterStartUserScriptUri,
+      clusterImages.map(_.imageType).filterNot(_ == Welder).headOption
+    ).toJson.asJsObject.fields.mapValues(labelValue => labelValue.convertTo[String])
```
rtitle: Should this use `DefaultLabels.toMap`?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeService.scala
```diff
@@ -0,0 +1,277 @@
...
+import org.broadinstitute.dsde.workbench.leonardo.model.Cluster.getClusterUrl
+import RuntimeService._
+import org.broadinstitute.dsde.workbench.google2.GoogleStorageService
+import org.broadinstitute.dsde.workbench.leonardo.http.api.RuntimeServiceAPIContext
+import org.broadinstitute.dsde.workbench.leonardo.monitor.LeoPubsubMessage
+import org.broadinstitute.dsde.workbench.model.google
+
+import scala.concurrent.ExecutionContext
+
+class RuntimeService[F[_]: ContextShift](config: RuntimeServiceConfig,
```
rtitle: Question: do we need a separate `RuntimeService` class? I was thinking `LeonardoService` would handle create/get/stop/start/delete runtime requests, and call the appropriate `Helper` methods depending on whether it's Dataproc or GCE.

We could also rename `LeonardoService` to `RuntimeService`.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/RuntimeRoutes.scala
Qi77Qi: regular scala doesn't allow us to define an implicit value in a for comp...this syntax is from https://github.com/oleg-py/better-monadic-for#define-implicits-in-for-comprehensions-or-matches

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/RuntimeRoutes.scala
Qi77Qi: I always thought `POST` is for creating a new entity, `PUT` is for updating a whole entity. Why do we want `PUT` for creating a runtime?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
Qi77Qi: I didn't want to replace existing code too much. But I guess this one should be pretty safe to do

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeService.scala
Qi77Qi: I started in `LeonardoService`, but then realize there will be lots of rewrite..and I didn't want to mess with dataproc route too much. I'm hoping once things are more finalized in RuntimeService, we can slowly reorg the code in leonardoservice and runtimeservice a bit

I think the decision whether creating a dataproc cluster or gce vm will be in `LeoPubsubMessageSubscriber`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/RuntimeRoutes.scala
Qi77Qi: yea, I'm thinking this PR will be just create runtime since it's the most complicated one..I may add `get runtime` if I need it in test. But I was planning to add the other ones in a separate one

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/RuntimeRoutes.scala
kyuksel: good to know, thanks

rtitle:
I haven't pulled in the routes yet, but a lot of the conf/model stuff here is done in https://github.com/DataBiosphere/leonardo/pull/1281

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/package.scala
```diff
@@ -4,16 +4,4 @@ package object leonardo {
   type LabelMap = Map[String, String]
   //this value is the default for autopause, if none is specified. An autopauseThreshold of 0 indicates no autopause
   final val autoPauseOffValue = 0
-
-  /**
-   * Aliases. These exist solely for the purpose of making refactoring easier,
```
rtitle, line 9: Heh thanks, these aliases were me being lazy

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/runtimeModels.scala
```diff
@@ -276,14 +276,26 @@ object RuntimeUI {
...
                          creator: WorkbenchEmail,
                          clusterServiceAccount: Option[WorkbenchEmail],
                          notebookServiceAccount: Option[WorkbenchEmail],
                          notebookUserScript: Option[UserScriptPath],
                          notebookStartUserScript: Option[UserScriptPath],
-                         tool: Option[RuntimeImageType])
+                         tool: Option[RuntimeImageType]) {
+  def toMap: LabelMap =
+    Map(
+      "runtimeName" -> runtimeName.asString, //TODO: is this label renaming okay?
```
rtitle: I think we should keep the old key (maybe we can include both `clusterName` and `runtimeName`). Terra UI and AoU both make use of labels.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/HttpRoutes.scala
```diff
@@ -0,0 +1,115 @@
...
+)(implicit timer: Timer[IO], ec: ExecutionContext, ac: ActorSystem, cs: ContextShift[IO]) {
+
+  private val swaggerRoutes = new SwaggerRoutes(swaggerConfig)
+  private val statusRoutes = new StatusRoutes(statusService)
+  private val corsSupport = new CorsSupport(contentSecurityPolicy)
+  private val proxyRoutes = new ProxyRoutes(proxyService, corsSupport)
+  private val leonardoRoutes = new LeoRoutes(leonardoService, userInfoDirectives)
+  private val runtimeRoutes = new RuntimeRoutes(runtimeService, userInfoDirectives)
+
+  //TODO: is this really necessary?
```
rtitle: No - we can remove it I think

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/RuntimeRoutes.scala
rtitle: Hm it does sound like it should be `POST` because it's not idempotent. TIL

https://restfulapi.net/rest-put-vs-post/
https://stackoverflow.com/questions/107390/whats-the-difference-between-a-post-and-a-put-http-request

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/RuntimeRoutes.scala
```diff
@@ -0,0 +1,145 @@
...
+    }
+  }
+
+  private[api] def createRuntimeHandler(userInfo: UserInfo,
+                                        googleProject: GoogleProject,
+                                        runtimeName: RuntimeName, //TODO: rename this to RuntimeName once Rob's PR is in
+                                        req: CreateRuntime2Request): IO[ToResponseMarshallable] =
+    for {
+      traceId <- IO(UUID.randomUUID())
+      implicit0(context: ApplicativeAsk[IO, RuntimeServiceContext]) <- timer.clock
```
rtitle, line 60: Ah thanks for putting this implicit at the highest level

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/RuntimeRoutes.scala
```diff
@@ -0,0 +1,145 @@
...
+          }
+        }
+      }
+    }
+  }
+
+  private[api] def createRuntimeHandler(userInfo: UserInfo,
+                                        googleProject: GoogleProject,
+                                        runtimeName: RuntimeName, //TODO: rename this to RuntimeName once Rob's PR is in
+                                        req: CreateRuntime2Request): IO[ToResponseMarshallable] =
```
rtitle, line 57: Should this return some specific response type instead of `ToResponseMarshallable`?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/RuntimeRoutes.scala
```diff
@@ -0,0 +1,145 @@
...
+            }
+          }
+        }
+      }
+    }
+  }
+
+  private[api] def createRuntimeHandler(userInfo: UserInfo,
+                                        googleProject: GoogleProject,
+                                        runtimeName: RuntimeName, //TODO: rename this to RuntimeName once Rob's PR is in
```
rtitle: Remove TODO

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/RuntimeRoutes.scala
Qi77Qi: I see this function as the helper that fits into `complete(???)`...so I think it should be `ToResponseMarshallable`, at "service" level, each function is returning api specific responses

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/RuntimeRoutes.scala
```diff
@@ -0,0 +1,145 @@
...
+        userInfo,
+        googleProject,
+        runtimeName,
+        req
+      )
+    } yield StatusCodes.Accepted
+}
+
+object RuntimeRoutes {
+  implicit val gceConfigDecoder: Decoder[RuntimeConfigRequest.GceConfig] = Decoder.forProduct2(
```
rtitle, line 77: Can use use https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/LeoRoutesJsonCodec.scala instead and put any new decoders/encoders in there?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/RuntimeRoutes.scala
```diff
@@ -0,0 +1,145 @@
...
+  }
+
+  implicit val createRuntimeRequestDecoder: Decoder[CreateRuntime2Request] = Decoder.instance { c =>
+    for {
+      l <- c.downField("labels").as[Option[LabelMap]]
+      je <- c.downField("jupyterExtensionUri").as[Option[GcsPath]]
+      jus <- c.downField("jupyterUserScriptUri").as[Option[UserScriptPath]]
+      jsus <- c.downField("jupyterStartUserScriptUri").as[Option[UserScriptPath]]
+      rc <- c.downField("runtimeConfig").as[Option[RuntimeConfigRequest]]
+      p <- c.downField("properties").as[Option[LabelMap]]
```
rtitle: You mentioned putting `properties` inside of `DataprocConfig` -- that's probably a good idea

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/RuntimeRoutes.scala
```diff
@@ -0,0 +1,145 @@
...
+      uje <- c.downField("userJupyterExtensionConfig").as[Option[UserJupyterExtensionConfig]]
+      a <- c.downField("autopause").as[Option[Boolean]]
+      apt <- c.downField("autopauseThreshold").as[Option[Int]]
+      dc <- c.downField("defaultClientId").as[Option[String]]
+      tdi <- c.downField("toolDockerImage").as[Option[ContainerImage]]
+      wdi <- c.downField("welderDockerImage").as[Option[ContainerImage]]
+      s <- c.downField("scopes").as[Option[Set[String]]]
+      c <- c.downField("customClusterEnvironmentVariables").as[Option[LabelMap]]
```
rtitle: Maybe we can rename this to `customEnvironmentVariables`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/SwaggerRoutes.scala
```diff
@@ -9,11 +9,9 @@ import akka.stream.scaladsl.Flow
 import akka.util.ByteString
 import org.broadinstitute.dsde.workbench.leonardo.config.SwaggerConfig
 
-trait SwaggerRoutes {
+class SwaggerRoutes(swaggerConfig: SwaggerConfig) {
   private val swaggerUiPath = "META-INF/resources/webjars/swagger-ui/2.2.5"
```
rtitle, line 13: Not for this PR, but latest swagger UI is 3.22 :)

https://github.com/swagger-api/swagger-ui/releases?after=v3.22.2
https://swagger.io/tools/swagger-ui/

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -169,21 +169,30 @@ class LeoPubsubMessageSubscriber[F[_]: Async: Timer: ContextShift: Concurrent](
...
+                                        clusterResult.customDataprocImage.asString,
+                                        now)
+            // Save dataproc image in the database
+            _ <- dbRef.inTransaction(clusterImageQuery.save(msg.id, clusterImage))
+            _ <- logger.info(
+              s"Cluster ${msg.clusterProjectAndName} was successfully created. Will monitor the creation process."
+            )
+          } yield ()
+        case CloudService.GCE =>
+          Async[F].unit //TODO: Rob will replace this with appropriate implementation
```
rtitle, line 194: Here it is! I can address this in my PR

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -974,4 +910,82 @@ object LeonardoService {
...
+      googleProject,
+      creator,
+      serviceAccountInfo.clusterServiceAccount,
+      serviceAccountInfo.notebookServiceAccount,
+      request.jupyterUserScriptUri,
+      request.jupyterStartUserScriptUri,
+      clusterImages.map(_.imageType).filterNot(_ == Welder).headOption
+    )
+
+    val defaultLabelsMap = defaultLabels.asJson.asObject
```
rtitle: I see you added a `DefaultLabels.toMap` -- use that instead instead of json encoding to build a map?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeService.scala
```diff
@@ -0,0 +1,15 @@
...
+
+import cats.mtl.ApplicativeAsk
+import org.broadinstitute.dsde.workbench.leonardo.http.api.{CreateRuntime2Request, RuntimeServiceContext}
+import org.broadinstitute.dsde.workbench.model.UserInfo
+import org.broadinstitute.dsde.workbench.model.google.GoogleProject
+
+trait RuntimeService[F[_]] {
+  def createRuntime(userInfo: UserInfo,
+                    googleProject: GoogleProject,
+                    runtimeName: RuntimeName, //TODO: rename this to RuntimeName once Rob's PR is in
```
rtitle: Remove TODO

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeService.scala
```diff
@@ -0,0 +1,15 @@
...
+                    req: CreateRuntime2Request)(implicit as: ApplicativeAsk[F, RuntimeServiceContext]): F[Unit]
```
rtitle, line 14: Is it intended that this returns `F[Unit]` instead of some response? I am wondering if any clients rely on the response returned by `createCluster`.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeServiceInterp.scala
```diff
@@ -0,0 +1,316 @@
...
+import scala.concurrent.ExecutionContext
+
+class RuntimeServiceInterp[F[_]](blocker: Blocker,
+                                 semaphore: Semaphore[F],
+                                 config: RuntimeServiceConfig,
+                                 authProvider: LeoAuthProvider[F],
+                                 serviceAccountProvider: ServiceAccountProvider[F],
+                                 dockerDAO: DockerDAO[F],
+                                 googleStorageService: GoogleStorageService[F],
+                                 publisherQueue: fs2.concurrent.Queue[F, LeoPubsubMessage])(implicit F: Async[F],
```
rtitle: Just curious why have 'explicit' implicits here rather than `F[_]: Async: Logger: Timer: ContextShift` (not sure I have opinion, just trying to learn the convention).

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeServiceInterp.scala
```diff
@@ -0,0 +1,316 @@
...
+                                                                                            log: StructuredLogger[F],
+                                                                                            timer: Timer[F],
+                                                                                            cs: ContextShift[F],
+                                                                                            dbReference: DbReference[F],
+                                                                                            ec: ExecutionContext)
+    extends RuntimeService[F] {
+
+  def createRuntime(userInfo: UserInfo,
+                    googleProject: GoogleProject,
+                    runtimeName: RuntimeName, //TODO: rename this to RuntimeName once Rob's PR is in
```
rtitle: remove TODO

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeServiceInterp.scala
```diff
@@ -0,0 +1,316 @@
...
+                    req: CreateRuntime2Request)(implicit as: ApplicativeAsk[F, RuntimeServiceContext]): F[Unit] =
+    for {
+      context <- as.ask
+      implicit0(traceId: ApplicativeAsk[F, TraceId]) <- F.pure(ApplicativeAsk.const[F, TraceId](context.traceId)) //TODO: is there a better way to do this?
```
rtitle, line 55: Is this needed since the implicit ev is passed to this method?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeServiceInterp.scala
```diff
@@ -0,0 +1,316 @@
...
+        .getNotebookServiceAccount(userInfo, googleProject)
+      serviceAccountInfo = ServiceAccountInfo(clusterServiceAccountOpt, notebookServiceAccountOpt)
+
+      clusterOpt <- clusterQuery.getActiveClusterByNameMinimal(googleProject, runtimeName).transaction
+
+      _ <- clusterOpt match {
+        case Some(c) => F.raiseError[Unit](RuntimeAlreadyExistsException(googleProject, runtimeName, c.status))
+        case None =>
+          for {
+            internalId <- F.pure(RuntimeInternalId(context.traceId.asString)) //TODO: is there any downside to reusing traceId as internalId?
```
rtitle: Hm.. this is also used as a Sam id, are there security issues with exposing internal Sam resource ids in error messages, etc? On the other hand it's just an opaque UUID so it's probably fine

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeServiceInterp.scala
```diff
@@ -0,0 +1,316 @@
...
+                    .toList
+              )
+              .getOrElse(List.empty) ++ req.jupyterUserScriptUri.map(_.asString) ++ req.jupyterStartUserScriptUri.map(
+              _.asString
+            )
+            _ <- petToken.traverse(
+              t =>
+                gcsObjectUrisToValidate
+                  .traverse_(s => validateBucketObjectUri(userInfo.userEmail, t, s, context.traceId))
+            ) //TODO figure out implicits required for parTraverse
```
rtitle: I think `Parallel` typeclass

rtitle:
Did a first pass, overall looks good! See comments above. I'll do another pass when addressed.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
Qi77Qi: ah, this was addressed previously...bad merge

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeServiceInterp.scala
Qi77Qi: I found a little annoying to always type `Async[F].delay` or `Logger[F].info` etc...I think it makes more sense to have them as context bounds if the function body itself is not directly referencing them...There's a library that allows as to do things like `F.delay` when using context bound, but I haven't looked at it yet (will paste a link if I find it again...saw it on twitter)....

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeServiceInterp.scala
Qi77Qi: oh, this was left over before I moved it to route

Qi77Qi: actually, this is needed because `ApplicativeAsk[F, RuntimeServiceContext]` is different from `ApplicativeAsk[F, TraceId]`....there might be a way to derive `ApplicativeAsk[F, TraceId]` from `ApplicativeAsk[F, RuntimeServiceContext]` , but I didn't spend time look into it

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeServiceInterp.scala
Qi77Qi: I'll use UUID then just in case

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeServiceInterp.scala
Qi77Qi: yea, what's the import tho?

rtitle: `import cats.Parallel` is what I used in https://github.com/DataBiosphere/leonardo/pull/1309/files#diff-2b3a7f343790bfa0bfab5213c8dcfb90

Yeah it's weird it's not cats-effect..

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeService.scala
Qi77Qi: yep...intended...it'll be easier to add fields but harder to remove fields later

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/SwaggerRoutes.scala
Qi77Qi: 3.25.0 it is! https://github.com/swagger-api/swagger-ui/releases :)

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/api/TestLeoRoutes.scala
```diff
@@ -122,14 +137,48 @@ trait TestLeoRoutes {
...
+    userInfoDirectives,
+    contentSecurityPolicy
+  )
+  val timedHttpRoutes = new HttpRoutes(swaggerConfig,
+                                       statusService,
+                                       proxyService,
+                                       leonardoService,
+                                       runtimeService,
+                                       timedUserInfoDirectives,
+                                       contentSecurityPolicy)
```
kyuksel, line 181: Took me some staring to notice the difference between `httpRoutes` and `timedHttpRoutes`. Would have been easier with 

```scala
val timedHttpRoutes = httpRoutes.copy(userInfoDirectives = timedUserInfoDirectives)
```

with `HttpRoutes` turned into a `case class`.

Qi77Qi: case class are usually used for data types....`HttpRoutes` is not a data type at all

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -1242,6 +1236,12 @@ definitions:
         description: |
           Optional, the number of preemptible workers. If unspecified, the default number is 0. Ignored if numberOfWorkers is 0.
           For more information, see https://cloud.google.com/compute/docs/instances/preemptible
+      dataprocProperties:
```
rtitle: It looks like it expects `properties` in the json unless I'm mistaken?

Qi77Qi: you're right...updated

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/RuntimeRoutes.scala
```diff
@@ -0,0 +1,142 @@
...
+object RuntimeRoutes {
+  implicit val gceConfigDecoder: Decoder[RuntimeConfigRequest.GceConfig] = Decoder.forProduct2(
+    "machineType",
+    "diskSize"
+  )((mt, ds) => RuntimeConfigRequest.GceConfig(mt, ds))
+
+  implicit val runtimeConfigDecoder: Decoder[RuntimeConfigRequest] = Decoder.instance { x =>
+    //For newer version of requests, we use `cloudService` field to distinguish whether user is
+    for {
+      cloudService <- x.downField("cloudService").as[CloudService]
```
rtitle, line 85: Should `cloudService` be optional and default to GCE?

rtitle: Eh maybe this is fine -- the whole `runtimeConfig` is optional in the request and defaults to GCE… if you specifiy a `runtimeConfig` you should also specify the type.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeServiceInterp.scala
```diff
@@ -0,0 +1,316 @@
...
+import org.broadinstitute.dsde.workbench.leonardo.http.service.RuntimeServiceInterp._
+import org.broadinstitute.dsde.workbench.leonardo.model._
+import org.broadinstitute.dsde.workbench.leonardo.monitor.LeoPubsubMessage
+import org.broadinstitute.dsde.workbench.leonardo.monitor.LeoPubsubMessage.CreateCluster
+import org.broadinstitute.dsde.workbench.model.google.GoogleProject
+import org.broadinstitute.dsde.workbench.model.{TraceId, UserInfo, WorkbenchEmail, google}
+
+import scala.concurrent.ExecutionContext
+
+class RuntimeServiceInterp[F[_]: Parallel](blocker: Blocker,
```
rtitle, line 34: So the intent is to keep `LeonardoService` for the legacy routes, and build up `RuntimeService` / `RuntimeServiceInterp` as we go for the new routes, right? That's probably fine, but I hope we can have clients switch over soon and delete `LeoRoutes` and `LeonardoService` to avoid too much code duplication. Also we should add a comment to `LeonardoService`/`LeoRoutes` explaining the situation.

rtitle:
This overall looks good I think… I'm interested to see how tests do, jenkins retest

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/RuntimeRoutes.scala
Qi77Qi: yea, also this is for new routes, so I think it should be okay.

Qi77Qi:
@rtitle automation tests failures are from labelCheck..should be fixed now

rtitle (approved):
ok 👍 pending tests passing

PR #1285, opened 2020-02-19
[IA-1596] GCE Image Generator

kyuksel:
This is the first of the two PRs for [IA-1596](https://broadworkbench.atlassian.net/browse/IA-1596). It includes a wrapper script (`create_gce_image.sh`) that I used for development purposes and thought may be good to check in as reference and documentation, too. It runs [Daisy](https://github.com/GoogleCloudPlatform/compute-image-tools/tree/master/daisy), which is a GCP tool that can run various workflows. The workflow this PR defines (`leo_custom_gce_image.wf.json`)
- spins up a VM
- runs the tool installation script that this PR also adds (`prepare_custom_leonardo_gce_image.sh`)
- creates an image out of it
- deletes the VM

The follow-up PR will make changes to our Jenkins flow to automate it all.


kyuksel:
Looks like an irrelevant Selenium failure from `NotebookBioconductorKernelSpec`

jenkins retest

jenkins/gce-custom-images/create_gce_image.sh
```diff
@@ -0,0 +1,40 @@
...
+
+set -e -x
+
+# This script is a wrapper around the command that runs the custom GCE image
+# creation tool - Daisy - for local development and documentation purposes.
+# It should be run from the root of the Leonardo repo.
+# It assumes that a Daisy image was pulled; e.g. docker pull gcr.io/compute-image-tools/daisy:latest
+
+# The date-time suffix is in the format yymmdd-hh-mm but it could be anything.
+OUTPUT_IMAGE_NAME="leo-custom-gce-image-200219-16-05"
```
rtitle, line 11: Eventually this will be Jenkins-ized and these params will turn into Jenkins input parameters, right?

jenkins/gce-custom-images/prepare_custom_leonardo_gce_image.sh
```diff
@@ -0,0 +1,180 @@
...
+openidc_proxy="broadinstitute/openidc-proxy:2.3.1_2"
+anvil_rstudio_base="us.gcr.io/anvil-gcr-public/anvil-rstudio-base:0.0.2"
+anvil_rstudio_bioconductor="us.gcr.io/anvil-gcr-public/anvil-rstudio-bioconductor:0.0.3"
+
+# this array determines which of the above images are baked into the custom image
+# the entry must match the var name above, which must correspond to a valid docker URI
+docker_image_var_names="welder_server leonardo_jupyter terra_jupyter_base terra_jupyter_python terra_jupyter_r terra_jupyter_bioconductor terra_jupyter_gatk openidc_proxy anvil_rstudio_base anvil_rstudio_bioconductor"
+
+# The version of python to install
+python_version="3.7.4"
```
rtitle: This var isn't used I think

jenkins/gce-custom-images/prepare_custom_leonardo_gce_image.sh
```diff
@@ -0,0 +1,180 @@
+#!/usr/bin/env bash
```
rtitle, line 1: It would be good to bake the "hardening" stuff that Albano linked into this script

jenkins/gce-custom-images/create_gce_image.sh
```diff
@@ -0,0 +1,40 @@
...
+
+# Set this to the tag of the Daisy image you had pulled
+DAISY_IMAGE_TAG="latest"
+
+# The source directory should contain `application_default_credentials.json`
```
rtitle, line 16: When we Jenkins-ize this we probably want to use a service account key, rather than application default credentials.

rtitle (approved):
A few comments above, but overall this looks great. Nice approach with Daisy. 👍 

Remaining work is to integrate this into Jenkins and add the hardening commands -- anything else I'm missing?

jenkins/gce-custom-images/create_gce_image.sh
kyuksel: right

jenkins/gce-custom-images/create_gce_image.sh
kyuksel: Correct. This script may not end up being used for Jenkins-izing. I think it's still useful for manual image creation and reference purposes, though.

jenkins/gce-custom-images/prepare_custom_leonardo_gce_image.sh
kyuksel: Right. I wanted to see what that hardening stuff entails and then decide whether it makes sense to subsume it into this script, call it from this script or call it from Jenkins as a subsequent step. It might be good to keep the hardening script (or whatever it turns out to be) a separate thing if it applies to other image creations in the future as well.

jenkins/gce-custom-images/prepare_custom_leonardo_gce_image.sh
kyuksel: 👍 thanks!

kyuksel (commented):
> Remaining work is to integrate this into Jenkins and add the hardening commands

That's correct.

PR #1289, opened 2020-02-20
IA-1672 only monitor cluster once it's been created in google

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-1672

---

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorSupervisor.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorSupervisor.scala
@@ -322,7 +322,7 @@ class ClusterMonitorSupervisor(
 
             case c if c.status == ClusterStatus.Updating => IO(self ! ClusterUpdated(c))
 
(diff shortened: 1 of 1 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
```diff
@@ -423,7 +423,7 @@ class ClusterMonitorActor(
     for {
       dbCluster <- getDbCluster
       next <- dbCluster.status match {
-        case status if status.isMonitored =>
+        case status if status.isMonitored && dbCluster.dataprocInfo.isDefined =>
```
rtitle: Perhaps the check could be here instead, to prevent the cluster from even being monitored?

https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorSupervisor.scala#L325

Qi77Qi: ah cool 👍 updated!

rtitle (approved):

PR #1291, opened 2020-02-20
[IA-1696] Disallow worker updates on stopped clusters

jdcanas:
Relatively small if/else + a unit test

### DIFF
--- a/http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoServiceSpec.scala
+++ b/http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoServiceSpec.scala
@@ -1072,15 +1072,14 @@ class LeonardoServiceSpec
   }
 
   it should "throw an exception when trying to resize a stopped cluster" in isolatedDbTest {
-    // create the clusterx
+    // create a stopped cluster record cluster
     val stoppedCluster: Cluster = makeCluster(3).copy(status = ClusterStatus.Stopped).save()
-//    val cluster =
-    // check that the cluster was created
+
+    // check that the cluster is stopped
     val dbCluster = dbFutureValue { clusterQuery.getClusterById(stoppedCluster.id) }
     dbCluster.map(_.status) shouldBe Some(ClusterStatus.Stopped)
 
-    // set the cluster to Running
-
+    //the below block will fail if the exception we expect isn't thrown. We
     val caught = the[ClusterCannotBeUpdatedException] thrownBy {
       leo
         .updateCluster(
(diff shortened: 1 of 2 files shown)


http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoServiceSpec.scala
```diff
@@ -1071,6 +1071,34 @@ class LeonardoServiceSpec
...
+          testClusterRequest.copy(
+            runtimeConfig = Some(
+              RuntimeConfigRequest.DataprocConfig(numberOfWorkers = Some(2), masterMachineType = None, masterDiskSize = None)
+            )
+          )
+        )
+        .unsafeRunSync()
+    }
+
+    logger.info(s"exception: ${caught.getMessage()}")
```
Qi77Qi: do we really need this log line?

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoServiceSpec.scala
```diff
@@ -1071,6 +1071,34 @@ class LeonardoServiceSpec
...
+          userInfo,
+          stoppedCluster.googleProject,
+          stoppedCluster.clusterName,
+          testClusterRequest.copy(
+            runtimeConfig = Some(
+              RuntimeConfigRequest.DataprocConfig(numberOfWorkers = Some(2), masterMachineType = None, masterDiskSize = None)
+            )
+          )
+        )
+        .unsafeRunSync()
```
Qi77Qi, line 1094: I would do the assersion like this
```
val res = leo
        .updateCluster(
          userInfo,
          stoppedCluster.googleProject,
          stoppedCluster.clusterName,
          testClusterRequest.copy(
            runtimeConfig = Some(
              RuntimeConfigRequest.DataprocConfig(numberOfWorkers = Some(2), masterMachineType = None, masterDiskSize = None)
            )
          )
        ).attempt.unsafeRunSync
res shouldBe Left(ClusterCannotBeUpdatedException("expected message"))
```

jdcanas: I like this approach since it is only dependent on the user hint portion of the expected message, while I feel the overall exception message can change independently

jdcanas: I like this approach since it is only dependent on the user hint portion of the expected message, while I feel the overall exception message can change independently

rtitle (approved):

PR #1302, opened 2020-02-27
IA-1645 Don't leak db connection in unit tests

rtitle:
https://broadworkbench.atlassian.net/browse/IA-1645

My approach here was to remove `DbSingleton` and use `BeforeAndAfterAll` to init/close the slick database between tests. This is how slick recommends to do it. I also made it so we don't have to run liquibase every time.

It seems to fix the issue for me, and it doesn't seem to add too much time to the test run.

---

In all cases:


rtitle:
jenkins retest

rtitle:
jenkins retest

rtitle:
jenkins retest

rtitle:
@Qi77Qi fyi, this is working great for me

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/db/TestComponent.scala
```diff
@@ -2,23 +2,60 @@ package org.broadinstitute.dsde.workbench.leonardo.db
...
+  override def afterAll(): Unit = {
+    dbRef.close()
+    super.afterAll()
+  }
+
+  // This is a bit duplicative of DbReference.init but that method returns a Resource[F, DbRef[F]]
+  // which doesn't play nicely with ScalaTest BeforeAndAfterAll. This version returns an F[DbRef[F]].
+  private def initDbRef: IO[DbRef[IO]] =
+    for {
+      concurrentPermits <- Semaphore[IO](100)
```
kyuksel: Should `100` come from a config?

kyuksel (approved):
Verified `http/test` succeeds when run within an `sbt` shell 👍 
One minor comment inline.
Thanks a lot for fixing this 🙏

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/db/TestComponent.scala
rtitle: Ah yeah good catch, this is already in reference.conf. Will update.

Qi77Qi:
🎉 👍 thanks for fixing this!

PR #1307, opened 2020-03-03
IA-1716 Support aou Leo image in GCE workflow

calbach:
Sister PR to https://github.com/DataBiosphere/terra-docker/pull/106

### DIFF
--- a/jenkins/Jenkinsfile
+++ b/jenkins/Jenkinsfile
@@ -136,6 +136,7 @@ pipeline {
     string(name: "terra_jupyter_python", defaultValue: "0.0.6")
     string(name: "terra_jupyter_bioconductor", defaultValue: "0.0.9")
     string(name: "terra_jupyter_gatk", defaultValue: "0.0.8")
+    string(name: "terra_jupyter_aou", defaultValue: "0.0.1")
     string(name: "anvil_rstudio_base", defaultValue: "0.0.2")
     string(name: "anvil_rstudio_bioconductor", defaultValue: "0.0.3")
     string(name: "leonardo_jupyter", defaultValue: "5c51ce6935da",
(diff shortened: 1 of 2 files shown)


rtitle:
CI failed because of the slash in the branch name:
```
Error parsing reference: "gcr.io/broad-dsp-gcr-public/leonardo:ch/aou-image" is not a valid repository/tag: invalid reference format
```

(btw I closed/re-opened this PR by mistake)

jenkins/Jenkinsfile
```diff
@@ -136,6 +136,7 @@ pipeline {
     string(name: "terra_jupyter_python", defaultValue: "0.0.6")
     string(name: "terra_jupyter_bioconductor", defaultValue: "0.0.9")
     string(name: "terra_jupyter_gatk", defaultValue: "0.0.8")
+    string(name: "terra_jupyter_aou", defaultValue: "0.0.1")
```
rtitle, line 139: If I'm understanding correctly these params are not used by the default case -- instead this script tries to parse https://github.com/DataBiosphere/leonardo/blob/develop/automation/src/test/resources/reference.conf to find the versions to pull. I wonder if we should update this conf and add a test case for the AoU image similar to https://github.com/DataBiosphere/leonardo/blob/develop/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookGATKSpec.scala

jdcanas: Yeah, a test isn't strictly necessary, but I think it would be good in this PR or a follow-up one. A conf entry is preferred to updating the version in `prepare_custom_leonardo_gce_image.sh` each time as then the PR updating the version will automatically test the new image for you. The will also make it so any updates to your Dockerfile in `terra-docker` get automatically built and baked into a new custom image in our CI.

kyuksel: > A conf entry is preferred to updating the version in prepare_custom_leonardo_gce_image.sh each time

Asking to make sure I'm not missing anything: The conf entry is what determines how the corresponding image's version will be updated in `prepare_custom_leonardo_gce_image.sh`, no? In other words, `prepare_custom_leonardo_gce_image.sh` **is** updated each time.

calbach:
Oops, I forgot Leo Jenkins has issues with `/` branches , recreated PR with a friendlier branch name: https://github.com/DataBiosphere/leonardo/pull/1308

Please resume the discussion there.

PR #1309, opened 2020-03-05
IA-1591 Add GCE+Dataproc algebra to Back Leo

rtitle:
Most of the meat of GCE is in here. High level changs:

* Removed `ClusterHelper`, `ComputeHelper` -- created `RuntimeAlgebra`, `DataprocInterpreter`, `GceInterpreter` which is the main integration point with Google/wb-libs.
* Made `ClusterMonitorActor`/`ClusterMonitorSupervisor` handle both GCE and Dataproc
* Made `LeoPubsubMessageSubscriber` handle both GCE and Dataproc
* Made `BucketHelper`, `VPCHelper` take an `F[_]` type parameter instead of `IO`
* Some pubsub model refactoring/renaming

TODO:
- ~Test the actual GCE creation~
- Add routes for GCE stop/start/delete/etc
- Zombie monitor is still Dataproc specific
- Error handling -- implement GCE getErrorDetails() and clean up on failure
   - In fact GCE doesn't fail if the init script fails, how to signal errors?
- Create dedicated VPC subnet instead of using default
- Merge wb-libs PR once all is working & tested
- Improve unit tests around moved/new files
- (Eventually) add Swagger page for new routes

---

In all cases:


rtitle:
jenkins retest

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
```diff
@@ -77,7 +79,8 @@ object ClusterMonitorActor {
     google2StorageDAO: GoogleStorageService[IO],
     dbRef: DbReference[IO],
     authProvider: LeoAuthProvider[IO],
-    clusterHelper: ClusterHelper,
+    dataprocAlg: DataprocAlgebra[IO],
+    gceAlg: RuntimeAlgebra[IO],
```
Qi77Qi: should it be `GceAlgebra` 🤔 ?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -555,7 +540,7 @@ class LeonardoService(
       // Note: instance must be stopped in order to change machine type
       case Some(updatedMasterMachineType) if existingCluster.status == Stopped =>
         for {
-          _ <- clusterHelper.updateMasterMachineType(existingCluster, updatedMasterMachineType)
+          _ <- dataprocAlg.updateMachineType(UpdateMachineTypeParams(existingCluster, updatedMasterMachineType))
```
Qi77Qi: should it be `updateMasterMachineType` still? it seems a bit ambiguous if it's just `updateMachineType`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/BaseRuntimeInterpreter.scala
```diff
@@ -0,0 +1,224 @@
...
+                  s"Failed to flush welder cache for ${params.runtimeAndRuntimeConfig.runtime.projectNameString}"
+                )
+            )
+        } else Async[F].unit
+
+        // Stop the cluster in Google
+        _ <- stopGoogleRuntime(params.runtimeAndRuntimeConfig.runtime, params.runtimeAndRuntimeConfig.runtimeConfig)
+
+        // Update the cluster status to Stopping
+        now <- Async[F].delay(Instant.now)
```
Qi77Qi: shall we have `now` being passed in since it'll exist in api's context?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreter.scala
```diff
@@ -0,0 +1,644 @@
...
+  metrics: NewRelicMetrics[F],
+  dbRef: DbReference[F])
+    extends BaseRuntimeInterpreter[F](config, welderDao)
+    with DataprocAlgebra[F]
+    with LazyLogging
+    with Retry {
+
+  import dbRef._
+
+  implicit val contextShiftIO = IO.contextShift(blocker.blockingContext)
```
Qi77Qi: why this cs is backed by blockingContext?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/VPCHelper.scala
```diff
@@ -12,33 +14,34 @@ import org.broadinstitute.dsde.workbench.model.google.GoogleProject
...
-                                                                //  log: Logger[IO]
-) {
+class VPCHelper[F[_]: Async: ContextShift: Logger](config: VPCHelperConfig,
+                                                   googleProjectDAO: GoogleProjectDAO,
+                                                   googleComputeService: GoogleComputeService[F],
+                                                   blocker: Blocker) {
 
-  private def getVPCSettingsFromProjectLabel(googleProject: GoogleProject): IO[Option[VPCConfig]] =
-    IO.fromFuture(IO(googleProjectDAO.getLabels(googleProject.value))).map { labelMap =>
+  implicit val contextShiftIO = IO.contextShift(blocker.blockingContext)
```
Qi77Qi: same question, why this needs to be backed by a blocking context? also the global `ContextShift[F]` is in scope via contextbound as well

rtitle: Yeah this probably isn't correct.. basically for `IO.fromFuture(…)` I need a `ContextShift[IO]`. But I only have a `ContextShift[F]`. I couldn't figure out how to create a `ContextShift[IO]` from a `ContextShift[F]` so I defined one in this way using the blocking context.

However now I think it might be better for this class to take an implicit `ContextShift[IO]` in its constructor and have `Boot`/`IOApp` provide it.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
rtitle: So the way I've modeled it is:
- `RuntimeAlgebra` contains all base functionality for runtimes
- `DataprocAlgebra` extends `RuntimeAlgebra` to add Dataproc extensions to the algebra 
- `GceInterpreter` interprets `RuntimeAlgebra`
- `DataprocInterpreter` interprets `DataprocAlgebra`

However there will eventually be GCE extensions to the algebra I think (e.g. for PDs, GPUs, etc) so maybe I will go ahead and make a `GceAlgebra` extending `RuntimeAlgebra`.

rtitle:
jenkins retest

http/src/main/resources/init-resources/gce-init.sh
```diff
@@ -200,7 +200,11 @@ END
 </source>
 END
 
-    service google-fluentd reload
+    # restarting instead of `service google-fluentd-reload` because of bug:
+    # https://github.com/GoogleCloudPlatform/google-fluentd/issues/232
+    service google-fluentd restart
```
Qi77Qi, line 205: so fluentd does actually come with gce instances by default?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -140,20 +139,30 @@ class LeoPubsubMessageSubscriber[F[_]: Async: Timer: ContextShift: Concurrent](
               savedMasterMachineType match {
                 case Some(machineType) =>
                   for {
+                    runtimeConfig <- dbRef.inTransaction(
+                      RuntimeConfigQueries.getRuntimeConfig(resolvedCluster.runtimeConfigId)
+                    )
                     // perform gddao and db updates for new resources
-                    _ <- Async[F].liftIO(clusterHelper.updateMasterMachineType(resolvedCluster, machineType))
-                    now <- Timer[F].clock.realTime(TimeUnit.MILLISECONDS)
+                    _ <- runtimeConfig.cloudService match {
```
Qi77Qi: I wonder if it would make sense to have a type class something like 
```
trait RuntimeAlg[F] {
   def updateMachineType(UpdateMachineTypeParams(resolvedCluster, machineType, now)): F[Unit]
   def startRuntime(StartRuntimeParams(resolvedCluster, now)): F[Unit]
/// other common functions
}
```

And then we can do ```runtimeAlg.updateMachineType(xxx)``` etc

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/BaseRuntimeInterpreter.scala
```diff
@@ -0,0 +1,230 @@
...
+  RuntimeStatus,
+  WelderAction
+}
+import org.broadinstitute.dsde.workbench.model.TraceId
+import org.broadinstitute.dsde.workbench.newrelic.NewRelicMetrics
+
+import scala.concurrent.ExecutionContext
+import scala.util.Try
+
+abstract private[util] class BaseRuntimeInterpreter[F[_]: Async: ContextShift: Logger](
```
Qi77Qi: maybe turn this into a typeclass instead? 🤔

http/src/main/resources/init-resources/gce-init.sh
rtitle: No.. before I installed it I was getting `service not found: google-fluentd`. I think this is a bug in the latest version.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
rtitle: Trying to understand how a typeclass would work here. With a typeclass I guess I'd summon the instances like this:
```
cloudService match {
  case GCE => RuntimeAlg[F, GCE].updateMachineType(..)
  case Dataproc => RuntimeAlg[F, Dataproc].updateMachineType(..)
}
```
That might be a good way to do it instead of passing each interpreter to the constructor. But I don't see how it replaces the abstract class `BaseRuntimeInterpter` which is used for common functionality between GCE and Dataproc. Am I missing something?

rtitle:
> Is `sudo` needed?

Nope, fixed :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
Qi77Qi: maybe not get rid of `BaseRuntimeInterpter `, this is what I was thinking roughly

```
trait RuntimeAlgebra[F[_], A]  {
  def createRuntime(params: CreateRuntimeParams)(implicit ev: ApplicativeAsk[F, TraceId]): F[CreateRuntimeResponse]
  def deleteRuntime(params: DeleteRuntimeParams)(implicit ev: ApplicativeAsk[F, TraceId]): F[Unit]
  def finalizeDelete(params: FinalizeDeleteParams)(implicit ev: ApplicativeAsk[F, TraceId]): F[Unit]
  def stopRuntime(params: StopRuntimeParams)(implicit ev: ApplicativeAsk[F, TraceId]): F[Unit]
  def startRuntime(params: StartRuntimeParams)(implicit ev: ApplicativeAsk[F, TraceId]): F[Unit]
  def updateMachineType(params: UpdateMachineTypeParams)(implicit ev: ApplicativeAsk[F, TraceId]): F[Unit]
  def updateDiskSize(params: UpdateDiskSizeParams)(implicit ev: ApplicativeAsk[F, TraceId]): F[Unit]
}

// instances
object RuntimeInstances {
 def gce[F[_]: Async: ContextShift: Logger: DbReference: NewRelicMetrics](config: RuntimeInterpreterConfig.GceInterpreterConfig,
                                            welderDao: WelderDAO[F], ,
                                                                          gceAlgebra: GceAlgebra[F])(implicit ec: ExecutionContext): RuntimeAlgebra[F, CloudService.GCE.type] = new BaseRuntimeInterpreter(config, welderDao){
   override protected def stopGoogleRuntime(runtime: Runtime, runtimeConfig: RuntimeConfig)(implicit ev: ApplicativeAsk[F, TraceId]): F[Unit] = ???

   override protected def startGoogleRuntime(runtime: Runtime, welderAction: WelderAction, runtimeConfig: RuntimeConfig)(implicit ev: ApplicativeAsk[F, TraceId]): F[Unit] = ???

   override protected def setMachineTypeInGoogle(runtime: Runtime, machineType: MachineTypeName)(implicit ev: ApplicativeAsk[F, TraceId]): F[Unit] = ???

   override def createRuntime(params: CreateRuntimeParams)(implicit ev: ApplicativeAsk[F, TraceId]): F[CreateRuntimeResponse] = ???

   override def deleteRuntime(params: DeleteRuntimeParams)(implicit ev: ApplicativeAsk[F, TraceId]): F[Unit] = ???

   override def finalizeDelete(params: FinalizeDeleteParams)(implicit ev: ApplicativeAsk[F, TraceId]): F[Unit] = ???

   override def updateDiskSize(params: UpdateDiskSizeParams)(implicit ev: ApplicativeAsk[F, TraceId]): F[Unit] = ???
 }
  
  def dataproc[F[_]: Async: ContextShift: Logger: DbReference: NewRelicMetrics](config: RuntimeInterpreterConfig.DataprocInterpreterConfig,
                                                                                welderDao: WelderDAO[F], ,
                                                                         dataprocAlgebra: dataprocAlgebra[F])(implicit ec: ExecutionContext): RuntimeAlgebra[F, CloudService.Dataproc.type] = new BaseRuntimeInterpreter(config, welderDao){
    override protected def stopGoogleRuntime(runtime: Runtime, runtimeConfig: RuntimeConfig)(implicit ev: ApplicativeAsk[F, TraceId]): F[Unit] = ???

    override protected def startGoogleRuntime(runtime: Runtime, welderAction: WelderAction, runtimeConfig: RuntimeConfig)(implicit ev: ApplicativeAsk[F, TraceId]): F[Unit] = ???

    override protected def setMachineTypeInGoogle(runtime: Runtime, machineType: MachineTypeName)(implicit ev: ApplicativeAsk[F, TraceId]): F[Unit] = ???

    override def createRuntime(params: CreateRuntimeParams)(implicit ev: ApplicativeAsk[F, TraceId]): F[CreateRuntimeResponse] = ???

    override def deleteRuntime(params: DeleteRuntimeParams)(implicit ev: ApplicativeAsk[F, TraceId]): F[Unit] = ???

    override def finalizeDelete(params: FinalizeDeleteParams)(implicit ev: ApplicativeAsk[F, TraceId]): F[Unit] = ???

    override def updateDiskSize(params: UpdateDiskSizeParams)(implicit ev: ApplicativeAsk[F, TraceId]): F[Unit] = ???
  }
}
```

http/src/main/resources/init-resources/jupyter-docker-compose-gce.yaml
```diff
@@ -14,7 +14,7 @@ services:
     network_mode: host
     volumes:
       # shared with welder
-      - "${NOTEBOOKS_DIR}":"${NOTEBOOKS_DIR}"
+      - "/work:${NOTEBOOKS_DIR}"
```
Qi77Qi, line 17: is this intentional?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/google/HttpGoogleDataprocDAO.scala
```diff
@@ -412,8 +412,8 @@ class HttpGoogleDataprocDAO(
     }
   }
 
-  private def getOperationUUID(dop: DataprocOperation): UUID =
-    UUID.fromString(dop.getMetadata.get("clusterUuid").toString)
+  private def getOperationUUID(dop: DataprocOperation): String =
```
Qi77Qi: change this to `private def getGoogleId(dop: DataprocOperation): GoogleId`?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
```diff
@@ -140,7 +150,8 @@ class ClusterMonitorActor(
   val google2StorageDAO: GoogleStorageService[IO],
   val dbRef: DbReference[IO],
   val authProvider: LeoAuthProvider[IO],
-  val clusterHelper: ClusterHelper,
+  val dataprocAlg: DataprocAlgebra[IO],
```
Qi77Qi: maybe we can just pass in `RuntimeAlgebra[IO]` here? it doesn't seem like there's anything gce or dataproc specific things going on in this class. Do you foresee dataproc or gce specific ops needed in `ClusterMonitorActor`? 🤔

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -1,51 +1,43 @@
...
+import org.broadinstitute.dsde.workbench.model.{ErrorReport, TraceId, WorkbenchException}
 
 import scala.concurrent.ExecutionContext
+import scala.util.control.NoStackTrace
 
 class LeoPubsubMessageSubscriber[F[_]: Async: Timer: ContextShift: Concurrent](
   subscriber: GoogleSubscriber[F, LeoPubsubMessage],
-  clusterHelper: ClusterHelper
+  dataprocAlg: DataprocAlgebra[F],
+  gceAlg: GceAlgebra[F]
```
Qi77Qi: pass in `RuntimeAlgebra` seems will be sufficient for this class?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/leoPubsubMessageSubscriberModels.scala
```diff
@@ -16,39 +15,39 @@ sealed trait LeoPubsubMessage {
 }
 
 object LeoPubsubMessage {
-  final case class StopUpdate(updatedMachineConfig: RuntimeConfig, clusterId: Long, traceId: Option[TraceId])
+  final case class StopUpdateMessage(updatedMachineConfig: RuntimeConfig, runtimeId: Long, traceId: Option[TraceId])
```
Qi77Qi, line 18: nit and personal preference, the `Message` suffix seems a bit redundant in the context of `LeoPubsubMessage`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -635,20 +626,10 @@ class LeonardoService(
   def internalDeleteCluster(userEmail: WorkbenchEmail,
                             cluster: Runtime)(implicit ev: ApplicativeAsk[IO, TraceId]): IO[Unit] =
     if (cluster.status.isDeletable) {
+      val hasDataprocInfo = cluster.asyncRuntimeFields.isDefined
       for {
-        // Delete the notebook service account key in Google, if present
-        keyIdOpt <- clusterQuery.getServiceAccountKeyId(cluster.googleProject, cluster.runtimeName).transaction
-        _ <- clusterHelper
-          .removeServiceAccountKey(cluster.googleProject, cluster.serviceAccountInfo.notebookServiceAccount, keyIdOpt)
```
Qi77Qi, line 642: is this still needed for dataproc?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -683,30 +664,10 @@ class LeonardoService(
                                   throw403 = true)
 
       runtimeConfig <- RuntimeConfigQueries.getRuntimeConfig(cluster.runtimeConfigId).transaction
-      _ <- clusterHelper.stopCluster(cluster, runtimeConfig)
+      now <- IO(Instant.now)
+      _ <- dataprocAlg.stopRuntime(StopRuntimeParams(RuntimeAndRuntimeConfig(cluster, runtimeConfig), now))
     } yield ()
 
-  def internalStopCluster(cluster: Runtime)(implicit ev: ApplicativeAsk[IO, TraceId]): IO[Unit] =
```
Qi77Qi, line 689: 👍

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -721,8 +682,8 @@ class LeonardoService(
                                   RuntimeProjectAndName(cluster.googleProject, cluster.runtimeName),
                                   throw403 = true)
 
-      now <- timer.clock.realTime(TimeUnit.MILLISECONDS)
-      _ <- clusterHelper.startCluster(cluster, Instant.ofEpochMilli(now))
+      now <- IO(Instant.now)
```
Qi77Qi: I think using `Timer` is better than `Instant`. With `Timer`, we can manipulate the time more easily in unit tests if we decided to check time somewhere

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/BaseRuntimeInterpreter.scala
```diff
@@ -0,0 +1,230 @@
...
+          clusterQuery.updateClusterStatus(updatedRuntime.id, RuntimeStatus.Starting, params.now)
+        }
+      } yield ()
+    } else
+      Async[F].raiseError(
+        RuntimeCannotBeStartedException(params.runtime.googleProject, params.runtime.runtimeName, params.runtime.status)
+      )
+
+  private def getWelderAction(runtime: Runtime): WelderAction =
+    if (runtime.welderEnabled) {
```
Qi77Qi, line 112: minor, wonder if we can remove `NoAction` case, and have this function return `Option[WelderAction]`. This way, caller can do `getWelderAction(???).traverse(action => ???)`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/BaseRuntimeInterpreter.scala
```diff
@@ -0,0 +1,230 @@
...
+        "disableDelocalization" -> (welderAction == DisableDelocalization).toString
+      )
+
+    TemplateHelper
+      .templateResource[F](replacements, config.clusterResourcesConfig.startupScript, blocker)
+      .compile
+      .to[Array]
+      .map { bytes =>
+        Map(googleKey -> new String(bytes, StandardCharsets.UTF_8))
+      }
```
Qi77Qi, line 194: this can be 
```
TemplateHelper
      .templateResource[F](replacements, config.clusterResourcesConfig.startupScript, blocker)
      .through(fs2.text.utf8Decode)
      .compile
      .string
      .map { s =>
        Map(googleKey -> s)
      }
```
`string` operator combines `Stream[F, String]` to a single `String`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/BaseRuntimeInterpreter.scala
```diff
@@ -0,0 +1,230 @@
...
+      config.welderConfig,
+      config.proxyConfig,
+      config.clusterFilesConfig,
+      config.clusterResourcesConfig,
+      None,
+      RuntimeOperation.Stopping
+    )
+    val replacements = RuntimeTemplateValues(templateConfig).toMap
+
+    TemplateHelper
```
Qi77Qi, line 216: same comment as above

http/src/main/resources/init-resources/jupyter-docker-compose-gce.yaml
rtitle: Yeah.. I think it was a typo before, we should be mounting `/work` to `${NOTEBOOKS_DIR}`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
rtitle: So for things like `finalizeDelete` we need to choose the right interpreter depending on whether it's Dataproc or GCE (Dataproc needs to clean up some IAM stuff; GCE doesn't). That's why I'm passing 2 instances here.

How would it look like in the above approach? Maybe:
```
RuntimeInstances.fromCloudService(runtimeConfig.cloudService).finalizeDelete(params)
```
That might be a bit cleaner for callers..

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
rtitle: Same as above, maybe we don't pass in any algebras and use `RuntimeInstances.fromCloudService` to retrieve instances?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/leoPubsubMessageSubscriberModels.scala
rtitle: Hm, I changed it because it wasn't clear to me always that this was a pubsub message since we often `import org.broadinstitute.dsde.workbench.leonardo.monitor.LeoPubsubMessage._`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
rtitle: I moved it to `DataprocInterpreter.deleteRuntime`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/BaseRuntimeInterpreter.scala
rtitle: Hm, I guess `NoAction` could be represented as `None`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
rtitle: I think I see.. want to see how the code looks

http/src/main/resources/init-resources/jupyter-docker-compose-gce.yaml
kyuksel, line 17: Not specific to this PR but a general question: Why are we placing Welder files in the same directory as user files like notebooks? I'd think we'd want to separate files that users may want to view vs. those that are helpful to us for debugging/auditing purposes.

rtitle: Welder needs to have a shared volume with the Jupyter container because of its design -- it syncs notebooks on the filesystem with GCS. It's true that `welder.log` is placed in the same directory but I see that as a benefit as it gives debugging info to the user. I don't think there are any other welder-specific files that are shared (unless I'm forgetting something)..

kyuksel: I can see `welder.log` be useful for power users.

rtitle:
@Qi77Qi I think I applied all feedback and cleaned stuff up, please take another look.

See especially the `RuntimeInstances` class, I added the syntax:
```
cloudService.interpreter.deleteRuntime(params)
```
(similar to how `DbRef` `.transaction` works)

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/runtimeModels.scala
```diff
@@ -85,15 +84,17 @@ object RuntimeStatus extends Enum[RuntimeStatus] {
...
-  val activeStatuses: Set[RuntimeStatus] = Set(Unknown, Creating, Running, Updating, Stopping, Stopped, Starting)
+  val activeStatuses: Set[RuntimeStatus] =
+    Set(Unknown, Creating, Running, Updating, Stopping, Stopped, Starting, Staging)
 
   // Can a user delete this runtime? Contains everything except Creating, Deleting, Deleted.
   val deletableStatuses: Set[RuntimeStatus] = Set(Unknown, Running, Updating, Error, Stopping, Stopped, Starting)
 
   // Non-terminal statuses. Requires monitoring via ClusterMonitorActor.
-  val monitoredStatuses: Set[RuntimeStatus] = Set(Unknown, Creating, Updating, Deleting, Stopping, Starting)
+  val monitoredStatuses: Set[RuntimeStatus] = Set(Unknown, Creating, Updating, Deleting, Stopping, Starting, Staging)
```
Qi77Qi: we're introducing a new status? can we have newline for each of the status and add some comment about what each status means?

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/runtimeModels.scala
```diff
@@ -350,3 +351,19 @@ object VPCConfig {
   final case class VPCNetwork(value: String) extends VPCConfig
   final case class VPCSubnet(value: String) extends VPCConfig
 }
+
+sealed trait RuntimeOperation extends Product with Serializable {
```
Qi77Qi, line 356: is this only used for init script?

Qi77Qi (approved):

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/runtimeModels.scala
rtitle: yes.. could move this model to `RuntimeTemplateValues.scala` I guess

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/runtimeModels.scala
rtitle: Hm, `RuntimeStatus` is kind of overloaded. It's used for both _Leo_ runtime status and _Google_ status for the Dataproc cluster/GCE instance. I added `Staging` because that's a Google status for GCE, but we never actually update the Leo runtime status to `Staging` (we still use `Creating`). So all this change does is prevent `ClusterMonitorActor` from logging "Unknown status" log messages.

I think there's some work to be done to revisit our statuses and monitoring code. I'd prefer to do that in a separate ticket/PR though.

kyuksel: +1 that when we revisit statuses, it'd be clarifying to separate our statuses from Google statuses.

rtitle: Yeah, maybe could be done as part of bullet 2 in https://broadworkbench.atlassian.net/browse/IA-1747

Qi77Qi: yea...I was planning to address this as part of that ticket

kyuksel: 👍 Added a note to that ticket.

PR #1310, opened 2020-03-05
[No Ticket] Fix custom image build

kyuksel:
Was getting this error:

```
startup-script: Error response from daemon: manifest for us.gcr.io/broad-dsp-gcr-public/welder-server:welderContainerName not found: manifest unknown: Failed to fetch "welderContainerName" from request "/v2/broad-dsp-gcr-public/welder-server/manifests/welderContainerName".
```

because the `grep`ing in `Jenkinsfile

```
"cat $configFile | grep $image | tr ':' '\n' | tail -1 | tr -d '\"' | awk '{print \$1}' "
```

was matching the last line containing `welder-server` (`welderContainerName = "welder-server"`). The fix is `grep`ing for `/welder-server` instead. It doesn't feel robust but it's less brittle. Trying to match something more specific like `us.gcr.io/broad-dsp-gcr-public/welder-server` doesn't work because `getVersionMap()`s in `Jenkinsfile` go through some empty `image` strings which causes wrong matches. There seems to be tech debt incurred in those methods but I don't want to invest in refactoring them right now.


Qi77Qi (approved):

rtitle (approved):
Seems reasonable, thanks for fixing

jenkins/Jenkinsfile
```diff
@@ -76,14 +76,16 @@ LinkedHashMap<String,String> getVersionMap(Set<String> imageNames, String[] conf
...
     LinkedHashMap<String,String> versionMap = [:]
 
-    // For example, if $image is 'welder-server', this line would isolate 'a4f0e04' from the line below in a $configFile.
+    // For example, if $image is 'welder-server', the 'script' line below would isolate 'a4f0e04' from the line below in a $configFile.
     // welderDockerImage = "us.gcr.io/broad-dsp-gcr-public/welder-server:a4f0e04"
     // The awk at the end ensures that the line can have comments or other whitespace after the terminating quote.
     // Escaping '$' inside the 'awk' command is to appease Jenkins.
     for (String image: imageNames) {
+        // Without '/' in the regex, 'grep' is more susceptible to ordering in the conf files
+        String imageRegex = "/${image}"
```
jdcanas: can we append ':' to this regex to make it more strict? I'd say we should even ensure there's a 7-char hash after the :, but I don't think that's possible with grep without nasty [locale-dependent logic]((https://unix.stackexchange.com/questions/147582/matching-string-with-a-fixed-number-of-characters-using-grep)). We could at least match any number of lowercase alphanumeric chars?

So something like this (which I tested) `grep -E '/${image}:[[:alnum:]]+'`

jdcanas (approved):

kyuksel:
jenkins retest

PR #1317, opened 2020-03-11
IA-1684  allow data uri

Qi77Qi:
Add `data:` to `script-src`

Screen Shot 2020-03-10 at 8 36 46 PM

---

In all cases:


rtitle (approved):
Approving but I wonder if there are any security issues with including arbitrary `data:` in `'unsafe-eval'`..

kyuksel (commented):
I wonder if `data:` could be a more restricted regex if there is a known pattern that would suffice.

Qi77Qi:
@rtitle @kyuksel updated it to `script-src 'self' data:text/javascript 'unsafe-inline' 'unsafe-eval' https://apis.google.com`

kyuksel:
> updated it to script-src 'self' data:text/javascript 'unsafe-inline' 'unsafe-eval' https://apis.google.com

That looks better but I don't know if it may still pose too much of a security risk. Reading [this thread](https://security.stackexchange.com/questions/94993/is-including-the-data-scheme-in-your-content-security-policy-safe) didn't make me feel good. I wonder if @adrazhi would like to weigh in.

Qi77Qi:
had a conversation with @adrazhi , and we're good to go...created https://broadworkbench.atlassian.net/browse/IA-1763, and https://broadworkbench.atlassian.net/browse/IA-1762 as follow up for hardening CSP.

PR #1318, opened 2020-03-11
IA-1747 refactor zombie monitor to fs2

Qi77Qi:
---

In all cases:


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/Boot.scala
```diff
@@ -213,18 +208,15 @@ object Boot extends IOApp {
...
-        Stream.eval(logger.info(s"starting subscriber ${subscriberConfig.projectTopicName} in boot")) ++ pubsubSubscriber.process
-          .handleErrorWith(
-            error => Stream.eval(logger.error(error)("Failed to initialize message processor in Boot. "))
-          )
-      } else Stream.eval(logger.info(s"Not starting subscriber in boot"))
 
-      val startSubscriber = if (leoExecutionModeConfig.backLeo) {
-        Stream.eval(appDependencies.subscriber.start)
-      } else Stream.eval(IO.unit)
+      // only needed for backleo
```
rtitle: It looks like this is outside of the `if (leoExecutionModeConfig.backLeo) {}` block?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ZombieRuntimeMonitor.scala
```diff
@@ -0,0 +1,169 @@
...
+  googleProjectDAO: GoogleProjectDAO
+)(implicit F: Concurrent[F],
+  metrics: NewRelicMetrics[F],
+  logger: Logger[F],
+  dbRef: DbReference[F],
+  ec: ExecutionContext,
+  cs: ContextShift[IO]) {
+
+  val process: Stream[F, Unit] =
+    (Stream.sleep[F](config.zombieCheckPeriod) ++ Stream.eval(F.delay(println("evaluating")))++ Stream.eval(zombieCheck)).repeat
```
rtitle: maybe `Logger` instead of `println`?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ZombieRuntimeMonitor.scala
```diff
@@ -0,0 +1,169 @@
...
+
+  private[monitor] val zombieCheck: F[Unit] =
+    for {
+      start <- Timer[F].clock.realTime(TimeUnit.MILLISECONDS)
+      uuid <- F.delay(UUID.randomUUID())
+      implicit0(traceId: ApplicativeAsk[F, TraceId]) = ApplicativeAsk.const[F, TraceId](TraceId(uuid)) //we don't care much about traceId in unit tests, hence providing a constant UUID here
```
rtitle: Remove comment, probably

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ZombieRuntimeMonitor.scala
```diff
@@ -0,0 +1,169 @@
...
+  }
+
+  private def handleZombieCluster(runtime: PotentialZombieRuntime, now: Instant): F[Unit] =
+    for {
+      _ <- logger.info(s"Deleting zombie cluster: ${runtime.googleProject} / ${runtime.runtimeName}")
+      _ <- metrics.incrementCounter("zombieClusters")
+      _ <- dbRef.inTransaction {
+        for {
+          _ <- clusterQuery.completeDeletion(runtime.id, now)
+          error = RuntimeError("An underlying resource was removed in Google. Cluster has been marked deleted in Leo.",
```
rtitle: Nit: this message is user facing, maybe use `Runtime` instead of `Cluster`?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ZombieRuntimeMonitor.scala
```diff
@@ -0,0 +1,169 @@
...
+    val milliSecondsSinceClusterCreation: Long = now.toEpochMilli - runtime.auditInfo.createdDate.toEpochMilli
+    // this or'd with the google cluster status gives creating clusters a grace period before they are marked as zombies
+    if (runtime.status == RuntimeStatus.Creating && milliSecondsSinceClusterCreation < config.creationHangTolerance.toMillis) {
+      F.pure(true)
+    } else {
+      val runtimeStatus: F[RuntimeStatus] = runtime.cloudService match {
+        case CloudService.GCE =>
+          gce.getInstance(runtime.googleProject, config.gceZoneName, InstanceName(runtime.runtimeName.asString)).map{
+            instance =>
+              instance.flatMap(s => RuntimeStatus.withNameInsensitiveOption(s.getStatus)).getOrElse(RuntimeStatus.Unknown)
```
rtitle, line 121: Some GCE statuses aren't captured in `RuntimeStatus`. So this will resolve to `Unknown` but I think that should work fine since `Unknown` is in `activeStatuses`..

rtitle:
Some nits and CI isn't happy, but looks good overall -- nice use of fs2.

I feel like automation test isn't required for this since it's more about maintenance than user-critical functionality. Plus the zombie poll period is 4 hours so might be hard to auto-test anyway.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/Boot.scala
Qi77Qi: yea...initializing the object is okay...I can see if moving it inside the if is okay...it was a bit messier previously

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ZombieRuntimeMonitor.scala
Qi77Qi: will remove...we don't need this..was there when I was setting up things in unit test

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ZombieRuntimeMonitor.scala
Qi77Qi: I actually run into this in unit tests...added all gce statuses

Qi77Qi:
@rtitle yea that makes sense...I initially thought it's worth automation test mostly I wasn't sure about statuses returned from google...But I guess it's documented in the lib, so we should be able to trust it

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/runtimeModels.scala
```diff
@@ -73,24 +73,33 @@ sealed trait RuntimeStatus extends EnumEntry
...
+
+  case object Provisioning extends RuntimeStatus // gce
+  case object Staging extends RuntimeStatus // gce
+  case object Suspending extends RuntimeStatus // gce
+  case object Terminated extends RuntimeStatus // gce
 
   // A user might need to connect to this notebook in the future. Keep it warm in the DNS cache.
   val activeStatuses: Set[RuntimeStatus] =
-    Set(Unknown, Creating, Running, Updating, Stopping, Stopped, Starting)
+    Set(Unknown, Creating, Running, Updating, Stopping, Stopped, Starting, Staging)
```
rtitle: Add `Provisioning`? 🤔

rtitle (approved):
:shipit:

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/runtimeModels.scala
Qi77Qi: updated..I also updated things like `startableStatuses`, `updatableStatuses`...those probably will be covered in your stop/resume PR..but I guess it doesn't hurt to add them here

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ZombieRuntimeMonitor.scala
```diff
@@ -0,0 +1,169 @@
...
+import org.broadinstitute.dsde.workbench.model.TraceId
+import org.broadinstitute.dsde.workbench.model.google.GoogleProject
+import org.broadinstitute.dsde.workbench.newrelic.NewRelicMetrics
+
+import scala.concurrent.ExecutionContext
+
+/**
+ * This monitor periodically sweeps the Leo database and checks for clusters which no longer exist in Google.
+ */
+class ZombieClusterMonitor[F[_]: Parallel: ContextShift: Timer](
```
Qi77Qi, line 31: class ZombieClusterMonitor[F[_]: Parallel: ContextShift: Timer]()  ----> context bound

class ZombieClusterMonitor[F[_]](implicit pf: Parallel[F], cs: ContextShift[F], t: Timer[F])

PR #1319, opened 2020-03-11
PROD-364 build correct subnet uri

rtitle:
VPC issue round 2: https://broadworkbench.atlassian.net/browse/PROD-364

Basically I think we were building the wrong URI when using subnets when creating the notebooks firewall rule. Again, only AoU uses subnets in high security projects. Terra uses the default network but there is a desire to start using dedicated subnets as well.

This only affected AoU projects created after 3/5 because firewall setup is a one-time project action.

Added a unit test for this but we should really think about adding an automation test that uses VPC-SC.

---

In all cases:

### DIFF
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -153,6 +153,9 @@ proxy {
   networkTag = "leonardo"
   projectVPCNetworkLabel = "vpc-network-name"
   projectVPCSubnetLabel = "vpc-subnetwork-name"
+  # Subnets are regional and independent of the runtime region. Networks on the other hand are global.
+  # See https://cloud.google.com/vpc/docs/vpc
+  projectVPCSubnetRegion = "us-central1"
   dnsPollPeriod = 15 seconds
   tokenCacheExpiryTime = 60 minutes
   tokenCacheMaxSize = 10000
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -39,6 +39,7 @@ dataproc {
 }
 
 gce {
+  region = "us-central1"
   customGceImage = "projects/broad-dsp-gcr-public/global/images/custom-leo-image-gce-debian9-2d78533"
   defaultScopes = [
     "https://www.googleapis.com/auth/userinfo.email",
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -154,9 +154,6 @@ proxy {
   networkTag = "leonardo"
   projectVPCNetworkLabel = "vpc-network-name"
   projectVPCSubnetLabel = "vpc-subnetwork-name"
-  # Subnets are regional and independent of the runtime region. Networks on the other hand are global.
-  # See https://cloud.google.com/vpc/docs/vpc
-  projectVPCSubnetRegion = "us-central1"
   dnsPollPeriod = 15 seconds
   tokenCacheExpiryTime = 60 minutes
   tokenCacheMaxSize = 10000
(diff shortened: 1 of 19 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/VPCHelper.scala
```diff
@@ -55,13 +55,20 @@ class VPCHelper(config: VPCHelperConfig,
       )
       .build
 
-  private def networkUri(googleProject: GoogleProject, vpcConfig: VPCConfig): String =
-    s"projects/${googleProject.value}/global/networks/${vpcConfig.value}"
+  def buildNetworkUri(googleProject: GoogleProject, vpcConfig: VPCConfig): String =
```
rtitle: The crux of this fix -- build the right uri for the network or subnet.

This wasn't an issue in the old `com.google.api.services.compute` library -- but it seems like `com.google.cloud.compute.v1` is more strict.

http/src/main/resources/reference.conf
```diff
@@ -153,6 +153,9 @@ proxy {
   networkTag = "leonardo"
   projectVPCNetworkLabel = "vpc-network-name"
   projectVPCSubnetLabel = "vpc-subnetwork-name"
+  # Subnets are regional and independent of the runtime region. Networks on the other hand are global.
```
kyuksel: Does ^^ contradict

>Because subnets are regional objects, the region that you select for a resource determines the subnets that it can use 

from the link below?

http/src/main/resources/reference.conf
```diff
@@ -153,6 +153,9 @@ proxy {
...
+  # See https://cloud.google.com/vpc/docs/vpc
+  projectVPCSubnetRegion = "us-central1"
```
Qi77Qi: does this region apply to all subnet configs?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/VPCHelper.scala
kyuksel:
```suggestion
  private def buildNetworkUri(googleProject: GoogleProject, vpcConfig: VPCConfig): String =
```

http/src/main/resources/reference.conf
rtitle: Yeah it's globally set in Leo (as is the Dataproc/GCE region). If/when we support other regions we might want to start choosing this dynamically.

kyuksel (approved):
Looks good 👍 Minor comment and suggestion inline.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/VPCHelper.scala
Qi77Qi: I would do `private[util]` and test the logic in unit test

Qi77Qi (approved):

http/src/main/resources/reference.conf
rtitle: Ugh yeah good catch. Maybe instead of this config it should just use the region of the cluster/instance.

kyuksel: Using the region of the resource sounds more ideal. Though, I'd think the current fix will be fine for now since we're creating resources only in `us-central1` but it may break if we start allowing other regions in the future.

rtitle: I think I will update the PR to do that

rtitle:
Just updated the PR based on feedback @kyuksel @Qi77Qi (thanks)

http/src/main/resources/reference.conf
```diff
@@ -39,6 +39,7 @@ dataproc {
 }
 
 gce {
+  region = "us-central1"
```
kyuksel, line 42: Should we rename `dataprocDefaultRegion` as `region` as well so they are consistent? Also, we only support one region so `default` in the name might be misleading.

kyuksel (approved):
👍 with a nit

http/src/main/resources/reference.conf
rtitle: That makes sense.. I'm going to have to rebase https://github.com/DataBiosphere/leonardo/pull/1309 anyway so I can do it there

kyuksel: 👍

PR #1320, opened 2020-03-11
Fix firewall creation by passing network

rtitle:
I manually deployed this to dev in order to test with AoU… it seems to work.

---

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/VPCHelper.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/VPCHelper.scala
@@ -4,7 +4,7 @@ import cats.effect.{ContextShift, IO}
 import cats.mtl.ApplicativeAsk
 import com.google.cloud.compute.v1.{Allowed, Firewall}
 import org.broadinstitute.dsde.workbench.google.GoogleProjectDAO
-import org.broadinstitute.dsde.workbench.google2.{FirewallRuleName, GoogleComputeService, RegionName}
+import org.broadinstitute.dsde.workbench.google2.{FirewallRuleName, GoogleComputeService}
 import org.broadinstitute.dsde.workbench.leonardo.{NetworkTag, VPCConfig}
 import org.broadinstitute.dsde.workbench.leonardo.VPCConfig.{VPCNetwork, VPCSubnet}
 import org.broadinstitute.dsde.workbench.model.TraceId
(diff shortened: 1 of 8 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/VPCHelper.scala
```diff
@@ -16,40 +16,43 @@ class VPCHelper(config: VPCHelperConfig,
...
     for {
-      rule <- googleComputeService.getFirewallRule(googleProject, config.firewallRuleName)
-      _ <- rule.fold(
-        googleComputeService.addFirewallRule(googleProject, buildFirewall(googleProject, regionName, vpcConfig))
-      )(
-        _ => IO.unit
-      )
+      ruleOpt <- googleComputeService.getFirewallRule(googleProject, config.firewallRuleName)
+      _ <- if (ruleOpt.isEmpty) {
+        // Firewall rules don't work with a subnet, so we have to pass a network. This is
```
rtitle, line 38: The issue in the last PR is that firewall rules require a network, not a subnet… sigh

Qi77Qi (approved):

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/VPCHelper.scala
Qi77Qi: so what was the initial bug before last PR?

rtitle: I'm a bit confused how it used to work before we used `VPCHelper`/`GoogleComputeService`. Looking at old code, it seemed we passed the subnet to the firewall creation request in `HttpGoogleComputeDAO` (which is why I initially did that in my last PR). It's possible that the old code wasn't correct either and it was working by accident somehow.

kyuksel: This is so tricky 😞

kyuksel (approved):
Thanks for fixing this. Given how tricky it is and how much trouble it's created, we should probably try to do more regression testing around it. I wonder if [Firewall Rules Logging](https://cloud.google.com/vpc/docs/firewall-rules-logging) may help too.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/VPCHelper.scala
rtitle: It really is.. before the post-mortem I'll try and better understand this change in behavior.

rtitle:
> Given how tricky it is and how much trouble it's created, we should probably try to do more regression testing around it. I wonder if Firewall Rules Logging may help too.

Yeah, I would say this definitely highlighted a big testing gap. That link is interesting, I wonder if we could do more monitoring of notebooks traffic at the network level.

PR #1321, opened 2020-03-13
IA-1732 Add GCE routes

rtitle:
Added to `RuntimeRoutes`:
- get runtime
- list runtimes
- delete runtime
- stop runtime
- start runtime

Additionally for stop/start/delete I also added PubSub messaging so the Google operations are done by back Leo.

TODO:
- update runtime (patch endpoint -- may do this in a separate PR)
- ~fill in missing unit tests~
- "real" testing in a fiab/automation test

---

In all cases:


core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dataprocModels.scala
```diff
@@ -7,8 +7,6 @@ import enumeratum.{Enum, EnumEntry}
 import org.broadinstitute.dsde.workbench.google2.{InstanceName, ZoneName}
 import org.broadinstitute.dsde.workbench.model.google.GoogleProject
 
-final case class CustomDataprocImage(asString: String) extends AnyVal
```
rtitle, line 10: Moved `CustomDataprocImage` and `GceCustomImage` to an ADT in `runtimeModels`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GceInterpreter.scala
```diff
@@ -156,7 +156,7 @@ class GceInterpreter[F[_]: Async: Parallel: ContextShift: Logger](
                                               OperationName(operation.getName),
                                               stagingBucketName,
                                               None)
-    } yield CreateRuntimeResponse(asyncRuntimeFields, initBucketName, None, CustomDataprocImage("foo"))
```
rtitle, line 159: Whoops, missed this in my last PR

rtitle:
Added unit tests, coverage now above 70%. I think this PR is good to review @Qi77Qi, I won't be adding to it.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/RuntimeRoutes.scala
```diff
@@ -57,20 +133,77 @@ class RuntimeRoutes(runtimeService: RuntimeService[IO], userInfoDirectives: User
...
       )
     } yield StatusCodes.Accepted
+
+  private[api] def getRuntimeHandler(userInfo: UserInfo,
+                                     googleProject: GoogleProject,
+                                     runtimeName: RuntimeName): IO[ToResponseMarshallable] =
+    for {
+      traceId <- IO(UUID.randomUUID())
+      now <- nowInstant
+      implicit0(context: ApplicativeAsk[IO, RuntimeServiceContext]) = ApplicativeAsk.const[IO, RuntimeServiceContext](
```
Qi77Qi: maybe we can create a helper for creating the context

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -150,7 +154,13 @@ class LeoPubsubMessageSubscriber[F[_]: Async: Timer: ContextShift: Concurrent](
                     // start cluster
                     _ <- runtimeConfig.cloudService.interpreter.startRuntime(StartRuntimeParams(resolvedCluster, now))
                     // clean-up info from follow-up table
-                    _ <- dbRef.inTransaction { followupQuery.delete(message.runtimeFollowupDetails) }
+                    _ <- dbRef.inTransaction {
+                      followupQuery.delete(message.runtimeFollowupDetails) >> clusterQuery.updateClusterStatus(
```
Qi77Qi: I think we don't need to delete record from `followupQuery` anymore?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -205,6 +217,58 @@ class LeoPubsubMessageSubscriber[F[_]: Async: Timer: ContextShift: Concurrent](
...
+
+  private[monitor] def handleDeleteRuntimeMessage(msg: DeleteRuntimeMessage, now: Instant)(
+    implicit traceId: ApplicativeAsk[F, TraceId]
+  ): F[Unit] =
+    for {
+      runtimeOpt <- clusterQuery.getClusterById(msg.runtimeId).transaction
+      runtime <- runtimeOpt.fold(
+        Async[F].raiseError[Runtime](PubsubHandleMessageError.ClusterNotFound(msg.runtimeId, msg))
+      )(Async[F].pure)
+      _ <- if (runtime.status != RuntimeStatus.Deleting)
```
Qi77Qi, line 229: just curious, when do we set runtime's state to `Deleting`?

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriberSpec.scala
```diff
@@ -157,6 +157,23 @@ class LeoPubsubMessageSubscriberSpec
     }
   }
 
+  // TODO
```
Qi77Qi: shall we add these todo?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/leoPubsubMessageSubscriberModels.scala
```diff
@@ -104,13 +116,25 @@ object LeoPubsubCodec {
...
     for {
       messageType <- message.downField("messageType").as[String]
       value <- messageType match {
         case "stopUpdate"         => message.as[StopUpdateMessage]
         case "transitionFinished" => message.as[RuntimeTransitionMessage]
         case "createRuntime"      => message.as[CreateRuntimeMessage]
+        case "deleteRuntime"      => message.as[DeleteRuntimeMessage]
+        case "stopRuntime"        => message.as[StopRuntimeMessage]
+        case "startRuntime"       => message.as[StartRuntimeMessage]
         case other                => Left(DecodingFailure(s"invalid message type: $other", List.empty))
```
Qi77Qi: not in your change, but I think we should get rid of this `other` case...

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -669,9 +683,21 @@ class LeonardoService(
...
+      else
+        IO.raiseError[Unit](RuntimeCannotBeStoppedException(cluster.googleProject, cluster.runtimeName, cluster.status))
+
+      // Update the cluster status to Stopping in the DB
+      now <- nowInstant
+      _ <- clusterQuery.setToStopping(cluster.id, now).transaction
+
+      // stop the runtime
+      traceId <- ev.ask
+      _ <- publisherQueue.enqueue1(StopRuntimeMessage(cluster.id, Some(traceId)))
```
Qi77Qi, line 700: will this impact ui in anyway? since the api will be changed from sync to async?

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/CommonTestData.scala
```diff
@@ -226,10 +226,10 @@ object CommonTestData {
     proxyUrl = Runtime.getProxyUrl(proxyUrlBase, project, name1, Set(jupyterImage), Map.empty),
     status = RuntimeStatus.Unknown,
     labels = Map(),
-    jupyterExtensionUri = Some(GcsPath(GcsBucketName("bucketName"), GcsObjectName("extension"))),
-    jupyterUserScriptUri = Some(UserScriptPath.Gcs(GcsPath(GcsBucketName("bucketName"), GcsObjectName("userScript")))),
+    jupyterExtensionUri = Some(GcsPath(GcsBucketName("bucket-name"), GcsObjectName("extension"))),
```
Qi77Qi, line 232: any specific reason for this change?

Qi77Qi (commented):
some questions and minor comments...lgtm overall 👍

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
rtitle: Hm why not? I didn't change any behavior w.r.t. `followupQuery` here. I added the `updateClusterStatus` calls because `RuntimeService.start/stopRuntime` no longer updates the runtime status.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
rtitle: So front-leo delete/stop/start runtime sets the DB status to `Deleting`/`Stopping`/`Starting` and publishes a message. The pubsub subscriber (back-Leo) handles the message asynchronously.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/leoPubsubMessageSubscriberModels.scala
rtitle: Why? Since it's matching a `String` I thought we should leave it there.. wouldn't we get a `MatchError` otherwise?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
Qi77Qi: gotcha..so same way as creating

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
rtitle: I don't think so.. Stopping was always async, now we just do the Google calls in back leo instead of front Leo. The API and DB behavior should be the same.

I am testing with Terra UI anyway though..

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
Qi77Qi: I think the original reason why we need to do this clean up is we're doing `insert` when adding `followup` record, but I changed it to `upsert` at one point, and removed this clean up...

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriberSpec.scala
rtitle: Yeah, I was struggling with how to test those because they just call `RuntimeService` methods and return `IO[Unit]`. Feels like I would need to set up mocks and capture the calls to them somehow.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
rtitle: yup

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/leoPubsubMessageSubscriberModels.scala
Qi77Qi: oh yea...my bad..I thought it's matching an object...maybe worth adding an ADT for `MessageType`...

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/CommonTestData.scala
rtitle: Yeah, the route tests were encoding/decoding runtime requests and this was failing bucket name validation because capital letters aren't allowed:

https://github.com/broadinstitute/workbench-libs/blob/develop/model/src/main/scala/org/broadinstitute/dsde/workbench/model/google/GcsPathParser.scala#L15

I think `HttpRoutesSpec` was the first time we were decoding JSON -> model objects using the circe decoders.

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriberSpec.scala
Qi77Qi: I think we can also just check some logic branching is working as expected...like for `handleDeleteRuntimeMessage`, we can check the `ClusterInvalidState` path

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/CommonTestData.scala
Qi77Qi: ah..that's a cool...

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
rtitle: Ah gotcha.. I can remove this then

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriberSpec.scala
rtitle: sounds good, will add

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/leoPubsubMessageSubscriberModels.scala
rtitle: good idea

rtitle:
@Qi77Qi I think I addressed feedback, also added the new routes to swagger

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -239,6 +526,9 @@ paths:
             - openid
             - email
             - profile
+
+## Legacy Cluster API ##
+
   "/api/cluster/v2/{googleProject}/{clusterName}":
```
Qi77Qi, line 531: shall we add `deprecated: true` to this route?

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -63,14 +63,129 @@ paths:
               schema:
                 $ref: "#/components/schemas/LeonardoVersion"
       security: []
+
+## Runtime paths ##
+
+  /api/runtimes:
```
Qi77Qi: should this be `api/google/v1/runtimes`? also, shall we rename the other path to `api/google/v1/runtimes` as well as opposed to singular form? I think by convention, the resource collection is usually plural.

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -63,14 +63,129 @@ paths:
...
+            default: false
+      responses:
+        "200":
+          description: List of runtimes
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: "#/components/schemas/ListClusterResponse"
```
Qi77Qi, line 104: I kinda wanna expose different objects just so we don't have to carry over fields that don't make sense....but we can do that in a follow up PR

http/src/main/resources/swagger/api-docs.yaml
rtitle: Let's wait to do that until we're actually confident in the new routes - maybe after we switch automation tests over

http/src/main/resources/swagger/api-docs.yaml
rtitle: Yeah makes sense.. good catch

Qi77Qi (approved):

http/src/main/resources/swagger/api-docs.yaml
rtitle: Yeah maybe do that in a separate PR.. because currently in Leo code we are using the same objects/serializers for both paths

rtitle:
fiab testing looks good, Terra UI works and I called all the new swagger routes :) Will merge once tests go green.

kyuksel:
Great work!

PR #1324, opened 2020-03-17
[no ticket] Fix RuntimeImageType enum and add DB migration

rtitle:
Fix issue on dev with existing clusters:

image

---

In all cases:

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/runtimeModels.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/runtimeModels.scala
@@ -8,7 +8,7 @@ import enumeratum.{Enum, EnumEntry}
 import org.broadinstitute.dsde.workbench.google2.MachineTypeName
 import org.broadinstitute.dsde.workbench.leonardo.DataprocRole.SecondaryWorker
 import org.broadinstitute.dsde.workbench.leonardo.RuntimeContainerServiceType.JupyterService
-import org.broadinstitute.dsde.workbench.leonardo.RuntimeImageType.{Custom, Jupyter, RStudio, Welder}
+import org.broadinstitute.dsde.workbench.leonardo.RuntimeImageType.{Jupyter, RStudio, VM, Welder}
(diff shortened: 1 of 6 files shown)


core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/runtimeModels.scala
```diff
@@ -246,7 +246,7 @@ object RuntimeImageType extends Enum[RuntimeImageType] {
   case object Jupyter extends RuntimeImageType
   case object RStudio extends RuntimeImageType
   case object Welder extends RuntimeImageType
-  case object Custom extends RuntimeImageType
+  case object VM extends RuntimeImageType
```
rtitle, line 249: I figure `RuntimeImageType.VM` is a better name than `RuntimeImageType.Custom` since it represents a VM image.

rtitle:
I have a Stopped cluster on dev showing this issue; I can validate after merge.

akarukappadath (approved):
lgtm!

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/runtimeModels.scala
kyuksel: I agree that `RuntimeImageType.VM` is a better name. 

I also wonder if we should distinguish between `VM` and `Docker` images for clarity. Perhaps there could be a `ImageType` as grandparent which could have children `VMImage` and `DockerImageType` and then `Jupyter`, `RStudio`, etc. could be the grandchildren.

kyuksel (approved):
👍 one question/comment inline

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/runtimeModels.scala
rtitle: Yeah I never liked how this enum mixes Docker and VM images. This might be some tech debt as there are other similar models which also track image types -- maybe they could be consolidated. I'd prefer to do it outside of this PR though.

rtitle:

PR #1328, opened 2020-03-20
[IA-1765] Use AppSec's Playbook for GCE Image Hardening

kyuksel:
[JIRA ticket](https://broadworkbench.atlassian.net/browse/IA-1765)

The main change is due to that `prepare_custom_leonardo_gce_image.sh` no longer does the image hardening directly but instead calls another script that does it. The latter script lives in a new repo that the AppSec team has created and will maintain. Since it will probably be an important part of our process going forward and it's not a big repo, it felt right to submodule it within the Leo repo.

The updated flow to generate GCE VM images is now as follows:
- Jenkins runs a Daisy container
- Daisy spins up a VM and uses the `prepare_custom_leonardo_gce_image` script to install things on the VM
- except for image hardening which is handled by the new `harden-images.sh` script that `prepare_custom_leonardo_gce_image.sh` calls


.gitmodules
```diff
@@ -1,3 +1,6 @@
 [submodule "jenkins/dataproc-custom-images/dataproc-custom-images"]
 	path = jenkins/dataproc-custom-images/dataproc-custom-images
 	url = https://github.com/GoogleCloudPlatform/dataproc-custom-images.git
+[submodule "jenkins/gce-custom-images/cis-harden-images"]
+	path = jenkins/gce-custom-images/cis-harden-images
+	url = https://github.com/broadinstitute/CIS-harden-images.git
```
kyuksel: This entry is created (along with a local one in `.git/config`) when we manually run `git submodule add ...` (more info [here](https://broadworkbench.atlassian.net/wiki/spaces/IA/pages/104399223/Callisto+Developer+Handbook#CallistoDeveloperHandbook-AddingaGitRepoasSubmodule)). Something mysterious happened around git auth'ing on the Jenkins node which caused an error and then went away after changing the `https://` url to `git@github`. However,  `.gitmodule` and `.git/config` on the Jenkins node was still showing `https://` for both of the submodules so I'm not convinced that was the issue. Something non-idempotent must be going on. Will keep an eye on it.

jenkins/Jenkinsfile
```diff
@@ -140,6 +140,8 @@ pipeline {
             description: "The path for the bucket where the custom GCE images will be stored")
         string(name: "DATAPROC_VERSIONS", defaultValue: "1.2.79-debian9 1.4.15-debian9",
             description: "A custom image will be build for each of these dataproc versions")
+        string(name: "IMAGE_HARDENING_LIB_VERSION", defaultValue: "8277c6e",
+            description: "The hash of the 'broadinstitute/CIS-harden-images' repo branch that contains libraries for CIS-hardening GCE images")
```
kyuksel: This parameter will allow us to point to newer versions of the hardening library if the AppSec team makes any updates to it.

jenkins/Jenkinsfile
```diff
@@ -340,6 +342,15 @@ pipeline {
             }
     }
 
+    stage('Init git submodules') {
+        when { expression { !shouldAbortBuild } }
+        steps {
+            sh """
+                git submodule update --init --recursive
```
kyuksel: This is the same as `git submodule init && git submodule update`. I couldn't find a way to insert it before the forking happens within the next `stage` so had to create a new `stage`  😞

.gitmodules
jdcanas: more modules == more fun :)

jenkins/Jenkinsfile
jdcanas: I think this is fine, no problem with more separation of code/logs in this massive job

jenkins/Jenkinsfile
```diff
@@ -385,24 +392,28 @@ pipeline {
...
                                 sh """
+                                    cd jenkins/gce-custom-images/cis-harden-images
+                                    git checkout $IMAGE_HARDENING_LIB_VERSION 
+
                                     # Create the Daisy scratch bucket if it doesn't exist. The Daisy workflow will clean it up at the end.
                                     gsutil ls $GCE_IMAGE_BUCKET || gsutil mb -p $GOOGLE_PROJECT -l $REGION $GCE_IMAGE_BUCKET
 
                                     docker pull $daisyImage
-                                    docker run -i --rm -v ${WORKSPACE}/jenkins/gce-custom-images:/daisy_source_files \
+                                    docker run -i --rm -v ${WORKSPACE}/jenkins/gce-custom-images:/gce-custom-images \
```
jdcanas, line 402: if this gets much bigger it may be worth thinking of consolidating it along with the docker run command in `create_gce_image.sh`. There seems to be a bit of duplication

jdcanas (approved):

kyuksel:
jenkins retest

jenkins/Jenkinsfile
kyuksel: Yeah, that's a good idea.

jenkins/Jenkinsfile
```diff
@@ -193,7 +195,7 @@ pipeline {
 
     stage('Pull Leonardo') {
         steps {
-            git credentialsId: 'jenkins-ssh-github', url: 'git@github.com:DataBiosphere/leonardo.git', branch: 'rt-add-algebra-2'  // TODO revert me
+            git credentialsId: 'jenkins-ssh-github', url: 'git@github.com:DataBiosphere/leonardo.git', branch: 'ky_ia_1765_image_hardening_using_appsec_libraries'  // TODO revert me
```
rtitle: Oops did I leave this by mistake? Reminder to change to `develop` before merge.

rtitle (approved):

jenkins/Jenkinsfile
kyuksel: No worries. It's too easy to forget. I probably will too 😝

PR #1329, opened 2020-03-21
[IA-1754] GCE: Create dedicated VPC subnet instead of using default

rtitle:
JIRA: https://broadworkbench.atlassian.net/browse/IA-1754
Depends on: https://github.com/broadinstitute/workbench-libs/pull/288/

Implement VPC security best practices as documented here: https://docs.google.com/document/d/1adV0LC2f_GIpl3A1AeoQuNiwcP59NToIt6VYT3xRCkU/edit

Summary of changes:
* We now create VMs in a dedicated Leo VPC instead of the default network
* We now restrict ssh access to VMs to Broad-internal IPs
* High security project (read: AoU) behavior is unchanged

---

In all cases:


rtitle:
jenkins retest

rtitle:
jenkins retest, looks like unrelated fiab-start issue

rtitle:
transient failure? jenkins retest

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/Config.scala
```diff
@@ -87,6 +91,39 @@ object Config {
...
+      config.as[NetworkName]("network"),
+      config.as[List[IpRange]]("sourceRanges"),
+      config.as[List[Allowed]]("allowed")
+    )
+  }
+
+  implicit val vpcConfigReader: ValueReader[VPCConfig] = ValueReader.relative { config =>
+    VPCConfig(
+      config.getString("highSecurityProjectNetworkLabel"),
+      config.getString("highSecurityProjectSubnetworkLabel"),
```
Qi77Qi: shall we have custom data type for `NetworkLabel` vs `SubnetWorkLabel`?

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/CommonTestData.scala
```diff
@@ -88,13 +93,25 @@ object CommonTestData {
...
   val monitorConfig = config.as[MonitorConfig]("monitor")
   val clusterBucketConfig = config.as[ClusterBucketConfig]("clusterBucket")
-  val vpcHelperConfig = VPCHelperConfig(
-    proxyConfig.projectVPCNetworkLabel,
-    proxyConfig.projectVPCSubnetLabel,
-    FirewallRuleName(proxyConfig.firewallRuleName),
-    proxyConfig.proxyProtocol,
-    proxyConfig.proxyPort,
-    List(NetworkTag(proxyConfig.networkTag))
+  val vpcInterpreterConfig = VPCInterpreterConfig(vpcConfig)
```
Qi77Qi: are these needed or we can just use `Config.vpcInterpreterConfig`? `Config` is reading from configuration file anyway

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/Config.scala
rtitle: I could.. but these are label keys which to me are semantically `String`s anyway.

http/src/main/resources/reference.conf
```diff
@@ -56,6 +56,66 @@ gce {
...
+vpc {
+  highSecurityProjectNetworkLabel = "vpc-network-name"
+  highSecurityProjectSubnetworkLabel = "vpc-subnetwork-name"
+  networkName = "leonardo-network"
+  networkTag = "leonardo"
+  autoCreateSubnetworks = false
+  # note the below 3 fields are not used if autoCreateSubnetworks is true
+  subnetworkName = "leonardo-subnetwork"
+  subnetworkRegion = "us-central1"
+  subnetworkIpRange = "10.128.0.0/20"
```
kyuksel: Where was the IP range specified?

http/src/main/resources/reference.conf
```diff
@@ -56,6 +56,66 @@ gce {
...
+      network = ${vpc.networkName}
+      sourceRanges = ["0.0.0.0/0"]
+      allowed = [
+        {
+          protocol = "tcp"
+          port = "443"
+        }
+      ]
+    },
+    # allows dataproc nodes to communicate with each other via internal IP
```
kyuksel: Dataproc nodes within the same cluster, I assume? What ensures that?

http/src/main/resources/reference.conf
```diff
@@ -56,6 +56,66 @@ gce {
...
+      sourceRanges= ["69.173.112.0/21", "69.173.127.232/29", "69.173.127.128/26", "69.173.127.0/25", "69.173.127.240/28", "69.173.127.224/30", "69.173.127.230/31", "69.173.120.0/22", "69.173.127.228/32", "69.173.126.0/24", "69.173.96.0/20", "69.173.64.0/19", "69.173.127.192/27", "69.173.124.0/23"]
+      allowed = [
+        {
+          protocol = "tcp"
+          port = 22
+        }
+      ]
+    }
+  ]
+  firewallsToRemove = ["default-allow-rdp", "default-allow-ssh", "leonardo-notebooks-rule"]
```
kyuksel, line 121: I also see `default-allow-icmp` on the [list](https://cloud.google.com/vpc/docs/firewalls#more_rules_default_vpc). Do we care about that?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/VPCInterpreter.scala
```diff
@@ -0,0 +1,161 @@
...
+
+final case class SubnetworkNotReadyException(project: GoogleProject, subnetwork: SubnetworkName)
+    extends LeoException(
+      s"Subnetwork ${subnetwork.value} in project ${project.value} not ready within the specified time"
+    )
+
+final case class FirewallNotReadyException(project: GoogleProject, firewall: FirewallRuleName)
+    extends LeoException(s"Firewall ${firewall.value} in project ${project.value} not ready within the specified time")
+
+class VPCInterpreter[F[_]: Async: Parallel: ContextShift: Logger](
```
kyuksel: Make it `final` if we don't expect `VPCInterpreter` to be extended?

kyuksel (commented):
Looks good overall. Questions inline. Also,

1. Should we have an integration test for scenarios like the one that cause issues for AoU recently?
2. Should we have eyes from the AppSec team regarding the security settings?

http/src/main/resources/reference.conf
rtitle: This is the ip range when creating the subnet. It can be any ip range but I chose the range for the `us-central1` subnet in [auto mode](https://cloud.google.com/vpc/docs/vpc#ip-ranges).

By the way, I experimented with auto mode and custom mode subnet creation ([docs](https://cloud.google.com/vpc/docs/vpc#vpc_networks_and_subnets)). Either way works -- but this PR is using custom mode because:
* we only need subnets in 1 region (auto mode creates one in _every_ region)
* we have control over the IP range which is useful for creating fw rules
* [these docs](https://cloud.google.com/vpc/docs/vpc#auto-mode-considerations) recommend using custom mode for production

Maybe I'll synthesize this in a comment in the `reference.conf` file..

http/src/main/resources/reference.conf
rtitle: Only nodes within the same cluster _need_ to communicate with each other, but this fw rule allows internal communication over the whole `10.128.0.0/20` address space (which is the same way the default network is set up).

http/src/main/resources/reference.conf
rtitle: I am not sure, I guess that enables `ping`. I was going by [this doc](https://docs.google.com/document/d/1adV0LC2f_GIpl3A1AeoQuNiwcP59NToIt6VYT3xRCkU/edit) which doesn't mention icmp traffic so I'd just as soon not touch it.

rtitle:
> Should we have an integration test for scenarios like the one that cause issues for AoU recently?

Yes definitely. I think that may be outside the scope of this PR though, as it requires changes to our automation infrastructure to create high-security gpalloc projects. I think we have a separate ticket for it.

> Should we have eyes from the AppSec team regarding the security settings?

Sure. cc @adrazhi, this PR attempts to make Leo network configuration conform to: https://docs.google.com/document/d/1adV0LC2f_GIpl3A1AeoQuNiwcP59NToIt6VYT3xRCkU/edit. See `reference.conf` and `VPCInterpreter.scala` for main changes.

rtitle:
@Qi77Qi @kyuksel applied PR feedback so far

Qi77Qi (commented):
👍 lgtm...ideally replace the snap version with published version

rtitle:
Will do.. will make sure things are working then merge wb-libs and update this PR.

http/src/main/resources/reference.conf
kyuksel: I see. The comment mentions `dataproc nodes`. Does this not apply to GCE nodes?

kyuksel (approved):
Looks good 👍

http/src/main/resources/reference.conf
rtitle: It applies to GCE as well (anything on the network) but it’s only a requirement for dataproc. I can adjust the comment to make this more clear.

rtitle:
jenkins retest

rtitle:
a lot of fiab-start issues today jenkins retest

rtitle:
jenkins retest

PR #1330, opened 2020-03-22
[IA-1681] isUpdating flag on runtimes during patch

akarukappadath:

---

In all cases:


http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20200320_is_updating.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="leonardo" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+    <changeSet logicalFilePath="leonardo" author="azmina" id="is_updating">
+        <addColumn tableName="CLUSTER">
```
rtitle: Could this be:
```
<addColumn tableName="CLUSTER">
   <column name="isUpdating" type="boolean" defaultValue="false">
      <constraints nullable="false" />
   </column>
 </addColumn>
```
Then we don't need the 2 changeSets below?

akarukappadath: Would that take care of previous clusters that don't have this column? I was under the impression that the default value was for future rows of the database

rtitle: I think adding a default populates all existing null values… haven't tested with liquibase though.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/LeoRoutesSprayJsonCodec.scala
```diff
@@ -109,6 +109,7 @@ object LeoRoutesSprayJsonCodec extends DefaultJsonProtocol {
       "defaultClientId" -> obj.defaultClientId.toJson,
       "stopAfterCreation" -> obj.stopAfterCreation.toJson,
       "welderEnabled" -> obj.welderEnabled.toJson,
+      "isUpdating" -> obj.isUpdating.toJson,
```
rtitle: Should this be in `GetRuntimeFormat` as well?

rtitle:
Looks good, just 2 comments above

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/LeoRoutesSprayJsonCodec.scala
akarukappadath: The original ticket was only for listClusters so I didn't add the field to the getCluster response - I didn't think it was necessary in getCluster since Terra UI bases the Runtime Manager off of the listCluster calls, but I can add it to getCluster if that would be helpful for AoU/otherwise

rtitle: My intuition is that it should be in getCluster even if not used.. I think of the listCluster response as a subset of getCluster.

akarukappadath:
jenkins retest

jdcanas:
Sorry, I didn't include it in the tickets acceptance critera, but I think its worth including a test for this flag in the existing patch automation test. I would say its not worth it normally, but since we already have an auto test that goes through the flow necessary to exercise this functionality it seems low-cost to add additional checks that the flag is set/cleaned-up as expected. `https://github.com/DataBiosphere/leonardo/blob/develop/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/cluster/ClusterPatchSpec.scala`

rtitle (approved):
Looks good, you may have a compilation error in `automation` though

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20200320_is_updating.xml
Qi77Qi: I don't think the field should be added here..it should be in `followupDetails` table

Qi77Qi:
can you link the ticket

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20200320_is_updating.xml
rtitle: I thought we wanted it on the CLUSTER table? It specifies whether a cluster has an update in progress, and is returned by listClusters/getCluster.

Qi77Qi (commented):
thanks for addressing comments...this looks good to me 👍

akarukappadath:
jenkins retest

PR #1333, opened 2020-03-23
IA-1777 use different encoder for responses

Qi77Qi:
Notice `cluster` is being used in get/list runtime apis when working with Wilson on automation test. Hence created https://broadworkbench.atlassian.net/browse/IA-1777

Use different encoder for `GetRuntimeResponse` and `ListRuntimeResponse` in runtime routes so that we're not returning fields like `clusterName`.

What's in swagger doc is not a complete list of what we return currently. It should be easy to add those additional fields if we need to.

for `dataprocInstances` field previously returned in `getClusterResponse`, I'm leaving it out in new route's encoder because I'd like to refactor `Instance` table to reference `RuntimeConfig` table instead of `Cluster`. And we can add `dataprocInstances` in `runtimeConfig` field once that's done.

---

In all cases:

### DIFF
--- a/http/src/main/resources/swagger/api-docs.yaml
+++ b/http/src/main/resources/swagger/api-docs.yaml
@@ -161,7 +161,7 @@ paths:
               schema:
                 type: array
                 items:
-                  $ref: "#/components/schemas/ListClusterResponse"
+                  $ref: "#/components/schemas/ListRuntimeResponse"
         "400":
           description: Bad Request
           content:
@@ -207,7 +207,7 @@ paths:
           content:
             application/json:
               schema:
-                $ref: "#/components/schemas/Cluster"
+                $ref: "#/components/schemas/GetRuntimeResponse"
         "403":
           description: User does not have permission to perform action on runtime
         "404":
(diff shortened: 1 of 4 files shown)


http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -1528,6 +1528,61 @@ components:
...
+        - runtimeConfig
+        - status
+        - labels
+        - autopauseThreshold
+        - defaultClientId
+        - scopes
+      properties:
+        id:
+          type: string
+          description: The internally-referenced ID of the cluster
```
rtitle: Maybe let's s/cluster/runtime in the descriptions?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/RuntimeRoutes.scala
```diff
@@ -250,6 +253,89 @@ object RuntimeRoutes {
       c.getOrElse(Map.empty)
     )
   }
+
+  //TODO: add `dataprocInstances` when it's added to `RuntimeConfig` https://broadworkbench.atlassian.net/browse/IA-1778
```
rtitle: Agreed `dataprocInstances` makes sense in `RuntimeConfig`. Will need to do a DB migration.

rtitle: As far as I know clients don't use this field, so should be fine to leave it out for now.

rtitle (changes requested):
1 swagger change requested; plus CI is failing

Qi77Qi:
@rtitle I addressed the comment. Let me know if you have more comments

rtitle (approved):

PR #1336, opened 2020-03-25
[No Ticket] Image Hardening CI Bug Fix

kyuksel:
Fix for the bug where not `git fetch`ing the custom image hardening submodule before `git checkout`ing a specified hash was causing an error when the hash wasn't locally tracked.

### DIFF
--- a/jenkins/Jenkinsfile
+++ b/jenkins/Jenkinsfile
@@ -140,7 +140,7 @@ pipeline {
             description: "The path for the bucket where the custom GCE images will be stored")
         string(name: "DATAPROC_VERSIONS", defaultValue: "1.2.79-debian9 1.4.15-debian9",
             description: "A custom image will be build for each of these dataproc versions")
-        string(name: "IMAGE_HARDENING_LIB_VERSION", defaultValue: "f45a1a2",
+        string(name: "IMAGE_HARDENING_LIB_VERSION", defaultValue: "fb388c5",
             description: "The hash of the 'broadinstitute/dsp-appsec-base-image-hardening' repo branch that contains libraries for CIS-hardening GCE images")
         booleanParam(name: "USE_CUSTOM_DOCKER_IMAGE_VERSIONS", defaultValue: false,
             description: "Check if you want to specify custom image versions below")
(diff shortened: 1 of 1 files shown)


Qi77Qi (approved):

rtitle (approved):

jenkins/Jenkinsfile
```diff
@@ -393,7 +393,7 @@ pipeline {
                                 String daisyImage = "gcr.io/compute-image-tools/daisy:latest"
                                 sh """
                                     cd jenkins/gce-custom-images/cis-harden-images
-                                    git checkout $IMAGE_HARDENING_LIB_VERSION 
+                                    git fetch && git checkout $IMAGE_HARDENING_LIB_VERSION
```
jdcanas, line 396: nit:

this can be done by a one time update to the .gitmodules file to track a specific branch via
`git config -f .gitmodules submodule.[submoduleName].branch [branchname]`
and then adding a `--remote` flag to the `submodule update` call

details were here https://git-scm.com/book/en/v2/Git-Tools-Submodules

not necessarily advocating to change it here since this works too (and I don't want you to have to change anymore at this point :) ), but I think it makes it more explicit that the intent is to update the submodule, as opposed to a fetch/checkout

jdcanas (approved):

jenkins/Jenkinsfile
kyuksel: Oh wow..thanks for the info. I'm just hesitant to add any more git magic to it at this point, and that testing it thoroughly may not be so straightforward.

kyuksel:
leonardo build failure with `denied: requested access to the resource is denied`

jenkins retest

jenkins/Jenkinsfile
jdcanas: yeah, I totally agree, if its working lets keep it, but a TIL for me when we have to do other submodule shenanigans

kyuksel:
Something similar happened in the past due to issues with certain slaves (outdated Docker version?) so trying another jenkins retest to see if it could land on another slave and test the hypothesis.

PR #1337, opened 2020-03-26
IA-1675 add metrics for gce

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-1675

Screen Shot 2020-03-28 at 9 46 09 PM

---

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/Boot.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/Boot.scala
@@ -9,7 +9,12 @@ import cats.implicits._
 import com.google.protobuf.ByteString
 import com.google.pubsub.v1.PubsubMessage
 import com.typesafe.sslconfig.akka.util.AkkaLoggerFactory
-import com.typesafe.sslconfig.ssl.{ConfigSSLContextBuilder, DefaultKeyManagerFactoryWrapper, DefaultTrustManagerFactoryWrapper, SSLConfigFactory}
+import com.typesafe.sslconfig.ssl.{
+  ConfigSSLContextBuilder,
+  DefaultKeyManagerFactoryWrapper,
+  DefaultTrustManagerFactoryWrapper,
+  SSLConfigFactory
+}
 import fs2.concurrent.InspectableQueue
 import fs2.{Pipe, Stream}
 import io.chrisdavenport.log4cats.slf4j.Slf4jLogger
@@ -17,8 +22,20 @@ import io.chrisdavenport.log4cats.{Logger, StructuredLogger}
 import io.circe.syntax._
 import javax.net.ssl.SSLContext
 import org.broadinstitute.dsde.workbench.google.GoogleCredentialModes.{Json, Token}
-import org.broadinstitute.dsde.workbench.google.{GoogleStorageDAO, HttpGoogleDirectoryDAO, HttpGoogleIamDAO, HttpGoogleProjectDAO, HttpGoogleStorageDAO}
-import org.broadinstitute.dsde.workbench.google2.{Event, GoogleComputeService, GooglePublisher, GoogleStorageService, GoogleSubscriber}
+import org.broadinstitute.dsde.workbench.google.{
+  GoogleStorageDAO,
+  HttpGoogleDirectoryDAO,
+  HttpGoogleIamDAO,
+  HttpGoogleProjectDAO,
+  HttpGoogleStorageDAO
+}
+import org.broadinstitute.dsde.workbench.google2.{
+  Event,
+  GoogleComputeService,
+  GooglePublisher,
+  GoogleStorageService,
+  GoogleSubscriber
+}
 import org.broadinstitute.dsde.workbench.leonardo.auth.sam.{PetClusterServiceAccountProvider, SamAuthProvider}
 import org.broadinstitute.dsde.workbench.leonardo.config.Config._
 import org.broadinstitute.dsde.workbench.leonardo.dao._
@@ -291,7 +308,8 @@ object Boot extends IOApp {
       // This is for sending custom metrics to stackdriver. all custom metrics starts with `OpenCensus/leonardo/`.
       // Typing in `leonardo` in metrics explorer will show all leonardo custom metrics.
       // As best practice, we should have all related metrics under same prefix separated by `/`
-      openTelemetry <- OpenTelemetryMetrics.resource[F](applicationConfig.leoServiceAccountJsonFile, applicationConfig.applicationName, blocker)
+      openTelemetry <- OpenTelemetryMetrics
+        .resource[F](applicationConfig.leoServiceAccountJsonFile, applicationConfig.applicationName, blocker)
     } yield AppDependencies(
       storage,
       dbRef,
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/Boot.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/Boot.scala
@@ -9,12 +9,7 @@ import cats.implicits._
 import com.google.protobuf.ByteString
 import com.google.pubsub.v1.PubsubMessage
 import com.typesafe.sslconfig.akka.util.AkkaLoggerFactory
-import com.typesafe.sslconfig.ssl.{
-  ConfigSSLContextBuilder,
-  DefaultKeyManagerFactoryWrapper,
-  DefaultTrustManagerFactoryWrapper,
-  SSLConfigFactory
-}
+import com.typesafe.sslconfig.ssl.{ConfigSSLContextBuilder, DefaultKeyManagerFactoryWrapper, DefaultTrustManagerFactoryWrapper, SSLConfigFactory}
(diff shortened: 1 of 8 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/Boot.scala
```diff
@@ -9,33 +9,16 @@ import cats.implicits._
...
-}
-import org.broadinstitute.dsde.workbench.google2.{
-  Event,
-  GoogleComputeService,
-  GooglePublisher,
-  GoogleStorageService,
-  GoogleSubscriber
-}
+import org.broadinstitute.dsde.workbench.google.{GoogleStorageDAO, HttpGoogleDirectoryDAO, HttpGoogleIamDAO, HttpGoogleProjectDAO, HttpGoogleStorageDAO}
+import org.broadinstitute.dsde.workbench.google2.{Event, GoogleComputeService, GooglePublisher, GoogleStorageService, GoogleSubscriber}
```
kyuksel: Have you run `scalafmt` on this?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
```diff
@@ -762,15 +752,16 @@ class ClusterMonitorActor(
 
   private def recordStatusTransitionMetrics(clusterUI: RuntimeUI,
```
kyuksel: Rename `clusterUI` as `runtimeUI`?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
```diff
@@ -762,15 +752,16 @@ class ClusterMonitorActor(
...
-      baseName = s"ClusterMonitor/${clusterUI.asString}/${origStatus}->${finalStatus}"
-      counterName = s"${baseName}/count"
-      timerName = s"${baseName}/timer"
+      metricsName = s"monitor/${clusterUI.asString}/transition/${origStatus}_to_${finalStatus}"
       duration = (endTime - startTime).millis
-      _ <- metrics.incrementCounter(counterName).runAsync(_ => IO.unit).toIO
-      _ <- metrics.recordResponseTime(timerName, duration).runAsync(_ => IO.unit).toIO
+      tags = Map("cloudService" -> cloudService.asString, "ui_client" -> clusterUI.asString)
+      _ <- openTelemetryMetrics.incrementCounter(metricsName, 1, tags)
+      histoBuckets = List(30000.0, 40000.0, 50000.0, 60000.0, 72000.0, 90000.0, 120000.0, 150000.0, 180000.0, 210000.0, 240000.0, 270000) //histogram buckets from 1 min to 6 min
```
kyuksel: May want to add to comments why 1 min to 6 mins. Also what the increment pattern is and why.

Could `histoBuckets` be more strongly typed? Could perhaps use a time-related library? It may also make reading easier with units.

May be worth spelling out `histogram` as `histo` may not be an obvious short form :)

`histoBuckets` is used in two places. Does it make sense to factor it out of the methods so it's defined in a single place?

project/Dependencies.scala
```diff
@@ -22,6 +22,7 @@ object Dependencies {
   val workbenchGoogle2V = "0.7-f3b5a5a"
   val workbenchMetricsV = "0.3-c5b80d2"
   val workbenchNewRelicV = "0.3-8bae8e8"
+  val workbenchOpenTelemetryV = "0.1-2587881-SNAP"
```
kyuksel: Moving to non-SNAP version before merging?

kyuksel: For some reason my IntelliJ can't import the SNAP version  🤷‍♂

kyuksel (commented):
The output looks nice! It may be a good walkthrough topic :) 

One of the acceptance criteria is that "We can measure the count of hail images versus GCE runtimes over time." Does this PR enable that?

Some comments inline.

project/Dependencies.scala
Qi77Qi: it's not published to jfrog :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
Qi77Qi: they're 2 different buckets. Buckets are metrics specific, like starting -> running should have different buckets from runtimeCreation

 I thought about making them `FiniteDuration`, but I think millis is a bit more granular.

Qi77Qi:
@kyuksel 
> The output looks nice! It may be a good walkthrough topic :)

yep, already added to walkthrough page

> One of the acceptance criteria is that "We can measure the count of hail images versus GCE runtimes over time." Does this PR enable that?

I think the AC is a bit more scoped than what this PR enables. This PR will enable measuring runtime creation broken down by all images, and cloudService (gce, dataproc)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
kyuksel: > they're 2 different buckets.

Sorry, not sure how I missed that.

kyuksel (approved):
👍 May still be good to explain the increment pattern in `histoBuckets` in the comments.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/Boot.scala
```diff
@@ -303,6 +288,7 @@ object Boot extends IOApp {
       subscriber <- GoogleSubscriber.resource(subscriberConfig, subscriberQueue)
 
       googleComputeService <- GoogleComputeService.resource(pathToCredentialJson, blocker, semaphore)
+      openTelemetry <- OpenTelemetryMetrics.resource[F](applicationConfig.leoServiceAccountJsonFile, "leonardo", blocker)
```
rtitle: Maybe use `applicationConfig.applicationName` instead of hardcoded string

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
```diff
@@ -762,15 +752,16 @@ class ClusterMonitorActor(
...
                                             origStatus: RuntimeStatus,
-                                            finalStatus: RuntimeStatus): IO[Unit] =
+                                            finalStatus: RuntimeStatus,
+                                            cloudService: CloudService): IO[Unit] =
     for {
       endTime <- IO(System.currentTimeMillis)
-      baseName = s"ClusterMonitor/${clusterUI.asString}/${origStatus}->${finalStatus}"
-      counterName = s"${baseName}/count"
-      timerName = s"${baseName}/timer"
+      metricsName = s"monitor/${clusterUI.asString}/transition/${origStatus}_to_${finalStatus}"
```
rtitle: Just curious - we used slash-separated names because they were wildcard-able in NewRelic.. are there different metric naming conventions for Stackdriver?

Also I was thinking it would be good to have a reference page with all our metric names.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
rtitle: Looks like it's 30 seconds to 4.5 minutes.. agreed it would be good to break this variable out to another place (`package.scala`?) since it's being reused and add a comment.

rtitle:
A few comments above but this looks cool.. can't wait to play around with it.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
Qi77Qi: it doesn't work the way newrelic does...we can group more flexibly with "tags" opencensus term (label in stackdriver). Using `/` is more of just naming convention thing.

I think reference page can grow out of sync pretty easily and tedious to maintain. All leo custom metrics starts with `OpenCensus/leonardo/xxx`, and if you type `leonardo`, you'll see all of them.  Maybe I can leave general comment about how metrics names will look like

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
Qi77Qi: this is not really being reused...each metric should have their own distribution

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/Boot.scala
Qi77Qi: yep..will do

PR #1340, opened 2020-03-27
[IA-1776] Implement runtime PATCH route

rtitle:
https://broadworkbench.atlassian.net/browse/IA-1776

This adds `PATCH /api/google/v1/runtimes/{project}/{name}` with equivalent functionality to `PATCH /api/cluster/{project}/{name}`.

I also made a couple of improvements/modifications in the new version:
1. Made dedicated models for `PATCH` requests, instead of overloading `RuntimeRequest`
2. Moved more work to Back Leo. Now `RuntimeServiceInterp` (front Leo) only does request validation and DB updates, and the actual GCP updates are done by the PubSub message subscriber (back Leo).

TODO:
- Fill in `LeoPubsubMessageSubscriberSpec`
- Swagger
- fiab testing

---

In all cases:


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeServiceInterp.scala
```diff
@@ -354,6 +406,122 @@ class RuntimeServiceInterp[F[_]: Parallel](blocker: Blocker,
         }
     }
   }
+
+  /**
+   * This method validates an UpdateRuntimeConfigRequest against the current state of the runtime,
+   * and potentially sends a PubSub message (or throws an error). The PubSub receiver does not
+   * do validation; it is assumed all validation happens here.
+   */
+  private[service] def processUpdateRuntimeConfigRequest(req: UpdateRuntimeConfigRequest,
```
rtitle: I think this logic is the same as in `LeonardoService`. I added a lot of unit tests to make sure.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/RuntimeRoutes.scala
```diff
@@ -201,13 +210,30 @@ class RuntimeRoutes(runtimeService: RuntimeService[IO], userInfoDirectives: User
       )
       _ <- runtimeService.startRuntime(userInfo, googleProject, runtimeName)
     } yield StatusCodes.Accepted
+
+  private[api] def updateRuntimeHandler(userInfo: UserInfo,
```
Qi77Qi, line 214: I wonder if it makes sense to put `UserInfo`, `GoogleProject`, `RuntimeName` in `RuntimeServiceContext`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/RuntimeRoutes.scala
```diff
@@ -201,13 +210,30 @@ class RuntimeRoutes(runtimeService: RuntimeService[IO], userInfoDirectives: User
...
 
 object RuntimeRoutes {
-  implicit val gceConfigDecoder: Decoder[RuntimeConfigRequest.GceConfig] = Decoder.forProduct2(
-    "machineType",
-    "diskSize"
-  )((mt, ds) => RuntimeConfigRequest.GceConfig(mt, ds))
+  implicit val gceConfigDecoder: Decoder[RuntimeConfigRequest.GceConfig] = Decoder.instance { x =>
+    for {
+      machineType <- x.downField("machineType").as[Option[MachineTypeName]]
+      diskSize <- x
```
Qi77Qi, line 231: if we change `diskSize` type to dedicated type(think there's one already?), we don't have to do the negative check here.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/RuntimeRoutes.scala
```diff
@@ -254,6 +280,56 @@ object RuntimeRoutes {
     )
   }
 
+  implicit val updateGceConfigDecoder: Decoder[UpdateRuntimeConfigRequest.GceConfig] = Decoder.instance { x =>
+    for {
+      machineType <- x.downField("updatedMachineType").as[Option[MachineTypeName]]
+      diskSize <- x
+        .downField("updatedDiskSize")
```
Qi77Qi, line 286: same comment

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/RuntimeRoutes.scala
```diff
@@ -254,6 +280,56 @@ object RuntimeRoutes {
...
+      }
+    } yield r
+  }
+
+  implicit val updateRuntimeRequestDecoder: Decoder[UpdateRuntimeRequest] = Decoder.instance { x =>
+    for {
+      rc <- x.downField("updatedRuntimeConfig").as[Option[UpdateRuntimeConfigRequest]]
+      as <- x.downField("allowStop").as[Option[Boolean]]
+      ap <- x.downField("updatedAutopause").as[Option[Boolean]]
+      at <- x.downField("updatedAutopauseThreshold").as[Option[Int]]
```
Qi77Qi, line 326: use  `Option[FiniteDuration]` for `updatedAutopauseThreshold`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/leoPubsubMessageSubscriberModels.scala
```diff
@@ -108,12 +114,21 @@ object LeoPubsubMessage {
     val messageType: LeoPubsubMessageType = LeoPubsubMessageType.StartRuntime
   }
 
-  final case class RuntimeFollowupDetails(runtimeId: Long, runtimeStatus: RuntimeStatus)
-      extends Product
-      with Serializable
+  final case class UpdateRuntimeMessage(runtimeId: Long,
+                                        newMachineType: Option[MachineTypeName],
+                                        stopUpdateMachineType: Boolean,
```
Qi77Qi: what does `stopUpdateMachineType` mean? maybe add a comment?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeServiceInterp.scala
```diff
@@ -281,6 +293,46 @@ class RuntimeServiceInterp[F[_]: Parallel](blocker: Blocker,
...
+    googleProject: GoogleProject,
+    runtimeName: RuntimeName,
+    req: UpdateRuntimeRequest
+  )(implicit as: ApplicativeAsk[F, RuntimeServiceContext]): F[Unit] =
+    for {
+      // throw 404 if not existent
+      runtimeOpt <- clusterQuery.getActiveClusterByNameMinimal(googleProject, runtimeName).transaction
+      runtime <- runtimeOpt.fold(F.raiseError[Runtime](RuntimeNotFoundException(googleProject, runtimeName)))(F.pure)
+      // throw 404 if no GetClusterStatus permission
+      hasPermission <- authProvider.hasNotebookClusterPermission(runtime.internalId,
```
Qi77Qi, line 315: should this be 403?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/RuntimeRoutes.scala
```diff
@@ -363,3 +439,25 @@ final case class CreateRuntime2Request(labels: LabelMap,
...
+                                  updatedMasterDiskSize: Option[Int],
+                                  updatedNumberOfWorkers: Option[Int],
+                                  updatedNumberOfPreemptibleWorkers: Option[Int])
+      extends UpdateRuntimeConfigRequest {
+    val cloudService: CloudService = CloudService.Dataproc
+  }
+}
+
+final case class UpdateRuntimeRequest(updatedRuntimeConfig: Option[UpdateRuntimeConfigRequest],
+                                      allowStop: Option[Boolean],
```
Qi77Qi: can we make this `Boolean` instead of `Option[Boolean]`? We're defaulting it to `false` anyway, and having it as `Boolean` makes it a cleaner on `caller` side (we don't have to do `.getOrElse(false)`` everywhere

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeServiceInterp.scala
```diff
@@ -354,6 +406,122 @@ class RuntimeServiceInterp[F[_]: Parallel](blocker: Blocker,
...
+                                         _,
+                                         existingNumPreemptibles,
+                                         _),
+            UpdateRuntimeConfigRequest.DataprocConfig(reqMasterMachineType,
+                                                      reqMasterDiskSize,
+                                                      reqNumOfWorkers,
+                                                      reqNumPreemptibles)) =>
+        for {
+          // should num workers be updated?
+          updatedNumWorkers <- reqNumOfWorkers.flatTraverse[F, Int] {
```
Qi77Qi: minor rename `updated` to `target`? `updated` reads to me as if the update has already happened

Qi77Qi (commented):
a few minor comments..
Looks great! so much better than the old patch endpoint. Very glad to see the changes :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/RuntimeRoutes.scala
rtitle: Not all routes have `GoogleProject` and `RuntimeName` (e.g. list clusters)… also I feel like `UserInfo` is better left explicit.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/RuntimeRoutes.scala
rtitle: Yeah that's a good idea
update: I did this

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/leoPubsubMessageSubscriberModels.scala
rtitle: I can add a comment.. if `stopUpdateMachineType == true` then the runtime should go through a stop-update transition (via `FollowupDetails`). If `false` then the machine type can be changed immediately.

(This also implies that there is no more use for the `StopUpdate` message type, as `UpdateRuntimeMessage` will handle the stop-update flow.)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeServiceInterp.scala
rtitle: Our pattern is to 404 (e.g. pretend the runtime doesn't exist) if the user doesn't have `GetClusterStatus` permission. We return 403 if the user can _view_ the runtime but can't perform some other action.

rtitle:
Thanks for the review @Qi77Qi! I'll apply those comments (am also updating Swagger)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeServiceInterp.scala
Qi77Qi: gotcha, that's very helpful...can you add the comment in code as well :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/leoPubsubMessageSubscriberModels.scala
Qi77Qi: shall we delete `StopUpdate` message then?

rtitle: It's still used by `LeonardoService`.. I really want to delete `LeoRoutes` and `LeonardoService` ;)

Qi77Qi: I'm thinking once we migrate terra-ui to use new route, we can at least delete the patch implementation in leonardoservice since patch is added recently and mostly likely no one is using it

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/RuntimeRoutes.scala
rtitle: I left this as `Int` because we interpret it as `minutes` and convert to `FiniteDuration` in the yield block.

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -316,6 +316,57 @@ paths:
...
+          required: true
+          schema:
+            type: string
+        - in: path
+          name: name
+          description: runtimeName. only lowercase alphanumeric characters, numbers and dashes are allowed
+          required: true
+          schema:
+            type: string
+        requestBody:
```
kyuksel: I get an indentation error for the above line when I copy/paste the `yaml` content onto http://editor.swagger.io.

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -1849,6 +1900,74 @@ components:
         customClusterEnvironmentVariables:
           type: object
           description: Optional environment variables to be set on the cluster.
+    UpdateRuntimeConfig:
```
kyuksel: add `description` here too?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -269,6 +271,46 @@ class LeoPubsubMessageSubscriber[F[_]: Async: Timer: ContextShift: Concurrent](
...
+                .save(followupDetails, Some(m))
+            )
+            _ <- runtimeConfig.cloudService.interpreter
+              .stopRuntime(StopRuntimeParams(RuntimeAndRuntimeConfig(runtime, runtimeConfig), now))
+          } yield ()
+        } else {
+          runtimeConfig.cloudService.interpreter.updateMachineType(UpdateMachineTypeParams(runtime, m, now))
+        }
+      }
+    } yield ()
```
kyuksel, line 330: Do we need to add/update `isUpdating` flag?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/leoPubsubMessageSubscriberModels.scala
kyuksel: I wonder if `stopBeforeUpdatingMachineType` or `stopToUpdateMachineType` may be more explanatory without before overly verbose  🤔

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeServiceInterp.scala
kyuksel: The method is quite long. Does it make sense to factor out the `case`s so it's easier to follow at a glance? It also may (or may not) help factoring out some lines that are duplicated.

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeServiceInterpSpec.scala
```diff
@@ -359,4 +361,262 @@ class RuntimeServiceInterpSpec extends FlatSpec with LeonardoTestSuite with Test
...
+                                            false,
+                                            None,
+                                            None,
+                                            None,
+                                            Some(traceId))
+    }
+    res.unsafeRunSync()
+  }
+
+  it should "update a GCE machine type in Running state" in {
```
kyuksel, line 448: Would it be good to test states other than `Running` and `Stopped`?

kyuksel (commented):
Looks good overall. Questions/comments inline.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeServiceInterp.scala
rtitle: Yeah, I'll see if anything makes sense to factor out. However having a lot of separate methods is not (to me) always easier to follow if the logic is complex.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/leoPubsubMessageSubscriberModels.scala
rtitle: I like `stopToUpdateMachineType`, I can change it

http/src/main/resources/swagger/api-docs.yaml
rtitle: should be fixed now

http/src/main/resources/swagger/api-docs.yaml
rtitle: done

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
rtitle: Yeah… that PR isn't merged yet, will probably need to coordinate with @akarukappadath about who goes first. (I'm happy to rebase this PR and incorporate that flag.)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeServiceInterp.scala
Qi77Qi: yea. I don't like looking at code with too many unnecessary small functions where I have to jump around a lot

kyuksel: I find it easier to look at "table of contents" before diving into the "doc" but I see yous' perspective as well  ☺️

rtitle:
@Qi77Qi will double check on the Swagger stuff. But here is my thinking:
1. I saw that some route definitions had `security:` and some didn't. I think if it's not present it just uses the security definition at the top level, and can be specified per-route for overrides. So I just removed all of them where it's not needed.
2. I noticed that Swagger UI wasn't displaying all the fields for `runtimeConfig`. Based on https://swagger.io/docs/specification/data-models/inheritance-and-polymorphism/ I changed it to:
```
updatedRuntimeConfig:
  oneOf:
    - $ref: "#/components/schemas/UpdateGceConfig"
    - $ref: "#/components/schemas/UpdateDataprocConfig"
  discriminator:
    propertyName: cloudService
```
That seems to make it show up in the UI properly.

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeServiceInterpSpec.scala
rtitle: Like Starting, Stopping, Creating, Deleting? Hm maybe we should disallow updates in those states (and test for it).

kyuksel: makes sense to me

kyuksel (approved):
Thanks for the refactoring, status checking and the test additions 👍

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -174,17 +169,12 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/ErrorReport"
-      security:
```
Qi77Qi, line 177: are these not needed?

rtitle: I think it inherits from the top level: https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/resources/swagger/api-docs.yaml#L22

Per-route security settings are only needed for overrides.

Looking on dev the swagger routes are authenticated correctly: https://leonardo.dsde-dev.broadinstitute.org/#/runtimes/listRuntimesByProject

PR #1342, opened 2020-03-31
[no ticket] Fix GCE resume by removing unneeded metadata entry

rtitle:
I think this fixes the problem with GCE `resume`. Previously we had duplicated startup scripts defined via instance metadata:
* `startup-script-url` -- used only for creation and points to `gce-init.sh`
* `startup-script` -- used for resume

This PR removes the `startup-script-url` metadata entry at start time when it's no longer needed.

See https://github.com/broadinstitute/workbench-libs/pull/293

---

In all cases:

### DIFF
--- a/docker/build.sh
+++ b/docker/build.sh
@@ -151,14 +151,14 @@ function make_jar()
     GIT_HASH=$(git log -n 1 --pretty=format:%h)
 
     # Make jar & cache sbt dependencies.
+    EXIT_CODE=0
     docker run --rm --link $DB_CONTAINER:mysql \
                           -e GIT_HASH=$GIT_HASH \
                           -v $PWD:/working \
                           -v jar-cache:/root/.ivy \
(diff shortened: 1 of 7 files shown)


Qi77Qi (approved):

rtitle:
jenkins retest

docker/build.sh
```diff
@@ -151,13 +151,13 @@ function make_jar()
     GIT_HASH=$(git log -n 1 --pretty=format:%h)
 
     # Make jar & cache sbt dependencies.
-    JAR_CMD="$(docker run --rm --link $DB_CONTAINER:mysql \
+    docker run --rm --link $DB_CONTAINER:mysql \
```
rtitle, line 155: This makes all build output display in Jenkins `leonardo-build`.

…although Jenkins crashed right after I did this 😰

rtitle:
jenkins retest

PR #1343, opened 2020-04-02
IA-1589 IA-1798 Fix a bug where user script and startup script failure doesn't get propagated properly

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-1798

Status from gce doesn't indicate whether or not a user script or startup script runs successfully. Hence we're checking metadata for both script output files in order to tell whether runtime creation is successful or not.

For user script, since the output file name is the same (since we only create the runtime once), it's easy to check the file metadata. But for user startup script, since every time runtime restarts, we generate a different output file, hence this PR persists the output path to instance metadata during runtime creation time (only in the case of gce)

Screen Shot 2020-04-03 at 11 15 29 AM
Screen Shot 2020-04-03 at 11 15 20 AM

Tested manually...File format is updated a bit to make things easier to find

---

In all cases:

### DIFF
--- a/http/src/main/resources/init-resources/gce-init.sh
+++ b/http/src/main/resources/init-resources/gce-init.sh
@@ -102,7 +102,7 @@ export JUPYTER_SERVER_NAME=$(jupyterServerName)
 export JUPYTER_DOCKER_IMAGE=$(jupyterDockerImage)
 export JUPYTER_START_USER_SCRIPT_URI=$(jupyterStartUserScriptUri)
 # Include a timestamp suffix to differentiate different startup logs across restarts.
-export JUPYTER_START_USER_SCRIPT_OUTPUT_URI="$(jupyterStartUserScriptOutputBaseUri)-$(date -u "+%Y.%m.%d-%H.%M.%S").txt"
+export JUPYTER_START_USER_SCRIPT_OUTPUT_URI=$(jupyterStartUserScriptOutputUri)
 export NOTEBOOKS_DIR=$(notebooksDir)
 export WELDER_SERVER_NAME=$(welderServerName)
 export WELDER_DOCKER_IMAGE=$(welderDockerImage)
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/models.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/models.scala
@@ -7,7 +7,6 @@ import cats.effect.{IO, Timer}
 import org.broadinstitute.dsde.workbench.leonardo.http.nowInstant
 import org.broadinstitute.dsde.workbench.model.TraceId
 
-
 final case class AppContext(traceId: TraceId, now: Instant)
 object AppContext {
   def generate(implicit timer: Timer[IO]): IO[AppContext] =
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/models.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/models.scala
@@ -1,0 +1,18 @@
+package org.broadinstitute.dsde.workbench.leonardo
+
+import java.time.Instant
+import java.util.UUID
+
+import cats.effect.{IO, Timer}
+import org.broadinstitute.dsde.workbench.leonardo.http.nowInstant
+import org.broadinstitute.dsde.workbench.model.TraceId
+
+
+final case class AppContext(traceId: TraceId, now: Instant)
+object AppContext {
+  def generate(implicit timer: Timer[IO]): IO[AppContext] =
+    for {
+      traceId <- IO(UUID.randomUUID())
+      now <- nowInstant[IO]
+    } yield AppContext(TraceId(traceId), now)
+}
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
@@ -17,24 +17,8 @@ import cats.mtl.ApplicativeAsk
 import com.typesafe.scalalogging.LazyLogging
 import io.grpc.Status.Code
 import org.broadinstitute.dsde.workbench.google.GoogleStorageDAO
-import org.broadinstitute.dsde.workbench.google2.{
-  GcsBlobName,
-  GetMetadataResponse,
-  GoogleComputeService,
-  GoogleStorageService,
-  InstanceName
-}
-import org.broadinstitute.dsde.workbench.leonardo.RuntimeStatus.{
-  Creating,
-  Deleted,
-  Deleting,
-  Error,
-  Running,
-  Starting,
-  Stopping,
-  Unknown,
-  Updating
-}
+import org.broadinstitute.dsde.workbench.google2.{GcsBlobName, GetMetadataResponse, GoogleComputeService, GoogleStorageService, InstanceName}
+import org.broadinstitute.dsde.workbench.leonardo.RuntimeStatus.{Creating, Deleted, Deleting, Error, Running, Starting, Stopping, Unknown, Updating}
 import org.broadinstitute.dsde.workbench.leonardo.config._
 import org.broadinstitute.dsde.workbench.leonardo.dao.ToolDAO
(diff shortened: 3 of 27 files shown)


http/src/main/resources/init-resources/gce-init.sh
```diff
@@ -102,7 +102,7 @@ export JUPYTER_SERVER_NAME=$(jupyterServerName)
 export JUPYTER_DOCKER_IMAGE=$(jupyterDockerImage)
 export JUPYTER_START_USER_SCRIPT_URI=$(jupyterStartUserScriptUri)
 # Include a timestamp suffix to differentiate different startup logs across restarts.
-export JUPYTER_START_USER_SCRIPT_OUTPUT_URI="$(jupyterStartUserScriptOutputBaseUri)-$(date -u "+%Y.%m.%d-%H.%M.%S").txt"
+export JUPYTER_START_USER_SCRIPT_OUTPUT_URI=$(jupyterStartUserScriptOutputUri)
```
rtitle, line 105: Hm - do we need a timestamp per this comment? AoU might be consuming these logs.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
```diff
@@ -495,76 +480,103 @@ class ClusterMonitorActor(
...
+            runtimeStatus = instanceStatusToRuntimeStatus(instanceOpt)
+            userScript = runtimeAndRuntimeConfig.runtime.asyncRuntimeFields.map(_.stagingBucket).map(b => RuntimeTemplateValues.jupyterUserScriptOutputUriPath(b))
+            userStartupScript = instanceOpt.flatMap(x => Option(x.getMetadata.getFieldValue(userScriptStartupOutputUriMetadataKey)))
+              .flatMap(s => org.broadinstitute.dsde.workbench.model.google.parseGcsPath(s.toString).toOption)
+
+            userScriptSuccess <- checkUserScripts(userScript, google2StorageDAO)
+            startUpScriptSuccess <- checkUserScripts(userStartupScript, google2StorageDAO)
+            r <- (userScriptSuccess, startUpScriptSuccess) match {
+              case (true, true) => continueCheckRuntimeInGoogle(runtimeAndRuntimeConfig, runtimeStatus)
+              case (false, true) => IO.pure(FailedCluster(runtimeAndRuntimeConfig, RuntimeErrorDetails(-1, Some("user script failed")), Set.empty): ClusterMonitorMessage)
```
rtitle: It might be nice if the error message(s) pointed to the script output file

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
```diff
@@ -495,76 +480,103 @@ class ClusterMonitorActor(
...
       }
 
-      runningInstanceCount = dataprocInstances.count(_.status == GceInstanceStatus.Running)
-      stoppedInstanceCount = dataprocInstances.count(
-        i => i.status == GceInstanceStatus.Stopped || i.status == GceInstanceStatus.Terminated
-      )
+  private def continueCheckRuntimeInGoogle(
+                                            runtimeAndRuntimeConfig: RuntimeAndRuntimeConfig,
+                                            runtimeStatus: RuntimeStatus
+                                          )(implicit ev: ApplicativeAsk[IO, TraceId]): IO[ClusterMonitorMessage]  = for {
```
rtitle: May need scalafmt

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
```diff
@@ -495,76 +480,103 @@ class ClusterMonitorActor(
...
+                RuntimeStatus.Running,
+                dataprocInstances,
+                Some("Could not retrieve master IP"))
+            )
+        }
+      // Take care we don't fail a Deleting or Stopping cluster if google hasn't updated their status yet
+      case Error
+        if runtimeAndRuntimeConfig.runtime.status != Deleting && runtimeAndRuntimeConfig.runtime.status != Stopping =>
+        runtimeAndRuntimeConfig.runtimeConfig.cloudService match {
+          // TODO implement getErrorDetails in wb-libs
```
rtitle: Now I remember running into this -- there's no such thing as `Error` status in GCE:
https://cloud.google.com/compute/docs/reference/rest/v1/instances/get
```
enum

[Output Only] The status of the instance. One of the following values: PROVISIONING, STAGING, RUNNING, STOPPING, STOPPED, SUSPENDING, SUSPENDED, and TERMINATED.
```

So this is a Dataproc-only concept..

http/src/main/resources/init-resources/gce-init.sh
rtitle: I see we are using GCS object metadata to check on the result, so want a consistent output URI.

What if we wrote 2 files every time:
- `$jupyterStartUserScriptOutputUri` -- always contains the 'latest' output, can be used to store "passed" metadata flag
- `$jupyterStartUserScriptOutputUri-$(timestamp)` -- tracks output over time

Another idea is to use GCS object versioning.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/models.scala
```diff
@@ -0,0 +1,18 @@
...
+
+import java.time.Instant
+import java.util.UUID
+
+import cats.effect.{IO, Timer}
+import org.broadinstitute.dsde.workbench.leonardo.http.nowInstant
+import org.broadinstitute.dsde.workbench.model.TraceId
+
+
+final case class AppContext(traceId: TraceId, now: Instant)
```
rtitle, line 10: So `AppContext` can now be used everywhere instead of needing to convert across different types of `ApplicativeAsk`? I like this.

http/src/main/resources/init-resources/gce-init.sh
rtitle: Oh wait.. we're still appending the timestamp, just in `RuntimeTemplateValues` instead of here. Do I have that right?

I do think `startup.sh` needs to be updated then, in addition to this script.

rtitle (changes requested):
I am still not sure I understand the flow for tracking the startup script output filename. It looks like we generate a filename in `RuntimeTemplateValues` and store the value in instance metadata. But I don't see how it changes over time -- e.g. I would expect every time the instance is resumed a new filename is generated and the metadata field is updated. I don't think I saw that -- let me know if I'm missing something.

Also: `gce-init.sh` is only used for instance creation -- `startup.sh` is used for instance resume so that probably needs to be updated too.

http/src/main/resources/init-resources/gce-init.sh
Qi77Qi: yea...I'm moving this timestamp out of the script so that leo knows about which file to look up later

Qi77Qi:
yea, I missed the startup script...will update

Qi77Qi:
@rtitle , think the issue should be addressed. I created a gce runtime, and then stop and restart the runtime, 2 startup script output files were generated, and the instance metadata is updated to most recent output url

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/models.scala
Qi77Qi: yep..that's the idea, I think it should be used in most routes api calls going forward.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorActor.scala
Qi77Qi: yea, but we're mapping gce statuses to `RuntimeStatus`, and deciding what to do for both gce and dataproc...It's not obvious what would be cleaner since branching out depends on gce and dataproc means we'll need to determine what to do based on gce or dataproc specific statuses as well

Qi77Qi: now that I think about it..Maybe a better way to address this is to have different monitoring code for gce vs dataproc....since we also want different polling interval for them as well

rtitle (approved):
Looks good

PR #1349, opened 2020-04-06
IA-1816 IA-1821 A-1803 Optimize monitoring for gce runtime

Qi77Qi:
Use fs2 stream for gce runtime monitoring...

Question: Do we have a process where a runtime is deleted in leo db, and we then delete the runtime in google as well?

Depends on https://github.com/broadinstitute/workbench-libs/pull/297

Things changed:
1. Use fs2.Stream for any polling instead of ClusterMonitorActor
2. traceId will be logged throughout createRuntime call including monitoring part
3. Autopause for GCE will send an event to subscriber to trigger stopping instead of relying on clusterMonitorActor
4. convert error code to string representation so that it's easier to understand their meaning, and add it as a label in `runtimeCreationFailure` metric
5. remove googleStorageDao usage everywhere.
6. Reenabled previously failing 2 gce automation tests (patch and scopes)
7. Break autopause into its own class and initiated in boot just like any other infinite Stream
8. Replaced all use of newrelic to openTelemetry
9. Added an automation test for runtime creation with bad user startup script

TODO: 
~~1. need to account for patch during startup~~
~~2. use traceId more through out the call stack~~
~~3. remove recreate runtime logic~~
~~4. stopRuntime, deleteRuntime~~

---

In all cases:


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/GceRuntimeMonitorInterp.scala
```diff
@@ -0,0 +1,690 @@
...
+  ec: ExecutionContext,
+  openTelemetry: OpenTelemetryMetrics[F])
+    extends GceRuntimeMonitor[F] {
+  def process(runtimeId: Long)(implicit ev: ApplicativeAsk[F, TraceId]): Stream[F, Unit] =
+    for {
+      // building up a stream that will terminate when gce runtime is ready
+      traceId <- Stream.eval(ev.ask)
+      startMonitoring <- Stream.eval(nowInstant[F])
+      monitorContext = MonitorContext(startMonitoring, runtimeId, traceId)
+      // 20 seconds is from experience. No status transition took less than 20 seconds. Hence wait 20 seconds before start polling
```
kyuksel: May want to move the comment above to where `initialDelay` is defined.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/GceRuntimeMonitorInterp.scala
```diff
@@ -0,0 +1,690 @@
...
+                runtimeAndRuntimeConfig: RuntimeAndRuntimeConfig,
+                operation: com.google.cloud.compute.v1.Operation,
+                action: RuntimeStatus)(implicit ev: ApplicativeAsk[F, TraceId]): F[Unit] =
+    for {
+      // building up a stream that will terminate when gce runtime is ready
+      traceId <- ev.ask
+      startMonitoring <- nowInstant
+      monitorContext = MonitorContext(startMonitoring, runtimeAndRuntimeConfig.runtime.id, traceId)
+      // 20 seconds is from experience. No status transition took less than 20 seconds. Hence wait 20 seconds before start polling
+      _ <- Timer[F].sleep(30 seconds)
```
kyuksel: Move `30 seconds` to a config along with the comment (which probably needs updating).

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/GceRuntimeMonitorInterp.scala
```diff
@@ -0,0 +1,690 @@
...
+            for {
+              userScriptSuccess <- checkUserScripts(userScript)
+              startUpScriptSuccess <- checkUserScripts(userStartupScript)
+              r <- (userScriptSuccess, startUpScriptSuccess) match {
+                case (true, true) =>
+                  getInstanceIP(i) match {
+                    case Some(ip) =>
+                      // It takes a bit for jupyter to startup, hence wait 5 seconds before we check jupyter
+                      Timer[F]
+                        .sleep(8 seconds) >> handleCheckTools(monitorContext, runtimeAndRuntimeConfig, ip, List.empty)
```
kyuksel, line 225: Move to config

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/GceRuntimeMonitorInterp.scala
```diff
@@ -0,0 +1,690 @@
...
+          case GceInstanceStatus.Running =>
+            val userStartupScript = getUserScript(i)
+
+            for {
+              startUpScriptSuccess <- checkUserScripts(userStartupScript)
+              r <- if (startUpScriptSuccess)
+                getInstanceIP(i) match {
+                  case Some(ip) =>
+                    // It takes a bit for jupyter to startup, hence wait 5 seconds before we check jupyter
+                    Timer[F].sleep(8 seconds) >> handleCheckTools(monitorContext, runtimeAndRuntimeConfig, ip, List.empty)
```
kyuksel: Move to config

kyuksel (commented):
Not a comprehensive pass yet. Just wanted to point out the hardcoded sleep times that I'd noticed during the walkthrough 😊

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/GceRuntimeMonitorInterp.scala
Qi77Qi: these are largely experimental...they'll be moved to config before review/merge

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookGCECustomizationSpec.scala
```diff
@@ -99,35 +94,36 @@ final class NotebookGCECustomizationSpec extends GPAllocFixtureSpec with Paralle
     }
 
     //TODO Renable test once this is fixed
-//    "should give cluster user-specified scopes" in { billingProject =>
```
rtitle, line 103: Need rebase?

http/src/main/resources/reference.conf
```diff
@@ -54,6 +54,22 @@ gce {
     diskSize = 100
   }
   gceReservedMemory = 1g
+
+  monitor {
```
rtitle, line 58: I notice this is inside the `gce` config. When Dataproc eventually uses this will we move it out? Or will dataproc have separate `monitor` configurations?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/Boot.scala
```diff
@@ -210,13 +172,63 @@ object Boot extends IOApp {
...
       val allStreams = {
-        val extra =
-          if (leoExecutionModeConfig.backLeo) {
-            // only needed for backleo
-            val pubsubSubscriber: LeoPubsubMessageSubscriber[IO] =
-              new LeoPubsubMessageSubscriber(appDependencies.subscriber)
-            List(pubsubSubscriber.process, Stream.eval(appDependencies.subscriber.start), zombieClusterMonitor.process)
-          } else List.empty[Stream[IO, Unit]]
+        val extra = if (leoExecutionModeConfig.backLeo) {
+          implicit def clusterToolToToolDao =
```
rtitle: Minor, just wondering if this can be a `val` instead of a `def`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/Config.scala
```diff
@@ -381,7 +382,31 @@ object Config {
   val clusterToolMonitorConfig = config.as[ClusterToolConfig](path = "clusterToolMonitor")
   val clusterDnsCacheConfig = config.as[ClusterDnsCacheConfig]("clusterDnsCache")
   val leoExecutionModeConfig = config.as[LeoExecutionModeConfig]("leoExecutionMode")
-  val clusterBucketConfig = config.as[ClusterBucketConfig]("clusterBucket")
+  val clusterBucketConfig = config.as[RuntimeBucketConfig]("clusterBucket")
+
+
+  implicit val gceMonitorConfigReader: ValueReader[GceMonitorConfig] = ValueReader.relative { config =>
```
rtitle, line 388: Maybe put the `ValueReader[GceMonitorConfig]` next to the `ValueReader[GceConfig]` in this file?

rtitle: Also if we intend to use this with Dataproc (and GKE?) too then let's not name the class `GceMonitorConfig`. Maybe `StreamingMontitorConfig` or something?

http/src/main/resources/reference.conf
```diff
@@ -54,6 +54,22 @@ gce {
...
+    initialDelay = 20 seconds
+    pollingInterval = 15 seconds
+    checkToolsDelay = 8 seconds
+    pollCheckMaxAttempts = 120 # 15 seconds * 120 is 30 min
+    maxRetries = -1  # means retry forever
```
rtitle: I think we decided to get rid of `maxRetries`. Any other removable fields?

http/src/main/resources/reference.conf
rtitle: Might make sense for different services to have different monitor configs but use the same class.. similar to `gce.runtimeDefaults` and `dataproc.runtimeDefaults`.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpJupyterDAO.scala
```diff
@@ -20,14 +20,16 @@ class HttpJupyterDAO[F[_]: Timer: ContextShift: Concurrent](val clusterDnsCache:
...
+          .successful(
+            Request[F](
+              method = Method.GET,
+              uri = Uri.unsafeFromString(
+                s"https://${targetHost.toString}/notebooks/${googleProject.value}/${runtimeName.asString}/api/status"
+              )
             )
           )
-        )
+          .handleError(_ => false)
```
rtitle, line 32: Does this make it not retry?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
```diff
@@ -309,6 +310,22 @@ object clusterQuery extends TableQuery(new ClusterTable(_)) {
       )
     }
 
+  def listMonitoredGceOnly(implicit ec: ExecutionContext): DBIO[Seq[Runtime]] =
+    clusterQuery
+      .filter { _.status inSetBind RuntimeStatus.monitoredStatuses.map(_.toString) }
+      .joinLeft(runtimeConfigs)
```
rtitle: Could this be an inner join instead of left join? I think we expect all runtimes to have runtime configs (and if not, we probably don't need to process them here).

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
```diff
@@ -309,6 +310,22 @@ object clusterQuery extends TableQuery(new ClusterTable(_)) {
...
+      .on(_.runtimeConfigId === _.id)
+      .filter {
+        case (_, runtimeConfig) =>
+          runtimeConfig.map(x => x.cloudService.asColumnOf[String] === CloudService.GCE.asString)
```
rtitle: Do we have an index on `RUNTIME_CONFIG.cloudService` in the DB?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/AutopauseMonitor.scala
```diff
@@ -0,0 +1,84 @@
...
+import org.broadinstitute.dsde.workbench.leonardo.dao.JupyterDAO
+import org.broadinstitute.dsde.workbench.leonardo.db._
+import org.broadinstitute.dsde.workbench.leonardo.http.{dbioToIO, nowInstant}
+import org.broadinstitute.dsde.workbench.model.TraceId
+import org.broadinstitute.dsde.workbench.openTelemetry.OpenTelemetryMetrics
+
+import scala.concurrent.ExecutionContext
+
+/**
+ * This monitor periodically sweeps the Leo database and checks for clusters which no longer exist in Google.
```
rtitle: This comment is wrong, seems copied from the zombie monitor

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/AutopauseMonitor.scala
```diff
@@ -0,0 +1,84 @@
...
+)(implicit F: Async[F],
+  metrics: OpenTelemetryMetrics[F],
+  logger: Logger[F],
+  dbRef: DbReference[F],
+  ec: ExecutionContext) {
+
+  val process: Stream[F, Unit] =
+    (Stream.sleep[F](config.autoFreezeCheckInterval) ++ Stream.eval(autoPauseCheck)).repeat
+
+  private[monitor] val autoPauseCheck: F[Unit] =
```
rtitle, line 34: Cool to break this into its own class/stream. Logic looks correct to me, am glad we have an autopause automation test now. :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorSupervisor.scala
```diff
@@ -333,21 +263,29 @@ class ClusterMonitorSupervisor(
       .flatMap {
         case Right(clusters) =>
           val clustersNotAlreadyBeingMonitored = clusters.filterNot(c => monitoredClusterIds.contains(c.id))
+          clustersNotAlreadyBeingMonitored.toList
+            .filterA { c =>
+              RuntimeConfigQueries
+                .getRuntimeConfig(c.runtimeConfigId)
+                .transaction
+                .map(x => x.cloudService == CloudService.Dataproc) //only monitor dataproc runtimes
```
rtitle: Could this filtering be done in the DB query (`clusterQuery.listMonitoredClusterOnly`)?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterToolMonitor.scala
```diff
@@ -36,15 +32,14 @@ object ClusterToolMonitor {
 }
 
 /**
+ * TODO: Is this needed?
```
rtitle: I think this was added to alert on things like Jupyter/welder crashing for a running cluster. Don't think anyone is currently looking at the logs/metrics but I'm wondering if we could use it with Stackdriver alerting..

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/GceRuntimeMonitor.scala
```diff
@@ -0,0 +1,54 @@
...
+import org.broadinstitute.dsde.workbench.model.TraceId
+import org.broadinstitute.dsde.workbench.model.google.GoogleProject
+
+import scala.concurrent.duration._
+
+/**
+  * Monitor for runtime status transition: Starting, Creating, Deleting, Stopping
+  * It doesn't trigger any of the action but only responsible for monitoring the progress and make necessary cleanup when the transition is done
+  */
+trait GceRuntimeMonitor[F[_]] {
```
rtitle, line 19: Haven't reviewed it all yet, but I'm wondering how much of this is GCE-specific. What would it take to abstract this monitoring approach to Dataproc and GKE?

Should we have something similar to `RuntimeAlgebra[F[_]]` which has interpreters for GCE and Dataproc?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/GceRuntimeMonitor.scala
```diff
@@ -0,0 +1,54 @@
...
+                runtimeAndRuntimeConfig: RuntimeAndRuntimeConfig,
+                operation: com.google.cloud.compute.v1.Operation,
+                action: RuntimeStatus)(implicit ev: ApplicativeAsk[F, TraceId]): F[Unit]
+}
+
+object GceRuntimeMonitor {
+  type CheckResult = (Unit, Option[GcsMonitorState])
+}
+
+sealed abstract class GcsMonitorState extends Product with Serializable
```
rtitle: Did you mean `Gce` instead of `Gcs`?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -234,15 +239,28 @@ class LeoPubsubMessageSubscriber[F[_]: Async: Timer: ContextShift: Concurrent](
...
+        F.raiseError[Unit](
           PubsubHandleMessageError.ClusterInvalidState(msg.runtimeId, runtime.projectNameString, runtime, msg)
         )
-      else Async[F].unit
+      else F.unit
       runtimeConfig <- RuntimeConfigQueries.getRuntimeConfig(runtime.runtimeConfigId).transaction
-      _ <- runtimeConfig.cloudService.interpreter.deleteRuntime(DeleteRuntimeParams(runtime))
+      _ <- if (runtimeConfig.cloudService == CloudService.GCE) {
+        googleComputeService
+          .deleteInstance(runtime.googleProject, config.gceZoneName, InstanceName(runtime.runtimeName.asString))
```
rtitle: I think you should call `CloudService.GCE.interpreter.deleteRuntime()` instead of `googleComputeService.deleteInstance()` -- the GCE interpreter might orchestrate other Google calls as well.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -251,16 +269,33 @@ class LeoPubsubMessageSubscriber[F[_]: Async: Timer: ContextShift: Concurrent](
...
-      _ <- runtimeConfig.cloudService.interpreter
-        .stopRuntime(StopRuntimeParams(RuntimeAndRuntimeConfig(runtime, runtimeConfig), now))
+      _ <- if (runtime.welderEnabled) {
+        welderDAO
+          .flushCache(runtime.googleProject, runtime.runtimeName)
+          .handleErrorWith(e => logger.error(e)(s"Failed to flush welder cache for ${runtime.projectNameString}"))
+      } else F.unit
+      _ <- if (runtimeConfig.cloudService == CloudService.GCE)
+        for {
+          operation <- googleComputeService.stopInstance(runtime.googleProject,
```
rtitle: Same as above -- use the interpreter instead of making Google calls directly

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -323,6 +362,23 @@ class LeoPubsubMessageSubscriber[F[_]: Async: Timer: ContextShift: Concurrent](
             _ <- dbRef.inTransaction(clusterQuery.setToStopping(msg.runtimeId, now))
             _ <- runtimeConfig.cloudService.interpreter
               .stopRuntime(StopRuntimeParams(RuntimeAndRuntimeConfig(runtime, runtimeConfig), now))
+             _ <- if (runtimeConfig.cloudService == CloudService.GCE) //TODO: we should cover dataproc as well in the future
+              // We're running this asynchronously because we don't want to hold up acking event since Stopping an instance can take a while
+               F.runAsync(gceRuntimeMonitor.process(msg.runtimeId).compile.drain) {
```
rtitle: Kinda cool we can express the whole stop -> update -> start transition here without generating separate pubsub messages.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GceInterpreter.scala
```diff
@@ -104,6 +98,7 @@ class GceInterpreter[F[_]: Async: Parallel: ContextShift: Logger](
 
       templateValues = RuntimeTemplateValues(templateParams, Some(ctx.now))
 
+      _ <- Logger[F].info(s"1111 scopes: ${params.scopes.toList}")
```
rtitle: debug msg

rtitle (changes requested):
Pretty cool - I've been wanting to make changes like this for years. ;-)

Still there are a lot of changes here. I made some comments above -- I still want to review the actual monitoring code more closely before merge. 

Also the current approach is very GCE specific. Which is fine to start with but it's worth giving some thought to how Dataproc and GKE would use the new monitoring approach.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
Qi77Qi: Reason why I'm not using `GceInterpreter` is we need an `Operation` back from the `deleteRuntime`, which doesn't apply to current `DataprocInterpreter`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
Qi77Qi: same as `deleteRuntime`, we need an `Operation` back here for polling

http/src/main/resources/reference.conf
Qi77Qi: oops, forgot to remove it from config...I think all the other ones are needed

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpJupyterDAO.scala
Qi77Qi: it still retries since the retry logic is baked in the http client.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
rtitle: I see.. but the interpreter does extra stuff besides `googleComputeService.stopInstance`. It also flushes the welder cache, builds the shutdown script, and sets instance metadata. I don't think we want to duplicate that logic here. Is there a way to make the interpreter return the operation? Or should we just not use Operations?

http/src/main/resources/reference.conf
Qi77Qi: yea...what I have in mind is we'll have an entire different class for dataproc...hence they'll have different configurations. If it turns out both of them have lots of shared code, we can optimize later

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/Config.scala
Qi77Qi: I'm thinking GKE, dataproc, gce will have different monitors...maybe they all have same algebra...

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
Qi77Qi: if later we add GKE, which is probably not tracked in RuntimeConfig, we'll probably need to left join?

> and if not, we probably don't need to process them here

What do you mean? It's used in filtering out dataproc clusters

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
Qi77Qi: we don't...any filter against non index column is slow?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/AutopauseMonitor.scala
Qi77Qi: yea, this is mostly straight copy out of ClusterMonitorSupervisor class. Those unit tests for autopause were living in `ClusterMonitorSupervisorSpec` as well, which is now deleted since all it had is autopause unit tests

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/GceRuntimeMonitor.scala
Qi77Qi: yea...I think eventually we'll have something like `MonitorAlgebra`...but each implementation is completely cloud service specific

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
Qi77Qi: ah I see...I was wondering where welder flush cache logic has moved to...right now, I moved flush welder cache to pubsub function...I'll think about it...I'd still like to use Operation if possible since it does simplify the logic a bit for `delete` and `stop`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterToolMonitor.scala
Qi77Qi: hmmkay...I removed all use of newrelic in this PR, so we should be able to add alerts pretty easily

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ClusterMonitorSupervisor.scala
Qi77Qi: updated

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
rtitle: This query is filtering to GCE-only anyway which should always have a runtimeConfig. So I don't think we need to left join.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
rtitle: Talked offline, I don't think an index is really needed here.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/GceRuntimeMonitorInterp.scala
```diff
@@ -0,0 +1,698 @@
...
+          instance match {
+            case None => deletedRuntime(monitorContext, runtimeAndRuntimeConfig)
+            case _    => checkAgain(monitorContext, runtimeAndRuntimeConfig, Some("Instance hasn't been deleted yet"))
+          }
+        case RuntimeStatus.Starting =>
+          startingRuntime(instance, monitorContext, runtimeAndRuntimeConfig)
+        // This is needed because during boot time, we no longer have reference to the previous delete operation
+        // But this path should only happen during boot and there's on-going delete
+        case RuntimeStatus.Stopping =>
+          val gceStatus = instance.flatMap(i => GceInstanceStatus.withNameInsensitiveOption(i.getStatus))
```
rtitle, line 164: Wondering if the logic in this case could be put into a `stoppingRuntime` method for consistency sake with the other cases

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/GceRuntimeMonitorInterp.scala
```diff
@@ -0,0 +1,698 @@
...
+  private[monitor] def creatingRuntime(
+    instance: Option[Instance],
+    monitorContext: MonitorContext,
+    runtimeAndRuntimeConfig: RuntimeAndRuntimeConfig
+  )(implicit ev: ApplicativeAsk[F, AppContext]): F[CheckResult] = instance match {
+    case None =>
+      checkAgain(monitorContext, runtimeAndRuntimeConfig, Some(s"Can't retrieve instance yet"))
+    case Some(i) =>
+      for {
+        gceStatus <- F.fromEither(GceInstanceStatus.withNameInsensitiveOption(i.getStatus).toRight(new Exception(s"Unknown GCE instance status ${i.getStatus}"))) //TODO: Ideally we should have sentry report this case
```
rtitle, line 195: If there is an unexpected status then the stream will log an error and just stop, right? Agreed about having alerting about this.

I think the old monitor would just restart if an error occurred. Which meant that we had some really old monitor actor instances continually logging errors. I think I like failing fast better (and alerting).

rtitle: Thinking out loud: should we map to `failedRuntime` in this case instead of throwing?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/GceRuntimeMonitorInterp.scala
```diff
@@ -0,0 +1,698 @@
...
+      now <- nowInstant
+      implicit0(traceId: ApplicativeAsk[F, AppContext]) = ApplicativeAsk.const[F, AppContext](AppContext(monitorContext.traceId, now))
+      timeElapsed = (now.toEpochMilli - monitorContext.start.toEpochMilli).millis
+      res <- config.monitorStatusTimeouts.get(runtimeAndRuntimeConfig.runtime.status) match {
+        case Some(timeLimit) if timeElapsed > timeLimit =>
+          for {
+            _ <- logger.info(
+              s"Detected that ${runtimeAndRuntimeConfig.runtime.projectNameString} has been stuck in status ${runtimeAndRuntimeConfig.runtime.status} too long."
+            )
+            r <- // Take care not to Error out a cluster if it timed out in Starting status
```
rtitle, line 399: scalafmt?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/GceRuntimeMonitorInterp.scala
Qi77Qi: `stopRuntime` is called in `pollCheck` as well, and the additional code won't make sense in the case of `pollCheck`...we can potentially have another function wrapping this code, but I kinda prefer to have less additional functions if this code doesn't grow further.

rtitle (approved):
OK approving.. I walked through the new monitor logic and it makes sense to me. It's a nice improvement over the actor stuff, nice work. 👍 

It's a lot of changes but tests are passing (good) and it really only affects GCE.

My one concern is that we're accruing tech debt by making the new monitoring GCE specific -- we should have a tech debt ticket to move Dataproc over as well and delete the legacy `ClusterMonitorActor`/`ClusterMonitorSupervisor` code. We'll probably need to rename some classes at that time and define a monitor algebra that works for both GCE and Dataproc.

Qi77Qi:
@rtitle thanks for reviewing!...for cleaning up tech debt https://broadworkbench.atlassian.net/browse/IA-1832 ...I'm hoping we can get to it once perf tests are done to GCE

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/GceRuntimeMonitorInterp.scala
Qi77Qi: failRuntime will delete the runtime. I'm thinking this case will happen if gce introduces a new status that we don't know about, so I don't think we should blindly delete the runtime

Qi77Qi: oh for previous comment, if this errors, the stream will terminate

```
@ Stream.unfoldLoopEval[IO, String, String]("")(s => if(s.length < 10) IO.raiseError(new Exception("error")) else IO.pure((s, None)))
res6: Stream[IO, String] = Stream(..)

@ res6.compile.toList
res7: IO[List[String]] = Map(
  Bind(Delay(fs2.Stream$CompileOps$$Lambda$2686/1839813660@641c8ba4), fs2.Stream$Compiler$$anon$3$$Lambda$2689/506027903@52667676),
  fs2.Stream$CompileOps$$Lambda$2688/876029877@3fad8636,
  0
)

@ res7.unsafeRunSync
java.lang.Exception: error
  ammonite.$sess.cmd6$.$anonfun$res6$1(cmd6.sc:1)
  fs2.Stream$.$anonfun$unfoldLoopEval$1(Stream.scala:3669)
  fs2.Stream$.$anonfun$unf
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/GceRuntimeMonitorInterp.scala
Qi77Qi: mmm...yea, this is weird formatting 🤷‍♀️

PR #1350, opened 2020-04-07
[IA-1762] [Security] Make CSP policy environment specific

rtitle:
As part of this I decomposed the CSP into separate configs/models. Hopefully this will make it easier to maintain going forward.

Corresponding fc-develop PR: https://github.com/broadinstitute/firecloud-develop/pull/2090

---

In all cases:


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/ContentSecurityPolicyConfig.scala
```diff
@@ -0,0 +1,44 @@
+package org.broadinstitute.dsde.workbench.leonardo.config
```
rtitle, line 1: Unrelated scalafmt

Qi77Qi (commented):
👍 looks great!

Qi77Qi (approved):

PR #1360, opened 2020-04-17
IA-1855 improve subscriber

Qi77Qi:
1. set timeout on the IO that we’re running before ack msgs 
2. process messages from queue with some concurrency

LeoPubsubMessageSubscriber.scala  has main fix

TODO:
~~bump wb-libs with other improvements~~

---

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/Config.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/Config.scala
@@ -353,7 +353,8 @@ object Config {
   implicit val leoPubsubMessageSubscriberConfigReader: ValueReader[LeoPubsubMessageSubscriberConfig] =
     ValueReader.relative { config =>
       LeoPubsubMessageSubscriberConfig(
-        config.getInt("concurrency")
+        config.getInt("concurrency"),
+        toScalaDuration(config.getDuration("timeout"))
       )
     }
 
@@ -422,12 +423,14 @@ object Config {
 
   val pubsubConfig = config.as[PubsubConfig]("pubsub")
   val vpcConfig = config.as[VPCConfig]("vpc")
+  val topic = ProjectTopicName.of(pubsubConfig.pubsubGoogleProject.value, pubsubConfig.topicName)
 
-  val topicName = ProjectTopicName.of(pubsubConfig.pubsubGoogleProject.value, pubsubConfig.topicName)
-  val subscriberConfig: SubscriberConfig =
-    SubscriberConfig(applicationConfig.leoServiceAccountJsonFile.toString, topicName, 1 minute, None)
+  val subscriberConfig: SubscriberConfig = SubscriberConfig(
+    applicationConfig.leoServiceAccountJsonFile.toString,
+    topic,
+    1 minute,
+    None)
 
-  private val topic = ProjectTopicName.of(pubsubConfig.pubsubGoogleProject.value, pubsubConfig.topicName)
   private val retryConfig = GoogleTopicAdminInterpreter.defaultRetryConfig
   val publisherConfig: PublisherConfig =
     PublisherConfig(applicationConfig.leoServiceAccountJsonFile.toString, topic, retryConfig)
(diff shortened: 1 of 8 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/Config.scala
```diff
@@ -350,6 +350,13 @@ object Config {
   implicit val styleSrcReader: ValueReader[StyleSrc] = traversableReader[List, String].map(StyleSrc)
   implicit val connectSrcReader: ValueReader[ConnectSrc] = traversableReader[List, String].map(ConnectSrc)
   implicit val objectSrcReader: ValueReader[ObjectSrc] = traversableReader[List, String].map(ObjectSrc)
+  implicit val leoPubsubMessageSubscriberConfigReader: ValueReader[LeoPubsubMessageSubscriberConfig] =
+    ValueReader.relative { config =>
+      LeoPubsubMessageSubscriberConfig(
+        config.getInt("concurrency"),
+        toScalaDuration(config.getDuration("timeout"))
```
rtitle: Can do `config.as[FiniteDuration]("timeout")`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -52,11 +54,12 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift](
   private[monitor] def messageHandler: Pipe[F, Event[LeoPubsubMessage], Unit] = in => {
     in.evalMap { event =>
       val traceId = event.traceId.getOrElse(TraceId("None"))
-
       val now = Instant.ofEpochMilli(com.google.protobuf.util.Timestamps.toMillis(event.publishedTime))
       implicit val appContext = ApplicativeAsk.const[F, AppContext](AppContext(traceId, now))
       val res = for {
-        res <- messageResponder(event.msg, now).attempt
+        res <- messageResponder(event.msg, now)
+          .timeout(config.timeout)
```
rtitle, line 61: Trying to understand `timeout` semantics. As I understand it if a timeout occurs it will throw `TimeoutException` and will end up matching this case:
```
case _ =>
  logger.error(e)("Fail to process non-retryable pubsub message") >> ack(event)
```

Maybe we should add another case to specifically handle timeouts.

Also if a timeout occurs is the underlying `IO` cancelled or does it keep running? Thinking 55 seconds might be too short to fail cluster creations if some Google calls get retried.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -81,9 +84,11 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift](
     }
   }
 
-  val process: Stream[F, Unit] =
-    (Stream.eval(logger.info(s"starting subscriber ${subscriberConfig.projectTopicName}")) ++ (subscriber.messages through messageHandler))
-      .handleErrorWith(error => Stream.eval(logger.error(error)("Failed to initialize message processor")))
+  val process: Stream[F, Unit] = (subscriber.messages through messageHandler)
```
rtitle, line 87: Cool

rtitle (commented):
See above.. I like the concurrency change, kind of want to understand the `timeout` semantics better to make sure it doesn't cause issues for us.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
Qi77Qi: we've been using 1 min as subscriber's ack timeout. What do you think should happen if time out happens other than logging the error and acking the message?

We can bump both the subscriber's ack deadline and this timeout if 1 min is not enough.

According to cats-effect doc, the underlying IO is canceled in the event of time out

rtitle: Yeah agreed, not much we can do besides log the error. Maybe trigger a stackdriver alert?

I think 1 minute is probably too short -- a single Google call can retry for more than 1 minute.  Maybe 5 minutes?

Qi77Qi: yea, I wish we've set up the ability to error reporting (like if we have sentry set up properly, anything logged as `error` will go to alert). Added https://broadworkbench.atlassian.net/browse/IA-1857.

I'll up the ack deadline

rtitle (approved):

PR #1362, opened 2020-04-21
[IA-1823] [IA-1852] Integrate RStudio Pro license file

rtitle:
https://broadworkbench.atlassian.net/browse/IA-1823
https://broadworkbench.atlassian.net/browse/IA-1852

This PR adds templating logic for the RStudio Pro license file. It also fixes some proxy configs to make it work with RSP.

This PR should be mergeable before adding the actual file to vault/firecloud-develop.

Tested in a fiab + local Terra. All of the images work:
- Jupyter (gatk image)
- RStudio open source (anvil image)
- RStudio Pro (https://github.com/DataBiosphere/terra-docker/pulls/111)
   - To test I manually added the temp license file to `/etc` on my fiab
   - With no license file the RSP UI displays an error

---

In all cases:


http/src/main/resources/init-resources/cluster-site.conf
```diff
@@ -21,15 +25,15 @@
     RewriteCond %{REQUEST_URI} /proxy/[^/]*/[^/]*/rstudio/.* [NC]
     RewriteRule /proxy/[^/]*/[^/]*/rstudio/(.*) http://127.0.0.1:8001/$1 [P,L]
 
-    ProxyPass / http://127.0.0.1:8001/ retry=10
```
rtitle, line 24: Turns out we didn't need `ProxyPass` since we have `RewriteRule`s defined for each proxy path. 

`ProxyPassReverse` is needed for RStudio but not Jupyter.

http/src/main/resources/init-resources/gce-init.sh
```diff
@@ -213,6 +215,18 @@ END
       gsutil cp ${CUSTOM_ENV_VARS_CONFIG_URI} /etc
     fi
 
+    # Install RStudio license file, if specified
+    if [ ! -z "$RSTUDIO_DOCKER_IMAGE" ] ; then
+      STAT_EXIT_CODE=0
+      gsutil -q stat ${RSTUDIO_LICENSE_FILE} || STAT_EXIT_CODE=$?
```
rtitle, line 224: Tolerate the license not being there, so this PR can be merged ahead of fc-develop

rtitle:
RStudioSpec failed but it works in my fiab 🤔… jenkins retest

rtitle:
Tests passed, manual testing looks good too :)

rtitle:
fc-develop PR: https://github.com/broadinstitute/firecloud-develop/pull/2105

http/src/main/resources/init-resources/gce-init.sh
Qi77Qi: can we add a TODO for removing this once fc-develop PR is merged?

http/src/main/resources/init-resources/init-actions.sh
```diff
@@ -210,6 +212,18 @@ END
       echo "" > /etc/google_application_credentials.env
     fi
 
+    # Install RStudio license file, if specified
+    if [ ! -z "$RSTUDIO_DOCKER_IMAGE" ] ; then
```
Qi77Qi, line 217: if rstudio is not gonna use spark, we probably don't need this?

http/src/main/resources/init-resources/proxy-docker-compose.yaml
```diff
@@ -13,3 +13,6 @@ services:
     environment:
       HTTPD_PORT: '80'
       SSL_HTTPD_PORT: '443'
+      GOOGLE_PROJECT: "${GOOGLE_PROJECT}"
+      CLUSTER_NAME: "${CLUSTER_NAME}"
```
Qi77Qi: should this be `RUNTIME_NAME`?

http/src/main/resources/init-resources/rstudio-docker-compose.yaml
```diff
@@ -16,8 +16,10 @@ services:
       WORKSPACE_NAMESPACE: "${GOOGLE_PROJECT}"
       CLUSTER_NAME: "${CLUSTER_NAME}"
       OWNER_EMAIL: "${OWNER_EMAIL}"
+    privileged: true
```
Qi77Qi: just for my education, what does this do?

http/src/main/resources/reference.conf
```diff
@@ -191,11 +191,11 @@ clusterResources {
 }
 
 clusterFiles {
-  configFolderPath = "/etc/"
-  jupyterServerCrt = "jupyter-server.crt"
-  jupyterServerKey = "jupyter-server.key"
-  jupyterRootCaPem = "rootCA.pem"
-  jupyterRootCaKey = "rootCA.key"
+  proxyServerCrt = "/etc/jupyter-server.crt"
```
Qi77Qi, line 194: 👍  on this change

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/ProxyConfig.scala
```diff
@@ -9,4 +11,9 @@ case class ProxyConfig(proxyDomain: String,
                        tokenCacheExpiryTime: FiniteDuration,
                        tokenCacheMaxSize: Int,
                        internalIdCacheExpiryTime: FiniteDuration,
-                       internalIdCacheMaxSize: Int)
+                       internalIdCacheMaxSize: Int) {
+  def getProxyServerHostName: String = {
+    val url = new URL(proxyUrlBase)
+    s"${url.getProtocol}://${url.getHost}" + (if (url.getPort == -1) "" else s":${url.getPort.toString}")
```
Qi77Qi: when would it be `-1`?

http/src/main/resources/init-resources/gce-init.sh
rtitle: Will do.. was thinking if there's ever a case where we still want to support RStudio _without_ a license (in which case we'd want to keep this). But I don't think so.

http/src/main/resources/init-resources/init-actions.sh
rtitle: We currently support RStudio on Dataproc so I think it makes sense to install the license too.. we could decide in the future to remove RStudio support from Dataproc. Although maybe there are some R tool which use Spark 🤷‍♂️

http/src/main/resources/init-resources/rstudio-docker-compose.yaml
rtitle: Hm, it gives the container deeper control of all VM devices. It was in the example docker-compose RStudio sent. I also see this:

https://github.com/rstudio/rstudio-docker-products/blob/3f22fbff6c35f185253b507746187a8f07023e1c/README.md#privileged-containers

I can see if things work without it, and also potentially ask RSP team if it's really necessary.

Also cc @adrazhi, do we have concerns about running docker containers in user VMs in `privileged` mode?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/ProxyConfig.scala
rtitle: From javadoc:
```
    /**
     * Gets the port number of this {@code URL}.
     *
     * @return  the port number, or -1 if the port is not set
     */
```

It's `-1` if the URL string is `https://notebooks.firecloud.org`. But it's set if the URL is `https://leonardo-fiab.dsde-dev.broadinstitute.org:30443` (I think fiab is the only time we explicitly set a port).

http/src/main/resources/init-resources/rstudio-docker-compose.yaml
rtitle: I actually removed it for now

http/src/main/resources/init-resources/init-actions.sh
Qi77Qi: I see...UI still needs to make decision which runtime to create, are we creating GCE VMs for RStudio by default?

Qi77Qi (approved):

http/src/main/resources/init-resources/init-actions.sh
rtitle: Yeah, IIRC it defaults to GCE VM but the user can still choose to use Spark

rtitle:
jenkins retest

http/src/main/resources/init-resources/rstudio-docker-compose.yaml
rtitle: Leaving `--privileged` out of this PR, opened https://broadworkbench.atlassian.net/browse/IA-1868 for future investigation

adrazhi: > just for my education, what does this do?

@Qi77Qi  You can read about it more [here](https://dsp-security.broadinstitute.org/platform-security-categories/docker/attacking-dockers/capabilities#capabilities). In short it's never really a good idea as it gives the container root CAPABILITIES to the host machine.

PR #1365, opened 2020-04-22
[IA-1583] Client code for Sam persistent disk resource type

akarukappadath:
See Sam side of things: https://github.com/broadinstitute/sam/pull/422

Note: I based much of this work off of what already exists for the notebook-cluster resource. However, I noticed that many of these functions pass around a RuntimeName without actually using it, so I elected not to pass around a DiskName for the functions that deal with PDs. 

---

In all cases:


core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/runtimeModels.scala
```diff
@@ -379,6 +379,8 @@ final case class OperationName(value: String) extends ValueObject
 final case class GoogleOperation(name: OperationName, id: GoogleId)
 final case class GoogleId(value: String) extends AnyVal
 
+final case class PersistentDiskInternalId(asString: String) extends AnyVal
```
Qi77Qi, line 382: shall we name this `PersistentDiskSamId`? `internalId` is pretty confusing name

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/IamProxyAuthProvider.scala
```diff
@@ -168,4 +182,35 @@ class IamProxyAuthProvider(config: Config, saProvider: ServiceAccountProvider[Fu
...
   )(implicit ev: ApplicativeAsk[Future, TraceId]): Future[Unit] = Future.unit
 
+  /**
+    * Leo calls this method to notify the auth provider that a new persistent disk has been created.
+    * The returned future should complete once the provider has finished doing any associated work.
+    * Leo will wait, so be timely!
+    *
+    * @param internalId     The internal ID for the disk (i.e. used for Sam resources)
+    * @param creatorEmail   The email address of the user in question
+    * @param googleProject  The Google project the disk was created in
```
akarukappadath, line 192: Note: For the functions that deal with runtimes, a RuntimeName is being passed around but not used, so I elected not to pass around a DiskName for the ones that deal with PDs

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeoAuthProvider.scala
```diff
@@ -24,6 +25,15 @@ object NotebookClusterActions {
   val allActions = Seq(GetClusterStatus, ConnectToCluster, SyncDataToCluster, DeleteCluster, StopStartCluster)
 }
 
+object PersistentDiskActions {
+  sealed trait PersistentDiskAction extends LeoAuthAction
```
Qi77Qi: can we move `sealed trait PersistentDiskAction extends LeoAuthAction` outside the object itself.
For `allActions`, can you use `sealerate`
```
import ca.mrvisser.sealerate
sealerate.values[PersistentDiskActions]
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeoAuthProvider.scala
```diff
@@ -101,4 +125,35 @@ abstract class LeoAuthProvider[F[_]] {
...
+    * @param googleProject  The Google project the disk was created in
+    * @return A Future that will complete when the auth provider has finished doing its business.
+    */
+  def notifyPersistentDiskCreated(internalId: PersistentDiskInternalId,
+                                  creatorEmail: WorkbenchEmail,
+                                  googleProject: GoogleProject)(implicit ev: ApplicativeAsk[F, TraceId]): F[Unit]
+
+  /**
+    * Leo calls this method to notify the auth provider that a persistent disk has been deleted.
+    * The returned future should complete once the provider has finished doing any associated work.
```
Qi77Qi, line 160: comments should be updated if we want to keep them

Qi77Qi (commented):
some minor comments..but looks great overall 👍

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/runtimeModels.scala
akarukappadath: I based it off what we have for Runtime - When we create the PersistentDisk type in Leo, I'm guessing the object will need an `internalId` field the same way as `Runtime`. I'd vote for keeping things consistent between the two, so if we want to change to something like `PersistentDiskSamId` I can create a ticket to change the `internalId` in the `Runtime` object to something like `RuntimeSamId` as well

rtitle: btw I'm creating `persistentDiskModels.scala` with some PD models for IA-1756. I can move this class in my PR.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/sam/SamAuthProvider.scala
```diff
@@ -72,12 +73,25 @@ class SamAuthProvider[F[_]: Effect: Logger](samDao: SamDAO[F],
...
+
   private val projectActionMap: Map[LeoAuthAction, String] = Map(
     GetClusterStatus -> "list_notebook_cluster",
     CreateClusters -> "launch_notebook_cluster",
     SyncDataToCluster -> "sync_notebook_cluster",
     DeleteCluster -> "delete_notebook_cluster",
-    StopStartCluster -> "stop_start_notebook_cluster"
+    StopStartCluster -> "stop_start_notebook_cluster",
+    CreatePersistentDisk -> "create_persistent_disk",
+    ModifyPersistentDisk -> "modify_persistent_disk",
```
rtitle: Didn't we get rid of `modify` at this project level… should this be `list_persistent_disk`?

rtitle (changes requested):
Overall looks good.. one comment above about `modify` at the project level.

Also would be good to add some unit tests to `SamAuthProviderSpec` to cover the added methods. (We also have `AuthProviderSpec` but that test is a bit of a mess -- I'd be ok with removing that honestly.)

akarukappadath:
jenkins retest

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpSamDAO.scala
```diff
@@ -179,6 +179,91 @@ class HttpSamDAO[F[_]: Effect](httpClient: Client[F], config: HttpSamDaoConfig,
...
+    * The returned future should complete once the provider has finished doing any associated work.
+    * Returning a failed Future will prevent the disk from being created, and will call notifyPersistentDiskDeleted for the same disk.
+    * Leo will wait, so be timely!
+    *
+    * @param internalId     The internal ID for the disk (i.e. used for Sam resources)
+    * @param creatorEmail   The email address of the user in question
+    * @param googleProject  The Google project the disk was created in
+    * @return A Future that will complete when the auth provider has finished doing its business.
+    */
+  def createPersistentDiskResource(internalId: PersistentDiskInternalId,
```
rtitle, line 194: Is the only difference between `create/deletePersistentDiskResource` and `create/deleteClusterResource` the Sam URL? Maybe some code could be shared.

rtitle (approved):
Looks good! The permission model seems correct to me. Nice work

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpSamDAO.scala
akarukappadath: If as part of your PD data model PR you're refactoring some types out to be used both by PDs and runtimes, we can add a `SamResourceId` trait there, and then `PersistentDiskSamResourceId` and `RuntimeSamResourceId` can both extend from that - I think that would help in making this code more easily sharable between resource types

rtitle: Yeah  good point -- I think we can merge this PR as is and then I can rebase my PR and refactor the models.

PR #1372, opened 2020-04-23
[IA-1801] update swagger

jdcanas:
Good thing we had this, these entries were not very accurate. I copied directly from the encoder and tested the calls myself. 

Some other entries may be out of date, but I tried to clean up what I could think of and what is most relevant to consumers.

I'm not too sure why ListRuntime and ListCluster are so different...

### DIFF
--- a/http/src/main/resources/swagger/api-docs.yaml
+++ b/http/src/main/resources/swagger/api-docs.yaml
@@ -1574,6 +1574,7 @@ components:
         - runtimeName
         - googleProject
         - serviceAccount
+        - auditInfo
         - runtimeConfig
         - status
         - labels
@@ -1582,8 +1583,8 @@ components:
         - scopes
       properties:
         id:
-          type: string
-          description: The internally-referenced ID of the runtime
+          type: long
+          description: The leonardo ID of the runtime
         runtimeName:
           type: string
           description: The user-supplied name for the runtime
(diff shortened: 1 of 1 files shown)


http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -1582,8 +1583,8 @@ components:
         - scopes
       properties:
         id:
-          type: string
-          description: The internally-referenced ID of the runtime
+          type: long
```
Qi77Qi, line 1586: 👍

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -1694,47 +1742,36 @@ components:
...
+        - runtimeConfig
         - status
         - labels
-        - autopauseThreshold
-        - defaultClientId
       properties:
-        internalId:
-          type: string
-          description: Internal resource ID of the runtime
+        id:
```
Qi77Qi, line 1753: we have both `id` and `internalId`, just want to double check that this is returning `id` but not `internalId`?

Qi77Qi (commented):
one question...otherwise lgtm 👍

http/src/main/resources/swagger/api-docs.yaml
jdcanas: Yeah, if you trace it, it corresponds to this encoder https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/RuntimeRoutes.scala#L379

verified here:
Screen Shot 2020-04-23 at 5 55 12 PM

Qi77Qi (approved):

PR #1381, opened 2020-04-28
[IA-1756] PD data model

rtitle:
Data model, slick mapping, and liquibase migration for PDs.

See:
https://broadworkbench.atlassian.net/browse/IA-1756
https://broadworkbench.atlassian.net/wiki/spaces/IA/pages/398295082/2020-04-22+Data+Model+for+Persistent+Disks

---

In all cases:


core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodec.scala
```diff
@@ -111,6 +112,14 @@ object JsonCodec {
     "errorCode",
     "timestamp"
   )(x => RuntimeError.unapply(x).get)
+  implicit val diskStatusEncoder: Encoder[DiskStatus] = Encoder.encodeString.contramap(_.toString)
+  implicit val diskAuditInfoEncoder: Encoder[DiskAuditInfo] = Encoder.forProduct4(
```
Qi77Qi: this looks about the same as existing AuditInfo, can we reuse the same case class?

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/diskModels.scala
```diff
@@ -1,6 +1,62 @@
 package org.broadinstitute.dsde.workbench.leonardo
 
+import java.time.Instant
+
+import enumeratum.{Enum, EnumEntry}
+import org.broadinstitute.dsde.workbench.google2.{DiskName, ZoneName}
+import org.broadinstitute.dsde.workbench.model.WorkbenchEmail
+import org.broadinstitute.dsde.workbench.model.google.GoogleProject
+
+final case class PersistentDisk(id: Long,
```
Qi77Qi: can we use a `DiskId` type, just so we will never mix up runtime id with this?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/PersistentDiskComponent.scala
```diff
@@ -0,0 +1,150 @@
...
+                                      googleProject: GoogleProject,
+                                      zone: ZoneName,
+                                      name: DiskName,
+                                      googleId: Option[GoogleId],
+                                      samResourceId: DiskSamResourceId,
+                                      status: DiskStatus,
+                                      creator: WorkbenchEmail,
+                                      createdDate: Instant,
+                                      destroyedDate: Instant,
+                                      dateAccessed: Instant,
```
Qi77Qi, line 26: can we move these fields into `AuditInfo`?

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/CommonTestData.scala
```diff
@@ -287,6 +291,21 @@ object CommonTestData {
     instance.copy(key = modifyInstanceKey(instance.key), googleId = instance.googleId + 1)
   def modifyInstanceKey(instanceKey: DataprocInstanceKey): DataprocInstanceKey =
     instanceKey.copy(name = InstanceName(instanceKey.name.value + "_2"))
+
+  def makePersistentDisk(id: Int): PersistentDisk = PersistentDisk(
```
Qi77Qi: shall we start making scalacheck generators instead?

Qi77Qi (commented):
👍 lgtm in general..left some minor comments

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodec.scala
rtitle: That would be nice… I noticed that the existing `AuditInfo` had 1 field that's not relevant to disks. Will see if I can refactor so we can reuse it.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/PersistentDiskComponent.scala
rtitle: We could.. but this lets us write `<> (PersistentDiskRecord.tupled, PersistentDiskRecord.unapply)` below. I kind of see this class as a direct mapping to columns in the database, whereas the `PersistentDisk` class has more structure.

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/CommonTestData.scala
rtitle: Hmm.. yeah maybe

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodec.scala
jdcanas: I'm running into the same thing with kubernetes right now. I think creator, createdDate, destroyedDate are the core fields between all 3 use cases

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/PersistentDiskComponent.scala
```diff
@@ -0,0 +1,150 @@
...
+import org.broadinstitute.dsde.workbench.model.google.GoogleProject
+
+import scala.concurrent.ExecutionContext
+
+final case class PersistentDiskRecord(id: Long,
+                                      googleProject: GoogleProject,
+                                      zone: ZoneName,
+                                      name: DiskName,
+                                      googleId: Option[GoogleId],
+                                      samResourceId: DiskSamResourceId,
```
jdcanas, line 21: why do we use this here, but not do the same for GCE runtimes/Dataproc clusters?

rtitle: You mean `DiskSamResourceId`? Runtimes have a field for this but it's called `RuntimeInternalId` which IMO is a bit non-descriptive because it's only used for Sam IDs.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/PersistentDiskComponent.scala
Qi77Qi: I don't feel strongly either way..But if we see this as direct mapping, it's easy to run into 22 fields limit tho.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/PersistentDiskComponent.scala
Qi77Qi: I'm wondering, is it reasonable to make a `SamResourceId` data type and use it for all different sam resources

rtitle: I thought about that, was thinking it's good to have type safety to not confuse runtime and disk sam IDs. 

Maybe we could add an ADT like:
```
sealed trait SamResource {
  def resourceId: String
  def resourceType: SamResourceType
}
object SamResource {
  final case class RuntimeSamResource(resourceId: String) {
    override def resourceType: SamResourceType.Runtime
  }
  final case class PersistentDiskSamResource(resourceId: String) {
    override def resourceType: SamResourceType.PersistentDisk
  }
}
```

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20200427_create_persistent_disk_table.xml
```diff
@@ -0,0 +1,71 @@
...
+                <constraints nullable="false"/>
+            </column>
+        </createTable>
+        <addUniqueConstraint columnNames="googleProject, name, destroyedDate" constraintName="IDX_PERSISTENT_DISK_UNIQUE" tableName="PERSISTENT_DISK"/>
+        <createIndex indexName="IDX_PERSISTENT_DISK_STATUS" tableName="PERSISTENT_DISK">
+            <column name="status"/>
+        </createIndex>
+    </changeSet>
+    <changeSet logicalFilePath="leonardo" author="rtitle" id="persistent-disk-label-table">
+        <createTable tableName="PERSISTENT_DISK_LABEL">
```
rtitle: Thinking more, I think I'm going to try and repurpose the existing `LABEL` table instead of adding a new `PERSISTENT_DISK_LABEL`

jdcanas: ➕ I was starting to write a comment along the lines of this

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/PersistentDiskComponent.scala
```diff
@@ -0,0 +1,150 @@
...
+                                      status: DiskStatus,
+                                      creator: WorkbenchEmail,
+                                      createdDate: Instant,
+                                      destroyedDate: Instant,
+                                      dateAccessed: Instant,
+                                      size: DiskSize,
+                                      diskType: DiskType,
+                                      blockSize: BlockSize)
+
+class PersistentDiskTable(tag: Tag) extends Table[PersistentDiskRecord](tag, "PERSISTENT_DISK") {
```
jdcanas, line 31: I think you need to define an `index` to correspond to `IDX_PERSISTENT_DISK_UNIQUE`

rtitle: We typically haven't represented all our indexes in the slick mapping code. The indexes are still present in the database. IIRC they are only useful in slick when generating the schema from slick code, which we don't do

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/PersistentDiskComponent.scala
rtitle: So I started down this path and it led to a lot of other refactoring… I still think it's valuable but I think I will split it into a separate PR, so this can be merged sooner. Will still apply the other feedback in this PR.

rtitle: Opened https://broadworkbench.atlassian.net/browse/IA-1893

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/diskModels.scala
```diff
@@ -1,6 +1,55 @@
...
+                                zone: ZoneName,
+                                name: DiskName,
+                                googleId: Option[GoogleId],
+                                samResourceId: DiskSamResourceId,
+                                status: DiskStatus,
+                                auditInfo: AuditInfo,
+                                size: DiskSize,
+                                diskType: DiskType,
+                                blockSize: BlockSize,
+                                labels: LabelMap)
```
akarukappadath, line 18: Since we have a `Failed` disk status, should we have an `errors` field here?

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20200427_create_persistent_disk_table.xml
akarukappadath: If disks can get errored out, I wonder if we should do the same for the `CLUSTER_ERROR` table

akarukappadath (approved):
Looks good to me - just a question about PD errors

Qi77Qi (approved):
lgtm

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20200427_create_persistent_disk_table.xml
rtitle: That's an excellent question… will think about it but probably punt to a separate PR if so.

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/diskModels.scala
rtitle: Yeah, I didn't make `errors` at all.. was thinking of punting that till we have more understanding of what errors can occur.

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/runtimeModels.scala
```diff
@@ -25,6 +25,7 @@ final case class Runtime(id: Long,
                          serviceAccount: WorkbenchEmail,
                          asyncRuntimeFields: Option[AsyncRuntimeFields],
                          auditInfo: AuditInfo,
+                         kernelFoundBusyDate: Option[Instant],
```
rtitle, line 28: Moved `kernelFoundBusyDate` out of `AuditInfo` and into `Runtime`

PR #1382, opened 2020-04-28
[no ticket] Fix flakey LeoPubsubMessageSubscriberSpec

rtitle:
I think there was a race condition in the tests for `handleUpdateRuntimeMessage` because the stop-start happens asynchronously. This PR adds tests for the following:

1. process `UpdateRuntimeMessage`, transitions to `Stopping`, doesn't do the update yet
2. process `UpdateRuntimeMessage`, transition from `Stopping` -> `Stopped` -> `Starting`, makes the update
3. process `UpdateRuntimeMessage`, does not need to transition, makes the update immediately

---

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
@@ -389,10 +389,7 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift](
             }
           } yield ()
         } else {
-          for {
-            _ <- runtimeConfig.cloudService.interpreter.updateMachineType(UpdateMachineTypeParams(runtime, m, now))
-            _ <- RuntimeConfigQueries.updateMachineType(runtime.runtimeConfigId, m, now).transaction
-          } yield ()
+          runtimeConfig.cloudService.interpreter.updateMachineType(UpdateMachineTypeParams(runtime, m, now))
         }
       }
     } yield ()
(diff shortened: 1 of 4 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -386,10 +389,7 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift](
             }
           } yield ()
         } else {
-          for {
-            _ <- runtimeConfig.cloudService.interpreter.updateMachineType(UpdateMachineTypeParams(runtime, m, now))
-            _ <- RuntimeConfigQueries.updateMachineType(runtime.runtimeConfigId, m, now).transaction
```
rtitle, line 391: I don't think we needed to update the DB here because `runtimeConfig.cloudService.interpreter.updateMachineType` already does that

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriberSpec.scala
```diff
@@ -268,15 +276,56 @@ class LeoPubsubMessageSubscriberSpec
...
+      // machine type should not be updated yet
+      updatedRuntimeConfig shouldBe 'defined
+      updatedRuntimeConfig.get.machineType shouldBe MachineTypeName("n1-standard-4")
     }
 
     res.unsafeRunSync()
   }
 
-  it should "handle UpdateRuntimeMessage and update the runtime config in the database" in isolatedDbTest {
+  it should "handle UpdateRuntimeMessage and go through a stop-start transition" in isolatedDbTest {
```
akarukappadath, line 294: Is it worth having both of these two tests? It seems like the first one is checking to see that a machine-type change stops the runtime, and this one checks to see that it is eventually starting later - would just the starting test suffice, since a runtime can't be in starting status without having been stopped?

akarukappadath (approved):
lgtm! Just left a comment about potential redundancy. Thanks for looking into this!

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriberSpec.scala
rtitle: Yeah I thought about that and I'm on the fence.. I kind of like that the `Stopping` test verifies that the machine type change has NOT been made yet.

jdcanas (approved):

jdcanas:
Thanks for digging into this. A level of comprehensiveness in addressing this and improving the code at the same time to aspire to 😃

PR #1392, opened 2020-04-29
Minimum disk size

gpcarr:
Not ready for review. WIP. Setting up this draft PR to run unit tests until I have a computer.

---

In all cases:


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeServiceInterp.scala
```diff
@@ -395,6 +414,11 @@ class RuntimeServiceInterp[F[_]: Parallel](config: RuntimeServiceConfig,
     }
   }
 
+  private[service] def validateMinimumDiskSize(diskSize: DiskSize): F[Unit] = {
```
Qi77Qi, line 417: we should just need to update https://github.com/DataBiosphere/leonardo/blob/develop/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodec.scala#L136...decoding error will result in bad request response

gpcarr: oh man lol i complicated this way more than i needed to. thanks for the headups, going to start a new branch and make this small change!

PR #1394, opened 2020-04-29
IA-1299 Enforce minimum disk size of 50

gpcarr:
When we decode both GCE and Dataproc config, we are now enforcing a minimum disk size of 50GB. These changes reflect that and add two unit tests for validation.
---

In all cases:

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodec.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodec.scala
@@ -19,6 +19,7 @@ import org.broadinstitute.dsde.workbench.model.google.{
 object JsonCodec {
   // Errors
   val negativeNumberDecodingFailure = DecodingFailure("Negative number is not allowed", List.empty)
+  val minimumDiskSizeDecodingFailure = DecodingFailure("Minimum required disk size is 50GB", List.empty)
   val oneWorkerSpecifiedDecodingFailure = DecodingFailure(
     "Google Dataproc does not support clusters with 1 non-preemptible worker. Must be 0, 2 or more.",
     List.empty
@@ -134,7 +135,7 @@ object JsonCodec {
   implicit val urlDecoder: Decoder[URL] =
     Decoder.decodeString.emap(s => Either.catchNonFatal(new URL(s)).leftMap(_.getMessage))
   implicit val diskSizeDecoder: Decoder[DiskSize] =
-    Decoder.decodeInt.emap(d => if (d < 50) Left("Minimum disk size is 50GB.") else Right(DiskSize(d)))
+    Decoder.decodeInt.emap(d => if (d < 50) Left("Minimum required disk size is 50GB") else Right(DiskSize(d)))
   implicit val workbenchEmailDecoder: Decoder[WorkbenchEmail] = Decoder.decodeString.map(WorkbenchEmail)
   implicit val runtimeImageTypeDecoder: Decoder[RuntimeImageType] = Decoder.decodeString.emap(s =>
     RuntimeImageType.stringToRuntimeImageType.get(s).toRight(s"invalid RuntimeImageType ${s}")
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodec.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodec.scala
@@ -134,7 +134,7 @@ object JsonCodec {
   implicit val urlDecoder: Decoder[URL] =
     Decoder.decodeString.emap(s => Either.catchNonFatal(new URL(s)).leftMap(_.getMessage))
   implicit val diskSizeDecoder: Decoder[DiskSize] =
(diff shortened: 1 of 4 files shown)


core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodec.scala
```diff
@@ -134,7 +134,7 @@ object JsonCodec {
   implicit val urlDecoder: Decoder[URL] =
     Decoder.decodeString.emap(s => Either.catchNonFatal(new URL(s)).leftMap(_.getMessage))
   implicit val diskSizeDecoder: Decoder[DiskSize] =
-    Decoder.decodeInt.emap(d => if (d < 0) Left("Negative number is not allowed") else Right(DiskSize(d)))
+    Decoder.decodeInt.emap(d => if (d < 50) Left("Minimum disk size is 50GB.") else Right(DiskSize(d)))
```
gpcarr: @Qi77Qi would we want to keep the "Negative number is not allowed" error and add a new one for minimum disk size or replace it like I've done in this PR? looking at fixing unit tests now and want to make sure I don't get rid of functionality that we may still want

Qi77Qi: I think just erroring min disk is 50GB is clear enough...thanks for the quick fix 👍 ...maybe reword this to `Minimum required disk size is 50GB.`?

Just to double check, this is true for GCE VMs runtimes as well? cc @rtitle  for confirmation

Qi77Qi (approved):

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodec.scala
gpcarr: Sounds good, will update! thank you :)

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/api/LeoRoutesJsonCodecSpec.scala
```diff
@@ -67,7 +67,24 @@ class LeoRoutesJsonCodecSpec extends FlatSpec with Matchers {
       json <- io.circe.parser.parse(inputString)
       r <- json.as[RuntimeConfigRequest.DataprocConfig]
     } yield r
-    decodeResult.leftMap(_.getMessage) shouldBe Left("Negative number is not allowed: DownField(masterDiskSize)")
+    decodeResult.leftMap(_.getMessage) shouldBe Left("Minimum required disk size is 50GB: DownField(masterDiskSize)")
+  }
+
+  it should "fail with minimumDiskSizeDecodingFailure when masterDiskSize is less than 50" in {
```
Qi77Qi: this is fine too...I would probably add the test in https://github.com/DataBiosphere/leonardo/blob/develop/core/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodecSpec.scala#L8 for just `diskSizeDecoder`

gpcarr: I added `minimumDiskSizeDecodingFailure`  because we use the `negativeNumberDecodingFailure` in other places. and good point, I'll move my test now!

rtitle (approved):
looks good to me!

rtitle:
btw we usually put the JIRA ticket number in the PR description -- we have a process to generate release notes based on commit logs.

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/api/LeoRoutesJsonCodecSpec.scala
gpcarr: @Qi77Qi I moved my unit test to `JsonCodecSpec` and have a unit test for a GCE config and a Dataproc config!

gpcarr:
> btw we usually put the JIRA ticket number in the PR description -- we have a process to generate release notes based on commit logs.

Updated the PR description :)

core/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodecSpec.scala
```diff
@@ -49,4 +53,39 @@ class JsonCodecSpec extends LeonardoTestSuite with Matchers with FlatSpecLike {
...
+      """
+        |{
+        |   "machineType": "n1-standard-8",
+        |   "diskSize": 35,
+        |   "cloudService": "GCE"
+        |}
+        |""".stripMargin
+
+    val res = decode[RuntimeConfig](inputString)
+    res.left shouldBe LeftProjection(Left(DecodingFailure("Minimum required disk size is 50GB", List(DownField("diskSize")))))
```
Qi77Qi, line 68: btw, you should be able to do the check like this
```
res shouldBe Left(DecodingFailure("Minimum required disk size is 50GB", List(DownField("diskSize"))))
```

PR #1395, opened 2020-04-30
IA-1832 Replace ClusterMonitorActor

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-1832

1. Remove all actor, actor system usage
- `ClusterMonitorActor` is replaced with `RuntimeMonitor`
- `DateAccessActor` is replaced with `DateAccessedUpdater`. Optimized the updater a bit by grouping all messages first, and we only apply the latest update for the same runtime

2. Use google2 dataproc lib in monitoring code. Related PR https://github.com/broadinstitute/workbench-libs/pull/300

3. Made all config value reader implicits private so that it can't be used outside Config.scala

4. Introduced `AsyncTaskProcessor`, whose sole responsibility is to run IOs asynchronously. This is helpful because previously in LeoPubsubSubscriber we had a few places we needed to do `F.runAsync(xxx)` just to make the monitoring part asynchronously to message acking.

5. Add logic for terminating a monitoring process if the db status is changed (for instance when a new status transition request comes in)

6. Fix a bug where welder is not being shut down properly

Screen Shot 2020-05-01 at 7 48 39 PM

Feels okay with new code test coverage...not sure why that 15% drop since I didn't remove any unit tests for `GceRuntimeMonitor`

TODO:
~~1. Replace wb-lib version to published one~~
~~2. Add async task processor~~
~~3. Map RuntimeStatus in contract~~

---

In all cases:


http/src/main/resources/reference.conf
```diff
@@ -67,10 +83,17 @@ gce {
       creating = 30 minutes
       starting = 20 minutes
       deleting = 30 minutes
+      stopping = 30 minutes
```
rtitle: What's the behavior if a runtime times out in `Stopping` status? I think I avoided adding a timeout for this case because I didn't want Leo to auto-delete the runtime. But probably good to at least alert if this happens.

Qi77Qi: ah good point..I think I had a comment somewhere wondering if this is safe to do...maybe I'll treat stopping time out differently if it's easy to do. Otherwise I might just leave this out for now since the reason I did this is because I saw clusters stuck in `Stopping` in perf tests a few times, but I think it's probably related to https://broadinstitute.slack.com/archives/GB1VAM14J/p1588953652207800

http/src/main/resources/reference.conf
```diff
@@ -383,3 +392,8 @@ ui {
   terraLabel = "saturnAutoCreated"
   allOfUsLabel = "all-of-us"
 }
+
+async-task-processor {
```
rtitle, line 402: Just a nit: I notice that all our other configs are camelCase :)

Qi77Qi: yea...I've been doing camelCase in leonardo as well...I just thought if we're serious about casing, I think it's recommended to do hyphen-separated style, see https://github.com/lightbend/config/blob/master/HOCON.md#hyphen-separated-vs-camelcase... also this is default style pureconfig uses, which I'm thinking maybe one day we'll move over to pureconfig :P

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/InstanceComponent.scala
```diff
@@ -107,7 +107,7 @@ object instanceQuery extends TableQuery(new InstanceTable(_)) {
       googleId = BigDecimal(instance.googleId),
       status = instance.status.entryName,
       ip = instance.ip.map(_.value),
-      dataprocRole = Some(instance.dataprocRole.entryName),
+      dataprocRole = Some(instance.dataprocRole.toString),
```
rtitle, line 110: Just checking, does this change the DB contract at all? We should keep the same string value if possible (else migrate existing records).

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BaseCloudServiceRuntimeMonitor.scala
```diff
@@ -0,0 +1,367 @@
...
+    } yield ()
+
+  def pollCheck(googleProject: GoogleProject,
+                runtimeAndRuntimeConfig: RuntimeAndRuntimeConfig,
+                operation: com.google.cloud.compute.v1.Operation,
+                action: RuntimeStatus)(implicit ev: ApplicativeAsk[F, TraceId]): F[Unit]
+
+  private def handler(monitorContext: MonitorContext, monitorState: MonitorState): F[CheckResult] =
+    for {
+      now <- nowInstant
```
rtitle, line 59: Just curious if we need `MonitorContext` _and_ `ApplicationContext`.. could just `MonitorContext` suffice?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BaseCloudServiceRuntimeMonitor.scala
```diff
@@ -0,0 +1,367 @@
...
+  def readyRuntime(runtimeAndRuntimeConfig: RuntimeAndRuntimeConfig,
+                   publicIp: IP,
+                   monitorContext: MonitorContext,
+                   dataprocInstances: Set[DataprocInstance]): F[CheckResult] =
+    for {
+      now <- nowInstant
+      _ <- persistInstances(runtimeAndRuntimeConfig, dataprocInstances)
+      // update DB after auth futures finish
+      _ <- clusterQuery.setToRunning(runtimeAndRuntimeConfig.runtime.id, publicIp, now).transaction
+      // Record metrics in NewRelic
```
rtitle: Maybe fix comment to not say NewRelic :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/InstanceComponent.scala
Qi77Qi: it should be the same...but I'll double check

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BaseCloudServiceRuntimeMonitor.scala
Qi77Qi: so `MonitorContext`'s now is fixed to when the monitoring starts, but `AppContext`'s now is more like when the function is being called, which might change throughout the monitoring process

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/InstanceComponent.scala
Qi77Qi:
```
scala> Master.toString
res1: String = Master

scala> Worker.toString
res2: String = Worker

scala> SecondaryWorker.toString
res3: String = SecondaryWorker
```

It matches what's in database currently

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/runtimeModels.scala
```diff
@@ -78,6 +78,7 @@ object RuntimeStatus extends Enum[RuntimeStatus] {
 
   // NOTE: Remember to update the definition of this enum in Swagger when you add new ones
   // transition statuses
+  case object PreCreating extends RuntimeStatus //This status exists when we saves a runtime after createRuntime request, but we haven't published the even to back leo
```
rtitle: Maybe `Provisioning`? 🤔 

I like this idea of this status but we should double-check with API consumers (Terra UI, AoU). Our status codes are a contract that we provide and we should make sure adding a new one doesn't break anything.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/LeoPublisher.scala
```diff
@@ -0,0 +1,56 @@
...
+import fs2.{Pipe, Stream}
+import io.chrisdavenport.log4cats.Logger
+import io.circe.syntax._
+import org.broadinstitute.dsde.workbench.google2.GooglePublisher
+import org.broadinstitute.dsde.workbench.leonardo.db.{clusterQuery, DbReference}
+import org.broadinstitute.dsde.workbench.leonardo.http.{dbioToIO, nowInstant}
+import org.broadinstitute.dsde.workbench.leonardo.monitor.LeoPubsubCodec._
+import org.broadinstitute.dsde.workbench.leonardo.monitor.LeoPubsubMessage
+
+final class LeoPublisher[F[_]: Logger: Timer](
```
rtitle, line 24: Can you add a comment about what this class is for? I think it's used in the front-leo `service` layer to publish messages to PubSub, right?

Can we put it in `service` package instead of top-level `leonardo` package?

Qi77Qi: it basically provides a publishing stream (`process`)...It's not being used in anywhere else other than `boot`...

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BaseCloudServiceRuntimeMonitor.scala
```diff
@@ -0,0 +1,367 @@
...
+                    instances: Set[DataprocInstance])(
+    implicit ev: ApplicativeAsk[F, AppContext]
+  ): F[CheckResult]
+
+  def readyRuntime(runtimeAndRuntimeConfig: RuntimeAndRuntimeConfig,
+                   publicIp: IP,
+                   monitorContext: MonitorContext,
+                   dataprocInstances: Set[DataprocInstance]): F[CheckResult] =
+    for {
+      now <- nowInstant
```
rtitle, line 98: Could this method get `now` from `ApplicationContext`?

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/runtimeModels.scala
Qi77Qi: I didn't use `Provisioning` just because gce is already using the word.

For contract, I think client doesn't need to know this status, we can map it to `Creating` in contract

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BaseCloudServiceRuntimeMonitor.scala
```diff
@@ -0,0 +1,367 @@
...
+      _ <- if (runtimeAndRuntimeConfig.runtime.status == RuntimeStatus.Creating)
+        RuntimeMonitor.recordClusterCreationMetrics(
+          runtimeAndRuntimeConfig.runtime.auditInfo.createdDate,
+          runtimeAndRuntimeConfig.runtime.runtimeImages,
+          monitorConfig.imageConfig,
+          runtimeAndRuntimeConfig.runtimeConfig.cloudService
+        )
+      else F.unit
+      timeElapsed = (now.toEpochMilli - monitorContext.start.toEpochMilli).milliseconds
+      // Finally pipe a shutdown message to this actor
```
rtitle: Remove comment

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BaseCloudServiceRuntimeMonitor.scala
```diff
@@ -0,0 +1,367 @@
...
+      _ <- logger.info(
+        s"${monitorContext} | Runtime ${runtimeAndRuntimeConfig.runtime.projectNameString} is ready for use after ${timeElapsed.toSeconds} seconds!"
+      )
+    } yield ((), None)
+
+  def checkAgain(
```
rtitle, line 122: I'm a bit confused as to the purpose of the `checkAgain` method -- can you elaborate?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/DataprocRuntimeMonitor.scala
```diff
@@ -0,0 +1,524 @@
...
+      result <- runtimeAndRuntimeConfig.runtime.status match {
+        case RuntimeStatus.Creating =>
+          creatingRuntime(cluster, monitorContext, runtimeAndRuntimeConfig)
+        // This is needed because during boot time, we no longer have reference to the previous delete operation
+        // But this path should only happen during boot and there's on-going delete
+        case RuntimeStatus.Deleting =>
+          deletedRuntime(cluster, monitorContext, runtimeAndRuntimeConfig)
+        case RuntimeStatus.Starting =>
+          startingRuntime(cluster, monitorContext, runtimeAndRuntimeConfig)
+        // This is needed because during boot time, we no longer have reference to the previous delete operation
```
rtitle: Fix comment

rtitle:
I've been going through this but there's a lot of changes and it's a little intimidating. I want to make sure I fully understand the monitoring flows before thumbing. What do you think of doing a walkthrough for this PR, to describe the monitor class organization and state transition flows? A design doc would probably be good too.

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/runtimeModels.scala
rtitle: 👍 to map it to `Creating` in contract

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/LeoPublisher.scala
rtitle: I see.. it's not called directly by the `service` classes. The current location next to `Boot` is fine I guess.

Qi77Qi:
@rtitle sure yea..I've been meaning to have a walkthrough...I'll think about design doc, not sure exactly what to put there yet, but maybe after walkthrough I'll have a better understanding what needs explanation.

rtitle:
Sounds good re walkthrough.. maybe we can do it tomorrow (5/12)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/LeoPublisher.scala
Qi77Qi: yea, so this is just for dequeueing from publisherQueue and transform it into a native google pubsub message; After pubsub message is published, we update database when necessary..I added this to comment...

And services that need to publish event, they do so by enqueue a message into the internal publisherQueue (a fs2.Queue)

http/src/main/resources/reference.conf
Qi77Qi: removing the timeout for now just to introduce less changes in this PR

Qi77Qi:
@rtitle  it's one of the items carried over from last week...so it's already on the agenda...we can make sure it happens this time :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BaseCloudServiceRuntimeMonitor.scala
Qi77Qi: it's used anywhere the transition is not complete yet...Before setting the state to either `Check` or `CheckTools` which will cause the stream the sleep some interval and continue, it also checks if we should timeout the monitoring process

Qi77Qi:
fiab-start failure ...jenkins retest

Qi77Qi:
jenkins retest

Qi77Qi:
fiab-start failure..jenkins retest

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/cluster/ClusterStatusTransitionsSpec.scala
```diff
@@ -48,6 +48,7 @@ class ClusterStatusTransitionsSpec extends GPAllocFixtureSpec with ParallelTestE
       // delete the cluster, but don't wait
       deleteCluster(billingProject, clusterName, monitor = false)
 
+      Thread.sleep(1000)
```
rtitle: Is this needed? I think the intent is that the cluster should be `Deleting` immediately after the `delete` call.

Qi77Qi: since the DB update is async to the API call now (after event published when publisherQueue is dequeued), DB might not have been updated yet right after the API call finishes

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/LeoPublisher.scala
```diff
@@ -0,0 +1,60 @@
...
+          .newBuilder()
+          .setData(byteString)
+          .putAttributes("traceId", msg.traceId.map(_.asString).getOrElse("null"))
+          .build()
+      }
+
+  private def updateDatabase(msg: LeoPubsubMessage): F[Unit] =
+    for {
+      now <- nowInstant[F]
+      _ <- msg match {
```
rtitle, line 59: Hm does this need to handle `LeoPubSubMessage.UpdateRuntimeMessage`? We do set the DB status to `Updating` in some cases I think.

Qi77Qi: I don't see db update where `UpdateRuntimeMessage` is being used..maybe I'm missing something?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/LeoRoutesSprayJsonCodec.scala
```diff
@@ -145,6 +145,14 @@ object LeoRoutesSprayJsonCodec extends DefaultJsonProtocol {
     }
   }
 
+  implicit val runtimeStatusFormat: RootJsonWriter[RuntimeStatus] = (status: RuntimeStatus) => {
```
rtitle: Could we put this logic in circe `JsonCodec`? Have been trying to not add new spray encoders -- this is just used for the legacy routes anyway, the new `/runtimes` routes use all circe encoding/decoding.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/RuntimeRoutes.scala
```diff
@@ -326,6 +326,13 @@ object RuntimeRoutes {
     } yield UpdateRuntimeRequest(rc, as.getOrElse(false), ap, at.map(_.minutes))
   }
 
+  implicit val runtimeStatusEncoder: Encoder[RuntimeStatus] = Encoder.encodeString.contramap { x =>
```
rtitle, line 328: Oh here it is. Could this live in `core`?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/LeoRoutesSprayJsonCodec.scala
Qi77Qi: this is just to support the old route...new route uses circle (https://github.com/DataBiosphere/leonardo/pull/1395/files#diff-9bef6acafa788ef59ee57f5d1618439aR329)...I didn't put it in `JsonCodec` in core is because this encoding is contract specific...most likely anywhere else we don't want to do this encoding (for instance, in Leo pubsub, it's encoded differently)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/Config.scala
```diff
@@ -47,7 +48,7 @@ import scala.concurrent.duration._
 object Config {
   val config = ConfigFactory.parseResources("leonardo.conf").withFallback(ConfigFactory.load()).resolve()
 
-  implicit val applicationConfigReader: ValueReader[ApplicationConfig] = ValueReader.relative { config =>
+  implicit private val applicationConfigReader: ValueReader[ApplicationConfig] = ValueReader.relative { config =>
```
rtitle, line 51: +1 on making all these `private`

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/cluster/ClusterStatusTransitionsSpec.scala
rtitle: Hm - so this is a change from our original behavior. Previously callers were guaranteed that the status is updated transactionally as part of calling a `POST`/`PATCH`/`DELETE` API. Now the status change may happen asynchronously some time _after_ the API returns.

Am wondering about possible effects here. For example when the UI pauses/deletes a runtime, it calls `POST` then `GET` immediately to get the new status. If there's now a race condition it may lead to delays in updating the status in the UI.

I understand why we did this: to recover better from pubsub publishing errors. But I'm wondering if there's another way to solve this while still making the status change transactional within the request. (Maybe we could error out the runtime instead if publishing fails or takes too long)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/LeoPublisher.scala
rtitle: Oh - `Updating` DB status is set in the `LeoPubSubMessageSubscriber`, because we only need to do it for certain types of updates.  This seems fine then.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/LeoRoutesSprayJsonCodec.scala
rtitle: Yeah I see now.. thanks

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/RuntimeRoutes.scala
rtitle: Answered above

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/google/HttpGoogleDataprocDAO.scala
```diff
@@ -132,19 +131,6 @@ class HttpGoogleDataprocDAO(
     transformed.value.handleGoogleException(googleProject, clusterName)
   }
 
-  override def getClusterInstances(googleProject: GoogleProject,
```
rtitle, line 135: Assuming this moved to wb-libs

Qi77Qi: yea https://github.com/broadinstitute/workbench-libs/pull/300/files#diff-28a9648a50715239edaa8ca1bfc1a263R130

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/cluster/ClusterStatusTransitionsSpec.scala
jdcanas: This ties into the need for the `CLUSTER_PATCH` table, since a patch that requires a status transition doesn't happen transactionally with the request. 

I don't think we've implemented yet UI-side, but the goal is the UI will poll frequently when it GETs and sees the `isUpdating` flag.

We will definitely start getting complaints from the UI team and users if other endpoints start behaving inconsistently. It can be particularly disruptive, since the UI polls very infrequently when in certain statuses (such as 'Running') since it relies on this transaction assumption and immediately refreshes after all requests, so it can manifest in users not seeing changed statuses or receiving any feedback on transitions for minutes

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/LeoRoutesSprayJsonCodec.scala
Qi77Qi: I'd prefer not to make frontend do publishing to google synchronous to API calls everywhere. If we're worried about the delay in UI, I think a better way is to introduce more `pre` statuses persisted synchronous to API call

rtitle: does this belong in the other thread? :)

I just think the transactional nature of these APIs is a contract that we shouldn't break. `pre` statuses could be a solution. Or maybe a flag `isPublished` in the DB which is initially `false` but flipped to `true` once the google msg has been published.

Qi77Qi: oops yea...got threads mixed up.

I kinda prefer `pre`, cuz we'll still need `PreCreating` even if we introduce `isPublished`...So I feel introducing `Pre` for all transition requests seems more consistent.

Qi77Qi:
build failed...jenkins retest

Qi77Qi:
another resolution error somehow

Qi77Qi:
another resolution error somehow...jenkins retest

Qi77Qi:
build failure...jenkins retest....

Qi77Qi:
another resolution error. jenkins retest

Qi77Qi:
what's wrong with jenkins...jenkins retest

Qi77Qi:
fiab start failed this time...jenkins retest

Qi77Qi:
a different test failed...jenkins retest...

Qi77Qi:
in https://fc-jenkins.dsp-techops.broadinstitute.org/view/Leonardo/job/LeoPerfTestRunner/302/, at peak, latency for asynTaskProcessor can be up to almost 1 min (seems this only happens when int the beginning of the test where we're creating a ton runtimes)...once traffic slows down, latency becomes close to 0 seconds.

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/AsyncTaskProcessor.scala
```diff
@@ -0,0 +1,84 @@
...
+
+final class AsyncTaskProcessor[F[_]: Timer](config: AsyncTaskProcessor.Config, asyncTasks: InspectableQueue[F, Task[F]])(
+  implicit logger: Logger[F],
+  F: Concurrent[F],
+  metrics: OpenTelemetryMetrics[F]
+) {
+  def process: Stream[F, Unit] = {
+    val asyncTaskStream = asyncTasks
+      .dequeueChunk(config.maxConcurrentTasks)
+      .parEvalMapUnordered(config.maxConcurrentTasks)(task => handler(task))
```
rtitle, line 25: Curious about the bug you mentioned in standup.. is the fix essentially calling `handler` in `parEvalMapUnordered`? Maybe we can talk details @ walkthrough :)

Qi77Qi: yea, this is the bug...It's definitely interesting to show...I added it on walkthrough agenda...I plan to run the unit tests and it'll become clear what happened

rtitle (approved):
Thumbing this.. code looks good to me, and I'm glad that auto & perf test results look good.

Once on dev I'd love to do some manual testing of status transitions in Terra UI, and autopause.

Weirdly travis didn't seem to run on this PR 🤔

Qi77Qi:
think this PR is ready now...https://fc-jenkins.dsp-techops.broadinstitute.org/view/Leonardo/job/LeoPerfTestRunner/338/ looks good, and it's actually much better than dev run, dev run took 30 min, this one only took half the time...

The bug previously causing perf issue is due to https://github.com/DataBiosphere/leonardo/pull/1395/files#diff-44372e68c769f56432d025c8a7d171cfR204, (previously we have a very low queue bound, hence this call will block when we get too many proxy requests), so to fix this, I increased queue bound and also increased the frequency of updating `dateAccessed` field. (I'll talk more about this during walkthrough next week)

@rtitle let me know if you any concerns with merging. The change worth mentioning after your approval last week is bumping Akka version (I did this when I thought there might be a bug causing the proxy exception I shared in slack). I checked newrelic and it gets metrics from perf just fine.

rtitle:
@Qi77Qi I did another pass and this looks good to me. Nice work! I'm impressed you were able to get rid of the `akka-actor` dependency. Wanna do the same for Cromwell? 😅 

I think you can merge this, it will be good to have it bake in dev for a while. Let team members know they'll probably have to rebase.

rtitle (approved):
:shipit:

Qi77Qi:
@rtitle I think akka-actor might still be used under the hood by akka-http somewhere. But at least we don't directly use it anymore :)....thanks for the review!

PR #1399, opened 2020-05-01
IA-1888 Terra's content security policy and use of JavaScript libraries

rtitle:
@adrazhi looking for your input here (not sure you're gonna like this one..)

cc @sharmatime and @deflaux as well.

We've had multiple user complaints of JS libraries not loading in Terra notebooks due to CSP. Specifically:

1. [plotly](https://plotly.com/r/) and [facets](https://github.com/PAIR-code/facets)
   * For some reason unbeknownst to me these libraries load scripts from `data:` URIs. Our current CSP has `data:text/javascript` but I discovered you can't specify the content type after the scheme. Setting just `data:` fixes both of these libraries.
   * CSP evaluator doesn't like this, it says:
      ```
      data: URI in script-src allows the execution of unsafe scripts.
      ```

2. [altair](https://altair-viz.github.io/) 
   * This library needs `https://cdn.jsdelivr.net` added to `script-src`. CSP evaluator doesn't like this either, it says:
      ```
      cdn.jsdelivr.net is known to host Angular libraries which allow to bypass this CSP.
      ```

3. Separately, I tried to remove `'unsafe-inline'` from `script-src` but this prevents notebooks from loading at all. Apparently Jupyter needs this in order to work at all:
https://jupyter-notebook.readthedocs.io/en/stable/public_server.html#content-security-policy-csp

So I worry I'm adding even more insecure things to CSP. I'm open to any alternatives. If these changes are acceptable though I'd like to push them through to unblock some users trying to use Terra notebooks.

---

In all cases:

### DIFF
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -278,6 +278,9 @@ contentSecurityPolicy {
   objectSrc = [
     "'none'"
   ]
+  reportUri = [
+    "https://terra.report-uri.com/r/d/csp/reportOnly"
+  ]
 }
 
 swagger {
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -256,7 +256,7 @@ contentSecurityPolicy {
     "'unsafe-inline'",
     "'unsafe-eval'",
     "https://apis.google.com",
-    "https://cdn.jsdelivr.net/npm"
+    "https://cdn.jsdelivr.net"
   ]
   styleSrc = [
     "'self'",
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -255,7 +255,8 @@ contentSecurityPolicy {
     "data:",
     "'unsafe-inline'",
     "'unsafe-eval'",
-    "https://apis.google.com"
+    "https://apis.google.com",
+    "https://cdn.jsdelivr.net/npm"
   ]
   styleSrc = [
     "'self'",
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -248,11 +248,13 @@ contentSecurityPolicy {
   frameAncestors = [
     "'none'"
   ]
-  # TODO: try to remove unsafe-inline
-  # see https://broadworkbench.atlassian.net/browse/IA-1763
   scriptSrc = [
     "'self'",
+    # Note: data: is insecure but needed to support tools like plotly and facets.
+    # See discussion in https://github.com/DataBiosphere/leonardo/pull/1399
     "data:",
+    # Note: 'unsafe-inline' in insecure but is needed by Jupyter UI.
+    # See https://broadworkbench.atlassian.net/browse/IA-1763
     "'unsafe-inline'",
     "'unsafe-eval'",
     "https://apis.google.com",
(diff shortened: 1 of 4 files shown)


rtitle:
Example generated CSP with these changes:
```
Content-Security-Policy: frame-ancestors 'self' http://localhost:3000 http://localhost:4200 https://localhost:443 https://bvdp-saturn-dev.appspot.com https://all-of-us-workbench-test.appspot.com dev.terra.biodatacatalyst.nhlbi.nih.gov; script-src 'self' data: 'unsafe-inline' 'unsafe-eval' https://apis.google.com https://cdn.jsdelivr.net; style-src 'self' 'unsafe-inline' data:; connect-src 'self' wss://*.broadinstitute.org:* wss://notebooks.firecloud.org:* *.googleapis.com https://*.npmjs.org https://data.broadinstitute.org https://s3.amazonaws.com/igv.broadinstitute.org/ https://s3.amazonaws.com/igv.org.genomes/ https://raw.githubusercontent.com/PAIR-code/facets/1.0.0/facets-dist/facets-jupyter.html https://cdnjs.cloudflare.com/ajax/libs/webcomponentsjs/1.3.3/webcomponents-lite.js; object-src 'none'
```

(note this is a dev version; prod doesn't have dev URLs in `frame-ancestors`.)

rtitle:
And just to show examples of things working:

plotly: 
image

facets:
image

altair:
image

adrazhi:
@rtitle This is probably not a surprise suggestion since CSP evaluator complains about it too but I was wondering, can we use  `strict-dynamic' in combination with CSP nonces or hashes so we don't rely much on whitelists?

rtitle:
Yeah.. will look into that and update back here.

rtitle:
@adrazhi I looked into `strict-dynamic` and unless I'm missing something it doesn't really fit this use case. Users want to load arbitrary JS (like `plotly`) into Jupyter notebooks. To use `strict-dynamic` with hashes we'd have to whitelist all the hashes of scripts users might want to use. This seems unmaintainable since these things probably change often upstream. And using nonces doesn't seem possible because nonces aren't specified when these scripts are loaded into Jupyter components.

As an example, here is the DOM of a notebook with a `plotly` chart loaded:

image

Looks like it renders an iframe with a bunch of `<script src=data:...>` tags. I poked around a bit but couldn't see a great way to make plotly NOT load scripts from `data:` URIs (I'm guessing they do that for efficiency reasons).

So I'm not sure how to allow this except by putting `script-src data:` in our CSP. Our users really need access to these tools. Can you think of another way to secure this, or is what I have in this PR acceptable?

Thanks.

adrazhi:
@rtitle  Yeah I think for now it should be fine to merge this. 
Also could you please add this to the `report-uri` directive of CSP: 

- `report-uri: https://terra.report-uri.com/r/d/csp/reportOnly`

The endpoint will help us with data on potential CSP "violations" and better observe what users are doing in case a violation occurs.

rtitle:
Will do, thank you!

Qi77Qi (approved):

rtitle:
Retested latest changes: confirmed all visualizations still work and no CSP issues due to added `report-uri` field. Merging this.

rtitle:
@adrazhi down the line I'd love to see data about things flagged by `report-uri`. Might give us good insight into what tools our users are _trying_ to use. For example I just discovered that `https://cdn.pydata.org/bokeh/release/bokeh-tables-1.2.0.min.js` is also blocked by CSP, which is a tool Hail users might try to load.

adrazhi:
@rtitle Yeah I'll keep you posted.

PR #1402, opened 2020-05-01
[IA-1813] Inactive zombie monitoring

akarukappadath:

---

In all cases:


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ZombieMonitorQueries.scala
```diff
@@ -3,41 +3,53 @@ package db
...
+  }
 }
 
-final case class PotentialZombieRuntime(id: Long,
-                                        googleProject: GoogleProject,
-                                        runtimeName: RuntimeName,
-                                        status: RuntimeStatus,
-                                        auditInfo: AuditInfo,
-                                        cloudService: CloudService)
+final case class ZombieCandidate(runtime: Runtime,
```
akarukappadath: In order for the runtime to be deleted through a CloudService interpreter, we need to pass in the full runtime, which is why `ZombieCandidate` now has a full Runtime in it. The other option was to have two types of `ZombieCandidate`s - `ActiveZombieCandidate` and `InactiveZombieCandidate` but I thought it made the code in the monitor a lot less easily readable

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ZombieRuntimeMonitor.scala
```diff
@@ -48,37 +54,69 @@ class ZombieRuntimeMonitor[F[_]: Parallel: ContextShift: Timer](
...
+          semaphore.withPermit(
+            // If the cluster is less than one minute old, it may not exist in Google because it is still Provisioning, so it's not a zombie
+            isRuntimeActiveInGoogle(runtime.googleProject, runtime.runtimeName, candidate.cloudService, startInstant)
+              .ifA[Option[ZombieCandidate]](zombieIfOlderThanOneMinute(candidate, startInstant), updateRuntimeAsConfirmedDeleted(runtime.id))
+          )
+      }
+
+      // Delete runtime on Leo side if active zombie, or on Google side if inactive zombie
+      _ <- activeZombies.parTraverse(zombie => semaphore.withPermit(handleActiveZombieRuntime(zombie, startInstant)))
+      _ <- inactiveZombies.parTraverse(zombie => semaphore.withPermit(handleInactiveZombieRuntime(zombie, startInstant)))
```
Qi77Qi: would be good to add some metrics..like number of activeZombies (I think we have this already) and number of inactiveZombies

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ZombieRuntimeMonitor.scala
```diff
@@ -48,37 +54,69 @@ class ZombieRuntimeMonitor[F[_]: Parallel: ContextShift: Timer](
...
+      _ <- logger.info(
+        s"Starting inactive zombie detection within ${unconfirmedDeletedRuntimes.size} runtimes with concurrency of ${config.concurrency}"
+      )
+
+      inactiveZombies <- unconfirmedDeletedRuntimes.traverseFilter {
+        candidate =>
+          val runtime = candidate.runtime
+          semaphore.withPermit(
+            // If the cluster is less than one minute old, it may not exist in Google because it is still Provisioning, so it's not a zombie
+            isRuntimeActiveInGoogle(runtime.googleProject, runtime.runtimeName, candidate.cloudService, startInstant)
```
Qi77Qi: we probably don't even need to check status in this case...I think as long as the runtime exists in Google, we should delete it whatever state that runtime is in 🤔

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ZombieMonitorQueries.scala
Qi77Qi: Delete really shouldn't need the whole `Runtime`...if you look at both gce and dataproc interpreters, they only really need googleProject, runtimeName, and zoneName (if gce)...Let's refactor `deleteRuntime` signature a bit? :)

Qi77Qi (commented):
Looks great 👍 ...Left some minor comments

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ZombieRuntimeMonitor.scala
akarukappadath: Yup, they're both in the `handle` function for each

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ZombieMonitorQueries.scala
```diff
@@ -3,41 +3,53 @@ package db
...
+          r._1.googleProject -> Chain(
+            ZombieCandidate(runtime, cloudService)
           )
         )
       }
     }
   }
 
+  private def unconfirmedRuntimeLabels: Query[LabelTable, LabelRecord, Seq] =
+    labelQuery.filter(_.key === Config.zombieClusterMonitorConfig.deletionConfirmationLabelKey).filter(_.value === "false")
```
rtitle: Existing records won't have this label key at all.. does this code take that into account?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ZombieMonitorQueries.scala
```diff
@@ -3,41 +3,53 @@ package db
...
+
+  def listInactiveZombieQuery(implicit ec: ExecutionContext): DBIO[List[ZombieCandidate]] = {
+    val runtimes = clusterQuery.filter(_.status === RuntimeStatus.Deleted.toString)
+    val unconfirmedRuntimeQuery = runtimes.join(unconfirmedRuntimeLabels).on(_.id === _.clusterId)
+    val joinedQuery = unconfirmedRuntimeQuery.joinLeft(RuntimeConfigQueries.runtimeConfigs).on(_._1.runtimeConfigId === _.id)
```
rtitle: Maybe do `join` on runtime configs because we expect every record to have one

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ZombieRuntimeMonitor.scala
```diff
@@ -48,37 +54,69 @@ class ZombieRuntimeMonitor[F[_]: Parallel: ContextShift: Timer](
...
-      _ <- zombies.parTraverse(zombie => semaphore.withPermit(handleZombieRuntime(zombie, startInstant)))
+
+      // Get all deleted cluster that haven't been confirmed from the Leo DB
+      unconfirmedDeletedRuntimes <- ZombieMonitorQueries.listInactiveZombieQuery.transaction
+
+      _ <- logger.info(
+        s"Starting inactive zombie detection within ${unconfirmedDeletedRuntimes.size} runtimes with concurrency of ${config.concurrency}"
+      )
+
+      inactiveZombies <- unconfirmedDeletedRuntimes.traverseFilter {
```
rtitle: should maybe be `parTraverseFilter` since we're using a `semaphore`?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ZombieRuntimeMonitor.scala
```diff
@@ -48,37 +54,69 @@ class ZombieRuntimeMonitor[F[_]: Parallel: ContextShift: Timer](
...
+          s"Elapsed time = ${duration} milli seconds."
       )
     } yield ()
 
+  // This puts a label on a deleted runtime to show that we've confirmed the google runtime's deletion
+  private def updateRuntimeAsConfirmedDeleted(runtimeId: Long): F[Option[ZombieCandidate]] = {
+    dbRef.inTransaction {
+      labelQuery.save(runtimeId, config.deletionConfirmationLabelKey, "true")
+    }.void
+    F.pure(None)
```
rtitle: Can we make this method return `F[Unit]` and remove this line?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -656,6 +657,8 @@ class LeonardoService(
         now <- nowInstant
         _ <- if (hasDataprocInfo) clusterQuery.markPendingDeletion(cluster.id, now).transaction
         else clusterQuery.completeDeletion(cluster.id, now).transaction
+        // save label for zombie monitor actor to confirm deletion of google runtime in the future
+        _ <- labelQuery.save(cluster.id, zombieRuntimeMonitorConfig.deletionConfirmationLabelKey, "false").transaction
```
rtitle: Please make the same change in `RuntimeServiceInterp.scala`

rtitle:
Few comments above.. overall the approach looks good, will re-review before merge.

One question: are we expecting this to loop through _all_ currently deleted clusters the first time this runs? There are a lot of them in prod:
```
MySQL [leonardo]> select count(*) from CLUSTER where status = 'Deleted';
+----------+
| count(*) |
+----------+
|    14690 |
+----------+
```
I'm thinking it should be OK since we control concurrency with a Semaphore but we might just want to double check how long it'll take, and make sure we don't blow up any Google quota or anything.

Another approach is just to manually set `deletionConfirmed==true` on all existing deleted clusters via a liquibase migration. Then the new inactive zombie monitoring will only take place for runtimes going forward. We can process all legacy runtimes offline with a script.

Qi77Qi:
@rtitle I was thinking semaphore will protect us, but I think what you're suggesting with migration is probably even better.

akarukappadath:
@rtitle I'm not sure we'd be able to do a liquibase migration for existing runtimes since it's a label (db row) rather than a db column. If we're worried about a google quota, we could write a script to take care of the existing runtimes

rtitle:
@akarukappadath we can run arbitrary SQL in liquibase migrations, example: https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20200317_fix_custom_image_type.xml#L5

Something like
```
INSERT INTO LABEL (resourceId, resourceType, key, value) SELECT id, 'runtime', 'deletionConfirmed', 'true` FROM CLUSTER c where c.status = 'Deleted'
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ZombieRuntimeMonitor.scala
akarukappadath: I'm worried this might lead to a weird state - what if a runtime is `Deleting` on Google? I think trying to delete a `Deleting` runtime would cause an error right?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ZombieRuntimeMonitor.scala
```diff
@@ -48,37 +54,70 @@ class ZombieRuntimeMonitor[F[_]: Parallel: ContextShift: Timer](
...
+      unconfirmedDeletedRuntimes <- ZombieMonitorQueries.listInactiveZombieQuery.transaction
+
+      _ <- logger.info(
+        s"Starting inactive zombie detection within ${unconfirmedDeletedRuntimes.size} runtimes with concurrency of ${config.concurrency}"
+      )
+
+      inactiveZombies <- unconfirmedDeletedRuntimes.parTraverse {
+        candidate =>
+          semaphore.withPermit {
+            isRuntimeActiveInGoogle(candidate.googleProject, candidate.runtimeName, candidate.cloudService, startInstant).flatMap { isActive =>
```
akarukappadath: Is there a more scala-esque way to do this? `ifA` didn't work here since it looks like it evaluates both parameters and returns the relevant one, which meant that the database was being updated regardless of the boolean value

Qi77Qi: I don't think so...you're probably have to use `flatMap` and then if else

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20200506_label_pk.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="leonardo" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+    <changeSet logicalFilePath="leonardo" author="azmina" id="label_pk">
+        <addPrimaryKey tableName="LABEL" columnNames="resourceId, resourceType, key"/>
```
rtitle: I'm fine with this change but I'm curious if there is a specific reason for using primary key over unique constraint? Kinda wish I did it in my last PR

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
```diff
@@ -690,13 +690,13 @@ object clusterQuery extends TableQuery(new ClusterTable(_)) {
...
 
   private def unmarshalCluster(clusterRecord: ClusterRecord,
-                               instanceRecords: Seq[InstanceRecord],
-                               errors: List[ClusterErrorRecord],
-                               labels: LabelMap,
-                               userJupyterExtensionConfig: List[ExtensionRecord],
-                               clusterImageRecords: List[ClusterImageRecord],
-                               scopes: List[ScopeRecord],
-                               patch: List[PatchRecord]): Runtime = {
+                                         instanceRecords: Seq[InstanceRecord],
```
rtitle: Can you run `scalafmt` and `test:scalafmt`?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/LabelComponent.scala
```diff
@@ -16,19 +16,18 @@ class LabelTable(tag: Tag) extends Table[LabelRecord](tag, "LABEL") {
   def key = column[String]("key", O.Length(254))
   def value = column[String]("value", O.Length(254))
 
-  def uniqueKey = index("IDX_LABEL_UNIQUE", (resourceId, resourceType, key), unique = true)
+  def pk = primaryKey("label_pk", (resourceId, resourceType, key))
```
rtitle: Maybe specify the constraintName in liquibase so the name here matches. Seems you can do:
```
<addPrimaryKey tableName="LABEL" columnNames="resourceId, resourceType, key" contstraintName="PK_LABEL" />
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/LabelComponent.scala
```diff
@@ -38,6 +37,7 @@ object labelQuery extends TableQuery(new LabelTable(_)) {
     }
 
   def deleteAllForResource(resourceId: Long, resourceType: LabelResourceType): DBIO[Int] =
+
```
rtitle: Remove blank line (`scalafmt` might do this for you)

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20200506_label_pk.xml
akarukappadath: If we want to use `insertOrUpdate` on the label table (in order to change existing labels) there needs to be a primary key for the table. I slacked about it earlier this week, sorry if I wasn't clear!

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ZombieRuntimeMonitor.scala
```diff
@@ -48,37 +54,70 @@ class ZombieRuntimeMonitor[F[_]: Parallel: ContextShift: Timer](
...
             isProjectActiveInGoogle(project).flatMap {
               case true =>
-                // If the project is active, check each individual cluster
-                clusters.toList.traverseFilter { cluster =>
-                  isRuntimeActiveInGoogle(cluster, startInstant).ifA(F.pure(None), F.pure(Some(cluster)))
+                // If the project is active, check each individual runtime
+                zombieCandidates.toList.traverseFilter { candidate =>
+                  // If the runtime is less than one minute old, it may not exist in Google because it is still Provisioning, so it's not a zombie
+                  isRuntimeActiveInGoogle(candidate.googleProject, candidate.runtimeName, candidate.cloudService, startInstant)
+                    .ifA[Option[ZombieCandidate]](F.pure(None), F.pure(zombieIfOlderThanOneMinute(candidate, startInstant)))
```
rtitle: How did we decide on 1 minute? Seems a little aggressive -- I could imaging a 1 minute lag between when front Leo receives the create request and back Leo actually creates the runtime. Maybe 10 minutes would be safer.

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20200506_label_pk.xml
rtitle: Yeah I think I realized that, makes sense 👍

rtitle:
Sorry a few more comments, but aside from those this looks great -- all the logic in `ZombieRuntimeMonitor` looks correct to me. Glad we're adding metrics too.

Will thumb once the above is addressed.

rtitle (approved):
Looks good but as discussed let's hold off merging until after the release tomorrow. Thank you!

PR #1405, opened 2020-05-04
[IA-1872] Kubernetes database code (slick/liquibase)

jdcanas:
Looking for some feedback
This PR:

- adds a Cluster, Nodepool, and Namespace table
- takes into account the recent audit info and label generalization that has been done
- provides unit tests for the liquibase and slick code above
- renames things as per kaan's wb-libs PR

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
@@ -1,8 +1,8 @@
 package org.broadinstitute.dsde.workbench.leonardo
 
 import enumeratum.{Enum, EnumEntry}
-import org.broadinstitute.dsde.workbench.google2.GKEModels.{KubernetesClusterId, KubernetesClusterName, NodePoolName}
-import org.broadinstitute.dsde.workbench.google2.KubernetesModels.KubernetesMasterIP
+import org.broadinstitute.dsde.workbench.google2.GKEModels.{KubernetesClusterId, KubernetesClusterName, NodepoolName}
+import org.broadinstitute.dsde.workbench.google2.KubernetesModels.KubernetesApiServerIp
 import org.broadinstitute.dsde.workbench.google2.KubernetesSerializableName.KubernetesNamespaceName
 import org.broadinstitute.dsde.workbench.google2.{Location, MachineTypeName, NetworkName, SubnetworkName}
 import org.broadinstitute.dsde.workbench.model.WorkbenchEmail
@@ -44,7 +44,7 @@ case class SaveKubernetesCluster(googleProject: GoogleProject,
 
 case class KubernetesClusterSamResource(resourceId: String)
 
-case class KubernetesClusterAsyncFields(apiServerIp: Option[KubernetesMasterIP],
+case class KubernetesClusterAsyncFields(apiServerIp: Option[KubernetesApiServerIp],
                                         networkInfo: Option[NetworkFields]
                                        )
 
@@ -88,7 +88,7 @@ case class KubernetesNamespaceId(id: Long)
 
 case class Nodepool(id: NodepoolLeoId,
                     clusterId: KubernetesClusterLeoId,
-                    nodepoolName: NodePoolName,
+                    nodepoolName: NodepoolName,
                     status: NodepoolStatus,
                     auditInfo: AuditInfo,
                     machineType: MachineTypeName,
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
@@ -1,0 +1,123 @@
+package org.broadinstitute.dsde.workbench.leonardo
+
+import java.time.Instant
+
+import enumeratum.{Enum, EnumEntry}
+import org.broadinstitute.dsde.workbench.google2.GKEModels.{KubernetesClusterId, KubernetesClusterName, NodePoolName}
+import org.broadinstitute.dsde.workbench.google2.KubernetesModels.KubernetesMasterIP
+import org.broadinstitute.dsde.workbench.google2.KubernetesSerializableName.KubernetesNamespaceName
+import org.broadinstitute.dsde.workbench.google2.{Location, MachineTypeName, NetworkName, SubnetworkName}
+import org.broadinstitute.dsde.workbench.model.WorkbenchEmail
+import org.broadinstitute.dsde.workbench.model.google.GoogleProject
+
+/** Google Container Cluster statuses
+ *  see: https://cloud.google.com/kubernetes-engine/docs/reference/rest/v1/projects.locations.clusters#Cluster.Status
+ */
+
+//TODO: labels
+case class KubernetesCluster(id: KubernetesClusterLeoId,
+                             googleProject: GoogleProject,
+                             clusterName: KubernetesClusterName,
+                             location: Location,
+                             status: KubernetesClusterStatus,
+                             serviceAccountInfo: WorkbenchEmail,
+                             samResourceId: KubernetesClusterSamResource,
+                             auditInfo: KubernetesAuditInfo,
+                             asyncFields: KubernetesClusterAsyncFields,
+                             namespaces: Set[KubernetesNamespaceName],
+                             labels: LabelMap,
+                             nodepools: Set[Nodepool]
+                            )
+
+object KubernetesCluster {
+  implicit class EnrichedKubernetesCluster(cluster: KubernetesCluster) {
+    def getGKEClusterId: KubernetesClusterId = KubernetesClusterId(cluster.googleProject, cluster.location, cluster.clusterName)
+  }
+}
+
+case class SaveKubernetesCluster(googleProject: GoogleProject,
+                                 clusterName: KubernetesClusterName,
+                                 location: Location,
+                                 status: KubernetesClusterStatus,
+                                 serviceAccountInfo: WorkbenchEmail,
+                                 samResourceId: KubernetesClusterSamResource,
+                                 auditInfo: KubernetesAuditInfo,
+                                 labels: LabelMap,
+                                 initialNodepool: Nodepool) //the clusterId specified here isn't used, and will be replaced by the id of cluster saved beforehand
+
+case class KubernetesClusterSamResource(resourceId: String)
+
+case class KubernetesAuditInfo(creator: WorkbenchEmail,
+                               createdDate: Instant,
+                               destroyedDate: Option[Instant])
+
+case class KubernetesClusterAsyncFields(apiServerIp: Option[KubernetesMasterIP],
+                                        networkInfo: Option[NetworkFields]
+                                       )
+
+case class NetworkFields(networkName: NetworkName,
+                         subNetworkName: SubnetworkName,
+                         subNetworkIpRange: IpRange)
+
+
+//TODO: possibly consolidate Kubernetes status when monitoring is implemented
+sealed trait KubernetesClusterStatus extends EnumEntry with Product with Serializable
+object KubernetesClusterStatus extends Enum[KubernetesClusterStatus] {
+  val values = findValues
+
+  // NOTE: Remember to update the definition of this enum in Swagger when you add new ones
+  case object Status_Unspecified extends KubernetesClusterStatus
+  case object Provisioning extends KubernetesClusterStatus
+  case object Running extends KubernetesClusterStatus
+  case object Reconciling extends KubernetesClusterStatus
+  case object Stopping extends KubernetesClusterStatus
+  case object Error extends KubernetesClusterStatus
+  case object Degraded extends KubernetesClusterStatus
+}
+
+//TODO: possibly consolidate Kubernetes status when monitoring is implemented
+sealed trait NodepoolStatus extends EnumEntry with Product with Serializable
+object NodepoolStatus extends Enum[NodepoolStatus] {
+  val values = findValues
+
+  // NOTE: Remember to update the definition of this enum in Swagger when you add new ones
+  case object Status_Unspecified extends NodepoolStatus
+  case object Provisioning extends NodepoolStatus
+  case object Running extends NodepoolStatus
+  case object Running_with_Error extends NodepoolStatus //consistent with Degraded from KubernetesClusterStatus
+  case object Reconciling extends NodepoolStatus
+  case object Stopping extends NodepoolStatus
+  case object Error extends NodepoolStatus
+}
+
+case class KubernetesClusterLeoId(id: Long)
+case class KubernetesNamespaceId(id: Long)
+
+case class Nodepool(id: NodepoolLeoId,
+                    clusterId: KubernetesClusterLeoId,
+                    nodepoolName: NodePoolName,
+                    status: NodepoolStatus,
+                    auditInfo: KubernetesAuditInfo,
+                    machineType: MachineTypeName,
+                    numNodes: NumNodes,
+                    autoScalingEnabled: Boolean,
+                    autoscalingConfig: Option[NodepoolAutoscaling])
+
+//TODO: remove?
+//case class SaveNodePool(clusterId: KubernetesClusterLeoId,
+//                    nodePoolName: NodePoolName,
+//                    status: NodePoolStatus,
+//                    auditInfo: KubernetesAuditInfo,
+//                    machineType: MachineTypeName,
+//                    numNodes: NumNodes,
+//                    autoScalingEnabled: Boolean,
+//                    autoscalingConfig: Option[NodePoolAutoscaling])
+
+case class NodepoolAutoscaling(autoScalingMin: AutoScalingMin,
+                               autoScalingMax: AutoScalingMax)
+
+case class NodepoolLeoId(id: Long)
+case class NumNodes(amount: Int)
+case class AutoScalingMin(amount: Int)
+case class AutoScalingMax(amount: Int)
+
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
@@ -1,6 +1,6 @@
 package org.broadinstitute.dsde.workbench.leonardo
 
-import enumeratum.{Enum, EnumEntry}
+import ca.mrvisser.sealerate
 import org.broadinstitute.dsde.workbench.google2.GKEModels.{KubernetesClusterId, KubernetesClusterName, NodepoolName}
 import org.broadinstitute.dsde.workbench.google2.KubernetesModels.KubernetesApiServerIp
 import org.broadinstitute.dsde.workbench.google2.KubernetesSerializableName.KubernetesNamespaceName
@@ -23,20 +23,18 @@ case class KubernetesCluster(id: KubernetesClusterLeoId,
                             )
 
 object KubernetesCluster {
-
   implicit class EnrichedKubernetesCluster(cluster: KubernetesCluster) {
     def getGkeClusterId: KubernetesClusterId = KubernetesClusterId(cluster.googleProject, cluster.location, cluster.clusterName)
   }
-
 }
 
-case class KubernetesClusterSamResource(resourceId: String)
+final case class KubernetesClusterSamResource(resourceId: String)
 
-case class KubernetesClusterAsyncFields(apiServerIp: KubernetesApiServerIp,
+final case class KubernetesClusterAsyncFields(apiServerIp: KubernetesApiServerIp,
                                         networkInfo: NetworkFields
                                        )
 
-case class NetworkFields(networkName: NetworkName,
+final case class NetworkFields(networkName: NetworkName,
                          subNetworkName: SubnetworkName,
                          subNetworkIpRange: IpRange)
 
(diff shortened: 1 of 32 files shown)


core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
```diff
@@ -0,0 +1,106 @@
...
+                             nodepools: Set[Nodepool]
+                            )
+
+object KubernetesCluster {
+  implicit class EnrichedKubernetesCluster(cluster: KubernetesCluster) {
+    def getGKEClusterId: KubernetesClusterId = KubernetesClusterId(cluster.googleProject, cluster.location, cluster.clusterName)
+  }
+}
+
+case class SaveKubernetesCluster(googleProject: GoogleProject,
```
Qi77Qi: If this is database specific model, I think we can move it to DB layer. Also I don't think the comment is accurate

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
```diff
@@ -0,0 +1,106 @@
...
+                                 clusterName: KubernetesClusterName,
+                                 location: Location,
+                                 status: KubernetesClusterStatus,
+                                 serviceAccountInfo: WorkbenchEmail,
+                                 samResourceId: KubernetesClusterSamResource,
+                                 auditInfo: AuditInfo,
+                                 labels: LabelMap,
+                                 initialNodepool: Nodepool) //the clusterId specified here isn't used, and will be replaced by the id of cluster saved beforehand
+
+case class KubernetesClusterSamResource(resourceId: String)
```
Qi77Qi: can we add `final` to add case classes that we don't expect anything to extend from

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
```diff
@@ -0,0 +1,106 @@
...
+
+case class KubernetesClusterAsyncFields(apiServerIp: Option[KubernetesApiServerIp],
```
Qi77Qi: is there really 4 different permutations of these 2 fields? or can we make this 
```
final case class KubernetesClusterAsyncFields(apiServerIp: KubernetesApiServerIp,networkInfo: NetworkFields)
```
and at call site this asyncFields as a whole is optional?

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
```diff
@@ -0,0 +1,106 @@
...
+                             auditInfo: AuditInfo,
+                             asyncFields: KubernetesClusterAsyncFields,
+                             namespaces: Set[KubernetesNamespaceName],
+                             labels: LabelMap,
+                             nodepools: Set[Nodepool]
+                            )
+
+object KubernetesCluster {
+  implicit class EnrichedKubernetesCluster(cluster: KubernetesCluster) {
+    def getGKEClusterId: KubernetesClusterId = KubernetesClusterId(cluster.googleProject, cluster.location, cluster.clusterName)
```
kyuksel: Naming-wise, it'd be nice to follow what we've done with `GCE`, which is not capitalizing `C` and `E` (i.e. `getGkeClusterId`)

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
kyuksel: Would `SavedKubernetesCluster` read better?

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
```diff
@@ -0,0 +1,106 @@
...
+                         subNetworkIpRange: IpRange)
+
+
+//TODO: possibly consolidate Kubernetes status when monitoring is implemented
+sealed trait KubernetesClusterStatus extends EnumEntry with Product with Serializable
+object KubernetesClusterStatus extends Enum[KubernetesClusterStatus] {
+  val values = findValues
+
+  // NOTE: Remember to update the definition of this enum in Swagger when you add new ones
+  case object Status_Unspecified extends KubernetesClusterStatus
```
kyuksel: Should ^^ just be `Unspecified`? Same below for `NodepoolStatus`.

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
```diff
@@ -0,0 +1,106 @@
...
+  case object Provisioning extends KubernetesClusterStatus
+  case object Running extends KubernetesClusterStatus
+  case object Reconciling extends KubernetesClusterStatus
+  case object Stopping extends KubernetesClusterStatus
+  case object Error extends KubernetesClusterStatus
+  case object Degraded extends KubernetesClusterStatus
+}
+
+//TODO: possibly consolidate Kubernetes status when monitoring is implemented
```
kyuksel:
```suggestion
//TODO: possibly consolidate Nodepool status when monitoring is implemented
```

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
```diff
@@ -0,0 +1,106 @@
...
+sealed trait NodepoolStatus extends EnumEntry with Product with Serializable
+object NodepoolStatus extends Enum[NodepoolStatus] {
+  val values = findValues
+
+  // NOTE: Remember to update the definition of this enum in Swagger when you add new ones
+  case object Status_Unspecified extends NodepoolStatus
+  case object Provisioning extends NodepoolStatus
+  case object Running extends NodepoolStatus
+  case object Running_with_Error extends NodepoolStatus //consistent with Degraded from KubernetesClusterStatus
```
kyuksel: `RunningWithError` instead?

kyuksel (commented):
Ran out of time as I have to step out for a doctor appointment 😞 Will continue when I'm back. Mainly nits so far.

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
jdcanas: this class is used when saving a cluster to the DB, since not all fields in a `KubernetesCluster` should be specified at creation time. Its analagous to the `SaveCluster` ADT we have. I'll move it to the `KubernetesClusterComponent` class so the intent is more clear

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
jdcanas: These words map directly (underscores and all for the below) to the google container statuses. The comment linking to the documentation got move as I wrote more code, I moved it to be next to these

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
jdcanas: The intent here is that it would be nice to consolidate the statuses we have here such that we don't have to do too much work reimplementing monitoring, since both Kubernetes Nodepools and Clusters use google `Operation`s, which GCE also uses

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
jdcanas: I think its necessary to do it like this since this is the flow

1. Front Leo inserts initial record, `networkInfo` and `apiServerIp` aren't specified
2. Back Leo creates the vpc network and submits the call to google. Now `networkInfo` is specified and `apiServerIp` is pending the creation in google
3. finally, once the creation is complete in google, both `networkFields` and `apiServerIp` will be specified

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
Qi77Qi: I'd prefer in scala code, casing is right...and add a 
```
def asString: String
```
To `NodepoolStatus`, and define it in all product types...Note this is not just to make casing right, it's also to decouple these case classes as types vs their String representation from Google

jdcanas: These words (and thus the strings the underlying enum implementation provides) map directly to the google container statuses.

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
jdcanas: Jinx 😄 

Not sure which comment you mean.

With regard to `//the clusterId specified here isn't used, and will be replaced by the id of cluster saved beforehand`, this is because we want to enforce a `KubernetesCluster` be created with an initial Nodepool, since this is what google requires. However, since the cluster isn't created we replace the `clusterId` in the `Nodepool` object here `nodepool <- nodepoolQuery.saveForCluster(saveCluster.initialNodepool.copy(clusterId = clusterId))`.

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
Qi77Qi: can we make wait until 2,3 to finish before we update DB?...

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
kyuksel: I'm still trying to understand this. Perhaps we can talk offline.

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
kyuksel: Why do our Scala code names have to mirror Google's style exactly?

Qi77Qi: yea. I know this is how current code has been. But we really shouldn't...We should never rely on the `case object` name for its String representation as contract with another service. It's very error prone, if one day someone decides to rename the case object, they shouldn't need to worry that "oh, would this break the contract as well?"...this is the exact reason why we have explicit encoder/decoder for contract case classes, it's not technically required, but it's much safer than relying to magical reflection or other means to derive string representation in contract

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
jdcanas: Hmm.. It's possible, but since they are discrete pollable Operations, I don't see why we wouldn't update them separately as they come available? It seems this would allow better representation of state incase an error occurs at any one of the steps, and clean-up is required of those resources that were successfully created. 

I'm open to doing them together and changing to `Option[AsyncFields]` to be consistent with our `createRuntime` and `CreateRuntimeResponse`, just curious why we chose to bunch all the db updates for the end of creation instead of doing them as each pollable action finishes.

Qi77Qi: my thoughts on this is
1. I don't like we're making it 4 different permutations of just 2 fields. If we're going to add yet another field, it'll be 9 different permuatations. This makes things difficult for caller of this data type because they need to accommodate all cases.
2. As far as I can see, there's not much point to persist networkInfo if apiServerIp failed to create. We should probably clean things up if this happens, so why bother persisting networkInfo in the first case.

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
jdcanas: The premise behind the `Enum` implementation we are using is that your case object name maps via `.withName`, which makes decoding trivial. Is it really worth decoupling these, only to move the coupling to a mapping table we maintain? That would require strings being returned from an `asString` coupled to a map. it seems like unnecessary code to maintain instead of using built-in enum features. Let's jump on a call at some point

Qi77Qi: Yes it's worth the effort. And we shouldn't use `withName` because it can throw exception if one day google decides to add a new status. An example of the approach I'm talking about https://github.com/DataBiosphere/leonardo/blob/87db66a47d2ba1e2946505984fbd5cedded7415a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dataprocModels.scala#L96.

http/src/main/resources/logback.xml
```diff
@@ -64,19 +64,15 @@
...
-    <!--
-    <logger name="slick.jdbc.JdbcBackend.statement" level="DEBUG" additivity="false">
-        <appender-ref ref="file"/>
-        <appender-ref ref="console"/>
-    </logger>
-    -->
+<!--    <logger name="slick.jdbc.JdbcBackend.statement" level="DEBUG" additivity="false">-->
+<!--        <appender-ref ref="file"/>-->
+<!--        <appender-ref ref="console"/>-->
+<!--    </logger>-->
```
jdcanas, line 70: This is how intelliJ does it with `cmd + /`, I don't like typing 😄

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
```diff
@@ -0,0 +1,150 @@
...
+                             auditInfo: AuditInfo,
+                             asyncFields: Option[KubernetesClusterAsyncFields],
+                             namespaces: Set[KubernetesNamespaceName],
+                             labels: LabelMap,
+                             nodepools: Set[Nodepool]
+                            )
+
+object KubernetesCluster {
+  implicit class EnrichedKubernetesCluster(cluster: KubernetesCluster) {
+    def getGkeClusterId: KubernetesClusterId = KubernetesClusterId(cluster.googleProject, cluster.location, cluster.clusterName)
```
rtitle: Could this just be a `def` in the class itself instead of an implicit extension class?

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
```diff
@@ -0,0 +1,150 @@
...
+  }
+}
+
+final case class KubernetesClusterSamResource(resourceId: String)
```
rtitle: Nit: can we make single-parameter case classes extend `AnyVal`?

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
```diff
@@ -0,0 +1,150 @@
...
+
+final case class KubernetesClusterAsyncFields(apiServerIp: KubernetesApiServerIp,
+                                        networkInfo: NetworkFields
```
rtitle: Maybe could use a `scalafmt`

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
akarukappadath: nit: to keep consistent with the persistent disk one, could we call this `KubernetesClusterSamResourceId`?

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
```diff
@@ -0,0 +1,150 @@
...
+                             clusterName: KubernetesClusterName,
+                             location: Location,
+                             status: KubernetesClusterStatus,
+                             serviceAccountInfo: WorkbenchEmail,
+                             samResourceId: KubernetesClusterSamResource,
+                             auditInfo: AuditInfo,
+                             asyncFields: Option[KubernetesClusterAsyncFields],
+                             namespaces: Set[KubernetesNamespaceName],
+                             labels: LabelMap,
+                             nodepools: Set[Nodepool]
```
akarukappadath: Does this need an `errors` field?

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
```diff
@@ -45,44 +43,81 @@ case class NetworkFields(networkName: NetworkName,
  *  see: https://cloud.google.com/kubernetes-engine/docs/reference/rest/v1/projects.locations.clusters#Cluster.Status
  */
 
-//TODO: possibly consolidate Kubernetes status when monitoring is implemented
```
kyuksel: I finally understand what you meant by the comment above (better late than never, eh 😛). I agree with it and would vote for keeping it (perhaps with "Kubernetes status" replaced by "KubernetesClusterStatus'es and NodepoolStatus'es").

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
```diff
@@ -98,14 +105,20 @@ object NodepoolStatus {
     override def toString: String = "RECONCILING"
   }
 
-  case object Stopping extends NodepoolStatus {
+  //in kubernetes statuses, they label "deleting' resources as 'stopping'
```
kyuksel, line 104: That sounds confusing. Should we map it to `stopping` internally?

kyuksel (commented):
Looks good. One question inline about `Stopping` vs. `Deleting`.

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
```diff
@@ -0,0 +1,150 @@
...
+
+  case object RunningWithError extends NodepoolStatus {
+    override def toString: String = "RUNNING_WITH_ERROR"
+  }
+
+  def values: Set[NodepoolStatus] = sealerate.values[NodepoolStatus]
+  def stringToObject: Map[String, NodepoolStatus] = values.map(v => v.toString -> v).toMap
+}
+
+final case class KubernetesClusterLeoId(id: Long)
```
Qi77Qi: `extend AnyVal` for all single param case classes so that they don't have extra boxing in runtime (in most cases)

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20200501_kubernetes_tables.xml
```diff
@@ -0,0 +1,113 @@
...
+            <column name="clusterName" type="VARCHAR(254)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="location" type="VARCHAR(254)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="status" type="VARCHAR(254)" defaultValue="STATUS_UNSPECIFIED">
+                <constraints nullable="false"/>
+            </column>
+            <column name="serviceAccountInfo" type="VARCHAR(254)">
```
rtitle: Should we just name this `serviceAccount`? `Info` seems a little confusing 🤔

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20200501_kubernetes_tables.xml
```diff
@@ -0,0 +1,113 @@
...
+            </column>
+            <column name="dateAccessed" type="TIMESTAMP(6)" defaultValue="0000-00-00 00:00:00.000000">
+                <constraints nullable="false"/>
+            </column>
+            <column name="apiServerIp" type="VARCHAR(254)"/>
+            <column name="networkName" type="VARCHAR(254)"/>
+            <column name="subNetworkName" type="VARCHAR(254)"/>
+            <column name="subNetworkIpRange" type="VARCHAR(254)"/>
+        </createTable>
+        <createIndex indexName="IDX_KUBERNETES_CLUSTER_ID" tableName="KUBERNETES_CLUSTER">
```
rtitle: This index may not be needed since it's already the primary key

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20200501_kubernetes_tables.xml
```diff
@@ -0,0 +1,113 @@
...
+                <constraints nullable="false"/>
+            </column>
+            <column name="autoScalingEnabled" type="boolean">
+                <constraints nullable="false"/>
+            </column>
+            <column name="autoScalingMin" type="INT"/>
+            <column name="autoScalingMax" type="INT"/>
+        </createTable>
+        <addForeignKeyConstraint baseColumnNames="clusterId" baseTableName="NODEPOOL" constraintName="FK_NODEPOOL_CLUSTER_ID" deferrable="false" initiallyDeferred="false" onDelete="NO ACTION" onUpdate="NO ACTION" referencedColumnNames="id" referencedTableName="KUBERNETES_CLUSTER"/>
+        <createIndex indexName="IDX_NODEPOOL_ID" tableName="NODEPOOL">
```
rtitle: Same here, think this is unnecessary

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20200501_kubernetes_tables.xml
Qi77Qi: yea, I think serviceAccountInfo was a deprecated case class name (we had it when we had notebookSA)

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20200501_kubernetes_tables.xml
```diff
@@ -0,0 +1,113 @@
...
+                <constraints nullable="false"/>
+            </column>
+        </createTable>
+        <addForeignKeyConstraint baseColumnNames="clusterId" baseTableName="NAMESPACE" constraintName="FK_NAMESPACE_CLUSTER_ID" deferrable="false" initiallyDeferred="false" onDelete="NO ACTION" onUpdate="NO ACTION" referencedColumnNames="id" referencedTableName="KUBERNETES_CLUSTER"/>
+        <createIndex indexName="IDX_NAMESPACE_ID" tableName="NAMESPACE">
+            <column name="id"/>
+        </createIndex>
+        <addUniqueConstraint columnNames="clusterId, namespaceName" constraintName="IDX_NAMESPACE_UNIQUE" tableName="NAMESPACE"/>
+    </changeSet>
+</databaseChangeLog>
```
rtitle, line 104: Migration seems fine, but as a sanity check I like to do:
```
DESCRIBE <tableName>
SHOW INDEX FROM <tableName>
```
on any new/changed tables. Can do this on the local unit-test MySQL DB.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesClusterComponent.scala
```diff
@@ -0,0 +1,262 @@
...
+import nodepoolQuery.unmarshalNodepool
+import org.broadinstitute.dsde.workbench.google2.KubernetesSerializableName.KubernetesNamespaceName
+import org.broadinstitute.dsde.workbench.leonardo.db.LeoProfile.{dummyDate, unmarshalDestroyedDate}
+
+import scala.concurrent.ExecutionContext
+
+case class KubernetesClusterRecord(id: KubernetesClusterLeoId,
+                                   googleProject: GoogleProject,
+                                   clusterName: KubernetesClusterName,
+                                   // the GKE API actually supports a location (e.g. us-central1 or us-central1-a)
```
rtitle: This is a nice comment -- can we put it in `core/kubernetesModels.scala` which is probably where people will mostly look?

rtitle:
Did a pass -- overall looks good! Just a few nits above. I can do another pass before merge.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesClusterComponent.scala
kyuksel: Agreed. Reading this comment also made me wonder if we should reflect the fact that a `Location` can be a `Region` or a `Zone` in our workbench-libs modeling of it  (via an ADT or such) so it's more explicit. I remember being thrown off when I first did anything with it.

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
jdcanas: Yeah, that's essentially what I've done. If google returns a status of "STOPPING", then it really means "Deleting" as far as we are concerned.

In other words, the way we serializable a value of "STOPPING" from google will be as our "Deleting" case object

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20200501_kubernetes_tables.xml
jdcanas: huh, I didn't know primary keys were automatically indexed in SQL, TIL

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
kyuksel: Does that work with 

```scala
case KubernetesClusterStatus.Deleting => Deleting
```

in `fromKubernetesClusterStatus()`? It felt like it should be

```scala
case KubernetesClusterStatus.Deleting => Stopping
```

🤔

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesClusterComponent.scala
jdcanas: Yeah... next pr to wb-libs we should fix it, and make some sort of hierarchy where Location is an interface and `Zone` and `Region` inherit from it. Currently, `Location` already exists as an ADT in some experimental service file. 

I held off on it because it falls under a larger category that I want to address of 'stray google2 models that are applicable to more things than the Service/Interp where they are located'

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
jdcanas: fromKubernetesClusterStatus maps a KubernetesClusterStatus to a RuntimeStatus

So what `case KubernetesClusterStatus.Deleting => Deleting` is saying is that `KubernetesClusterStatus.Deleting` is analogous to `RuntimeStatus.Deleting`. In other words the `STOPPIN` status that google returns maps to `KubernetesClusterStatus.Deleting` which maps to `RuntimeStatus.Deleting`

kyuksel (approved):

rtitle (approved):
Looks good to me!

One request though: can you run `scalafmt` and `test:scalafmt` from an sbt shell, in all modules? It just feels like it'll pick some things up.

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
```diff
@@ -0,0 +1,154 @@
...
+                             location: Location,
+                             status: KubernetesClusterStatus,
+                             serviceAccount: WorkbenchEmail,
+                             samResourceId: KubernetesClusterSamResourceId,
+                             auditInfo: AuditInfo,
+                             asyncFields: Option[KubernetesClusterAsyncFields],
+                             namespaces: Set[KubernetesNamespaceName],
+                             labels: LabelMap,
+                             nodepools: Set[Nodepool],
+                             //TODO: populate this
```
Qi77Qi, line 27: is the comment still valid?

Qi77Qi (approved):

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
jdcanas: yeah, I punted on error handling here

Qi77Qi: hmm, I didn't even realize this is for error handling...can you make the comment a bit more concrete

jdcanas:
> Looks good to me!
> 
> One request though: can you run `scalafmt` and `test:scalafmt` from an sbt shell, in all modules? It just feels like it'll pick some things up.

Huh... I've been running `sbt "project http" scalafmt`... it caught some of the things in the modules but not all of them for some reason

Qi77Qi:
when you do `sbt "project http" scalafmt`, you're telling it to only care about `http` project

jdcanas:
> when you do `sbt "project http" scalafmt`, you're telling it to only care about `http` project

Are tests not part of this project? I thought they were. I've been running it for `core` too

Qi77Qi:
`sbt scalafmt` and `sbt test:scalafmt` should suffice, you don't need to run it on individual projects...tests are treatedly different l think

rtitle:
I open the `sbt` shell and then run `scalafmt` and `test:scalafmt`. That formats `core`, `http`, and `automation` modules.

kyuksel:
Curious if anyone has experienced IntelliJ scalafmt plug-in miss any formatting from any of the modules?

PR #1407, opened 2020-05-06
IA-1897 listCluster performance regression

rtitle:
https://broadworkbench.atlassian.net/browse/IA-1897

The real fix here is to make `LABEL.resourceId` the same type as `CLUSTER.id` and `PERSISTENT_DISK.id`. MySQL can join columns with different types but there's a significant performance penalty as it has to convert one type to the other for each row.

Also improved some slick code although I don't think slick was really to blame in this case. (Still pro moving to writing direct SQL though)

I tested this on dev by manually making the schema changes (will revert before merge) and running some slick-generated SQL queries. 

---

In all cases:

### DIFF
--- a/http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20200505_change_label_resource_id_type.xml
+++ b/http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20200505_change_label_resource_id_type.xml
@@ -1,0 +1,22 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="leonardo" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+    <!-- See https://broadworkbench.atlassian.net/browse/IA-1897 -->
+    <changeSet logicalFilePath="leonardo" author="rtitle" id="change-label-resource-id-type">
+        <!-- Add `resourceId2` column with correct type and populate it from `resourceId` -->
+        <addColumn tableName="LABEL">
+            <column name="resourceId2" type="bigint(20)">
+                <constraints nullable="true" />
+            </column>
+        </addColumn>
+        <sql>UPDATE LABEL SET `resourceId2` = CAST(`resourceId` AS UNSIGNED INTEGER)</sql>
+
+        <!-- Drop the existing `IDX_LABEL_UNIQUE` constraint and `resourceId` column -->
+        <dropUniqueConstraint uniqueColumns="resourceId, resourceType, key" tableName="LABEL" constraintName="IDX_LABEL_UNIQUE" />
+        <dropColumn tableName="LABEL" columnName="resourceId" />
+
+        <!-- Rename `resourceId2` to `resourceId` and re-add not-null and unique constraints -->
+        <renameColumn tableName="LABEL" oldColumnName="resourceId2" newColumnName="resourceId" columnDataType="bigint(20)" />
+        <addNotNullConstraint columnName="resourceId" columnDataType="bigint(20)" tableName="LABEL"/>
+        <addUniqueConstraint tableName="LABEL" columnNames="resourceId, resourceType, key" constraintName="IDX_LABEL_UNIQUE" />
+    </changeSet>
+</databaseChangeLog>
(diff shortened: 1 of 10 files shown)


http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20200505_change_label_resource_id_type.xml
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
rtitle, line 1: Table looks good locally after this migration:
```
mysql> DESCRIBE LABEL;
+--------------+--------------+------+-----+---------+-------+
| Field        | Type         | Null | Key | Default | Extra |
+--------------+--------------+------+-----+---------+-------+
| key          | varchar(254) | NO   | PRI | NULL    |       |
| value        | varchar(254) | NO   |     | NULL    |       |
| resourceType | varchar(254) | NO   | PRI | NULL    |       |
| resourceId   | bigint(20)   | NO   | PRI | NULL    |       |
+--------------+--------------+------+-----+---------+-------+
4 rows in set (0.00 sec)

mysql> SHOW INDEX FROM LABEL;
+-------+------------+------------------+--------------+--------------+-----------+-------------+----------+--------+------+------------+---------+---------------+
| Table | Non_unique | Key_name         | Seq_in_index | Column_name  | Collation | Cardinality | Sub_part | Packed | Null | Index_type | Comment | Index_comment |
+-------+------------+------------------+--------------+--------------+-----------+-------------+----------+--------+------+------------+---------+---------------+
| LABEL |          0 | IDX_LABEL_UNIQUE |            1 | resourceId   | A         |           0 |     NULL | NULL   |      | BTREE      |         |               |
| LABEL |          0 | IDX_LABEL_UNIQUE |            2 | resourceType | A         |           0 |     NULL | NULL   |      | BTREE      |         |               |
| LABEL |          0 | IDX_LABEL_UNIQUE |            3 | key          | A         |           0 |     NULL | NULL   |      | BTREE      |         |               |
+-------+------------+------------------+--------------+--------------+-----------+-------------+----------+--------+------+------------+---------+---------------+
3 rows in set (0.00 sec)
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
```diff
@@ -197,7 +197,10 @@ object clusterQuery extends TableQuery(new ClusterTable(_)) {
   //   left join label l on c.id = l.clusterId
   val clusterLabelQuery: Query[(ClusterTable, Rep[Option[LabelTable]]), (ClusterRecord, Option[LabelRecord]), Seq] = {
     for {
-      (cluster, label) <- clusterQuery joinLeft labelQuery.runtimeLabels on (_.id === _.resourceId)
```
rtitle, line 200: I removed `labelQuery.runtimeLabels` -- I think it's more efficient to do the `lbl.resourceType === LabelResourceType.runtime` filter in the join as opposed to a separate inner query.

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/db/RuntimeServiceDbQueriesSpec.scala
```diff
@@ -0,0 +1,124 @@
...
+  RuntimeServiceDbQueries,
+  TestComponent
+}
+import org.broadinstitute.dsde.workbench.leonardo.http.api.ListRuntimeResponse2
+import org.scalatest.FlatSpecLike
+import org.scalatest.concurrent.ScalaFutures
+
+import scala.concurrent.ExecutionContext.Implicits.global
+
+class RuntimeServiceDbQueriesSpec extends FlatSpecLike with TestComponent with GcsPathUtils with ScalaFutures {
```
rtitle, line 23: Added this because there wasn't an existing spec for `RuntimeServiceDbQueries` (just `LeonardoServiceDbQueries`).

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/db/RuntimeServiceDbQueriesSpec.scala
```diff
@@ -0,0 +1,124 @@
...
+      list3 <- RuntimeServiceDbQueries.listClusters(Map.empty, false, None).transaction
+    } yield {
+      list1 shouldEqual List.empty
+      val c1Expected = toListRuntimeResponse(c1, Map.empty)
+      val c2Expected = toListRuntimeResponse(c2, Map.empty)
+      list2 shouldEqual List(c1Expected)
+      list3.toSet shouldEqual Set(c1Expected, c2Expected)
+    }
+
+    res.unsafeRunSync()
```
Qi77Qi, line 46: would it make sense to add a check to make sure the query doesn't take more than say 5 seconds?

Qi77Qi (approved):
lgtm 👍

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/db/RuntimeServiceDbQueriesSpec.scala
rtitle: Sure could add something like that :)

PR #1413, opened 2020-05-12
[IA-1580] Persistent disk CRUD routes

akarukappadath:

---

In all cases:

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/diskModels.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/diskModels.scala
@@ -20,7 +20,7 @@ final case class PersistentDisk(id: DiskId,
   def projectNameString: String = s"${googleProject.value}/${name.value}"
 }
 
-final case class DiskId(id: Long) extends AnyVal
+final case class DiskId(value: Long) extends AnyVal
 final case class DiskSamResourceId(asString: String) extends AnyVal
 
 /** Default persistent disk labels */
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/diskModels.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/diskModels.scala
@@ -48,6 +48,21 @@ object DiskStatus extends Enum[DiskStatus] {
   final case object Ready extends DiskStatus
   final case object Deleting extends DiskStatus
   final case object Deleted extends DiskStatus
+
+  //TODO: confirm these
+  val activeStatuses: Set[DiskStatus] =
+    Set(Creating, Restoring, Ready)
+
+  val deletableStatuses: Set[DiskStatus] =
+    Set(Failed, Ready)
+
+  val updatableStatuses: Set[DiskStatus] = Set(Ready)
+
+  implicit class EnrichedDiskStatus(status: DiskStatus) {
+    def isActive: Boolean = activeStatuses contains status
+    def isDeletable: Boolean = deletableStatuses contains status
+    def isUpdatable: Boolean = updatableStatuses contains status
+  }
 }
 
 // Disks are always specified in GB, it doesn't make sense to support other units
@@ -69,4 +84,9 @@ object DiskType extends Enum[DiskType] {
   final case object SSD extends DiskType {
     override def googleString: String = "pd-ssd"
   }
+  def stringToDiskType(string: String): DiskType =
+    string match {
+      case "pd-standard" => Standard
+      case "pd-ssd"      => SSD
+    }
 }
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/diskModels.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/diskModels.scala
@@ -24,16 +24,12 @@ final case class DiskId(id: Long) extends AnyVal
 final case class DiskSamResourceId(asString: String) extends AnyVal
 
 /** Default persistent disk labels */
-case class DefaultDiskLabels(diskName: DiskName,
-                             googleProject: GoogleProject,
-                             creator: WorkbenchEmail,
-                             serviceAccount: WorkbenchEmail) {
+case class DefaultDiskLabels(diskName: DiskName, googleProject: GoogleProject, creator: WorkbenchEmail) {
   def toMap: LabelMap =
     Map(
       "diskName" -> diskName.value,
       "googleProject" -> googleProject.value,
-      "creator" -> creator.value,
-      "clusterServiceAccount" -> serviceAccount.value
+      "creator" -> creator.value
     ).filterNot(_._2 == null)
 }
 
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/diskModels.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/diskModels.scala
@@ -38,6 +38,7 @@ sealed trait DiskStatus extends EnumEntry
 object DiskStatus extends Enum[DiskStatus] {
   val values = findValues
 
+  //TODO: Create Pre statuses once https://github.com/DataBiosphere/leonardo/pull/1395/files#diff-4101c04c4a7015e058bf48267899df0bR92 is merged
   final case object Creating extends DiskStatus
   final case object Restoring extends DiskStatus
   final case object Failed extends DiskStatus
@@ -45,7 +46,6 @@ object DiskStatus extends Enum[DiskStatus] {
   final case object Deleting extends DiskStatus
   final case object Deleted extends DiskStatus
 
-  //TODO: confirm these
   val activeStatuses: Set[DiskStatus] =
     Set(Creating, Restoring, Ready)
 
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/diskModels.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/diskModels.scala
@@ -1,6 +1,5 @@
 package org.broadinstitute.dsde.workbench.leonardo
 
-import ca.mrvisser.sealerate
 import enumeratum.{Enum, EnumEntry}
 import org.broadinstitute.dsde.workbench.google2.{DiskName, ZoneName}
 import org.broadinstitute.dsde.workbench.model.WorkbenchEmail
@@ -69,11 +68,11 @@ final case class DiskSize(gb: Int) extends AnyVal {
 
 final case class BlockSize(bytes: Int) extends AnyVal
 
-sealed trait DiskType extends Product with Serializable {
+sealed trait DiskType extends EnumEntry {
   def googleString: String
 }
-object DiskType {
-  val allDiskTypes = sealerate.values[DiskType]
+object DiskType extends Enum[DiskType] {
+  val values = findValues
 
   final case object Standard extends DiskType {
     override def googleString: String = "pd-standard"
@@ -81,9 +80,4 @@ object DiskType {
   final case object SSD extends DiskType {
     override def googleString: String = "pd-ssd"
   }
-  def stringToDiskType(string: String): Option[DiskType] =
-    allDiskTypes
-      .map(obj => obj.googleString -> obj)
-      .toMap
-      .get(string)
 }
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/diskModels.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/diskModels.scala
@@ -2,6 +2,7 @@ package org.broadinstitute.dsde.workbench.leonardo
 
 import enumeratum.{Enum, EnumEntry}
 import org.broadinstitute.dsde.workbench.google2.{DiskName, ZoneName}
+import org.broadinstitute.dsde.workbench.model.WorkbenchEmail
 import org.broadinstitute.dsde.workbench.model.google.GoogleProject
 
 final case class PersistentDisk(id: DiskId,
@@ -15,10 +16,26 @@ final case class PersistentDisk(id: DiskId,
                                 size: DiskSize,
                                 diskType: DiskType,
                                 blockSize: BlockSize,
-                                labels: LabelMap)
+                                labels: LabelMap) {
+  def projectNameString: String = s"${googleProject.value}/${name.value}"
+}
 
 final case class DiskId(id: Long) extends AnyVal
 final case class DiskSamResourceId(asString: String) extends AnyVal
+
+/** Default persistent disk labels */
+case class DefaultDiskLabels(diskName: DiskName,
+                             googleProject: GoogleProject,
+                             creator: WorkbenchEmail,
+                             serviceAccount: WorkbenchEmail) {
+  def toMap: LabelMap =
+    Map(
+      "diskName" -> diskName.value,
+      "googleProject" -> googleProject.value,
+      "creator" -> creator.value,
+      "clusterServiceAccount" -> serviceAccount.value
+    ).filterNot(_._2 == null)
+}
 
 // See https://cloud.google.com/compute/docs/reference/rest/v1/disks
 sealed trait DiskStatus extends EnumEntry
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/diskModels.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/diskModels.scala
@@ -1,5 +1,6 @@
 package org.broadinstitute.dsde.workbench.leonardo
 
+import ca.mrvisser.sealerate
 import enumeratum.{Enum, EnumEntry}
 import org.broadinstitute.dsde.workbench.google2.{DiskName, ZoneName}
 import org.broadinstitute.dsde.workbench.model.WorkbenchEmail
@@ -68,11 +69,11 @@ final case class DiskSize(gb: Int) extends AnyVal {
 
 final case class BlockSize(bytes: Int) extends AnyVal
 
-sealed trait DiskType extends EnumEntry {
+sealed trait DiskType extends Product with Serializable {
   def googleString: String
 }
-object DiskType extends Enum[DiskType] {
-  val values = findValues
+object DiskType {
+  val allDiskTypes = sealerate.values[DiskType]
 
   final case object Standard extends DiskType {
     override def googleString: String = "pd-standard"
@@ -80,9 +81,9 @@ object DiskType extends Enum[DiskType] {
   final case object SSD extends DiskType {
     override def googleString: String = "pd-ssd"
   }
-  def stringToDiskType(string: String): DiskType =
-    string match {
-      case "pd-standard" => Standard
-      case "pd-ssd"      => SSD
-    }
+  def stringToDiskType(string: String): Option[DiskType] =
+    allDiskTypes
+      .map(obj => obj.googleString -> obj)
+      .toMap
+      .get(string)
 }
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -372,7 +372,7 @@ zombieRuntimeMonitor {
 persistentDisk {
   defaultDiskSizeGB = 500
   defaultDiskType = "pd-standard"
-  defaultBlockSizeBytes = 32 # TODO: figure out what this really should be
+  defaultBlockSizeBytes = 4096
 }
 
 clusterToolMonitor {
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -369,6 +369,12 @@ zombieRuntimeMonitor {
   concurrency = 100
 }
 
+persistentDisk {
+  defaultDiskSizeGB = 500
+  defaultDiskType = "pd-standard"
+  defaultBlockSizeBytes = 32 # TODO: figure out what this really should be
+}
+
 clusterToolMonitor {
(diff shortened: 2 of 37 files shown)


core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/diskModels.scala
```diff
@@ -15,11 +16,27 @@ final case class PersistentDisk(id: DiskId,
...
+}
 
 final case class DiskId(id: Long) extends AnyVal
 final case class DiskSamResourceId(asString: String) extends AnyVal
 
+/** Default persistent disk labels */
+case class DefaultDiskLabels(diskName: DiskName,
+                             googleProject: GoogleProject,
+                             creator: WorkbenchEmail,
+                             serviceAccount: WorkbenchEmail) {
```
rtitle: Not sure `serviceAccount` really makes sense as a disk label.. you don't need a service account to create a disk.

Qi77Qi: what's the purpose of these labels? are they get created as google labels?

http/src/main/resources/reference.conf
```diff
@@ -369,6 +369,12 @@ zombieRuntimeMonitor {
   concurrency = 100
 }
 
+persistentDisk {
+  defaultDiskSizeGB = 500
+  defaultDiskType = "pd-standard"
+  defaultBlockSizeBytes = 32 # TODO: figure out what this really should be
```
rtitle: From docs the allowed values are 4096 and 16384. Probably use 4096 as a default.

https://cloud.google.com/compute/docs/reference/rest/v1/disks/insert

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/DiskRoutes.scala
```diff
@@ -0,0 +1,237 @@
...
+import org.broadinstitute.dsde.workbench.leonardo.http.api.DiskRoutes._
+import org.broadinstitute.dsde.workbench.model.{TraceId, UserInfo}
+
+class DiskRoutes(diskService: DiskService[IO], userInfoDirectives: UserInfoDirectives)(
+  implicit timer: Timer[IO]
+) {
+  val routes: server.Route = userInfoDirectives.requireUserInfo { userInfo =>
+    CookieSupport.setTokenCookie(userInfo, CookieSupport.tokenCookieName) {
+      implicit val traceId = ApplicativeAsk.const[IO, TraceId](TraceId(UUID.randomUUID()))
+      pathPrefix("google" / "v1" / "disk") {
```
rtitle: I think we wanted to use plural routes, so `/api/google/v1/disks`

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/diskModels.scala
akarukappadath: @rtitle thanks - meant to leave a comment asking if SAs were relevant for disks. 
@Qi77Qi the disk labels don't get persisted to google (similar to runtimes) - it's just another way to filter disks like runtimes

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -49,6 +49,7 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift](
         handleStartRuntimeMessage(msg, now)
       case msg: UpdateRuntimeMessage =>
         handleUpdateRuntimeMessage(msg, now)
+      case _ => ??? //TODO: put disk messages here
```
rtitle: Cool, so handling this will be the work for https://broadworkbench.atlassian.net/browse/IA-1757 (cc @gpcarr)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/DiskServiceInterp.scala
```diff
@@ -0,0 +1,233 @@
...
+    googleProject: GoogleProject,
+    diskName: DiskName,
+    req: CreateDiskRequest
+  )(implicit as: ApplicativeAsk[F, AppContext]): F[Unit] =
+    for {
+      ctx <- as.ask
+      hasPermission <- authProvider.hasProjectPermission(userInfo, CreatePersistentDisk, googleProject)
+      _ <- if (hasPermission) F.unit else F.raiseError[Unit](AuthorizationError(Some(userInfo.userEmail)))
+      // Grab the service accounts from serviceAccountProvider for use later
+      clusterServiceAccountOpt <- serviceAccountProvider
```
rtitle: Don't think we need to get a pet to create a disk

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/DiskServiceInterp.scala
```diff
@@ -0,0 +1,233 @@
...
+            )
+            _ <- authProvider
+              .notifyPersistentDiskCreated(samResourceId, userInfo.userEmail, googleProject)
+              .handleErrorWith { t =>
+                log.error(t)(
+                  s"[${ctx.traceId}] Failed to notify the AuthProvider for creation of persistent disk ${disk.projectNameString}"
+                ) >> F.raiseError(t)
+              }
+
+            _ <- persistentDiskQuery.save(disk).transaction
```
rtitle: We may want to enqueue _before_ persisting the DB record. Reason is we only want to save the record if enqueuing succeeds.

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/diskModels.scala
```diff
@@ -15,11 +16,27 @@ final case class PersistentDisk(id: DiskId,
...
+      "diskName" -> diskName.value,
+      "googleProject" -> googleProject.value,
+      "creator" -> creator.value,
+      "clusterServiceAccount" -> serviceAccount.value
+    ).filterNot(_._2 == null)
+}
+
 // See https://cloud.google.com/compute/docs/reference/rest/v1/disks
 sealed trait DiskStatus extends EnumEntry
 object DiskStatus extends Enum[DiskStatus] {
```
rtitle, line 38: Wondering if we'll want a `PreCreating` status similar to runtimes..

rtitle:
I know this is WIP but added a few suggestions. Looks great overall! I'll re-review when ready but I think this will be close to merging once tests are added 🚀

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/diskModels.scala
Qi77Qi: in that case...do we still need labels like `diskName, googleProject, creator`? We can filter disks by these already without labels I think?

rtitle: Still think it's useful even if you can filter by googleProject, diskName in other ways. For instance Terra UI lists runtimes like this:
```
https://notebooks.firecloud.org/api/google/v1/runtimes?saturnAutoCreated=true&googleProject=broad-firecloud-dsde&creator=rtitle%40broadinstitute.org
```
They could change to this: 
```
https://notebooks.firecloud.org/api/google/v1/runtimes/broad-firecloud-dsde?saturnAutoCreated=true&creator=rtitle%40broadinstitute.org
```
but either way the DB query is pretty fast. It's nice to support both options.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/DiskServiceInterp.scala
akarukappadath: We need to get `diskId` by saving the disk before we enqueue it, so I don't think I can flip the order

akarukappadath:
jenkins restart

rtitle (approved):
Looks good!

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/DiskServiceInterp.scala
Qi77Qi: in my other PR, we're introducing a `pre` status to represent the period when the even is not being pushed to pubsub yet...I would either wait for my other PR (cuz there're some extra stuff needed for this to happen, or at least add a comment to address this)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/LeoRoutes.scala
```diff
@@ -145,10 +146,11 @@ class LeoRoutes(
 }
 
 object LeoRoutes {
-  private val clusterNameReg = "([a-z|0-9|-])*".r
+  private val resourceNameReg = "([a-z|0-9|-])*".r
```
Qi77Qi: I think we're slowly deprecating this file...We don't really need to add stuff this file anymore...Can we move the validation to `DisksRoutes`'s companion object?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/Config.scala
```diff
@@ -274,6 +274,14 @@ object Config {
     )
   }
 
+  implicit val persistentDiskConfigReader: ValueReader[PersistentDiskConfig] = ValueReader.relative { config =>
```
Qi77Qi: make this private...we shouldn't use this reader outside this file

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/PersistentDiskConfig.scala
```diff
@@ -0,0 +1,9 @@
+package org.broadinstitute.dsde.workbench.leonardo.config
+
+import org.broadinstitute.dsde.workbench.leonardo.{BlockSize, DiskSize, DiskType}
+
+case class PersistentDiskConfig(
```
Qi77Qi: minor, add `final`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/DiskServiceInterp.scala
akarukappadath: Gotcha, I can hold off for now and add a `precreating` status to disks before merging.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
Qi77Qi: can you add something like `throw new NotImplementedError("disk events are not implemented yet")`

Qi77Qi (commented):
are you going to add swagger doc as well? or at least some doc describing the APIs would be nice

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/diskModels.scala
```diff
@@ -52,4 +80,9 @@ object DiskType extends Enum[DiskType] {
   final case object SSD extends DiskType {
     override def googleString: String = "pd-ssd"
   }
+  def stringToDiskType(string: String): DiskType =
```
Qi77Qi: can you use `sealerate.values.map(obj => obj.googleString -> obj` for `stringToDiskType`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/Config.scala
akarukappadath: should I go ahead and make all of the readers here private then?

Qi77Qi: nah...I did that in my other PR...

akarukappadath: making this private made it inaccessible for the configs in `CommonTestData` - did you move those around in your PR as well?

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/diskModels.scala
akarukappadath: Decided to revert back to Enum and use the given `withNameOption` method instead of writing our own here

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -422,7 +424,279 @@ paths:
...
+            default: false
+      responses:
+        "200":
+          description: List of persistent disks
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: "#/components/schemas/PersistentDisk"
```
Qi77Qi: can you name this to `listPersistentDiskResponse`...I'd like to prevent us from reusing this model for other routes so that routes can evolve independent of each other and we won't accidentally update contract for one route but end up updating multiple routes.

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -422,7 +424,279 @@ paths:
...
+          required: true
+          schema:
+            type: string
+      responses:
+        "200":
+          description: Persistent disk found, here are the details
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/PersistentDisk"
```
Qi77Qi: same comment, can this be `GetPersistentDiskResponse`

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -422,7 +424,279 @@ paths:
...
+              schema:
+                $ref: "#/components/schemas/ErrorReport"
+        "500":
+          description: Internal Error
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/ErrorReport"
+    patch:
+      summary: Updates the configuration of a persistent disk
```
Qi77Qi, line 657: if the disk is being used by a running VM, will we be able to update the disk?

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -1382,10 +1656,9 @@ components:
           schema:
             $ref: "#/components/schemas/CreateRuntimeRequest"
           example:
-            runtimeConfig:
-              cloudService: gce
-              machineType: n1-standard-4
-              diskSize: 100
+            size: 500
```
Qi77Qi: is the size increase intentional?

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -1781,6 +2082,57 @@ components:
...
           description: Whether there is a patch in progress on the runtime. Is used to indicate updates that require status transitions.
+    PersistentDisk:
+      description: ""
+      required:
+        - id
+        - googleProject
+        - zone
+        - name
+        - googleId
+        - samResourceId
```
Qi77Qi, line 2129: are we not return `error` for  GetPersistentDisk route?...I don't think `labels` is required...
Not all these fields are applicable to `listPersistentDiskResponse` I think.

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -2087,6 +2439,38 @@ components:
...
+          type: integer
+          description: The block size of the persistent disk to be created in bytes.
+    UpdateDiskRequest:
+      description: "Updates an existing persistent disk"
+      type: object
+      properties:
+        labels:
+          type: object
+          description: The labels to be added or updated on the persistent disk. Of type Map[String,String]
+        newSize:
```
Qi77Qi: I think terra ui preferred these names remain the same...like `size`, `diskType`, `blockSize`...

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/Config.scala
Qi77Qi: yea that's exactly the point...we shouldn't use it in `CommonTestData`...you can use `Config.persistentDisk` (whatever name you give it) in test the same way you do in app code...That way, our unit tests will actually be testing the Config code as well instead of doing its own thing in test..see https://github.com/DataBiosphere/leonardo/pull/1395/files#diff-4101c04c4a7015e058bf48267899df0bR92

http/src/main/resources/swagger/api-docs.yaml
rtitle: +1

http/src/main/resources/swagger/api-docs.yaml
rtitle: +1

http/src/main/resources/swagger/api-docs.yaml
rtitle: You can resize a disk while it's attached to a running VM.

My PR involves adding validation/documentation about what operations are/aren't allowed with disks & VMs.

http/src/main/resources/swagger/api-docs.yaml
rtitle: Hm.. let's keep `CreateRuntimeRequest` unchanged in this PR, I am updating it in my PR

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -1397,6 +1670,25 @@ components:
               cloudService: gce
               machineType: n1-highmem-16
               diskSize: 500
+    CreateDiskRequest:
+      content:
+        application/json:
+          schema:
+            $ref: "#/components/schemas/CreateDiskRequest"
+          example:
+            runtimeConfig:
```
rtitle: Don't think this example is right, there is no `runtimeConfig` or `cloudService`

http/src/main/resources/swagger/api-docs.yaml
rtitle: +1

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -2170,86 +2554,86 @@ components:
     GceConfig:
       description: Configuration for Google Compute Engine instances.
       allOf:
-      - $ref: '#/components/schemas/RuntimeConfig'
```
rtitle, line 2167: Hm.. was spacing wrong here? I touched this recently, would be good to make sure it still displays correctly

rtitle:
Few more comments in swagger docs.. rest LGTM. Also this PR has some conflicts 😑

http/src/main/resources/swagger/api-docs.yaml
akarukappadath: Makes sense. To clarify, this is only in the swagger? Or should that be a type of its own in `DiskRoutes.scala`?

http/src/main/resources/swagger/api-docs.yaml
akarukappadath: We don't have errors for PDs yet, I think we were holding off until we have a better idea of what kinds of errors can come up. @rtitle would that be after some of the back Leo work?

http/src/main/resources/swagger/api-docs.yaml
Qi77Qi: I think both..We should have dedicated response type in code as well

http/src/main/resources/swagger/api-docs.yaml
rtitle: Yeah we haven't added any disk errors yet… let's hold off till another PR, I don't want to keep increasing scope of this one.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/DiskRoutes.scala
```diff
@@ -0,0 +1,316 @@
...
+    Directive { inner =>
+      validateDiskName(diskNameString) match {
+        case Left(e)  => failWith(e)
+        case Right(c) => inner(Tuple1(c))
+      }
+    }
+
+}
+
+final case class ListPersistentDiskResponse(id: DiskId,
```
akarukappadath, line 276: As of the moment, these are the same as the `PersistentDisk` type. Once we include `errors` in PDs, I think that will go in the `get` response and not the `list` response. Other potential fields we could remove from the `list` response are `googleId`, `samResourceId`, and `labels` - let me know what you think

Qi77Qi: sounds good to me...are you removing the fields you mentioned? or you're just saying that can be done in the future? (I'd prefer just removing them if no particular reason to keep..either this PR or follow up PRs)

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -1778,6 +2079,108 @@ components:
...
           description: Whether there is a patch in progress on the runtime. Is used to indicate updates that require status transitions.
+    ListPersistentDiskResponse:
+      description: ""
+      required:
+        - id
+        - googleProject
+        - zone
+        - name
+        - googleId
+        - samResourceId
```
Qi77Qi: I thought you were going to remove this `samResourceId`?

Qi77Qi (approved):

http/src/main/resources/swagger/api-docs.yaml
akarukappadath: It turns out we use it ourselves to check whether or not the disk is visible for the user before `list`ing, so I think it makes sense to keep in there

PR #1414, opened 2020-05-13
[No ticket] leo should not return stack traces to user

jdcanas:
If we don't specify a statuscode for LeoException it defaults to 500 which results in users seeing a stacktrace 

https://github.com/DataBiosphere/leonardo/blob/develop/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeoException.scala

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/HttpRoutes.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/HttpRoutes.scala
@@ -62,7 +62,7 @@ class HttpRoutes(
         complete(StatusCodes.InternalServerError -> report)
       case e: Throwable =>
         //NOTE: this needs SprayJsonSupport._, ErrorReportJsonSupport._, and errorReportSource all imported to work
-        complete(StatusCodes.InternalServerError -> ErrorReport(e))
+        complete(StatusCodes.InternalServerError -> ErrorReport(e.getMessage, Some(StatusCodes.InternalServerError), ErrorReport.causes(e), Seq(), Some(e.getClass)))
     }
   }
 
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/HttpRoutes.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/HttpRoutes.scala
@@ -62,7 +62,7 @@ class HttpRoutes(
         complete(StatusCodes.InternalServerError -> report)
       case e: Throwable =>
         //NOTE: this needs SprayJsonSupport._, ErrorReportJsonSupport._, and errorReportSource all imported to work
-        complete(StatusCodes.InternalServerError -> ErrorReport(e.getMessage, Some(StatusCodes.InternalServerError), ErrorReport.causes(e), Seq(), Some(e.getClass)))
+        complete(StatusCodes.InternalServerError -> ErrorReport(e.getMessage, Some(StatusCodes.InternalServerError), Seq(), Seq(), Some(e.getClass)))
     }
   }
 
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpSamDAO.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpSamDAO.scala
@@ -6,6 +6,8 @@ import java.util.UUID
 import java.util.concurrent.TimeUnit
 
 import _root_.io.circe.{Decoder, Json, KeyDecoder}
+import akka.http.scaladsl.model.{StatusCode, StatusCodes}
+import akka.http.scaladsl.model.StatusCode._
 import ca.mrvisser.sealerate
 import cats.effect.implicits._
 import cats.effect.{Blocker, ContextShift, Effect, Resource}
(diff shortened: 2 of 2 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpSamDAO.scala
```diff
@@ -423,5 +425,5 @@ object ResourceTypeName {
 }
 
 final case object NotFoundException extends NoStackTrace
-final case class AuthProviderException(traceId: TraceId, msg: String)
+final case class AuthProviderException(traceId: TraceId, msg: String, statusCode: StatusCode)
```
rtitle: Needs something like 
```
final case class AuthProviderException(traceId: TraceId, msg: String, override val statusCode: StatusCode)
  extends LeoException(message = s"${traceId} | AuthProvider error: $msg")
```
I guess

rtitle (approved):
Looks fine to me (barring compilation error). 401 seems reasonable if we can't get the pet token.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpSamDAO.scala
jdcanas: yeah I didn't bother compiling, but I guess thats flying too close to the sun

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpSamDAO.scala
```diff
@@ -423,5 +425,5 @@ object ResourceTypeName {
 }
 
 final case object NotFoundException extends NoStackTrace
-final case class AuthProviderException(traceId: TraceId, msg: String)
-    extends LeoException(message = s"${traceId} | AuthProvider error: $msg")
+final case class AuthProviderException(traceId: TraceId, msg: String, code: StatusCode)
```
Qi77Qi, line 428: are there cases where status code is not `StatusCodes.Unauthorized` for `AuthProviderException`? also I'm not sure which change corresponds to the title 🤔

jdcanas: This is for a user liason issue. 
Sometimes SAM can return other codes, for example, for the issue I'm investigating now SAM is returning a 400. I don't want to eat the status from SAM in that case.

rtitle: Hm, why _are_ we returning the stack trace currently? Is there some logic where we return the stack trace for 500 but not other status codes?

Qi77Qi (approved):

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpSamDAO.scala
jdcanas: Yeah, that is here, but I'm also a little confused because it should match `case leoException` which does not pass the stacktrace along.  https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/HttpRoutes.scala#L65

The logic is a little fuzzy, but since that's the only case in the handler that provides a stacktrace, I think its the culprit

Qi77Qi: you can also just extend `NoStackTrace`, that way stacktrace is not calculated in the first place, which is more desirable if we don't care about stacktrace here.

rtitle: Yeah makes me wonder if something more subtle is going on. Like maybe we _do_ throw `AuthProviderException` here but something downstream throws a `Throwable` wrapping the original error, causing a stack trace to be returned. Just speculating though

jdcanas: yeah... I just modified the generic `Throwable` case in `HttpRoutes` to not pass the stacktrace into the `ErrorReport`. 
This should work for all exceptions and accomplish the same thing as @Qi77Qi suggestion, but for all exceptions by default

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/HttpRoutes.scala
```diff
@@ -62,7 +62,7 @@ class HttpRoutes(
         complete(StatusCodes.InternalServerError -> report)
       case e: Throwable =>
         //NOTE: this needs SprayJsonSupport._, ErrorReportJsonSupport._, and errorReportSource all imported to work
-        complete(StatusCodes.InternalServerError -> ErrorReport(e))
+        complete(StatusCodes.InternalServerError -> ErrorReport(e.getMessage, Some(StatusCodes.InternalServerError), ErrorReport.causes(e), Seq(), Some(e.getClass)))
```
jdcanas: this ErrorReport required some digging since its overloaded a bunch, but this should correspond to the following signature
` def apply(message: String, statusCode: Option[StatusCode], causes: Seq[ErrorReport], stackTrace: Seq[StackTraceElement], exceptionClass: Option[Class[_]])(implicit source: ErrorReportSource): ErrorReport =
    ErrorReport(source.source, message, statusCode, causes, stackTrace, exceptionClass)`

this is what `LeoException` uses via 
` ErrorReport(Option(getMessage).getOrElse(""), Some(statusCode), Seq(), Seq(), Some(this.getClass))`

jdcanas:
jenkins retest

PR #1416, opened 2020-05-19
[IA-1901] warm up R in jupyter startup script

rtitle:
https://broadworkbench.atlassian.net/browse/IA-1901

Two fixes here:
1. Adopt https://github.com/broadinstitute/workbench-libs/pull/307 to fix a NPE that has been happening for a while (an AoU user reported seeing it though it seems to happen for non-AoU as well)
2. Try to work around the issue where the R kernel is hanging on VM startup. It seems to _not_ hang if you run a dummy R command before the Jupyter kernel starts up. Not sure exactly why

---

In all cases:

### DIFF
--- a/http/src/main/resources/init-resources/startup.sh
+++ b/http/src/main/resources/init-resources/startup.sh
@@ -124,7 +124,7 @@ if [ ! -z "$JUPYTER_DOCKER_IMAGE" ] ; then
     # See IA-1901: Jupyter UI stalls indefinitely on initial R kernel connection after cluster create/resume
     # The intent of this is to "warm up" R at VM creation time to hopefully prevent issues when the Jupyter
     # kernel tries to connect to it.
-    docker exec ${JUPYTER_SERVER_NAME} /bin/bash -c "R -e '1+1'" || true
+    docker exec $JUPYTER_SERVER_NAME /bin/bash -c "R -e '1+1'" || true
 
     docker exec -d $JUPYTER_SERVER_NAME /bin/bash -c "export WELDER_ENABLED=$WELDER_ENABLED && export NOTEBOOKS_DIR=$NOTEBOOKS_DIR && (/etc/jupyter/scripts/run-jupyter.sh $NOTEBOOKS_DIR || /usr/local/bin/jupyter notebook)"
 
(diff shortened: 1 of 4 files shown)


http/src/main/resources/init-resources/startup.sh
```diff
@@ -121,6 +121,11 @@ if [ ! -z "$JUPYTER_DOCKER_IMAGE" ] ; then
     # update container MEM_LIMIT to reflect VM's MEM_LIMIT
     docker update $JUPYTER_SERVER_NAME --memory $MEM_LIMIT
 
+    # See IA-1901: Jupyter UI stalls indefinitely on initial R kernel connection after cluster create/resume
```
Qi77Qi, line 124: I wonder if we should revert the previous PR that changed the proxy service if that didn't really fix things 🤔

Qi77Qi (approved):

http/src/main/resources/init-resources/startup.sh
rtitle: Yeah seems reasonable.. I reverted it

rtitle:
fiab-start failing with quay.io issues.. jenkins retest

Qi77Qi:
just occurred to me...will this mean `terra-jupyter-base`, `terra-jupyter-python` will fail to start? do they have `R` installed?

rtitle:
That's why I included `|| true` -- the failing command shouldn't fail the script. Will double check in a fiab

http/src/main/resources/init-resources/startup.sh
Qi77Qi: oh cool..did't see `||`

project/Dependencies.scala
```diff
@@ -18,7 +18,8 @@ object Dependencies {
 
   val workbenchUtilV = "0.5-4c7acd5"
   val workbenchModelV = "0.13-31cacc4"
-  val workbenchGoogleV = "0.21-96ad43c"
+  // TODO: revert to non-SNAP and remove this comment once https://github.com/broadinstitute/workbench-libs/pull/307 is merged
```
Qi77Qi: think https://github.com/broadinstitute/workbench-libs/pull/307 can be merged?

rtitle: Yeah will do.. was just waiting for auto tests to pass

rtitle:
More quay.io issues in fiab-start. jenkins retest

rtitle:
fiab-start is broken but just going to merge anyway.. tests passed prior to https://github.com/DataBiosphere/leonardo/pull/1416/commits/11e359d3791ce639fb0d6320ca0f8d79eae13bb7

PR #1417, opened 2020-05-21
[no ticket] Apply R fix to creation scripts too

rtitle:
AoU is still seeing the R issue. I think they are seeing it for freshly-created clusters. I had added the fix to `startup.sh` which is used for cluster _resumes_, but neglected to add it to the init scripts used for cluster _creation_. I reproduced the same behavior in Terra.

---

In all cases:

### DIFF
--- a/http/src/main/resources/init-resources/gce-init.sh
+++ b/http/src/main/resources/init-resources/gce-init.sh
@@ -422,6 +422,11 @@ END
         done
       fi
 
+      # See IA-1901: Jupyter UI stalls indefinitely on initial R kernel connection after cluster create/resume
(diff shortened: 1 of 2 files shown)


http/src/main/resources/init-resources/gce-init.sh
```diff
@@ -430,6 +435,9 @@ END
     # Do this asynchronously so it doesn't hold up cluster creation
     log 'Pruning docker images...'
     docker image prune -a -f &
+    
+# TODO (RT): I'm pretty sure this block is never used because we have a dedicated startup.sh script
+# used for starting runtimes. We could confirm this and remove this block.
```
rtitle, line 440: Didn't want to delete this now but it seems like dead code

Qi77Qi (approved):

PR #1418, opened 2020-05-21
IA-1757 Back Leo: add disk PubSub messaging and algebra

gpcarr:
PR to add message handler methods to LeoPubsubSubscriber for `CreateDisk`, `DeleteDisk` and `UpdateDisk` 

---

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/DiskRoutes.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/DiskRoutes.scala
@@ -191,10 +191,8 @@ object DiskRoutes {
   implicit val updateDiskRequestDecoder: Decoder[UpdateDiskRequest] = Decoder.instance { x =>
     for {
       l <- x.downField("labels").as[LabelMap]
-      us <- x.downField("updateSize").as[Option[DiskSize]]
-      ud <- x.downField("updateDiskType").as[Option[DiskType]]
-      ub <- x.downField("updateBlockSize").as[Option[BlockSize]]
-    } yield UpdateDiskRequest(l, us, ud, ub)
+      us <- x.downField("size").as[DiskSize]
+    } yield UpdateDiskRequest(l, us)
   }
 
   implicit val getDiskResponseEncoder: Encoder[GetPersistentDiskResponse] = Encoder.forProduct12(
@@ -303,7 +301,4 @@ final case class CreateDiskRequest(labels: LabelMap,
                                    diskType: Option[DiskType],
                                    blockSize: Option[BlockSize])
 
-final case class UpdateDiskRequest(labels: LabelMap,
-                                   size: Option[DiskSize],
-                                   diskType: Option[DiskType],
-                                   blockSize: Option[BlockSize])
+final case class UpdateDiskRequest(labels: LabelMap, size: DiskSize)
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/PersistentDiskComponent.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/PersistentDiskComponent.scala
@@ -112,6 +112,12 @@ object persistentDiskQuery extends TableQuery(new PersistentDiskTable(_)) {
       .map(d => (d.status, d.destroyedDate, d.dateAccessed))
       .update((DiskStatus.Deleted, destroyedDate, destroyedDate))
 
+  def updateGoogleId(id: DiskId, googleId: GoogleId, dateAccessed: Instant) =
+    findByIdQuery(id).map(d => (d.googleId, d.dateAccessed)).update((Some(googleId), dateAccessed))
+
+  def updateSize(id: DiskId, newSize: DiskSize, dateAccessed: Instant) =
+    findByIdQuery(id).map(d => (d.size, d.dateAccessed)).update((newSize, dateAccessed))
+
   // TODO add other queries as needed
 
   private[db] def marshalPersistentDisk(disk: PersistentDisk): PersistentDiskRecord =
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
@@ -8,10 +8,11 @@ import cats.effect.implicits._
 import cats.effect.{ConcurrentEffect, ContextShift, Timer}
 import cats.implicits._
 import cats.mtl.ApplicativeAsk
-import fs2.Stream
 import fs2.concurrent.InspectableQueue
-import org.broadinstitute.dsde.workbench.google2.{Event, GoogleSubscriber, MachineTypeName}
 import org.broadinstitute.dsde.workbench.leonardo.AsyncTaskProcessor.Task
+import com.google.cloud.compute.v1.Disk
+import fs2.Stream
+import org.broadinstitute.dsde.workbench.google2.{Event, GoogleDiskService, GoogleSubscriber, MachineTypeName}
 import org.broadinstitute.dsde.workbench.leonardo.db._
 import org.broadinstitute.dsde.workbench.leonardo.http.{cloudServiceSyntax, _}
 import org.broadinstitute.dsde.workbench.leonardo.model.LeoException
@@ -27,7 +28,8 @@ import scala.util.control.NoStackTrace
 class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift](
   config: LeoPubsubMessageSubscriberConfig,
   subscriber: GoogleSubscriber[F, LeoPubsubMessage],
-  asyncTasks: InspectableQueue[F, Task[F]]
+  asyncTasks: InspectableQueue[F, Task[F]],
+  googleDiskService: GoogleDiskService[F]
 )(implicit executionContext: ExecutionContext,
   F: ConcurrentEffect[F],
   logger: StructuredLogger[F],
@@ -47,7 +49,12 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift](
         handleStartRuntimeMessage(msg, now)
       case msg: UpdateRuntimeMessage =>
         handleUpdateRuntimeMessage(msg, now)
-      case _ => throw new NotImplementedError("disk events are not implemented yet") //TODO: put disk messages here
+      case msg: CreateDiskMessage =>
+        handleCreateDiskMessage(msg, now)
+      case msg: DeleteDiskMessage =>
+        handleDeleteDiskMessage(msg, now)
+      case msg: UpdateDiskMessage =>
+        handleUpdateDiskMessage(msg, now)
     }
 
   private[monitor] def messageHandler(event: Event[LeoPubsubMessage]): F[Unit] = {
@@ -328,6 +335,62 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift](
       _ <- runtimeConfig.cloudService.process(runtime.id, RuntimeStatus.Starting).compile.drain
     } yield ()
 
+  private[monitor] def handleCreateDiskMessage(msg: CreateDiskMessage, now: Instant)(
+    implicit traceId: ApplicativeAsk[F, AppContext]
+  ): F[Unit] = {
+    val createDisk = for {
+      diskResult <- googleDiskService.createDisk(
+        msg.googleProject,
+        msg.zone,
+        Disk
+          .newBuilder()
+          .setName(msg.name.value)
+          .setSizeGb(msg.size.asString)
+          .setZone(msg.zone.value)
+          .setType(msg.diskType.googleString)
+          .setPhysicalBlockSizeBytes(msg.blockSize.bytes.toString)
+          .build()
+      )
+      _ <- persistentDiskQuery.updateGoogleId(msg.diskId, GoogleId(diskResult.getTargetId), now).transaction[F]
+    } yield ()
+    createDisk.handleErrorWith {
+      case e =>
+        for {
+          _ <- logger.error(e)(s"Failed to create disk ${msg.name.value} in Google project ${msg.googleProject.value}")
+          _ <- persistentDiskQuery.updateStatus(msg.diskId, DiskStatus.Failed, now).transaction[F]
+        } yield ()
+    }
+  }
+
+  private[monitor] def handleDeleteDiskMessage(msg: DeleteDiskMessage, now: Instant)(
+    implicit traceId: ApplicativeAsk[F, AppContext]
+  ): F[Unit] =
+    for {
+      diskOpt <- persistentDiskQuery.getById(msg.diskId).transaction
+      disk <- diskOpt.fold(
+        F.raiseError[PersistentDisk](PubsubHandleMessageError.DiskNotFound(msg.diskId, msg))
+      )(F.pure)
+      _ <- if (disk.status != DiskStatus.Deleting)
+        F.raiseError[Unit](
+          PubsubHandleMessageError.DiskInvalidState(msg.diskId, disk.projectNameString, disk, msg)
+        )
+      else F.unit
+      _ <- googleDiskService.deleteDisk(disk.googleProject, disk.zone, disk.name)
+    } yield ()
+
+  private[monitor] def handleUpdateDiskMessage(msg: UpdateDiskMessage, now: Instant)(
+    implicit traceId: ApplicativeAsk[F, AppContext]
+  ): F[Unit] =
+    for {
+      diskOpt <- persistentDiskQuery.getById(msg.diskId).transaction
+      disk <- diskOpt.fold(
+        F.raiseError[PersistentDisk](PubsubHandleMessageError.DiskNotFound(msg.diskId, msg))
+      )(F.pure)
+      _ <- googleDiskService.resizeDisk(disk.googleProject, disk.zone, disk.name, msg.newSize.gb)
+      // TODO: needs to be updated after monitoring is added
+      _ <- persistentDiskQuery.updateSize(msg.diskId, msg.newSize, now).transaction[F]
+    } yield ()
+
   private def logError(projectAndName: String): Throwable => F[Unit] =
     t => logger.error(t)(s"Fail to monitor ${projectAndName}")
 }
@@ -341,6 +404,7 @@ object PubsubHandleMessageError {
       s"Unable to process transition finished message ${message} for cluster ${clusterId} because it was not found in the database"
     val isRetryable: Boolean = false
   }
+
   final case class ClusterNotStopped(clusterId: Long,
                                      projectName: String,
                                      clusterStatus: RuntimeStatus,
@@ -350,6 +414,7 @@ object PubsubHandleMessageError {
       s"Unable to process message ${message} for cluster ${clusterId}/${projectName} in status ${clusterStatus.toString}, when the monitor signalled it stopped as it is not stopped."
     val isRetryable: Boolean = false
   }
+
   final case class ClusterInvalidState(clusterId: Long,
                                        projectName: String,
                                        cluster: Runtime,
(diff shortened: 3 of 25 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -49,6 +51,8 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift](
         handleStartRuntimeMessage(msg, now)
       case msg: UpdateRuntimeMessage =>
         handleUpdateRuntimeMessage(msg, now)
+      case msg: CreateDiskMessage =>
+        handleCreateDiskMessage(msg, now)
       case _ => throw new NotImplementedError("disk events are not implemented yet") //TODO: put disk messages here
```
Qi77Qi: remove `case _` and add the `DeleteDiskMessage` and `UpdateDiskMessage`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/leoPubsubMessageSubscriberModels.scala
```diff
@@ -167,11 +167,7 @@ object LeoPubsubMessage {
     val messageType: LeoPubsubMessageType = LeoPubsubMessageType.UpdateRuntime
   }
 
-  final case class UpdateDiskMessage(diskId: DiskId,
-                                     newSize: Option[DiskSize],
-                                     newDiskType: Option[DiskType],
-                                     newBlockSize: Option[BlockSize],
-                                     traceId: Option[TraceId])
+  final case class UpdateDiskMessage(diskId: DiskId, newSize: Option[DiskSize], traceId: Option[TraceId])
```
Qi77Qi: if `newSize` is None, is there still point of having this message?

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/dao/google/MockGoogleDiskService.scala
```diff
@@ -0,0 +1,29 @@
...
+
+import cats.effect.IO
+import cats.mtl.ApplicativeAsk
+import com.google.cloud.compute.v1.{Disk, Operation}
+import fs2.Stream
+import org.broadinstitute.dsde.workbench.google2.{DiskName, GoogleDiskService, ZoneName}
+import org.broadinstitute.dsde.workbench.model.TraceId
+import org.broadinstitute.dsde.workbench.model.google.GoogleProject
+
+class MockGoogleDiskService extends GoogleDiskService[IO] {
```
Qi77Qi, line 12: we should probably have added this in wb-libs similar to https://github.com/broadinstitute/workbench-libs/blob/develop/google2/src/test/scala/org/broadinstitute/dsde/workbench/google2/mock/FakeGoogleStorageInterpreter.scala

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/leoPubsubMessageSubscriberModels.scala
gpcarr: I guess not.. should we make `newSize` not be an Option?

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/dao/google/MockGoogleDiskService.scala
gpcarr: I can add a `TODO` similar to https://github.com/DataBiosphere/leonardo/blob/develop/http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/dao/google/MockGoogleComputeService.scala#L24 and we can create a tech debt ticket to move them both to wb-libs

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -415,6 +422,51 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift](
...
+
+  private[monitor] def handleDeleteDiskMessage(msg: DeleteDiskMessage, now: Instant)(
+    implicit traceId: ApplicativeAsk[F, AppContext]
+  ): F[Unit] =
+    for {
+      diskOpt <- persistentDiskQuery.getById(msg.diskId).transaction
+      disk <- diskOpt.fold(
+        F.raiseError[PersistentDisk](PubsubHandleMessageError.DiskNotFound(msg.diskId, msg))
+      )(F.pure)
+      // TODO: do I need to check disk status before deleting?
```
gpcarr: @akarukappadath @rtitle do I need to check disk status before deleting? and what about for updating disk? what statuses are we checking for?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -415,6 +422,51 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift](
...
+  private[monitor] def handleCreateDiskMessage(msg: CreateDiskMessage, now: Instant)(
+    implicit traceId: ApplicativeAsk[F, AppContext]
+  ): F[Unit] =
+    for {
+      _ <- googleDiskService.createDisk(
+        msg.googleProject,
+        msg.zone,
+        Disk
+          .newBuilder()
+          .setId(msg.diskId.toString)
```
rtitle: I don't think you need to set an id in the create request. This field is output only, meaning Google will generate its own ID.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -415,6 +422,51 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift](
...
+          .setCreationTimestamp(now.toString)
```
rtitle: Same here, I think `creationTimestamp` is output only

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -415,6 +422,51 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift](
...
+          .setName(msg.name.toString)
```
rtitle: Probably want to use `.asString` or `.value` for these case classes. `.toString` may not generate the format we are expecting.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -415,6 +422,51 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift](
       _ <- gceRuntimeMonitor.process(runtime.id).compile.drain
     } yield ()
 
+  private[monitor] def handleCreateDiskMessage(msg: CreateDiskMessage, now: Instant)(
+    implicit traceId: ApplicativeAsk[F, AppContext]
+  ): F[Unit] =
+    for {
```
rtitle: I'm picturing a flow like the following here (psuedocode):
```
val createDisk = for {
  googleResp <- googleService.createDisk(…)
  _ <- // update persistent disk in the DB with async fields like `googleId`
  _ <- // monitor disk creation
} yield ()

createDisk.handleErrorWith { 
  // if an error occurred, log and set the disk status to `Error`
}
```

See `handleCreateRuntimeMessage` as a model. I think monitoring stuff is tracked as a separate ticket (and will change once https://github.com/DataBiosphere/leonardo/pull/1395/ is merged).

rtitle: I'm also wondering if we want to make Google calls inline here in the pubsub subscriber, or create a separate algebra/interpreter similar to `RuntimeAlgebra` and `Gce/DataprocInterpreter`. I'm not sure I see the need for a separate algebra yet, so maybe let's keep things simple for now and just make the calls here in `LeoPubsubMessageSubscriber`.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
rtitle: I think we expect the disk to be in `Deleting` state by the time the subscriber receives the `delete` message. So maybe check the record for `Deleting` status, and raise an error otherwise (this is similar to `handleDeleteRuntimeMessage`).

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -415,6 +422,51 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift](
...
+          .build()
+      )
+    } yield ()
+
+  private[monitor] def handleDeleteDiskMessage(msg: DeleteDiskMessage, now: Instant)(
+    implicit traceId: ApplicativeAsk[F, AppContext]
+  ): F[Unit] =
+    for {
+      diskOpt <- persistentDiskQuery.getById(msg.diskId).transaction
+      disk <- diskOpt.fold(
```
rtitle, line 370: Just FYI, another way to write this more concisely is:
```
F.fromEither(diskOpt.toRight(PubsubHandleMessageError.DiskNotFound(msg.diskId, msg)))
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/leoPubsubMessageSubscriberModels.scala
Qi77Qi: yea...make it not optional so that if we're sending this event, it's an event that we do care.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -415,6 +422,51 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift](
...
+
+  private[monitor] def handleUpdateDiskMessage(msg: UpdateDiskMessage, now: Instant)(
+    implicit traceId: ApplicativeAsk[F, AppContext]
+  ): F[Unit] =
+    for {
+      diskOpt <- persistentDiskQuery.getById(msg.diskId).transaction
+      disk <- diskOpt.fold(
+        F.raiseError[PersistentDisk](PubsubHandleMessageError.DiskNotFound(msg.diskId, msg))
+      )(F.pure)
+      _ <- if (msg.newSize.isDefined) {
```
rtitle: Maybe 
```
_ <- msg.newSize.traverse_ { newSize =>
  googleDiskService.resizeDisk(disk.googleProject, disk.zone, disk.name, newSize.gb)
}
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -439,6 +492,7 @@ object PubsubHandleMessageError {
       s"Unable to process message ${message} for cluster ${clusterId}/${projectName} in status ${clusterStatus.toString}, when the monitor signalled it stopped as it is not stopped."
     val isRetryable: Boolean = false
   }
+
   final case class ClusterInvalidState(clusterId: Long,
```
rtitle, line 418: Maybe we make a `DiskInvalidState` exception type for the status exceptions above

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/leoPubsubMessageSubscriberModels.scala
rtitle: That's a good point, maybe we should make it non-optional.
If there are multiple potential fields to update, then we would make them optional.

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriberSpec.scala
```diff
@@ -20,7 +20,7 @@ import org.broadinstitute.dsde.workbench.leonardo.CommonTestData._
 import org.broadinstitute.dsde.workbench.leonardo.RuntimeImageType.VM
 import org.broadinstitute.dsde.workbench.leonardo.config.Config
 import org.broadinstitute.dsde.workbench.leonardo.dao.WelderDAO
-import org.broadinstitute.dsde.workbench.leonardo.dao.google.MockGoogleComputeService
+import org.broadinstitute.dsde.workbench.leonardo.dao.google.{MockGoogleComputeService, MockGoogleDiskService}
```
rtitle, line 22: Probably want to add some test cases in this spec for the new `handle` methods in `LeoPubsubMessageSubscriber`

rtitle:
Left some suggestions above, this is a good first pass though. It really feels like we're getting closer with the PD work.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
gpcarr: good to know, will update!

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/leoPubsubMessageSubscriberModels.scala
gpcarr: should I also make it non-optional in `UpdateDiskRequest`? https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/DiskRoutes.scala#L305 (I removed `diskType` and `blockSize` from here as well)

Qi77Qi: it needs to be optional in request because it's possible that they just want to update `diskType` or `blockSize`

gpcarr: Oh so I shouldn't remove `diskType` or `blockSize` from the request? the reason i removed them is because Google doesn't support updating `diskType` or `blockSize`

gpcarr: would we want to allow users to update `diskType` snd `blockSize` just in our DB?

Qi77Qi: hmmm if that's the case..yea probably should just remove them from the request and make size non optional...cc @rtitle for confirmation

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/DiskRoutes.scala
```diff
@@ -190,10 +190,8 @@ object DiskRoutes {
   implicit val updateDiskRequestDecoder: Decoder[UpdateDiskRequest] = Decoder.instance { x =>
     for {
       l <- x.downField("labels").as[LabelMap]
-      us <- x.downField("updateSize").as[Option[DiskSize]]
-      ud <- x.downField("updateDiskType").as[Option[DiskType]]
-      ub <- x.downField("updateBlockSize").as[Option[BlockSize]]
-    } yield UpdateDiskRequest(l, us, ud, ub)
+      us <- x.downField("updateSize").as[DiskSize]
```
Qi77Qi: we should just call this `size`...being an `UpdateDiskRequest` already indicates this is for updating the disk size. terra UI preferred the field matches other requests so that they don't have to adjust the field name differently depending on the call they're making

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -415,6 +422,49 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift](
       _ <- gceRuntimeMonitor.process(runtime.id).compile.drain
     } yield ()
 
+  private[monitor] def handleCreateDiskMessage(msg: CreateDiskMessage, now: Instant)(
+    implicit traceId: ApplicativeAsk[F, AppContext]
+  ): F[Unit] =
+    for {
+      diskResult <- googleDiskService.createDisk(
```
Qi77Qi, line 342: use `_` since doesn't seem like you care about `diskResult` here

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/DiskRoutes.scala
gpcarr: will update :)

gpcarr:
jenkins retest

gpcarr:
jenkins retest

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/PersistentDiskComponent.scala
```diff
@@ -108,6 +108,12 @@ object persistentDiskQuery extends TableQuery(new PersistentDiskTable(_)) {
       .map(d => (d.status, d.destroyedDate, d.dateAccessed))
       .update((DiskStatus.Deleted, destroyedDate, destroyedDate))
 
+  def updateGoogleId(id: DiskId, googleId: GoogleId, dateAccessed: Instant) =
```
Qi77Qi, line 115: just curious, when does `googleId` change?

Qi77Qi (approved):
lgtm 👍

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/PersistentDiskComponent.scala
gpcarr: We create the Disk in our db, then send the api call to google which creates the `googleId` so we have to update the `googleId` in our db after we successfully create the disk in google

gpcarr:
@Qi77Qi @rtitle looks like the jenkins tests failed because the `fiab-start` job failed but the `leo` build was successful (https://fc-jenkins.dsp-techops.broadinstitute.org/job/swatomation-pipeline/34686/). seems like I should wait until this job is fixed before merging according to a thread in #dsp-devops-champions

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
gpcarr: this is no longer an optional field so no need to check anymore if it exists :)

akarukappadath (approved):
lgtm!

gpcarr:
jenkins retest

gpcarr:
jenkins retest

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -328,6 +335,60 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift](
...
 
+  private[monitor] def handleCreateDiskMessage(msg: CreateDiskMessage, now: Instant)(
+    implicit traceId: ApplicativeAsk[F, AppContext]
+  ): F[Unit] = {
+    val createDisk = for {
+      diskResult <- googleDiskService.createDisk(
+        msg.googleProject,
+        msg.zone,
+        Disk
+          .newBuilder()
```
rtitle, line 346: For `createDisk`, don't we have `blockSize`, `diskType` fields too?

rtitle (approved):
Looks great!

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
gpcarr: oh we do, the corresponding fields in google would be `type` and `physicalBlockSizeBytes` I believe. I need to add this to wb-libs as well (it'll be a different `diskClient.insertDisk` method we need to use), should I create a new ticket to add this to wb-libs and then add `blockSize` and `diskType` in the leo create disk call? Alternatively, i can hold off on merging this PR, make the changes in wb-libs and then add the leo changes to this PR.

gpcarr: Oh wait, nevermind I think I can just make the changes in leo since we pass a `Disk` object to the google API call

rtitle: Yup, latest change looks good 👍

PR #1424, opened 2020-05-22
IA-1907 consolidate sam calls

xuscrbw:
We were looking for a way to consolidate the Sam calls, which checked for Notebook Action permissions. We built out a new method with would call the resources API to return of list of permissions that a particular user had for a cluster. 

Using that list, we would check to see if a particular permission was within that list. This way, we did not have to call Sam more than once per request.

This is still a WIP.
---

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/sam/SamAuthProvider.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/sam/SamAuthProvider.scala
@@ -152,10 +152,30 @@ class SamAuthProvider[F[_]: Effect: Logger](samDao: SamDAO[F],
     checkPersistentDiskPermissionWithProjectFallback(samResource, authorization, action, googleProject)
   }
 
-  def getNotebookClusterActions(internalId: RuntimeInternalId,
-                                userInfo: UserInfo)(implicit ev: ApplicativeAsk[F, TraceId]): F[List[String]] = {
+  private def convertNotebookClusterActionsToString(action: String): NotebookClusterAction = {
+    action match {
+      case "connect" => NotebookClusterAction.ConnectToCluster
+      case "modify" => NotebookClusterAction.ModifyCluster
+      case "sync" => NotebookClusterAction.SyncDataToCluster
+      case "stop_start" =>  NotebookClusterAction.StopStartCluster
+      case "delete" => NotebookClusterAction.DeleteCluster
+    }
+  }
+
+  def getNotebookClusterActions(internalId: RuntimeInternalId, userInfo: UserInfo)(
+    implicit ev: ApplicativeAsk[F, TraceId]
+  ): F[List[NotebookClusterAction]] = {
+
     val authorization = Authorization(Credentials.Token(AuthScheme.Bearer, userInfo.accessToken.token))
-    samDao.getListOfResourcePermissions(internalId.asString, ResourceTypeName.NotebookCluster, authorization)
+
+    for{
+      listOfPermissions <- samDao
+                              .getListOfResourcePermissions(internalId.asString, ResourceTypeName.NotebookCluster, authorization)
+
+
+    } yield {
+      listOfPermissions.map(x => convertNotebookClusterActionsToString(x))
+    }
   }
 
   private def checkRuntimePermissionWithProjectFallback(
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/sam/SamAuthProvider.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/sam/SamAuthProvider.scala
@@ -66,7 +66,7 @@ class SamAuthProvider[F[_]: Effect: Logger](samDao: SamDAO[F],
         )
       )
 
-  private def getNotebookActionString(action: LeoAuthAction): Either[Throwable, String] =
+  private def getNotebookActionString(action: RuntimeAction): Either[Throwable, String] =
     runtimeActionMap
       .get(action)
       .toRight(
@@ -75,7 +75,7 @@ class SamAuthProvider[F[_]: Effect: Logger](samDao: SamDAO[F],
         )
       )
 
-  private def getPersistentDiskActionString(action: LeoAuthAction): Either[Throwable, String] =
+  private def getPersistentDiskActionString(action: PersistentDiskAction): Either[Throwable, String] =
     persistentDiskActionMap
       .get(action)
       .toRight(
@@ -94,7 +94,7 @@ class SamAuthProvider[F[_]: Effect: Logger](samDao: SamDAO[F],
     DeletePersistentDisk -> "delete_persistent_disk"
   )
 
-  private val runtimeActionMap: Map[LeoAuthAction, String] = Map(
+  private val runtimeActionMap: Map[RuntimeAction, String] = Map(
     GetRuntimeStatus -> "status",
     ConnectToRuntime -> "connect",
     SyncDataToRuntime -> "sync",
@@ -103,7 +103,7 @@ class SamAuthProvider[F[_]: Effect: Logger](samDao: SamDAO[F],
     StopStartRuntime -> "stop_start"
   )
 
-  private val persistentDiskActionMap: Map[LeoAuthAction, String] = Map(
+  private val persistentDiskActionMap: Map[PersistentDiskAction, String] = Map(
     ReadPersistentDisk -> "read",
     AttachPersistentDisk -> "attach",
     ModifyPersistentDisk -> "modify",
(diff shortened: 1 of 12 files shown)


xuscrbw:
jenkins retest

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpSamDAOSpec.scala
```diff
@@ -24,7 +24,7 @@ import org.http4s.client.middleware.{Retry, RetryPolicy}
 
 import scala.util.control.NoStackTrace
 
-class HttpSamDAOSpec extends FlatSpec with Matchers with BeforeAndAfterAll {
+class   HttpSamDAOSpec extends FlatSpec with Matchers with BeforeAndAfterAll {
```
Qi77Qi, line 27: can you run `scalafmt`

Qi77Qi (approved):
👍 lgtm..one minor comment

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeoAuthProvider.scala
```diff
@@ -65,6 +65,16 @@ abstract class LeoAuthProvider[F[_]] {
...
 
+  /**
+   *
+   *
+   * @param internalId
+   * @return returns a list of Cluster Actions that the user has access to
+   */
+  def getNotebookClusterActions(internalId: RuntimeInternalId, userInfo: UserInfo)(
+    implicit ev: ApplicativeAsk[F, TraceId]
+  ): F[List[String]]
```
rtitle: Perhaps this should return a `List[NotebookClusterAction]` instead of a `List[String]`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/IamProxyAuthProvider.scala
```diff
@@ -159,4 +159,14 @@ class IamProxyAuthProvider(config: Config, saProvider: ServiceAccountProvider[Fu
     creatorEmail: WorkbenchEmail,
     googleProject: GoogleProject
   )(implicit ev: ApplicativeAsk[Future, TraceId]): Future[Unit] = Future.unit
+
+  /**
+   *
+   *
+   * @param internalId
```
rtitle: I think @akarukappadath removed this class in her PR.. so this is fine, but you'll probably have to rebase.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/WhitelistAuthProvider.scala
```diff
@@ -87,4 +87,14 @@ class WhitelistAuthProvider(config: Config, saProvider: ServiceAccountProvider[I
...
+
+  /**
+   *
+   *
+   * @param internalId
+   * @return returns a list of Cluster Actions that the user has access to
+   */
+  override def getNotebookClusterActions(internalId: RuntimeInternalId, userInfo: UserInfo)(
+    implicit ev: ApplicativeAsk[IO, TraceId]
+  ): IO[List[String]] = IO.pure(List.empty)
```
rtitle: Instead of stubbing this out we should probably implement this the right way… something like
```
if (checkWhitelist(userInfo)) NotebookClusterAction.allActions else List.empty
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpSamDAO.scala
```diff
@@ -80,6 +80,17 @@ class HttpSamDAO[F[_]: Effect](httpClient: Client[F], config: HttpSamDaoConfig,
       )(onError)
     } yield res
 
+  def getListOfResourcePermissions(resourceId: String, resourceTypeName: ResourceTypeName, authHeader: Authorization)(
```
rtitle: I think this will also conflict with @akarukappadath's PR as she's making `SamDAO` methods take a `SamResource` instead of a `resourceId: String, resourceTypeName: ResourceTypeName`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeServiceInterp.scala
```diff
@@ -185,23 +185,21 @@ class RuntimeServiceInterp[F[_]: Parallel](config: RuntimeServiceConfig,
...
       // GetClusterStatus permission. We return 403 if the user can view the runtime but can't perform some other action.
-      hasPermission <- authProvider.hasNotebookClusterPermission(runtime.internalId,
-                                                                 userInfo,
-                                                                 GetClusterStatus,
-                                                                 googleProject,
-                                                                 runtimeName)
-      _ <- if (hasPermission) F.unit else F.raiseError[Unit](RuntimeNotFoundException(googleProject, runtimeName))
+
+      listOfPermissions <- authProvider.getNotebookClusterActions(runtime.internalId, userInfo)
+      hasStatusPermission = listOfPermissions.contains("status")
```
rtitle: Minor nitpick - can we do `.toSet.contains(..)`?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeServiceInterp.scala
```diff
@@ -222,6 +220,7 @@ class RuntimeServiceInterp[F[_]: Parallel](config: RuntimeServiceConfig,
               config.zombieRuntimeMonitorConfig.deletionConfirmationLabelKey,
               "false")
         .transaction
+      _ <- log.info(s"AT THE END OF DELETE")
```
rtitle: Remove log before merging

rtitle:
A few comments above, overall looks good! Please coordinate with @akarukappadath on who is going to merge first, rebase, etc.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/sam/SamAuthProvider.scala
```diff
@@ -211,7 +226,7 @@ class SamAuthProvider[F[_]: Effect: Logger](samDao: SamDAO[F],
 
   private def hasRuntimePermissionInternal(
     samResource: RuntimeSamResource,
-    action: LeoAuthAction,
```
Qi77Qi, line 214: why did we introduce LeoAuthAction?

xuscrbw: This is part of the changes Azmina made in her PR. 

LeoAuthAction gets extended by RuntimeAction/ProjectAction/PersistentDiskAction

xuscrbw:
jenkins retest

xuscrbw:
jenkins retest

xuscrbw:
jenkins retest

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/sam/SamAuthProvider.scala
```diff
@@ -152,6 +152,21 @@ class SamAuthProvider[F[_]: Effect: Logger](samDao: SamDAO[F],
...
+  ): F[List[RuntimeAction]] = {
+
+    val authorization = Authorization(Credentials.Token(AuthScheme.Bearer, userInfo.accessToken.token))
+
+    for {
+      listOfPermissions <- samDao
+        .getListOfResourcePermissions(samResource, authorization)
+
+    } yield {
+      listOfPermissions.map(x => runtimeActionMap.find(_._2 == x).map(_._1).get)
```
rtitle: This `Option.get` doesn't seem ideal.. and this probably isn't efficient since it's traversing `runtimeActionMap` each time.

What if we declared this in the class:
```
private val runtimeActionMapReversed = runtimeActionMap.map(_.swap)
```
Then here we could just yield:
```
listOfPermissions.flatMap(x => runtimeActionMapReversed.get(x))
// Or more concisely:
// listOfPermissions.flatMap(runtimeActionMapReversed.get)
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/sam/SamAuthProvider.scala
```diff
@@ -152,6 +152,21 @@ class SamAuthProvider[F[_]: Effect: Logger](samDao: SamDAO[F],
     checkPersistentDiskPermissionWithProjectFallback(samResource, authorization, action, googleProject)
   }
 
+  def getRuntimeActions(samResource: SamResource, userInfo: UserInfo)(
```
rtitle: Since this method is Runtime-specific, should it take `RuntimeSamResource` instead?

(Also maybe we want a PD version of this method as well)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/sam/SamAuthProvider.scala
rtitle: Yeah.. makes sense that these methods take specific types depending on resource

Qi77Qi: I might try to remove `LeoAuthAction` in a future PR...It makes inheritance more complicated and doesn't bring much value

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/sam/SamAuthProvider.scala
rtitle: @xuscrbw I wonder if this is where that `None.get` error you were seeing is coming from :)

rtitle (approved):

PR #1426, opened 2020-05-27
[IA-1594] Front Leo: incorporate disks to the runtime APIs

rtitle:
Adds an optional `diskConfig` parameter to the "create runtime" request JSON. The `diskConfig` has 2 forms:
1. create -- creates a new disk along with the runtime
2. reference -- attaches an existing disk to the new runtime

Examples:
```
// creates a new disk named my-new-disk along with the runtime
POST /api/google/v1/runtimes/project/name
{
  …,
  "diskConfig": {
    "create": true,
    "name": "my-new-disk",
    "size": 100
  }
}
```
```
// attaches an existing disk named my-existing-disk to the new runtime
POST /api/google/v1/runtimes/project/name
{
  …,
  "diskConfig": {
    "create": false,
    "name": "my-existing-disk"
  }
}
```

Also added disk information to `GetRuntimeResponse` and `ListRuntimeResponse`, and fixed a couple bugs.

---

In all cases:


core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodec.scala
```diff
@@ -221,5 +221,5 @@ object JsonCodec {
   implicit val diskStatusDecoder: Decoder[DiskStatus] =
     Decoder.decodeString.emap(x => DiskStatus.withNameOption(x).toRight(s"Invalid disk status: $x"))
   implicit val diskTypeDecoder: Decoder[DiskType] =
-    Decoder.decodeString.emap(x => DiskType.withNameInsensitiveOption(x).toRight(s"Invalid disk type: $x"))
+    Decoder.decodeString.emap(x => DiskType.googleStringToValues.get(x).toRight(s"Invalid disk type: $x"))
```
rtitle, line 228: We were serializing with `_.googleString`, need to deserialize the same way

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20200515_add_runtime_disk_foreign_key.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="leonardo" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+    <changeSet logicalFilePath="leonardo" author="rtitle" id="add_runtime_disk_foreign_key">
+        <addColumn tableName="CLUSTER">
+            <column name="persistentDiskId" type="BIGINT">
+                <constraints nullable="true" />
```
rtitle: I made it a nullable foreign key, because existing runtime records in the DB won't have a disk record. Nullable foreign keys are OK in MySQL.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/RuntimeConfigQueries.scala
```diff
@@ -6,26 +6,26 @@ import java.time.Instant
...
 
   val runtimeConfigs = TableQuery[RuntimeConfigTable]
 
-  def clusterLabelRuntimeConfigQuery(baseQuery: ClusterJoinLabel): ClusterJoinLabelJoinRuntimeConfigJoinPatch =
+  def runtimeLabelRuntimeConfigQuery(baseQuery: RuntimeJoinLabel): RuntimeJoinLabelJoinRuntimeConfigJoinPatch =
     for {
-      (((cluster, label), runTimeConfig), patch) <- baseQuery
-        .joinLeft(runtimeConfigs)
+      (((cluster, label), runtimeConfig), patch) <- baseQuery
+        .join(runtimeConfigs)
```
rtitle, line 24: Don't need to `joinLeft` on `runtimeConfigs` because it's a required association

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/RuntimeServiceDbQueries.scala
```diff
@@ -52,47 +77,56 @@ object RuntimeServiceDbQueries {
...
-            patchInProgress
+            patchInProgress,
+            diskConfigOpt
           )
       }.toList
     }
   }
+
+  def isDiskAttachedToRuntime(disk: PersistentDisk)(implicit ec: ExecutionContext): DBIO[Boolean] =
+    clusterQuery.filter(x => x.persistentDiskId.isDefined && x.persistentDiskId === disk.id).length.result.map(_ > 0)
```
rtitle, line 135: Reason #2032452346 I hate slick: you use `===` to compare optional and non-optional values

rtitle:
jenkins retest

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeServiceInterp.scala
```diff
@@ -92,21 +92,30 @@ class RuntimeServiceInterp[F[_]: Parallel](config: RuntimeServiceConfig,
...
             clusterImages <- getRuntimeImages(petToken, context.now, req.toolDockerImage, req.welderDockerImage)
-            gceRuntimeDefaults = config.gceConfig.runtimeConfigDefaults
+            diskOpt <- req.diskConfig.traverse(x =>
+              processDiskConfigRequest(x,
+                                       googleProject,
+                                       req.runtimeConfig.map(_.cloudService).getOrElse(CloudService.GCE),
+                                       userInfo)
+            )
+            // For GCE, make diskConfig.size take precedence over runtimeConfig.diskSize
+            gceRuntimeDefaults = diskOpt match {
```
Qi77Qi, line 102: do we still need the old config if we now have diskConfig?

rtitle: We still need `runtimeConfig`, just `GceRuntimeConfig.diskSize` is redundant now. We can remove it eventually but we need to keep it for now until disks are fully implemented and clients move over to using it.

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -2788,3 +2799,65 @@ components:
             Minimum size is 10GB. Note: disk size cannot be
 
             decreased in an update request, only increased.
+
+    DiskConfig:
+      description: The configuration of a persistent disk, returned in runtime responses
+      type: object
+      required:
+        - name
```
Qi77Qi, line 2697: are they all required? or do we have default values as well?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/RuntimeRoutes.scala
```diff
@@ -392,7 +429,8 @@ object RuntimeRoutes {
     "proxyUrl",
     "status",
     "labels",
-    "patchInProgress"
+    "patchInProgress",
+    "diskConfig"
```
Qi77Qi: is `diskConfig` needed for listRuntime?

http/src/main/resources/swagger/api-docs.yaml
rtitle: This `DiskConfig` is returned in responses -- all 4 fields are always present.

`DiskConfigRequest` is sent in requests and has optional fields.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/RuntimeRoutes.scala
rtitle: I think so, `listRuntime` should return info about the attached disk I think. It also returns `runtimeConfig`.

Qi77Qi: it's returning `runtimeConfig` only because it has been returning I think....last time I talked to Isaac, I think they only really need `listRuntime` to have `status` and image.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeServiceInterp.scala
```diff
@@ -513,6 +530,58 @@ class RuntimeServiceInterp[F[_]: Parallel](config: RuntimeServiceConfig,
...
+    userInfo: UserInfo
+  )(implicit as: ApplicativeAsk[F, AppContext]): F[PersistentDisk] =
+    (req, cloudService) match {
+      case (DiskConfigRequest.Reference(name), CloudService.GCE) =>
+        for {
+          diskOpt <- persistentDiskQuery.getActiveByName(googleProject, name).transaction
+          disk <- F.fromEither(diskOpt.toRight(DiskNotFoundException(googleProject, name)))
+          attached <- RuntimeServiceDbQueries.isDiskAttachedToRuntime(disk).transaction
+          _ <- if (attached) F.raiseError[Unit](DiskAlreadyAttachedException(googleProject, name)) else F.unit
+          hasAttachPermission <- authProvider.hasPersistentDiskPermission(disk.samResourceId,
```
Qi77Qi: do we need to check if this user is owner/creator for the disk as well?...

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/RuntimeRoutes.scala
rtitle: Ok, could check with the UI team then.. maybe I could leave it out of `listRuntime` for now, and add it later if needed.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeServiceInterp.scala
rtitle: We could but I don't think it's necessary -- the `attach` permission in Sam should handle that. In general I think Leo should delegate all access control questions to Sam and not try to make decisions based on owner email in its database.

Qi77Qi: does hasAttach sam call check if the caller(runtime's creator) is owner of the disk as well?

rtitle: This is called in a `create runtime` context.. so the permission check is as follows:
- does the caller have permission to create a runtime in this project?
- does the caller have permission to attach this disk (to any runtime)?

Maybe this permission should be called `attach-on-create` or something… if we supported attaching disks to _existing_ runtimes then yeah we'd probably need some additional permission checks.

akarukappadath (approved):
looks good to me! 
just wondering, does it make sense to completely stop supporting runtimes without PDs? Thinking about the case in which someone wants to just go into notebooks in playground mode - feels like requiring a PD for all runtimes would put unnecessary overhead on some users

rtitle:
Thanks @akarukappadath! Rebased and applied feedback so far.
> just wondering, does it make sense to completely stop supporting runtimes without PDs? Thinking about the case in which someone wants to just go into notebooks in playground mode - feels like requiring a PD for all runtimes would put unnecessary overhead on some users

It's a good question, not sure 100%. My impression is that in the future all GCE-based runtimes would use PDs since Terra UI would automatically manage them for you. But in Leo code we need to model everything as optional to support existing runtimes without PDs. Also Dataproc won't use PDs. Maybe there's a use case in the future for creating a GCE runtime with no detachable PD..

rtitle:

rtitle:
jenkins retest

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeServiceInterp.scala
```diff
@@ -91,6 +96,31 @@ class RuntimeServiceInterp[F[_]: Parallel](config: RuntimeServiceConfig,
             _ <- context.span.traverse(s => F.delay(s.addAnnotation("Done Sam getAccessToken")))
             clusterImages <- getRuntimeImages(petToken, context.now, req.toolDockerImage, req.welderDockerImage)
             _ <- context.span.traverse(s => F.delay(s.addAnnotation("Done get runtime images")))
+            diskOpt <- req.diskConfig.traverse(x =>
+              processDiskConfigRequest(x,
+                                       googleProject,
+                                       req.runtimeConfig.map(_.cloudService).getOrElse(CloudService.GCE),
+                                       userInfo)
+            )
+            _ <- context.span.traverse(s => F.delay(s.addAnnotation("Done process disk config")))
```
Qi77Qi, line 100: oh nice...I was gonna say that we can add some tracing for new routes now that tracing PR is merged

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeServiceInterp.scala
Qi77Qi: okay...so I guess we're just saying disks can be attached to anyone's runtime is by design then. Would be nice to mention this in some doc. (I think @akarukappadath is going to add a doc around sam permissions for disks, would be nice to add this there)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeServiceInterp.scala
```diff
@@ -629,3 +702,15 @@ final case class WrongCloudServiceException(runtimeCloudService: CloudService, u
       s"Bad request. This runtime is created with ${runtimeCloudService.asString}, and can not be updated to use ${updateCloudService.asString}",
       StatusCodes.Conflict
     )
+
+final case object DiskNotSupportedException
+    extends LeoException(
+      s"Persistent disks are not supported on Google Cloud Dataproc",
```
Qi77Qi: maybe add `traceId` in exceptions?

Qi77Qi (approved):

rtitle:
jenkins retest

rtitle:
sigh fiab-start issues. jenkins retest

rtitle:
jenkins retest

PR #1429, opened 2020-05-28
IA-1951 fix bug for patchInProgress

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-1951

---

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/MonitorAtBoot.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/MonitorAtBoot.scala
@@ -1,24 +1,25 @@
 package org.broadinstitute.dsde.workbench.leonardo
 package monitor
 
-import cats.effect.Async
+import cats.effect.{Async, Timer}
 import cats.implicits._
 import cats.mtl.ApplicativeAsk
 import fs2.Stream
 import io.chrisdavenport.log4cats.Logger
-import org.broadinstitute.dsde.workbench.leonardo.db.{clusterQuery, DbReference}
+import org.broadinstitute.dsde.workbench.leonardo.db.{clusterQuery, patchQuery, DbReference}
 import org.broadinstitute.dsde.workbench.leonardo.http.dbioToIO
 import org.broadinstitute.dsde.workbench.model.TraceId
-import org.broadinstitute.dsde.workbench.leonardo.db.patchQuery
 import org.broadinstitute.dsde.workbench.leonardo.http.cloudServiceSyntax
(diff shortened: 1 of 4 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/MonitorAtBoot.scala
```diff
@@ -36,8 +37,24 @@ class MonitorAtBoot[F[_]](implicit F: Async[F],
                     statusOpt <- clusterQuery.getClusterStatus(c.id).transaction
                     tid <- traceId.ask
                     s <- F.fromEither(
-                      statusOpt.toRight(new Exception(s"${c.id} not found after transition. This is very weird!"))
+                      statusOpt
+                        .toRight(new Exception(s"${tid} | ${c.id} not found after transition. This is very weird!"))
                     )
+                    now <- nowInstant[F]
+                    _ <- if (s != RuntimeStatus.Running) {
+                      // There's slight chance where pubsub message is never published during a redeploy.
```
akarukappadath, line 45: In this case, wouldn't there be a row in the DB with the patch details? As we were talking about yesterday, maybe this can be part of a series of DB checks after redeploy (until we have no-downtime deployments)

Qi77Qi: I didn't think we persist target machine type...just checked database, seems there is `masterMachineType`, is that the target machineType? and we only have patchInProgress when we need to change machine type?

akarukappadath: Yeah, that's the target machine type. If `patchInProgress` is true for a row in the DB but there's no `UpdateRuntimeMessage` in the queue after redeploy, then we should be able to make a new message for the queue using the `masterMachineType` in the `PatchTable`

akarukappadath: Ah but this might only be for updates that need a stop-start.. @jdcanas do you know?

Qi77Qi: yea, we won't be able to construct a complete UpdateRuntimeMessage because it has other fields like size, etc

rtitle:
Code looks fine I think, seems there's a unit test failure though

Qi77Qi:
@akarukappadath @rtitle  any additional comments for this PR?

rtitle (approved):
Looks good

PR #1432, opened 2020-06-01
IA-1953 globally exclude guava from automation

rtitle:
See https://broadworkbench.atlassian.net/browse/IA-1953

Globally exclude guava from all automation dependencies. It seems to fix it locally. Haven't run all tests yet.

(I just learned about this sbt feature, see https://www.scala-sbt.org/1.x/docs/Library-Management.html#Exclude+Transitive+Dependencies)

---
Have you read [CONTRIBUTING.md](https://github.com/DataBiosphere/leonardo/blob/develop/CONTRIBUTING.md) lately? If not, do that first.

I, the developer opening this PR, do solemnly pinky swear that:

- [ ] I've documented my API changes in Swagger

In all cases:

- [ ] Get a thumbsworth of review and PO signoff if necessary
- [ ] Verify all tests go green
- [ ] Squash and merge; Delete your branch after this
- [ ] Test this change deployed correctly and works on dev environment after deployment

PR #1435, opened 2020-06-03
* also not a real PR *

rtitle:
guava exclusion cherry-picked to last known "working" commit (to see if the global guava exclusion is causing an issue)

---

In all cases:

### DIFF
--- a/project/Settings.scala
+++ b/project/Settings.scala
@@ -115,6 +115,7 @@ object Settings {
 
   val automationSettings = commonSettings ++ List(
     libraryDependencies ++= automationDependencies,
+    excludeDependencies += excludeGuava,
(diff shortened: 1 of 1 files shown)


rtitle:
This passed. I don't think the global exclusion itself is causing issues; it's probably something else we pulled in in https://github.com/DataBiosphere/leonardo/pull/1395

Qi77Qi:
What's the difference between this PR and https://github.com/DataBiosphere/leonardo/pull/1432? diff looks exactly the same

rtitle:
This branch is based on the commit right before https://github.com/DataBiosphere/leonardo/pull/1395 was merged to develop

PR #1438, opened 2020-06-03
[no ticket] check project permissions as well

Qi77Qi:
Tested in fiab

The way I tested this:

- Add Ron to a billing project that Herminone is project owner. 
- Login as Ron in terra UI, and create a runtime
- Login as Hermione in swagger and try to delete the runtime Ron just created

---

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeServiceInterp.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeServiceInterp.scala
@@ -206,8 +206,9 @@ class RuntimeServiceInterp[F[_]: Parallel](config: RuntimeServiceConfig,
       // throw 404 if no GetClusterStatus permission
       // Note: the general pattern is to 404 (e.g. pretend the runtime doesn't exist) if the caller doesn't have
       // GetClusterStatus permission. We return 403 if the user can view the runtime but can't perform some other action.
-
-      listOfPermissions <- authProvider.getRuntimeActions(runtime.samResource, userInfo)
+      listOfPermissions <- authProvider.getRuntimeActionsWithProjectFallback(googleProject,
+                                                                             runtime.samResource,
+                                                                             userInfo)
       hasStatusPermission = listOfPermissions.toSet.contains(RuntimeAction.GetRuntimeStatus)
 
       _ <- ctx.span.traverse(s => F.delay(s.addAnnotation("Sam | Done first hasNotebookClusterPermission")))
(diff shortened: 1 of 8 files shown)


http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/dao/MockSamDAO.scala
```diff
@@ -105,7 +105,29 @@ class MockSamDAO extends SamDAO[IO] {
   override def getStatus(implicit ev: ApplicativeAsk[IO, TraceId]): IO[StatusCheckResponse] =
     IO.pure(StatusCheckResponse(true, Map.empty))
 
-  override def getListOfResourcePermissions(resource: SamResource, authHeader: Authorization)(implicit ev: ApplicativeAsk[IO, TraceId]): IO[List[String]] = IO.pure(List.empty)
+  override def getListOfResourcePermissions(resource: SamResource, authHeader: Authorization)(
```
xuscrbw, line 108: question on this implementation, is there a reason why we have a pattern match here, but not for the same method in httpsamdao?

xuscrbw (approved):
I think this looks good. Just had a question on the implementation

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/dao/MockSamDAO.scala
Qi77Qi: this is just mock...in httpsamdao, we'd be getting the info from sam.

rtitle:
👍 thanks for fixing

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeServiceInterp.scala
```diff
@@ -260,7 +261,9 @@ class RuntimeServiceInterp[F[_]: Parallel](config: RuntimeServiceConfig,
       // Note: the general pattern is to 404 (e.g. pretend the runtime doesn't exist) if the caller doesn't have
       // GetClusterStatus permission. We return 403 if the user can view the runtime but can't perform some other action.
 
-      listOfPermissions <- authProvider.getRuntimeActions(runtime.samResource, userInfo)
+      listOfPermissions <- authProvider.getRuntimeActionsWithProjectFallback(googleProject,
```
rtitle, line 264: Minor comment: technically we shouldn't need the project fallback for stop/start/update because there is no project-level permission for those actions. So we could avoid a Sam call for those cases. It's probably fine as is though

PR #1439, opened 2020-06-03
[no ticket] fix a bug where when host is down leonardo thinks kernel is still active

Qi77Qi:
Autopause works for usual cases, but then we can't get the host IP, Leonardo mistakenly thinks that indicates the kernel is not idle, which prevents it from autopausing the cluster (not sure how this ends up happening, but I saw one case from user liaison today, where master node is stopped, but Leo thinks the cluster still has active kernel.)

---

In all cases:


Qi77Qi:
jenkins retest...fiab failure

rtitle (approved):
Interesting.. also wondering how this occurs in practice but the change makes sense to me

PR #1443, opened 2020-06-08
[IA-1871][IA-1910] Jc k8 service

jdcanas:
This PR
- Adds a lot to the kuberenetes data model. See https://drive.google.com/file/d/1M6ZzmaTOCgyZt-qa7ZG2ILBUHGzoV-uD/view for references
- Adds routes for kubernetes
- adds the kubernetes service
- extensively tests all of the above

---

In all cases:


automation/gce-deploy/leonardo.conf
```diff
@@ -118,7 +118,6 @@ sam {
 serviceAccounts {
   # To implement your own service account provisioning, subclass org.broadinstitute.dsde.workbench.leonardo.model.ServiceAccountProvider
   # and specify your concrete class here
-  providerClass = "org.broadinstitute.dsde.workbench.leonardo.auth.DefaultServiceAccountProvider"
```
jdcanas, line 121: I know this is in an auto-generated file, but I just wanted to remember that I removed it for firecloud-develop cleanup, and It doesn't seem to be used anywhere.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/DefaultServiceAccountProvider.scala
```diff
@@ -1,28 +0,0 @@
-package org.broadinstitute.dsde.workbench.leonardo.auth
```
jdcanas, line 1: this isn't used anywhere I'm fairly sure, and it makes figuring out whats going on with service accounts hard

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/Config.scala
```diff
@@ -458,17 +463,85 @@ object Config {
...
+  }
+
+  implicit val releaseNameReader: ValueReader[ReleaseName] = stringValueReader.map(ReleaseName)
+  implicit val namespaceNameReader: ValueReader[NamespaceName] = stringValueReader.map(NamespaceName)
+
+  implicit val serviceReader: ValueReader[ServiceConfig] = ValueReader.relative { config =>
+    ServiceConfig(
+      config.as[ServiceName]("name"),
+      config.as[KubernetesServiceKindName]("kind"),
+      List() //TODO fill this out if we need ports
```
jdcanas, line 542: In my initial testing, I haven't had to specify a port number when I construct the URL with the service name. However, a service can have multiple ports, so we may need to specify this when we get further down the implementation path for more complex use cases

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/AppComponent.scala
```diff
@@ -0,0 +1,165 @@
...
+    )
+
+  def save(saveApp: SaveApp)(implicit ec: ExecutionContext): DBIO[App] = {
+    val namespaceName = saveApp.app.appResources.namespace.name
+    for {
+      nodepool <- nodepoolQuery
+        .getMinimalById(saveApp.app.nodepoolId)
+        .map(
+          _.getOrElse(
+            throw new SQLIntegrityConstraintViolationException(
```
jdcanas: we do this instead of allowing the slick `.save` to handle the constraint violation because we want to use the `clusterId` associated with the `nodepoolId` to save the app's namespace, instead of allowing the caller to specify the `clusterId` for the namespace.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesServiceDbQueries.scala
```diff
@@ -0,0 +1,256 @@
...
+  private[db] def joinFullAppAndUnmarshal(
+    baseQuery: Query[KubernetesClusterTable, KubernetesClusterRecord, Seq],
+    nodepoolQuery: Query[NodepoolTable, NodepoolRecord, Seq],
+    appQuery: Query[AppTable, AppRecord, Seq],
+    labelFilter: LabelMap = Map()
+  )(implicit ec: ExecutionContext): DBIO[List[KubernetesCluster]] =
+    joinFullApp(baseQuery, nodepoolQuery, appQuery, labelFilter).result
+      .map(rec => aggregateJoinedApp(rec).toList)
+
+  private[db] def joinFullApp(baseClusterQuery: Query[KubernetesClusterTable, KubernetesClusterRecord, Seq],
```
jdcanas, line 144: the `join`s and `joinLeft`s here work such we will not return anything if for a nodepool/cluster if it has no apps. See this for an example test https://github.com/DataBiosphere/leonardo/pull/1443/files#diff-b2baac6bc0230858b5892b940ccb9a55R144

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesServiceDbQueries.scala
```diff
@@ -0,0 +1,256 @@
...
+      .map {
+        case (serviceRec, portRecs) =>
+          serviceQuery.unmarshalService(serviceRec, portRecs.toList.toSet.toList)
+      }
+      .toSet
+      .toList
+}
+
+//minimal cluster has the nodepools, but no namespaces or apps
+case class SaveClusterResult(minimalCluster: KubernetesCluster, doesActiveClusterExist: Boolean)
```
jdcanas, line 286: this boolean is to allow the front leo service `createApp` call to differentiate the case in which it needs to tell back leo to create a cluster from when it just needs to create a nodepool

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesServiceDbQueries.scala
```diff
@@ -0,0 +1,256 @@
...
+                            nodepoolQuery.filter(_.destroyedDate === dummyDate),
+                            appQuery.findActiveByNameQuery(appName),
+                            labelFilter)
+      .map(clusters =>
+        clusters.headOption.fold[Option[GetAppResult]](
+          None
+        )(cluster =>
+          Some(
+            GetAppResult(
+              //we are guaranteed exactly one app and nodepool here if the query doesn't return None due to the way the query JOINs and database unique enforcement
```
jdcanas: This solution took some iteration to arrive on. From the call-site, it is highly desirable not to need to traverse the cluster-> nodepool -> app structure (a cluster contains a list of nodepools, and each nodepool has a list of apps) when we expect one 'app' to be returned.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/RuntimeServiceDbQueries.scala
```diff
@@ -86,7 +86,11 @@ object RuntimeServiceDbQueries {
             // .filter { lbl => (lbl.key, lbl.value) inSetBind labelMap.toSet }
             // Unfortunately slick doesn't support inSet/inSetBind for tuples.
             // https://github.com/slick/slick/issues/517
-            .filter(lbl => labelMap.map { case (k, v) => lbl.key === k && lbl.value === v }.reduce(_ || _))
+            .filter(lbl =>
+              labelMap
+                .map { case (k, v) => lbl.key === k && lbl.value === v }
+                .fold[Rep[Boolean]](false)(_ || _)
```
jdcanas, line 93: I noticed that this code has a bug in that it would fail if the labelMap passed to it is empty when writing a similar query

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeoKubernetesServiceInterp.scala
```diff
@@ -0,0 +1,412 @@
...
+
+      //TODO implement SAM check
+      hasPermission <- F.pure(true)
+      _ <- if (hasPermission) F.unit else F.raiseError[Unit](AuthorizationError(Some(userInfo.userEmail)))
+
+      canDelete = AppStatus.deletableStatuses.contains(appResult.app.status)
+      _ <- if (canDelete) F.unit
+      else F.raiseError[Unit](AppCannotBeDeletedException(googleProject, appName, appResult.app.status, ctx.traceId))
+//
+      //TODO: do atomically and send message at the same time?
```
jdcanas: I'm deferring this decision for when we design/implement the back leo messaging

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeoKubernetesServiceInterp.scala
```diff
@@ -0,0 +1,412 @@
...
+    )
+  }
+
+}
+
+object LeoKubernetesServiceInterp {
+  private[service] def processDiskConfigRequest[F[_]: Parallel](
+    req: DiskConfigRequest,
+    googleProject: GoogleProject,
+    isDataproc: Boolean,
```
jdcanas: this paramater isn't needed for this interp, but its used to illustrate what is needed to generalize to other interps

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/leonardoServiceContractModel.scala
```diff
@@ -298,3 +300,62 @@ object GetRuntimeResponse {
...
+        val errors = c.errors ++ n.errors ++ a.errors
+        ListAppResponse(
+          c.googleProject,
+          KubernetesRuntimeConfig(
+            n.numNodes,
+            n.machineType,
+            n.autoscalingEnabled
+          ),
+          errors,
+          if (errors.isEmpty) a.status else AppStatus.Error, //TODO: aggregate?
```
jdcanas, line 287: so, for now, based on discussions with Adrian, I think the UI will only care about the app status, unless there is an error somewhere else. We may want to aggregate in a more intelligent way down the line though

project/Dependencies.scala
```diff
@@ -166,6 +166,7 @@ object Dependencies {
     workbenchGoogleTest,
     "org.typelevel" %% "cats-mtl-core"  % "0.7.0",
     "org.typelevel" %% "cats-effect"    % "2.0.0", //forcing cats 2.0.0
+    "com.rms.miu" %% "slick-cats" % "0.10.1",
```
jdcanas, line 170: this allows us to `.traverse` with `DBIO` by providing the necessary implicit.

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodec.scala
```diff
@@ -29,6 +41,10 @@ object JsonCodec {
     "Google Dataproc does not support clusters with 1 non-preemptible worker. Must be 0, 2 or more.",
     List.empty
   )
```
kyuksel, line 43: I know it's not from this PR but was wondering where the _50GB_ min disk size comes from ([it seems to be 10GB](https://cloud.google.com/compute/docs/disks#introduction)) and why it's hardcoded in the message.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/LeoRoutes.scala
```diff
@@ -162,11 +164,19 @@ object LeoRoutes {
...
-  def validateClusterNameDirective(clusterNameString: String): Directive1[RuntimeName] =
+  def validateKubernetesName[A](nameString: String, apply: String => A): Directive1[A] =
     Directive { inner =>
-      validateClusterName(clusterNameString) match {
+      KubernetesName.withValidation(nameString, apply) match {
         case Left(e)  => failWith(e)
         case Right(c) => inner(Tuple1(c))
       }
     }
+
```
kyuksel: `validateNameDirective` and `validateName` seem to be specifically for clusters. Does it make sense to keep `cluster` in the method names?

kyuksel:
I don't see updates to the swagger doc. Is it not part of this ticket?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesServiceDbQueries.scala
```diff
@@ -0,0 +1,256 @@
...
+import scala.concurrent.ExecutionContext
+import org.broadinstitute.dsde.workbench.leonardo.db.LeoProfile.api._
+import cats.data.Chain
+import LeoProfile.mappedColumnImplicits._
+import org.broadinstitute.dsde.workbench.leonardo.db.LeoProfile.{dummyDate}
+import cats.implicits._
+import org.broadinstitute.dsde.workbench.leonardo.db.nodepoolQuery.unmarshalNodepool
+
+object KubernetesServiceDbQueries {
+  def listFullApps(googleProject: Option[GoogleProject],
```
kyuksel, line 19: What does `Full` in `listFullApps` signify?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeoKubernetesServiceInterp.scala
```diff
@@ -0,0 +1,412 @@
...
+    for {
+      ctx <- as.ask
+      appOpt <- KubernetesServiceDbQueries.getActiveFullAppByName(googleProject, appName).transaction
+      app <- F.fromOption(appOpt, AppNotFoundException(googleProject, appName, ctx.traceId))
+
+      //TODO ask SAM
+      hasPermission <- F.pure(true)
+      _ <- if (hasPermission) F.unit else F.raiseError[Unit](AppNotFoundException(googleProject, appName, ctx.traceId))
+    } yield GetAppResponse.fromDbResult(app)
+
```
kyuksel: Wondering how `listApp` performs? Trying to have an idea of its time complexity in terms of the number of clusters, nodepools and apps.

kyuksel: `listFullApps` does a `listByProject`, which means we never iterate over anything that doesn't belong to the GoogleProject at hand?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeoKubernetesServiceInterp.scala
```diff
@@ -0,0 +1,412 @@
...
+  case class LeoKubernetesConfig(serviceAccountConfig: ServiceAccountProviderConfig,
+                                 clusterConfig: KubernetesClusterConfig,
+                                 nodepoolConfig: NodepoolConfig,
+                                 galaxyAppConfig: GalaxyAppConfig,
+                                 diskConfig: PersistentDiskConfig)
+}
+case class AppNotFoundException(googleProject: GoogleProject, appName: AppName, traceId: TraceId)
+    extends LeoException(
+      s"Kubernetes cluster ${googleProject.value}/${appName.value} not found. Trace ID: ${traceId.asString}",
```
kyuksel:
```suggestion
      s"Kubernetes app ${googleProject.value}/${appName.value} not found. Trace ID: ${traceId.asString}",
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeoKubernetesServiceInterp.scala
```diff
@@ -0,0 +1,412 @@
...
+  )(
+    implicit as: ApplicativeAsk[F, AppContext]
+  ): F[Unit] =
+    for {
+      ctx <- as.ask
+      //TODO check SAM permissions
+      hasPermission <- F.pure(true)
+      _ <- if (hasPermission) F.unit else F.raiseError[Unit](AuthorizationError(Some(userInfo.userEmail)))
+
+      appOpt <- KubernetesServiceDbQueries.getActiveFullAppByName(googleProject, appName).transaction
```
kyuksel: Are we not allowing users to create apps with the same name (as long as all but one are `Deleted`)?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeoKubernetesServiceInterp.scala
```diff
@@ -0,0 +1,412 @@
...
+      _ <- appOpt.fold(F.unit)(c =>
+        F.raiseError[Unit](AppAlreadyExistsException(googleProject, appName, c.app.status, ctx.traceId))
+      )
+
+      samResourceId <- F.delay(AppSamResourceId(UUID.randomUUID().toString))
+//   TODO: notify SAM
+//      _ <- authProvider.notifyKubernetesClusterCreated(samResourceId)
+
+      saveCluster <- F.fromEither(getSavableCluster(userInfo, googleProject, ctx.now))
+      saveClusterResult <- KubernetesServiceDbQueries.saveOrGetForApp(saveCluster).transaction
```
kyuksel: I'm trying to understand how `saveOrGetForApp()` is a no-op for the case where we want to create an app on an existing cluster because `getSavableCluster()` appears to define a `dummyNodepool` with a random name the first time `createApp` is called so the second time `createApp()` is called on the same cluster, `getSavableCluster()` will define another random `dummyNodepool` name, which won't match the previous one so `saveOrGetForApp()` will think there is no such existing cluster. What am I missing?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeoKubernetesServiceInterp.scala
```diff
@@ -0,0 +1,412 @@
...
+
+        } yield savedDisk
+      case (_, true) =>
+        as.ask.flatMap(ctx => F.raiseError(DiskNotSupportedException(ctx.traceId)))
+    }
+
+  case class LeoKubernetesConfig(serviceAccountConfig: ServiceAccountProviderConfig,
+                                 clusterConfig: KubernetesClusterConfig,
+                                 nodepoolConfig: NodepoolConfig,
+                                 galaxyAppConfig: GalaxyAppConfig,
```
kyuksel: Should this be `appConfig: AppConfig` instead of `GalaxyAppConfig` (which would extend `AppConfig`) so we can support non-Galaxy apps, too?

kyuksel: Same for `GalaxyNodepoolConfig`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeoKubernetesServiceInterp.scala
```diff
@@ -0,0 +1,412 @@
...
+                                          userInfo: UserInfo,
+                                          req: CreateAppRequest,
+                                          now: Instant): Either[Throwable, Nodepool] = {
+    val auditInfo = AuditInfo(userInfo.userEmail, now, None, now)
+
+    val machineConfig = req.kubernetesRuntimeConfig.getOrElse(
+      KubernetesRuntimeConfig(
+        leoKubernetesConfig.nodepoolConfig.galaxyNodepoolConfig.numNodes,
+        leoKubernetesConfig.nodepoolConfig.galaxyNodepoolConfig.machineType,
+        leoKubernetesConfig.nodepoolConfig.galaxyNodepoolConfig.autoscalingEnabled
```
kyuksel: If I'm reading ^^ correctly, the config parameters are coming from `CreateAppRequest` but shouldn't they be coming from a general config as the default (dummy) nodepool will always have the same parameters that _we_ define (i.e. single-node, inexpensive machine)?

kyuksel (commented):
My brain got saturated by the time I went through `create` and `list` routes 🤯  
Will come back for round#2 later but hopefully there's some food for thought in the meantime (mostly questions trying to understand) 😊

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodec.scala
jdcanas: This is a boot disk requirement, since we take ~30GB of it for our images with the biggest set of runtime images possible. This probably should change with the change to a boot/user disk @Qi77Qi  ?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesServiceDbQueries.scala
jdcanas: its analogous to the word in our cluster queries, where 'full' does all joins possible. There is a 'minimal' query that only does a few joins, specified in a comment above

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodec.scala
kyuksel: That makes sense. I thought it was implied that the limitation was due to Google.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesServiceDbQueries.scala
kyuksel: Having a high-level method name reflect how the queries it does were implemented doesn't sound right/necessary to me.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/LeoRoutes.scala
```diff
@@ -151,9 +153,9 @@ class LeoRoutes(
 object LeoRoutes {
   private val clusterNameReg = "([a-z|0-9|-])*".r
```
kyuksel: Should these `validate*()` methods live outside `LeoRoutes` given they're used by `KubernetesRoutes` too? That way `KubernetesRoutes` would have no dependency on `LeoRoutes`.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/leonardoServiceContractModel.scala
```diff
@@ -298,3 +300,62 @@ object GetRuntimeResponse {
...
+
+final case class CreateAppRequest(kubernetesRuntimeConfig: Option[KubernetesRuntimeConfig],
+                                  appType: AppType,
+                                  diskConfig: Option[DiskConfigRequest],
+                                  labels: LabelMap = Map.empty)
+
+final case class GetAppResponse(kubernetesRuntimeConfig: KubernetesRuntimeConfig,
+                                errors: List[KubernetesError],
+                                status: AppStatus, //TODO: do we need some sort of aggregate status?
+                                proxyUrls: Map[ServiceName, URL],
```
kyuksel, line 246: Will we have more than one `proxyUrl`s?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesServiceDbQueries.scala
kyuksel: Convert the comment into assertions (that `nodepools` and `apps` `size`s must be 1) so we can get meaningful error messages if the assertions happen to be violated?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/GalaxyAppConfig.scala
```diff
@@ -0,0 +1,6 @@
+package org.broadinstitute.dsde.workbench.leonardo.config
+
+import org.broadinstitute.dsde.workbench.google2.KubernetesSerializableName.NamespaceName
+import org.broadinstitute.dsde.workbench.leonardo.{ReleaseName, ServiceConfig}
+
+case class GalaxyAppConfig(releaseName: ReleaseName, namespaceNameSuffix: NamespaceName, services: List[ServiceConfig])
```
kyuksel, line 6: Why `namespaceNameSuffix` and not `namespaceName`?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeoKubernetesServiceInterp.scala
```diff
@@ -0,0 +1,412 @@
...
+        Left(IllegalLabelKeyException(includeDeletedKey))
+      else
+        Right(allLabels)
+      //galaxy apps need a disk
+      disk <- if (req.appType == AppType.Galaxy && diskOpt.isEmpty)
+        Left(AppRequiresDiskException(googleProject, appName, req.appType, ctx.traceId))
+      else Right(diskOpt)
+      namespaceName <- KubernetesName.withValidation(
+        s"${appName.value}-${leoKubernetesConfig.galaxyAppConfig.namespaceNameSuffix.value}",
+        NamespaceName.apply
```
kyuksel: Why are we not using the namespace name as is?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeoKubernetesServiceInterp.scala
```diff
@@ -0,0 +1,412 @@
...
+              }
+              .filterNot(_.apps.isEmpty)
+          )
+        }
+        .filterNot(_.nodepools.isEmpty)
+        .flatMap(c => ListAppResponse.fromCluster(c))
+        .toVector
+    }
+
+  override def deleteApp(userInfo: UserInfo, googleProject: GoogleProject, appName: AppName)(
```
kyuksel: Do we need to add another input parameter to differentiate between soft/hard deletions, leaving/deleting the disk?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeoKubernetesServiceInterp.scala
```diff
@@ -0,0 +1,412 @@
...
+      _ <- if (hasPermission) F.unit else F.raiseError[Unit](AuthorizationError(Some(userInfo.userEmail)))
+
+      canDelete = AppStatus.deletableStatuses.contains(appResult.app.status)
+      _ <- if (canDelete) F.unit
+      else F.raiseError[Unit](AppCannotBeDeletedException(googleProject, appName, appResult.app.status, ctx.traceId))
+//
+      //TODO: do atomically and send message at the same time?
+      _ <- nodepoolQuery.markPendingDeletion(appResult.nodepool.id).transaction
+      _ <- appQuery.markPendingDeletion(appResult.app.id).transaction
+      //TODO queue stuff
```
kyuksel: Do we need another `TODO` for disk-related work?

kyuksel (commented):
Round#2: Comments and questions inline around `get` and `delete` flows.
Will do another overall go-through.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeoKubernetesServiceInterp.scala
jdcanas: I put the namespace in a config, but we can't use just that since a namespace must be unique at the cluster-level, so we must use the appname as well

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/GalaxyAppConfig.scala
jdcanas: see below comment

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/leonardoServiceContractModel.scala
jdcanas: Yeah, so since we need to support multiple services, we will need multiple proxy urls. For example, to support routing to both cloudman and galaxy for an app. At first, there should only be one, but I want the UI to get used to looking it up in a map by a static service name so we don't have to change that contract down the line

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/LeoRoutes.scala
jdcanas: Originally I wanted to have it so that all routes validating 'leo entities' the same way, but kubernetes names have a different regex they need to follow. I think I'll keep it and apply the pattern to disk routes though, since the same validation applies

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesServiceDbQueries.scala
jdcanas: I think it helps, because it makes the caller aware of what they are going to be getting back, and consider if the alternative lower-cost query is desirable. I'll add some comments. 

The alternative to this is specifying two different ADTs that have 1 field different (one having a list of apps, and the other not), which I think would be more confusing since it would require different handling at call site.

Its worth mentioning that the minimal call is currently only used in testing, but it will be needed when/if we layer in ZombieMonitor-like cleanup of nodepools/cluster separate from the deletion endpoint

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesServiceDbQueries.scala
jdcanas: Done. Its a bit ugly, but I wanted the assertions to follow the `DBIO.failure` pattern.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeoKubernetesServiceInterp.scala
jdcanas: `listByProject` takes an `Option[GoogleProject]` and handles it via 

```
googleProject match {
        case Some(project) => kubernetesClusterQuery.filter(_.googleProject === project)
        case None          => kubernetesClusterQuery
      }
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeoKubernetesServiceInterp.scala
jdcanas: hmm... so right now I only do 'soft deletes'. I think I'll address this with my follow-up PR regarding pub/sub messaging. To ease my cognitive burden, I punted since on the difference since, right now, front leo only marks things as 'deleting', and never 'deleted'; the deleting->deleted transition will be implemented in back-leo and need to be included in the pub/sub messaging. I think you're right, in that the follow-up ticket should add an affordance to this endpoint for it as well.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeoKubernetesServiceInterp.scala
jdcanas: yeah. I updated the TODO to read     `//TODO queue disk/cluster/nodepool/app creation `

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeoKubernetesServiceInterp.scala
jdcanas: Ah, this function is poorly named from an early iteration. I renamed it to `getUserNodepool`. The dummy nodepool is created from a different config within the `getSavableCluster` function, since saving a cluster explicitly requires the dummy nodepool in the `SaveCluster` ADT

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeoKubernetesServiceInterp.scala
jdcanas: I'm imagining that as these are extended, we will have a map of [AppType,A], but I think we should avoid designing the configs for this until we know how close/extendable our current config code is. In other words, I want to punt on how exactly we handle multiple app configs since I didn't want to be too predictive in terms of what they need. Similarly for nodepool configs, I think that we may want to have a map when we have more than one, and be really specific to galaxy here

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeoKubernetesServiceInterp.scala
jdcanas: oops.. too much pasta

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodec.scala
```diff
@@ -6,12 +6,22 @@ import java.time.Instant
...
 import org.broadinstitute.dsde.workbench.leonardo.SamResource.{
   PersistentDiskSamResource,
   ProjectSamResource,
   RuntimeSamResource
 }
+import org.broadinstitute.dsde.workbench.google2.{DiskName, ZoneName}
+import org.broadinstitute.dsde.workbench.google2.GKEModels.{KubernetesClusterName, NodepoolName}
+import org.broadinstitute.dsde.workbench.google2.KubernetesModels.KubernetesApiServerIp
+import org.broadinstitute.dsde.workbench.google2.KubernetesSerializableName.{NamespaceName, ServiceName}
+import org.broadinstitute.dsde.workbench.google2.{
```
akarukappadath, line 18: nit: add `DiskName` and `ZoneName` here to consolidate

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
```diff
@@ -18,20 +28,16 @@ case class KubernetesCluster(id: KubernetesClusterLeoId,
                              location: Location,
                              status: KubernetesClusterStatus,
                              serviceAccount: WorkbenchEmail,
-                             samResourceId: KubernetesClusterSamResourceId,
```
akarukappadath, line 21: could you remind me why we don't want Sam to manage k8s clusters?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/AppComponent.scala
```diff
@@ -0,0 +1,190 @@
...
+import org.broadinstitute.dsde.workbench.model.google.GoogleProject
+
+import scala.concurrent.ExecutionContext
+
+final case class AppRecord(id: AppId,
+                           nodepoolId: NodepoolLeoId,
+                           appType: AppType,
+                           appName: AppName,
+                           status: AppStatus,
+                           samResourceId: AppSamResourceId,
```
akarukappadath: Can we use the `RuntimeSamResource` type here? Or create a new `KubernetesAppSamResource` type in `samModels`?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/AppComponent.scala
```diff
@@ -0,0 +1,190 @@
...
+  ): Query[AppTable, AppRecord, Seq] =
+    appQuery
+      .filter(_.appName === appName)
+      .filter(_.destroyedDate === dummyDate)
+}
+
+case class SaveApp(app: App)
+case class AppExistsForProjectException(appName: AppName, googleProject: GoogleProject)
+    extends LeoException(
+      s"An app with name ${appName} already exists for the prokect ${googleProject}.",
```
akarukappadath: nit: typo in "project"

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesClusterComponent.scala
```diff
@@ -246,14 +210,30 @@ object kubernetesClusterQuery extends TableQuery(new KubernetesClusterTable(_))
...
-) //the clusterId specified here isn't used, and will be replaced by the id of cluster saved beforehand
+case class KubernetesAppCreationException(message: String) extends Exception
+
+case class SaveKubernetesCluster(googleProject: GoogleProject,
+                                 clusterName: KubernetesClusterName,
+                                 location: Location,
+                                 status: KubernetesClusterStatus,
+                                 serviceAccount: WorkbenchEmail,
+                                 auditInfo: AuditInfo,
+                                 dummyNodepool: Nodepool) {
```
akarukappadath: is this being used anywhere?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesServiceDbQueries.scala
```diff
@@ -0,0 +1,288 @@
...
+      } joinLeft
+        namespaceQuery on (_._1._2.id === _.clusterId) join
+        namespaceQuery on (_._1._1._1._1.namespaceId === _.id) joinLeft
+        serviceQuery on (_._1._1._1._1._1.id === _.appId) joinLeft
+        portQuery on (_._2.map(_.id) === _.serviceId) joinLeft
+        persistentDiskQuery on (_._1._1._1._1._1._1._1.diskId === _.id) joinLeft
+        labelQuery on {
+        case (c, lbl) =>
+          lbl.resourceId
+            .mapTo[DiskId] === c._1._1._1._1._1._1._1._1.diskId && lbl.resourceType === LabelResourceType.persistentDisk
```
akarukappadath, line 167: this is hilarious

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeoKubernetesServiceInterp.scala
```diff
@@ -0,0 +1,412 @@
...
+import org.broadinstitute.dsde.workbench.leonardo.SamResource.PersistentDiskSamResource
+import org.broadinstitute.dsde.workbench.leonardo.http.api.{CreateDiskRequest, DiskConfigRequest}
+import org.broadinstitute.dsde.workbench.leonardo.http.service.LeoKubernetesServiceInterp.LeoKubernetesConfig
+import org.broadinstitute.dsde.workbench.leonardo.model.PersistentDiskAction.AttachPersistentDisk
+import org.broadinstitute.dsde.workbench.leonardo.model.ProjectAction.CreatePersistentDisk
+import org.broadinstitute.dsde.workbench.leonardo.service.LeoKubernetesService
+
+import scala.concurrent.ExecutionContext
+
+class LeoKubernetesServiceInterp[F[_]: Parallel](
```
akarukappadath: nit: could we call this just `KubernetesServiceInterp` like `DiskServiceInterp` and `RuntimeServiceInterp`?

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
jdcanas: If a user creates an app, and a kubernetes cluster does not exist, Leo will automagically create the cluster. In other words, the fact that an app needs a cluster is more of a leo implementation detail, and its assumed that if a user has the ability to create apps in SAM that they can create the cluster. Deleting a cluster is also something managed by Leo, and not explicitly exposed on the UI.

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodec.scala
jdcanas: huh.. I'm surprised scalafmt didn't catch this.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/AppComponent.scala
jdcanas: oh yeah. I started this before the samModels PR, thanks for reminding me to follow up :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesClusterComponent.scala
jdcanas: Yeah, its used here https://github.com/DataBiosphere/leonardo/pull/1443/files#diff-aa9679f7f59adb5590ab00dfff5fe0e6R187. 

Its because a cluster needs a nodepool to be created in kubernetes, so I wanted to enforce it in the ADT

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesServiceDbQueries.scala
jdcanas: I'm just glad there were existing examples 😂

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeoKubernetesServiceInterp.scala
jdcanas: I avoided this initially because the the wb-libs class is also `KubernetesService`. I guess we can overload it and be explicit when the google one would conflict for consistency

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeoKubernetesServiceInterp.scala
jdcanas: Yeah, we enforce uniqueness on (AppName, GoogleProject) here for active apps https://github.com/DataBiosphere/leonardo/pull/1443/files#diff-71f91335642354f5fe16600d475c12d8R130, as opposed to enforcing global uniqueness on appName

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeoKubernetesServiceInterp.scala
jdcanas: in `saveOrGetForApp`, we do `kubernetesClusterQuery.getMinimalActiveClusterByName(saveKubernetesCluster.googleProject)`, which just looks for any active clusters in the project. We do erroneously create the dummy nodepool in `getSavableCluster ` and its more or less discarded if a cluster exists, but its not used to check for cluster existence, and fairly low overhead

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeoKubernetesServiceInterp.scala
kyuksel: It makes sense that we need a unique name for the Galaxy instance. I just couldn't understand why we need the concatenation on that line. The `App` object we're `yield`ing already contains an `AppName` and a `NamespaceName`, which, when combined, should be unique, no?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/KubernetesServiceInterp.scala
```diff
@@ -0,0 +1,412 @@
...
+
+  case class LeoKubernetesConfig(serviceAccountConfig: ServiceAccountProviderConfig,
+                                 clusterConfig: KubernetesClusterConfig,
+                                 nodepoolConfig: NodepoolConfig,
+                                 galaxyAppConfig: GalaxyAppConfig,
+                                 diskConfig: PersistentDiskConfig)
+}
+case class AppNotFoundException(googleProject: GoogleProject, appName: AppName, traceId: TraceId)
+    extends LeoException(
+      s"Kubernetes cluster ${googleProject.value}/${appName.value} not found. Trace ID: ${traceId.asString}",
```
kyuksel:
```suggestion
      s"App ${googleProject.value}/${appName.value} not found. Trace ID: ${traceId.asString}",
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesClusterComponent.scala
kyuksel: The term GCP uses for the nodepool that needs to be created as part of the cluster is `default nodepool`. How would you feel about renaming `dummy` as `default` so it's one less thing to know.

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
```diff
@@ -133,13 +155,137 @@ final case class Nodepool(id: NodepoolLeoId,
...
+  def values: Set[AppType] = sealerate.values[AppType]
+  def stringToObject: Map[String, AppType] = values.map(v => v.toString -> v).toMap
+}
+
+final case class ReleaseName(value: String) extends AnyVal
+
+final case class AppId(id: Long) extends AnyVal
+final case class AppName(value: String) extends AnyVal
+//These are async from the perspective of Front Leo saving the app record, but both must exist before the helm command is executed
+final case class AppResources(namespace: Namespace, disk: Option[PersistentDisk], services: List[KubernetesService])
```
kyuksel, line 222: Where are we modeling namespace/app-specific Kubernetes service accounts? Should it be part of `AppResources`?

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
```diff
@@ -133,13 +155,137 @@ final case class Nodepool(id: NodepoolLeoId,
...
+      service.config.name ->
+        new URL(
+          s"${apiServerIp.url}/api/v1/namespaces/${appResources.namespace.name.value}/services/${service.config.name}/proxy/"
+        )
+    }.toMap
+}
+
+sealed abstract class AppStatus
+object AppStatus {
+  case object StatusUnspecified extends AppStatus {
```
kyuksel: Would `Unspecified` suffice?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/DiskRoutes.scala
```diff
@@ -57,7 +55,7 @@ class DiskRoutes(diskService: DiskService[IO], userInfoDirectives: UserInfoDirec
               }
             } ~
               pathPrefix(Segment) { diskNameString =>
-                validateDiskNameDirective(diskNameString) { diskName =>
+                LeoRoutes.validateNameDirective(diskNameString, DiskName.apply) { diskName =>
```
kyuksel: Feels like it'd be cleaner if all `validate*NameDirective`s lived in one separate file which could then be imported by `*Routes`.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/LeoRoutes.scala
```diff
@@ -151,22 +152,22 @@ class LeoRoutes(
...
         Left(
           RequestValidationError(
-            s"invalid cluster name ${clusterNameString}. Only lowercase alphanumeric characters, numbers and dashes are allowed in cluster name"
+            s"Invalid name ${clusterNameString}. Only lowercase alphanumeric characters, numbers and dashes are allowed in leo names"
           )
         )
     }
 
-  def validateClusterNameDirective(clusterNameString: String): Directive1[RuntimeName] =
+  def validateNameDirective[A](clusterNameString: String, apply: String => A): Directive1[A] =
```
kyuksel: This method is called by `{Disk,Runtime}Routes` as well so I'd rename `clusterNameString` as `NameString`. 
Same for `private def validateName`
Also see previous comment on moving these to a non-LeoRoutes-specific file for the same reason.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/DummyNodepoolConfig.scala
```diff
@@ -0,0 +1,6 @@
+package org.broadinstitute.dsde.workbench.leonardo.config
+
+import org.broadinstitute.dsde.workbench.google2.MachineTypeName
+import org.broadinstitute.dsde.workbench.leonardo.NumNodes
+
+case class DummyNodepoolConfig(machineType: MachineTypeName, numNodes: NumNodes, autoscalingEnabled: Boolean)
```
kyuksel: Won't `DummyNodepoolConfig` always have `numNodes = 1` and `autoscalingEnabled = false`?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/KubernetesServiceInterp.scala
```diff
@@ -0,0 +1,412 @@
...
+
+      diskOpt <- req.diskConfig.traverse(diskReq =>
+        LeoKubernetesServiceInterp.processDiskConfigRequest(
+          diskReq,
+          googleProject,
+          false,
+          userInfo,
+          authProvider,
+          serviceAccountProvider,
+          leoKubernetesConfig.diskConfig
```
kyuksel, line 100: Shouldn't `PersistentDiskConfig` value be coming from user's app creation request? `leoKubernetesConfig` seems to be defined at Leo boot time?

kyuksel (commented):
Still going through the diff's. Will resume tomorrow morning 😪

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/AppComponent.scala
```diff
@@ -0,0 +1,190 @@
...
+
+  def save(saveApp: SaveApp)(implicit ec: ExecutionContext): DBIO[App] = {
+    val namespaceName = saveApp.app.appResources.namespace.name
+    for {
+      nodepool <- nodepoolQuery
+        .getMinimalById(saveApp.app.nodepoolId)
+        .map(
+          _.getOrElse(
+            throw new SQLIntegrityConstraintViolationException(
+              "Apps must be saved with an nodepool ID that exists in the DB. FK_APP_NODEPOOL_ID"
```
kyuksel: Nit:
```suggestion
              "Apps must be saved with a nodepool ID that exists in the DB. FK_APP_NODEPOOL_ID"
```

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesClusterComponentSpec.scala
```diff
@@ -57,52 +58,58 @@ class KubernetesClusterComponentSpec extends FlatSpecLike with TestComponent {
...
-    //list deleted
-    val listCluster2 =
-      dbFutureValue(kubernetesClusterQuery.listFullClusters(savedCluster1.googleProject, includeDeleted = true))
-    val getCluster3 = dbFutureValue(kubernetesClusterQuery.getFullClusterById(savedCluster3.id)).get
-    listCluster2.size shouldBe 3
-    listCluster2 should contain(savedCluster1)
-    listCluster2 should contain(savedCluster2)
-    listCluster2 should contain(getCluster3)
-  }
+  //TODO: move to app
```
kyuksel: app?

kyuksel (commented):
Finished going through 😊 Looks good overall. A couple of more comments inline.
Marked as resolved the questions/comments addressed previously.

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
jdcanas: I held off on kubernetes service account handling since the design wasn't finalized when I was coding this up. 

Not sure exactly what you mean by the namespace modeling, this ADT does have a field for the app namespace if that's what you are referring to?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/DiskRoutes.scala
jdcanas: Yeah... I held off because the object this is contained in is an overloaded name, which makes intellij unhappy with refactoring... I'll bite the bullet

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/DummyNodepoolConfig.scala
jdcanas: Yes, but I wanted to leave it configurable via a `.conf` in case that needs to change. The alternative is hard-coding it, so I'd prefer to have the configurability here

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesClusterComponent.scala
jdcanas: I originally had it as default, but we've been calling it 'dummy' as a team, so I was thinking it could be more approachable this way. I'm fine with changing it back

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesClusterComponentSpec.scala
jdcanas: ah, this was leftover from when I was originally designing the routes to support create/get/list cluster. I kept this as a reference since the list app test would be similar. This is no longer needed :)

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
kyuksel: > I held off on kubernetes service account handling since the design wasn't finalized when I was coding this up.

Cool. It'd be good to capture that somewhere (ideally on an existing or new ticket?).

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/DiskRoutes.scala
kyuksel: thanks 🙏

automation/gce-deploy/leonardo.conf
rtitle: +1, I suspect the stuff in `gce-deploy` is probably dead code and doesn't work anyway

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20200520_kubernetes_refactor.xml
```diff
@@ -0,0 +1,104 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="leonardo" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+    <changeSet logicalFilePath="leonardo" author="jcanas" id="kubernetes_drop_refactor">
```
rtitle, line 3: Do these tables have any data in dev/alpha/perf/prod? Sometimes migrations work on an empty DB but fail when there is data -- want to make sure we avoid that here.

(Now that I'm typing this, we probably don't have any k8s records in dev/etc because the routes don't exist yet)

http/src/main/resources/reference.conf
```diff
@@ -162,6 +162,37 @@ groups {
   #dataprocImageProjectGroupEmail = ${groups.dataprocImageProjectGroupName}"@{{$appsSubdomain}}"
 }
 
+gke {
```
rtitle, line 166: Nit: we tend to indent by 2 spaces in this file

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
jdcanas: I have a placeholder for it in this ticket https://broadworkbench.atlassian.net/browse/IA-1911. I just added details

rtitle:
👍 
Not my most detailed review, but I scanned the changes and all looks reasonable. I'm pro on merging this and iterating in separate PRs if there are issues/followup work.

However seems this PR has some conflicts (maybe due to @Qi77Qi's recent model changes) and also some unit test failures.

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
```diff
@@ -133,13 +155,137 @@ final case class Nodepool(id: NodepoolLeoId,
...
                           machineType: MachineTypeName,
                           numNodes: NumNodes,
-                          autoScalingEnabled: Boolean,
-                          autoscalingConfig: Option[NodepoolAutoscaling],
-                          errors: List[RuntimeError])
+                          autoscalingEnabled: Boolean,
+                          autoscalingConfig: Option[AutoscalingConfig],
+                          errors: List[KubernetesError],
+                          apps: List[App])
+object KubernetesNameUtils {
```
Qi77Qi, line 162: is this a test util?

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
```diff
@@ -133,13 +155,137 @@ final case class Nodepool(id: NodepoolLeoId,
...
+object ErrorSource {
+  case object Cluster extends ErrorSource {
+    override def toString: String = "Cluster"
+  }
+
+  case object Nodepool extends ErrorSource {
+    override def toString: String = "Nodepool"
+  }
+
+  case object App extends ErrorSource {
```
Qi77Qi, line 199: should we give this a name. Think it'd be useful to know which app it is

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20200520_kubernetes_refactor.xml
```diff
@@ -0,0 +1,104 @@
...
+                <constraints nullable="false"/>
+            </column>
+            <column name="diskId" type="BIGINT"/>
+        </createTable>
+        <addForeignKeyConstraint baseColumnNames="nodepoolId" baseTableName="APP" constraintName="FK_APP_NODEPOOL_ID" deferrable="false" initiallyDeferred="false" onDelete="NO ACTION" onUpdate="NO ACTION" referencedColumnNames="id" referencedTableName="NODEPOOL"/>
+        <addForeignKeyConstraint baseColumnNames="diskId" baseTableName="APP" constraintName="FK_APP_DISK_ID" deferrable="false" initiallyDeferred="false" onDelete="NO ACTION" onUpdate="NO ACTION" referencedColumnNames="id" referencedTableName="PERSISTENT_DISK"/>
+        <addForeignKeyConstraint baseColumnNames="namespaceId" baseTableName="APP" constraintName="FK_APP_NAMESPACE_ID" deferrable="false" initiallyDeferred="false" onDelete="NO ACTION" onUpdate="NO ACTION" referencedColumnNames="id" referencedTableName="NAMESPACE"/>
+        <addUniqueConstraint columnNames="appName, nodepoolId" constraintName="IDX_APP_UNIQUE" tableName="APP"/>
+
+        <createTable tableName="SERVICE">
```
Qi77Qi, line 63: what is a "service"?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/RouteValidation.scala
```diff
@@ -0,0 +1,44 @@
...
+import scala.util.control.NoStackTrace
+
+object RouteValidation {
+  private val leoNameReg = "([a-z|0-9|-])*".r
+
+  final case class RequestValidationError(message: String) extends NoStackTrace {
+    override def getMessage: String = message
+  }
+
+  private def validateName(nameString: String): Either[Throwable, String] =
```
Qi77Qi, line 18: minor, probably useful to add a String input param that indicates what name this is for , say `disk name`, `gce VM name` etc

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/Config.scala
```diff
@@ -458,17 +463,85 @@ object Config {
...
   val dataprocMonitorConfig = config.as[DataprocMonitorConfig]("dataproc.monitor")
   val uiConfig = config.as[ClusterUIConfig]("ui")
-  val serviceAccountProviderClass = config.as[String]("serviceAccounts.providerClass")
   val samAuthConfig = config.as[SamAuthProviderConfig]("auth.providerConfig")
   val httpSamDap2Config = config.as[HttpSamDaoConfig]("auth.providerConfig")
   val liquibaseConfig = config.as[LiquibaseConfig]("liquibase")
   val welderConfig = config.as[WelderConfig]("welder")
   val dbConcurrency = config.as[Long]("mysql.concurrency")
 
+  implicit val kubeClusterConfigReader: ValueReader[KubernetesClusterConfig] = ValueReader.relative { config =>
```
Qi77Qi, line 499: not directly related to this PR....but I'm thinking we should probably make the effort to pureconfig one day. just so we don' have to add this many `configReaders`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/DefaultNodepoolConfig.scala
```diff
@@ -0,0 +1,6 @@
+package org.broadinstitute.dsde.workbench.leonardo.config
+
+import org.broadinstitute.dsde.workbench.google2.MachineTypeName
+import org.broadinstitute.dsde.workbench.leonardo.NumNodes
+
+case class DefaultNodepoolConfig(machineType: MachineTypeName, numNodes: NumNodes, autoscalingEnabled: Boolean)
```
Qi77Qi, line 6: kinda prefer to have these configs in one file so that it's easier to see relationships between them if any

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/KubernetesRoutes.scala
```diff
@@ -0,0 +1,202 @@
...
+import org.broadinstitute.dsde.workbench.leonardo.http.api.DiskConfigRequest.{Create, Reference}
+import org.broadinstitute.dsde.workbench.leonardo.service.KubernetesService
+
+class KubernetesRoutes(kubernetesService: KubernetesService[IO], userInfoDirectives: UserInfoDirectives)(
+  implicit timer: Timer[IO]
+) {
+
+  val routes: server.Route = userInfoDirectives.requireUserInfo { userInfo =>
+    CookieSupport.setTokenCookie(userInfo, CookieSupport.tokenCookieName) {
+      implicit val traceId = ApplicativeAsk.const[IO, TraceId](TraceId(UUID.randomUUID()))
```
Qi77Qi: I don't think this is needed...AppContext has as traceId

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/AppComponent.scala
Qi77Qi: I think you can do `.flatMap{x => x.fold(DBIO.failed(exception), DBIO.successful)}

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/AppComponent.scala
```diff
@@ -0,0 +1,190 @@
...
+              "Apps must be saved with a nodepool ID that exists in the DB. FK_APP_NODEPOOL_ID"
+            )
+          )
+        )
+
+      cluster <- kubernetesClusterQuery
+        .getMinimalClusterById(nodepool.clusterId)
+        .map(
+          _.getOrElse(
+            throw new SQLIntegrityConstraintViolationException(
```
Qi77Qi: same comment as above

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesServiceDbQueries.scala
```diff
@@ -0,0 +1,288 @@
...
+          cluster.status match {
+            case KubernetesClusterStatus.Precreating  => failure
+            case KubernetesClusterStatus.Provisioning => failure
+            case _                                    => DBIO.successful(SaveClusterResult(cluster, true))
+          }
+        case _ => kubernetesClusterQuery.save(saveKubernetesCluster).map(c => SaveClusterResult(c, false))
+      }
+    } yield eitherClusterOrError
+
+  case class GetAppAssertion(message: String) extends Exception
```
Qi77Qi: move this out of the object

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesServiceDbQueries.scala
```diff
@@ -0,0 +1,288 @@
...
+                                        Map[String, Chain[String]],
+                                        Map[PersistentDiskRecord, Map[String, Chain[String]]])]]
+  ): List[Nodepool] =
+    nodepools
+      .map {
+        case (nodepoolRec, appMap) =>
+          unmarshalNodepool(nodepoolRec, unmarshalAppMap(appMap))
+      }
+      .toSet
+      .toList
```
Qi77Qi, line 245: can we get rid of `toSet` here?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/RuntimeServiceDbQueries.scala
Qi77Qi: yea..`reduce` will throw exception when collection is empty

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/KubernetesServiceInterp.scala
```diff
@@ -0,0 +1,412 @@
...
+
+  private[service] def getSavableApp(googleProject: GoogleProject,
+                                     appName: AppName,
+                                     userInfo: UserInfo,
+                                     samResourceId: AppSamResource,
+                                     req: CreateAppRequest,
+                                     diskOpt: Option[PersistentDisk],
+                                     nodepoolId: NodepoolLeoId,
+                                     ctx: AppContext): Either[Throwable, SaveApp] = {
+    val now = ctx.now
```
Qi77Qi, line 257: minor, pass in `now: Instant` directly?

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/service/MockKubernetesService.scala
```diff
@@ -0,0 +1,31 @@
...
+object MockKubernetesServiceInterp extends KubernetesService[IO] {
+  override def createApp(userInfo: UserInfo, googleProject: GoogleProject, appName: AppName, req: CreateAppRequest)(
+    implicit as: ApplicativeAsk[IO, AppContext]
+  ): IO[Unit] =
+    IO.unit
+
+  override def getApp(userInfo: UserInfo, googleProject: GoogleProject, appName: AppName)(
+    implicit as: ApplicativeAsk[IO, AppContext]
+  ): IO[GetAppResponse] =
+    IO(getAppResponse)
```
Qi77Qi: minor...use `IO.pure` to be explicit that `getAppResponse` is pure....`IO()` is lazy, which is unnecessary here

Qi77Qi (approved):
looks good overall... 👍  once the API change that we talked about on call

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
jdcanas: no, its needed for database constraint enforcement

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
jdcanas: right now, this is used to differentiate where the error occurred when returning errors to the UI. Currently, we handle it fairly simply by just saying "return all errors", and we don't even persistent any yet. I'd prefer punting on any complex handling until we've started to store errors in some way.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/KubernetesServiceInterp.scala
jdcanas: deep down in the call tree, in the DiskServiceInterp, we consolidate the request and the config such that defaults are used if the req has None in various places

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/KubernetesServiceInterp.scala
jdcanas: some of the exceptions print need the traceId from the ctx to print more helpful messages

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesServiceDbQueries.scala
jdcanas: we need to .toSet this, because the database queries do not GroupBy, and thus duplicates are returned on all tables due to the permutations SQL gives by default on joins

jdcanas:
jenkins  retest

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/leonardoServiceContractModel.scala
```diff
@@ -231,3 +234,62 @@ object GetRuntimeResponse {
     diskConfig
   )
 }
+
+final case class CreateAppRequest(kubernetesRuntimeConfig: Option[KubernetesRuntimeConfig],
+                                  appType: AppType,
+                                  diskConfig: Option[PersistentDiskRequest],
```
Qi77Qi, line 240: shouldn't this be a non optional?

Qi77Qi: also would be nice to move these contract data types to core..similar to https://github.com/DataBiosphere/leonardo/blob/develop/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/diskRoutesModels.scala...it's a lot more work to move them later

PR #1450, opened 2020-06-12
[no ticket] Handle error during zombie detection

rtitle:
I think this was causing issues in Back Leo on alpha. There are a couple of inactive dataproc clusters that were returning a 403 from Google for both `getClusterStatus` and `deleteCluster`. I think this error actually wasn't handled in the delete case and caused Back Leo to shut down [here](https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/Boot.scala#L252).

The fix: 
   * added error handling in `handleInactiveZombieRuntime`.
   * moved error handling _out_ of `isRuntimeActiveInGoogle` to the actual call site. Since this method is used in both active and inactive cases, we want different error handling behavior for each case.

---

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ZombieRuntimeMonitor.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ZombieRuntimeMonitor.scala
@@ -82,6 +82,7 @@ class ZombieRuntimeMonitor[F[_]: Parallel: ContextShift: Timer](
                         case Right(true) =>
                           F.pure(None)
                         case Right(false) =>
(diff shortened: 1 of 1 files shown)


jdcanas:
that's some mature zombie monitoring code there

jdcanas (approved):

Qi77Qi:
shall we also add a handleError wrapping `zombieCheck` for unexpected error...

rtitle:
Sure -- I think this PR addresses the specific case on alpha right now, but it's probably a good idea to handle unexpected errors too

Qi77Qi (approved):
thanks!..would be nice to if we can get alerts about these unexpected errros

PR #1457, opened 2020-06-16
IA-1759 Mount PD in GCE VM

gpcarr:
Adds  formatting and mounting PD to GCE VM so that user data persists. Adds `formattedBy` to `persistent_disk_table` 

---

In all cases:


http/src/main/resources/init-resources/gce-init.sh
```diff
@@ -153,6 +154,16 @@ if [[ "$RUNTIME_OPERATION" == 'creating' ]]; then
 
     mkdir -p /work
     mkdir -p /certs
+    # Mount persisent disk
+    export DISK_DEVICE_ID=$(lsblk -o name,serial | grep 'user-disk' | awk '{print $1}')
+    # Only format disk is it hasn't already been formatted
+    if [ "$IS_GCE_FORMATTED" == "false" ] ; then
+      mkfs.ext4 -m 0 -E lazy_itable_init=0,lazy_journal_init=0,discard /dev/${DISK_DEVICE_ID}
```
gpcarr, line 161: formatting options according to google suggestions: https://cloud.google.com/compute/docs/disks/optimizing-pd-performance#formatting_parameters

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20200609_add_persistent_disk_formattedBy.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="leonardo" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+    <changeSet logicalFilePath="leonardo" author="gabriela" id="add_persistent_disk_formattedBy">
+        <addColumn tableName="PERSISTENT_DISK">
+            <column name="formattedBy" type="VARCHAR(255)" defaultValue="NULL">
```
Qi77Qi: I think we were saying making this a ENUM type, similar to https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20200123_cluster_table_gce_refactor.xml#L21

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/diskModels.scala
```diff
@@ -91,3 +92,17 @@ object DiskType extends Enum[DiskType] {
...
   }
 }
+
+sealed trait FormattedBy extends EnumEntry with Product with Serializable {
+  def asString: String
+}
+object FormattedBy extends Enum[FormattedBy] {
+  val values = findValues
+
+  final case object Runtime extends FormattedBy {
```
Qi77Qi: shall we call this Gce? just to be more specific

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GceInterpreter.scala
```diff
@@ -70,6 +62,18 @@ class GceInterpreter[F[_]: Parallel: ContextShift: Logger](
                                                     config.gceConfig.zoneName,
                                                     params.runtimeConfig.machineType)
 
+      // Determine is user persisent disk has been formatted before
+      isGceFormatted <- params.runtimeConfig match {
```
Qi77Qi, line 74: I was thinking return `isGceFomatted` as part of the block below...since we're already doing this check in https://github.com/DataBiosphere/leonardo/pull/1457/files#diff-44f5fd773b99b0ea53e82a7b38b3d8e2R140

Qi77Qi (commented):
would be nice to have automation tests merged in along with this PR....but I guess it's fine since we have automation tests as a separate ticket explicitly.

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20200609_add_persistent_disk_formattedBy.xml
rtitle: +1 on using enum

http/src/main/resources/init-resources/gce-init.sh
```diff
@@ -153,6 +154,16 @@ if [[ "$RUNTIME_OPERATION" == 'creating' ]]; then
...
+    fi
+    mount -o discard,defaults /dev/${DISK_DEVICE_ID} /work
+    # Ensure peristent disk re-mounts if runtime stops and restarts
+    cp /etc/fstab /etc/fstab.backup
+    echo UUID=`blkid -s UUID -o value /dev/${DISK_DEVICE_ID}` /work ext4 discard,defaults,nofail 0 2 | tee -a /etc/fstab
```
rtitle, line 166: Cool - so this means we don't need any changes to `startup.sh`?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GceInterpreter.scala
rtitle: Think this is needed here in order to pass `isGceFormatted` to `RuntimeTemplateValuesConfig.fromCreateRuntimeParams`?

rtitle (approved):
Couple minor comments above, overall looks good!

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GceInterpreter.scala
Qi77Qi: yea it will...just moving things around a bit....so `RuntimeTemplateValuesConfig.fromCreateRuntimeParams` will be moved down as well...

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20200609_add_persistent_disk_formattedBy.xml
gpcarr: will update

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/diskModels.scala
gpcarr: yeah that makes sense, i can update

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GceInterpreter.scala
gpcarr: I was thinking to keep as is since we do all the templating first (defining `resourceConstraints`, `templateParams`, and `templateValues`) then build the disks and instance. if i moved it, i would also need to move the templating to happen after we build the disk but before we build the instance (since the instance `templateValues` when building it) and it seemed better to keep the logic flow the way it is

http/src/main/resources/init-resources/gce-init.sh
gpcarr: yep, i tested and validated that a PD gets mounted after restarting an instance

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20200609_add_persistent_disk_formattedBy.xml
gpcarr: `NULL` is an invalid default value for ENUM so looks like we should keep `VARCHAR(255)` since `formattedBy` is an optional field. All other enums use varchar except the one in the link above that Qi mentioned

Qi77Qi: NULL should be valid, see https://www.tutorialspoint.com/what-are-the-different-attributes-of-MySQL-ENUM-data-type#:~:text=NOT%20NULL%20%2D%20In%20ENUM%20type,while%20describing%20the%20ENUM%20column...



I tried

```
        <addColumn tableName="PERSISTENT_DISK">
            <column name="formattedBy" type="ENUM('GALAXY', 'GCE')">
                <constraints nullable="true"/>
            </column>
        </addColumn>
```
this should work

Qi77Qi (approved):

PR #1461, opened 2020-06-17
[no ticket] make persistent disk default size to 30G

Qi77Qi:
Set default user disk size to 20G instead of 500G....Old default is 50G in total for the one disk...

Tested in fiab

Per https://broadinstitute.slack.com/archives/GB1VAM14J/p1592423352248900

---

In all cases:

### DIFF
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -66,7 +66,7 @@ gce {
   ]
   runtimeDefaults {
     machineType = "n1-standard-4"
-    diskSize = 20 # This is default size for just user disk
+    diskSize = 30 # This is default size for just user disk
   }
   gceReservedMemory = 1g
   bootDiskSize = 50
@@ -427,7 +427,7 @@ zombieRuntimeMonitor {
 }
 
 persistentDisk {
-  defaultDiskSizeGB = 20
+  defaultDiskSizeGB = 30
   defaultDiskType = "pd-standard"
   defaultBlockSizeBytes = 4096
   zone = "us-central1-a"
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -66,7 +66,7 @@ gce {
   ]
   runtimeDefaults {
     machineType = "n1-standard-4"
-    diskSize = 50
+    diskSize = 20 # This is default size for just user disk
   }
   gceReservedMemory = 1g
   bootDiskSize = 50
@@ -427,7 +427,7 @@ zombieRuntimeMonitor {
 }
 
 persistentDisk {
-  defaultDiskSizeGB = 500
+  defaultDiskSizeGB = 20
   defaultDiskType = "pd-standard"
   defaultBlockSizeBytes = 4096
   zone = "us-central1-a"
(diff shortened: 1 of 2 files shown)


http/src/main/resources/reference.conf
```diff
@@ -66,7 +66,7 @@ gce {
   ]
   runtimeDefaults {
     machineType = "n1-standard-4"
-    diskSize = 50
+    diskSize = 20 # This is default size for just user disk
```
Qi77Qi: default size for user disk when user is not opting to use persistent disk

http/src/main/resources/reference.conf
```diff
@@ -392,7 +392,7 @@ zombieRuntimeMonitor {
 }
 
 persistentDisk {
-  defaultDiskSizeGB = 500
+  defaultDiskSizeGB = 20
```
Qi77Qi: default size for persistent disk

http/src/main/resources/reference.conf
jdcanas: Is this for what we've been calling the 'boot disk'? I thought the max set of images could be around ~30GB?

http/src/main/resources/reference.conf
jdcanas: It may be best to specify a bit higher of a default. I think individual files users often stream/interact with can be around 20GB

http/src/main/resources/reference.conf
Qi77Qi: no...boot disk is fixed to 50G because it needs to host the custom VM image we built. This disk is solely for storing user specific data....mounted in `/work`

kyuksel: can we name it `bootDiskSize` for clarity?

http/src/main/resources/reference.conf
Qi77Qi: are you thinking 20MB or 20GB for individual file?

http/src/main/resources/reference.conf
Qi77Qi: this is NOT for boot disk......there's a separate entry for `bootDiskSize`...

http/src/main/resources/reference.conf
jdcanas: 20GB. When doing the DRS work recently, a common use case a user told me about to test was copying over 20GB .cram files. In general, files containing genetic data can be very large

Qi77Qi: do they usually specify disk size in this case? the current disk size in total is 50G for gce VMs, which gives users about 37G usable space...

jdcanas: They can specify a size if this doesn't work I guess. I'm just thinking, in a typical exploratory analysis use-case, 20GB may be a bit small since individual files in the genomic space could exceed this size.

http/src/main/resources/reference.conf
kyuksel: My bad. How about `persistentDiskSize` or `userDiskSize`? It's so easy to falsely assume.

Qi77Qi: I think that can be confusing because `diskSize` is what we call it in request payload (not a new field), and this config value is the default value for that field if the field is not specified

kyuksel: Request is user-facing where we expose only the user disk (so no confusion there) whereas this config is internal where we have two types of disks so not sure if there should be an expectation of a one-to-one match between the config and the request parameter. Take it or leave it 😊

rtitle (approved):

PR #1464, opened 2020-06-18
IA-1966 Automation test for mounting disk in VM, data should persist

gpcarr:
Adds to existing test `"create runtime and attach an existing a persistent disk"` and validates that when a runtime is created with an existing PD, that the data on that PD persists

---

In all cases:

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/runtimes/RuntimeCreationPdSpec.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/runtimes/RuntimeCreationPdSpec.scala
@@ -1,23 +1,28 @@
 package org.broadinstitute.dsde.workbench.leonardo
 package runtimes
 
+import java.time.Instant
+
 import cats.effect.IO
 import org.broadinstitute.dsde.workbench.google2.Generators.genDiskName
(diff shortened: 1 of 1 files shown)


automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/runtimes/RuntimeCreationDiskSpec.scala
```diff
@@ -108,6 +110,7 @@ class RuntimeCreationDiskSpec
   "create runtime and attach an existing a persistent disk" in { googleProject =>
     val randomeName = randomClusterName
     val runtimeName = randomeName.copy(asString = randomeName.asString + "pd-spec") // just to make sure the test runtime name is unique
+    val runtimeWithDataName = randomeName.copy(asString = randomeName.asString + "pd-spec-data-persist")
```
Qi77Qi, line 113: `.copy(asString = randomeName.asString + "pd-spec-data-persist")` isn't actually needed btw :) I added it in line 112 when I was trying to figure out a weird bug..but `randomeName` itself should be able to generate unique names

Qi77Qi (approved):
👍  lgtm

akarukappadath (approved):

PR #1466, opened 2020-06-19
IA-2003  changes around formatted by

Qi77Qi:
We should check formattedBy before granting attaching disk permission as we discussed during walkthrough 2 weeks ago

---

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/PersistentDiskComponent.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/PersistentDiskComponent.scala
@@ -130,10 +130,29 @@ object persistentDiskQuery extends TableQuery(new PersistentDiskTable(_)) {
   def updateSize(id: DiskId, newSize: DiskSize, dateAccessed: Instant) =
     findByIdQuery(id).map(d => (d.size, d.dateAccessed)).update((newSize, dateAccessed))
 
-  def getIsGceFormatted(id: DiskId, dateAccessed: Instant)(implicit ec: ExecutionContext): DBIO[Option[FormattedBy]] =
+  def getFormattedBy(id: DiskId)(implicit ec: ExecutionContext): DBIO[Option[FormattedBy]] =
     findByIdQuery(id).map(_.formattedBy).result.headOption.map(_.flatten)
 
-  // TODO add other queries as needed
+  def isDiskAlreadyAttached(diskId: DiskId)(implicit ec: ExecutionContext): DBIO[Boolean] =
+    for {
+      formattedBy <- getFormattedBy(diskId)
+      r <- formattedBy match {
+        case None =>
+          for {
+            isAttachedToRuntime <- RuntimeConfigQueries.isDiskAttached(diskId)
+            isAttached <- if (isAttachedToRuntime) DBIO.successful(true)
+            else appQuery.isDiskAttached(diskId)
+          } yield isAttached
+        case Some(FormattedBy.Galaxy) =>
+          appQuery.filter(x => x.diskId.isDefined && x.diskId === diskId).length.result.map(_ > 0)
+        case Some(FormattedBy.GCE) =>
+          RuntimeConfigQueries.runtimeConfigs
+            .filter(x => x.persistentDiskId.isDefined && x.persistentDiskId === diskId)
+            .length
+            .result
+            .map(_ > 0)
+      }
+    } yield r
 
   private[db] def marshalPersistentDisk(disk: PersistentDisk): PersistentDiskRecord =
     PersistentDiskRecord(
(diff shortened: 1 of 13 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/KubernetesServiceInterp.scala
```diff
@@ -96,6 +96,7 @@ class LeoKubernetesServiceInterp[F[_]: Parallel](
           googleProject,
           userInfo,
           petSA,
+          FormattedBy.Galaxy,
```
gpcarr, line 99: why are we setting `FormattedBy` here? shouldn't we only update this when we know that the disk has been formatted ie when during monitoring we see that the runtime was created successfully? otherwise we might be setting it too early and during the startup script, it will think the disk is already formatted

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeServiceInterp.scala
```diff
@@ -110,6 +110,7 @@ class RuntimeServiceInterp[F[_]: Parallel](config: RuntimeServiceConfig,
                                                       googleProject,
                                                       userInfo,
                                                       petSA,
+                                                      FormattedBy.GCE,
```
gpcarr, line 113: same question as above

gpcarr: oh wait, i see what you're doing after reading the rest of the diff. you're not actually updating the DB 👍

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/PersistentDiskComponent.scala
```diff
@@ -130,10 +130,29 @@ object persistentDiskQuery extends TableQuery(new PersistentDiskTable(_)) {
...
+      formattedBy <- getFormattedBy(diskId)
+      r <- formattedBy match {
+        case None =>
+          for {
+            isAttachedToRuntime <- RuntimeConfigQueries.isDiskAttached(diskId)
+            isAttached <- if (isAttachedToRuntime) DBIO.successful(true)
+            else appQuery.isDiskAttached(diskId)
+          } yield isAttached
+        case Some(FormattedBy.Galaxy) =>
+          appQuery.filter(x => x.diskId.isDefined && x.diskId === diskId).length.result.map(_ > 0)
```
kyuksel: Wouldn't this return `true` if we had done the following?
- created a disk and attached it to a Galaxy app
- deleted the app but kept the disk

In other words, is it correct to assume that if a disk was formatted by Galaxy, it is attached?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/PersistentDiskComponent.scala
```diff
@@ -130,10 +130,29 @@ object persistentDiskQuery extends TableQuery(new PersistentDiskTable(_)) {
   def updateSize(id: DiskId, newSize: DiskSize, dateAccessed: Instant) =
     findByIdQuery(id).map(d => (d.size, d.dateAccessed)).update((newSize, dateAccessed))
 
-  def getIsGceFormatted(id: DiskId, dateAccessed: Instant)(implicit ec: ExecutionContext): DBIO[Option[FormattedBy]] =
+  def getFormattedBy(id: DiskId)(implicit ec: ExecutionContext): DBIO[Option[FormattedBy]] =
     findByIdQuery(id).map(_.formattedBy).result.headOption.map(_.flatten)
 
-  // TODO add other queries as needed
+  def isDiskAlreadyAttached(diskId: DiskId)(implicit ec: ExecutionContext): DBIO[Boolean] =
```
kyuksel: Nit: `Already` in the method name may be redundant

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeServiceInterp.scala
```diff
@@ -667,6 +668,7 @@ object RuntimeServiceInterp {
     googleProject: GoogleProject,
     userInfo: UserInfo,
     serviceAccount: WorkbenchEmail,
+    willByUsedBy: FormattedBy,
```
kyuksel:
```suggestion
    willBeUsedBy: FormattedBy,
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/PersistentDiskComponent.scala
Qi77Qi: No. If we delete app, we "dettach" (meaning deleting the diskId reference from either app table or cluster table)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeServiceInterp.scala
Qi77Qi: yea, this param is called `willBeUsedBy`, so we're checking if `willBeUsedBy` matches are `formattedBy` in DB

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/PersistentDiskComponent.scala
kyuksel: 👍

kyuksel (approved):
LGTM 👍

Qi77Qi:
flaky test
```
[info] - should aggregate all sub tables on get, and clean up all tables on delete *** FAILED ***
```

cc @jdcanas , this is still happening

Qi77Qi:
fiab-start failure...jenkins retest

PR #1468, opened 2020-06-19
[No ticket] sam dao stubbing and fixing test

jdcanas:
fix test that doesn't always work on jenkins

---

In all cases:

### DIFF
--- a/http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/dao/MockSamDAO.scala
+++ b/http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/dao/MockSamDAO.scala
@@ -5,7 +5,7 @@ import java.util.UUID
 
 import cats.effect.IO
 import cats.mtl.ApplicativeAsk
-import org.broadinstitute.dsde.workbench.leonardo.SamResource.{PersistentDiskSamResource, RuntimeSamResource}
+import org.broadinstitute.dsde.workbench.leonardo.SamResource.{AppSamResource, PersistentDiskSamResource, RuntimeSamResource}
 import org.broadinstitute.dsde.workbench.leonardo.dao.MockSamDAO._
 import org.broadinstitute.dsde.workbench.model.google.GoogleProject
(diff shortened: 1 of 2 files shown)


http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/dao/MockSamDAO.scala
```diff
@@ -75,7 +82,8 @@ class MockSamDAO extends SamDAO[IO] {
         IO(runtimes += (r, userEmailToAuthorization(creatorEmail)) -> runtimeActions)
       case r: PersistentDiskSamResource =>
         IO(persistentDisks += (r, userEmailToAuthorization(creatorEmail)) -> diskActions)
-      case _ => IO(throw new Exception("Invalid resource to create"))
+      case r: AppSamResource => IO.unit //TODO
+      case _                 => IO(throw new Exception("Invalid resource to create"))
```
Qi77Qi: let's not put `_` here, we lose the benefit of using `sealed` this way, because we do want to get warnings if we don't exhaust the cases properly

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/dao/MockSamDAO.scala
```diff
@@ -65,6 +70,8 @@ class MockSamDAO extends SamDAO[IO] {
         IO.pure(projectOwners.get(authHeader).map(_.toList).getOrElse(List.empty)).map(_.asInstanceOf[List[A]])
       case SamResourceType.PersistentDisk =>
         IO.pure(diskCreators.get(authHeader).map(_.toList).getOrElse(List.empty)).map(_.asInstanceOf[List[A]])
+      case SamResourceType.App =>
+        IO.pure(List[A]()) //TODO
```
Qi77Qi, line 74: can you add what some words indicating what the TODO is? simply something like "fix me" will do compared to nothing

Qi77Qi (approved):

PR #1471, opened 2020-06-23
Jc k8 proxy

jdcanas:
I wanted to throw this up before I disappear. 

**This is a draft. It is not working and has bugs. Has not been testing. Existing tests not passing**


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/ProxyRoutes.scala
```diff
@@ -53,7 +59,7 @@ class ProxyRoutes(proxyService: ProxyService, corsSupport: CorsSupport)(
                 }
               }
             }
-          } ~
+          } ~ path("jupyter" | "rstudio") {
```
jdcanas, line 62: this doesn't actually work. How its designed, /jupyter/x will not get picked up by anything. I need to add another block after this to capture `jupyter/*`

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/api/ProxyRoutesSpec.scala
```diff
@@ -69,21 +69,16 @@ class ProxyRoutesSpec
...
   val prefix = "proxy"
 
-  "ProxyRoutes" should s"listen on /$prefix/{project}/{name}/... ($prefix)" in {
-    Get(s"/$prefix/$googleProject/$clusterName").addHeader(Cookie(tokenCookie)) ~> proxyRoutes.route ~> check {
-      handled shouldBe true
-      status shouldEqual StatusCodes.OK
-      validateCors()
-    }
-    Get(s"/$prefix/$googleProject/$clusterName/foo").addHeader(Cookie(tokenCookie)) ~> proxyRoutes.route ~> check {
+  "ProxyRoutes" should s"listen on /$prefix/{project}/{name}/jupyter)" in {
```
jdcanas, line 72: So, I changed the proxy route so `/project/name/x` won't match to jupyter, and we instead use 'x' to determine if it should go to jupyter/rstudio (existing proxy service) or if its a service that should go to kubernetes. These tests aren't working, because the proxy route path matching is wrong currently. I will most likely refactor these tests a bit as well, so ignore everything changed in this file

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/ProxyRoutes.scala
rtitle: `pathPrefix` might help here

PR #1477, opened 2020-06-24
[IA-1952] welderRegistry field for newest welder image from desired container registry

akarukappadath:

---

In all cases:


core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodec.scala
```diff
@@ -205,6 +206,8 @@ object JsonCodec {
   implicit val ipDecoder: Decoder[IP] = Decoder.decodeString.map(IP)
   implicit val containerImageDecoder: Decoder[ContainerImage] =
     Decoder.decodeString.emap(s => ContainerImage.fromString(s).toRight(s"invalid container image ${s}"))
+  implicit val containerRegistryDecoder: Decoder[ContainerRegistry] =
+    Decoder.decodeString.emap(s => ContainerRegistry.withNameOption(s).toRight(s"Unsupported container registry ${s}"))
```
Qi77Qi: can we use `withNameInsensitiveOption`

http/src/main/resources/reference.conf
```diff
@@ -195,7 +195,8 @@ gke {
 }
 
 image {
-  welderImage = "us.gcr.io/broad-dsp-gcr-public/welder-server:7ae1a62"
+  welderGcrImage = "us.gcr.io/broad-dsp-gcr-public/welder-server:7ae1a62"
```
Qi77Qi: I thought we're doing something like
```
welderGcrImage = "us.gcr.io/broad-dsp-gcr-public/welder-server"
welderDockerHubImage = "broadinstitute/welder-server"
welderHash = "7ae1a62"
```
This way, the jenkins job that updates this hash only needs to look for `welderHash`

akarukappadath, line 198: The other option is to have three config values here: 
```
welderGcrUri = "us.gcr.io/broad-dsp-gcr-public/welder-server"
welderDockerHubUri = "broadinstitute/welder-server"
welderHash = "7ae1a62"
```
The benefit of doing it this way is that the Jenkins job that updates this hash would only have one value to change. However, we would need to construct the relevant image URI with the hash in Leo, and it adds another value to a growing `reference.conf` - should we be doing that instead?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/BaseRuntimeInterpreter.scala
```diff
@@ -97,8 +97,8 @@ abstract private[util] class BaseRuntimeInterpreter[F[_]: Async: ContextShift: L
       val labelFound = config.welderConfig.updateWelderLabel.exists(runtime.labels.contains)
 
       val imageChanged = runtime.runtimeImages.find(_.imageType == Welder) match {
-        case Some(welderImage) if welderImage.imageUrl != config.imageConfig.welderImage => true
-        case _                                                                           => false
+        case Some(welderImage) if welderImage.imageUrl != config.imageConfig.welderGcrImage => true
```
Qi77Qi: If welderImage is from dockerhub, it'll never equal to `welderGceImage`, and we shouldnt' try to update the image...

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/runtimeRoutesModels.scala
```diff
@@ -84,6 +85,7 @@ final case class CreateRuntime2Request(labels: LabelMap,
                                        autopauseThreshold: Option[FiniteDuration],
                                        defaultClientId: Option[String],
                                        toolDockerImage: Option[ContainerImage],
+                                       welderRegistry: Option[ContainerRegistry],
```
Qi77Qi: can you add a comment about removing `welderDockerImage` in the future

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/BaseRuntimeInterpreter.scala
akarukappadath: Right now, the label on which we update welder is `saturnVersion` - until we change that we won't be supporting updating welder from Dockerhub anyways. AoU also didn't seem to care much about having welder auto-updated

Qi77Qi: I realize that's the assumption here...but once aou start using the new flag, welder image URL will be something like "broadinstitute/welder-server:7ae1a62", so this condition will always be true, but we should have it as false cuz we don't want to update it in this case

akarukappadath: It's the same as it is now though. We won't be updating the image either way because both the label needs to be there too (see line 104)

akarukappadath:
jenkins retest

akarukappadath:
jenkins retest

akarukappadath:
jenkins retest

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/containerModels.scala
```diff
@@ -1,51 +1,47 @@
 package org.broadinstitute.dsde.workbench.leonardo
 
 import ca.mrvisser.sealerate
+import enumeratum.{Enum, EnumEntry}
 
 import scala.util.matching.Regex
 
 /** Container registry, e.g. GCR, Dockerhub */
-sealed trait ContainerRegistry extends Product with Serializable {
+sealed trait ContainerRegistry extends EnumEntry with Product with Serializable {
```
Qi77Qi, line 9: this data type already has `allRegistries`, why are we adding `EnumEntry`?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/BaseRuntimeInterpreter.scala
```diff
@@ -97,8 +97,8 @@ abstract private[util] class BaseRuntimeInterpreter[F[_]: Async: ContextShift: L
       val labelFound = config.welderConfig.updateWelderLabel.exists(runtime.labels.contains)
 
       val imageChanged = runtime.runtimeImages.find(_.imageType == Welder) match {
-        case Some(welderImage) if welderImage.imageUrl != config.imageConfig.welderImage => true
-        case _                                                                           => false
+        case Some(welderImage) if welderImage.hash != Option(config.imageConfig.welderHash) => true
```
Qi77Qi: use `Some`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/BaseRuntimeInterpreter.scala
```diff
@@ -121,12 +121,22 @@ abstract private[util] class BaseRuntimeInterpreter[F[_]: Async: ContextShift: L
...
+      newWelderImageUrl = runtime.runtimeImages
+        .find(_.imageType == Welder)
+        .map(x =>
+          x.registry match {
+            case Some(ContainerRegistry.GCR)       => config.imageConfig.welderGcrImage.imageUrl
+            case Some(ContainerRegistry.DockerHub) => config.imageConfig.welderDockerHubImage.imageUrl
+            case None                              => throw new Exception(s"Unable to update Welder: registry for ${x.imageUrl} not parsable")
+          }
+        )
+        .getOrElse(throw new Exception(s"Unable to update welder because current welder image is not available"))
```
Qi77Qi: can we not throw here and use `F.fromEither` instead

Qi77Qi (commented):
merge conflicts and you'll need to update https://github.com/broadinstitute/dsp-jenkins/blob/master/jobs/fc-jenkins/WelderUpdateHash.groovy#L32 as well

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/containerModels.scala
akarukappadath: I added it so that I could use `withNameInsensitiveOption` for the Decoder and `entryName` for the Encoder. Is there a drawback to using `EnumEntry`?

Qi77Qi: it's just duplicate since we're using sealed trait there as well...for `withNameInsensitiveOption`, we can use all lowercase in the string representation, and in decoder, just do `_.toLowerCase`...guess I don't feel strongly here, take or leave it

Qi77Qi (approved):
👍

PR #1478, opened 2020-06-24
IA-2019 support deleteDisk query param for deleteRuntime API

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-2019

---

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/RuntimeRoutes.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/RuntimeRoutes.scala
@@ -224,7 +224,7 @@ class RuntimeRoutes(runtimeService: RuntimeService[IO], userInfoDirectives: User
       deleteDisk = params
         .get("deleteDisk")
         .map(s => if (s == "true") true else false)
-        .getOrElse(true) //if `deleteDisk` is explicitly set to true, then we delete disk; otherwise, we don't
+        .getOrElse(false) //if `deleteDisk` is explicitly set to true, then we delete disk; otherwise, we don't
       request = DeleteRuntimeRequest(userInfo, googleProject, runtimeName, deleteDisk)
       _ <- runtimeService.deleteRuntime(request)
     } yield StatusCodes.Accepted: ToResponseMarshallable
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/RuntimeRoutes.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/RuntimeRoutes.scala
@@ -23,7 +23,11 @@ import org.broadinstitute.dsde.workbench.leonardo.SamResource.RuntimeSamResource
 import org.broadinstitute.dsde.workbench.leonardo.api.CookieSupport
 import org.broadinstitute.dsde.workbench.leonardo.http.api.LeoRoutesJsonCodec.dataprocConfigDecoder
 import org.broadinstitute.dsde.workbench.leonardo.http.api.RuntimeRoutes._
-import org.broadinstitute.dsde.workbench.leonardo.http.service.{GetRuntimeResponse, RuntimeService}
+import org.broadinstitute.dsde.workbench.leonardo.http.service.{
+  DeleteRuntimeRequest,
+  GetRuntimeResponse,
+  RuntimeService
+}
 import org.broadinstitute.dsde.workbench.model.google.GoogleProject
 import org.broadinstitute.dsde.workbench.model.{TraceId, UserInfo}
 
@@ -114,15 +118,18 @@ class RuntimeRoutes(runtimeService: RuntimeService[IO], userInfoDirectives: User
                         }
                       } ~
                       delete {
-                        traceRequestForService(serviceData) { span =>
-                          complete(
-                            deleteRuntimeHandler(
-                              userInfo,
-                              googleProject,
-                              runtimeName,
-                              span
+                        parameterMap { params =>
+                          traceRequestForService(serviceData) { span =>
+                            complete(
+                              deleteRuntimeHandler(
+                                userInfo,
+                                googleProject,
+                                runtimeName,
+                                params,
+                                span
+                              )
                             )
-                          )
+                          }
                         }
                       }
                   } ~
@@ -210,10 +217,16 @@ class RuntimeRoutes(runtimeService: RuntimeService[IO], userInfoDirectives: User
   private[api] def deleteRuntimeHandler(userInfo: UserInfo,
                                         googleProject: GoogleProject,
                                         runtimeName: RuntimeName,
+                                        params: Map[String, String],
                                         span: Span): IO[ToResponseMarshallable] = {
     val res = for {
       implicit0(ctx: ApplicativeAsk[IO, AppContext]) <- AppContext.lift[IO](Some(span))
-      _ <- runtimeService.deleteRuntime(userInfo, googleProject, runtimeName)
+      deleteDisk = params
+        .get("deleteDisk")
+        .map(s => if (s == "true") true else false)
+        .getOrElse(true) //if `deleteDisk` is explicitly set to true, then we delete disk; otherwise, we don't
+      request = DeleteRuntimeRequest(userInfo, googleProject, runtimeName, deleteDisk)
+      _ <- runtimeService.deleteRuntime(request)
     } yield StatusCodes.Accepted: ToResponseMarshallable
 
     spanResource[IO](span, "deleteRuntime")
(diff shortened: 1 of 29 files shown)


xuscrbw (approved):
Looks good. Would like someone to verify the PubsubSubscriber changes just in case

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/RuntimeRoutes.scala
```diff
@@ -210,10 +217,16 @@ class RuntimeRoutes(runtimeService: RuntimeService[IO], userInfoDirectives: User
...
                                         runtimeName: RuntimeName,
+                                        params: Map[String, String],
                                         span: Span): IO[ToResponseMarshallable] = {
     val res = for {
       implicit0(ctx: ApplicativeAsk[IO, AppContext]) <- AppContext.lift[IO](Some(span))
-      _ <- runtimeService.deleteRuntime(userInfo, googleProject, runtimeName)
+      deleteDisk = params
+        .get("deleteDisk")
+        .map(s => if (s == "true") true else false)
+        .getOrElse(true) //if `deleteDisk` is explicitly set to true, then we delete disk; otherwise, we don't
```
gpcarr: why do we say `.getOrElse(true)`, if `deleteDisk` is not in `params` then wouldn't we want to set `deleteDisk` to false so we don't delete the disk?

Qi77Qi: ah nice catch 👍

Qi77Qi:
patchspec passes locally..jenkins retest

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/api/HttpRoutesSpec.scala
```diff
@@ -130,6 +139,57 @@ class HttpRoutesSpec
...
+    }
+  }
+
+  it should "not delete a runtime and disk if deleteDisk is false" in {
+    val runtimeService = new BaseMockRuntimeServiceInterp {
+      override def deleteRuntime(deleteRuntimeRequest: DeleteRuntimeRequest)(
+        implicit as: ApplicativeAsk[IO, AppContext]
+      ): IO[Unit] = IO {
+        val expectedDeleteRuntime =
+          DeleteRuntimeRequest(timedUserInfo, GoogleProject("googleProject1"), RuntimeName("runtime1"), true)
```
akarukappadath: should `deleteDisk` here be `false`?

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/api/HttpRoutesSpec.scala
```diff
@@ -130,6 +139,57 @@ class HttpRoutesSpec
...
+        deleteRuntimeRequest shouldBe expectedDeleteRuntime
+      }
+    }
+    val routes = fakeRoutes(runtimeService)
+    Delete("/api/google/v1/runtimes/googleProject1/runtime1?deleteDisk=true") ~> routes.route ~> check {
```
akarukappadath: same as above - shouldn't this `deleteDisk` be `false`?

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/api/HttpRoutesSpec.scala
```diff
@@ -130,6 +139,57 @@ class HttpRoutesSpec
...
+      status shouldEqual StatusCodes.Accepted
+      validateRawCookie(header("Set-Cookie"))
+    }
+  }
+
+  it should "not delete a runtime and disk if deleteDisk is false" in {
```
akarukappadath: could we rename this test to "should only delete runtime if deleteDisk is false" or "should not delete disk if deleteDisk is false"?

akarukappadath (changes requested):

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/api/HttpRoutesSpec.scala
Qi77Qi: nice catch 👍

akarukappadath (approved):

Qi77Qi:
jenkins retest

rtitle (approved):

rtitle:
Looks good.. btw `gcloud` has similar functionality: you can pass a `--delete-disks` flag to `gcloud compute instances delete`.

Qi77Qi:
@rtitle good point...I didn't find an option on `deleteRuntime` call directly, but maybe if we set disk to auto delete before making delete call, it'll work...I'll give it a try

rtitle:
I think the current approach is fine, I didn’t see a —delete-disk option in the GCE API either. I suspect the gcloud cli does something similar where it first deletes the disks, then the instance.

Qi77Qi:
@rtitle  setting the flag `auto-delete` to true works...see https://github.com/broadinstitute/workbench-libs/pull/323, tested in console

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpSamDAO.scala
```diff
@@ -123,19 +123,21 @@ class HttpSamDAO[F[_]: Effect](httpClient: Client[F], config: HttpSamDaoConfig,
       _ <- logger.info(
         s"${traceId} | creating ${resource.resourceType.asString} resource in sam for ${googleProject}/${resource.resourceId}"
       )
-      _ <- httpClient.fetch[Unit](
-        Request[F](
-          method = Method.POST,
-          uri = config.samUri
-            .withPath(s"/api/resources/v1/${resource.resourceType.asString}/${resource.resourceId}"),
-          headers = Headers.of(authHeader)
+      _ <- httpClient
```
Qi77Qi, line 126: `fetch` is deprecated in latest http4s

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BaseCloudServiceRuntimeMonitor.scala
```diff
@@ -139,8 +141,7 @@ abstract class BaseCloudServiceRuntimeMonitor[F[_]] {
     runtimeAndRuntimeConfig: RuntimeAndRuntimeConfig,
     dataprocInstances: Set[DataprocInstance], //only applies to dataproc
     message: Option[String],
-    ip: Option[IP] = None,
-    toolsToCheck: List[RuntimeImageType] = List.empty // If empty, then check all tools; if defined, then only check specified tools
+    ip: Option[IP] = None
```
Qi77Qi, line 144: previously `toolsToCheck` is an optimization that allows us not to re-check a tool if we've already checked it, but now that we're using `streamFUntilDone`, we don't need this anymore, which simplifies the monitoring logic

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeServiceInterp.scala
```diff
@@ -237,7 +239,7 @@ class RuntimeServiceInterp[F[_]: Parallel](config: RuntimeServiceConfig,
       _ <- if (runtime.asyncRuntimeFields.isDefined) {
         clusterQuery.updateClusterStatus(runtime.id, RuntimeStatus.PreDeleting, ctx.now).transaction >> publisherQueue
           .enqueue1(
-            DeleteRuntimeMessage(runtime.id, Some(ctx.traceId))
+            DeleteRuntimeMessage(runtime.id, req.deleteDisk, Some(ctx.traceId))
```
rtitle, line 242: I thought we were going to detach the disk in front Leo? Or is that a separate PR?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GceInterpreter.scala
```diff
@@ -303,9 +303,12 @@ class GceInterpreter[F[_]: Parallel: ContextShift: Logger](
...
     params: DeleteRuntimeParams
   )(implicit ev: ApplicativeAsk[F, TraceId]): F[Option[Operation]] =
-    if (params.asyncRuntimeFields.isDefined)
+    if (params.isAsyncRuntimeFields)
       googleComputeService
-        .deleteInstance(params.googleProject, config.gceConfig.zoneName, InstanceName(params.runtimeName.asString))
+        .deleteInstance(params.googleProject,
+                        config.gceConfig.zoneName,
+                        InstanceName(params.runtimeName.asString),
+                        params.autoDeletePersistentDisk.to[Set])
```
rtitle, line 311: I see, so wb-libs is calling `instanceClient.setDiskAutoDeleteInstance` for each disk then deleting the instance

rtitle (approved):
LGTM

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeServiceInterp.scala
Qi77Qi: yea, that'll be a separate PR. This is just to support `deleteDisk` flag

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GceInterpreter.scala
Qi77Qi: yep...

Qi77Qi:
going to merge this PR as is and bump wb-libs PR in the PR for detaching disk

PR #1483, opened 2020-06-29
[No ticket][IA-1891] make exception messaging more helpful and refactor svc acct

jdcanas:
Was encountering unhelpful internal server errors while playing around with the API in a FIAB

### DIFF
--- a/automation/hotswap.sh
+++ b/automation/hotswap.sh
@@ -14,12 +14,12 @@ fi
 FIAB=$1
 ENV=${2:-dev}
 
-printf "Generating the Leo jar. This will fail silently here if you have any compilation errors, or print a success message otherwise\n\n"
+printf "Generating the Leo jar...\n\n"
 # Example: /Users/qi/workspace/leonardo/http/target/scala-2.12/http-assembly-0.1-437ee4a9-SNAPSHOT.jar
 sbt -Dsbt.log.noformat=true "project http" clean assembly
 LEO_JAR_PATH=$(ls http/target/scala-2.12/http*)
 
-printf "Jar successfully generated."
+printf "\n\nJar successfully generated."
 
 LEO_JAR_NAME=$(basename $LEO_JAR_PATH)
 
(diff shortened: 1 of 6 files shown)


automation/hotswap.sh
```diff
@@ -14,12 +14,12 @@ fi
 FIAB=$1
 ENV=${2:-dev}
 
-printf "Generating the Leo jar. This will fail silently here if you have any compilation errors, or print a success message otherwise\n\n"
```
jacmrob, line 17: does this no longer silently fail?

jacmrob (commented):
The config stuff LGTM, you probably want more experienced eyes on the scala stuff

automation/hotswap.sh
jdcanas: yeah,  it was fixed a while ago but the message never updated. Also, must have fat fingered you as a reviewer, sorry 😄

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesServiceDbQueries.scala
```diff
@@ -43,17 +44,16 @@ object KubernetesServiceDbQueries {
...
-        )
-      )
       eitherClusterOrError <- clusterOpt match {
         case Some(cluster) =>
           cluster.status match {
-            case KubernetesClusterStatus.Precreating  => failure
-            case KubernetesClusterStatus.Provisioning => failure
-            case _                                    => DBIO.successful(SaveClusterResult(cluster, true))
+            case s if KubernetesClusterStatus.creatingStatuses contains s =>
+              DBIO.failed(
```
rtitle, line 51: Minor, but I wonder if this should be handled at the service layer (e.g. `LeoKubernetesServiceInterp`) instead of the DB layer since it feels more like business logic instead of something that should be enforced in the DB.

rtitle (approved):

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesServiceDbQueries.scala
jdcanas: Yeah, I can see the argument for it. I view these `ServiceDbQueries` objects as database business logic for the service, so I like having the low-level logic here to keep the flow in the ServiceInterp simplified

PR #1484, opened 2020-06-29
[IA-1986] Update attached PD when updating runtime

akarukappadath:
1. Fix a bug in existing update disk size where we're updating boot disk size
2. Fix a bug where update disk size doesn't work because runtime is not restarted.
3. Add automation tests to validate update disk size works for gce, gce with pd, dataproc

---

In all cases:

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/runtimes/RuntimePatchSpec.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/runtimes/RuntimePatchSpec.scala
@@ -5,21 +5,28 @@ import cats.effect.IO
 import cats.implicits._
 import org.broadinstitute.dsde.workbench.DoneCheckable
 import org.broadinstitute.dsde.workbench.auth.AuthToken
-import org.broadinstitute.dsde.workbench.google2.{streamFUntilDone, MachineTypeName}
+import org.broadinstitute.dsde.workbench.google2.{streamFUntilDone, DiskName, MachineTypeName}
 import org.broadinstitute.dsde.workbench.leonardo.GPAllocFixtureSpec.gpallocProjectKey
 import org.broadinstitute.dsde.workbench.leonardo.LeonardoApiClient._
-import org.broadinstitute.dsde.workbench.leonardo.http.RuntimeConfigRequest
+import org.broadinstitute.dsde.workbench.leonardo.http.{
+  PersistentDiskRequest,
+  RuntimeConfigRequest,
+  UpdateRuntimeConfigRequest,
+  UpdateRuntimeRequest
+}
+import org.broadinstitute.dsde.workbench.leonardo.notebooks.{NotebookTestUtils, Python3}
 import org.broadinstitute.dsde.workbench.model.google.GoogleProject
 import org.broadinstitute.dsde.workbench.service.util.Tags
 import org.http4s.headers.Authorization
 import org.http4s.{AuthScheme, Credentials}
 import org.scalatest.{DoNotDiscover, Outcome}
 import org.scalatest.freespec.FixtureAnyFreeSpec
+
 import scala.concurrent.duration._
 
 // extending fixture.Freespec just so we can use `taggedAs`. Not really need fixture for anything
 @DoNotDiscover
-class RuntimePatchSpec extends FixtureAnyFreeSpec with LeonardoTestUtils with LeonardoTestSuite {
+class RuntimePatchSpec extends FixtureAnyFreeSpec with LeonardoTestUtils with LeonardoTestSuite with NotebookTestUtils {
   implicit val ronToken: AuthToken = ronAuthToken
   implicit val auth: Authorization = Authorization(Credentials.Token(AuthScheme.Bearer, ronCreds.makeAuthToken().value))
 
@@ -31,16 +38,22 @@ class RuntimePatchSpec extends FixtureAnyFreeSpec with LeonardoTestUtils with Le
     // create a new GCE runtime
     val runtimeName = randomClusterName
 
-    val newMasterMachineType = "n1-standard-2"
-    val newRuntimeConfig = UpdateRuntimeConfigRequestCopy.GceConfig(
-      Some(newMasterMachineType),
-      None
-    )
+    val newMasterMachineType = MachineTypeName("n1-standard-2")
+    val newDiskSize = DiskSize(20)
+    val updateRuntimeRequest = UpdateRuntimeRequest(Some(
+                                                      UpdateRuntimeConfigRequest.GceConfig(
+                                                        Some(newMasterMachineType),
+                                                        Some(newDiskSize)
+                                                      )
+                                                    ),
+                                                    true,
+                                                    None,
+                                                    None)
     val createRuntimeRequest = defaultCreateRuntime2Request.copy(
       runtimeConfig = Some(
         RuntimeConfigRequest.GceConfig(
           Some(MachineTypeName("n1-standard-4")),
-          None
+          Some(DiskSize(10))
         )
       )
     )
@@ -54,13 +67,7 @@ class RuntimePatchSpec extends FixtureAnyFreeSpec with LeonardoTestUtils with Le
 
       for {
         _ <- createRuntimeWithWait(googleProject, runtimeName, createRuntimeRequest)
-        _ <- IO {
-          Leonardo.cluster.updateRuntime(
-            googleProject,
-            runtimeName,
-            request = UpdateRuntimeRequestCopy(Some(newRuntimeConfig), true, None, None)
-          )
-        }
+        _ <- updateRuntime(googleProject, runtimeName, updateRuntimeRequest)
         _ <- testTimer.sleep(30 seconds) //We need this because DB update happens in subscriber for update API.
         ioa = LeonardoApiClient.getRuntime(googleProject, runtimeName)
         getRuntimeResult <- ioa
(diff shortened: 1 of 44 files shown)


akarukappadath:
jenkins retest

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeServiceInterp.scala
```diff
@@ -530,6 +532,53 @@ class RuntimeServiceInterp[F[_]: Parallel](config: RuntimeServiceConfig,
...
+    } yield msg
+
+  private[service] def processUpdateGceWithPdConfigRequest(req: UpdateRuntimeConfigRequest.GceConfig,
+                                                           allowStop: Boolean,
+                                                           runtime: Runtime,
+                                                           gceConfig: RuntimeConfig.GceWithPdConfig,
+                                                           traceId: TraceId): F[Option[UpdateRuntimeMessage]] =
+    for {
+      // should machine type be updated?
+      targetMachineType <- traverseIfChanged(req.updatedMachineType, gceConfig.machineType) { mt =>
```
Qi77Qi: can you make this into a function so that it can be reused in `processUpdateGceConfigRequest`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeServiceInterp.scala
```diff
@@ -530,6 +532,53 @@ class RuntimeServiceInterp[F[_]: Parallel](config: RuntimeServiceConfig,
...
+          Async[F].raiseError[(MachineTypeName, Boolean)](RuntimeMachineTypeCannotBeChangedException(runtime))
+        else Async[F].pure((mt, true))
+      }
+      // should disk size be updated?
+      diskId <- F.fromEither(
+        gceConfig.persistentDiskId.toRight(RuntimeDiskNotFound(runtime.googleProject, runtime.runtimeName, traceId))
+      )
+      diskOpt <- persistentDiskQuery.getById(diskId).transaction
+      disk <- F.fromEither(diskOpt.toRight(DiskNotFoundByIdException(diskId, traceId)))
+      targetDiskSize <- traverseIfChanged(req.updatedDiskSize, disk.size) { d =>
```
Qi77Qi: while you're in here, can you test to see if disk size change require restart? (there was a bug ticket that updating disk size didn't take effect until machine is restarted

akarukappadath: can you elaborate on the bug? I just tried updating a disk size and it looks like it went through fine, but was the machine unable to use the disk?

Qi77Qi: https://broadworkbench.atlassian.net/browse/IA-2015 is the ticket...

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BaseCloudServiceRuntimeMonitor.scala
```diff
@@ -111,6 +111,7 @@ abstract class BaseCloudServiceRuntimeMonitor[F[_]] {
             _ <- persistentDiskQuery
               .updateStatusAndIsFormatted(diskId, DiskStatus.Ready, FormattedBy.GCE, now)
               .transaction
+            _ <- persistentDiskQuery.updateNeedsResize(diskId, false).transaction
```
Qi77Qi: why is this needed?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -528,6 +533,65 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift](
...
+            ctx.now
+          )
+        }
+        _ <- runtimeConfig.cloudService
+          .process(runtime.id, RuntimeStatus.Starting)
+          .compile
+          .drain
+      } yield ()
+      _ <- asyncTasks.enqueue1(
+        Task(ctx.traceId, task, Some(logError(runtime.projectNameString, "restarting runtime")), ctx.now)
```
Qi77Qi: we don't have to run this async to `restartRuntime`....reason why we used asyncTaskProcessor in a main message handler is because we want this whole waiting happen async to message acknowledgement (we don't want to wait say 20 min before acking a message, cuz things get weird with pubsub in that case)...

Here the whole pollOperation is already async to the message acking, so you don't need to run ilne 556 async to `restartRuntime` anymore

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -514,7 +515,11 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift](
           config.persistentDiskMonitorConfig.create.maxAttempts,
           None
         )(
-          nowInstant.flatMap(now => persistentDiskQuery.updateSize(msg.diskId, msg.newSize, now).transaction[F]).void,
+          for {
+            _ <- nowInstant.flatMap(now => persistentDiskQuery.updateSize(msg.diskId, msg.newSize, now).transaction[F])
+            _ <- persistentDiskQuery.updateNeedsResize(msg.diskId, true).transaction[F]
+            _ <- msg.restartRuntimeId.traverse(restartRuntime(_))
```
Qi77Qi: is this to account for when a `updateDisk` api called..if the disk is attached to a runtime, then we try to stop the attached runtime if it happens to be running? in other words, `restartRuntimeId` only exists when there's a runtime associated with the disk, and the runtime happens to be running?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeServiceInterp.scala
```diff
@@ -598,6 +599,53 @@ class RuntimeServiceInterp[F[_]: Parallel](config: RuntimeServiceConfig,
...
+          Async[F].raiseError[DiskSize](RuntimeDiskSizeCannotBeDecreasedException(runtime.projectNameString))
+        else if (!allowStop)
+          Async[F].raiseError[DiskSize](RuntimeDiskSizeCannotBeChangedException(runtime.projectNameString))
+        else
+          Async[F].pure(d)
+      }
+
+      // if target disk size is defined, enqueue pubsub messages to change the PD size and restart the runtime
+      _ <- targetDiskSize.traverse { size =>
+        val diskMessage = UpdateDiskMessage(diskId, size, Some(runtime.id), Some(traceId))
```
Qi77Qi: why don't we merge this intent in `UpdateRuntimeMessage` instead of a separate message for updating diskMesage? won't be get in race condition this way since `UpdateRuntimeMessage` can potentially trigger a restart as well?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BaseCloudServiceRuntimeMonitor.scala
akarukappadath: once a runtime is ready (either the init or start-up script is complete) the disk will have been resized

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
akarukappadath: `restartRuntimeId` only exists if the `UpdateDiskMessage` came from a `updateRuntime` request. If it's coming from a `updateDisk` request, `restartRuntimeId` will be `None`. The reasoning behind this is because we have the `allowStop` field in `updateRuntime,` but it doesn't make sense for the `updateDisk` endpoint. So, if someone updates the disk size through the `updateRuntime` endpoint it will restart the runtime for the user if `allowStop` is true (or throw an error if `allowStop` is false). If someone updates the disk size through the `updateDisk` endpoint, the runtime will not be restarted but the disk will have `needsResize=true` until they restart the runtime themselves

akarukappadath:
jenkins retest

http/src/main/resources/init-resources/startup.sh
```diff
@@ -154,3 +154,7 @@ if [ "$WELDER_ENABLED" == "true" ] ; then
     echo "Starting Welder on cluster $GOOGLE_PROJECT / $CLUSTER_NAME..."
     docker exec -d $WELDER_SERVER_NAME /bin/bash -c "export STAGING_BUCKET=$STAGING_BUCKET && /opt/docker/bin/entrypoint.sh"
 fi
+
+# Resize persistent disk if needed.
+echo "Resizing persistent disk attached to runtime $GOOGLE_PROJECT / $CLUSTER_NAME if disk size changed..."
+resize2fs /dev/sdb
```
Qi77Qi, line 160: this does nothing if the disk doesn't need resize...Azmina originally added `needsResize` variable, but it was added to persistent disk table, which is not the right place because we need this flag for non pd gce runtimes too. I removed `needsResize` entirely since it doesn't seem to me worth the extra complexity.

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/runtimes/RuntimePatchSpec.scala
```diff
@@ -70,39 +77,132 @@ class RuntimePatchSpec extends FixtureAnyFreeSpec with LeonardoTestUtils with Le
...
+          withWebDriver { implicit driver =>
+            withNewNotebook(clusterCopy, Python3) { notebookPage =>
+              //all other packages cannot be tested for their versions in this manner
+              //warnings are ignored because they are benign warnings that show up for python2 because of compilation against an older numpy
+              val res = notebookPage
+                .executeCell(
+                  "! df -H |grep sdb"
+                )
+                .get
+              res should include("22G") //disk output is always a few more gb than what's specified
```
rtitle, line 95: Really.. we specify `DiskSize(20)` and get `22G` from `df -H`?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
```diff
@@ -369,6 +368,15 @@ object clusterQuery extends TableQuery(new ClusterTable(_)) {
     }
   }
 
+  def getActiveClusterRecordByName(project: GoogleProject,
```
rtitle, line 370: Curious why we need this instead of `getActiveClusterByNameMinimal` which seems to be equivalent.. our usual practice is to not return `*Record` objects outside of DB code

Qi77Qi: I added this because originally we were using `getActiveClusterByNameMinimal`, which stubs out fields outside cluster table. I didn't want to keep using `getActiveClusterByNameMinimal` because this caused a bug where I was using dataproc instances from `Runtime` object, but because it was stubbed out, it wasn't doing what I expected it's doing. By using  `getActiveClusterRecordByName`, the return type is exactly what it is, and there's no chance that caller might accidentally be using a stubbed out field

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/runtimes/RuntimePatchSpec.scala
Qi77Qi: yea...not sure why

rtitle (approved):
LGTM after merge conflicts are addressed

PR #1490, opened 2020-07-01
IA-1993 IA-1793 Galaxy proxy routes

rtitle:
https://broadworkbench.atlassian.net/browse/IA-1993
https://broadworkbench.atlassian.net/browse/IA-1793

### Data model changes:
* Dropped `PORT` table -- it is no longer necessary
* Renamed `KUBERNETES_CLUSTER.apiServiceIp` to `KUBERNETES_CLUSTER.externalIp`
* Left `SERVICE` table as is -- it is still useful to track many services for an App.

### Route changes:
* Left the current `/proxy/{googleProject}/{runtimeName}` routes as is
* Made new routes for proxying apps: `/proxy/google/v1/apps/{googleProject}/{appName}/{serviceName}`
   * I didn't want to overload the existing proxy routes because it might be hard to distinguish intent (e.g. is a path segment a runtime name or an app name?)
   * This pattern has nice harmony with the API routes: `/api/google/v1/apps/{googleProject}/{appName}/...`. We could think about migrating our runtime proxy routes to this model as well.

### Service changes:
* `ProxyService` now handles both runtime and app cases
* Renamed `ClusterDnsCache` to `RuntimeDnsCache`, and made a new `KubernetesDnsCache`. Their functionality is similar, but there is slightly different logic.

### Test changes
* Refactored for the above changes
* Removed `AuthProviderSpec`
* Added cases to `ProxyRoutesSpec` to cover apps routes

---

In all cases:


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dns/JupyterNameService.scala
```diff
@@ -25,3 +27,8 @@ class JupyterNameServiceDescriptor extends NameServiceDescriptor {
   override def getProviderName: String = "Jupyter"
   override def getType: String = "dns"
 }
+
+object HostToIpMapping {
+  private[dns] val hostToIpMapping: ValueBox[Map[Host, IP]] = ValueBox(Map.empty)
```
Qi77Qi: what does `ValueBox` do?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/ProxyService.scala
```diff
@@ -161,91 +166,95 @@ class ProxyService(
...
-   *         server IP could not be found.
-   */
   def proxyRequest(userInfo: UserInfo, googleProject: GoogleProject, runtimeName: RuntimeName, request: HttpRequest)(
     implicit ev: ApplicativeAsk[IO, TraceId]
   ): IO[HttpResponse] =
     for {
       _ <- authCheck(userInfo, googleProject, runtimeName, ConnectToRuntime)
-      now <- timer.clock.realTime(TimeUnit.MILLISECONDS)
-      r <- proxyInternal(googleProject, runtimeName, request, Instant.ofEpochMilli(now))
+      now <- nowInstant
```
Qi77Qi: can probably require `implicit ev: ApplicativeAsk[IO, AppContext]` instead of using `nowInstant`

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20200701_kubernetes_drop_port_table.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="leonardo" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+
+    <changeSet logicalFilePath="leonardo" author="rtitle" id="kubernetes_drop_port_table">
+        <dropTable cascadeConstraints="true" tableName="PORT"/>
```
jdcanas, line 5: you may need to explicitly drop the FK constraints before dropping the table 🤔

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dns/KubernetesDnsCache.scala
```diff
@@ -0,0 +1,69 @@
...
+            } yield hostStatus
+
+            res.toIO.unsafeRunSync
+          }
+        }
+      )
+
+  private def host(cluster: KubernetesCluster): Host = {
+    val prefix =
+      Base64.getUrlEncoder().encodeToString(cluster.getGkeClusterId.toString.getBytes(StandardCharsets.UTF_8))
```
jdcanas: huh... whats the point of getting the `s"projects/${project.value}/locations/${location.value}/clusters/${clusterName.value}"` string here?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dns/JupyterNameService.scala
rtitle: So we need to use this global, static variable `hostToIpMapping` to be able to plug into Java DNS stuff. `ValueBox` provides some synchronization around mutating this. I think originally we just used `@volatile` but we need to do a get-and-update so extra synchronization is needed. I think @kyuksel originally added `ValueBox`.

It'd be nice to use `IO` or maybe `Ref` instead but I recall this being hard to plug into `sun.net.spi.nameservice.NameService` which is some magic Java stuff (but maybe it's possible).

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20200701_kubernetes_drop_port_table.xml
rtitle: Maybe - I was wondering if `cascadeConstraints` will work. Will test

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dns/JupyterNameService.scala
Qi77Qi: maybe try `Ref[IO, Map[Host, IP]]`? https://typelevel.org/cats-effect/concurrency/ref.html...this is what welder is using for keeping the cache map

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dns/KubernetesDnsCache.scala
rtitle: Basically I need a URL-safe string to uniquely identify a cluster to use as our "fake" DNS record to match the wildcard cert. For runtimes we use the `googleId` (which is a string of digits). Here I decided to base-64 encode the `GkeClusterId` string. Other approaches are possible too -- the string just needs to be unique per cluster and URL-safe.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dns/JupyterNameService.scala
rtitle: Yeah.. maybe will try that

rtitle: Using `Ref[IO, Map[Host, IP]]` in my latest commit.. it seems to work

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20200701_kubernetes_drop_port_table.xml
rtitle: This seems to successfully drop the table and leave no constraints behind

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dns/JupyterNameService.scala
```diff
@@ -3,25 +3,33 @@ package org.broadinstitute.dsde.workbench.leonardo.dns
...
 }
 
 class JupyterNameServiceDescriptor extends NameServiceDescriptor {
   override def createNameService(): NameService = new JupyterNameService
   override def getProviderName: String = "Jupyter"
   override def getType: String = "dns"
 }
+
+object HostToIpMapping {
+  private[dns] val hostToIpMapping: Ref[IO, Map[Host, IP]] = Ref.unsafe(Map.empty)
```
Qi77Qi, line 43: usually this is initialized in `boot`, do `Ref[IO].of(Map.empty[Host, IO])`, and pass it to constructors for classes that need it

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dns/RuntimeDnsCache.scala
```diff
@@ -0,0 +1,92 @@
...
+  private val projectClusterToHostStatus = CacheBuilder
+    .newBuilder()
+    .expireAfterWrite(cacheConfig.cacheExpiryTime.toSeconds, TimeUnit.SECONDS)
+    .maximumSize(cacheConfig.cacheMaxSize)
+    .recordStats
+    .build(
+      new CacheLoader[RuntimeDnsCacheKey, HostStatus] {
+        def load(key: RuntimeDnsCacheKey): HostStatus = {
+          val res = for {
+            _ <- Effect[F]
```
Qi77Qi: Use `Logger` instead of `LazyLogging`?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dns/RuntimeDnsCache.scala
```diff
@@ -0,0 +1,92 @@
...
+
+          res.toIO.unsafeRunSync
+        }
+      }
+    )
+
+  private def host(r: Runtime): Host = {
+    val googleId = r.asyncRuntimeFields.map(_.googleId)
+    val assumption = s"Google ID for Google project/cluster ${r.googleProject}/${r.runtimeName} must not be undefined."
+    assert(googleId.isDefined, assumption)
```
Qi77Qi: this will throw exception...shall we make `host` to return `Either`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/ProxyService.scala
```diff
@@ -161,91 +170,96 @@ class ProxyService(
...
+
+  def proxyAppRequest(userInfo: UserInfo,
+                      googleProject: GoogleProject,
+                      appName: AppName,
+                      serviceName: ServiceName,
+                      request: HttpRequest)(
+    implicit ev: ApplicativeAsk[IO, AppContext]
+  ): IO[HttpResponse] =
+    for {
+      _ <- IO(logger.info(s"authorizing ${userInfo.userEmail}")) // TODO placeholder for auth check
```
Qi77Qi, line 200: is TODO still valid?

Qi77Qi (approved):
left some minor comments...lgtm otherwise

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dns/RuntimeDnsCache.scala
rtitle: Yeah this existed before.. but I think I will remove the `assert`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/ProxyService.scala
rtitle: Yeah, we haven't done the SAM work yet

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dns/JupyterNameService.scala
rtitle: The issue here is that `JupyterNameService` needs it, but that class needs to be initialized via a no-arg constructor, so we can't pass things to it. This is the reason we define this `Map` statically in an object. I can add a comment about this.

jdcanas (approved):

PR #1491, opened 2020-07-01
IA-1953 QA-1197 Enable all tests

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-1953
https://broadworkbench.atlassian.net/browse/QA-1197

Root cause of the flakiness is recent privacy change in chrome incognito mode blocks all third parties cookies. We turned this off in https://github.com/broadinstitute/workbench-libs/pull/326

1. consoliate tests in biocondutorSpec a bit so that we're opening less webdrivers at the same time
2. Update the sbt container that we've been using for building automation test container image so that the sbt version available on the image matches up with leonardo.
3. Update scalatest to 3.2.0
4. Looks like `NotebookClusterMonitoringSpec` and `NotebookGCEClusterMonitoringSpec` are failing due to real bug introduced lately. I haven't had time to look yet..but will do after this PR

---

In all cases:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM openjdk:8
+FROM oracle/graalvm-ce:20.0.0-java8
 
 EXPOSE 8080
 EXPOSE 5050
--- a/docker/build.sh
+++ b/docker/build.sh
@@ -15,7 +15,6 @@ HELP_TEXT="$(cat <<EOF
    jar : build Leonardo jar
    -d | --docker : (default: no action) provide either "build" or "push" to
            build or push a docker image.  "push" will also perform build.
-   -dr | --dockerhub-registry: The dockerhub registry to push to
    -gr | --gcr-registry: The GCR registry to push to
    -n | --notebook-repo: (default: --project) the repo to push the notebooks
            image. Can be a dockerhub or GCR repo.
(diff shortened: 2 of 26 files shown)


Qi77Qi:
where is jenkins run? jenkins retest

Qi77Qi:
jenkins retest...fiab-start failure

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM openjdk:8
+FROM oracle/graalvm-ce:20.0.0-java8
```
rtitle, line 1: Hm… I'm not opposed to it but are we confident in the switch from OpenJDK -> GraalVM?

Qi77Qi: welder has been using graalvm....so I think it should be fine...I talked to Albano before about graalvm, he's fine with it too

docker/build.sh
```diff
@@ -75,11 +73,6 @@ while [ "$1" != "" ]; do
             RUN_DOCKER=true
             DOCKER_CMD="$1"
             ;;
-        -dr | --dockerhub-registry)
```
rtitle, line 78: Could remove `-ui` above too I guess

project/Settings.scala
```diff
@@ -132,7 +132,7 @@ object Settings {
       * Specify that all tests will be executed in a single external JVM.
       * By default, tests executed in a forked JVM are executed sequentially.
       */
-    Test / fork := true,
+    Test / fork := false,
```
rtitle, line 135: Sounds good to not fork.. does this increase test execution time?

rtitle (approved):
LGTM, good find on the Chrome fix

Dockerfile
rtitle: ok

xuscrbw:
Looks good. Includes the update on the assertion for that issue we were running into

project/Settings.scala
Qi77Qi: doesn't seem like it increases execution time much..it's still around 18 min

PR #1487, opened 2020-06-30
IA-2000 IA-1999 Add dettach

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-2000
https://broadworkbench.atlassian.net/browse/IA-1999

1. add dettach disk
2. explicitly delete disk because once disk is detached, deleting runtime won't automatically delete disks anymore
3. Fix an edge case where when user interrupt `Stopping` with `Deleting`, we weren't doing deleting. Added https://broadworkbench.atlassian.net/browse/IA-2026 as well for addressing `Deleting` in the middle of `Starting`

---
Have you read [CONTRIBUTING.md](https://github.com/DataBiosphere/leonardo/blob/develop/CONTRIBUTING.md) lately? If not, do that first.

I, the developer opening this PR, do solemnly pinky swear that:

- [ ] I've documented my API changes in Swagger

In all cases:

- [ ] Get a thumbsworth of review and PO signoff if necessary
- [ ] Verify all tests go green
- [ ] Squash and merge; Delete your branch after this
- [ ] Test this change deployed correctly and works on dev environment after deployment

PR #1492, opened 2020-07-01
[IA-1959] front leo publishing K8 messages and default nodepool refactor

jdcanas:
- Added pubsub messaging for kubernetes
- refactored nodepool logic and table a bit so DefaultNodepool is an explicit ADT
- Added tests

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
@@ -138,6 +138,10 @@ object NodepoolStatus {
     override def toString: String = "PRECREATING"
   }
 
+  final case object Predeleting extends NodepoolStatus {
+    override def toString: String = "PREDELETING"
+  }
+
   def values: Set[NodepoolStatus] = sealerate.values[NodepoolStatus]
   def stringToObject: Map[String, NodepoolStatus] = values.map(v => v.toString -> v).toMap
 
@@ -159,11 +163,30 @@ final case class Nodepool(id: NodepoolLeoId,
                           autoscalingEnabled: Boolean,
                           autoscalingConfig: Option[AutoscalingConfig],
                           errors: List[KubernetesError],
-                          apps: List[App])
+                          apps: List[App],
+                          isDefault: Boolean)
 object KubernetesNameUtils {
   //UUID almost works for this use case, but kubernetes names must start with a-z
   def getUniqueName[A](apply: String => A): Either[Throwable, A] =
     KubernetesName.withValidation('k' + UUID.randomUUID().toString.toLowerCase.substring(1), apply)
+}
+
+case class DefaultNodepool(id: NodepoolLeoId,
+                           clusterId: KubernetesClusterLeoId,
+                           nodepoolName: NodepoolName,
+                           status: NodepoolStatus,
+                           auditInfo: AuditInfo,
+                           machineType: MachineTypeName,
+                           numNodes: NumNodes,
+                           autoscalingEnabled: Boolean,
+                           autoscalingConfig: Option[AutoscalingConfig])  {
+  def toNodepool(): Nodepool = {
+    Nodepool(id, clusterId, nodepoolName, status, auditInfo, machineType, numNodes, autoscalingEnabled, autoscalingConfig, List.empty, List.empty, true)
+  }
+}
+object DefaultNodepool {
+  def fromNodepool(n: Nodepool) =
+    DefaultNodepool(n.id, n.clusterId, n.nodepoolName, n.status, n.auditInfo, n.machineType, n.numNodes, n.autoscalingEnabled, n.autoscalingConfig)
 }
 
 final case class AutoscalingConfig(autoscalingMin: AutoscalingMin, autoscalingMax: AutoscalingMax)
@@ -260,6 +283,10 @@ object AppStatus {
     override def toString: String = "PRECREATING"
   }
 
+  final case object Predeleting extends AppStatus {
+    override def toString: String = "PREDELETING"
+  }
+
   final case object Provisioning extends AppStatus {
     override def toString: String = "PROVISIONING"
   }
@@ -277,3 +304,6 @@ final case class ServiceConfig(name: ServiceName, kind: KubernetesServiceKindNam
 final case class KubernetesServiceKindName(value: String) extends AnyVal
 
 final case class KubernetesRuntimeConfig(numNodes: NumNodes, machineType: MachineTypeName, autoscalingEnabled: Boolean)
+
+//used in pubsub messaging to indicate the cluster and dummy nodepool to be created
+final case class CreateCluster(clusterId: KubernetesClusterLeoId, nodepoolId: NodepoolLeoId)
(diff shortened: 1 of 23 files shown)


jdcanas:
jenkins retest

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20200701_kubernetes_default_nodepool.xml
```diff
@@ -0,0 +1,8 @@
+<databaseChangeLog logicalFilePath="leonardo" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+    <changeSet logicalFilePath="leonardo" author="jcanas" id="add_default_nodepool_flag">
+        <addColumn tableName="NODEPOOL">
+            <column name="isDefault" type="boolean" defaultValue="false">
```
rtitle, line 4: Can we make this not-null?
```
<constraints nullable="false" /> 
```

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
```diff
@@ -158,13 +162,50 @@ final case class Nodepool(id: NodepoolLeoId,
...
-                          apps: List[App])
+                          apps: List[App],
+                          isDefault: Boolean)
 object KubernetesNameUtils {
   //UUID almost works for this use case, but kubernetes names must start with a-z
   def getUniqueName[A](apply: String => A): Either[Throwable, A] =
     KubernetesName.withValidation('k' + UUID.randomUUID().toString.toLowerCase.substring(1), apply)
 }
 
+case class DefaultNodepool(id: NodepoolLeoId,
```
rtitle, line 174: This class seems a bit duplicative of `Nodepool`. Is there a strong need for a separate type given we also have `Nodepool.isDefault`?

rtitle:
Overall looks good, just a couple comments above

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
jdcanas: I think it is helpful to have to be able to enforce the type of nodepool being passed around with type safety, as opposed to validating a boolean. Without it, there are some scenarios where you need to specify Apps/Errors and it doesn't make sense. Prior to this, it was easy to accidentally pass the wrong nodepool around

The use case for the boolean is to filter db results, which have no notion of the ADT differentiation

rtitle (approved):

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
```diff
@@ -159,13 +163,50 @@ final case class Nodepool(id: NodepoolLeoId,
...
+             machineType,
+             numNodes,
+             autoscalingEnabled,
+             autoscalingConfig,
+             List.empty,
+             List.empty,
+             true)
+}
+object DefaultNodepool {
+  def fromNodepool(n: Nodepool) =
```
Qi77Qi, line 198: you can do something similar to https://github.com/DataBiosphere/leonardo/pull/1487/files#diff-35d0241c3de10d4f5cbec9c74325cb1aR46-R57

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesServiceDbQueries.scala
```diff
@@ -113,6 +116,32 @@ object KubernetesServiceDbQueries {
     joinFullAppAndUnmarshal(kubernetesClusterQuery.findByIdQuery(id), nodepoolQuery, appQuery)
       .map(_.headOption)
 
+  def markPendingCreating(nodepoolId: NodepoolLeoId, appId: AppId, clusterId: Option[CreateCluster])(
```
Qi77Qi: shouldn't `clusterId` be something like `Option[ClusterId]`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -460,6 +465,10 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift](
       )
     } yield ()
 
+  def handleCreateAppMessage(msg: CreateAppMessage): F[Unit] = F.unit
```
Qi77Qi, line 468: add TODO or use `NotImplemented` exception

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/leoPubsubMessageSubscriberModels.scala
```diff
@@ -120,6 +127,25 @@ object LeoPubsubMessage {
...
+                                    traceId: Option[TraceId])
+      extends LeoPubsubMessage {
+    val messageType: LeoPubsubMessageType = LeoPubsubMessageType.CreateApp
+  }
+
+  final case class DeleteAppMessage(appId: AppId,
+                                    nodepoolId: NodepoolLeoId,
+                                    project: GoogleProject,
+                                    deleteDisk: Boolean,
+                                    traceId: Option[TraceId])
```
Qi77Qi, line 144: `traceId` probably doesn't need to be optional

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
jdcanas: ill be implementing this in the ticket I have now, so it won't stay like this for long

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/leoPubsubMessageSubscriberModels.scala
jdcanas: that param is inherited from the trait. Not sure why its optional, I think some LeoPubsubMessages may be a bit older and not pass along a traceId. I'll look at it in depth in my next PR using the messages, and see if I can make it non-optional in the parent trait

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesServiceDbQueries.scala
jdcanas: ah, old param name. Renamed it to something more fitting. The `createCluster` ADT is needed, since creating a cluster involves creating a default nodepool along with it

Qi77Qi: don't you also have `nodepoolId` passed in already? or that's a different `nodepoolId`?

PR #1496, opened 2020-07-07
IA-2031 Have Leo pass auth headers to Galaxy

rtitle:
https://broadworkbench.atlassian.net/browse/IA-2031

Not actually tested in a fiab yet, will do that next.

---

In all cases:

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
@@ -257,6 +257,8 @@ object AppType {
 }
 
 final case class ReleaseName(value: String) extends AnyVal
+final case class RemoteUserName(value: String) extends AnyVal
+final case class RemoteUserSecret(value: String) extends AnyVal
 
 final case class AppId(id: Long) extends AnyVal
 final case class AppName(value: String) extends AnyVal
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
@@ -258,7 +258,6 @@ object AppType {
 
 final case class ReleaseName(value: String) extends AnyVal
 final case class RemoteUserName(value: String) extends AnyVal
-final case class RemoteUserSecret(value: String) extends AnyVal
 
 final case class AppId(id: Long) extends AnyVal
 final case class AppName(value: String) extends AnyVal
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -168,29 +168,30 @@ gke {
     location = "us-central1-a"
   }
   defaultNodepool {
-    machineType = "g1-small",
-    numNodes = 1,
+    machineType = "g1-small"
+    numNodes = 1
     autoscalingEnabled = false
   }
   galaxyNodepool {
-    machineType = "n2-standard-8",
-    numNodes = 1,
-    autoscalingEnabled = true,
-    autoscalingConfig = {
-       autoscalingMin = 0,
+    machineType = "n2-standard-8"
+    numNodes = 1
+    autoscalingEnabled = true
+    autoscalingConfig {
+       autoscalingMin = 0
        autoscalingMax = 1
     }
   }
-  app {
-    releaseName = "release1",
-    namespaceNameSuffix = "namespace",
+  galaxyApp {
+    releaseName = "release1"
+    namespaceNameSuffix = "namespace"
     services = [
       {
-        name = "galaxy-svc",
-        kind = "ClusterIP",
-        ports = []
+        name = "galaxy-web"
+        kind = "ClusterIP"
       }
     ]
+    remoteUserName = "galaxy-user"
+    remoteUserSecret = "galaxy-secret"
   }
 }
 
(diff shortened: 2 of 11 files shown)


http/src/main/resources/reference.conf
```diff
@@ -168,29 +168,30 @@ gke {
     location = "us-central1-a"
   }
   defaultNodepool {
-    machineType = "g1-small",
```
rtitle, line 171: Config syntax cleanup (don't need commas)

http/src/main/resources/reference.conf
```diff
@@ -168,29 +168,30 @@ gke {
...
+    autoscalingEnabled = true
+    autoscalingConfig {
+       autoscalingMin = 0
        autoscalingMax = 1
     }
   }
-  app {
-    releaseName = "release1",
-    namespaceNameSuffix = "namespace",
+  galaxyApp {
```
rtitle, line 184: Renamed `app` to `galaxyApp` since it really contains Galaxy-specific configs (and we may need to add more Galaxy configs in the future).

http/src/main/resources/reference.conf
```diff
@@ -168,29 +168,30 @@ gke {
...
   }
   defaultNodepool {
-    machineType = "g1-small",
-    numNodes = 1,
+    machineType = "g1-small"
+    numNodes = 1
     autoscalingEnabled = false
   }
   galaxyNodepool {
-    machineType = "n2-standard-8",
```
Qi77Qi, line 176: probably should add unit test in https://github.com/DataBiosphere/leonardo/blob/develop/http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/config/ConfigSpec.scala#L14

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
```diff
@@ -216,6 +216,8 @@ object AppType {
 }
 
 final case class ReleaseName(value: String) extends AnyVal
+final case class RemoteUserName(value: String) extends AnyVal
```
Qi77Qi, line 260: just for my own understanding, what will be the value for this remote user name and user secret? and will they be the same for all users and stored in vault?

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/api/ProxyRoutesSpec.scala
```diff
@@ -317,6 +318,14 @@ class ProxyRoutesSpec
     }
   }
 
+  it should "add Galaxy remote user headers" in {
+    Get(s"/proxy/google/v1/apps/$googleProject/$appName/$serviceName")
+      .addHeader(Cookie(tokenCookie)) ~> httpRoutes.route ~> check {
+      responseAs[Data].headers.toList should contain allElementsOf Map("HTTP_REMOTE_USER" -> "galaxy-user",
```
Qi77Qi: so we're setting user and secret in response and galaxy UI knows how to read this from response header?

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/config/ConfigSpec.scala
```diff
@@ -1,12 +1,26 @@
...
-import org.broadinstitute.dsde.workbench.google2.ZoneName
+import org.broadinstitute.dsde.workbench.google2.KubernetesSerializableName.{NamespaceName, ServiceName}
+import org.broadinstitute.dsde.workbench.google2.{Location, MachineTypeName, ZoneName}
 import org.broadinstitute.dsde.workbench.leonardo.monitor.{
   LeoPubsubMessageSubscriberConfig,
   PersistentDiskMonitor,
   PersistentDiskMonitorConfig
 }
-import org.broadinstitute.dsde.workbench.leonardo.{BlockSize, DiskSize, DiskType}
+import org.broadinstitute.dsde.workbench.leonardo.{
```
Qi77Qi: this won't be needed if package line is written as

```
package config
```

Qi77Qi (approved):

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
rtitle: I was planning to just send `galaxy-user` as the user and `galaxy-secret` as the secret for all users. However this isn't really working on the Galaxy end currently so the approach may change.

Good question on vault -- I wasn't really planning to store the secret in vault but it may be a good idea..

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/api/ProxyRoutesSpec.scala
rtitle: Yeah, we set it in the request header to Galaxy and the secret is also configured on the Galaxy side:

https://github.com/galaxyproject/galaxy/blob/d6ba85ae48072bc9c15f17955588ef1576dc5393/lib/galaxy/config/sample/galaxy.yml.sample#L1397

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
Qi77Qi: I see...if it's just dummy value, probably not worth storing in vault then..I thought this will be actual secret

PR #1497, opened 2020-07-08
IA-1958

xuscrbw:
Updated the authentication error message to be more descriptive so users will seek to be added to the billing project if they have correct permissions already.

Removed the Option typing since we should never have a case where userInfo doesn't give us an email. If the access token is invalid, it will fail earlier when trying to get the userInfo from the access token.

---

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
@@ -45,7 +45,8 @@ import scala.concurrent.ExecutionContext
 import scala.concurrent.duration._
 
 case class AuthorizationError(email: Option[WorkbenchEmail] = None)
-    extends LeoException(s"${email.map(e => s"'${e.value}'").getOrElse("Your account")} is unauthorized",
+    extends LeoException(s"${email.map(e => s"'${e.value}'").getOrElse("Your account")} is unauthorized." +
+      "If you have proper permissions to use the workspace, make sure you are also added to the billing account",
                          StatusCodes.Forbidden)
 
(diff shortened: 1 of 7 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala
```diff
@@ -45,7 +45,8 @@ import scala.concurrent.ExecutionContext
 import scala.concurrent.duration._
 
 case class AuthorizationError(email: Option[WorkbenchEmail] = None)
-    extends LeoException(s"${email.map(e => s"'${e.value}'").getOrElse("Your account")} is unauthorized",
+    extends LeoException(s"${email.map(e => s"'${e.value}'").getOrElse("Your account")} is unauthorized." +
```
Qi77Qi: if `email` is `None`, does it mean a different type of error? like we can't extract user's email or something?

Qi77Qi: I would remove the `Option` around email

xuscrbw: Tried to dig through to see when we would not be able to extract the user's email. 

I think the only case would be if the user's access token is invalid and doesn't return the user's email. In that case, I think we would run into an error much earlier in the process (specifically when we get to getUserInfoAndExpirationFromAccessToken). This function will spit out its own Authentication Error. So I think we can do what you recommended and remove the option since we should never get to this point

Qi77Qi (approved):
👍

xuscrbw:
jenkins retest

Qi77Qi:
@xuscrbw think this can be merged?

xuscrbw:
yep this should be good to be merged

PR #1499, opened 2020-07-10
[IA-1973] Fix jupyter extension installs

gpcarr:
Now that we docker compose with env var `PIP_TARGET` we need to update the `gce-init` script to use `PIP_TARGET` instead of `PIP_USER` when installing packages

Tested the following:
- `NotebookGCECustomizationSpec` now passes
- `NotebookPyKernelSpec` still passes
- determined `ROOT_USER_PIP_DIR` and `JUPYTER_USER_PIP_DIR` by reverting my FIAB to before I updated user package installation persistence and seeing where `userJupyterExtensionConfig` get installed to

---

In all cases:

### DIFF
--- a/http/src/main/resources/init-resources/jupyter-docker-compose-gce.yaml
+++ b/http/src/main/resources/init-resources/jupyter-docker-compose-gce.yaml
@@ -28,6 +28,7 @@ services:
       PIP_USER: "false"
       PIP_TARGET: "${NOTEBOOKS_DIR}/packages"
       R_LIBS: "${NOTEBOOKS_DIR}/packages"
+      PYTHONPATH: "${PYTHONPATH:+$PYTHONPATH:}${NOTEBOOKS_DIR}/packages"
     env_file:
       - /etc/google_application_credentials.env
       - /etc/custom_env_vars.env
--- a/http/src/main/resources/init-resources/jupyter-docker-compose-gce.yaml
+++ b/http/src/main/resources/init-resources/jupyter-docker-compose-gce.yaml
@@ -28,7 +28,6 @@ services:
       PIP_USER: "false"
       PIP_TARGET: "${NOTEBOOKS_DIR}/packages"
       R_LIBS: "${NOTEBOOKS_DIR}/packages"
-      PYTHONPATH: "${PYTHONPATH:+$PYTHONPATH:}${NOTEBOOKS_DIR}/packages"
     env_file:
       - /etc/google_application_credentials.env
       - /etc/custom_env_vars.env
--- a/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoSuite.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoSuite.scala
@@ -99,7 +99,7 @@ final class LeonardoSuite
       new ClusterStatusTransitionsSpec,
       new LabSpec,
       new NotebookClusterMonitoringSpec,
-//      new NotebookCustomizationSpec,
+      new NotebookCustomizationSpec,
       new NotebookDataSyncingSpec,
       new LeoPubsubSpec,
       new ClusterAutopauseSpec,
--- a/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoSuite.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoSuite.scala
@@ -107,7 +107,7 @@ final class LeonardoSuite
       new RuntimePatchSpec,
       new RuntimeStatusTransitionsSpec,
       new NotebookGCEClusterMonitoringSpec,
-//      new NotebookGCECustomizationSpec,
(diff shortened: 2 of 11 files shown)


Qi77Qi:
does https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/resources/init-resources/init-actions.sh need the same change?

gpcarr:
> does https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/resources/init-resources/init-actions.sh need the same change?

is `init-actions.sh` only used for GCE VMs? what about `startup.sh`?

gpcarr:
> does https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/resources/init-resources/init-actions.sh need the same change?

seems like `init-actions.sh` is only for dataproc clusters https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/resources/init-resources/init-actions.sh#L6

gpcarr:
jenkins retest

Qi77Qi:
@gpcarr yea, gce-init.sh is for gce vms, and init-actions.sh is for dataproc clusters, they both use the same images. regarding `startup.sh`, if we're installing Jupiter extension there, then we probably should also update it

gpcarr:
jenkins retest

gpcarr:
> @gpcarr yea, gce-init.sh is for gce vms, and init-actions.sh is for dataproc clusters, they both use the same images. regarding `startup.sh`, if we're installing Jupiter extension there, then we probably should also update it

the changes that broke the gce vms were not in the images but in `jupyter-docker-compose-gce` where I set `PIP_USER` to false and set a `PIP_TARGET`, all of the images in `terra-docker` still have `ENV PIP_USER=true` at the end of them

gpcarr:
this comment in `startup.sh` makes it seem like it's only for dataproc clusters but i'm not sure that's true https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/resources/init-resources/startup.sh#L6 @rtitle does `startup.sh` also run on gce vms? if so, it does some installations for jupyter and i'll need to update it as well along with passing a template param telling us if it's running on gce or dataproc because the change is only for gce

gpcarr:
fiab-start failure, jenkins retest

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookBioconductorKernelSpec.scala
```diff
@@ -108,19 +108,19 @@ class NotebookBioconductorKernelSpec extends RuntimeFixtureSpec with NotebookTes
             notebookPage.executeCell("library('BiocSklearn')")
 
             notebookPage.executeCell("""BiocManager::install('rhdf5')""", installTimeout)
-            notebookPage.executeCell("library(rhdf5)")
+            notebookPage.executeCell("library('rhdf5')")
```
Qi77Qi: @nturaga just curious, is `'` needed here?

gpcarr: it shouldn't be https://stackoverflow.com/questions/25210455/how-can-library-accept-both-quoted-and-unquoted-strings

Qi77Qi:
just occurred to me, startup script is shared between dataproc and gce VM, so not sure how you would differentiate the two

gpcarr:
> just occurred to me, startup script is shared between dataproc and gce VM, so not sure how you would differentiate the two

just pushed a change to account for differentiating the two. testing it now

http/src/main/resources/init-resources/jupyter-docker-compose-gce.yaml
```diff
@@ -28,6 +28,7 @@ services:
       PIP_USER: "false"
       PIP_TARGET: "${NOTEBOOKS_DIR}/packages"
       R_LIBS: "${NOTEBOOKS_DIR}/packages"
+      PYTHONPATH: "${PYTHONPATH:+$PYTHONPATH:}${NOTEBOOKS_DIR}/packages"
```
rtitle: Trying to understand this syntax.. does this result in a `PYTHONPATH` of `/etc/jupyter/custom:/home/jupyter-user/notebooks` in the container?

rtitle (approved):
This looks good to me

http/src/main/resources/init-resources/jupyter-docker-compose-gce.yaml
gpcarr: yeah, i added this in this PR so we can account for users that use old docker images that don't have the `/packages` dir in their `PYTHONPATH`

rtitle: Yeah makes sense to add it here, was just confused about the bash syntax

gpcarr: actually removed this line and punting on it for now. it was breaking gce vm creation so need to figure out how to do it correctly but for now going to get the original fix in

PR #1501, opened 2020-07-10
IA-2021 Auto-kill bitcoin processes

rtitle:
https://broadworkbench.atlassian.net/browse/IA-2021

cc @adrazhi 

General approach is:
1. Added https://github.com/broadinstitute/terra-cryptomining-security-alerts as a new submodule of Leo
2. Updated `Jenkinsfile` for custom image building to copy scripts and configs from submodule
3. Updated `prepare_custom_*_image.sh` scripts for GCE and Dataproc to install the falco agent
4. Agent runs as root on the VM:
   ```
   root@saturn-05ec9869-74ac-45cc-8e4c-1434ecd57047:/etc/falco# ps -ef | grep falco
   root       803     1  3 16:45 ?        00:00:22 /usr/bin/falco --daemon --pidfile=/var/run/falco.pid
   ```

Here it is working in a fiab:
image

The [action](https://github.com/broadinstitute/terra-cryptomining-security-alerts/blob/master/report.py) currently kills the mining process and shuts down the VM. Eventually Leo notices and moves the VM to Stopped state, where it can be started again. I'd be interested in any other ideas of actions to take, keeping in mind that we can't use elevated permissions since this has to happen on the VM (we can't delete the VM for example).

---

In all cases:

### DIFF
--- a/jenkins/Jenkinsfile
+++ b/jenkins/Jenkinsfile
@@ -141,7 +141,9 @@ pipeline {
         string(name: "DATAPROC_VERSIONS", defaultValue: "1.2.79-debian9 1.4.15-debian9",
             description: "A custom image will be build for each of these dataproc versions")
         string(name: "IMAGE_HARDENING_LIB_VERSION", defaultValue: "100d620",
-            description: "The hash of the 'broadinstitute/dsp-appsec-base-image-hardening' repo branch that contains libraries for CIS-hardening GCE images")
+            description: "The branch or hash of the 'broadinstitute/dsp-appsec-base-image-hardening' repo that contains libraries for CIS-hardening GCE images")
+        string(name: "CRYPTOMINING_LIB_VERSION", defaultValue: "master",
+            description: "The branch or hash of the 'broadinstitute/terra-cryptomining-security-alerts' repo that contains libraries for detecting cryptomining processes")
         booleanParam(name: "USE_CUSTOM_DOCKER_IMAGE_VERSIONS", defaultValue: false,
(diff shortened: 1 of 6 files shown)


jenkins/Jenkinsfile
```diff
@@ -345,9 +347,12 @@ pipeline {
     stage('Init git submodules') {
         when { expression { !shouldAbortBuild } }
         steps {
-            sh """               
-                git submodule update --init --recursive                
-            """
+            sshagent(['jenkins-ssh-github']) {
```
jdcanas, line 350: huh... odd that this is needed now, before it was only required for the github write permissions it granted when pushing to a repo

It seems that the new submodule requires these permissions to even read?

rtitle: Yeah -- it's a private repo so it needs creds to read as well. Probably don't want our mining detection rules to be readable to the world.

rtitle:
Actually I've only tested GCE so far, need to test Dataproc as well..

jdcanas:
how exactly does leo detect that the VM is stopped in google when it didn't trigger the transition?

Qi77Qi (approved):
lgmt 👍

rtitle:
> how exactly does leo detect that the VM is stopped in google when it didn't trigger the transition?

It eventually auto-pauses and moves to Stopped. We could get fancier here but I don't think we need to care about a good UX for bitcoin miners 😂

jdcanas (approved):

rtitle:
Bioc test failed.. rebasing to hopefully fix it.

PR #1505, opened 2020-07-14
[no ticket] add more clear error messages

Qi77Qi:
Currently, it's hard to tell what error a runtime gets since we only print out status in console. It's also a bit hard to associate the test failure with a specific runtime. We should see error like https://github.com/DataBiosphere/leonardo/blob/develop/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoApiClient.scala#L121 after this PR

---

In all cases:

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoTestUtils.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoTestUtils.scala
@@ -320,7 +320,7 @@ trait LeonardoTestUtils
           )
         )
         resp <- if (monitor)
-          LeonardoApiClient.waitForCreation(googleProject, runtimeName, shouldError)
+          LeonardoApiClient.waitUntilRunning(googleProject, runtimeName, shouldError)
         else LeonardoApiClient.getRuntime(googleProject, runtimeName)
       } yield resp
     }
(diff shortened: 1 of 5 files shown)


automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoTestUtils.scala
```diff
@@ -569,67 +569,26 @@ trait LeonardoTestUtils
...
-      }
-
-      logger.info(s"Checking if runtime is proxyable yet")
-      val getResult = Try(Notebook.getApi(googleProject, runtimeName))
-      getResult.isSuccess shouldBe true
-      getResult.get should not include "ProxyException"
+    logger.info(s"Checking if runtime is proxyable yet")
+    val getResult = Try(Notebook.getApi(googleProject, runtimeName))
+    getResult.isSuccess shouldBe true
+    getResult.get should not include "ProxyException"
```
kyuksel, line 586: If the assertions above fail, will we know what the runtime name is? I think we were saying that'd be useful to know during the walkthrough today.

kyuksel (approved):
LGTM with one consideration inline.

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoTestUtils.scala
Qi77Qi: probably not 🤔  this is copied straight from existing code..I can add runtimeName on line 583

PR #1510, opened 2020-07-16
QA-1204 increase timeout for waiting for tools to startup'

Qi77Qi:
https://broadworkbench.atlassian.net/browse/QA-1204

---

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BaseCloudServiceRuntimeMonitor.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BaseCloudServiceRuntimeMonitor.scala
@@ -298,11 +298,11 @@ abstract class BaseCloudServiceRuntimeMonitor[F[_]] {
                                         dataprocInstances: Set[DataprocInstance]) // only applies to dataproc
   (
     implicit ev: ApplicativeAsk[F, AppContext]
-  ): F[CheckResult] =
+  ): F[CheckResult] = {
     // Update the Host IP in the database so DNS cache can be properly populated with the first cache miss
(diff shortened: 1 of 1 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BaseCloudServiceRuntimeMonitor.scala
```diff
@@ -319,7 +319,8 @@ abstract class BaseCloudServiceRuntimeMonitor[F[_]] {
                                runtimeAndRuntimeConfig.runtime.runtimeName).map(b => (imageType, b))
           )
       }
-      availableTools <- streamFUntilDone(checkTools, 10, 5 seconds).compile.lastOrError
+      // wait for 10 minutes for tools to start up before time out.
+      availableTools <- streamFUntilDone(checkTools, 120, 5 seconds).compile.lastOrError
```
kyuksel, line 323: If a tool needs that much time to come up, the runtime start time will be significantly affected, correct? Did we have some sort of alert for if that happens?

kyuksel (approved):
👍 Would be good to be alerted if start times go up significantly as a result which may warrant an investigation on why tools would take that long to come up.

rtitle (approved):

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BaseCloudServiceRuntimeMonitor.scala
Qi77Qi: 👍 added metrics

PR #1511, opened 2020-07-17
[IA-1889] Jc k8 back leo pubsub

jdcanas:

---

In all cases:


http/src/main/resources/reference.conf
```diff
@@ -167,10 +167,26 @@ groups {
...
+                          "69.173.127.232/29",
+                          "69.173.127.128/26",
+                          "69.173.96.0/20",
+                          "69.173.127.240/28",
+                          "69.173.112.0/21"
+    ]
   }
   defaultNodepool {
-    machineType = "g1-small"
+    machineType = "n1-standard-1"
```
jdcanas, line 190: testing shows we need at least .7 cpu as opposed to the .5 g1-small has for the default nodepool system pods. Bumping this machine size slightly

jdcanas:
jenkins retest

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
```diff
@@ -41,6 +42,9 @@ final case class KubernetesClusterAsyncFields(externalIp: IP, networkInfo: Netwo
 
 final case class NetworkFields(networkName: NetworkName, subNetworkName: SubnetworkName, subNetworkIpRange: IpRange)
 
+//should be in the format 0.0.0.0/0
+final case class CidrIP(value: String)
```
Qi77Qi: extends AnyVal

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
```diff
@@ -227,19 +229,41 @@ final case class DefaultKubernetesLabels(googleProject: GoogleProject,
     )
 }
 
-final case class KubernetesError(errorMessage: String, errorCode: Int, timestamp: Instant, errorSource: ErrorSource)
+final case class KubernetesErrorId(id: Long)
```
Qi77Qi: extends AnyVal. 

Is this this the auto incremented db key?

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesServiceDbQueriesSpec.scala
```diff
@@ -328,4 +333,24 @@ class KubernetesServiceDbQueriesSpec extends AnyFlatSpecLike with TestComponent
...
 
+  it should "save and get errors" in isolatedDbTest {
+    val savedCluster1 = makeKubeCluster(1).save()
+    val savedNodepool1 = makeNodepool(1, savedCluster1.id).save()
+
+    val savedApp1 = makeApp(1, savedNodepool1.id).save()
+    val now = Instant.now()
+    val error1 = KubernetesError("error1", now, ErrorAction.CreateGalaxyApp, ErrorSource.App, Some(1))
+    val error2 = KubernetesError("error2", now, ErrorAction.DeleteGalaxyApp, ErrorSource.Nodepool, Some(2))
+    println(f"now: $now")
```
Qi77Qi: remove println

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/service/MockGKEService.scala
```diff
@@ -0,0 +1,51 @@
...
+import cats.effect.IO
+import cats.mtl.ApplicativeAsk
+import com.google.container.v1.{Cluster, NodePool, Operation}
+import org.broadinstitute.dsde.workbench.DoneCheckable
+import org.broadinstitute.dsde.workbench.google2.{GKEModels, GKEService}
+import org.broadinstitute.dsde.workbench.model.TraceId
+
+import scala.concurrent.duration.FiniteDuration
+
+class MockGKEService extends GKEService[IO] {
```
Qi77Qi: can we move this to workbench-libs

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/service/MockKubernetesService.scala
```diff
@@ -1,38 +1,55 @@
...
-    IO.pure(listAppResponse)
-
-  override def deleteApp(params: DeleteAppParams)(
-    implicit as: ApplicativeAsk[IO, AppContext]
-  ): IO[Unit] =
-    IO.unit
+import org.broadinstitute.dsde.workbench.google2.{GKEModels, KubernetesModels}
+import org.broadinstitute.dsde.workbench.model.TraceId
+
+class MockKubernetesService extends org.broadinstitute.dsde.workbench.google2.KubernetesService[IO] {
```
Qi77Qi: can we move this to workbench-libs

http/src/main/resources/reference.conf
```diff
@@ -167,10 +167,26 @@ groups {
 
 gke {
   cluster {
-    location = "us-central1-a"
+    location = "us-central1-a",
+    region = "us-central1",
+    authorizedNetworks = ["69.173.127.0/25",
```
gpcarr, line 173: how are these networks determined?

jdcanas: via the list here https://dsp-security.broadinstitute.org/cloud-security/google-cloud-platform/gke. I'll add a comment

gpcarr:
this looks good to me. i would +1 for moving the mock services to wb-libs. also awesome unit testing!

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
jdcanas: yes

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/GKEInterpreter.scala
```diff
@@ -0,0 +1,390 @@
...
+  ): F[Unit] = {
+    val createApp = for {
+      ctx <- ev.ask
+      _ <- logger.info(s"beginning app creation for app ${msg.appId} | trace id : ${ctx.traceId}")
+      getAppOpt <- KubernetesServiceDbQueries.getActiveFullAppByName(msg.project, msg.appName).transaction
+      getApp <- F.fromOption(getAppOpt, AppNotFoundException(msg.project, msg.appName, ctx.traceId))
+      gkeClusterId = getApp.cluster.getGkeClusterId
+      namespaceName = getApp.app.appResources.namespace.name
+
+      _ <- kubeService.createNamespace(gkeClusterId, KubernetesNamespace(namespaceName))
```
rtitle: Wondering if there should be a check here, to not create the namespace/secrets if they already exist in the cluster. Namespace/secret creation should only be needed the first time the user deploys an app.

rtitle: I'm also wondering if in the future we should bake the namespace/secret creation into a `leo-helm` chart which we can deploy alongside the app helm chart. That might take care of the "create-if-it-doesn't-exist" semantics.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/GKEInterpreter.scala
```diff
@@ -0,0 +1,390 @@
...
+      //TODO create svc accts
+      //TODO helm create ingress
+      //TODO helm create galaxy
+      _ <- appQuery.updateStatus(msg.appId, AppStatus.Running).transaction
+    } yield ()
+
+    createApp.onError {
+      case e =>
+        for {
+          now <- nowInstant
```
rtitle: Can we use the `now` from the `AppContext` passed to this method instead of generating a new one?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/GKEInterpreter.scala
```diff
@@ -0,0 +1,390 @@
+package org.broadinstitute.dsde.workbench.leonardo
+package monitor
```
rtitle: I'm not certain this class belongs in `monitor` -- we have the Dataproc and GCE interpreters in the `util` package

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/GKEInterpreter.scala
```diff
@@ -0,0 +1,390 @@
...
+
+        (network, subnetwork) <- vpcAlg.setUpProjectNetwork(
+          SetUpProjectNetworkParams(dbCluster.googleProject)
+        )
+
+        kubeNetwork = KubernetesNetwork(dbCluster.googleProject, network).idString
+        kubeSubNetwork = KubernetesSubNetwork(dbCluster.googleProject, dbCluster.region, subnetwork).idString
+
+        cidrBuilder = CidrBlock.newBuilder()
+        _ <- F.delay(config.clusterConfig.authorizedNetworks.foreach(cidrIP => cidrBuilder.setCidrBlock(cidrIP.value)))
```
rtitle: This line doesn't seem quite correct -- seems like it would repeatedly call `setCidrBlock` for each entry in `authorizedNetworks`? Should it be set with a comma-separated list instead?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/GKEInterpreter.scala
```diff
@@ -0,0 +1,390 @@
...
+        getCluster <- F.fromOption(
+          getClusterOpt,
+          ClusterCreationException(
+            s"Failed to get cluster to retrieve external IP after successful creation polling for ${dbCluster.id}"
+          )
+        )
+        _ <- kubernetesClusterQuery
+          .updateAsyncFields(dbCluster.id,
+                             KubernetesClusterAsyncFields(
+                               IP(getCluster.getEndpoint()),
```
rtitle: Hm what does `getEndpoint()` give us? Not sure this is the right IP -- we really need the external IP of the nginx controller's LoadBalancer (which isn't in yet because it needs to be helm installed).

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/GKEInterpreter.scala
```diff
@@ -0,0 +1,390 @@
...
+      dbClusterOpt <- kubernetesClusterQuery.getMinimalClusterById(dbNodepool.clusterId).transaction
+      dbCluster <- F.fromOption(
+        dbClusterOpt,
+        KubernetesClusterNotFoundException(s"cluster with id ${dbNodepool.clusterId} not found in database")
+      )
+      req = KubernetesCreateNodepoolRequest(
+        dbCluster.getGkeClusterId,
+        getGoogleNodepool(dbNodepool)
+      )
+      op <- gkeService.createNodepool(req)
```
rtitle: Nodepool probably also needs create-if-not-exists.. any maybe fancier logic depending on how we resolve the concurrency issue.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/GKEInterpreter.scala
```diff
@@ -0,0 +1,390 @@
...
+    }
+  }
+
+  def deleteApp(getAppResult: GetAppResult)(
+    implicit ev: ApplicativeAsk[F, AppContext]
+  ): F[Unit] = {
+    //TODO: this should actually monitor the deletion went through successfully before updating the status
+    val deleteApp = for {
+      now <- nowInstant
+      _ <- kubeService.deleteNamespace(getAppResult.cluster.getGkeClusterId,
```
rtitle: Need to make sure deleting the namespace actually destroys all resources created by the app

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -528,9 +532,105 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift](
...
     } yield ()
 
-  def handleCreateAppMessage(msg: CreateAppMessage): F[Unit] = F.unit
+  def handleCreateAppMessage(msg: CreateAppMessage)(
+    implicit ev: ApplicativeAsk[F, AppContext]
+  ): F[Unit] = {
+    val createClusterAndNodepool =
+      for {
+        _ <- msg.cluster.fold(F.unit)(createClusterMessage => gkeInterp.createCluster(createClusterMessage, msg.appId))
+        _ <- gkeInterp.createNodepool(msg.nodepoolId, msg.appId)
```
rtitle: We won't _always_ need to create a nodepool during a `createApp` message -- wondering if we could represent that in a similar way as creating a cluster.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/GKEInterpreter.scala
jdcanas: I tried to use a comma separated list and setting the entire list. It seems that you need to set each cidrBlock separately, and a single ip is a 'cidrBlock` as far as the lib is concerned. it took a bit of iteration, and I verified that this works in the cluster via the consolle

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/GKEInterpreter.scala
jdcanas: Oh, I didn't know it wasn't the cluster IP anymore. I'll replace this with a TODO then, it seems we need to wait on the helm command to get this.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/GKEInterpreter.scala
jdcanas: I originally had it there, but I moved it because I wanted to unit test some private methods. 

It's a bit annoying because we test these util interps as part of the LeoPubsubMessageSubscriber, which is in monitor. I'll split those tests out into a separate file and move this to util.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/GKEInterpreter.scala
jdcanas: I've verified that it deletes all things currently created at least. sounds like a good automation test too cc @xuscrbw

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/GKEInterpreter.scala
jdcanas: I'd prefer to defer thinking about the nodepool concerns in this PR, and merge the base code. 

Having this PR live too long sounds like a bad idea if we want to start parallelizing things like getting the helm client calls in, the IAM calls, claiming nodepools, etc

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
jdcanas: See above comment in terms of the new nodepool concerns. I'd prefer to do all the refactoring at once since we'll need a moderate amount of new logic

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/GKEInterpreter.scala
rtitle: Can probably annotate methods as `private[leonardo]` to make available to tests

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/GKEInterpreter.scala
rtitle: Hm ok.. so this doesn't result in only the _last_ one in `config.clusterConfig.authorizedNetworks` being set?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/GKEInterpreter.scala
rtitle: Works for me

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/GKEInterpreter.scala
jdcanas: I think we should aim to delete these resources on deleteApp, so any duplicate resources should be taken care of in front leo when we detect if the app already exists

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/Config.scala
```diff
@@ -627,4 +635,40 @@ object Config {
 
   val leoPubsubMessageSubscriberConfig = config.as[LeoPubsubMessageSubscriberConfig]("pubsub.subscriber")
   val asyncTaskProcessorConfig = config.as[AsyncTaskProcessor.Config]("async-task-processor")
+
+  implicit val secretKeyReader: ValueReader[SecretKey] =
```
Qi77Qi: can you add private to these readers so that people won't accidentally using them directly outside this file

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/SecurityFilesConfig.scala
```diff
@@ -0,0 +1,9 @@
+package org.broadinstitute.dsde.workbench.leonardo.config
+
+import java.nio.file.Path
+
+case class SecurityFilesConfig(proxyServerCrt: Path,
```
Qi77Qi, line 5: will this be shared by both kubernetes app and jupyter/rstudio?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -528,9 +532,105 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift](
...
+  ): F[Unit] = {
+    val delete = for {
+      ctx <- ev.ask
+      _ <- handleDeleteDiskMessage(DeleteDiskMessage(diskId, Some(ctx.traceId)))
+    } yield ()
+
+    delete.onError {
+      case e =>
+        for {
+          now <- nowInstant
```
Qi77Qi: ctx <- ev.ask
now <- ctx.now

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -528,9 +532,105 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift](
...
+    implicit ev: ApplicativeAsk[F, AppContext]
+  ): F[Unit] = {
+    val create = msg.createDisk match {
+      case true =>
+        for {
+          _ <- logger.info(s"Beginning disk creation for app ${msg.appId}")
+          ctx <- ev.ask
+          getAppOpt <- KubernetesServiceDbQueries.getActiveFullAppByName(msg.project, msg.appName).transaction
+          getApp <- F.fromOption(getAppOpt, AppNotFoundException(msg.project, msg.appName, ctx.traceId))
+          disk <- F.fromOption(
```
Qi77Qi, line 693: 👍 didn't realize there's `fromOption`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -528,9 +532,105 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift](
...
+            )
+          )
+          _ <- handleCreateDiskMessage(CreateDiskMessage.fromDisk(disk, Some(ctx.traceId)))
+        } yield ()
+      case false => F.unit
+    }
+    create.onError {
+      case e =>
+        for {
+          now <- nowInstant
```
Qi77Qi: ctx <- ev.ask
now <- ctx.now

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -528,9 +532,105 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift](
...
+    implicit ev: ApplicativeAsk[F, AppContext]
+  ): F[Unit] = {
+    val task = for {
+      ctx <- ev.ask
+      _ <- logger.info(s"beginning delete app for app ${msg.project}/${msg.appName}")
+      getAppOpt <- KubernetesServiceDbQueries.getActiveFullAppByName(msg.project, msg.appName).transaction
+      getApp <- F.fromOption(getAppOpt, AppNotFoundException(msg.project, msg.appName, ctx.traceId))
+      _ <- gkeInterp.deleteNodepool(getApp)
+      _ <- gkeInterp.deleteApp(getApp)
+      _ <- msg.diskId.fold(F.unit)(diskId => deleteDiskForApp(diskId, getApp.app.id))
```
Qi77Qi: traverse?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -528,9 +532,105 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift](
...
+
+  def handleDeleteAppMessage(msg: DeleteAppMessage)(
+    implicit ev: ApplicativeAsk[F, AppContext]
+  ): F[Unit] = {
+    val task = for {
+      ctx <- ev.ask
+      _ <- logger.info(s"beginning delete app for app ${msg.project}/${msg.appName}")
+      getAppOpt <- KubernetesServiceDbQueries.getActiveFullAppByName(msg.project, msg.appName).transaction
+      getApp <- F.fromOption(getAppOpt, AppNotFoundException(msg.project, msg.appName, ctx.traceId))
+      _ <- gkeInterp.deleteNodepool(getApp)
```
Qi77Qi: Trigger of all google ops needs to be kicked off sync to the message acking.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -528,9 +532,105 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift](
       )
     } yield ()
 
-  def handleCreateAppMessage(msg: CreateAppMessage): F[Unit] = F.unit
+  def handleCreateAppMessage(msg: CreateAppMessage)(
+    implicit ev: ApplicativeAsk[F, AppContext]
+  ): F[Unit] = {
+    val createClusterAndNodepool =
+      for {
+        _ <- msg.cluster.fold(F.unit)(createClusterMessage => gkeInterp.createCluster(createClusterMessage, msg.appId))
```
Qi77Qi: we should kick off some google operation sync to the message instead of queueing them

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/SecurityFilesConfig.scala
jdcanas: yeah, they use the same crts

jdcanas:
jenkins retest

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/Config.scala
```diff
@@ -627,4 +635,40 @@ object Config {
 
   val leoPubsubMessageSubscriberConfig = config.as[LeoPubsubMessageSubscriberConfig]("pubsub.subscriber")
   val asyncTaskProcessorConfig = config.as[AsyncTaskProcessor.Config]("async-task-processor")
+
+  implicit private val secretKeyReader: ValueReader[SecretKey] =
+    stringValueReader.map(s => KubernetesName.withValidation(s, SecretKey).right.get)
```
Qi77Qi: can you throw an explicit exception here instead of doing `.get`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/AppErrorComponent.scala
```diff
@@ -0,0 +1,56 @@
...
+package db
+
+import java.time.Instant
+
+import LeoProfile.api._
+import LeoProfile.mappedColumnImplicits._
+
+import scala.concurrent.ExecutionContext
+
+case class AppErrorRecord(id: Long,
```
Qi77Qi: shall we have a custom type for appErrorRecord id so that we won't accidentally use other `id` for this id?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesServiceDbQueries.scala
```diff
@@ -64,12 +66,28 @@ object KubernetesServiceDbQueries {
 
   def getActiveFullAppByName(googleProject: GoogleProject, appName: AppName, labelFilter: LabelMap = Map())(
     implicit ec: ExecutionContext
+  ): DBIO[Option[GetAppResult]] =
+    getActiveFullApp(listClustersByProject(Some(googleProject)),
```
Qi77Qi, line 70: why does `listClustersByProject` takes an optional project argument? shouldn't it always take a project?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -528,9 +540,127 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift](
...
+            ctx <- ev.ask
+            _ <- logger.error(e)(s"Failed to delete app for app ${msg.appId}")
+            _ <- appErrorQuery
+              .save(msg.appId,
+                    KubernetesError(e.getMessage(), ctx.now, ErrorAction.CreateGalaxyApp, ErrorSource.Cluster, None))
+              .transaction
+          } yield ()
+      }
+      createClusterAndNodepool = for {
+        _ <- clusterResult.fold(F.unit)(result => gkeInterp.pollCluster(result))
```
Qi77Qi: clusterResult.traverse

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -528,9 +540,127 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift](
...
+      ctx <- ev.ask
+      _ <- handleDeleteDiskMessage(DeleteDiskMessage(diskId, Some(ctx.traceId)), true)
+    } yield ()
+
+    delete.onError {
+      case e =>
+        for {
+          ctx <- ev.ask
+          _ <- logger.error(e)(s"Failed to delete disk for app ${appId}")
+          _ <- appErrorQuery
```
Qi77Qi: I don't think we need to `onError` every single google op...I think it's better to `adaptError` after each google call, and outside for comp, do a `onError` and save the error in one place

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -528,9 +540,127 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift](
...
+        deleteResult <- gkeInterp.deleteNodepool(getApp)
+      } yield (getApp, deleteResult)
+      (getApp, deleteResult) <- initialGoogleCall.onError {
+        case e =>
+          for {
+            ctx <- ev.ask
+            _ <- logger.error(e)(s"Failed to delete app for app ${msg.appId}")
+            _ <- appErrorQuery
+              .save(msg.appId,
+                    KubernetesError(e.getMessage(), ctx.now, ErrorAction.DeleteGalaxyApp, ErrorSource.Nodepool, None))
```
Qi77Qi: same comment...`adaptError` here, and use `onError` on the whole fo comprehension. Also we put `ctx <- ev.ask` at the beginning of the for comp, so that it's available to all statements after, line 646 is duplicate

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/package.scala
```diff
@@ -32,6 +32,12 @@ package object http {
       .compile
       .lastOrError
 
+  def readFileToBytes[F[_]: Sync: ContextShift](path: Path, blocker: Blocker): F[List[Byte]] =
```
Qi77Qi, line 35: where is this used? also why do we need to convert `Stream[F, Byte]` into `F[List[Byte]]`?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesServiceDbQueries.scala
jdcanas: its reused by the generic listApps endpoint which takes an optional project. Figured it was simpler to make it optional in the service queries as well as opposed to making a similar method that only diiffers by a single filter

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
jdcanas: what exactly is the difference? The reason that any possible failures are surrounded by onError is because we care what step of the process the error occurred, as we want to save the `ErrorSource`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/AppErrorComponent.scala
jdcanas: I created it but never used it 🤔

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/package.scala
jdcanas: its used here https://github.com/DataBiosphere/leonardo/pull/1511/files#diff-d08c4643745ad6f27c31469d39980e33R94 because we require a java Array which we eventually to pass to a client lib, and don't actually consume the bytes in Leo. Its for certificate handling

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
Qi77Qi: now that I think about it, I think persisting to DB should happen in https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala#L83...

To your question, it removes some duplicate code, and we won't have db save and log.error everywhere randomly. You would still know the source because you'll do `adaptError(KuberErro(xxx, errorSource = ???, ???))`

jdcanas:
jenkins retest

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -534,9 +556,185 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift](
...
+            )
+        }
+      } yield ()
+
+      _ <- asyncTasks.enqueue1(
+        Task(ctx.traceId, task, Some(handleKubernetesError), ctx.now)
+      )
+    } yield ()
+
+  private def handleKubernetesError(e: Throwable): F[Unit] =
```
Qi77Qi, line 658: take a `PubsubKubernetesError` instead of `Throwable`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/KubernetesServiceInterp.scala
```diff
@@ -81,6 +82,13 @@ class LeoKubernetesServiceInterp[F[_]: Parallel](
 
       saveCluster <- F.fromEither(getSavableCluster(userInfo, googleProject, ctx.now))
       saveClusterResult <- KubernetesServiceDbQueries.saveOrGetForApp(saveCluster).transaction
+      _ <- if (saveClusterResult.minimalCluster.status == KubernetesClusterStatus.Error)
+        F.raiseError[Unit](
+          KubernetesAppCreationException(
+            s"You cannot create an app while a cluster is in status ${KubernetesClusterStatus.Error}"
```
Qi77Qi: should be `saveClusterResult.minimalCluster.status`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -0,0 +1,339 @@
...
+  gkeService: org.broadinstitute.dsde.workbench.google2.GKEService[F],
+  kubeService: org.broadinstitute.dsde.workbench.google2.KubernetesService[F],
+  blocker: Blocker
+)(implicit val executionContext: ExecutionContext,
+  timer: Timer[F],
+  logger: StructuredLogger[F],
+  dbRef: DbReference[F],
+  F: ConcurrentEffect[F]) {
+
+  def createAndPollApp(msg: CreateAppMessage)(
```
Qi77Qi, line 46: this function doesn't seem to poll?

project/Dependencies.scala
```diff
@@ -21,7 +21,7 @@ object Dependencies {
   val workbenchUtilV = "0.5-4c7acd5"
   val workbenchModelV = "0.13-31cacc4"
   val workbenchGoogleV = "0.21-2a218f3"
-  val workbenchGoogle2V = "0.11-f0a2ce1"
```
Qi77Qi, line 24: update hash to published version

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/AppErrorComponent.scala
```diff
@@ -0,0 +1,56 @@
...
+                                    error.googleErrorCode)
+
+  def get(appId: AppId)(implicit ec: ExecutionContext): DBIO[List[KubernetesError]] =
+    appErrorQuery.filter(_.appId === appId).result map { recs =>
+      val errors = recs map { rec => unmarshallAppErrorRecord(rec) }
+      errors.toList
+    }
+
+  def unmarshallAppErrorRecord(appErrorRecord: AppErrorRecord): KubernetesError =
+    KubernetesError(appErrorRecord.errorMessage,
```
Qi77Qi: shall we rename `KubernetesError` to `AppError`?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/package.scala
Qi77Qi: I think you can do `.compile.to(Array)` on a `Stream[F, Byte]` directly to get an Array of bytes

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -534,9 +556,185 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift](
...
+      case _ =>
+        F.raiseError(new RuntimeException("handleKubernetesError should not be used with a non kubernetes error"))
+    }
+
+  private def deleteDiskForApp(diskId: DiskId)(
+    implicit ev: ApplicativeAsk[F, AppContext]
+  ): F[Unit] =
+    for {
+      ctx <- ev.ask
+      _ <- handleDeleteDiskMessage(DeleteDiskMessage(diskId, Some(ctx.traceId)), true)
```
Qi77Qi: minor, but I would probably leave `handleDeleteDiskMessage` untouched, and create another function that takes a `sync` param, something like `private def deleteDisk(diskId: DiskId])(implicit ev: ApplicativeAsk[F, AppContext])` because I see `handleXXX` functions as top level function only called in `messageResponder`, and wouldn't expect it being used by other functions.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
jdcanas: it doesn't yet, but there's a TODO to add it once we have the helm client actually creating the app

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
jdcanas: I don't want to change the contract for a `Task` which takes `Throwable => F[Unit]`

Qi77Qi: I thought you're only using this in the pubsub error handler..I guess it's fine then if you're using it elsewhere too

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
jdcanas: I was a little hesitant, but for now the desired functionality is identical and I like not duplicating the code. There are unit tests to cover both success and failure for disks, so if they diverge and it breaks for kubernetes, we can evaluate if we want to split them

Qi77Qi: I don't mean to duplicate code...but mostly just to to break out body of `handleDeleteDiskMessage` to another function that can be shared

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -408,73 +421,99 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift](
...
-          _ <- logger.error(e)(s"Failed to create disk ${msg.name.value} in Google project ${msg.googleProject.value}")
-          _ <- persistentDiskQuery.updateStatus(msg.diskId, DiskStatus.Failed, ctx.now).transaction[F]
-        } yield ()
-    }
-  }
+
+  //this returns an F[F[Unit]. It kicks off the google operation, and then return an F containing the async polling task
+  private[monitor] def createDisk(msg: CreateDiskMessage)(
+    implicit ev: ApplicativeAsk[F, AppContext]
+  ): F[F[Unit]] =
```
Qi77Qi: I think having a `sync` param is a bit easier to reason about this code. can we just use a `isSync` param

Qi77Qi (approved):

jdcanas:
jenkins retest

jdcanas:
jenkins retest

PR #1513, opened 2020-07-20
[IA-2060] Add custom environment variables to `CreateAppRequest`

gpcarr:
This PR adds `customEnvironmentVariables` to `CreateAppRequest` and to `CreateAppMessage` and saves them to the db (`App` table) in front leo. 
---

In all cases:

### DIFF
--- a/http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20200719_add_app_custom_env_vars.xml
+++ b/http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20200719_add_app_custom_env_vars.xml
@@ -1,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="leonardo" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+    <changeSet logicalFilePath="leonardo" author="gabriela" id="add_app_custom_env_vars">
+        <addColumn tableName="APP">
+            <column name="customEnvironmentVariables" type="varchar(2048)" >
+                <constraints nullable="true"/>
+            </column>
+        </addColumn>
+    </changeSet>
+</databaseChangeLog>
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/AppComponent.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/AppComponent.scala
@@ -44,7 +44,8 @@ final case class AppRecord(id: AppId,
                            destroyedDate: Instant,
                            dateAccessed: Instant,
                            namespaceId: NamespaceId,
-                           diskId: Option[DiskId])
+                           diskId: Option[DiskId],
+                           customEnvironmentVariables: Option[Map[String, String]])
 
 class AppTable(tag: Tag) extends Table[AppRecord](tag, "APP") {
   //unique (appName, destroyedDate)
@@ -60,6 +61,7 @@ class AppTable(tag: Tag) extends Table[AppRecord](tag, "APP") {
   def dateAccessed = column[Instant]("dateAccessed", O.SqlType("TIMESTAMP(6)"))
   def namespaceId = column[NamespaceId]("namespaceId", O.Length(254))
   def diskId = column[Option[DiskId]]("diskId", O.Length(254))
+  def customEnvironmentVariables = column[Option[Map[String, String]]]("customEnvironmentVariables")
 
   def * =
     (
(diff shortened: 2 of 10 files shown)


http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20200719_add_app_custom_env_vars.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="leonardo" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+    <changeSet logicalFilePath="leonardo" author="gabriela" id="add_app_custom_env_vars">
+        <addColumn tableName="APP">
+            <column name="customEnvironmentVariables" type="varchar(2048)" >
```
Qi77Qi, line 5: shall we call it terra environment variables instead of `customEnvironmentVariables`? existing `customEnvironmentVariables` is for setting API clients to set custom environments...

rtitle: Not sure I understand? This `customEnvironmentVariables` is available to API clients and Terra UI

Qi77Qi: by environment variables, we're talking about `WORKSPACE_NAME` AND `WORKSPACE_BUCKET` here right? or we're talking about actual client set envs?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/AppComponent.scala
```diff
@@ -156,7 +160,8 @@ object appQuery extends TableQuery(new AppTable(_)) {
         saveApp.app.auditInfo.destroyedDate.getOrElse(dummyDate),
         saveApp.app.auditInfo.dateAccessed,
         namespaceId,
-        diskOpt.map(_.id)
+        diskOpt.map(_.id),
+        saveApp.app.customEnvironmentVariables
```
rtitle, line 164: Does this end up getting stored as JSON in the database?

rtitle: I think so because of https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/LeoProfile.scala#L75-L86

rtitle (approved):
Looks good

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20200719_add_app_custom_env_vars.xml
rtitle: I think we'll use it to set `WORKSPACE_NAME` AND `WORKSPACE_BUCKET` but callers could set arbitrary EVs if they want. They'll be stored in a `ConfigMap` in GKE and the galaxy charts just take a set of key/value pairs.

gpcarr: Not sure I understand your first comment @Qi77Qi. I modeled this after how we set custom environment variables for runtimes, the UI can pass whichever var it needs to and like Rob said, we're passing them to the helm install command as key/value pairs

gpcarr: https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala#L48 and https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/leonardoServiceContractModel.scala#L32 for reference

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/AppComponent.scala
gpcarr: yeah seems like it. is that okay? i modeled after how we save `customClusterEnvironmentVariables` for runtimes. that code is a bit more complicated so I'm not sure if the env vars also save as json there too. they're stored as a `Map` on the cluster object so i'm assuming so

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20200719_add_app_custom_env_vars.xml
Qi77Qi: I understand we're passing them to helm install...It's just the naming is confusing IMO because `WORKSPACE_NAME` AND `WORKSPACE_BUCKET` aren't `customEnvironmentVariables` in runtime routes. They're set by Jupiter image assuming some input from UI. 

If we call this just `environmentVariables`, then it makes more sense to me.

Anyways, now that you guys clarified this is for `WORKSPACE_NAME` AND `WORKSPACE_BUCKET` and potentially other envs, I'm fine with leaving it as is

PR #1514, opened 2020-07-20
[IA-2061] update swagger docs

jdcanas:


http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -1122,6 +1122,252 @@ paths:
               schema:
                 $ref: "#/components/schemas/ErrorReport"
 
+  ## Kubernetes App API ##
+  "/api/google/v1/app":
+    get:
+      tags:
+        - app
+      summary: List kubernetes app
```
Qi77Qi: can you add "Experimental" or something to indicate this API is under active development, and subject to change....also can you update https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/resources/swagger/api-docs.yaml#L7, since it's new feature, so probably a minor version bump

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -1640,6 +1886,14 @@ components:
...
           example:
             size: 600
+    CreateAppRequest:
+      content:
+        application/json:
+          schema:
+            $ref: "#/components/schemas/CreateAppRequest"
+          example:
+            diskConfig:
+              name: "disk1"
```
Qi77Qi, line 1962: wrong example?

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -2773,3 +3036,130 @@ components:
...
+      description: the configuration of an app
+      type: object
+      properties:
+        googleProject:
+          type: string
+          description: the project this app is associated with
+        kubernetesRuntimeConfig:
+          type: object
+          $ref: '#/components/schemas/KubernetesRuntimeConfig'
+        errors:
```
Qi77Qi, line 3111: I think for `listRuntime`, we intentionally removed `errors` field from listRuntime response

http/src/main/resources/swagger/api-docs.yaml
jdcanas: this is correct, and the usage for galaxy app creation

http/src/main/resources/swagger/api-docs.yaml
jdcanas: why? I think its helpful to include errors when listing

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -2582,7 +2844,7 @@ components:
         - type: object
           properties:
             persistentDisk:
-              $ref: "#/components/schemas/PersistentDiskConfig"
+              $ref: "#/components/schemas/PersistentDiskRequest"
```
Qi77Qi, line 2914: this isn't correct...`blockSize` in `PersistentDiskRequest` doesn't apply

http/src/main/resources/swagger/api-docs.yaml
Qi77Qi: reason back then was we weren't actually joining CLUSTER_ERROR table for listRuntime API, so it's always empty...And then we just decided to remove it, and also was leaning towards making list call more lean and having client to call `getRuntime` if they need to know about error

http/src/main/resources/swagger/api-docs.yaml
jdcanas: `PersistentDiskRequest` doesn't have a parameter `blockSize` here, I removed it.

It now corresponds to the ADT

```
final case class PersistentDiskRequest(name: DiskName,
                                       size: Option[DiskSize],
                                       diskType: Option[DiskType],
                                       labels: LabelMap)
```

http/src/main/resources/swagger/api-docs.yaml
jdcanas: To report the status for an app, we need to check if the cluster or nodepool for it has errored (i.e., if, during creation, we errorred before getting to app creation). So we already have to do the join,  I figure it can be helpful to return those errors

http/src/main/resources/swagger/api-docs.yaml
Qi77Qi: oh cool..I thought it's shared with createDisk API, which I double checked it's not

Qi77Qi (approved):

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -4,7 +4,7 @@ info:
   description: |
     Workbench notebooks service.
   # Follow semantic versioning https://semver.org/
-  version: "0.0.1"
+  version: "0.1.1"
```
rtitle: Why not 0.1.0?

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -1122,6 +1122,252 @@ paths:
               schema:
                 $ref: "#/components/schemas/ErrorReport"
 
+  ## Kubernetes App API ##
```
rtitle, line 1127: I think there's a `tags` section at the top to update too?

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -1122,6 +1122,252 @@ paths:
...
+  "/api/google/v1/app":
```
rtitle: Can we make our routes plural? E.g. we have `runtimes`, `disks`

rtitle: And I'm not sure we settled on whether we want `google` in it or not

rtitle (changes requested):
Few comments above, let's make sure we have the routes correct before we document in swagger to avoid further version bumps. :)

Also would be good to test this in a swagger validator before merging

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/api/ProxyRoutesSpec.scala
```diff
@@ -241,9 +241,6 @@ class ProxyRoutesSpec
       status shouldEqual StatusCodes.OK
       validateCors()
     }
-    Get(s"/api/google/v1/apps/foo").addHeader(Cookie(tokenCookie)) ~> httpRoutes.route ~> check {
```
jdcanas, line 244: not too sure why this was in the proxy route spec, but it succeeds now that we've changed app -> apps. Seems  out of place in the context of the rest of the test code.

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -22,6 +22,8 @@ tags:
     description: Notebooks API. Deprecated in favor of Proxy API.
   - name: service_info
     description: Service Info API
+  - name: apps
+    description: App API info
```
rtitle: Maybe "Experimental Apps API"?

rtitle: Also while we're in here, maybe `s/Experimental Persistent Disks API/Persistent Disks API` above?

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -1122,6 +1124,252 @@ paths:
...
+                $ref: "#/components/schemas/ErrorReport"
+      security:
+        - googleoauth:
+            - openid
+            - email
+            - profile
+  "/api/google/v1/apps/{googleProject}":
+    get:
+      summary: Experimental. List kubernetes app
+      description: List kubernetes apps the caller has access to, without specifying a project
```
rtitle: Update description, this route does specify a project

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -1122,6 +1124,252 @@ paths:
...
+                $ref: "#/components/schemas/ErrorReport"
+      security:
+        - googleoauth:
+            - openid
+            - email
+            - profile
+    get:
+      summary: Experimental. Get details of an App
+      description: Returns information about an existing App managed by Leo.
+        Poll this to find out when your app has finished starting up.
```
rtitle, line 1294: Is this multiline string valid yaml?

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -1122,6 +1124,252 @@ paths:
...
+        - in: path
+          name: appName
+          description: appName
+          required: true
+          schema:
+            type: string
+      requestBody:
+        $ref: "#/components/requestBodies/CreateAppRequest"
+      responses:
+        "200":
```
rtitle: I think a success returns 202?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/KubernetesRoutes.scala
```diff
@@ -32,7 +32,7 @@ class KubernetesRoutes(kubernetesService: KubernetesService[IO], userInfoDirecti
   val routes: server.Route = userInfoDirectives.requireUserInfo { userInfo =>
```
rtitle, line 32: Wondering if we could rename this class `AppRoutes` as well

http/src/main/resources/swagger/api-docs.yaml
rtitle: _Also_ while we're in here we should maybe update the Proxy section to include the `/proxy/google/v1/apps/{project}/{appName}/{serviceName}` route

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/api/ProxyRoutesSpec.scala
rtitle: Think it was trying to assert that the proxy routes don't listen on `/api`… but yeah, fine to remove it.

jdcanas: Yeah, I typically don't like removing tests, but since we use httpRoutes in this test code, the `AppRoutes` are included and it seems like too much of a negative to test

jdcanas:
jenkins retest

rtitle:
I plugged this into https://editor.swagger.io/ and got a few errors:

```
Errors
Hide
 
Semantic error at paths./api/google/v1/apps/{googleProject}.get.operationId
Operations must have unique operationIds.
Jump to line 1185

Semantic error at paths./proxy/google/v1/apps/{project}/{appName}/{serviceName}
Declared path parameter "project" needs to be defined as a path parameter at either the path or operation level
Jump to line 1826

Semantic error at paths./proxy/google/v1/apps/{project}/{appName}/{serviceName}.get.parameters.0.name
Path parameter "googleProject" must have the corresponding {googleProject} segment in the "/proxy/google/v1/apps/{project}/{appName}/{serviceName}" path
Jump to line 1839
```

rtitle (changes requested):
> jdcanas requested a review from rtitle 19 minutes ago

Overall the changes look good to me, I think there may be a few lingering Swagger syntax errors though (see my comment above).

jdcanas:
> I plugged this into https://editor.swagger.io/ and got a few errors:
> 
> ```
> Errors
> Hide
>  
> Semantic error at paths./api/google/v1/apps/{googleProject}.get.operationId
> Operations must have unique operationIds.
> Jump to line 1185
> 
> Semantic error at paths./proxy/google/v1/apps/{project}/{appName}/{serviceName}
> Declared path parameter "project" needs to be defined as a path parameter at either the path or operation level
> Jump to line 1826
> 
> Semantic error at paths./proxy/google/v1/apps/{project}/{appName}/{serviceName}.get.parameters.0.name
> Path parameter "googleProject" must have the corresponding {googleProject} segment in the "/proxy/google/v1/apps/{project}/{appName}/{serviceName}" path
> Jump to line 1839
> ```

huh... I didn't know it would render with errors, I tested in fiab

jdcanas:
did some other misc cleanup of warnings

rtitle (approved):

PR #1515, opened 2020-07-22
IA-1995 retrieve certs in startup

Qi77Qi:
Always pull new certs in startup script

this should work as expected now. From startup script:
```
Jul 23 15:51:15 saturn-8a0b8b81-9df2-44b1-be25-3ca1233b9a4f GCEMetadataScripts[1478]: 2020/07/23 15:51:15 GCEMetadataScripts: startup-script: ++ openssl x509 -enddate -noout -in /certs/jupyter-server.crt
Jul 23 15:51:15 saturn-8a0b8b81-9df2-44b1-be25-3ca1233b9a4f GCEMetadataScripts[1478]: 2020/07/23 15:51:15 GCEMetadataScripts: startup-script: + notAfter='notAfter=Jun 15 18:22:45 2023 GMT'
Jul 23 15:51:15 saturn-8a0b8b81-9df2-44b1-be25-3ca1233b9a4f GCEMetadataScripts[1478]: 2020/07/23 15:51:15 GCEMetadataScripts: startup-script: + [[ notAfter=Jun 15 18:22:45 2023 GMT != *\n\o\t\A\f\t\e\r\=\J\u\l\ \2\2* ]]
Jul 23 15:51:15 saturn-8a0b8b81-9df2-44b1-be25-3ca1233b9a4f GCEMetadataScripts[1478]: 2020/07/23 15:51:15 GCEMetadataScripts: startup-script: + gsutil cp gs://leoinit-saturn-8a0b8b81-9d-57373a80-33e9-4cd5-b887-64c994a61acf/jupyter-server.crt /certs
```

---

In all cases:

### DIFF
--- a/http/src/main/resources/init-resources/startup.sh
+++ b/http/src/main/resources/init-resources/startup.sh
@@ -64,6 +64,21 @@ export WELDER_MEM_LIMIT=$(welderMemLimit)
 export MEM_LIMIT=$(memLimit)
 export USE_GCE_STARTUP_SCRIPT=$(useGceStartupScript)
 
+# Overwrite old cert on restart
+SERVER_CRT=$(proxyServerCrt)
+SERVER_KEY=$(proxyServerKey)
+ROOT_CA=$(rootCaPem)
+
+if openssl x509 -checkend 86400 -noout -in file.pem
+then
+  echo "Certificate is good for another day!"
+else
+  echo "Certificate has expired or will do so within 24 hours! Going to retrieve new certs"
+  gsutil cp ${SERVER_CRT} /certs
+  gsutil cp ${SERVER_KEY} /certs
+  gsutil cp ${ROOT_CA} /certs
+fi
+
 JUPYTER_HOME=/etc/jupyter
 
 # TODO: remove this block once data syncing is rolled out to Terra
--- a/http/src/main/resources/init-resources/startup.sh
+++ b/http/src/main/resources/init-resources/startup.sh
@@ -70,10 +70,14 @@ SERVER_KEY=$(proxyServerKey)
 ROOT_CA=$(rootCaPem)
 
 ## This helps when we need to rotate certs.
-gsutil cp ${SERVER_CRT} /certs
-gsutil cp ${SERVER_KEY} /certs
-gsutil cp ${ROOT_CA} /certs
-docker-compose -f /etc/proxy-docker-compose.yaml restart
+notAfter=`openssl x509 -enddate -noout -in jupyter-server.crt` # output should be something like `notAfter=Jul 22 13:09:15 2023 GMT`
+
+if [[ $notAfter = 'notAfter=Jul 22'* ]] ; then
+  gsutil cp ${SERVER_CRT} /certs
+  gsutil cp ${SERVER_KEY} /certs
+  gsutil cp ${ROOT_CA} /certs
+  docker-compose -f /etc/proxy-docker-compose.yaml restart
+fi
 
 JUPYTER_HOME=/etc/jupyter
 
--- a/http/src/main/resources/init-resources/startup.sh
+++ b/http/src/main/resources/init-resources/startup.sh
@@ -83,7 +83,8 @@ ROOT_CA=$(rootCaPem)
 ## This helps when we need to rotate certs.
 notAfter=`openssl x509 -enddate -noout -in /certs/jupyter-server.crt` # output should be something like `notAfter=Jul 22 13:09:15 2023 GMT`
 
-if [[ $notAfter = 'notAfter=Jul 22'* ]] ; then
+## If cert is old, then pull latest certs. Update date if we need to rotate cert again
+if [[ "$notAfter" != *"notAfter=Jul 22"* ]] ; then
   gsutil cp ${SERVER_CRT} /certs
   gsutil cp ${SERVER_KEY} /certs
   gsutil cp ${ROOT_CA} /certs
@@ -121,7 +122,7 @@ if [ ! -z ${JUPYTER_START_USER_SCRIPT_URI} ] ; then
     docker exec --privileged -u root -e PIP_USER=false ${JUPYTER_SERVER_NAME} ${JUPYTER_HOME}/${JUPYTER_START_USER_SCRIPT} &> start_output.txt || EXIT_CODE=$?
   fi
 
-  failScript
+  failScriptIfError
 fi
 
 # By default GCE restarts containers on exit so we're not explicitly starting them below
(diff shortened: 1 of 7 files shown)


http/src/main/resources/init-resources/startup.sh
```diff
@@ -64,6 +64,21 @@ export WELDER_MEM_LIMIT=$(welderMemLimit)
...
+ROOT_CA=$(rootCaPem)
+
+if openssl x509 -checkend 86400 -noout -in file.pem
+then
+  echo "Certificate is good for another day!"
+else
+  echo "Certificate has expired or will do so within 24 hours! Going to retrieve new certs"
+  gsutil cp ${SERVER_CRT} /certs
+  gsutil cp ${SERVER_KEY} /certs
+  gsutil cp ${ROOT_CA} /certs
```
rtitle, line 90: Wondering if the apache container needs to be restarted too? E.g.
```
docker-compose -f /etc/proxy-docker-compose.yaml stop
docker-compose -f /etc/proxy-docker-compose.yaml rm -f
docker-compose -f /etc/proxy-docker-compose.yaml start
```

rtitle: Or maybe it'll dynamically pull the new files from `/certs`..

http/src/main/resources/init-resources/startup.sh
```diff
@@ -64,6 +64,17 @@ export WELDER_MEM_LIMIT=$(welderMemLimit)
 export MEM_LIMIT=$(memLimit)
 export USE_GCE_STARTUP_SCRIPT=$(useGceStartupScript)
 
+# Overwrite old cert on restart
+SERVER_CRT=$(proxyServerCrt)
```
rtitle, line 79: Do these match keys in `RuntimeTemplateValues`, and are they populated at runtime resume time? (I recall some templated values are optional and are only populated at creation time.)

http/src/main/resources/init-resources/startup.sh
```diff
@@ -64,6 +64,17 @@ export WELDER_MEM_LIMIT=$(welderMemLimit)
...
+SERVER_KEY=$(proxyServerKey)
+ROOT_CA=$(rootCaPem)
+
+## This helps when we need to rotate certs.
```
rtitle, line 83: Hm, do we want to do this on _every_ resume? It should be a one-time thing -- could we only do it if the runtime has old certs?

http/src/main/resources/init-resources/startup.sh
Qi77Qi: oh didn't realize some values are optional...I can check..maybe that's why automation test failed

http/src/main/resources/init-resources/startup.sh
Qi77Qi: it's a bit tricky to test if the cert needs to an update tho...I'll give it a try

http/src/main/resources/init-resources/startup.sh
```diff
@@ -64,6 +65,34 @@ export WELDER_MEM_LIMIT=$(welderMemLimit)
...
+notAfter=`openssl x509 -enddate -noout -in /certs/jupyter-server.crt` # output should be something like `notAfter=Jul 22 13:09:15 2023 GMT`
+
+## If cert is old, then pull latest certs. Update date if we need to rotate cert again
+if [[ "$notAfter" != *"notAfter=Jul 22"* ]] ; then
```
rtitle, line 87: Parsing dates seems a little error-prone.. wonder if something like this would be better: https://stackoverflow.com/questions/21297853/how-to-determine-ssl-cert-expiration-date-from-a-pem-encoded-certificate

Maybe:
```
EXPIRED=0
openssl x509 -checkend <3-months-in-sections> -noout -in /certs/jupyter-server.crt || EXPIRED=$?
if [ $EXPIRED -ne 0 ]; then
  # copy the cert
fi
```

http/src/main/resources/init-resources/startup.sh
```diff
@@ -64,6 +65,34 @@ export WELDER_MEM_LIMIT=$(welderMemLimit)
 export MEM_LIMIT=$(memLimit)
 export USE_GCE_STARTUP_SCRIPT=$(useGceStartupScript)
 
+function failScriptIfError() {
+  if [ $EXIT_CODE -ne 0 ]; then
+    echo "Fail to docker-compose start welder ${EXIT_CODE}. Output is saved to ${JUPYTER_START_USER_SCRIPT_OUTPUT_URI}"
```
rtitle, line 70: I thought `JUPYTER_START_USER_SCRIPT_OUTPUT_URI` was just intended for the "restart" user script -- do we want to overload it for other purposes like starting welder?

Also the log message here mentions welder but this function is used for other purposes..

http/src/main/resources/init-resources/startup.sh
```diff
@@ -77,27 +106,23 @@ if [ "$UPDATE_WELDER" == "true" ] ; then
     gcloud auth configure-docker
     docker-compose -f /etc/welder-docker-compose.yaml stop
     docker-compose -f /etc/welder-docker-compose.yaml rm -f
-    docker-compose -f /etc/welder-docker-compose.yaml up -d
+    docker-compose -f /etc/welder-docker-compose.yaml up -d &> start_output.txt || EXIT_CODE=$?
```
rtitle, line 109: Same question as above.. not sure if we want to reuse `start_output.txt` which is intended for user-specified start scripts

rtitle:
Scala changes look good.. couple more questions in `startup.sh`

http/src/main/resources/init-resources/startup.sh
Qi77Qi: I can...but this means we won't be able to validate things in dev...since dev cert won't expire for another almost 3 years (the one you created in June)...

http/src/main/resources/init-resources/startup.sh
Qi77Qi: yea..I realize the original purpose of the output is only for user input startup script..but we actually do care if startup script failed anywhere tho, and we should exit startup instead of currently things just stuck in starting

http/src/main/resources/init-resources/startup.sh
rtitle: Hm yeah that's true..

http/src/main/resources/init-resources/startup.sh
rtitle: Yeah true.. I guess users probably only look at the output log in case of errors, so it's useful to have more types of errors in there.

http/src/main/resources/init-resources/startup.sh
rtitle: Do all the new certs in all envs expire on July 22, 2023?

Qi77Qi: yep..I created them on the same date...

rtitle (approved):
OK, I think these changes look good. We should definitely do some "dry run"s of the migration in dev, the same way we plan to do it in prod. Just trying to be careful since this affects all users.

PR #1517, opened 2020-07-27
IA-2051 Sam client for k8s apps

rtitle:
https://broadworkbench.atlassian.net/browse/IA-2051

* Refactored `LeoAuthProvider`/`SamAuthProvider` a bit:
   * Removed duplicate method signatures (for example `hasRuntimePermission`, `hasPersistentPermission`) in favor of methods with generic types
   * Added `SamResource` and `SamResourceAction` typeclasses to constrain which actions are checkable against which resource types. For example calling `hasPermission` with a runtime resource and a disk action will not compile!
   * Made project-fallback actions and cacheable actions more explicit in method signatures/typeclasses

* Added "app" Sam resource type:
   * It's easy because `LeoAuthProvider` methods are generic: we just need a new typeclass instance
   * Apps in Sam are modeled a little differently: there are NO actions which fall back to the project level; instead we use "creator" and "manager" policies attached to the App resource. This results in fewer Sam calls by Leo and also makes it so permissions on 1 resource don't confer permissions to another resource. I think it would be worthwhile to migrate runtimes and disks to this model too in the future.
   * Related Sam PR: https://github.com/broadinstitute/sam/pull/444

Unit tests pass, opening to get integration test results. I also want to do some manual fiab testing.

---

In all cases:


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/ProxyRoutes.scala
```diff
@@ -52,19 +51,10 @@ class ProxyRoutes(proxyService: ProxyService, corsSupport: CorsSupport)(
             path("setCookie") {
               extractUserInfo { userInfo =>
                 get {
-                  // Check the user for ConnectToCluster privileges and set a cookie in the response
-                  onSuccess {
-                    val authCheck = for {
-                      implicit0(ctx: ApplicativeAsk[IO, AppContext]) <- AppContext.lift[IO]()
-                      _ <- proxyService.authCheck(userInfo, googleProject, runtimeName, ConnectToRuntime)
```
Qi77Qi, line 59: we don't need this authCheck anymore?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -225,6 +225,7 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift](
                                                               10,
                                                               None)(
                     persistentDiskQuery.delete(id, now).transaction.void,
+                    // TODO delete disk resource in Sam
```
rtitle: I noticed we never delete disk resources in Sam. Will add this in

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/DiskServiceInterp.scala
```diff
@@ -148,17 +147,16 @@ class DiskServiceInterp[F[_]: Parallel](config: PersistentDiskConfig,
       // throw 404 if no ReadPersistentDisk permission
       // Note: the general pattern is to 404 (e.g. pretend the disk doesn't exist) if the caller doesn't have
       // ReadPersistentDisk permission. We return 403 if the user can view the disk but can't perform some other action.
-      hasPermission <- authProvider.hasPersistentDiskPermission(disk.samResource,
-                                                                userInfo,
-                                                                ReadPersistentDisk,
-                                                                googleProject)
-      _ <- if (hasPermission) F.unit
+      listOfPermissions <- authProvider.getActionsWithProjectFallback(disk.samResource, googleProject, userInfo)
```
rtitle, line 149: Added the same optimizations for disks as we did for runtimes, to avoid calling Sam twice

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/samModels.scala
```diff
@@ -52,13 +52,118 @@ object AccessPolicyName {
...
+  case object CreatePersistentDisk extends ProjectAction {
+    val asString = "create_persistent_disk"
+  }
+  case object CreateApp extends ProjectAction {
+    val asString = "create_kubernetes_app"
+  }
+
+  // TODO we'd like to remove the below actions at the project level, and control these
+  // actions with policies at the resource level instead. App resources currently follow
+  // this model. See IA-XYZ.
```
rtitle, line 97: Will create an actual ticket for this

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/sam/SamAuthProvider.scala
```diff
@@ -41,273 +30,170 @@ class SamAuthProvider[F[_]: Effect: Logger](samDao: SamDAO[F],
...
-    for {
-      callerActions <- getRuntimeActions(samResource, userInfo)
-      projectPermissions <- samDao
-        .getListOfResourcePermissions(ProjectSamResource(googleProject), authorization)
-      projectActions = projectActionMap.collect { case (k, v) if projectPermissions.contains(v) => k }
-    } yield callerActions ++ projectActions
-  }
-
-  def getRuntimeActions(samResource: RuntimeSamResource, userInfo: UserInfo)(
+  private def checkPermission(samResourceType: SamResourceType,
```
Qi77Qi, line 69: 👍 deleting half the file :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/AppComponent.scala
```diff
@@ -3,30 +3,12 @@ package org.broadinstitute.dsde.workbench.leonardo.db
...
-  KubernetesCluster,
-  KubernetesService,
-  LabelMap,
-  Namespace,
-  NamespaceId,
-  Nodepool,
-  NodepoolLeoId,
-  PersistentDisk
-}
+import org.broadinstitute.dsde.workbench.leonardo._
```
Qi77Qi: we should just update package to
```
package db
```

Qi77Qi (approved):
lgtm..maybe run perf tests as well? since the cache change caused some perf issue before

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/ProxyRoutes.scala
rtitle: Yeah I don't think setCookie ever needed a `ConnectToRuntime` auth check.. it doesn't actually connect to the runtime, it just echoes back the user's auth token as a `Set-Cookie` response header. For k8s I actually moved this route to the top level: `/proxy/setCookie` with no auth check, but we need to keep this path for backwards compatibility.

rtitle:
Thanks @Qi77Qi -- good idea on perf tests, I'll run some (am also testing on a fiab).

rtitle:
Fiab testing looks good. Also ran performance tests:
1. `LeoPauseResumeSimulation`, 200 users
2. `LeoProxySimulation`, 200 users
3 `BVDPNotebooksNightlyPerformanceTest`, 200 users

In all cases success rate, latency seems consistent with previous runs.

PR #1520, opened 2020-07-28
IA-2086 Galaxy proxy error when importing a workflow

rtitle:
https://broadworkbench.atlassian.net/browse/IA-2086

I think the Leo proxy was not handling encoded query string parameters correctly, like: 
```
?message=Imported,%20but%20some%20steps%20in%20this%20workflow%20have%20validation%20errors.&status=success
```
Interestingly I think this bug has existed in Leo for a long time, but Galaxy is the first thing to expose it.

I manually deployed this to the `galaxy-dev-env` fiab, and it seems to work.

---

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/AppComponent.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/AppComponent.scala
@@ -45,7 +45,7 @@ final case class AppRecord(id: AppId,
                            dateAccessed: Instant,
                            namespaceId: NamespaceId,
                            diskId: Option[DiskId],
-                           customEnvironmentVariables: Map[String, String])
+                           customEnvironmentVariables: Option[Map[String, String]])
 
 class AppTable(tag: Tag) extends Table[AppRecord](tag, "APP") {
   //unique (appName, destroyedDate)
@@ -61,7 +61,7 @@ class AppTable(tag: Tag) extends Table[AppRecord](tag, "APP") {
   def dateAccessed = column[Instant]("dateAccessed", O.SqlType("TIMESTAMP(6)"))
   def namespaceId = column[NamespaceId]("namespaceId", O.Length(254))
   def diskId = column[Option[DiskId]]("diskId", O.Length(254))
-  def customEnvironmentVariables = column[Map[String, String]]("customEnvironmentVariables")
+  def customEnvironmentVariables = column[Option[Map[String, String]]]("customEnvironmentVariables")
 
   def * =
     (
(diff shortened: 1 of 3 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/AppComponent.scala
```diff
@@ -61,7 +61,7 @@ class AppTable(tag: Tag) extends Table[AppRecord](tag, "APP") {
   def dateAccessed = column[Instant]("dateAccessed", O.SqlType("TIMESTAMP(6)"))
   def namespaceId = column[NamespaceId]("namespaceId", O.Length(254))
   def diskId = column[Option[DiskId]]("diskId", O.Length(254))
-  def customEnvironmentVariables = column[Map[String, String]]("customEnvironmentVariables")
+  def customEnvironmentVariables = column[Option[Map[String, String]]]("customEnvironmentVariables")
```
rtitle, line 64: This was actually throwing an error when I deployed it to my fiab:
```
Caused by: slick.SlickException: Read NULL value (null) for ResultSet column <computed>
```
The column needs to be `Option` in slick because it's nullable in mysql.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/ProxyService.scala
```diff
@@ -284,7 +284,7 @@ class ProxyService(
     // 1. filter out headers not needed for the backend server
     val newHeaders = filterHeaders(request.headers)
     // 2. strip out Uri.Authority:
-    val newUri = Uri(path = rewrittenPath, queryString = request.uri.queryString())
+    val newUri = Uri(path = rewrittenPath, queryString = request.uri.rawQueryString)
```
rtitle, line 287: This is the actual fix: `queryString()` returns the _decoded_ query string, so we were mistakenly passing the decoded value to the Galaxy server

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/AppComponent.scala
```diff
@@ -45,7 +45,7 @@ final case class AppRecord(id: AppId,
                            dateAccessed: Instant,
                            namespaceId: NamespaceId,
                            diskId: Option[DiskId],
-                           customEnvironmentVariables: Map[String, String])
+                           customEnvironmentVariables: Option[Map[String, String]])
```
Qi77Qi, line 48: can we make remove `Option` in the data type, and only have `Option` in `def customEnvironmentVariables = column[Option[Map[String, String]]]("customEnvironmentVariables")`

rtitle: We could but then can't use `<> (AppRecord.tupled, AppRecord.unapply)`. I guess another approach is to have some data type for env vars instead of `Map[String, String]`

Qi77Qi: mm I see...

Qi77Qi (approved):

PR #1530, opened 2020-07-29
[no ticket] bake aou 1.0.4 version

Qi77Qi:
need to check image size after this PR to see actually custom image and update bootDiskSize if necessary

This PR will
1. cache aou 0.0.1 and 1.0.4 for dataproc image
2. cache latest aou for gce image

---

In all cases:

### DIFF
--- a/jenkins/Jenkinsfile
+++ b/jenkins/Jenkinsfile
@@ -258,7 +258,9 @@ pipeline {
                                 if (GENERATE_DATAPROC_IMAGE == "true" && dataprocImages.contains(image)) {
                                     println("Using version $version for $image in this build of custom Dataproc images")
                                     sh """
-                                        sed -i "s/$image.*/$image:$version\\"/" jenkins/dataproc-custom-images/prepare-custom-leonardo-jupyter-dataproc-image.sh
+                                       if [ $image != 'terra-jupyter-aou' ]; then
+                                          sed -i "s/$image.*/$image:$version\\"/" jenkins/dataproc-custom-images/prepare-custom-leonardo-jupyter-dataproc-image.sh
+                                       fi
                                     """
                                 }
 
@@ -266,9 +268,7 @@ pipeline {
                                     println("Using version $version for $image in this build of custom GCE images")
                                     // Not replace aou version because we want to fix bake 2 aou versions until aou migrate to new version
                                     sh """
-                                        if [ $image != 'terra-jupyter-aou' ]; then
-                                            sed -i "s/$image.*/$image:$version\\"/" jenkins/gce-custom-images/prepare_custom_leonardo_gce_image.sh
-                                        fi
+                                       sed -i "s/$image.*/$image:$version\\"/" jenkins/gce-custom-images/prepare_custom_leonardo_gce_image.sh
                                     """
                                 }
                             }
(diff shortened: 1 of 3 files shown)


jenkins/dataproc-custom-images/prepare-custom-leonardo-jupyter-dataproc-image.sh
```diff
@@ -23,18 +23,18 @@ terra_jupyter_r="us.gcr.io/broad-dsp-gcr-public/terra-jupyter-r:0.0.11"
 terra_jupyter_bioconductor="us.gcr.io/broad-dsp-gcr-public/terra-jupyter-bioconductor:0.0.12"
 terra_jupyter_hail="us.gcr.io/broad-dsp-gcr-public/terra-jupyter-hail:0.0.9"
 terra_jupyter_gatk="us.gcr.io/broad-dsp-gcr-public/terra-jupyter-gatk:0.0.13"
-terra_jupyter_aou="us.gcr.io/broad-dsp-gcr-public/terra-jupyter-aou:0.0.1"
+terra_jupyter_aou_old="us.gcr.io/broad-dsp-gcr-public/terra-jupyter-aou:0.0.1"
```
rtitle, line 26: Wondering if this will break the `sed` script in the `Jenkinsfile`:
```
 sed -i "s/$image.*/$image:$version\\"/" jenkins/dataproc-custom-images/prepare-custom-leonardo-jupyter-dataproc-image.sh
```
Would this mistakenly catch both `terra_jupyter_aou` and `terra_jupyter_aou_old` lines?

The sed script could probably be more robust..

Qi77Qi: hmm didn't realize there's a sed in Jenkinsfile...we should probably hold on this since aou is not sure r major version bump is desirable for them yet

Qi77Qi: I didn't update regex, but instead made sed conditional in jenkinsfile (basically if it's aou, we don't do version replacement)

jdcanas (approved):

jenkins/Jenkinsfile
```diff
@@ -260,14 +258,17 @@ pipeline {
...
+                                       fi
                                     """
                                 }
 
                                 if (GENERATE_GCE_IMAGE == "true" && gceImages.contains(image)) {
                                     println("Using version $version for $image in this build of custom GCE images")
+                                    // Not replace aou version because we want to fix bake 2 aou versions until aou migrate to new version
                                     sh """
-                                        sed -i "s/$image.*/$image:$version\\"/" jenkins/gce-custom-images/prepare_custom_leonardo_gce_image.sh
+                                       sed -i "s/$image.*/$image:$version\\"/" jenkins/gce-custom-images/prepare_custom_leonardo_gce_image.sh
```
jdcanas, line 271: you will need the if for AOU here as well

Qi77Qi: aou is still using dataproc..they'll probably migrate to newer aou image before migrating to gce..so I just made gce to cache latest aou image....

PR #1533, opened 2020-07-30
IA-2096 fix delete disk flag

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-2096

The bug is in frontend we've detached the persistent disk, hence in back leo, if we try to get the associated persistent disk, it won't be there to delete. Added automation test for deleteDisk flag

---

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
@@ -207,46 +207,38 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift](
         case None =>
           runtimeConfig.cloudService.process(runtime.id, RuntimeStatus.Deleting).compile.drain
       }
-      fa = if (msg.deleteDisk)
-        runtimeConfig match {
-          case rc: RuntimeConfig.GceWithPdConfig =>
-            for {
-              _ <- poll
-              now <- nowInstant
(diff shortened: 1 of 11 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -206,43 +206,35 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift](
...
+          disk <- F.fromEither(diskOpt.toRight(new RuntimeException(s"disk not found for ${id}")))
+
+          deleteDiskOp <- googleDiskService.deleteDisk(runtime.googleProject, disk.zone, disk.name)
+          _ <- computePollOperation.pollZoneOperation(runtime.googleProject,
+                                                      disk.zone,
+                                                      OperationName(deleteDiskOp.getName),
+                                                      2 seconds,
+                                                      10,
+                                                      None)(
+            persistentDiskQuery.delete(id, now).transaction.void,
```
rtitle: By the way, I updated this line in my PR.. I think the Sam disk resource needs to be deleted here as well (we currently never delete disk Sam resources). I can rebase if this goes first.

rtitle (approved):

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
Qi77Qi: that's good to know...I'll remember to add this if yours go in first

PR #1538, opened 2020-07-31
QA-1205 - adding function to override notebook changed modal

xuscrbw:
Disabling autosave didn't seem to work consistently.
I adapted the dismissNotebookChange function to now look specifically for the button rather than the modal (since the modal name is not unique. The button is unique). It will only trigger a click if it finds the button, which means this should cover all cases

---

In all cases:


Qi77Qi:
can you put a title on the PR...also shall we revert the previous change that disables saving?

xuscrbw:
removed the disable sleep line

Qi77Qi (approved):

rtitle (approved):
LGTM.. I think I added `dismissNotebookChanged`, feel free to remove it if this method is better

PR #1545, opened 2020-08-04
[QA-1221] Retry getDisk() in RuntimeCreationDiskSpec

kyuksel:
[JIRA ticket](https://broadworkbench.atlassian.net/browse/QA-1221)

At the end of a [RuntimeCreationDiskSpec test](https://github.com/DataBiosphere/leonardo/blob/2034bb8260dad82fae7250207a3d203a185d162e/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/runtimes/RuntimeCreationDiskSpec.scala#L124), we’re deleting a runtime as well as the attached disk. Once we find out that the runtime is deleted, we check if the disk is deleted. The flakiness appears to be due to that sometimes the disk isn’t yet deleted right then. Therefore, we’ve added a retry for a few more times. From a user experience perspective, it shouldn’t be a big deal if the deletion of their disk takes another 5-10 seconds.

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/runtimes/RuntimeCreationDiskSpec.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/runtimes/RuntimeCreationDiskSpec.scala
@@ -187,8 +187,9 @@ class RuntimeCreationDiskSpec
           }
         })
         _ <- deleteRuntimeWithWait(googleProject, runtimeWithDataName, deleteDisk = true)
-        ioa = getDisk(googleProject, diskName).attempt
-        diskResp <- streamFUntilDone(ioa, 5, 10 seconds).compile.lastOrError
+        getDiskAttempt = getDisk(googleProject, diskName).attempt
+        // Disk deletion may take some time so we're retrying to reduce flaky test failures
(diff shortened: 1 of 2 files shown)


Qi77Qi (approved):
lgtm 👍  thanks for fixing!

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/runtimes/RuntimeCreationDiskSpec.scala
```diff
@@ -186,9 +186,10 @@ class RuntimeCreationDiskSpec
             notebookPage.executeCell(persistedPackage).get should include("/home/jupyter-user/notebooks/packages")
           }
         })
-        _ <- testTimer.sleep(5 seconds) //runtime status updates to Deleted before disk status change, hence add a sleep
         _ <- deleteRuntimeWithWait(googleProject, runtimeWithDataName, deleteDisk = true)
-        diskResp <- getDisk(googleProject, diskName).attempt
+        getDiskAttempt = getDisk(googleProject, diskName).attempt
+        // Disk deletion may take some time so we're retrying to reduce flaky test failures
+        diskResp <- streamFUntilDone(getDiskAttempt, 5, 5 seconds).compile.lastOrError
```
Qi77Qi, line 192: no big deal..but maybe lower the interval to 3 seconds?

PR #1547, opened 2020-08-05
IA-2045 fix runtimes stuck in Stopping, Deleting

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-2045. 

* My theory on the bug is error occurred before we update DB, but after google Delete/Stop completed. Hence we never updated DB status properly (altho the chance seems to me pretty low)
* Made `MonitorAtBoot` to resend a message, if a runtime is in `Stopping`, `Deleting`, `Creating`, `Starting`. This should make things more robust.
* Updated error handle in subscriber for runtime messages so that errors happen in async task will be persisted in DB

---

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
@@ -813,52 +813,6 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift](
       )
     } yield ()
 
-  private def deleteDiskForApp(diskId: DiskId)(
-    implicit ev: ApplicativeAsk[F, AppContext]
-  ): F[Unit] =
-    deleteDisk(diskId, true)
-
-  private def createDiskForApp(msg: CreateAppMessage)(
-    implicit ev: ApplicativeAsk[F, AppContext]
-  ): F[Unit] =
-    msg.createDisk match {
-      case true =>
-        for {
-          _ <- logger.info(s"Beginning disk creation for app ${msg.appId}")
-          ctx <- ev.ask
-          getAppOpt <- KubernetesServiceDbQueries.getActiveFullAppByName(msg.project, msg.appName).transaction
-          getApp <- F.fromOption(getAppOpt, AppNotFoundException(msg.project, msg.appName, ctx.traceId))
-          disk <- F.fromOption(
-            getApp.app.appResources.disk,
-            AppCreationException(
-              s"create disk was true for create app message, but app ${getApp.app.id} does not have a disk id saved"
-            )
-          )
-          _ <- createDisk(CreateDiskMessage.fromDisk(disk, Some(ctx.traceId)), true)
-        } yield ()
-      case false => F.unit
-    }
-
-  private def handleKubernetesError(e: Throwable): F[Unit] =
-    e match {
-      case e: PubsubKubernetesError =>
-        for {
-          _ <- appErrorQuery.save(e.appId, e.dbError).transaction
-          _ <- appQuery.updateStatus(e.appId, AppStatus.Error).transaction
-          _ <- e.clusterId.traverse(clusterId =>
-            kubernetesClusterQuery.updateStatus(clusterId, KubernetesClusterStatus.Error).transaction
-          )
-          _ <- e.nodepoolId.traverse(nodepoolId =>
-            nodepoolQuery.updateStatus(nodepoolId, NodepoolStatus.Error).transaction
-          )
-          _ <- logger.error(e)(s"updating db state for an async error for app ${e.appId}")
-        } yield ()
-      case _ =>
-        F.raiseError(
-          new RuntimeException(s"handleKubernetesError should not be used with a non kubernetes error. Error: ${e}")
-        )
-    }
-
   private def createRuntimeErrorHandler(msg: CreateRuntimeMessage, now: Instant)(e: Throwable): F[Unit] =
     for {
       _ <- logger.error(e)(s"Failed to create runtime ${msg.runtimeProjectAndName} in Google")
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
@@ -13,6 +13,7 @@ import cats.mtl.ApplicativeAsk
 import com.google.cloud.compute.v1.Disk
 import fs2.Stream
 import fs2.concurrent.InspectableQueue
+import org.broadinstitute.dsde.workbench.errorReporting.ErrorReporting
 import org.broadinstitute.dsde.workbench.leonardo.AsyncTaskProcessor.Task
 import org.broadinstitute.dsde.workbench.google2.{
   ComputePollOperation,
@@ -23,6 +24,7 @@ import org.broadinstitute.dsde.workbench.google2.{
   MachineTypeName,
   OperationName
 }
+import org.broadinstitute.dsde.workbench.errorReporting.ReportWorthySyntax._
 import org.broadinstitute.dsde.workbench.leonardo.db._
 import org.broadinstitute.dsde.workbench.leonardo.http.{cloudServiceSyntax, _}
 import org.broadinstitute.dsde.workbench.leonardo.model.{LeoAuthProvider, LeoException}
@@ -42,7 +44,8 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift](
   googleDiskService: GoogleDiskService[F],
   computePollOperation: ComputePollOperation[F],
   authProvider: LeoAuthProvider[F],
-  gkeInterp: GKEInterpreter[F]
+  gkeInterp: GKEInterpreter[F],
+  errorReporting: ErrorReporting[F]
 )(implicit executionContext: ExecutionContext,
   F: ConcurrentEffect[F],
   parallel: Parallel[F],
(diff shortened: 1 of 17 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -762,6 +723,78 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift](
...
+          )
+          _ <- logger.error(e)(s"updating db state for an async error for app ${e.appId}")
+        } yield ()
+      case _ =>
+        F.raiseError(
+          new RuntimeException(s"handleKubernetesError should not be used with a non kubernetes error. Error: ${e}")
+        )
+    }
+
+  private def createRuntimeErrorHandler(msg: CreateRuntimeMessage, now: Instant)(e: Throwable): F[Unit] =
```
rtitle, line 816: Just curious.. if create runtime fails, do we clean up associated Google resources that might have been created (e.g. the staging/init buckets). That's probably the job of `Gce/DataprocInterpreter` not `LeoPubsubMessageSubscriber`.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/MonitorAtBoot.scala
```diff
@@ -30,12 +30,14 @@ class MonitorAtBoot[F[_]: Timer](implicit F: Async[F],
           clusters.toList.traverse_ {
             case c if c.status.isMonitored && c.status != RuntimeStatus.Unknown =>
               val r = for {
-                _ <- c.cloudService.process(c.id, c.status).compile.drain
+                tid <- traceId.ask
+                message <- runtimeStatusToMessage(c, tid)
+                // If a runtime is in transition status (Creating, Starting etc), then we're enqueue a pubsub message again
+                _ <- message.traverse(m => publisherQueue.enqueue1(m))
```
rtitle, line 36: So this will make Front Leo re-send messages for transition statuses? Back Leo should already start monitoring them at boot time, right? I guess it doesn't hurt to have both checks.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GceInterpreter.scala
```diff
@@ -314,8 +318,15 @@ class GceInterpreter[F[_]: Parallel: ContextShift: Logger](
     if (params.isAsyncRuntimeFields)
       googleComputeService
         .deleteInstance(params.googleProject, config.gceConfig.zoneName, InstanceName(params.runtimeName.asString))
-        .map(x => Some(x))
-    else Async[F].pure(None)
+        .map(x => Option(x))
+        .handleErrorWith {
+          case _: com.google.api.gax.rpc.NotFoundException =>
```
rtitle, line 323: Good catch.. probably would be good to have wb-libs `GoogleComputeInterpreter.deleteInstance` recover 404s. But can leave this here for now

rtitle (approved):

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
Qi77Qi: I don't think it clean up resources currently..but yea, I can look into it and add it if it's easy

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/MonitorAtBoot.scala
Qi77Qi: it doesn't involve front leo...Publisher stream runs on both front and back leo....We can technically enqueue this directly to subscriber's queue as well...but since currently subscriber is defined in a way that only processes google events, so it'll be a bit more refactoring to do that.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GceInterpreter.scala
Qi77Qi: I thought about adding it in `deleteInstance`, but if I do that, we'll need to return `F[Option[Operation]]` because we won't get a `Operation` when 404...I'm not entirely sure if `F[Option[Operation]]` is desirable, so I just added the handler here for now.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/MonitorAtBoot.scala
rtitle: Ah I see - thanks for the clarification

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GceInterpreter.scala
rtitle: I see - yeah that recover behavior in wb-libs is typically when it returns `F[Unit]`

Qi77Qi:
jenkins restart...fiab start failure

Qi77Qi:
jenkins retest...not restart

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -803,6 +813,38 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift](
...
+          None //this could happen when pubsub redelivers an event unexpectedly
+        case _ =>
+          Some(s"Failed to create cluster ${msg.runtimeProjectAndName} due to ${e.getMessage}")
+      }
+      _ <- errorMessage.traverse(m =>
+        (clusterErrorQuery.save(msg.runtimeId, RuntimeError(m, -1, now)) >>
+          clusterQuery.updateClusterStatus(msg.runtimeId, RuntimeStatus.Error, now)).transaction[F]
+      )
+      _ <- if (e.isReportWorthy)
+        errorReporting.reportError(e)
```
rtitle, line 833: Cool -- is `LeoPubsubMessageSubscriber` the only place where we do error reporting currently? I'd love to see a demo at some point (maybe you've given one already and I missed it..)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/MonitorAtBoot.scala
```diff
@@ -63,6 +65,73 @@ class MonitorAtBoot[F[_]: Timer](implicit F: Async[F],
...
+        for {
+          fullRuntime <- clusterQuery.getClusterById(runtime.id).transaction
+          rt <- F.fromOption(fullRuntime, new Exception(s"can't find ${runtime.id} in DB"))
+          rtConfig <- RuntimeConfigQueries.getRuntimeConfig(rt.runtimeConfigId).transaction
+          r = rtConfig match {
+            case x: RuntimeConfig.GceConfig =>
+              for {
+                bootDiskSize <- x.bootDiskSize.toRight(
+                  s"disk Size field not found for ${rt.id}. This should never happen"
+                ) //TODO: report error
```
rtitle, line 102: Remove TODO? Is this covered by the above error reporting?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
Qi77Qi: yea, currently I'm only reporting the edge cases that I had left TODO previously (they're from monitor class, but will bubble up to subscriber and handled here)...

I gave a demo in today's walkthrough, but I can demo again during sprint review.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/BucketHelper.scala
```diff
@@ -137,8 +127,10 @@ class BucketHelper[F[_]: Concurrent: ContextShift: Logger](config: BucketHelperC
           config.clusterFilesConfig.proxyRootCaPem
         )
       ) ++ rstudioLicenseFile
-      bytes <- Stream.eval(TemplateHelper.fileStream[F](f, blocker).compile.to(Array))
-      _ <- storeObject(initBucketName, GcsBlobName(f.getFileName.toString), bytes, "text/plain")
+      _ <- TemplateHelper.fileStream[F](f, blocker) through google2StorageDAO.streamUploadBlob(
```
rtitle, line 130: Nice, thanks for updating to not read all this data in memory

project/Dependencies.scala
```diff
@@ -131,7 +136,7 @@ object Dependencies {
     workbenchGoogle2Test,
     workbenchOpenTelemetry,
     workbenchOpenTelemetryTest,
-    "net.logstash.logback" % "logstash-logback-encoder" % "6.2", // for structured logging in logback
+//    "net.logstash.logback" % "logstash-logback-encoder" % "6.2", // for structured logging in logback
```
rtitle, line 139: This isn't needed for logstash? (I don't know where it came from originally)

rtitle (approved):
Latest changes LGTM

project/Dependencies.scala
Qi77Qi: think I added it a while ago for logging things in json format..since we're not doing that, we don't need this in...I didn't delete it just for future reference

Qi77Qi:
jenkins retest

Qi77Qi:
going merge this as is..this PR currently is using a snap version of https://github.com/broadinstitute/workbench-libs/pull/343, but I'll update google2 version pretty soon in another open PR

PR #1549, opened 2020-08-06
[no ticket] pre-create runtime in LeonardoSuite beforeAll

rtitle:
Create an "initial" runtime in `LeonardoSuite.beforeAll`. The idea is that this should set up necessary networks, subnets, etc to prevent concurrency issues. We can remove this once https://broadworkbench.atlassian.net/browse/IA-2121 is done.

---

In all cases:

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoSuite.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoSuite.scala
@@ -1,14 +1,12 @@
 package org.broadinstitute.dsde.workbench.leonardo
 
-import org.broadinstitute.dsde.workbench.fixture.BillingFixtures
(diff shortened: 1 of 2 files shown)


automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoSuite.scala
```diff
@@ -44,53 +48,120 @@ object GPAllocFixtureSpec {
...
+            IO(logger.warn(
+                 s"Failed to create initial runtime ${project.value} / ${initalRuntimeName.asString} with error"
+               ),
+               err)
+        }
+      } yield ()
+    }
+
+  private def deleteInitialRuntime(project: GoogleProject): IO[Unit] =
+    LeonardoApiClient.client.use { c =>
```
Qi77Qi: fyi, you can do `implicit c =>` directly so that you don't have to assign it to an implicit variable later

Qi77Qi (approved):

PR #1550, opened 2020-08-07
IA-2037- Adding appRoutes to LeoAPIClient and core module

xuscrbw:
Added in the appRoutes to the Core Module and APIClient so it can be used across multiple directory.

Still working on creating some initial tests to make sure the added routes can work properly.

Also added in a proxyUrl decoder
---

In all cases:

### DIFF
--- a/core/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/AppRoutesTestJsonCodec.scala
+++ b/core/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/AppRoutesTestJsonCodec.scala
@@ -24,21 +24,17 @@ object AppRoutesTestJsonCodec {
     )
   )
 
-  implicit val proxyUrlDecoder: Decoder[Map[ServiceName, URL]] = Decoder.decodeMap[ServiceName, URL](KeyDecoder.decodeKeyString.map(ServiceName), urlDecoder)
+  implicit val proxyUrlDecoder: Decoder[Map[ServiceName, URL]] =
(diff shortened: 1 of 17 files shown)


xuscrbw:
jenkins retest

xuscrbw:
jenkins retest

core/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/AppRoutesTestJsonCodec.scala
```diff
@@ -0,0 +1,40 @@
...
+
+import java.net.URL
+
+import io.circe.{Decoder, Encoder, KeyDecoder}
+import org.broadinstitute.dsde.workbench.google2.KubernetesSerializableName.ServiceName
+import org.broadinstitute.dsde.workbench.leonardo.JsonCodec._
+
+object AppRoutesTestJsonCodec {
+
+  implicit val createAppEncoder: Encoder[CreateAppRequest] = Encoder.forProduct5(
```
jdcanas, line 11: These may be only used here in tests, but we could also move these encoder/decoders to the core JsonCodec instead. Could be helpful later, up to you.

jdcanas (approved):
Looks good to me.

core/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/AppRoutesTestJsonCodec.scala
Qi77Qi: in app, I don't think we'll ever need to encode a `CreateAppRequest`....having it in test is good.

Qi77Qi (approved):
lgtm

PR #1551, opened 2020-08-10
[IA-2070] batch create nodepool/ claiming nodepool functionality

jdcanas:


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesServiceDbQueries.scala
```diff
@@ -144,10 +154,21 @@ object KubernetesServiceDbQueries {
           _ <- nodepoolQuery.updateStatus(createCluster.nodepoolId, NodepoolStatus.Provisioning)
```
rtitle: To check my understanding: is `createCluster.nodepoolId` the "dummy" nodepool, distinct from the `nodepoolId` passed to this method?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesServiceDbQueries.scala
```diff
@@ -144,10 +154,21 @@ object KubernetesServiceDbQueries {
...
         } yield ()
       )
-      _ <- nodepoolQuery.updateStatus(nodepoolId, NodepoolStatus.Provisioning)
+      _ <- nodepoolId.traverse(id => nodepoolQuery.updateStatus(id, NodepoolStatus.Provisioning))
       _ <- appQuery.updateStatus(appId, AppStatus.Provisioning)
     } yield ()
 
+  def markPendingCreating(clusterId: KubernetesClusterLeoId, nodepoolIds: List[NodepoolLeoId])(
```
rtitle: This is for the batch create, right? Could we maybe name the method differently instead of overloading `markPendingCreating`?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesServiceDbQueries.scala
```diff
@@ -309,7 +330,7 @@ object KubernetesServiceDbQueries {
         case (appRec, (services, namespaces, labelMap, diskMap, errors)) =>
           appQuery.unmarshalApp(
             appRec,
-            services.map(serviceQuery.unmarshalService).toList,
+            services.map(serviceQuery.unmarshalService).toList.toSet.toList,
```
rtitle, line 333: `services.map(serviceQuery.unmarshalService).toSet.toList`?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/NodepoolComponent.scala
```diff
@@ -86,27 +90,53 @@ object nodepoolQuery extends TableQuery(new NodepoolTable(_)) {
...
-          n.autoscalingConfig.map(_.autoscalingMax),
-          n.isDefault
-        )
+        fromNodepool(n)
     } yield n.copy(id = nodepoolId)
 
+  def markAsUnclaimed(ids: List[NodepoolLeoId])(implicit ec: ExecutionContext): DBIO[Unit] =
+    for {
+      _ <- nodepoolQuery
+        .filter(_.id.inSet(ids.toSet))
```
rtitle: Maybe use `inSetBind` -- IIRC `inSet` is the unsafe/unchecked version of the SQL `IN` clause

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/KubernetesServiceInterp.scala
```diff
@@ -97,8 +101,11 @@ class LeoKubernetesServiceInterp[F[_]: Parallel](
       else F.unit
 
       clusterId = saveClusterResult.minimalCluster.id
-      saveNodepool <- F.fromEither(getUserNodepool(clusterId, userInfo, req, ctx.now))
-      nodepool <- nodepoolQuery.saveForCluster(saveNodepool).transaction
+      claimedNodepoolOpt <- nodepoolQuery.claimNodepool(clusterId).transaction
+      nodepool <- claimedNodepoolOpt.fold(for {
+        saveNodepool <- F.fromEither(getUserNodepool(clusterId, userInfo, req.kubernetesRuntimeConfig, ctx.now))
+        savedNodepool <- nodepoolQuery.saveForCluster(saveNodepool).transaction
+      } yield savedNodepool)(n => log.info(s"claimed nodepool ${n.id}") >> F.pure(n))
```
rtitle: Might be good to put more information in the log statement, like `"Claimed nodepool {} for {project} / {app}"`

rtitle: Also I had a bit of trouble reading this, maybe a `match` statement would be clearer than `fold`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/KubernetesServiceInterp.scala
```diff
@@ -97,8 +101,11 @@ class LeoKubernetesServiceInterp[F[_]: Parallel](
       else F.unit
 
       clusterId = saveClusterResult.minimalCluster.id
-      saveNodepool <- F.fromEither(getUserNodepool(clusterId, userInfo, req, ctx.now))
-      nodepool <- nodepoolQuery.saveForCluster(saveNodepool).transaction
+      claimedNodepoolOpt <- nodepoolQuery.claimNodepool(clusterId).transaction
```
rtitle, line 104: So this should transactionally claim a nodepool if one exists -- I checked and our isolation level is `RepeatableRead` which I think should be good. 👍

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/KubernetesServiceInterp.scala
```diff
@@ -235,9 +243,43 @@ class LeoKubernetesServiceInterp[F[_]: Parallel](
...
+      msg = BatchNodepoolCreateMessage(createCluster.clusterId, allNodepoolIds, googleProject, Some(ctx.traceId))
+      _ <- publisherQueue.enqueue1(msg)
+    } yield ()
+
+  private[service] def getSavableCluster(
+    userInfo: UserInfo,
+    googleProject: GoogleProject,
+    now: Instant,
+    numNodepools: Option[NumNodepools]
+  ): Either[Throwable, SaveKubernetesCluster] = {
```
rtitle, line 289: Wondering why this returns `Either`.. looks like just because of `KubernetesNameUtils.getUniqueName`, but shouldn't we always be able to generate a valid name?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -118,18 +118,29 @@ class GKEInterpreter[F[_]: Parallel: ContextShift](
...
-  def createCluster(createMessage: CreateCluster, appId: AppId)(
+  def createCluster(clusterId: KubernetesClusterLeoId,
+                    nodepoolsToCreate: List[NodepoolLeoId],
+                    isNodepoolPrecreate: Boolean)(
     implicit ev: ApplicativeAsk[F, AppContext]
   ): F[CreateClusterResult] =
     for {
-      _ <- logger.info(s"beginning cluster creation for app ${appId}")
-      clusterOpt <- kubernetesClusterQuery.getMinimalClusterById(createMessage.clusterId).transaction
+      _ <- logger.info(s"beginning cluster creation for cluster ${clusterId}")
```
rtitle: Can we add project to this log msg?

(Thinking down the line to when we have to debug stuff :))

rtitle (approved):
Some nits above but nothing blocking -- looks good, nice work

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/DefaultNodepoolConfig.scala
```diff
@@ -3,4 +3,9 @@ package org.broadinstitute.dsde.workbench.leonardo.config
 import org.broadinstitute.dsde.workbench.google2.MachineTypeName
 import org.broadinstitute.dsde.workbench.leonardo.NumNodes
 
-case class DefaultNodepoolConfig(machineType: MachineTypeName, numNodes: NumNodes, autoscalingEnabled: Boolean)
+//this is the maximum number of nodepools that can run on a cluster the given default nodepool machine type, used to determine how many nodes the default nodepool needs
+case class MaxNodepoolsPerDefaultNode(value: Int)
```
Qi77Qi: final, extends AnyVal

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesServiceDbQueries.scala
```diff
@@ -134,7 +135,16 @@ object KubernetesServiceDbQueries {
...
 
-  def markPendingCreating(nodepoolId: NodepoolLeoId, appId: AppId, createCluster: Option[CreateCluster])(
+  def getAllNodepoolsForCluster(
+    clusterId: KubernetesClusterLeoId
+  )(implicit ec: ExecutionContext): DBIO[List[Nodepool]] =
+    for {
+      nodepools <- nodepoolQuery
+        .findActiveByClusterIdQuery(clusterId)
+        .result
+    } yield nodepools.map(rec => unmarshalNodepool(rec, List.empty)).toList
```
Qi77Qi, line 145: if we need to sub out `List.empty`, can we just return the actual data type we're retrieving?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/KubernetesServiceInterp.scala
```diff
@@ -235,9 +243,43 @@ class LeoKubernetesServiceInterp[F[_]: Parallel](
...
+      // create list of Precreating nodepools
+      eitherNodepoolsOrError = List.tabulate(req.numNodepools.value) { _ =>
+        getUserNodepool(createCluster.clusterId, userInfo, req.kubernetesRuntimeConfig, ctx.now)
+      }
+      nodepools <- eitherNodepoolsOrError.traverse(n => F.fromEither(n))
+      _ <- nodepoolQuery.saveAllForCluster(nodepools).transaction
+      dbNodepools <- KubernetesServiceDbQueries.getAllNodepoolsForCluster(createCluster.clusterId).transaction
+      allNodepoolIds = dbNodepools.map(_.id)
+      msg = BatchNodepoolCreateMessage(createCluster.clusterId, allNodepoolIds, googleProject, Some(ctx.traceId))
+      _ <- publisherQueue.enqueue1(msg)
```
Qi77Qi, line 281: we're enqueuing pub sub messages in the  `KubernetesServiceInterp` instead of front leo?

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -1237,6 +1237,45 @@ paths:
             - openid
             - email
             - profile
+  "/api/google/v1/apps/{googleProject}/batchNodepoolCreate":
```
kyuksel, line 1240: I like the term "batch" create as it conveys the full meaning without reference to any other information. I know we've been referring to it as "pre"create for historical reasons but "pre" is a relative term and I wonder if it's better to not have any such relativity as this endpoint is not concerned with how those batch created nodepools are utilized. If agreed, it'd be nice to use that term consistently across the implementation, too.

kyuksel (commented):
LGTM with pending comments from Qi and Rob. One semantic comment inline.

http/src/main/resources/swagger/api-docs.yaml
Qi77Qi: what's the relevant between `pre` and `batch` here?

kyuksel: Sorry, I should've made the comment stand-alone instead of inline. I was referring to the seemingly interchangeable usage of the terms _pre_ and _batch_ throughout the code, not specifically to the line above.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesServiceDbQueries.scala
jdcanas: its more expensive to include the list since its apps, and apps require a lot f joining (app, disk, label, service, namespace). Only af few queries need all the joins

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesServiceDbQueries.scala
jdcanas: Correct. I renamed it to `defaultNodepoolId` in another revision but must have stashed it. The optional nodepoolId is one for standalone nodepool creation if there are no pre-created ones

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesServiceDbQueries.scala
jdcanas: Since for performance reasons its a chain prior to this, and Chain doesn't support toSet, we have to do it this way.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/KubernetesServiceInterp.scala
jdcanas: The `KubernetesServiceInterp` is front leo. You may be thinking of the GKEInterpreter which we worked on recently which is back leo

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/KubernetesServiceInterp.scala
jdcanas: We generate a UUID that should always be valid, but the creation for the name case class enforces the validation and returns an Either. Most Kubernetes names follow this pattern because kubernetes does a regex validation of names.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesServiceDbQueries.scala
Qi77Qi: I understand joining is costly....I meant to return the data type without `List`...like straight-up db record instead

PR #1554, opened 2020-08-12
IA-2130 get rid of customizable submodule versions

rtitle:
Leo has submodules for `cis-harden-images` and `terra-cryptomining-security-alerts`. Previously the Jenkins job also took parameters to specify the versions of these libraries. I think this caused https://broadworkbench.atlassian.net/browse/IA-2130 because the submodule was the right version, but the job checked out an out-of-date `master` branch after running `git submodule init` and `git fetch`.

To simplify things I think we can just rely on the submodule versions for these tools and remove the extra params passed to the Jenkins job. If we want to update the version we should just need to update the submodule hash.

---

In all cases:

### DIFF
--- a/jenkins/Jenkinsfile
+++ b/jenkins/Jenkinsfile
@@ -141,10 +141,6 @@ pipeline {
             description: "The path for the bucket where the custom GCE images will be stored")
         string(name: "DATAPROC_VERSIONS", defaultValue: "1.2.79-debian9 1.4.15-debian9",
             description: "A custom image will be build for each of these dataproc versions")
-        string(name: "IMAGE_HARDENING_LIB_VERSION", defaultValue: "100d620",
-            description: "The branch or hash of the 'broadinstitute/dsp-appsec-base-image-hardening' repo that contains libraries for CIS-hardening GCE images")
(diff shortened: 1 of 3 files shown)


Qi77Qi (approved):

kyuksel:
With this change, are we no longer pegging the submodules to specific versions?

rtitle:
@kyuksel the submodules are still pegged to specific hashes, e.g.:

```
rtitle@leonardo[rt-jenkinsfile]$ git submodule status
 94277fe33c34f847b745582d9006965fcf951aa2 jenkins/dataproc-custom-images/dataproc-custom-images (heads/master)
 fb388c5efe715bfa89e0a9c51600d23b5cf488a6 jenkins/gce-custom-images/cis-harden-images (remotes/origin/ports-issue-3-gfb388c5)
 da0a85511508378fb17a14f604cb06e4704311d2 jenkins/gce-custom-images/terra-cryptomining-security-alerts (heads/master)
```

This PR removes the ability to override the submodule version via Jenkins input parameter.

So now, the lib versions are only defined by their submodule hash in Leo.

rtitle:
Successfully built an image with this -- merging.

PR #1563, opened 2020-08-14
Add Helm3 client and chart repos to Leo image

kyuksel:
[JIRA subtask](https://broadworkbench.atlassian.net/browse/IA-2136)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,7 +22,7 @@ RUN mkdir /leonardo
 COPY ./leonardo*.jar /leonardo
 COPY --from=helm-go-lib-builder /helm-go-lib-build/helm-scala-sdk/helm-go-lib /leonardo/helm-go-lib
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -22,5 +22,15 @@ RUN mkdir /leonardo
 COPY ./leonardo*.jar /leonardo
 COPY --from=helm-go-lib-builder /helm-go-lib-build/helm-scala-sdk/helm-go-lib /leonardo/helm-go-lib
 
+# Install the Helm3 CLI client from the source because installing it via the RHEL package managing didn't work
```
Qi77Qi: I think installing from script is a bit different from installing from source :)

Qi77Qi (approved):
👍

Dockerfile
kyuksel: picky picky 😛 
j/k you're right - corrected

rtitle:
👍

PR #1564, opened 2020-08-14
IA-2107- Adding Create App Automation tests

xuscrbw:
Adding automation tests for creating app without a cluster and then creating an app when a cluster already exists.

---

In all cases:

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetes/KubernetesSpec.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetes/KubernetesSpec.scala
@@ -40,55 +40,55 @@ class KubernetesSpec
       httpClient <- LeonardoApiClient.client
     } yield AppDependencies(httpClient)
 
-    "create app when cluster doesn't exist" in { googleProject =>
-      val appName = randomAppName
-      val appName2 = randomAppName
-
-      val createAppRequest = defaultCreateAppRequest.copy(
-        diskConfig = Some(
-          PersistentDiskRequest(
-            randomDiskName,
-            Some(DiskSize(30)),
-            None,
-            Map.empty
-          )
-        )
-      )
-      val res = dependencies.use { dep =>
-        implicit val client = dep.httpClient
-        val creatingDoneCheckable: DoneCheckable[GetAppResponse] =
-          x => x.status == AppStatus.Running
-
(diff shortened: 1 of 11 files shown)


automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetes/KubernetesSpec.scala
```diff
@@ -0,0 +1,243 @@
+package org.broadinstitute.dsde.workbench.leonardo
```
rtitle: Can we put this class in the `org.broadinstitute.dsde.workbench.leonardo.apps` package?

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetes/KubernetesSpec.scala
```diff
@@ -0,0 +1,243 @@
...
+          listApps = LeonardoApiClient.listApps(googleProject, true)
+
+          monitorApp1DeletionResult <- testTimer.sleep(30 seconds) >> streamFUntilDone(listApps, 120, 10 seconds)(
+            testTimer,
+            app1DeletedDoneCheckable
+          ).compile.lastOrError
+
+          _ <- loggerIO.info(s"app1 delete result: $monitorApp1DeletionResult")
+
+          _ <- testTimer.sleep(480 seconds)
```
rtitle: Do we need to create a second app here? It adds a lot of time to the test.. is creating 1 app enough to test `"create app when cluster doesn't exist"`?

rtitle:
Just thinking about the tests, there is some overlap I think.

Test A:
1. create app, cluster doesn't exist
2. delete app
3. create second app, cluster exists
4. delete second app

Test B:
1. create app, cluster doesn't exist
2. create second app, cluster exists
3. delete both apps

I'm wondering if we need both test cases, or if there is way to consolidate into 1 test. Will think more..

xuscrbw:
Would it make sense to remove the second test completely? 

I think for the first test, there might be a need to make sure that the cluster still remains in a good state after the app is deleted, but since we cover the "create cluster when app exists" in first test with the second app creation, I think the second test can be eliminated

rtitle (approved):
LGTM if tests pass

xuscrbw:
jenkins retest

PR #1565, opened 2020-08-16
[IA-2072] helm install nginx

rtitle:
* Added helm client dependency to Leonardo
* Integrated `helm install nginx` plus monitoring to `GKEInterpreter`
* Some general refactoring/cleanup around `GKEInterpreter` and `LeoPubsubMessageSubscriber`
* Fix some bugs based on Galaxy manual testing

Requires: 
https://github.com/broadinstitute/workbench-libs/pull/346/
https://github.com/broadinstitute/helm-scala-sdk/pull/4

Tested in a fiab. `createApp` sucessfully creates the cluster/nodepool/namespace/app, and nginx ingress appears to be up:
```
$ kubectl get svc -n nginx
NAME                                  TYPE           CLUSTER-IP     EXTERNAL-IP     PORT(S)                      AGE
nginx-nginx-ingress-controller        LoadBalancer   10.111.14.19   35.193.128.94   80:32070/TCP,443:32191/TCP   7m55s
nginx-nginx-ingress-default-backend   ClusterIP      10.111.1.58    <none>          80/TCP                       7m55s
```

---

In all cases:


http/src/main/resources/reference.conf
```diff
@@ -206,12 +206,10 @@ gke {
     namespaceNameSuffix = "namespace"
     services = [
       {
-        name = "galaxy-web"
+        name = "galaxy"
         kind = "ClusterIP"
       }
     ]
-    # Passed in galaxy config at helm install time
-    remoteUserName = "galaxy-user"
```
rtitle, line 214: I'm thinking we'll pass the real username as the Galaxy user since we have it anyway.

http/src/main/resources/reference.conf
```diff
@@ -414,7 +412,7 @@ helmConfig {
         name = "ca-secret"
         secretFiles = [
           {
-            name = "ca-crt"
+            name = "ca.crt"
```
rtitle: For whatever reason, the nginx ingress controller wants dots in these names, and I can't figure out how to change it.

https://kubernetes.github.io/ingress-nginx/examples/auth/client-certs/

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEAlgebra.scala
```diff
@@ -0,0 +1,34 @@
+package org.broadinstitute.dsde.workbench.leonardo.util
+
+import cats.mtl.ApplicativeAsk
+import org.broadinstitute.dsde.workbench.leonardo.{AppContext, AppId, AppName, KubernetesClusterLeoId, NodepoolLeoId}
+import org.broadinstitute.dsde.workbench.model.google.GoogleProject
+
+trait GKEAlgebra[F[_]] {
```
rtitle, line 12: We previously had a `GKEInterpreter` but not `GKEAlgebra`, so I added this. Also changed the methods a little bit.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -608,29 +608,20 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift](
...
-      clusterResult <- initialGoogleCall.adaptError {
-        case e =>
-          PubsubKubernetesError(
-            AppError(e.getMessage(), ctx.now, ErrorAction.CreateGalaxyApp, ErrorSource.Cluster, None),
-            Some(msg.appId),
-            false,
-            msg.cluster.map(_.defaultNodepoolId),
-            msg.cluster.map(_.clusterId)
+      // kick off cluster creation if necessary
+      createAndPollCluster = msg.cluster
```
Qi77Qi: I think the original code intend to keep some google calls being kicked off synchronous to messaging acking instead of queueing them, is that still true after the refactoring?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/package.scala
```diff
@@ -44,6 +45,13 @@ package object http {
       .compile
       .to(List)
 
+  def writeTempFile[F[_]: Sync: ContextShift](prefix: String, data: Array[Byte], blocker: Blocker): F[Path] =
```
Qi77Qi, line 48: what is this for?

project/Merging.scala
```diff
@@ -13,6 +13,12 @@ object Merging {
     case "module-info.class" =>
       MergeStrategy.discard // JDK 8 does not use the file module-info.class so it is safe to discard the file.
     case "reference.conf" => MergeStrategy.concat
-    case x                => oldStrategy(x)
+    // For the following error:
+    //[error] (assembly) deduplicate: different file contents found in the following:
+    //[error] logback.xml
+    //[error] /Users/rtitle/Library/Caches/Coursier/v1/https/broadinstitute.jfrog.io/broadinstitute/libs-release/org/broadinstitute/dsp/helm-scala-sdk_2.12/0.0.2-SNAPSHOT/helm-scala-sdk_2.12-0.0.2-SNAPSHOT.jar:logback.xml
+    // The Leo logback.xml should come first in the classpath, so okay to exclude the helm-scala-sdk one.
```
Qi77Qi: didn't realize I added logback in helm-sdk, I think we should delete it from the lib

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
rtitle: Yeah I thought about that -- right now everything is asynchronous. I wasn't sure the reasoning behind keeping some things synchronous to pubsub message processing...?

Especially in the createApp flow there is a lot of setup: create cluster -> monitor -> create nginx -> monitor -> create node pool -> monitor -> create galaxy -> monitor. Not sure where to draw the sync/async boundary.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/package.scala
rtitle: Yeah.. as far as I can tell the Go client only accepts the cert as a file path, not as a string/stream. So I'm writing a unique temp file in Leo immediately before invoking helm. 

Not ideal but I think it should work..

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
Qi77Qi: reason is if some initial google calls fail, then we can nack the message and pubsub will resend the message. Yea, when I discussed with Justin originally, we were thinking this retry would be best effort since that're lots of steps involved

project/Merging.scala
rtitle: Sounds good 👍 , I can do that in my PR

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
rtitle: I guess maybe we could do the very first create cluster call sync… but beyond that I don't think we want to resend the message since things could be in a bad state.

Also in some cases the cluster will already exist, in which case everything will still be async.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/package.scala
Qi77Qi: ah, bummer..okay..

Dockerfile
```diff
@@ -30,7 +33,8 @@ RUN curl -fsSL -o get_helm.sh https://raw.githubusercontent.com/helm/helm/master
 
 # Add the repos containing nginx and galaxy charts
 RUN helm repo add stable https://kubernetes-charts.storage.googleapis.com/ && \
-    helm repo add galaxy https://raw.githubusercontent.com/cloudve/helm-charts/anvil/
+    helm repo add galaxy https://raw.githubusercontent.com/cloudve/helm-charts/anvil/ && \
+    helm repo update
```
kyuksel, line 36: I wonder if we want to `helm dependency update` here as well.

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesClusterComponentSpec.scala
```diff
@@ -92,7 +92,7 @@ class KubernetesClusterComponentSpec extends AnyFlatSpecLike with TestComponent
   it should "update async fields" in isolatedDbTest {
     val savedCluster1 = makeKubeCluster(1).save()
 
-    val newAsyncFields = KubernetesClusterAsyncFields(externalIp, networkFields)
+    val newAsyncFields = KubernetesClusterAsyncFields(loadBalancerIp, apiServerIp, networkFields)
```
kyuksel, line 95: At some point we were planning on using `ClusterIP` as `Service` type. Why did we switch to `LoadBalancer` again?

http/src/main/resources/reference.conf
```diff
@@ -201,17 +201,49 @@ gke {
        autoscalingMax = 1
     }
   }
+  ingress {
+    namespace = "nginx"
+    release = "nginx"
+    chart = "stable/nginx-ingress"
+    loadBalancerService = "nginx-nginx-ingress-controller"
```
kyuksel, line 208: Should this be `nginx-ingress-controller`?

kyuksel (commented):
Started reviewing but didn't get far. To be continued... 😊

Dockerfile
rtitle: Hm, isn't `helm dependency update` just for local charts? At least when I try to run it it tries to look for a `Chart.yaml` file in the current directory. In Leo's case we'll just be using packaged charts in remote repositories.

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesClusterComponentSpec.scala
rtitle: The nginx-ingress-controller service type is `LoadBalancer` (it needs an external IP) -- but _all_ other services in the cluster are `ClusterIP`.

The `KUBERNETES_CLUSTER` table has 2 columns (which is what this code is checking):
- `loadBalancerIp` -- this is the external IP of the nginx-ingress-controller service, needed by the Leo proxy
- `apiServerIp` -- this is the IP of the k8s api server, needed by the helm client

Last comment: currently we have `serviceKind` in the Leo conf and database -- I'm not sure this is really needed since it'll always be `ClusterIP` for user-owned apps.

http/src/main/resources/reference.conf
rtitle: This is the literal name of the service, I guess in config I could probably specify it as `${release}-nginx-ingress-controller`

Dockerfile
kyuksel: Ah right..that makes sense.

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesClusterComponentSpec.scala
kyuksel: That makes sense. Thanks for the clarification.

> currently we have serviceKind in the Leo conf and database -- I'm not sure this is really needed since it'll always be ClusterIP for user-owned apps.

Yea, I can't think of a use case at the moment either.

rtitle:
jenkins retest

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEAlgebra.scala
```diff
@@ -0,0 +1,68 @@
...
+  /** Creates a GKE cluster but doesn't wait for its completion. */
+  def createCluster(params: CreateClusterParams)(implicit ev: ApplicativeAsk[F, AppContext]): F[CreateClusterResult]
+
+  /**
+   * Polls a creating GKE cluster for its completion and also does other cluster-wide set-up like
+   * install nginx ingress controller.
+   */
+  def pollCluster(params: PollClusterParams)(implicit ev: ApplicativeAsk[F, AppContext]): F[Unit]
+
+  /** Creates a GKE nodepool but doesn't want for its completion. */
```
kyuksel:
```suggestion
  /** Creates a GKE nodepool but doesn't wait for its completion. */
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -618,29 +616,29 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift](
...
-          )
-      }
-
-      createClusterAndNodepool = for {
-        _ <- clusterResult.traverse(result => gkeInterp.pollCluster(result)).adaptError {
+      // The "create app" flow does a number of things:
+      //  1. (optional) create, poll, and setup cluster
+      //  2. (optional) create and poll nodepool
+      //  3. (optional) create and poll disk
+      //  4. create and poll app
```
kyuksel, line 630: Thanks for thinking of including these helpful comments. I wonder if it'd be clearer to remove the word `(optional)` from the first three and append `, if it doesn't exist` to them since the steps aren't `optional` as much as `conditional`.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -618,29 +616,29 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift](
...
+
+      // Create the cluster synchronously
+      createClusterResultOpt <- msg.cluster
+        .traverse { c =>
+          gkeInterp.createCluster(CreateClusterParams(c.clusterId, msg.project, List(c.defaultNodepoolId), false))
+        }
+        .adaptError {
           case e =>
             PubsubKubernetesError(
               AppError(e.getMessage(), ctx.now, ErrorAction.CreateGalaxyApp, ErrorSource.Cluster, None),
```
kyuksel: > `e.getMessage()`

Not your change but I believe the [Scala convention](https://docs.scala-lang.org/style/method-invocation.html) is, methods with no side effects are called with no parentheses.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -618,29 +616,29 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift](
...
+      // The "create app" flow does a number of things:
+      //  1. (optional) create, poll, and setup cluster
+      //  2. (optional) create and poll nodepool
+      //  3. (optional) create and poll disk
+      //  4. create and poll app
+      //
+      // Numbers 1-3 are all Google calls; (4) is a helm call. If either of (1) or (2) are
+      // necessary then we will do the _initial_ Google call synchronous to the pubsub processing so
+      // we can nack the message on errors. Monitoring all creations will be asynchronous,
+      // and (3) and (4) will always be asynchronous.
```
kyuksel, line 635: Trying to understand why we're not doing the initial Google call synchronous to the pubsub processing, for (3) also, so we can nack the message on errors for that as well.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -696,13 +737,30 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift](
...
-  def handleBatchNodepoolCreateMessage(msg: BatchNodepoolCreateMessage)(
+  private[monitor] def handleBatchNodepoolCreateMessage(msg: BatchNodepoolCreateMessage)(
     implicit ev: ApplicativeAsk[F, AppContext]
   ): F[Unit] =
     for {
       ctx <- ev.ask
-      initialGoogleCall = gkeInterp.createCluster(msg.clusterId, msg.nodepools, true)
-      clusterResult <- initialGoogleCall.adaptError {
+
+      // Unlike handleCreateAppMessage, here we assume that the cluster does not exist so we always create it.
```
kyuksel, line 752: Does it make sense to check if the cluster exists before going through with the action?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -187,138 +124,330 @@ class GKEInterpreter[F[_]: Parallel: ContextShift](
...
+      _ <- logger.info(
+        s"Creating ${config.ingressConfig.namespace.value} namespace in cluster ${params.clusterId} | trace id: ${ctx.traceId}"
+      )
+      _ <- kubeService.createNamespace(dbCluster.getGkeClusterId, KubernetesNamespace(config.ingressConfig.namespace))
+
+      _ <- logger.info(
+        s"Installing ingress helm chart: ${config.ingressConfig.chart} in cluster ${params.clusterId} | trace id: ${ctx.traceId}"
+      )
+
+      // The helm client requires a Google acess token
```
kyuksel:
```suggestion
      // The helm client requires a Google access token
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -187,138 +124,330 @@ class GKEInterpreter[F[_]: Parallel: ContextShift](
...
+      // Resolve the cluster in Google
+      googleClusterOpt <- gkeService.getCluster(dbCluster.getGkeClusterId)
+      googleCluster <- F.fromOption(
+        googleClusterOpt,
+        ClusterCreationException(
+          s"Cluster not found in Google: ${dbCluster.getGkeClusterId} | trace id: ${ctx.traceId}"
+        )
+      )
+
+      // Install nginx igress controller
```
kyuksel: Would it be better to move nginx ingress controller installation and monitoring to own method(s)?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -187,138 +124,330 @@ class GKEInterpreter[F[_]: Parallel: ContextShift](
...
+      // Monitor nginx until public IP is accessible
+      loadBalancerIpOpt <- (
+        Stream.sleep_(config.monitorConfig.createIngress.interval) ++
+          Stream.eval(
+            kubeService.getServiceExternalIp(dbCluster.getGkeClusterId,
+                                             KubernetesNamespace(config.ingressConfig.namespace),
+                                             config.ingressConfig.loadBalancerService)
           )
+      ).repeatN(config.monitorConfig.createIngress.maxAttempts)
+        .takeThrough(!_.isDefined)
```
kyuksel: Nit
```suggestion
        .takeThrough(_.isEmpty)
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEAlgebra.scala
rtitle: heh thanks

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
rtitle: Sure, makes sense

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
rtitle: Yeah, the original code (`createDiskForApp`) had disk creation + monitoring async, so I decided not to change it. We COULD refactor to make the initial call sync, but I feel like it's not as important -- the GKE calls are the ones we want to nack and retry on errors.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
rtitle: I think front Leo checks this, it throws 409 on `batchNodepoolCreate` if a cluster already exists for the project.

However it might be a good idea to add a check here for safety as well..

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
rtitle: Yeah maybe… will try and see how it looks

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
rtitle: Actually I don't think we need an additional check -- the `GKEInterpreter` call will fail appropriately if the cluster already exists.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
kyuksel: Also 
```suggestion
      // Install nginx ingress controller
```

rtitle:
@kyuksel applied feedback so far :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -187,138 +124,330 @@ class GKEInterpreter[F[_]: Parallel: ContextShift](
...
+        s"Creating ${config.ingressConfig.namespace.value} namespace in cluster ${params.clusterId} | trace id: ${ctx.traceId}"
+      )
+      _ <- kubeService.createNamespace(dbCluster.getGkeClusterId, KubernetesNamespace(config.ingressConfig.namespace))
+
+      _ <- logger.info(
+        s"Installing ingress helm chart: ${config.ingressConfig.chart} in cluster ${params.clusterId} | trace id: ${ctx.traceId}"
+      )
+
+      // The helm client requires a Google acess token
+      _ <- F.delay(credentials.refreshIfExpired())
```
kyuksel: How are we deciding where and when to refresh the token?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -187,138 +124,330 @@ class GKEInterpreter[F[_]: Parallel: ContextShift](
...
+          .map(_.toMap)
+      } yield KubernetesSecret(
+        namespace,
+        secret.name,
+        secretFiles,
+        KubernetesSecretType.Generic
+      )
+    }
+
+  private[util] def getGoogleNodepool(nodepool: Nodepool): com.google.container.v1.NodePool = {
```
kyuksel: FWIW, the method name threw me off wondering how a Google nodepool would be different and also thought it would get nodepools from somewhere. At the same time I couldn't find a good name reflecting what the method achieves :(

kyuksel (commented):
Finished going through. Looks great overall. A few more comments inline.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
rtitle: I just call `refreshIfExpired` every time before invoking helm

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
rtitle: This is really building the google nodepool to pass in the create cluster request… hm maybe `buildGoogleNodepool`?

rtitle:
Thanks @kyuksel! Could I get a 👍 if you're good with it?

(I'd like to get this merged -- we'll need to iterate anyway to add the Galaxy helm install)

rtitle:
jenkins retest

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
kyuksel: If that's needed, I wonder if we could bake it into `HelmInterpreter` or such somehow so the caller doesn't need to know/remember to do that every time and/or add a comment to the appropriate places. Not a blocker but potentially something that could prevent issues in the future.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
kyuksel: Yea, I like `buildGoogleNodepool` better 😊

kyuksel (approved):
👍

rtitle:
weird issues downloading sbt dependencies.. jenkins retest

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -618,71 +616,114 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift](
...
-          PubsubKubernetesError(
-            AppError(e.getMessage(), ctx.now, ErrorAction.CreateGalaxyApp, ErrorSource.Disk, None),
-            Some(msg.appId),
-            false,
-            None,
-            None
-          )
-      }
-      parPreAppCreationSetup = List(createClusterAndNodepool, diskCreation).parSequence.flatMap(_ => F.unit)
+        // monitor cluster creation if necessary
```
jdcanas: previously we did the cluster monitoring and nodepool handling in parallel to disk creation and monitoring via `parPreAppCreationSetup = List(createClusterAndNodepool, diskCreation).parSequence.flatMap(_ => F.unit)`. Not sure how long disk creation/monitoring takes, but it seems good to not hold up other processing on that

rtitle: Good point, previously I thought monitoring this disk creation was async anyway so it didn't need to be parallel.. but now I see that this disk creation+monitoring is synchronous. So maybe it makes sense to do it in parallel to create cluster+nodepool. Will update.

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesClusterComponentSpec.scala
jdcanas: Yeah. The use case wasn't totally hammered down when we made the table. I figured it was worth specifying the type in case we need to support multiple and leo needs to decide how to route based on the AppType/Service type down the line. I'm not opposed to removing until it becomes relevant.

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
```diff
@@ -311,7 +311,8 @@ final case class App(id: AppId,
                      customEnvironmentVariables: Map[String, String]) {
   def getProxyUrls(project: GoogleProject, proxyUrlBase: String): Map[ServiceName, URL] =
     appResources.services.map { service =>
-      (service.config.name, new URL(s"$proxyUrlBase/$project/$appName/${service.config.name}"))
+      (service.config.name,
+       new URL(s"${proxyUrlBase}google/v1/apps/${project.value}/${appName.value}/${service.config.name.value}"))
```
jdcanas, line 315: Ah I just fixed this too, thanks :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
rtitle: Just added back parallel disk creation/monitoring

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
rtitle: Yeah currently the interface to `HelmInterpreter` is just a token, but one could imagine it taking a SA json in its constructor, similar to the wb-libs google2 interpreters. Maybe some improvements for the future.

PR #1567, opened 2020-08-18
QA-1205 - Fix for Click intercept modal

xuscrbw:
Adding in a sleep, as that seems to trigger the modal more often. Once that is triggered we will work through the modal. 

---

In all cases:

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookPage.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookPage.scala
@@ -372,7 +372,6 @@ class NotebookPage(val url: String)(implicit override val authToken: AuthToken,
   def clickOverrideNotebookChanged(): Unit =
     if (find(confirmNotebookSaveButton).size > 0) {
       click on confirmNotebookSaveButton
-      await notVisible jupyterModal
     }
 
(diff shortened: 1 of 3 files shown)


automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookPage.scala
```diff
@@ -372,7 +372,6 @@ class NotebookPage(val url: String)(implicit override val authToken: AuthToken,
   def clickOverrideNotebookChanged(): Unit =
     if (find(confirmNotebookSaveButton).size > 0) {
       click on confirmNotebookSaveButton
-      await notVisible jupyterModal
```
Qi77Qi, line 375: what does this do?

also can you do `git pull --rebase origin develop` and force push

xuscrbw: In the flow right now, there are two jupyterModals. When I was testing I noticed that even after we clicked on the confirmNotebookSaveButton, we would not get to the next step since it was awaiting for the modal underneath the first to not be visible as well. Removing this let's us get to the next step. I am a bit worried about needing a wait between the two clicks, but locally it seemed to work okay.

Qi77Qi (approved):

PR #1569, opened 2020-08-18
[IA-2146] Add GSA and KSA to App table and remove GSA from cluster table

gpcarr:
Currently we store the Leo dev SA as the service account for the kubernetes cluster. When it comes to workload identity, this won't work. We need the google service account of the user who created the app in order to link it to a kubernetes service account. we get the google service account from the create app request in front leo then back leo will create the kubernetes service account and update the db. 

This PR also removes `serviceAccount` from the kubernetes cluster table entirely because there are multiple users per cluster. Also, we were saving the leo dev SA which is not helpful in workload identity and not needed since leo will always be creating the cluster and we know what the leo dev SA is. 

---

In all cases:


Qi77Qi (approved):
lgtm

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20200817_add_ksa_and_gsa_to_app_table.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="leonardo" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+    <changeSet logicalFilePath="leonardo" author="gabriela" id="add_ksa_and_gsa_to_app_table">
+        <addColumn tableName="APP">
+            <column name="googleServiceAccount" type="VARCHAR(254)">
+                <constraints nullable="false"/>
```
rtitle, line 6: Just checked in dev and there are a couple of existing `APP` records. Not sure if adding a new not-null column will break w/ existing records.. maybe we should add some dummy default value just in case.

rtitle (approved):
Looks good

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20200817_add_ksa_and_gsa_to_app_table.xml
gpcarr: ah good point, i'll update then merge

kyuksel (approved):
Looks good 👍

PR #1570, opened 2020-08-18
IA-2138 disallow updating dataproc cluster when runtime is stopped

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-2138

---

In all cases:


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeServiceInterp.scala
```diff
@@ -665,6 +665,16 @@ class RuntimeServiceInterp[F[_]: Parallel](config: RuntimeServiceConfig,
       context <- ctx.ask
       // should num workers be updated?
       targetNumWorkers <- traverseIfChanged(req.updatedNumberOfWorkers, dataprocConfig.numberOfWorkers)(Async[F].pure)
+      _ <- targetNumWorkers.traverse(_ =>
+        if (runtime.status != RuntimeStatus.Running)
+          F.raiseError[Unit](
+            new LeoException(
+              s"${context.traceId.asString} | Bad request. Number of workers can only be updated if the dataproc cluster is Running. Please start your runtime before updating the cluster",
```
kyuksel: The last sentence in the error message sounds like it assumes the status is `Stopped` even though the predicate is any status but `Running`. I assume the code is correct, in which case we may want to add to the message what the status is and remove `Please start your runtime before updating the cluster`.

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/TestUtils.scala
```diff
@@ -21,6 +22,29 @@ object TestUtils extends Matchers {
     }
   }
 
+  implicit val leonardoExceptionEq = {
+    new Equality[LeoException] {
+      def areEqual(a: LeoException, b: Any): Boolean =
+        b match {
+          case bb: LeoException =>
+            println("comparinggggg")
```
kyuksel:
```suggestion
```

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/TestUtils.scala
```diff
@@ -21,6 +22,29 @@ object TestUtils extends Matchers {
...
+            a.message == bb.message && a.statusCode == bb.statusCode && a.cause == bb.cause
```
kyuksel, line 30: This looks right but I just wonder if there's a built-in way to check `Exception` equality that we could leverage instead of rolling our own in case there are edge cases we may not be aware of.

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeServiceInterpSpec.scala
```diff
@@ -766,6 +769,25 @@ class RuntimeServiceInterpSpec extends AnyFlatSpec with LeonardoTestSuite with T
     res.unsafeRunSync()
   }
 
+  it should "disallow updating dataproc cluster number of workers if runtime is Stopped" in {
```
kyuksel: How are we setting the runtime status to `Stopped` below?

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/TestUtils.scala
Qi77Qi: I don't think there's a "built-in" way to check `Exception`...

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeServiceInterp.scala
Qi77Qi: I'll update the check it `Stopped`

kyuksel: You mean changing it to 
```scala
if (runtime.status != RuntimeStatus.Stopped)
```
?

If so, what will happen when `runtime.status` is not `Running` or `Stopped`? My understanding was that we can update a cluster only if its status is `Running`, no?

Qi77Qi: > My understanding was that we can update a cluster only if its status is Running, no?

Yes. I changed the criteria to address just `Stopped` because the message is only applicable to `Stopped`, I didn't want to do a pattern match on the status and then depending on the status, giving a different message, cuz it seems too much for little gain, but I guess I can update it to `Stopped` vs other invalid statuses

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeServiceInterpSpec.scala
kyuksel: Should we test the other statuses where we should get a `Bad Request` as well?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeServiceInterp.scala
```diff
@@ -665,6 +665,16 @@ class RuntimeServiceInterp[F[_]: Parallel](config: RuntimeServiceConfig,
...
+      _ <- targetNumWorkers.traverse(_ =>
+        if (runtime.status == RuntimeStatus.Stopped)
+          F.raiseError[Unit](
+            new LeoException(
+              s"${context.traceId.asString} | Bad request. Number of workers can only be updated if the dataproc cluster is Running. Please start your runtime before updating the cluster.",
+              StatusCodes.BadRequest
+            )
+          )
+        else F.unit
+      )
```
kyuksel: I thought the original predicate (`runtime.status != RuntimeStatus.Running`) was correct. If we `raiseError` only when `runtime.status == RuntimeStatus.Stopped` then we wouldn't be handling the other non-Running statuses correctly, no? 

I understand that you're trying to make it clearer so if a user's cluster was stopped, they know to start it before attempting to update the cluster but I think the message "Number of workers can only be updated if the dataproc cluster is Running" is sufficient to convey that message.

kyuksel: In other words, I suggest having

```scala
        if (runtime.status != RuntimeStatus.Running)
          F.raiseError[Unit](
            new LeoException(
              s"${context.traceId.asString} | Bad request. The Dataproc cluster has status ${runtime.status} " +
              s"whereas number of workers can only be updated if the cluster has status ${RuntimeStatus.Running}.",
              StatusCodes.BadRequest
            )
          )
        else F.unit
```

Qi77Qi: I think the message is helpful, but I'm not gonna fight with you on this if you feel strongly :) will update

kyuksel (approved):
👍

PR #1575, opened 2020-08-24
[IA-2067] set up workload identity in back leo

gpcarr:
This PR does 2 things:
- [IA-2067] which sets up workload identity in back leo
- [IA-2189] which adds the leo side code to using the older google client lib to create clusters (wb-libs code can be found in this [PR](https://github.com/broadinstitute/workbench-libs/pull/350))

Testing:
- I used the `createApp` endpoint and verified that the cluster was successfully created in Google
- I created the first app as Hermione and then tested workload identity by doing the following:
creating a Pod with the KSA that runs the cloud-sdk container image, and connecting to it with an interactive session and then running `gcloud auth list` from within the pod and verified it was Hermione's Google SA that was returned:
Screen Shot 2020-09-17 at 1 03 40 PM
I then also ran command `kubectl get serviceaccounts/test1-galaxy-ksa -o yaml` to verify that Hermione's KSA is correctly annotated:
Screen Shot 2020-09-17 at 1 04 33 PM
I then did all of these steps again except authed as Ron and found that that works as expected as well

---

In all cases:

[IA-2067]: https://broadworkbench.atlassian.net/browse/IA-2067
[IA-2189]: https://broadworkbench.atlassian.net/browse/IA-2189


core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
```diff
@@ -158,7 +158,7 @@ object NodepoolStatus {
 final case class KubernetesClusterLeoId(id: Long) extends AnyVal
 final case class NamespaceId(id: Long) extends AnyVal
 final case class Namespace(id: NamespaceId, name: NamespaceName)
-final case class KubernetesServiceAccount(value: String) extends AnyVal
+final case class KubernetesServiceAccountName(value: String) extends AnyVal
```
gpcarr: changing this because we already have object `KubernetesServiceAccount` (which has both the SA account and its annotations) in wb-libs `kubernetesModels`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -319,6 +411,12 @@ class GKEInterpreter[F[_]: Parallel: ContextShift](
                                        KubernetesNamespace(getAppResult.app.appResources.namespace.name))
       _ <- appQuery.markAsDeleted(getAppResult.app.id, now).transaction
     } yield ()
+
+  private def retry[A](ioa: F[A], retryable: Throwable => Boolean): F[A] =
```
gpcarr: This method also exists in `DataprocInterpreter` (https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreter.scala#L617) was wondering where is a good place to pull this method out to so it can be used both here and in `DataprocInterpreter`

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
rtitle: Wonder if we should just use [ServiceAccountName](https://github.com/broadinstitute/workbench-libs/blob/develop/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/kubernetesModels.scala#L129) in Leo..

gpcarr: I think that's better, less to add to `kubernetesModels`

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
```diff
@@ -294,7 +294,7 @@ final case class AppName(value: String) extends AnyVal
 final case class AppResources(namespace: Namespace,
                               disk: Option[PersistentDisk],
                               services: List[KubernetesService],
-                              kubernetesServiceAccount: Option[KubernetesServiceAccount])
+                              kubernetesServiceAccountName: Option[KubernetesServiceAccountName])
```
rtitle: I'm wondering if we should make `kubernetesServiceAccountName` non-Optional in the DB and have front Leo always populate it. Back Leo would read it and create it in GKE. This is similar to how we handle namespaces.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -53,14 +78,81 @@ class GKEInterpreter[F[_]: Parallel: ContextShift](
       getApp <- F.fromOption(getAppOpt, AppNotFoundException(msg.project, msg.appName, ctx.traceId))
       gkeClusterId = getApp.cluster.getGkeClusterId
       namespaceName = getApp.app.appResources.namespace.name
+      googleProject = getApp.cluster.googleProject
```
rtitle: Might need to handle conflicts once https://github.com/DataBiosphere/leonardo/pull/1565 is merged (and tests are fixed).. I changegd this method to take `params: CreateAppParams` which includes `googleProject`.

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
gpcarr: hmmm the reason I didn't go with that approach was in the case that the call to `kubernetesService.createServiceAccount` fails in back leo

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -53,14 +78,81 @@ class GKEInterpreter[F[_]: Parallel: ContextShift](
...
+      gsa = getApp.app.googleServiceAccount
+      ksaName = KubernetesServiceAccountName(s"${namespaceName.value}-ksa")
```
rtitle: I think the namespace name is `s"${appName}-namespace"` so this would end up being `s"${appName}-namespace-ksa"`. Maybe just `s"${appName}-ksa"` would be cleaner?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -53,14 +78,81 @@ class GKEInterpreter[F[_]: Parallel: ContextShift](
...
+
+      // Create KSA
+      annotations = Map("iam.gke.io/gcp-service-account" -> gsa.value)
+      _ <- kubeService
+        .createServiceAccount(
+          gkeClusterId,
+          KubernetesServiceAccount(ServiceAccountName(ksaName.value), annotations),
+          KubernetesNamespace(namespaceName)
+        )
+        .handleErrorWith {
```
rtitle: Probably don't want `handleErrorWith` since that'll recover from the error. Maybe best to just let the error fall through -- it'll be caught and logged by `LeoPubsubMessageSubscriber`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -53,14 +78,81 @@ class GKEInterpreter[F[_]: Parallel: ContextShift](
...
+          case e =>
+            logger.error(e)(
+              s"Failed to create Kubernetes SA for namespace ${namespaceName.value} in cluster ${gkeClusterId.toString}"
+            )
+        }
+      // Persist KSA to DB
+      _ <- appQuery.updateKubernetesServiceAccount(getApp.app.id, ksaName).transaction
```
rtitle: If front Leo populates the KSA up front then this method will not be needed

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
rtitle: Yeah I wonder if we could call the wb-libs `retry` directly, which I did in `VPCInterpreter`: https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/VPCInterpreter.scala#L150

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
rtitle: Yeah that's true… ok I think it's fine to only populate it in the DB after it's created in k8s

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
gpcarr: i imagine i will have lots of merge conflicts hehe but i'll make sure you merge first since your PR is much bigger and dealing with merge conflicts would be more annoying!

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
gpcarr: hmmm yeah I can update. Is app name unique in a cluster?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
gpcarr: sounds good, will update!

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
gpcarr: I can try that out!

rtitle:
This looks good in general -- as a next step it would be good to update this PR to depend on https://github.com/broadinstitute/workbench-libs/pull/350, add `.setWorkloadIdentity` to the cluster object, and actually create the cluster using `legacyCreateClusterRec` to make sure it all works.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -100,44 +107,40 @@ class GKEInterpreter[F[_]: Parallel: ContextShift: Timer](
...
         )
-        .setIpAllocationPolicy( //otherwise it uses the legacy one, which is insecure. See https://cloud.google.com/kubernetes-engine/docs/how-to/alias-ips
-          IPAllocationPolicy
-            .newBuilder()
+        .setIpAllocationPolicy(
+          new com.google.api.services.container.model.IPAllocationPolicy()
             .setUseIpAliases(true)
         )
-        .build()
+        .setWorkloadIdentityConfig(
```
rtitle, line 133: woohoo :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -100,44 +107,40 @@ class GKEInterpreter[F[_]: Parallel: ContextShift: Timer](
       kubeNetwork = KubernetesNetwork(dbCluster.googleProject, network)
       kubeSubNetwork = KubernetesSubNetwork(dbCluster.googleProject, dbCluster.region, subnetwork)
 
-      createClusterReq = Cluster
-        .newBuilder()
+      legacyCreateClusterRec = new com.google.api.services.container.model.Cluster()
```
rtitle, line 110: This could be a separate ticket/PR but I bet we could also set GKE Sandbox now

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -100,44 +107,40 @@ class GKEInterpreter[F[_]: Parallel: ContextShift: Timer](
...
         )
-        .build()
+        .setWorkloadIdentityConfig(
+          new com.google.api.services.container.model.WorkloadIdentityConfig()
+            .setWorkloadPool(s"${params.googleProject.value}.svc.id.goog")
+        )
 
       // Submit request to GKE
-      req = KubernetesCreateClusterRequest(dbCluster.googleProject, dbCluster.location, createClusterReq)
+      // TODO: use legacy cluster once wb-libs is in place
```
rtitle, line 139: probably remove TODO, no?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -310,9 +314,11 @@ class GKEInterpreter[F[_]: Parallel: ContextShift: Timer](
       )
 
       // Create KSA
-      ksaName = KubernetesServiceAccount(s"${app.appName.value}-${config.galaxyAppConfig.serviceAccountSuffix}")
+      gsa = dbApp.app.googleServiceAccount
+      ksaName = ServiceAccountName(s"${app.appName.value}-${config.galaxyAppConfig.serviceAccountSuffix}")
```
rtitle: Sorry to cause conflicts but I changed the KSA name in the PR I just merged. It's now a static string from reference.conf, we don't need to embed the appName.

rtitle:
Overall looks great! Nice work. Just need to resolve conflicts, get tests passing, and then :shipit:

project/Dependencies.scala
```diff
@@ -16,12 +16,12 @@ object Dependencies {
...
   val opencensusV = "0.26.0"
-  val serviceTestV = "0.18-61887ce"
+  val serviceTestV = "0.18-ae64deb-SNAP"
 
   val workbenchUtilV = "0.5-4c7acd5"
   val workbenchModelV = "0.14-2e155f0"
-  val workbenchGoogleV = "0.21-2a218f3"
-  val workbenchGoogle2V = "0.12-2187bc0"
+  val workbenchGoogleV = "0.21-64a7b29"
+  val workbenchGoogle2V = "0.13-ae64deb-SNAP"
```
rtitle: You can update to non-SNAP once the wb-libs PR is merged and travis runs against dev

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
gpcarr: yep, will remove!

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
gpcarr: okay i'll work on rebasing now!

gpcarr:
jenkins retest ... leo build failed

rtitle (approved):

kyuksel (approved):
Looks great to me! 👍 

Becoming a big fan of your PR messages nicely detailing what testing you did (especially for tricky features like this) too 😊

Qi77Qi:
do we need any automation test for this?

gpcarr:
> do we need any automation test for this?

don't think so, we have unit tests for KSA creation + i tested everything in my FIAB

PR #1576, opened 2020-08-24
Switch to a standard enum ref for CreateDiskRequest.diskType

calbach:
allOf is apparently legal here, but Java codegen fails on it, and this is consistent with other usages in the Leo spec.

### DIFF
--- a/http/src/main/resources/swagger/api-docs.yaml
+++ b/http/src/main/resources/swagger/api-docs.yaml
@@ -2826,9 +2826,7 @@ components:
           type: integer
           description: The size of the persistent disk to be created in GB.
         diskType:
-          allOf:
-            - $ref: "#/components/schemas/DiskType"
-            - description: The type of the persistent disk to be created. Can be "Standard" or "SSD".
+          $ref: "#/components/schemas/DiskType"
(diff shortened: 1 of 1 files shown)


http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -2826,9 +2826,7 @@ components:
           type: integer
           description: The size of the persistent disk to be created in GB.
         diskType:
-          allOf:
```
Qi77Qi, line 2829: should this be `oneof`? 🤔

calbach: No - it's an enum so it should just reference the enum type. If there were a more complex disk configuration per type, then you may want a oneof on StandardDiskConfig and SolidStateDiskConfig, using a discriminator of (still) the diskType enum.

Qi77Qi (approved):

rtitle (approved):
LGTM

Note tests are failing due to some Jenkins infrastructure issues, I hope it's fixed soon

rtitle:
Rebased this to get the test fix

PR #1580, opened 2020-08-25
[IA-2154] Several runtimes in prod cannot be deleted in Error status

rtitle:
See https://broadworkbench.atlassian.net/browse/IA-2154

Make runtime monitor tool checking more configurable, and fixed a bug where deleting a runtime while tools are being monitored causes the runtime to go into a bad state.

Will test in a fiab.

---

In all cases:


http/src/main/resources/reference.conf
```diff
@@ -41,8 +41,9 @@ dataproc {
   monitor {
     initialDelay = 30 seconds
     pollingInterval = 15 seconds
-    checkToolsDelay = 8 seconds
```
rtitle, line 44: Previously `checkToolsDelay` was not used (the value was hardcoded in `BaseCloudServiceRuntimeMonitor`). I made it be used and also added `checkToolsMaxAttempts`.

Qi77Qi: the original intent of `checkToolsDelay` was to have an initial wait before we ping jupyter or welder, but I removed it later since I felt it wasn't benefiting much, and there're cases where the tools might be ready when we call checkTools. I think you're using it as `interval`, shall we rename it to `checkToolsInterval` instead?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BaseCloudServiceRuntimeMonitor.scala
```diff
@@ -320,20 +320,36 @@ abstract class BaseCloudServiceRuntimeMonitor[F[_]] {
...
+          // It's possible that the user explicitly deleted the runtime in which case
+          // we can ignore the error. Otherwise, move the runtime to Error status.
+          for {
+            curStatusOpt <- clusterQuery.getClusterStatus(runtimeAndRuntimeConfig.runtime.id).transaction
+            curStatus <- F.fromOption(
+              curStatusOpt,
+              new Exception(s"Cluster with id ${runtimeAndRuntimeConfig.runtime.id} not found in the database")
+            )
+            res <- curStatus match {
+              case RuntimeStatus.Deleted => F.pure[CheckResult](((), None))
```
Qi77Qi: shall we add the specific case in `failedRuntime` function instead of here? That way, no matter where we end up with `Deleted` status, `failedRuntime` won't try to put the cluster in `Error` status

http/src/main/resources/reference.conf
rtitle: Ah yeah I am using it as initial delay and interval.. sound good renaming it to `checkToolsInterval` (and maybe get rid of initial delay)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BaseCloudServiceRuntimeMonitor.scala
rtitle: Hmm.. yeah that probably makes sense

rtitle: I also notice that `GceRuntimeMonitor.failedRuntime` and `DataprocRuntimeMonitor.failedRuntime` are very similar.. the only difference is how errors are persisted to the DB. Considering moving some of this logic to `BaseCloudServiceRuntimeMonitor` to avoid duplication.

rtitle:
jenkins retest

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BaseCloudServiceRuntimeMonitor.scala
```diff
@@ -90,7 +93,68 @@ abstract class BaseCloudServiceRuntimeMonitor[F[_]] {
...
+              runtimeAndRuntimeConfig.runtime.runtimeName,
+              runtimeAndRuntimeConfig.runtime.asyncRuntimeFields.isDefined
+            )
+          )
+          .void, //TODO is this right when deleting or stopping fails?
+        //save cluster error in the DB
+        saveRuntimeError(
+          runtimeAndRuntimeConfig.runtime.id,
+          errorDetails,
+          None //TODO: pass in startup script path in the future
```
rtitle: This was `None` for both GCE and Dataproc, so I left this TODO here in the base class

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BaseCloudServiceRuntimeMonitor.scala
```diff
@@ -344,15 +410,52 @@ abstract class BaseCloudServiceRuntimeMonitor[F[_]] {
...
+                                 startUpScriptOutputFile: Option[GcsPath])(
+    implicit ev: ApplicativeAsk[F, AppContext]
+  ): F[Unit] = {
+    val result = for {
+      ctx <- ev.ask
+      errorToPersist <- errorDetails match {
+        case Some(error) =>
+          F.pure(error)
+        case None =>
+          startUpScriptOutputFile match {
```
rtitle: Moved this logic from `DataprocRuntimeMonitor`.. though it seems that `startUpScriptOutputFile` was always passed as `None` anyway

rtitle:
@Qi77Qi applied PR feedback when you have a chance.. still need to test in a fiab

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BaseCloudServiceRuntimeMonitor.scala
```diff
@@ -90,7 +93,68 @@ abstract class BaseCloudServiceRuntimeMonitor[F[_]] {
...
+      // Update the cluster status to Error only if the runtime is non-Deleted.
+      // If the user has explicitly deleted their runtime by this point then
+      // we don't want to move it back to Error status.
+      _ <- for {
+        curStatusOpt <- clusterQuery.getClusterStatus(runtimeAndRuntimeConfig.runtime.id).transaction
+        curStatus <- F.fromOption(
+          curStatusOpt,
+          new Exception(s"Cluster with id ${runtimeAndRuntimeConfig.runtime.id} not found in the database")
+        )
+        _ <- curStatus match {
```
rtitle, line 141: This is the main fix: to not change runtime status if it is `Deleted`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BaseCloudServiceRuntimeMonitor.scala
Qi77Qi: hmm, can't remember why I left the TODO anymore...I don't mind if we just remove it either...

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BaseCloudServiceRuntimeMonitor.scala
Qi77Qi: shall we log a message in this case?

Qi77Qi (approved):

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BaseCloudServiceRuntimeMonitor.scala
rtitle: Yeah.. it looks like this was mainly for reporting user script errors in the Dataproc case (the GCE case covers this). Maybe will see if I can fix this easily, and remove the TODO

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BaseCloudServiceRuntimeMonitor.scala
rtitle: 👍

rtitle:
jenkins retest

rtitle:
transient automation failures, jenkins retest

PR #1587, opened 2020-08-27
[IA-2088] [IA-2141] [IA-2089] [IA-2170] [IA-2140] Back Leo helm installs Galaxy

kyuksel:
This PR covers the following tickets:

* [IA-2088](https://broadworkbench.atlassian.net/browse/IA-2088) Back Leo installs Galaxy via Helm client lib
* [IA-2141](https://broadworkbench.atlassian.net/browse/IA-2141) Build string of override values
   * All known `--set` flags to the Galaxy install should now be present
* [IA-2089](https://broadworkbench.atlassian.net/browse/IA-2089) Update back Leo to uninstall charts
   * See https://github.com/broadinstitute/helm-scala-sdk/pull/5. 
   * The helm uninstall seems to work, but there is a known Galaxy issue with pods not terminating (Alex is looking into it)
* [IA-2170](https://broadworkbench.atlassian.net/browse/IA-2170) Add Role and Role Binding to helm install galaxy
   * This PR creates a KSA and passes it to `rbac` in the `galaxykubeman` chart
   * Note it does NOT set up Workload Identity yet (that will be a separate PR)
* [IA-2140](https://broadworkbench.atlassian.net/browse/IA-2140) Add code for monitoring waiting for app to come up
   * Added basic monitoring based on http polling. It works for now, but we can add k8s job polling in a future iteration.

Tested create/delete app flows in a fiab, and all seems to work.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,7 +9,7 @@ RUN mkdir /helm-go-lib-build && \
     cd /helm-go-lib-build && \
     git clone https://github.com/broadinstitute/helm-scala-sdk.git && \
     cd helm-scala-sdk && \
-    git checkout master && \
+    git checkout rt-uninstall && \
     cd helm-go-lib && \
     go build -o libhelm.so -buildmode=c-shared main.go
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,7 +9,7 @@ RUN mkdir /helm-go-lib-build && \
     cd /helm-go-lib-build && \
     git clone https://github.com/broadinstitute/helm-scala-sdk.git && \
     cd helm-scala-sdk && \
-    git checkout rt-uninstall && \
+    git checkout master && \
     cd helm-go-lib && \
     go build -o libhelm.so -buildmode=c-shared main.go
 
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -242,6 +242,10 @@ gke {
     chart = "galaxy/galaxykubeman"
     namespaceNameSuffix = "galaxy-ns"
     serviceAccountSuffix = "galaxy-ksa"
+    # Setting uninstallKeepHistory will cause the `helm uninstall` command to keep a record of
+    # the deleted release, which can make debugging and auditing easier. It should be safe for
+    # our use case because we generate unique release names anyway.
+    # See https://helm.sh/docs/intro/using_helm/#helm-uninstall-uninstalling-a-release
     uninstallKeepHistory = true
     services = [
       {
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -242,6 +242,7 @@ gke {
     chart = "galaxy/galaxykubeman"
     namespaceNameSuffix = "galaxy-ns"
     serviceAccountSuffix = "galaxy-ksa"
+    uninstallKeepHistory = true
     services = [
       {
         name = "galaxy"
@@ -487,6 +488,7 @@ pubsub {
     persistent-disk-monitor {
       create {
         max-attempts = 5
+
         interval = 3 seconds
       }
       delete {
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -469,6 +469,10 @@ pubsub {
       max-attempts = 100 # 3 seconds * 100 is 5 min
       interval = 3 seconds
     }
+    createApp {
+      interval = 10 seconds
+      max-attempts = 120 # 10 seconds * 120 = 20 min
+    }
   }
 
   subscriber {
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -238,7 +238,7 @@ gke {
     ]
   }
   galaxyApp {
-    releaseName = "galaxy-rls"
+    releaseNameSuffix = "galaxy-rls"
     chart = "galaxy/galaxykubeman"
     namespaceNameSuffix = "galaxy-ns"
     services = [
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -240,6 +240,12 @@ gke {
   galaxyApp {
     releaseName = "galaxy1"
     namespaceNameSuffix = "galaxy-ns"
+    services = [
+      {
+        name = "galaxy"
+        kind = "ClusterIP"
+      }
+    ]
     values = [
       "nfs.storageClass.name=nfs-galaxy1",
       "cvmfs.repositories.cvmfs-gxy-data-galaxy1=data.galaxyproject.org"
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -241,6 +241,7 @@ gke {
     releaseNameSuffix = "galaxy-rls"
     chart = "galaxy/galaxykubeman"
     namespaceNameSuffix = "galaxy-ns"
+    serviceAccountSuffix = "galaxy-ksa"
     services = [
       {
         name = "galaxy"
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -238,13 +238,23 @@ gke {
     ]
   }
   galaxyApp {
-    releaseName = "release1"
-    namespaceNameSuffix = "namespace"
-    services = [
-      {
-        name = "galaxy"
-        kind = "ClusterIP"
-      }
+    releaseName = "galaxy1"
+    namespaceNameSuffix = "galaxy-ns"
+    values = [
+      "nfs.storageClass.name=nfs-galaxy1",
+      "cvmfs.repositories.cvmfs-gxy-data-galaxy1=data.galaxyproject.org"
+      "cvmfs.repositories.cvmfs-gxy-main-galaxy1=main.galaxyproject.org"
+      "cvmfs.cache.alienCache.storageClass=nfs-galaxy1"
+      "galaxy.persistence.storageClass=nfs-galaxy1"
+      "galaxy.cvmfs.data.pvc.storageClassName=cvmfs-gxy-data-galaxy1"
+      "galaxy.cvmfs.main.pvc.storageClassName=cvmfs-gxy-main-galaxy1"
+      "nodeSelector.\"beta\\.kubernetes\\.io/nodeName=ron-pool"
+      "galaxy.ingress.path=/proxy/google/v1/apps/galaxy-dev-env/galaxy1/galaxy"
+      "galaxy.ingress.annotations.\"nginx\\.ingress\\.kubernetes\\.io/proxy-redirect-from=https://863027343.jupyter-dev.firecloud.org"
+      "galaxy.ingress.annotations.\"nginx\\.ingress\\.kubernetes\\.io/proxy-redirect-to=https://leonardo-anvil.dsde-dev.broadinstitute.org:30443"
+      "galaxy.ingress.hosts[0]=863027343.jupyter-dev.firecloud.org"
+      "galaxy.configs.\"galaxy\\.yml\".galaxy.single_user=ron.weasley@test.firecloud.org"
+      "galaxy.configs.\"galaxy\\.yml\".galaxy.admin_users=ron.weasley@test.firecloud.org"
     ]
   }
 }
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -474,8 +474,8 @@ pubsub {
       max-attempts = 120 # 10 seconds * 120 = 20 min
     }
     deleteApp {
-      interval = 30 seconds
-      max-attempts = 40 # 30 seconds * 40 = 20 min
+      interval = 10 seconds
+      max-attempts = 120 # 10 seconds * 120 = 20 min
     }
   }
 
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -473,6 +473,10 @@ pubsub {
       interval = 10 seconds
       max-attempts = 120 # 10 seconds * 120 = 20 min
     }
+    deleteApp {
+      interval = 30 seconds
+      max-attempts = 40 # 30 seconds * 40 = 20 min
+    }
   }
 
   subscriber {
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -238,29 +238,14 @@ gke {
     ]
   }
   galaxyApp {
-    releaseName = "galaxy1"
+    releaseName = "galaxy-rls"
+    chart = "galaxy/galaxykubeman"
     namespaceNameSuffix = "galaxy-ns"
     services = [
       {
         name = "galaxy"
         kind = "ClusterIP"
       }
-    ]
-    values = [
-      "nfs.storageClass.name=nfs-galaxy1",
-      "cvmfs.repositories.cvmfs-gxy-data-galaxy1=data.galaxyproject.org"
-      "cvmfs.repositories.cvmfs-gxy-main-galaxy1=main.galaxyproject.org"
-      "cvmfs.cache.alienCache.storageClass=nfs-galaxy1"
-      "galaxy.persistence.storageClass=nfs-galaxy1"
-      "galaxy.cvmfs.data.pvc.storageClassName=cvmfs-gxy-data-galaxy1"
-      "galaxy.cvmfs.main.pvc.storageClassName=cvmfs-gxy-main-galaxy1"
-      "nodeSelector.\"beta\\.kubernetes\\.io/nodeName=ron-pool"
-      "galaxy.ingress.path=/proxy/google/v1/apps/galaxy-dev-env/galaxy1/galaxy"
-      "galaxy.ingress.annotations.\"nginx\\.ingress\\.kubernetes\\.io/proxy-redirect-from=https://863027343.jupyter-dev.firecloud.org"
-      "galaxy.ingress.annotations.\"nginx\\.ingress\\.kubernetes\\.io/proxy-redirect-to=https://leonardo-anvil.dsde-dev.broadinstitute.org:30443"
-      "galaxy.ingress.hosts[0]=863027343.jupyter-dev.firecloud.org"
-      "galaxy.configs.\"galaxy\\.yml\".galaxy.single_user=ron.weasley@test.firecloud.org"
-      "galaxy.configs.\"galaxy\\.yml\".galaxy.admin_users=ron.weasley@test.firecloud.org"
     ]
   }
 }
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/GalaxyAppConfig.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/GalaxyAppConfig.scala
@@ -3,7 +3,7 @@ package org.broadinstitute.dsde.workbench.leonardo.config
 import org.broadinstitute.dsde.workbench.google2.KubernetesSerializableName.NamespaceName
 import org.broadinstitute.dsde.workbench.leonardo.{ChartName, ReleaseName, ServiceConfig}
 
-case class GalaxyAppConfig(releaseName: ReleaseName,
+case class GalaxyAppConfig(releaseNameSuffix: ReleaseName,
                            chart: ChartName,
                            namespaceNameSuffix: NamespaceName,
                            services: List[ServiceConfig])
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/GalaxyAppConfig.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/GalaxyAppConfig.scala
@@ -1,9 +1,10 @@
 package org.broadinstitute.dsde.workbench.leonardo.config
 
 import org.broadinstitute.dsde.workbench.google2.KubernetesSerializableName.NamespaceName
-import org.broadinstitute.dsde.workbench.leonardo.{ChartName, ReleaseName, ServiceConfig}
+import org.broadinstitute.dsde.workbench.leonardo.{ChartName, KubernetesServiceAccount, ReleaseName, ServiceConfig}
 
 case class GalaxyAppConfig(releaseNameSuffix: ReleaseName,
                            chart: ChartName,
                            namespaceNameSuffix: NamespaceName,
-                           services: List[ServiceConfig])
+                           services: List[ServiceConfig],
+                           serviceAccountSuffix: KubernetesServiceAccount)
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/GalaxyAppConfig.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/GalaxyAppConfig.scala
@@ -1,11 +1,10 @@
 package org.broadinstitute.dsde.workbench.leonardo.config
 
-import org.broadinstitute.dsde.workbench.google2.KubernetesSerializableName.NamespaceName
-import org.broadinstitute.dsde.workbench.leonardo.{ChartName, KubernetesServiceAccount, ReleaseName, ServiceConfig}
+import org.broadinstitute.dsde.workbench.leonardo.{ChartName, ServiceConfig}
 
-case class GalaxyAppConfig(releaseNameSuffix: ReleaseName,
+case class GalaxyAppConfig(releaseNameSuffix: String,
                            chart: ChartName,
-                           namespaceNameSuffix: NamespaceName,
+                           namespaceNameSuffix: String,
                            services: List[ServiceConfig],
-                           serviceAccountSuffix: KubernetesServiceAccount,
+                           serviceAccountSuffix: String,
                            uninstallKeepHistory: Boolean)
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/GalaxyAppConfig.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/GalaxyAppConfig.scala
@@ -1,6 +1,9 @@
 package org.broadinstitute.dsde.workbench.leonardo.config
 
 import org.broadinstitute.dsde.workbench.google2.KubernetesSerializableName.NamespaceName
-import org.broadinstitute.dsde.workbench.leonardo.{ReleaseName, ServiceConfig}
+import org.broadinstitute.dsde.workbench.leonardo.{ChartName, ReleaseName, ServiceConfig}
 
-case class GalaxyAppConfig(releaseName: ReleaseName, namespaceNameSuffix: NamespaceName, services: List[ServiceConfig])
+case class GalaxyAppConfig(releaseName: ReleaseName,
+                           chart: ChartName,
+                           namespaceNameSuffix: NamespaceName,
+                           services: List[ServiceConfig])
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/GalaxyAppConfig.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/GalaxyAppConfig.scala
@@ -7,4 +7,5 @@ case class GalaxyAppConfig(releaseNameSuffix: ReleaseName,
                            chart: ChartName,
                            namespaceNameSuffix: NamespaceName,
                            services: List[ServiceConfig],
-                           serviceAccountSuffix: KubernetesServiceAccount)
+                           serviceAccountSuffix: KubernetesServiceAccount,
+                           uninstallKeepHistory: Boolean)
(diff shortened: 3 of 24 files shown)


rtitle:
Just scanned, but overall looks great! Did you get past the `invalid value` issue? Maybe I'll give this branch a shot too

kyuksel:
Status as of the late morning: App creation fails with the error below

```
org.broadinstitute.dsde.workbench.leonardo.monitor.PubsubKubernetesError: An error occurred with a kubernetes operation from source app during action createGalaxyApp.
 
Original message: release galaxy-rls failed, and has been uninstalled due to atomic being set:

Deployment.apps "galaxy-rls-workflow" is invalid:

[spec.template.spec.nodeSelector: Invalid value: "\"cloud.google.com/gke-nodepool\"": prefix part a DNS-1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*'), 
spec.template.spec.nodeSelector: Invalid value: "\"cloud.google.com/gke-nodepool\"": name part must consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyName',  or 'my.name',  or '123-abc', regex used for validation is '([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9]')]
```

Curiously, `helm install`ing on the command line using the values string produced by the PR code doesn't reproduce the issue.

Haven't been able to reproduce the error on the sbt console using the Helm client repo either; though, I'm not certain if the issue I run into there may be happening before the values are parsed.

kyuksel:
Current status: Chart override string parsing appears to work now. Install ongoing 🤞 

>Did you get past the invalid value issue?

Yes. A few fixes that made the difference are included in [this diff](https://github.com/DataBiosphere/leonardo/pull/1587/files/2d6fe589d1f92b66e17d80c1c37ca650330a108a..c06a06020112d48c2c535c7632d4c26e98b0dfcf).

kyuksel:
Chart values passed in look like this:

```
nfs.storageClass.name=nfs-galaxy-rls-3

cvmfs.repositories.cvmfs-gxy-data-galaxy-rls-3=data.galaxyproject.org

cvmfs.repositories.cvmfs-gxy-main-galaxy-rls-3=main.galaxyproject.org

cvmfs.cache.alienCache.storageClass=nfs-galaxy-rls-3

galaxy.persistence.storageClass=nfs-galaxy-rls-3

galaxy.cvmfs.data.pvc.storageClassName=cvmfs-gxy-data-galaxy-rls-3

galaxy.cvmfs.main.pvc.storageClassName=cvmfs-gxy-main-galaxy-rls-3

galaxy.nodeSelector.cloud\.google\.com/gke-nodepool=k780ba7c-e503-45e9-a66d-ad6229a1e557

nfs.nodeSelector.cloud\.google\.com/gke-nodepool=k780ba7c-e503-45e9-a66d-ad6229a1e557

galaxy.ingress.path=/proxy/google/v1/apps/h-proj-2020-08-27-kyuksel/galaxy-rls-3/galaxy

galaxy.ingress.annotations.nginx\.ingress\.kubernetes\.io/proxy-redirect-from=https://437046468.jupyter-dev.firecloud.org

galaxy.ingress.annotations.nginx\.ingress\.kubernetes\.io/proxy-redirect-to=https://leonardo-fiab.dsde-dev.broadinstitute.org:30443

galaxy.ingress.hosts[0]=437046468.jupyter-dev.firecloud.org

galaxy.configs.galaxy\.yml.galaxy.single_user=hermione.owner@test.firecloud.org,galaxy.configs.galaxy\.yml.galaxy.admin_users=hermione.owner@test.firecloud.org
```

kyuksel:
Status update: The `ingress.path` fix doesn't seem to have fixed the two Galaxy pods coming up on the console

image

FWIW...also have been seeing the message `secrets "galaxy-rls-4-galaxy-secrets" not found` at the beginning of the installation although I can see that `secret` via `kubectl` on the command line  🤷‍♂️

kyuksel:
The final override values are below:

```
$ helm -n ky-glxy-app-0827-7-galaxy-ns get values galaxy-rls-4

USER-SUPPLIED VALUES:
cvmfs:
  cache:
    alienCache:
      storageClass: nfs-galaxy-rls-4
  repositories:
    cvmfs-gxy-data-galaxy-rls-4: data.galaxyproject.org
    cvmfs-gxy-main-galaxy-rls-4: main.galaxyproject.org
galaxy:
  configs:
    galaxy.yml:
      galaxy:
        admin_users: hermione.owner@test.firecloud.org
        single_user: hermione.owner@test.firecloud.org
  cvmfs:
    data:
      pvc:
        storageClassName: cvmfs-gxy-data-galaxy-rls-4
    main:
      pvc:
        storageClassName: cvmfs-gxy-main-galaxy-rls-4
  ingress:
    annotations:
      nginx.ingress.kubernetes.io/proxy-redirect-from: https://437046468.jupyter-dev.firecloud.org
      nginx.ingress.kubernetes.io/proxy-redirect-to: https://leonardo-fiab.dsde-dev.broadinstitute.org:30443
    hosts:
    - 437046468.jupyter-dev.firecloud.org
    path: /proxy/google/v1/apps/h-proj-2020-08-27-kyuksel/ky-glxy-app-0827-7/galaxy
  nodeSelector:
    cloud.google.com/gke-nodepool: k447d846-1c0a-4bd1-910d-65484f6809b8
  persistence:
    storageClass: nfs-galaxy-rls-4
nfs:
  nodeSelector:
    cloud.google.com/gke-nodepool: k447d846-1c0a-4bd1-910d-65484f6809b8
  storageClass:
    name: nfs-galaxy-rls-4
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/DateAccessedUpdater.scala
```diff
@@ -32,7 +32,7 @@ class DateAccessedUpdater[F[_]: ContextShift: Timer](
     (Stream.sleep[F](config.interval) ++ Stream.eval(check)).repeat
 
   private[monitor] val check: F[Unit] =
-    logger.info(s"Going to update dateAccessed") >> queue.tryDequeueChunk1(config.maxUpdate).flatMap { chunks =>
+    logger.debug(s"Going to update dateAccessed") >> queue.tryDequeueChunk1(config.maxUpdate).flatMap { chunks =>
```
kyuksel: This was causing a lot of verbosity 🙈

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -184,6 +186,8 @@ class GKEInterpreter[F[_]: Parallel: ContextShift: Timer](
         )
       )
 
+      _ <- logger.info(s"Successfully created cluster ${params.clusterId}!")
```
kyuksel: previously it was printing this message out late

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -456,6 +531,39 @@ class GKEInterpreter[F[_]: Parallel: ContextShift: Timer](
...
+                                                         userEmail: WorkbenchEmail): String = {
+    val release = config.galaxyAppConfig.releaseName
+    val proxyUrl = host(cluster, config.proxyConfig.proxyDomain).toString
+    val hostUrl = config.proxyConfig.getProxyServerHostName
+    // TODO Is the path below templated in any config? If not, consider adding
+    val ingressPath = s"/proxy/google/v1/apps/${cluster.googleProject.value}/${appName.value}/galaxy"
+
+    // Using the string interpolator raw""" since the chart keys include quotes to escape Helm
+    // value override special characters such as '.'
+    // https://helm.sh/docs/intro/using_helm/#the-format-and-limitations-of---set
```
kyuksel, line 683: `raw` interpolator may no longer be necessary since not using quotes in keys work the way we want. Even so, not sure if it buys anything to try to get rid of it 🤷‍♂️

kyuksel:
Also just noticed the following on my test cluster

image


which may explain `galaxy-rls-4-cvmfs-cache-preload-0` being `Unschedulable` (see previous screenshot above), which might in turn have prevented the web pod from coming up.

rtitle:
FYI this all worked and I demoed in in the IA sprint review! I still want to clean some stuff up next week, and then I'll move this to Ready for review.

kyuksel:
> FYI this all worked and I demoed in in the IA sprint review! I still want to clean some stuff up next week, and then I'll move this to Ready for review.

Yay! Thank you 😊

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -379,12 +381,12 @@ class GKEInterpreter[F[_]: Parallel: ContextShift: Timer](
...
       // Create namespace for nginx
       _ <- logger.info(
-        s"Creating namespace '${config.ingressConfig.namespace.value}' in cluster '${dbCluster.id}' | trace id: ${ctx.traceId}"
+        s"Creating namespace ${config.ingressConfig.namespace.value} in cluster ${dbCluster.id} | trace id: ${ctx.traceId}"
       )
       _ <- kubeService.createNamespace(dbCluster.getGkeClusterId, KubernetesNamespace(config.ingressConfig.namespace))
 
       _ <- logger.info(
-        s"Installing ingress helm chart '${config.ingressConfig.chart}' in cluster '${dbCluster.id}' | trace id: ${ctx.traceId}"
+        s"Installing ingress helm chart ${config.ingressConfig.chart} in cluster ${dbCluster.id} | trace id: ${ctx.traceId}"
```
kyuksel: Nit: Values read better to me in logs (i.e. `chart.value` instead of `chart`, and similarly for `id`)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -472,14 +474,24 @@ class GKEInterpreter[F[_]: Parallel: ContextShift: Timer](
...
+      // Poll galaxy until it starts up
+      _ <- (
+        Stream.sleep_(config.monitorConfig.createApp.interval) ++
+          Stream.eval(
+            galaxyDAO.isProxyAvailable(dbCluster.googleProject, appName)
+          )
+      ).repeatN(config.monitorConfig.createApp.maxAttempts)
+        .takeThrough(_ == false)
+        .compile
+        .lastOrError
```
kyuksel: This pattern looks nice and seems to be utilized in a few places so it may be nice to factor it out into a util method that accepts parameters for `Stream.eval()`, `repeatN()` and `takeThrough()`.

kyuksel (commented):
Looks great 👍 A couple of minor comments.

Also, what was the value string fix that made it work?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
rtitle: +1 me too - will add `.value`. 

Also maybe it's just me but I really don't like quoted strings in logs like `s"Installing ingress helm chart '${config.ingressConfig.chart}'"`. Not sure why.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -405,6 +434,65 @@ class GKEInterpreter[F[_]: Parallel: ContextShift: Timer](
...
+      // Invoke helm
+      _ <- helmClient
+        .installChart(
+          org.broadinstitute.dsp.Release(releaseName.value),
+          org.broadinstitute.dsp.Chart(config.galaxyAppConfig.chart.value),
+          org.broadinstitute.dsp.Values(chartValues)
+        )
+        .run(helmAuthContext)
+
+      // Poll galaxy until it starts up
```
kyuksel: Do we need a TODO to check the applicable Galaxy pod(s) as well when the dependencies are in place?

rtitle: Yeah - I added a TODO in `HttpGalaxyDAO` but maybe it belongs here.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
rtitle: Maybe… although I feel like the fs2 combinators are already pretty general and easy to read, not sure a util method would make things any simpler.

rtitle:
> Also, what was the value string fix that made it work?

I thought it was something around `galaxy.ingress.path` or `galaxy.ingress.host` but looking at commits I'm not sure. Another I found was that releaseName apparently needs to be unique within a cluster (not sure why) so I made it `releaseNameSuffix` in config.

Dockerfile
```diff
@@ -9,7 +9,7 @@ RUN mkdir /helm-go-lib-build && \
     cd /helm-go-lib-build && \
     git clone https://github.com/broadinstitute/helm-scala-sdk.git && \
     cd helm-scala-sdk && \
-    git checkout master && \
+    git checkout rt-uninstall && \
```
rtitle: TODO change back to master once https://github.com/broadinstitute/helm-scala-sdk/pull/5 is merged

project/Dependencies.scala
```diff
@@ -26,7 +26,7 @@ object Dependencies {
   val workbenchOpenTelemetryV = "0.1-e66171c"
   val workbenchErrorReportingV = "0.1-92fcd96"
 
-  val helmScalaSdkV = "0.0.1-RC1"
+  val helmScalaSdkV = "0.0.1-rt-uninstall-SNAPSHOT"
```
rtitle: TODO: change back to 0.0.1-RC1 once https://github.com/broadinstitute/helm-scala-sdk/pull/5 is merged

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreterSpec.scala
```diff
@@ -12,6 +12,7 @@ import org.broadinstitute.dsde.workbench.google2.mock.{
 import org.broadinstitute.dsde.workbench.leonardo.CommonTestData._
```
rtitle, line 22: TODO: add more specs to `GKEInterpreterSpec`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/GalaxyAppConfig.scala
```diff
@@ -1,6 +1,11 @@
...
 import org.broadinstitute.dsde.workbench.google2.KubernetesSerializableName.NamespaceName
-import org.broadinstitute.dsde.workbench.leonardo.{ReleaseName, ServiceConfig}
+import org.broadinstitute.dsde.workbench.leonardo.{ChartName, KubernetesServiceAccount, ReleaseName, ServiceConfig}
 
-case class GalaxyAppConfig(releaseName: ReleaseName, namespaceNameSuffix: NamespaceName, services: List[ServiceConfig])
+case class GalaxyAppConfig(releaseNameSuffix: ReleaseName,
+                           chart: ChartName,
+                           namespaceNameSuffix: NamespaceName,
+                           services: List[ServiceConfig],
+                           serviceAccountSuffix: KubernetesServiceAccount,
```
Qi77Qi: is this just suffix or a complete ksa?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/DateAccessedUpdater.scala
Qi77Qi: we can remove this line too...think I added this when I was debugging some issue

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -264,23 +279,69 @@ class GKEInterpreter[F[_]: Parallel: ContextShift: Timer](
...
+      // TODO populate annotations in KSA for Workload Identity once wb-libs GKE client is fixed
+      ksa = KubernetesModels.KubernetesServiceAccount(ServiceAccountName(ksaName.value), Map.empty)
+
+      _ <- logger.info(
+        s"Creating Kubernetes service account ${ksaName.value} for app  ${app.appName.value} in cluster ${gkeClusterId.toString} | trace id: ${ctx.traceId}"
+      )
+      _ <- kubeService.createServiceAccount(gkeClusterId, ksa, KubernetesNamespace(namespaceName))
+
+      // update KSA in DB
+      _ = appQuery.updateKubernetesServiceAccount(app.id, ksaName).transaction
```
Qi77Qi: should this be `_ <-`? if true, probably should add unit test for this

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -324,23 +387,44 @@ class GKEInterpreter[F[_]: Parallel: ContextShift: Timer](
...
       dbAppOpt <- KubernetesServiceDbQueries.getActiveFullAppByName(params.googleProject, params.appName).transaction
       dbApp <- F.fromOption(dbAppOpt, AppNotFoundException(params.googleProject, params.appName, ctx.traceId))
-      // TODO helm delete app and monitor for deletion
+
+      app = dbApp.app
+      namespaceName = app.appResources.namespace.name
+      dbCluster = dbApp.cluster
+      gkeClusterId = dbCluster.getGkeClusterId
+
+      _ <- logger.info(
```
Qi77Qi, line 410: should this be `debug`?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -400,11 +471,157 @@ class GKEInterpreter[F[_]: Parallel: ContextShift: Timer](
...
+    for {
+      ctx <- ev.ask
+
+      _ <- logger.info(
+        s"Installing helm chart ${config.galaxyAppConfig.chart.value} for app ${appName.value} in cluster ${dbCluster.getGkeClusterId.toString} | trace id: ${ctx.traceId}"
+      )
+
+      helmAuthContext <- getHelmAuthContext(googleCluster, dbCluster.id, namespaceName)
+
+      releaseName = ReleaseName(s"${appName.value}-${config.galaxyAppConfig.releaseNameSuffix.value}")
```
Qi77Qi: would it make sense to suffix a timestamp in the release name?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
Qi77Qi: you can use `streamFUntilDone` directly I think... you just need to provide a `DoneCheckable[Boolean]` instance

jdcanas: you could use streamFUntilDone https://github.com/broadinstitute/workbench-libs/blob/develop/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/package.scala#L121. Though it may be worth adding a param for onTimeout to it...

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -400,11 +471,157 @@ class GKEInterpreter[F[_]: Parallel: ContextShift: Timer](
...
+
+      // Invoke helm
+      _ <- helmClient
+        .uninstall(org.broadinstitute.dsp.Release(releaseName.value), config.galaxyAppConfig.uninstallKeepHistory)
+        .run(helmAuthContext)
+
+      isDone <- (Stream.sleep_(config.monitorConfig.deleteApp.interval) ++
```
Qi77Qi: same. you should be able to use `streamFUntilDone` and provide a `DoneCheckable[List[Pod]]` instance (hand wavy signature)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/GalaxyAppConfig.scala
rtitle: Just a suffix - I'm generating the ksa as `${appName}-${serviceAccountSuffix}`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
rtitle: Ahh good catch, should be `<-`. I'll try to add more specs to `GKEInterpreterSpec`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
rtitle: I dunno, I added a lot of `info` statements to make it clear what's going on when looking at logs. I feel like it's helpful.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
rtitle: I think it would - especially as we currently have the `keepHistory` option when deleting charts. Will update.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
rtitle: Forgot about `streamFUntilDone`! Thanks

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/GalaxyAppConfig.scala
rtitle: Maybe the suffixes should be `String`s? 🤔

Qi77Qi: I think `String` sounds good to me...there's no other `String` param in this data type anyway, so there's no chance we'll mix it up with another param

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
rtitle: Hm.. we need `releaseName` at both install and delete time, so if we add a timestamp then we should probably persist it to the database. This may be a good idea anyway, but I just added a TODO for now.

rtitle:
Thanks @Qi77Qi -- applied feedback so far I believe, still need to add some `GKEInterpreterSpec` tests

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
kyuksel: Single quotes in log messages allow my brain to pause, like with `,` and easily identify that it's a value and not a regular word in the sentence. That being said, it's a personal preference and I don't expect it to work the same way for others 😊

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
kyuksel: I find them helpful as well.

http/src/main/resources/reference.conf
```diff
@@ -241,6 +241,8 @@ gke {
     releaseNameSuffix = "galaxy-rls"
     chart = "galaxy/galaxykubeman"
     namespaceNameSuffix = "galaxy-ns"
+    serviceAccountSuffix = "galaxy-ksa"
+    uninstallKeepHistory = true
```
kyuksel, line 249: might want to consider adding a comment clarifying what history it is we're keeping

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -59,19 +62,20 @@ class GKEInterpreter[F[_]: Parallel: ContextShift: Timer](
...
-        s"Beginning cluster creation for cluster ${params.clusterId} in project ${params.googleProject} | trace id: ${ctx.traceId}"
-      )
 
       // Grab records from the database
       clusterOpt <- kubernetesClusterQuery.getMinimalClusterById(params.clusterId).transaction
       dbCluster <- F.fromOption(
         clusterOpt,
         KubernetesClusterNotFoundException(
-          s"Failed kubernetes cluster creation. Cluster with id ${params.clusterId} not found in database"
+          s"Failed kubernetes cluster creation. Cluster with id ${params.clusterId.id} not found in database | trace id: ${ctx.traceId}"
```
kyuksel, line 71: 👍

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -59,19 +62,20 @@ class GKEInterpreter[F[_]: Parallel: ContextShift: Timer](
...
         )
       )
 
+      _ <- logger.info(
+        s"Beginning cluster creation for cluster ${dbCluster.getGkeClusterId.toString} | trace id: ${ctx.traceId}"
+      )
```
kyuksel, line 77: Should we say "...for cluster with ID ${dbCluster.getGkeClusterId.toString}..."? Similarly below.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -414,53 +458,140 @@ class GKEInterpreter[F[_]: Parallel: ContextShift: Timer](
...
+        val msg =
+          s"Galaxy deletion has failed or timed out for app ${appName.value} in cluster ${dbCluster.getGkeClusterId.toString}. The following pods are not in a terminal state: ${last
+            .filterNot(isPodDone)
+            .map(_.name.value)
+            .mkString(", ")} | trace id: ${ctx.traceId}"
+        logger.error(msg)
+
+        // TODO (RT): currently there seems to be a bug causing Galaxy pods to not terminate when `helm delete` is called.
+        // So we're currently _not_ throwing an error if deletion times out, but we should uncomment the below
+        // line once the Galaxy bug is fixed.
```
kyuksel, line 576: Do we know if those `Terminating` pods are removed once the namespace is deleted?

kyuksel (commented):
The past two days' changes look good. Minor questions and comments are inline.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
rtitle: `getGkeClusterId.toString` returns a string like `s"projects/${project.value}/locations/${location.value}/clusters/${clusterName.value}"`. I like it because it's pretty imformative in logging, but because of this I don't feel that including "with ID" is necessary. (Perhaps `getGkeClusterId` itself should be named differently).

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
rtitle: They are not - the namespace itself gets stuck in Terminating and the pods also stay in Terminating.

It sounded like Alex had an idea about what the issue is.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
kyuksel: Oh right. No need for `with ID` then indeed. 

> Perhaps getGkeClusterId itself should be named differently)

Yea that and `KubernetesClusterId` seem misleading since that's not usually what we mean by `id`s in other places.

rtitle: I feel like there are several places where we might want to refactor the Leo/wb-libs k8s models but I'm holding off.. maybe can do as tech debt once the alpha release is stable.

kyuksel: Sounds good. Could add a TODO for future reference.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
rtitle: Added assertions to `LeoPubsubMessageSubscriberSpec` that the KSA is persisted to the DB

PR #1590, opened 2020-08-28
QA-1205: Making another merge to see if it resolves the ignore not applying

xuscrbw:
We noticed that the test we ignored in the previous PR is not still being ignored. This PR is for the purpose of seeing if this gets the ignore to apply.
---

In all cases:


Qi77Qi:
think we can close this now?

PR #1592, opened 2020-09-02
IA-2167 fix a bug where we're accidentally validating https path as gs path

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-2167

---

In all cases:

### DIFF
--- a/http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeServiceInterpSpec.scala
+++ b/http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeServiceInterpSpec.scala
@@ -2,6 +2,7 @@ package org.broadinstitute.dsde.workbench.leonardo
 package http
 package service
 
+import java.net.URL
 import java.time.Instant
 import java.util.UUID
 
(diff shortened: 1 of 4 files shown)


http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeServiceInterpSpec.scala
```diff
@@ -137,6 +138,39 @@ class RuntimeServiceInterpSpec extends AnyFlatSpec with LeonardoTestSuite with T
...
   }
 
+  it should "successfully accept https as user script and user startup script" in isolatedDbTest {
+    val userInfo = UserInfo(OAuth2BearerToken(""), WorkbenchUserId("userId"), WorkbenchEmail("user1@example.com"), 0) // this email is white listed
+    val googleProject = GoogleProject("googleProject")
+    val runtimeName = RuntimeName("clusterName2")
+    val request = emptyCreateRuntimeReq.copy(
+      jupyterUserScriptUri = Some(
+        UserScriptPath.Http(
+          new URL("https://api-dot-all-of-us-workbench-test.appspot.com/static/start_notebook_cluster.sh")
```
calbach, line 148: Will this endpoint still reject non-GCS and non http URIs?

calbach (commented):
Looks good AFAICT. I will let Rob approve though since I only have a surface-level understanding of Scala.

jdcanas (approved):
LGTM

rtitle (approved):
Looks good - so we are handling http paths in the init script, they were just failing validation in front Leo?

Qi77Qi:
@calbach Yes

@rtitle in the decoder, we're accepting both HTTP and Gcs. But in front leo, for all gcsPaths, we're validating 1. they're in correct format (this part overlaps decoder logic a bit..maybe we can remove it) 2. object actually exists in GCS. The bug was we were passing user startup script and user startup script to the gcs path validation code while we should only do that if these 2 are gcs path in the first place (the prism code will return `None` if it's not gcs path).

rtitle:
@Qi77Qi yeah that makes sense - just want to make sure that both http and gcs scripts are actually handled and applied in `init-actions.sh` and `gce-init.sh` and `startup.sh`.

Edit: seems like we are:
https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/resources/init-resources/gce-init.sh#L374-L378
https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/resources/init-resources/init-actions.sh#L391-L395

Qi77Qi:
ah yea...I think init scripts are shared between old and new apis, so they should be good

PR #1601, opened 2020-09-09
[IA-2182] adding cleanup code for app creation failures

jdcanas:
This PR adds cleanup for nodepools, disks, and app-related kubernetes entities when app creation fails.

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
@@ -26,10 +26,10 @@ import org.broadinstitute.dsde.workbench.google2.{
 }
 import org.broadinstitute.dsde.workbench.leonardo.AsyncTaskProcessor.Task
 import org.broadinstitute.dsde.workbench.leonardo.db._
-import org.broadinstitute.dsde.workbench.leonardo.http.service.AppNotFoundException
 import org.broadinstitute.dsde.workbench.leonardo.http.{cloudServiceSyntax, _}
 import org.broadinstitute.dsde.workbench.leonardo.model.{LeoAuthProvider, LeoException}
 import org.broadinstitute.dsde.workbench.leonardo.monitor.LeoPubsubMessage._
+import org.broadinstitute.dsde.workbench.leonardo.monitor.PubsubHandleMessageError.DiskNotFound
 import org.broadinstitute.dsde.workbench.leonardo.util._
 import org.broadinstitute.dsde.workbench.model.google.{GcsObjectName, GcsPath}
 import org.broadinstitute.dsde.workbench.model.{ErrorReport, TraceId, WorkbenchException}
@@ -537,21 +537,17 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift: Parallel](
   ): F[Unit] =
     for {
       ctx <- ev.ask
+      _ <- logger.info(s"Beginning disk deletion for ${diskId} | trace id: ${ctx.traceId}")
       diskOpt <- persistentDiskQuery.getById(diskId).transaction
       disk <- diskOpt.fold(
         F.raiseError[PersistentDisk](PubsubHandleMessageError.DiskNotFound(diskId))
       )(F.pure)
-      _ <- if (disk.status != DiskStatus.Deleting)
-        F.raiseError[Unit](
-          PubsubHandleMessageError.DiskInvalidState(diskId, disk.projectNameString, disk)
-        )
-      else F.unit
       operation <- googleDiskService.deleteDisk(disk.googleProject, disk.zone, disk.name)
       whenDone = persistentDiskQuery.delete(diskId, ctx.now).transaction[F].void >> authProvider.notifyResourceDeleted(
         disk.samResource,
         disk.auditInfo.creator,
         disk.googleProject
-      )
+      ) >> logger.info(f"Completed disk deletion for ${diskId} | trace id: ${ctx.traceId}")
       whenTimeout = F.raiseError[Unit](
         new TimeoutException(s"Fail to delete disk ${disk.name.value} in a timely manner")
       )
@@ -726,16 +722,41 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift: Parallel](
       task = for {
         _ <- parPreAppCreationSetup
         // create and monitor app
-        _ <- gkeInterp.createAndPollApp(CreateAppParams(msg.appId, msg.project, msg.appName)).adaptError {
-          case e =>
-            PubsubKubernetesError(
-              AppError(e.getMessage, ctx.now, ErrorAction.CreateGalaxyApp, ErrorSource.App, None),
-              Some(msg.appId),
-              false,
-              None,
-              None
-            )
-        }
+        _ <- gkeInterp
+          .createAndPollApp(CreateAppParams(msg.appId, msg.project, msg.appName))
+          .onError {
+            case _ =>
+              // clean-up resources in the event of an app creation error
+              for {
+                _ <- logger.info(
+                  s"Attempting to clean up resources due to app creation error for app ${msg.appName} in project ${msg.project}. | trace id: ${ctx.traceId}"
+                )
+                // we need to look up the app because we always want to clean up the nodepool associated with an errored app, even if it was pre-created
+                appOpt <- KubernetesServiceDbQueries.getFullAppByName(msg.project, msg.appId).transaction
+                // note that this will only clean up the disk if it was created as part of this app creation.
+                // it should not clean up the disk if it already existed
+                _ <- appOpt.traverse { app =>
+                  val deleteMsg =
+                    DeleteAppMessage(msg.appId, msg.appName, app.nodepool.id, msg.project, msg.createDisk, msg.traceId)
+                  // This is a good-faith attempt at clean-up. We do not want to take any action if clean-up fails for some reason.
+                  deleteApp(deleteMsg, true).handleErrorWith { e =>
+                    logger.error(e)(
+                      s"An error occurred during resource clean up for app ${msg.appName} in project ${msg.project}. | trace id: ${ctx.traceId}"
+                    )
+                  }
+                }
+              } yield ()
+          }
+          .adaptError {
+            case e =>
+              PubsubKubernetesError(
+                AppError(e.getMessage, ctx.now, ErrorAction.CreateGalaxyApp, ErrorSource.App, None),
+                Some(msg.appId),
+                false,
+                None,
+                None
+              )
+          }
       } yield ()
 
       _ <- asyncTasks.enqueue1(
@@ -785,6 +806,11 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift: Parallel](
   private[monitor] def handleDeleteAppMessage(msg: DeleteAppMessage)(
     implicit ev: ApplicativeAsk[F, AppContext]
   ): F[Unit] =
+    deleteApp(msg, false)
+
+  private[monitor] def deleteApp(msg: DeleteAppMessage, sync: Boolean)(
+    implicit ev: ApplicativeAsk[F, AppContext]
+  ): F[Unit] =
     for {
       ctx <- ev.ask
 
@@ -828,9 +854,11 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift: Parallel](
         _ <- deleteDisk
       } yield ()
 
-      _ <- asyncTasks.enqueue1(
-        Task(ctx.traceId, task, Some(handleKubernetesError), ctx.now)
-      )
+      _ <- if (sync) task
+      else
+        asyncTasks.enqueue1(
+          Task(ctx.traceId, task, Some(handleKubernetesError), ctx.now)
+        )
     } yield ()
 
   private def handleKubernetesError(e: Throwable)(implicit ev: ApplicativeAsk[F, AppContext]): F[Unit] =
@@ -863,21 +891,18 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift: Parallel](
     implicit ev: ApplicativeAsk[F, AppContext]
   ): F[Unit] =
     msg.createDisk match {
-      case true =>
+      case Some(diskId) =>
         for {
           ctx <- ev.ask
-          _ <- logger.info(s"Beginning disk creation for app ${msg.appId} | trace id: ${ctx.traceId}")
-          getAppOpt <- KubernetesServiceDbQueries.getActiveFullAppByName(msg.project, msg.appName).transaction
-          getApp <- F.fromOption(getAppOpt, AppNotFoundException(msg.project, msg.appName, ctx.traceId))
+          _ <- logger.info(s"Beginning disk creation for app ${msg.appName} | trace id: ${ctx.traceId}")
+          diskOpt <- persistentDiskQuery.getById(diskId).transaction
           disk <- F.fromOption(
-            getApp.app.appResources.disk,
-            AppCreationException(
-              s"create disk was true for create app message, but app ${getApp.app.id} does not have a disk id saved"
-            )
+            diskOpt,
+            DiskNotFound(diskId)
           )
           _ <- createDisk(CreateDiskMessage.fromDisk(disk, Some(ctx.traceId)), true)
         } yield ()
-      case false => F.unit
+      case None => F.unit
     }
 
   private def createRuntimeErrorHandler(msg: CreateRuntimeMessage, now: Instant)(e: Throwable): F[Unit] =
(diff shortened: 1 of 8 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -537,21 +537,17 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift: Parallel](
   ): F[Unit] =
     for {
       ctx <- ev.ask
+      _ <- logger.info(s"Beginning disk deletion for ${diskId} | trace id: ${ctx.traceId}")
       diskOpt <- persistentDiskQuery.getById(diskId).transaction
       disk <- diskOpt.fold(
         F.raiseError[PersistentDisk](PubsubHandleMessageError.DiskNotFound(diskId))
       )(F.pure)
-      _ <- if (disk.status != DiskStatus.Deleting)
```
jdcanas, line 544: the original intention here was to verify the state did not change between front leo queuing the message and back leo receiving it. Its a bit duplicative though, because front leo also does this verification. We now also use this function for clean-up of freshly-created disks without going through front Leo too, so removing the status check.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/leoPubsubMessageSubscriberModels.scala
```diff
@@ -210,7 +210,7 @@ object LeoPubsubMessage {
                                     appName: AppName,
                                     nodepoolId: Option[NodepoolLeoId],
                                     project: GoogleProject,
-                                    createDisk: Boolean,
+                                    createDisk: Option[DiskId],
```
jdcanas, line 213: this was a bit clunky as a Boolean as it duplicated the same disk lookup for the app in front leo and back leo.

jdcanas:
jenkins retest

jdcanas:
jenkins retest, unrelated failure

jdcanas:
jenkins retest

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
kyuksel: I'm trying to understand your comment. If the original intention was to verify the state did not change between front leo queuing the message and back leo receiving it then wouldn't this still be the right place to do it as front leo cannot do this verification by nature.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -726,16 +722,41 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift: Parallel](
...
+                )
+                // we need to look up the app because we always want to clean up the nodepool associated with an errored app, even if it was pre-created
+                appOpt <- KubernetesServiceDbQueries.getFullAppByName(msg.project, msg.appId).transaction
+                // note that this will only clean up the disk if it was created as part of this app creation.
+                // it should not clean up the disk if it already existed
+                _ <- appOpt.traverse { app =>
+                  val deleteMsg =
+                    DeleteAppMessage(msg.appId, msg.appName, app.nodepool.id, msg.project, msg.createDisk, msg.traceId)
+                  // This is a good-faith attempt at clean-up. We do not want to take any action if clean-up fails for some reason.
+                  deleteApp(deleteMsg, true).handleErrorWith { e =>
```
kyuksel: If we don't want to take any action when clean-up fails, why are we setting `deleteApp`'s `sync` parameter to `true`?

kyuksel (commented):
Ran out of time when the childcare duty called. Will resume thereafter.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -726,16 +722,41 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift: Parallel](
...
+            case _ =>
+              // clean-up resources in the event of an app creation error
+              for {
+                _ <- logger.info(
+                  s"Attempting to clean up resources due to app creation error for app ${msg.appName} in project ${msg.project}. | trace id: ${ctx.traceId}"
+                )
+                // we need to look up the app because we always want to clean up the nodepool associated with an errored app, even if it was pre-created
+                appOpt <- KubernetesServiceDbQueries.getFullAppByName(msg.project, msg.appId).transaction
+                // note that this will only clean up the disk if it was created as part of this app creation.
+                // it should not clean up the disk if it already existed
```
kyuksel, line 737: Makes sense. What tells us here that a disk already existed vs. was created for this app?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -828,9 +854,11 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift: Parallel](
         _ <- deleteDisk
```
kyuksel, line 856: Should be able to delete all three above in parallel?

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriberSpec.scala
```diff
@@ -566,7 +563,7 @@ class LeoPubsubMessageSubscriberSpec
     res.unsafeRunSync()
   }
 
-  it should "not handle DeleteDiskMessage when disk is not in Deleting status" in isolatedDbTest {
+  it should "handle DeleteDiskMessage when disk is not in Deleting status" in isolatedDbTest {
```
kyuksel, line 566: if we indeed don't need to check if disk is `Deleting` then we should update the test name above accordingly

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriberSpec.scala
```diff
@@ -1408,6 +1405,96 @@ class LeoPubsubMessageSubscriberSpec
...
+      queue <- InspectableQueue.bounded[IO, Task[IO]](10)
+      leoSubscriber = makeLeoSubscriber(asyncTaskQueue = queue, gkeInterp = gkeInterp)
+      asyncTaskProcessor = AsyncTaskProcessor(AsyncTaskProcessor.Config(10, 10), queue)
+      _ <- leoSubscriber.handleCreateAppMessage(msg)
+      _ <- withInfiniteStream(asyncTaskProcessor.process, assertions, maxRetry = 50)
+    } yield ()
+
+    res.unsafeRunSync()
+  }
+
```
kyuksel, line 1498: Would it be good to test the scenario where the disk pre-existed and wouldn't be cleaned up?

kyuksel (commented):
LGTM overall. A few questions and comments are inline.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -537,21 +537,17 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift: Parallel](
...
-          PubsubHandleMessageError.DiskInvalidState(diskId, disk.projectNameString, disk)
-        )
-      else F.unit
       operation <- googleDiskService.deleteDisk(disk.googleProject, disk.zone, disk.name)
       whenDone = persistentDiskQuery.delete(diskId, ctx.now).transaction[F].void >> authProvider.notifyResourceDeleted(
         disk.samResource,
         disk.auditInfo.creator,
         disk.googleProject
-      )
+      ) >> logger.info(f"Completed disk deletion for ${diskId} | trace id: ${ctx.traceId}")
```
rtitle: Maybe want `s".."` -- I think the `f` interpolater is for `printf`-like syntax

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
rtitle: This is already being done as an async task, so I think passing `sync = true` to `deleteApp` makes sense here.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
rtitle: I'm not sure (the TODO was from me). It's probably safer to wait for the `helm uninstall` to finish before deleting the underlying nodepool/disk.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -726,16 +722,41 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift: Parallel](
...
+              for {
+                _ <- logger.info(
+                  s"Attempting to clean up resources due to app creation error for app ${msg.appName} in project ${msg.project}. | trace id: ${ctx.traceId}"
+                )
+                // we need to look up the app because we always want to clean up the nodepool associated with an errored app, even if it was pre-created
+                appOpt <- KubernetesServiceDbQueries.getFullAppByName(msg.project, msg.appId).transaction
+                // note that this will only clean up the disk if it was created as part of this app creation.
+                // it should not clean up the disk if it already existed
+                _ <- appOpt.traverse { app =>
+                  val deleteMsg =
```
rtitle, line 739: One comment: I think `GKEInterpreter. deleteAndPollApp` will set the app status to `Deleted`. We probably want to leave it in `Error` status in this case, so the user can view the error and choose to delete the Error'd app themselves. Maybe `GKEInterpreter.deleteAndPollApp` should take a parameter or something to tell it whether to move to `Deleted` status.

rtitle:
Overall looks good. Just one real comment above: I think the app should remain in `Error` status if an error occurred -- e.g. the cleanup logic shouldn't move it to `Deleted`.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
kyuksel: oh I see

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
kyuksel: Good point. Justin and I had briefly touched on this when the issue of [recovering from cluster creation errors](https://broadworkbench.atlassian.net/browse/IA-2197) came up. As discussed there, I wonder if we should use a different status than `ERROR`/`DELETED` so offline clean-up jobs would know how to handle it? Or is this not necessary because resources in `ERROR` status (as far as Leo is concerned) but `DELETED` in Google are already properly handled?

Qi77Qi: is there a case where Leonardo care when a resource is in Error and it's `Deleted` in google, vs Error + still exists in google? To be clear, there should be only `Error` +`Deleted`, and `Error` + `Error` (`Error` + `Running` would be illegal, which I'm guessing this clean up code is addressing)

rtitle: Well our current pattern when an error occurs is to delete all underlying resources in Google but keep the record in Leo in Error status. That way the user can see that an error occurred, but are not spending money. Agree that the resource-validator will need to handle this situation but I feel like we can achieve this without the need for an additional status code (unless I'm missing something).

rtitle: I think currently this is the only pathway to `Error` -- so we only have `Error` + `Deleted`, there is no such thing (atm) as `Error` + `Error` for k8s apps.

So I feel like we don't need another status code at this time, maybe we can add one when there are more pathways to get an app into `Error` status (or we want to keep resources around in Google for whatever reason).

Qi77Qi: yea, to clarify, I don't think there's a need to introduce an additional status either...Just want to ask the question in case I'm missing something

kyuksel: Sounds like there is nothing new here that we aren't already handling as part of `Error` + `Deleted` scenarios 👍

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
jdcanas: Front Leo verifies that the disk can be deleted when it is requested. I don't know if there is a scenario in which the state can change between front leo messaging and back leo recieving it. Perhaps if an update request is made in the interim, but if a delete request is made and then an update request, I'm thinking Leo should still delete the disk

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
jdcanas: In the CreateAppMessage, disk ID is optional and only passed if disk creation is needed

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriberSpec.scala
jdcanas: Thats what I intended here with the wording change. Is it not clear?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
jdcanas: Oh, I was playing with python recently :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
jdcanas: So, deleteAndPollApp will set the app status to deleted. However, since an error occurred, this error will bubble up and the app will be marked as Error in the top-level handler.

There is a test to confirm this https://github.com/DataBiosphere/leonardo/pull/1601/files#diff-8e28b041fa5855e1d9f7498b905ed3f4R1469

rtitle: I thought of that but would that still result in `APP.destroyedDate` being set? I think the end result should be that `destroyedDate` is _unset_ so the Errord app is still queryable via GET endpoints

jdcanas: Yes, the destroyed date will be set. The alternative strikes me as a bit odd though, because it is indeed deleted in google and otherwise the user has to explicitly clean up the app in Leo. Its kind of like we are lying about it still existing for the purpose of showing them an error happened, while the list endpoint can still show it via the `includeDeleted` flag if it is desired. 

I'll change it just to avoid confusion from the UI team since this is how we currently handle errors.

rtitle: Yeah I see how it seems odd but that's how our queries work today -- e.g. `getActiveApp` queries `WHERE destroyedDate == dummyDate`. So any apps with a set `destroyedDate` are effectively invisible.

rtitle (approved):

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriberSpec.scala
kyuksel: Having `Deleting` status is no longer a special case so I'd think we not mention it in the test name, i.e.  "handle DeleteDiskMessage". What am I missing?

jdcanas:
jenkins retest

jdcanas:
jenkins retest

jdcanas:
jenkins retest... im seriously considering disabling `NotebookGCECustomizationSpec`, this is ridiculous. @xuscrbw , any luck so far?

xuscrbw:
Hm this is a bit strange. Right now we should have the "should install user specified notebook extensions" test in that spec disabled. The test that your last PR failed on is a problematic one that I will be merging a fix for soon. Can you see if that test is disabled?

PR #1602, opened 2020-09-10
[IA-2168] Enabling helm install'ing specific chart versions

kyuksel:
[JIRA ticket](https://broadworkbench.atlassian.net/browse/IA-2168)

This PR (along with the [companion helm-scala-sdk PR](https://github.com/broadinstitute/helm-scala-sdk/pull/6)) accomplishes both acceptance criteria. 

There is a remaining TODO to add chart name and version info to table
- `CLUSTER` for `ingress`
- `APP` for `galaxykubeman`
which I'll either append to this PR or include in a follow-up PR but that work is straightforward and I wanted to get feedback sooner rather than later.


core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/AsyncTaskProcessor.scala
```diff
@@ -29,14 +28,12 @@ final class AsyncTaskProcessor[F[_]: Timer](config: AsyncTaskProcessor.Config, a
...
-        .map(
-          Instant.ofEpochMilli
-        ) //TODO: use nowInstant once https://github.com/DataBiosphere/leonardo/pull/1404 is merged
+      now <- nowInstant[F]
       latency = (now.toEpochMilli - task.enqueuedTime.toEpochMilli).millis
       _ <- recordLatency(latency)
-      _ <- logger.info(s"executing ${task.traceId.asString}. Latency: ${latency.toSeconds} seconds")
+      _ <- logger.info(
+        s"Executing task with traceId ${task.traceId.asString} with latency of ${latency.toSeconds} seconds"
+      )
```
kyuksel, line 36: Unrelated to the ticket. Saw the TODO while looking at something else and couldn't resist removing it 😛

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
```diff
@@ -287,9 +287,6 @@ object AppType {
   def stringToObject: Map[String, AppType] = values.map(v => v.toString -> v).toMap
 }
 
-final case class ReleaseName(value: String) extends AnyVal
-final case class ChartName(value: String) extends AnyVal
-
```
kyuksel, line 292: couldn't see why we'd want to duplicate `Chart{Name,Version}` and `Release` which exist in `helm-scala-sdk`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/GalaxyAppConfig.scala
```diff
@@ -1,10 +1,15 @@
...
-                           uninstallKeepHistory: Boolean)
+final case class GalaxyAppConfig(releaseNameSuffix: String,
+                                 chartName: ChartName,
+                                 chartVersion: ChartVersion,
+                                 namespaceNameSuffix: String,
+                                 services: List[ServiceConfig],
+                                 serviceAccountSuffix: String,
+                                 uninstallKeepHistory: Boolean) {
+
+  def chartInfo: String = s"${chartName.asString}-${chartVersion.asString}"
```
kyuksel, line 14: convenience method that follows the format that we see with `helm ls` output to use in log messages

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -199,6 +199,7 @@ class GKEInterpreter[F[_]: Parallel: ContextShift: Timer](
         s"Successfully created cluster ${dbCluster.getGkeClusterId.toString}! | trace id: ${ctx.traceId}"
       )
 
+      // TODO: Should we move nginx installation to a separate step so it doesn't fail cluster creation?
```
kyuksel: I ran into a case where cluster was successfully created but nginx installation failed thereafter (because I'd specified an invalid nginx version). When I tried to create another app with the same project, it failed saying it can't create an app because the cluster is in `Error` status. It feels like we should **not** mark a cluster as `Error`ed just because ingress installation failed but rather incorporate checking if ingress is installed into the app creation flow. 

Any thoughts?

jdcanas: Hmm... I think the cluster transitioning to `Error` status in this case is correct. The ingress is a cluster-level resource, created once at cluster creation time. I don't think each app should concern itself with whether cluster-wide resources are set-up properly.

Whether this should block future app creation is another issue though. Perhaps we could consider doing something like allowing the creation of one non-errored cluster per project, as opposed to a blanket enforcement of 1 per project like we do now? I've ran into this before, and its fairly annoying to go into the DB and manually clean up everything. I'd imagine it will be even more painful if users run into this.

Just brainstorming, we could also proactively clean-up at cluster creation time if an error occurs by introducing a status along the lines of 'DELETED_WITH_ERROR' to capture that 1) it is deleted in google and 2) an error occurred (I feel 2 is an important thing to keep a record of with the cluster and why just transitioning it to `Deleted` isn't sufficient)

kyuksel: > The ingress is a cluster-level resource, created once at cluster creation time. I don't think each app should concern itself with whether cluster-wide resources are set-up properly.

That reasoning makes sense too.

I think, if we could recover from the case where we end up with an `Error`ed cluster so that users can still create apps somehow, this issue would be sufficiently addressed. Your suggestion 

> we could also proactively clean-up at cluster creation time if an error occurs by introducing a status along the lines of 'DELETED_WITH_ERROR' to capture that 1) it is deleted in google and 2) an error occurred (I feel 2 is an important thing to keep a record of with the cluster and why just transitioning it to Deleted isn't sufficient)

sounds like a good solution to me (perhaps `DELETED_AFTER_ERRORING` 😊)

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -1278,8 +1278,10 @@ paths:
                 $ref: "#/components/schemas/ErrorReport"
   "/api/google/v1/apps/{googleProject}/{appName}":
     post:
-      summary: Creates a new app in the given project with the given name
-      description: Default labels appName, googleProject, serviceAccount, and creator cannot be overridden.
+      summary: Creates a new app in the given project with the given appName
+      description: >
+        The specified appName is suffixed and the resulting string must adhere to Google's name validation regex ?:[a-z](?:[-a-z0-9]{0,38}[a-z0-9])?.
```
kyuksel, line 1283: ^^ is unrelated to the ticket and is to address a [comment from the UI team](https://broadinstitute.slack.com/archives/CA3NP1733/p1599759474089500)

kyuksel: supposedly one should be able to specify a regex `pattern` for parameter values so they are checked by Swagger but it didn't seem to work after a quick try 😞

jdcanas (approved):
LGTM. Let's make a follow-up ticket for recovering from a cluster error then?

kyuksel:
Thanks for the review @jdcanas!

> Let's make a follow-up ticket for recovering from a cluster error then?

Good idea. Ticket created: https://broadworkbench.atlassian.net/browse/IA-2197

http/src/main/resources/reference.conf
```diff
@@ -206,7 +206,10 @@ gke {
   ingress {
     namespace = "nginx"
     release = "nginx"
-    chart = "stable/nginx-ingress"
+    # TODO Switch to ingress-nginx because of:
```
rtitle, line 209: Oh thanks, I didn't realize `nginx-ingress` was deprecated

rtitle:
👍

PR #1603, opened 2020-09-10
Runtime API documentation fixes

calbach:


http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -2912,6 +2910,16 @@ components:
...
           type: string
           enum: [gce, dataproc]
+    OneOfRuntimeConfig
+      oneOf:
+        - $ref: "#/components/schemas/GceConfig"
+        - $ref: "#/components/schemas/GceWithPdConfig"
+        - $ref: "#/components/schemas/MachineConfig"
+      discriminator:
+        propertyName: cloudService
+        mapping:
```
calbach, line 2922: This corrects the usage of the discriminator field: https://swagger.io/specification/#discriminator-object . Note that you cannot have one value (gce) mapping to multiple oneOf values (GceConfig and GceConfigWithPD), so this still isn't really valid oneof usage. Would suggest just merging these two things into one.

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -3004,7 +3012,7 @@ components:
             workerDiskSize:
               type: integer
               description: >
-                Optional, the size in gigabytes of the disk on the master node.
+                Optional, the size in gigabytes of the disk on the worker nodes.
```
calbach, line 3017: Follow-up question here:

How does this parameter interact with `numberOfWorkerLocalSSDs` ?

rtitle: See https://cloud.google.com/dataproc/docs/reference/rest/v1/ClusterConfig#DiskConfig

`workerDiskSize` is passed to the Dataproc `bootDiskSizeGb` field and `numberOfWorkerLocalSSDs` is passed to `numLocalSsds`. 

I'm not completely confident in our support for `numberOfWorkerLocalSSDs` btw -- I don't think Terra UI exposes that option.

rtitle (approved):
LGTM

rtitle:
jenkins retest

rtitle:
jenkins retest

Qi77Qi (approved):

Qi77Qi:
tests failed on the extension one, which is commented out for now in develop branch..gonna rebase and push

Qi77Qi:
jenkins retest

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -2914,6 +2912,16 @@ components:
         cloudService:
           type: string
           enum: [gce, dataproc]
+    OneOfRuntimeConfig
```
rtitle, line 2915: Leo swagger isn't rendering this was merged, I think the cause is a missing `:` here… fixing.

rtitle: PR: https://github.com/DataBiosphere/leonardo/pull/1609

PR #1604, opened 2020-09-10
IA-2188 Fix terminal not loading issue with new images

Qi77Qi:
Tested in fiab. Notice in log, leo will check if the terminal exists first, and if not, creates a new terminal

```
[INFO] [19:16:50.882] [ioapp-compute-1] o.h.client.middleware.RequestLogger - HTTP/1.1 GET https://7071515981183754738.jupyter-dev.firecloud.org/notebooks/qi-billing/saturn-edc0a518-5341-4ca7-9297-806995bf96d0/api/terminals/1 Headers()
[INFO] [19:16:50.888] [pool-1-thread-5] o.h.client.middleware.ResponseLogger - HTTP/1.1 404 Not Found Headers(Date: Thu, 10 Sep 2020 19:16:50 GMT, Server: TornadoServer/5.1.1, Content-Type: application/json, X-Content-Type-Options: nosniff, Content-Security-Policy: frame-ancestors 'self'; report-uri /notebooks/qi-billing/saturn-edc0a518-5341-4ca7-9297-806995bf96d0/api/security/csp-report; default-src 'none', Access-Control-Allow-Origin: *, Content-Length: 52)
[INFO] [19:16:50.889] [ioapp-compute-5] o.h.client.middleware.RequestLogger - HTTP/1.1 POST https://7071515981183754738.jupyter-dev.firecloud.org/notebooks/qi-billing/saturn-edc0a518-5341-4ca7-9297-806995bf96d0/api/terminals Headers()
[INFO] [19:16:51.909] [pool-1-thread-5] o.h.client.middleware.ResponseLogger -
```

Previously the launch terminal button will work out of the box, but with new jupyter version, we need to do a API call to notebook for creating a terminal before we can navigate to the terminal URL. Hence, we're intercepting calls to `jupyter/terminals/{number}` and check if the terminal exists, if not, we create it before establishing websocket; otherwise, we navigate to terminal directly

---

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpJupyterDAO.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpJupyterDAO.scala
@@ -2,7 +2,7 @@ package org.broadinstitute.dsde.workbench.leonardo.dao
 
 import cats.effect.{Concurrent, ContextShift, Timer}
 import cats.implicits._
-import com.typesafe.scalalogging.LazyLogging
+import io.chrisdavenport.log4cats.Logger
 import io.circe.Decoder
 import org.broadinstitute.dsde.workbench.leonardo.RuntimeName
 import org.broadinstitute.dsde.workbench.leonardo.dao.ExecutionState.{Idle, OtherState}
@@ -14,9 +14,11 @@ import org.http4s.circe.CirceEntityDecoder._
 import org.http4s.client.Client
 import org.http4s.{Method, Request, Uri}
 
-class HttpJupyterDAO[F[_]: Timer: ContextShift: Concurrent](val runtimeDnsCache: RuntimeDnsCache[F], client: Client[F])
-    extends JupyterDAO[F]
-    with LazyLogging {
+//Jupyter server API doc https://github.com/jupyter/jupyter/wiki/Jupyter-Notebook-Server-API
+class HttpJupyterDAO[F[_]: Timer: ContextShift](val runtimeDnsCache: RuntimeDnsCache[F], client: Client[F])(
+  implicit F: Concurrent[F],
+  logger: Logger[F]
+) extends JupyterDAO[F] {
   def isProxyAvailable(googleProject: GoogleProject, runtimeName: RuntimeName): F[Boolean] =
     Proxy.getRuntimeTargetHost[F](runtimeDnsCache, googleProject, runtimeName) flatMap {
       case HostReady(targetHost) =>
@@ -30,7 +32,7 @@ class HttpJupyterDAO[F[_]: Timer: ContextShift: Concurrent](val runtimeDnsCache:
             )
           )
           .handleError(_ => false)
-      case _ => Concurrent[F].pure(false)
+      case _ => F.pure(false)
     }
 
   def isAllKernelsIdle(googleProject: GoogleProject, runtimeName: RuntimeName): F[Boolean] =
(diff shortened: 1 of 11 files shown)


jdcanas:
Would be good to unit test this functionality with mocks and the route itself. I understand the time crunch though, LGTM

jdcanas (approved):

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/ProxyService.scala
```diff
@@ -211,6 +211,32 @@ class ProxyService(
       r <- proxyInternal(hostContext, request)
     } yield r
 
+    res.onError {
+      case e =>
+        IO(
+          logger.warn(
+            s"${ctx.traceId} | proxy request failed for ${userInfo.userEmail.value} ${googleProject.value} ${runtimeName.asString}",
```
kyuksel, line 218: May be easier to parse the log message if we specify what `${...}` values are...something like

```suggestion
            s"${ctx.traceId} | proxy request failed for user/project/runtime ${userInfo.userEmail.value}/${googleProject.value}/${runtimeName.asString}"
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpJupyterDAO.scala
```diff
@@ -48,9 +50,43 @@ class HttpJupyterDAO[F[_]: Timer: ContextShift: Concurrent](val runtimeDnsCache:
...
+              uri = Uri.unsafeFromString(
+                s"https://${targetHost.toString}/notebooks/${googleProject.value}/${runtimeName.asString}/api/terminals"
+              )
+            )
+          )
+          .flatMap(res => if (res) F.unit else logger.error("Fail to create new terminal"))
+      case _ => F.unit
+    }
+
+  override def ifTerminalExist(googleProject: GoogleProject,
```
kyuksel: I wonder if 
```suggestion
  override def terminalExists(googleProject: GoogleProject,
```
would read more naturally?

kyuksel (approved):
LGTM 👍 Thanks for the fix!

kyuksel:
> Would be good to unit test this functionality with mocks and the route itself. I understand the time crunch though

Could create a follow-up ticket to add the unit tests

Qi77Qi:
@jdcanas I was thinking the unit test wouldn't be very useful cuz all it can verify is `post` API happened before we open websocket...but I guess I can add it if people find it useful still

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpJupyterDAO.scala
Qi77Qi: I tend to add `is` or `if` if the return type is a `Boolean`...I can add `s` at the end to make it reads better

kyuksel: I guess this may be a personal convention. I also take into account the call site, which, in the case of methods returning `Boolean` tends to have an `if` clause so the method name containing another `if` doesn't read well...but I don't feel strongly 🤷‍♂️

Qi77Qi: renamed

rtitle:
👍

PR #1605, opened 2020-09-10
QA-1252: Removing extensions

xuscrbw:
Automation tests are failing at the moment due to the following error:
Found this relevant error. AttributeError: module 'nbconvert.exporters' has no attribute 'WebPDFExporter'

This error is causing the notebook to throw a 500 internal server error. This WebPDFExporter is part of the new 6.0 version changes, so it basically can't find the module.

[https://nbconvert.readthedocs.io/en/latest/changelog.html)]

The changes here remove the two extensions that seem problematic. I've kept the translate extension, as it is what the test is really testing. I do realize that in removing the two extensions I am remove a check for that part of the test. I hope that once we update we can turn this back. Should be a straightforward fix.

Note: I am for some reason running into UI component differences on my local environment, so I am trying to test these changes in this PR. 
---

In all cases:

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoTestUtils.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoTestUtils.scala
@@ -77,9 +77,12 @@ trait LeonardoTestUtils
 
   val multiExtensionClusterRequest = UserJupyterExtensionConfig(
     //TODO: Re-enable the below two lines when nbconvert is updated to 6.0.1
-    //nbExtensions = Map("map" -> "gmaps"),
-    //combinedExtensions = Map("pizza" -> "pizzabutton")
+    nbExtensions = Map("map" -> "gmaps"),
+    combinedExtensions = Map("pizza" -> "pizzabutton")
   )
+
+  val emptyExtensionClusterRequest = UserJupyterExtensionConfig()
+
   val jupyterLabExtensionClusterRequest = UserJupyterExtensionConfig(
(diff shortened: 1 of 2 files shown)


automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoTestUtils.scala
```diff
@@ -76,9 +76,13 @@ trait LeonardoTestUtils
   val getAfterCreatePatience = PatienceConfig(timeout = scaled(Span(5, Minutes)), interval = scaled(Span(2, Seconds)))
 
   val multiExtensionClusterRequest = UserJupyterExtensionConfig(
+    //TODO: Re-enable the below two lines when nbconvert is updated to 6.0.1
```
Qi77Qi, line 79: can we use some other random extensions instead of disabling testing? ...I disabled the whole test in https://github.com/DataBiosphere/leonardo/pull/1604 temporarily btw just so that PR isn't blocked

rtitle: Yeah these extensions were picked randomly, we could try some other ones if these are problematic. Ideas:

* [toc](https://jupyter-contrib-nbextensions.readthedocs.io/en/latest/nbextensions/toc2/README.html)
* [codefolding](https://jupyter-contrib-nbextensions.readthedocs.io/en/latest/nbextensions/codefolding/readme.html)
* [igv](https://github.com/igvteam/igv-jupyter#:~:text=igv%2Djupyter%20is%20an%20extension,its%20API%20from%20the%20notebook.)

PR #1607, opened 2020-09-15
IA-2202: Changing Extensions to remove failure

xuscrbw:
The failure was due to the gmaps and pizzabuttons extensions not being compatible with the version of nbconvert.

Instead of bumping the version, we are opting to change the extension to one that is being used. Bumping the version has a possibility of breaking something else, so we are going to try this approach first
---

In all cases:

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoTestUtils.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoTestUtils.scala
@@ -76,8 +76,8 @@ trait LeonardoTestUtils
   val getAfterCreatePatience = PatienceConfig(timeout = scaled(Span(5, Minutes)), interval = scaled(Span(2, Seconds)))
 
   val multiExtensionClusterRequest = UserJupyterExtensionConfig(
-    nbExtensions = Map("map" -> "gmaps"),
-    combinedExtensions = Map("pizza" -> "pizzabutton")
+    nbExtensions = Map("saturn-iframe-extension" -> "https://app.terra.bio/jupyter-iframe-extension.js")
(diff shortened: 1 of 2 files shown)


xuscrbw:
jenkins retest

xuscrbw:
jenkins retest

Qi77Qi (approved):
👍

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoTestUtils.scala
```diff
@@ -76,8 +76,7 @@ trait LeonardoTestUtils
   val getAfterCreatePatience = PatienceConfig(timeout = scaled(Span(5, Minutes)), interval = scaled(Span(2, Seconds)))
 
   val multiExtensionClusterRequest = UserJupyterExtensionConfig(
-    nbExtensions = Map("map" -> "gmaps"),
-    combinedExtensions = Map("pizza" -> "pizzabutton")
+    nbExtensions = Map("saturn-iframe-extension" -> "https://app.terra.bio/jupyter-iframe-extension.js")
```
Qi77Qi, line 79: shall we still add another `combinedExtensions`?

xuscrbw: What is the difference between combined Extensions vs regular notebook? Could we use a notebook extension as a combined one?

Qi77Qi: > Could we use a notebook extension as a combined one?

I would give it a try..I don't think there's anything special about `pizzabutton`

PR #1610, opened 2020-09-16
[IA-2168] Persist helm chart names and versions

kyuksel:
This is the last PR to close out [IA-2168](https://broadworkbench.atlassian.net/browse/IA-2168).

- Adding
  - `chart` column to `APP` table
  - `ingressChart` column to `KUBERNETES_CLUSTER` table
- Front Leo will be populating those column values, which will be a composite string of chart name and version, at app and cluster creation time, respectively.


project/Dependencies.scala
```diff
@@ -141,6 +141,8 @@ object Dependencies {
     workbenchGoogle2Test,
     workbenchOpenTelemetry,
     workbenchOpenTelemetryTest,
+    helmScalaSdk,
```
rtitle, line 144: Should we remove these from `rootDependencies` since they're now in `coreDependencies`?

rtitle (approved):
Looks good to me!

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20200915_add_chart_to_app_and_cluster_tables.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="leonardo" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+    <changeSet logicalFilePath="leonardo" author="kyuksel" id="add_chart_to_app_table">
+        <addColumn tableName="APP">
+            <column name="chart" type="VARCHAR(254)" defaultValue="dummyChartName-x.y.z">
```
Qi77Qi: shall we run a `insert` cmd in current database, and remove the `defaultValue` here? just so there's no way going forward, a record will be populated with the dummy value

rtitle: That sounds good to me, could do something like:
1. add nullable collumn
2. INSERT statement to set the current values
3. make column not null

project/Dependencies.scala
kyuksel: `rootDependencies` correspond to `http` and `coreDependencies` apply to both `http` and `automation`?

Qi77Qi: yea, it's carry over name when leo was previously non multi project...feel free to rename it to HTTP...you can probably remove https://github.com/DataBiosphere/leonardo/blob/develop/build.sbt#L8 as well

kyuksel:
@Qi77Qi @rtitle Thank you for your reviews! Changes to address your comments are [here](https://github.com/DataBiosphere/leonardo/pull/1610/files/7c45f77b3e012913217297803648e3fdd35cce1b..911ae0bdecca4e72c9bfe179b8f71a4c1057ab1f) if you could take another look.

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20200915_add_chart_to_app_and_cluster_tables.xml
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="leonardo" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+    <changeSet logicalFilePath="leonardo" author="kyuksel" id="add_chart_to_app_table">
+        <addColumn tableName="APP">
+            <column name="chart" type="VARCHAR(254)">
+                <constraints nullable="true"/>
+            </column>
+        </addColumn>
+
+        <sql>UPDATE APP SET chart = 'dummyChartName-x.y.z' WHERE chart IS NULL</sql>
```
rtitle, line 10: Maybe we should set existing records to the actual value, e.g. `galaxy/galaxykubeman-0.5.3`?

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20200915_add_chart_to_app_and_cluster_tables.xml
```diff
@@ -0,0 +1,26 @@
...
+    </changeSet>
+
+    <changeSet logicalFilePath="leonardo" author="kyuksel" id="add_ingressChart_to_kubernetes_cluster_table">
+        <addColumn tableName="KUBERNETES_CLUSTER">
+            <column name="ingressChart" type="VARCHAR(254)">
+                <constraints nullable="true"/>
+            </column>
+        </addColumn>
+
+        <sql>UPDATE KUBERNETES_CLUSTER SET ingressChart = 'dummyChartName-x.y.z' WHERE ingressChart IS NULL</sql>
```
rtitle, line 22: Same here, maybe `nginx/nginx-1.41.3`

project/Version.scala
```diff
@@ -1,25 +1,13 @@
...
 object Version {
-  val baseVersion = "0.1"
-
-  def getVersionString = {
-
+  private def getVersionString =
     // either specify git model hash as an env var or derive it
     // if building from the broadinstitute/scala-baseimage docker image use env var
     // (scala-baseimage doesn't have git in it)
-    val lastCommit = sys.env.getOrElse("GIT_HASH", sys.process.Process("git rev-parse --short HEAD")!! ).trim()
```
rtitle, line 13: Just so I understand: does this logic exist somewhere else or are we changing the format of the Leo version string?


Current version strings look like this: `0.1-971dc87-SNAPSHOT`

(see https://notebooks.firecloud.org/version)

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20200915_add_chart_to_app_and_cluster_tables.xml
kyuksel: I was talking to Qi about this too. Thought using the dummy value may have two advantages:

1. I'm guessing we have some records that were created with older versions of `galaxykubeman`, and possibly `nginx` so the latest value may be wrong for them.
2. It could be good to know what records were created before the columns were added for debug purposes.

What do you think?

project/Version.scala
kyuksel:
```
sys.env.getOrElse("GIT_HASH", sys.process.Process("git rev-parse --short HEAD")!! ).trim()
```
is still in place. We're just removing `0.1` and `SNAPSHOT` from the version string because they're constant and don't seem to add value. The git hash is already unique. 

Or did I misunderstand your question?

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20200915_add_chart_to_app_and_cluster_tables.xml
rtitle: Sure, that sounds reasonable

project/Version.scala
rtitle: OK thanks -- so the format of the version string is changing, from `0.1-<hash>-SNAPSHOT` to just `<hash>`. I agree the `0.1` and `SNAPSHOT` don't add any value. Just thinking if anything depends on the current version string format, but I don't think anything does.

rtitle (approved):

Qi77Qi (commented):
lgtm 👍

project/Version.scala
kyuksel: Good thinking 👍  The only place I could think of was `hotswap` script but that seems to work.

PR #1611, opened 2020-09-16
[no ticket] Document setCookie endpoints in swagger

rtitle:
Marks the `/proxy/{googleProject}/{runtimeName}/setCookie` route as deprecated and documents `/proxy/setCookie`. Also changed some terminology from `cluster` -> `runtime` in the proxy routes in Swagger.

The rationale for the change is that Galaxy also requires the setCookie endpoint but we didn't want to overload the original (runtime-based) one for apps. In fact the old setCookie endpoint does not even use the parameters passed to it, so we just made it take no params, e.g. `/proxy/setCookie`.

Note the old version will continue to work -- this PR just updates the Swagger documentation.

Verified in https://editor.swagger.io/

---

In all cases:

### DIFF
--- a/http/src/main/resources/swagger/api-docs.yaml
+++ b/http/src/main/resources/swagger/api-docs.yaml
@@ -4,7 +4,7 @@ info:
   description: |
     Workbench notebooks service.
   # Follow semantic versioning https://semver.org/
-  version: "0.1.0"
+  version: "0.2.0"
   license:
     name: BSD
     url: http://opensource.org/licenses/BSD-3-Clause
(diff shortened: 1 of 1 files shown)


http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -1927,16 +1880,40 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/ErrorReport"
-  /proxy/invalidateToken:
+  "/proxy/setCookie":
```
Qi77Qi, line 1883: is this runtime specific or applicable to both runtime and galaxy?

Qi77Qi (approved):
think the answer to my question is both..but just wanted to confirm

http/src/main/resources/swagger/api-docs.yaml
rtitle: Yep, applicable to both runtime and galaxy

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -1927,16 +1880,40 @@ paths:
...
+    get:
+      summary: Sets a browser cookie needed to authorize connections to a Leonardo runtime
```
calbach, line 1885: Am I understanding the change correctly - that this endpoint is used to auth any runtime rather than just specific instance? Is there an implementation change associated with this - or was the project/namespace purely excess info to begin with?

calbach: nvm - it's called reading the PR description 🤦‍♂️

calbach (approved):
FWIW we are still using this one:

`/notebooks/{googleProject}/{clusterName}/setCookie`

PR #1614, opened 2020-09-17
[IA-2215] Decouple app name and helm release name

rtitle:
https://broadworkbench.atlassian.net/browse/IA-2215

Previously we used `appName` to generate both `namespaceName` and `releaseName`. This caused issues because `appName` (which is generated by the UI and "owned" by the Leo DB) was subject to hidden constraints due to how it was used by the Galaxy chart. This PR decouples `appName` from k8s/Galaxy internal names for release and namespace.

Unit tests pass, also manually tested app creation on a fiab.

---

In all cases:


http/src/main/resources/reference.conf
```diff
@@ -241,11 +241,11 @@ gke {
...
   }
   galaxyApp {
-    releaseNameSuffix = "galaxy-rls"
+    releaseNameSuffix = "gxy-rls"
     chartName = "galaxy/galaxykubeman"
     chartVersion = "0.5.3"
-    namespaceNameSuffix = "galaxy-ns"
-    serviceAccountSuffix = "galaxy-ksa"
+    namespaceNameSuffix = "gxy-ns"
+    serviceAccountName = "gxy-ksa"
```
rtitle, line 251: KSAs are scoped to the namespace, so don't need to be unique within a cluster

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -677,8 +679,8 @@ class GKEInterpreter[F[_]: Parallel: ContextShift: Timer](
         List(
           raw"""configs.$k=$v""",
           raw"""extraEnv[$i].name=$k""",
-          raw"""extraEnv[$i].valueFrom.configMapKeyRef.name=${release.asString}=galaxykubeman-configs""",
```
rtitle, line 681: I noticed these typos while in this code

http/src/main/resources/reference.conf
```diff
@@ -241,11 +241,11 @@ gke {
     ]
   }
   galaxyApp {
-    releaseNameSuffix = "galaxy-rls"
+    releaseNameSuffix = "gxy-rls"
```
rtitle, line 247: See comment in `KubernetesServiceInterp` for more context. Theoretically release names _should_ only need to be unique within a namespace, but something in the Galaxy chart requires them to be unique within a cluster.

rtitle:
jenkins retest

rtitle:
jenkins retest

rtitle:
Here's an example cluster where I successfully created 3 Galaxys:
image

Qi77Qi (approved):

http/src/main/resources/reference.conf
kyuksel: Might be helpful to add ^^ comment above line 244 too.

kyuksel (approved):
Looks good overall 👍 

I was finding it useful to know the namespace for filtering on the GKE console during development when I had multiple releases on the same cluster. Would we now be querying the DB to find out the `uid`?

While at it, I wonder if we should get rid of the duplicate `nginx` in `nginx-*` resources too 🤔

http/src/main/resources/reference.conf
rtitle: Sure

rtitle:
> Would we now be querying the DB to find out the uid?

Yeah the namespace name and release name are both in the DB, so would need to query the DB to tie them to a particular App.

rtitle:
> While at it, I wonder if we should get rid of the duplicate nginx in nginx-* resources too

I think this is because we name both the namespace and release `nginx` so we end up with `nginx-nginx-*`. I don't see this as critical since it not really causing any problems.

PR #1615, opened 2020-09-17
[no ticket] Remove special k8s name validation in routes

rtitle:
Missed this in my last PR -- since we're not using `appName` as k8s names anymore, we don't need to do custom k8s name validation in `AppRoutes`. Change `AppRoutes` to use the same Leo name validation logic as runtime names.

---

In all cases:


Qi77Qi (approved):

kyuksel (approved):
👍  Sorry that I didn't think of this while reviewing the preceding PR 😞

PR #1618, opened 2020-09-18
[IA-2222] Streamline cluster+nodepool creation

rtitle:
https://broadworkbench.atlassian.net/browse/IA-2222

This saves some time on app creation by passing in both the default nodepool and the user nodepool to the createCluster request, instead of creating the cluster and user nodepool serially.

I also refactored some of the models for the `createApp` pubsub message. Previously we determined behavior based on a few optional parameters in the message. I found this confusing, and not all permutations were actually valid states for the system to be in. So I added a sum type:
```
sealed trait ClusterNodepoolAction extends Product with Serializable {
  def actionType: ClusterNodepoolActionType
}
object ClusterNodepoolAction {
  final case class CreateClusterAndNodepool(clusterId: KubernetesClusterLeoId,
                                            defaultNodepoolId: NodepoolLeoId,
                                            nodepoolId: NodepoolLeoId)
      extends ClusterNodepoolAction {
    val actionType: ClusterNodepoolActionType = ClusterNodepoolActionType.CreateClusterAndNodepool

  }
  final case class CreateNodepool(nodepoolId: NodepoolLeoId) extends ClusterNodepoolAction {
    val actionType: ClusterNodepoolActionType = ClusterNodepoolActionType.CreateNodepool
  }
}
```
I think this is simpler as the cluster/nodepool action is now explicitly defined in the pubsub message.

---

In all cases:

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
@@ -389,9 +389,6 @@ final case class KubernetesServiceKindName(value: String) extends AnyVal
 final case class KubernetesRuntimeConfig(numNodes: NumNodes, machineType: MachineTypeName, autoscalingEnabled: Boolean)
 final case class NumNodepools(value: Int) extends AnyVal
 
-//used in pubsub messaging to indicate the cluster and dummy nodepool to be created
-final case class CreateCluster(clusterId: KubernetesClusterLeoId, defaultNodepoolId: NodepoolLeoId)
-
 final case class NodepoolNotFoundException(nodepoolLeoId: NodepoolLeoId) extends Exception {
   override def getMessage: String = s"nodepool with id ${nodepoolLeoId} not found"
 }
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
@@ -630,7 +630,8 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift: Parallel](
       // we can nack the message on errors. Monitoring all creations will be asynchronous,
       // and (3) and (4) will always be asynchronous.
 
-      // Create the cluster or nodepool synchronously if necessary
+      // Create the cluster or nodepool synchronously if necessary.
+      // The monitor operation is returned so it can be run asynchronously.
       monitorClusterOrNodepool <- msg.clusterNodepoolAction match {
         case Some(ClusterNodepoolAction.CreateClusterAndNodepool(clusterId, defaultNodepoolId, nodepoolId)) =>
           for {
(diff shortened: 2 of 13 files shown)


core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
```diff
@@ -386,9 +386,6 @@ final case class KubernetesServiceKindName(value: String) extends AnyVal
 final case class KubernetesRuntimeConfig(numNodes: NumNodes, machineType: MachineTypeName, autoscalingEnabled: Boolean)
 final case class NumNodepools(value: Int) extends AnyVal
 
-//used in pubsub messaging to indicate the cluster and dummy nodepool to be created
-final case class CreateCluster(clusterId: KubernetesClusterLeoId, defaultNodepoolId: NodepoolLeoId)
```
rtitle, line 393: I found there wasn't much use for this product type now that we have `ClusterNodepoolAction`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -630,37 +630,44 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift: Parallel](
...
-                false,
-                msg.nodepoolId,
-                None
+      // Create the cluster or nodepool synchronously if necessary
+      monitorClusterOrNodepool <- msg.clusterNodepoolAction match {
+        case Some(ClusterNodepoolAction.CreateClusterAndNodepool(clusterId, defaultNodepoolId, nodepoolId)) =>
+          for {
+            createClusterResult <- gkeInterp
+              .createCluster(
+                CreateClusterParams(clusterId, msg.project, List(defaultNodepoolId, nodepoolId), false)
```
rtitle, line 640: Note we're now passing 2 nodepools to the cluster creation request, which is explicitly specified in the message. This saves a few minutes off total app creation time.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -674,49 +681,8 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift: Parallel](
...
-              PubsubKubernetesError(
-                AppError(e.getMessage, ctx.now, ErrorAction.CreateGalaxyApp, ErrorSource.Cluster, None),
-                Some(msg.appId),
-                false,
-                msg.cluster.map(_.defaultNodepoolId),
-                msg.cluster.map(_.clusterId)
-              )
-          }
-
-        _ <- ((msg.nodepoolId, createNodepoolResultOpt) match {
```
rtitle, line 697: No need anymore for this decisioning based on multiple Options

rtitle:
jenkins retest, fiab-start issues

rtitle:
I tested all 3 cases on a fiab:
1. Some(create cluster)
2. Some(create nodepool)
3. None

All seems to work.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
Qi77Qi: just for my own understanding, are both nodepool already created at this point? or we're passing the nodepool names and `gkeInterp.createCluster` will create them concurrently if they don't exist?

Qi77Qi (approved):

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
rtitle: the nodepools don't exist yet -- we are passing them to GKE, and they will be created as part of cluster creation.

PR #1619, opened 2020-09-21
[IA-2197] Delete cluster on cluster error

jdcanas:
This PR allows users to attempt to create an app again if a create app requests results in a failed cluster creation attempt.

Previously, if the cluster errored out, users would not be able to create apps without manual intervention.

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
@@ -639,15 +639,24 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift: Parallel](
               .createCluster(
                 CreateClusterParams(clusterId, msg.project, List(defaultNodepoolId, nodepoolId), false)
               )
+              .onError {
+                case _ =>
+                  for {
+                    _ <- gkeInterp.deleteAndPollCluster(DeleteClusterParams(clusterId, msg.project))
+                  } yield ()
+              }
               .adaptError {
-                case e =>
+                case e => {
                   PubsubKubernetesError(
                     AppError(e.getMessage, ctx.now, ErrorAction.CreateGalaxyApp, ErrorSource.Cluster, None),
                     Some(msg.appId),
                     false,
+                    // We leave cluster id and default nodepool id as none here because we want the status to stay as DELETED and not transition to ERROR.
+                    // The app will have the error so the user can see it, delete their app, and try again
                     None,
-                    Some(clusterId)
+                    None
                   )
+                }
               }
             monitorOp = gkeInterp.pollCluster(PollClusterParams(clusterId, msg.project, false, createClusterResult))
           } yield monitorOp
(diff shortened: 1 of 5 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -633,17 +633,27 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift: Parallel](
...
-        .adaptError {
-          case e =>
-            PubsubKubernetesError(
-              AppError(e.getMessage, ctx.now, ErrorAction.CreateGalaxyApp, ErrorSource.Cluster, None),
-              Some(msg.appId),
-              false,
-              msg.cluster.map(_.defaultNodepoolId),
-              msg.cluster.map(_.clusterId)
-            )
+          gkeInterp
```
rtitle: Sorry might conflict a little with https://github.com/DataBiosphere/leonardo/pull/1618

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -633,17 +633,27 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift: Parallel](
...
+            .createCluster(CreateClusterParams(c.clusterId, msg.project, List(c.defaultNodepoolId), false))
+            .onError {
+              case _ =>
+                for {
+                  _ <- gkeInterp.deleteAndPollCluster(DeleteClusterParams(c.clusterId, msg.project))
```
rtitle: Does this do the deletion + polling synchronous to the pubsub request?

(I think that's fine, just curious)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -379,8 +379,33 @@ class GKEInterpreter[F[_]: Parallel: ContextShift: Timer](
     } yield ()
 
   override def deleteAndPollCluster(params: DeleteClusterParams)(implicit ev: ApplicativeAsk[F, AppContext]): F[Unit] =
-    // TODO not yet implemented
-    F.unit
+    for {
+      ctx <- ev.ask
+      dbClusterOpt <- kubernetesClusterQuery.getMinimalClusterById(params.clusterId).transaction
+      dbCluster <- F.fromOption(
```
rtitle, line 387: Should we mark the cluster as `Deleting` prior to the GKE delete+poll?

jdcanas: hmm.. we probably should so it can get picked up by MonitorAtBoot or the like if we don't poll to completion

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
jdcanas: Yes. it seemed a bit heavy to make an async task when this portion is sync already. It possible that this could trigger message resending from google, but It's set to 5 mins and this is usually sufficient to delete the cluster.  Thinking about this made me realize we should definitely `handleErrorWith` on this line so errors in cleaning up the cluster don't get bubbled up...

rtitle (approved):

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
rtitle: I made a ticket for integrating `MonitorAtBoot` :https://broadworkbench.atlassian.net/browse/IA-2224

rtitle:
Latest changes LGTM

PR #1621, opened 2020-09-22
IA-2227 drop_on_update_pd_table

Qi77Qi:
Currently `createdDate` in pd table has the following definition
```
`createdDate` timestamp(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6) ON UPDATE CURRENT_TIMESTAMP(6)
```

This makes `createdDate` updated every time the record is updated, which is not what we want.

---

In all cases:

### DIFF
--- a/http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changelog.xml
+++ b/http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changelog.xml
@@ -64,4 +64,5 @@
     <include file="changesets/20200817_kubernetes_cluster_api_server.xml" relativeToChangelogFile="true" />
     <include file="changesets/20200915_add_chart_to_app_and_cluster_tables.xml" relativeToChangelogFile="true" />
     <include file="changesets/20200917_add_release_to_app_table.xml" relativeToChangelogFile="true" />
+    <include file="changesets/20200921_drop_on_update_pd_table.xml" relativeToChangelogFile="true" />
 </databaseChangeLog>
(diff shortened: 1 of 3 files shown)


rtitle:
Good catch. Definitely not intended - I wonder how that happened? I wonder if the `on update` was intended for `dateAccessed`?
```
MySQL [leonardo]> DESCRIBE PERSISTENT_DISK;
+----------------+----------------------+------+-----+----------------------------+--------------------------------+
| Field          | Type                 | Null | Key | Default                    | Extra                          |
+----------------+----------------------+------+-----+----------------------------+--------------------------------+
| id             | bigint(20)           | NO   | PRI | NULL                       | auto_increment                 |
| googleProject  | varchar(255)         | NO   | MUL | NULL                       |                                |
| zone           | varchar(255)         | NO   |     | NULL                       |                                |
| name           | varchar(255)         | NO   |     | NULL                       |                                |
| googleId       | varchar(255)         | YES  |     | NULL                       |                                |
| samResourceId  | varchar(255)         | NO   |     | NULL                       |                                |
| status         | varchar(255)         | NO   | MUL | NULL                       |                                |
| creator        | varchar(255)         | NO   |     | NULL                       |                                |
| createdDate    | timestamp(6)         | NO   |     | CURRENT_TIMESTAMP(6)       | on update CURRENT_TIMESTAMP(6) |
| destroyedDate  | timestamp(6)         | NO   |     | 0000-00-00 00:00:00.000000 |                                |
| dateAccessed   | timestamp(6)         | NO   |     | 0000-00-00 00:00:00.000000 |                                |
| sizeGb         | int(11)              | NO   |     | NULL                       |                                |
| type           | varchar(255)         | NO   |     | NULL                       |                                |
| blockSizeBytes | int(11)              | NO   |     | NULL                       |                                |
| serviceAccount | varchar(255)         | NO   |     | NULL                       |                                |
| formattedBy    | enum('GALAXY','GCE') | YES  |     | NULL                       |                                |
+----------------+----------------------+------+-----+----------------------------+--------------------------------+
16 rows in set (0.048 sec)
```

I seem to recall that `ALTER TABLE` can't be used with liquibase -- is there a liquibase way to do this instead? I could be wrong though

Qi77Qi:
@rtitle yea I'm not sure how this happened....I thought about `dateAccessed`, but I think we shouldn't do that either because autopause rely on `dateAccessed` but autopause only expects this field to be updated when the runtime is accessed through proxy.

> I seem to recall that ALTER TABLE can't be used with liquibase 

hmm, I'll see what I can find

Qi77Qi:
seems `alter` does work...not sure why unit tests fail yet

rtitle:
Not quite sure how to check this, but last time I think I saw ALTER TABLE fail on dev with a db permission error. Permissions may be different on local MySQL (for unit tests/fiab) vs CloudSQL.

Maybe could log in to the dev db with the same user/pass as liquibase uses, and try ALTER TABLE in a transaction then roll back. Just to be safe, since it’s annoying back out a broken migration that’s been merged

Qi77Qi:
from local testing, `alter` in liquibase worked just fine

Qi77Qi:
@rtitle I looked through all possbile liquibase APIs, haven't been able to find anything that's promising....

rtitle:
I tried this on dev with the `leonardo` user and it worked, so maybe I was imagining things. I'm going to merge this and see how it goes.

rtitle (approved):

rtitle:
This seemed to work on dev

PR #1627, opened 2020-09-24
IA-2236 add ability to unset a Leo token cookie

rtitle:
https://broadworkbench.atlassian.net/browse/IA-2236

Changes the following 2 routes:
* `invalidateToken` will now unset the LeoToken cookie (in addition to invalidating caches as it did before)
* `setCookie` will now unset the LeoToken cookie if no auth information is passed to the request. Previously it would fail with a 401.

Either of the above routes can be called upon UI signout to prevent access to proxied apps running in different tabs.

cc @breilly2 @panentheos 

---

In all cases:

### DIFF
--- a/http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/api/ProxyRoutesSpec.scala
+++ b/http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/api/ProxyRoutesSpec.scala
@@ -246,6 +246,29 @@ class ProxyRoutesSpec
     result.futureValue shouldBe "Hello Leonardo!"
   }
 
+  it should "create terminal trying to access a terminal not created yet" in {
+    val jupyterDAO = mock[JupyterDAO[IO]]
+    when {
+      jupyterDAO.terminalExists(GoogleProject(googleProject), RuntimeName(clusterName), TerminalName("1"))
+    } thenReturn IO.pure(false)
+
+    when {
+      jupyterDAO.createTerminal(GoogleProject(googleProject), RuntimeName(clusterName))
+    } thenReturn IO.unit
+
+    val httpRoutes = createHttpRoute(jupyterDAO)
+
+    Get(s"/proxy/$googleProject/$clusterName/jupyter/terminals/1")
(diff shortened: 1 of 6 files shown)


rtitle:
jenkins retest

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/api/ProxyRoutesSpec.scala
```diff
@@ -246,6 +246,29 @@ class ProxyRoutesSpec
     result.futureValue shouldBe "Hello Leonardo!"
   }
 
+  it should "create terminal trying to access a terminal not created yet" in {
```
kyuksel: Should this be something like

```suggestion
  it should "create the terminal when trying to access a terminal that was not created yet" in {
```

kyuksel (commented):
The changes look good to me (with one minor comment inline). I just have a couple of high-level questions about the part around

> setCookie will now unset the LeoToken cookie if no auth information is passed to the request. Previously it would fail with a 401.

1. Why is ^^ necessary given `invalidateToken` can now be used for unsetting the cookie?
2. Is there any opening for a malicious use where someone may call `setCookie` without auth info on purpose to unset someone's cookie? Probably not a concern but wanted to be sure 😊

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/api/ProxyRoutesSpec.scala
rtitle: I just moved this test because it was in the wrong place.. but yeah, that name sounds better. Will fix.

rtitle:
1. True they are a bit redundant. The `setCookie` behavior was requested by the Terra UI team, but given that we also have `invaldiateToken` I thought I would just do both. 

2. Good question! I don't _think_ it's an issue because the worst thing one could do is delete their own cookie. I guess a malicious site could potentially unset the user's LeoToken cookie, but it could also potentially do other/worse things if signed into Google.

kyuksel (approved):

rtitle:
jenkins retest

PR #1628, opened 2020-09-29
[no ticket] Upgrade galaxykubeman to 0.7.1

rtitle:
This version has the ~startup optimizations~ and the Terra filepicker. I'm testing it in a fiab now.

---

In all cases:


Dockerfile
```diff
@@ -32,7 +32,7 @@ RUN curl -fsSL -o get_helm.sh https://raw.githubusercontent.com/helm/helm/master
 
 # Add the repos containing nginx and galaxy charts
 RUN helm repo add stable https://kubernetes-charts.storage.googleapis.com/ && \
-    helm repo add galaxy https://raw.githubusercontent.com/cloudve/helm-charts/anvil/ && \
+    helm repo add galaxy https://github.com/CloudVE/helm-charts/blob/sep29demo && \
```
rtitle: This will probably change back to the `anvil` branch -- Alex just pushed to `sep29demo` temporarily for testing + the anvil demo today @ 4.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -772,6 +772,7 @@ class GKEInterpreter[F[_]: Parallel: ContextShift: Timer](
       raw"""galaxy.ingress.hosts[0]=${k8sProxyHost}""",
       raw"""galaxy.configs.galaxy\.yml.galaxy.single_user=${userEmail.value}""",
       raw"""galaxy.configs.galaxy\.yml.galaxy.admin_users=${userEmail.value}""",
+      raw"""galaxy.rbac.serviceAccount=${ksa.value}""",
       raw"""rbac.serviceAccount=${ksa.value}""",
```
kyuksel, line 777: curious why there are two `serviceAccount` keys

kyuksel (approved):
👍

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
rtitle: Good question -- it was in the latest `values` that Alex sent me, but I'm not sure the purpose of `galaxy.rbac.*`. I just pinged Alex for clarification.

rtitle: Alexandru Mahmoud  10:07 AM
the enabled should be false from the default, but yes setting both rbac.serviceAccount and galaxy.rbac.serviceAccount  is needed
10:07
(P.S. it's disabled so that the Galaxy chart doesn't deploy a new Role/RoleBinding, but the KSA name will be used by the pods) (edited)

PR #1629, opened 2020-09-30
Update aou image refs to latest (1.0.11)

calbach:
For a follow-up, thoughts on supporting the "terra_jupyter_aou_old" pattern for GCE in addition to Dataproc? We're close to switching over to GCE, and this would be a nice transitional pattern for us.


rtitle:
I think we can use this PR and close https://github.com/DataBiosphere/leonardo/pull/1630.

Sounds reasonable to use the `terra_jupyter_aou_old` pattern for the GCE script. I can make a ticket.

EDIT: https://broadworkbench.atlassian.net/browse/IA-2247

rtitle (approved):
:shipit: if tests pass

calbach:
Thanks

PR #1635, opened 2020-10-01
IA-2248: Add Galaxy workspace configs + automation test stability fixes

rtitle:
https://broadworkbench.atlassian.net/browse/IA-2248

The actual change for IA-2248 is commented out, but I also added some auto test fixes which I'd like to merge. Comments inline.

---

In all cases:


automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/BatchNodepoolCreationSpec.scala
```diff
@@ -110,12 +110,13 @@ class BatchNodepoolCreationSpec
           _ <- loggerIO.info("About to get app1")
 
           getApp1 = LeonardoApiClient.getApp(googleProject, appName1)
-          monitorApp1CreationResult <- testTimer.sleep(30 seconds) >> streamFUntilDone(getApp1, 60, 10 seconds)(
+          monitorApp1CreationResult <- testTimer.sleep(30 seconds) >> streamFUntilDone(getApp1, 90, 10 seconds)(
```
rtitle: Giving apps 20 minutes to start up

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoApiClient.scala
```diff
@@ -223,33 +211,28 @@ object LeonardoApiClient {
...
           uri = rootUri
             .withPath(s"/api/google/v1/runtimes/${googleProject.value}/${runtimeName.asString}")
             .withQueryParam("deleteDisk", deleteDisk)
         )
-      )
-      .flatMap { success =>
-        if (success)
-          IO.unit
-        else
-          IO.raiseError(new RuntimeException(s"Fail to delete runtime ${googleProject.value}/${runtimeName.asString}"))
```
rtitle, line 251: I was annoyed all these http4s exceptions don't actually log the error response status code/body. So I improved the error handling and made it consistent.

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoSuite.scala
```diff
@@ -149,10 +149,11 @@ trait GPAllocBeforeAndAfterAll extends GPAllocUtils with BeforeAndAfterAll {
           case Right(_) =>
             IO(logger.info(s"Created initial runtime ${project.value} / ${initalRuntimeName.asString}"))
           case Left(err) =>
-            IO(logger.warn(
-                 s"Failed to create initial runtime ${project.value} / ${initalRuntimeName.asString} with error"
-               ),
-               err)
+            IO(
```
rtitle, line 153: Before this was `IO(logger.warn(msg), err)` of type `IO[(Unit, Throwable)]` -- not what we want!

It should actually be passing the error to the logger, e.g. `IO(logger.warn(msg, err))` of type `IO[Unit]`.

I wish the Scala compiler caught this somehow 🤦

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -777,6 +779,9 @@ class GKEInterpreter[F[_]: Parallel: ContextShift: Timer](
       raw"""galaxy.ingress.hosts[0]=${k8sProxyHost}""",
       raw"""galaxy.configs.galaxy\.yml.galaxy.single_user=${userEmail.value}""",
       raw"""galaxy.configs.galaxy\.yml.galaxy.admin_users=${userEmail.value}""",
+      // TODO add this when Galaxy version is upgraded
+      //raw"""galaxy.configs.file_sources_conf\.yml[0].workspace=${workspaceName}""",
```
rtitle: I'm coordinating with Alex to get the latest Galaxy packaged, at which point we can uncomment this.

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/BatchNodepoolCreationSpec.scala
```diff
@@ -110,12 +110,13 @@ class BatchNodepoolCreationSpec
...
             testTimer,
             appDoneCheckable
           ).compile.lastOrError
 
           _ <- loggerIO.info(s"app1 monitor result: ${monitorApp1CreationResult}")
+          _ = monitorApp1CreationResult.status shouldBe AppStatus.Running
```
rtitle, line 120: Important note about `streamFUntilDone` -- it doesn't fail if the condition is not met after the timeout! So it's totally possible for the app to still be `Provisioning` after `streamFUntilDone` terminates. So we should always assert the expected state at the end.

jdcanas: ahh, yeah I printed it to verify in testing but good to assert, thanks for fixing this I was about to make a PR

jdcanas (approved):

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoApiClient.scala
```diff
@@ -136,10 +133,7 @@ object LeonardoApiClient {
           headers = Headers.of(authHeader),
           uri = rootUri.withPath(s"/api/google/v1/runtimes/${googleProject.value}/${runtimeName.asString}/start")
         )
-      )(resp =>
-        resp.bodyText.compile.string
-          .flatMap(body => IO.raiseError(RestError(resp.status, body)))
-      )
+      )(onError(s"Failed to start runtime ${googleProject.value}/${runtimeName.asString}"))
```
jdcanas, line 136: I liked raising the error here, otherwise its hard to tell why the call actually failed (namely when I was debugging those SAM errors, the error message was hidden with this code)

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/BatchNodepoolCreationSpec.scala
jdcanas: Yeah, I was printing it to verify during testing but good to assert it as well. Thanks for fixing this, I was about to PR

xuscrbw: Thanks for this. I didn't quite understand why I was hitting errors when it didn't fully transition on my PR

xuscrbw: A quick question, I've noticed that we can't do the same sort of status assertion when we use streamFUntilDone with listApps. Is there anything we should do here to make sure the app is finished deleting properly?

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoApiClient.scala
```diff
@@ -223,33 +211,28 @@ object LeonardoApiClient {
...
       )
-    )
+    )(onError(s"Failed to get runtime ${googleProject.value}/${runtimeName.asString}"))
 
   def deleteRuntime(googleProject: GoogleProject,
                     runtimeName: RuntimeName,
                     deleteDisk: Boolean = true)(implicit client: Client[IO], authHeader: Authorization): IO[Unit] =
     client
-      .successful(
+      .expectOr[String](
```
rtitle: Ugh my http4s skills are not great, this is failing with "invalid json"

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/BatchNodepoolCreationSpec.scala
rtitle: `streamFUntilDone` is pretty generic -- it just takes an operation and a predicate and polls the operation until the predicate returns true (or a timeout occurs). The important behavior to be aware of is the timeout doesn't cause an exception to occur -- it just continues on. So if you _don't_ want to continue if the predicate didn't become true, you have to add an extra assertion.

So I think `streamFUntilDone` should be able to wait for deletion with listApps -- you just need to define the predicates correctly.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -777,6 +779,11 @@ class GKEInterpreter[F[_]: Parallel: ContextShift: Timer](
       raw"""galaxy.ingress.hosts[0]=${k8sProxyHost}""",
       raw"""galaxy.configs.galaxy\.yml.galaxy.single_user=${userEmail.value}""",
       raw"""galaxy.configs.galaxy\.yml.galaxy.admin_users=${userEmail.value}""",
+      raw"""galaxy.configs.file_sources_conf\.yml[0].type=anvil""",
```
rtitle, line 782: This seems to work with the current galaxy version, Alex confirmed the values

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
```diff
@@ -232,10 +232,10 @@ final case class DefaultKubernetesLabels(googleProject: GoogleProject,
                                          serviceAccount: WorkbenchEmail) {
   val toMap: LabelMap =
     Map(
-      "appName" -> appName.toString,
-      "googleProject" -> googleProject.toString,
-      "creator" -> creator.toString,
-      "clusterServiceAccount" -> serviceAccount.toString
+      "appName" -> appName.value,
```
rtitle, line 235: Noticed this in perf testing -- we were adding labels like `"AppName(name)"`

PR #1636, opened 2020-10-05
[IA-2171] Galaxy disk attachment

gpcarr:
Tested by creating an app using the `createApp` endpoint and verifying that the correct PV/PVC (both for the NFS disk and the postgres disk) pairs get created and that Galaxy successfully transitions to `RUNNING` 

We're defaulting the NFS disk to 500Gb because currently, Galaxy requests 300Gb in storage from the NFS provisioner and CVMFS cache requests 10Gb so the size of the NFS disk needs to be at least 310Gb. 500Gb was a suggestion from the Galaxy team -  if we want to lower this then we can but we just need to also lower the amount of storage Galaxy requests. 

---

In all cases:

### DIFF
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -261,6 +261,14 @@ gke {
       }
     ]
   }
+  galaxyDisk {
+    nfsPersistenceName = "nfs-disk"
+    postgresPersistenceName = "postgres-disk"
+    # TODO: remove post-alpha once persistence is in place for Galaxy
+    postgresDiskNameSuffix = "gxy-postres-disk"
+    postgresDiskSizeGB = 10
+    postgresDiskBlockSize = 4096
+  }
 }
 
 image {
@@ -540,6 +548,7 @@ persistentDisk {
   defaultDiskType = "pd-standard"
   defaultBlockSizeBytes = 4096
   zone = "us-central1-a"
+  defaultGalaxyNFSDiskSizeGB = 500
 }
 
 clusterToolMonitor {
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
@@ -15,6 +15,7 @@ import fs2.Stream
 import fs2.concurrent.InspectableQueue
 import org.broadinstitute.dsde.workbench.errorReporting.ErrorReporting
 import org.broadinstitute.dsde.workbench.errorReporting.ReportWorthySyntax._
+import org.broadinstitute.dsde.workbench.google2.KubernetesSerializableName.NamespaceName
 import org.broadinstitute.dsde.workbench.google2.{
   ComputePollOperation,
   DiskName,
@@ -22,16 +23,19 @@ import org.broadinstitute.dsde.workbench.google2.{
   GoogleDiskService,
   GoogleSubscriber,
   MachineTypeName,
-  OperationName
+  OperationName,
+  ZoneName
 }
+import org.broadinstitute.dsde.workbench.leonardo.AppType.Galaxy
 import org.broadinstitute.dsde.workbench.leonardo.AsyncTaskProcessor.Task
 import org.broadinstitute.dsde.workbench.leonardo.db._
+import org.broadinstitute.dsde.workbench.leonardo.http.service.AppNotFoundException
 import org.broadinstitute.dsde.workbench.leonardo.http.{cloudServiceSyntax, _}
 import org.broadinstitute.dsde.workbench.leonardo.model.{LeoAuthProvider, LeoException}
 import org.broadinstitute.dsde.workbench.leonardo.monitor.LeoPubsubMessage._
 import org.broadinstitute.dsde.workbench.leonardo.monitor.PubsubHandleMessageError._
 import org.broadinstitute.dsde.workbench.leonardo.util._
-import org.broadinstitute.dsde.workbench.model.google.{GcsObjectName, GcsPath}
+import org.broadinstitute.dsde.workbench.model.google.{GcsObjectName, GcsPath, GoogleProject}
 import org.broadinstitute.dsde.workbench.model.{ErrorReport, TraceId, WorkbenchException}
 
 import scala.concurrent.ExecutionContext
@@ -511,7 +515,7 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift: Parallel](
         asyncTasks.enqueue1(
           Task(ctx.traceId,
                task,
-               Some(logError(s"${ctx.traceId.asString} | ${msg.diskId.value}", "Creatiing Disk")),
+               Some(logError(s"${ctx.traceId.asString} | ${msg.diskId.value}", "Creating Disk")),
                ctx.now)
         )
       }
@@ -525,6 +529,51 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift: Parallel](
             s"Failed to create disk ${msg.name.value} in Google project ${msg.googleProject.value}"
           )
           _ <- persistentDiskQuery.updateStatus(msg.diskId, DiskStatus.Failed, ctx.now).transaction[F]
+        } yield ()
+    }
+  }
+
+  private[monitor] def createGalaxyPostgresDiskOnlyInGoogle(project: GoogleProject,
+                                                            zone: ZoneName,
+                                                            appName: AppName,
+                                                            namespaceName: NamespaceName)(
+    implicit ev: ApplicativeAsk[F, AppContext]
+  ): F[Unit] = {
+    // TODO: remove post-alpha release of Galaxy. For pre-alpha we are only creating the postgress disk in Google since we are not supporting persistence
+    // see: https://broadworkbench.atlassian.net/wiki/spaces/IA/pages/859406337/2020-10-02+Galaxy+disk+attachment+pre+post+alpha+release
+    val create = for {
+      ctx <- ev.ask
+      _ <- logger.info(s"Beginning postgres disk creation for app ${appName.value} | trace id: ${ctx.traceId}")
+      operation <- googleDiskService.createDisk(project,
+                                                zone,
+                                                gkeInterp.createGalaxyPostgresDisk(project, zone, namespaceName))
+      whenDone = logger.info(
+        s"Completed postgres disk creation for app ${appName.value} in project ${project.value} | trace id: ${ctx.traceId}"
+      )
+      _ <- computePollOperation.pollZoneOperation(
+        project,
+        zone,
+        OperationName(operation.getName),
+        config.persistentDiskMonitorConfig.create.interval,
+        config.persistentDiskMonitorConfig.create.maxAttempts,
+        None
+      )(
+        whenDone,
+        F.raiseError(
+          new TimeoutException(
+            s"Failed to create Galaxy postgres disk in a timely manner. Project: ${project.value}, AppName: ${appName.value}"
+          )
+        ),
+        F.unit
+      )
+    } yield ()
+
+    create.onError {
+      case e =>
+        for {
+          _ <- logger.error(e)(
+            s"Failed to create Galaxy postgres disk in Google project ${project.value}, AppName: ${appName.value}"
+          )
         } yield ()
     }
   }
@@ -577,6 +626,43 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift: Parallel](
                ctx.now)
         )
       }
+    } yield ()
+
+  private[monitor] def deleteGalaxyPostgresDiskOnlyInGoogle(project: GoogleProject,
+                                                            zone: ZoneName,
+                                                            appName: AppName,
+                                                            namespaceName: NamespaceName)(
+    implicit ev: ApplicativeAsk[F, AppContext]
+  ): F[Unit] =
+    // TODO: remove post-alpha release of Galaxy. For pre-alpha we are only deleting the postgress disk in Google since we are not supporting persistence
+    // see: https://broadworkbench.atlassian.net/wiki/spaces/IA/pages/859406337/2020-10-02+Galaxy+disk+attachment+pre+post+alpha+release
+    for {
+      ctx <- ev.ask
+      _ <- logger.info(
+        s"Beginning postres disk deletion for app ${appName.value} in project ${project.value} | trace id: ${ctx.traceId}"
+      )
+      operation <- googleDiskService.deleteDisk(project, zone, gkeInterp.getGalaxyPostgresDiskName(namespaceName))
+      whenDone = logger.info(
+        s"Completed postgres disk deletion for app ${appName.value} in project ${project.value} | trace id: ${ctx.traceId}"
+      )
+      whenTimeout = F.raiseError[Unit](
+        new TimeoutException(
+          s"Failed to delete postres disk in app ${appName.value} in project ${project.value} in a timely manner"
+        )
+      )
+      whenInterrupted = F.unit
+      task = operation match {
+        case Some(op) =>
+          computePollOperation.pollZoneOperation(project,
+                                                 zone,
+                                                 OperationName(op.getName),
+                                                 config.persistentDiskMonitorConfig.delete.interval,
+                                                 config.persistentDiskMonitorConfig.create.maxAttempts,
+                                                 None)(whenDone, whenTimeout, whenInterrupted)
+        case None =>
+          whenDone
+      }
+      _ <- task
     } yield ()
 
   private[monitor] def handleUpdateDiskMessage(msg: UpdateDiskMessage)(
@@ -703,8 +789,25 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift: Parallel](
           )
       }
 
+      dbAppOpt <- KubernetesServiceDbQueries.getActiveFullAppByName(msg.project, msg.appName).transaction
+      dbApp <- F.fromOption(dbAppOpt, AppNotFoundException(msg.project, msg.appName, ctx.traceId))
+      app = dbApp.app
+      createSecondDisk = if (msg.appType == Galaxy) {
+        createGalaxyPostgresDiskOnlyInGoogle(msg.project,
+                                             ZoneName("us-central1-a"),
+                                             msg.appName,
+                                             app.appResources.namespace.name).adaptError {
+          case e =>
+            PubsubKubernetesError(AppError(e.getMessage, ctx.now, ErrorAction.CreateGalaxyApp, ErrorSource.Disk, None),
+                                  Some(msg.appId),
+                                  false,
+                                  None,
+                                  None)
+        }
+      } else { F.unit }
+
       // parallelize disk creation and cluster/nodepool monitoring
-      parPreAppCreationSetup = List(createDisk, monitorClusterOrNodepool).parSequence_
+      parPreAppCreationSetup = List(createDisk, createSecondDisk, monitorClusterOrNodepool).parSequence_
 
       // build asynchronous task
       task = for {
@@ -826,7 +929,7 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift: Parallel](
             )
         }
 
-      deleteDisk = msg.diskId.traverse(diskId => deleteDiskForApp(diskId)).adaptError {
+      deleteDisk = msg.diskId.traverse_(diskId => deleteDiskForApp(diskId)).adaptError {
         case e =>
           PubsubKubernetesError(
             AppError(e.getMessage, ctx.now, ErrorAction.DeleteGalaxyApp, ErrorSource.Disk, None),
@@ -837,11 +940,29 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift: Parallel](
           )
       }
 
+      dbAppOpt <- KubernetesServiceDbQueries.getActiveFullAppByName(msg.project, msg.appName).transaction
+      dbApp <- F.fromOption(dbAppOpt, AppNotFoundException(msg.project, msg.appName, ctx.traceId))
+      app = dbApp.app
+      deletePostgresDisk = deleteGalaxyPostgresDiskOnlyInGoogle(msg.project,
+                                                                ZoneName("us-central1-a"),
+                                                                msg.appName,
+                                                                app.appResources.namespace.name)
+        .adaptError {
+          case e =>
+            PubsubKubernetesError(AppError(e.getMessage, ctx.now, ErrorAction.DeleteGalaxyApp, ErrorSource.Disk, None),
+                                  Some(msg.appId),
+                                  false,
+                                  None,
+                                  None)
+        }
+
+      deleteDisksInParallel = List(deleteDisk, deletePostgresDisk).parSequence_
+
       // TODO: can some of this be done in parallel?
       task = for {
         _ <- deleteApp
         _ <- deleteNodepool
-        _ <- deleteDisk
+        _ <- deleteDisksInParallel
       } yield ()
 
       _ <- if (sync) task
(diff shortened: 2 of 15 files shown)


http/src/main/resources/reference.conf
```diff
@@ -260,6 +260,8 @@ gke {
         kind = "ClusterIP"
       }
     ]
+    postgresDiskNameSuffix = "gxy-postres-disk"
+    postgresDiskSize = 10
```
kyuksel: perhaps indicate unit as comment (e.g. `postgresDiskSize = 10 // in GB`) or include it in key name (e.g. `postgresDiskSizeGB`)?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -527,6 +529,43 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift: Parallel](
...
+        zone,
+        OperationName(operation.getName),
+        config.persistentDiskMonitorConfig.create.interval,
+        config.persistentDiskMonitorConfig.create.maxAttempts,
+        None
+      )(
+        F.unit,
+        F.raiseError(
+          new TimeoutException(
+            s"Fail to create Galaxy postgres disk in a timely manner. Project: ${project.value}, AppName: ${appName.value}"
```
kyuksel:
```suggestion
            s"Failed to create Galaxy postgres disk in a timely manner. Project: ${project.value}, AppName: ${appName.value}"
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -835,11 +911,14 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift: Parallel](
...
       }
 
+      deletePostgresDisk = deleteDiskOnlyInGoogle(msg.project, ZoneName("us-central1-a"), msg.appName)
+
       // TODO: can some of this be done in parallel?
       task = for {
         _ <- deleteApp
         _ <- deleteNodepool
         _ <- deleteDisk
+        _ <- deletePostgresDisk
```
kyuksel: `deletePostgresDisk` could run in parallel to `deleteDisk`? Also wonder if the two disk deletions could both run in parallel to the nodepool deletion as well 🤔

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/leoPubsubMessageSubscriberModels.scala
```diff
@@ -680,14 +682,16 @@ object LeoPubsubCodec {
     }
 
   implicit val createAppMessageEncoder: Encoder[CreateAppMessage] =
-    Encoder.forProduct8(
+    // TODO: why is this 9? where is messageType coming from?
+    Encoder.forProduct9(
```
kyuksel, line 699: I think `messageType` is inherent to all `Message`s.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -361,6 +368,9 @@ class GKEInterpreter[F[_]: Parallel: ContextShift: Timer](
         )
       )
 
+      // TODO: is throwing an exception the correct behavior here?
+      nfsDisk = dbApp.app.appResources.disk.getOrElse(throw AppCreationException("NFS Disk not found in DB"))
```
kyuksel: Makes sense to me. Could add a unit test for it perhaps?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -493,6 +504,18 @@ class GKEInterpreter[F[_]: Parallel: ContextShift: Timer](
       else appQuery.updateStatus(dbApp.app.id, AppStatus.Error).transaction
     } yield ()
 
+  def createGalaxyPostresDisk(project: GoogleProject, zone: ZoneName, appName: AppName): Disk =
+    Disk
+      .newBuilder()
+      .setName(s"${appName.value}-${config.galaxyAppConfig.postgresDiskNameSuffix}")
+      .setZone(zone.value)
+      .setSizeGb(config.galaxyAppConfig.postgresDiskSize.gb.toString)
+      .setPhysicalBlockSizeBytes("4096")
```
kyuksel: Should the value `4096` come from a config?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -787,8 +813,12 @@ class GKEInterpreter[F[_]: Parallel: ContextShift: Timer](
...
       raw"""galaxy.rbac.serviceAccount=${ksa.value}""",
       raw"""rbac.serviceAccount=${ksa.value}""",
-      // TODO Update during https://broadworkbench.atlassian.net/browse/IA-2171
-      raw"""persistence={}"""
+      raw"""persistence.nfs.persistentVolume.extraSpec.gcePersistentDisk.pdName=${nfsDisk.name.value}""",
+      raw"""persistence.nfs.size=${nfsDisk.size.gb.toString}""",
+      raw"""persistence.postgres.persistentVolume.extraSpec.gcePersistentDisk.pdName=${appName.value}-${config.galaxyAppConfig.postgresDiskNameSuffix}""",
+      raw"""persistence.postgres.size=${config.galaxyAppConfig.postgresDiskSize.gb.toString}""",
+      raw"""nfs.persistence.existingClaim=nfs-disk-pvc""",
+      raw"""galaxy.postgresql.persistence.existingClaim=postgres-disk-pvc"""
```
kyuksel: Based on [Alex's comment](https://github.com/galaxyproject/galaxykubeman-helm/pull/7#issuecomment-689904354) from earlier, I wonder if we should set these two names ourselves instead of using Galaxy chart's defaults so if they happen to change the default, we wouldn't be affected; e.g. 

```
--set persistence.nfs.name=custom-nfs-name
--set persistence.postgres.name=custom-postgres-name
```

and then

```
--set nfs.persistence.existingClaim=custom-nfs-name-pvc
--set galaxy.postgresql.persistence.existingClaim=custom-postgres-name-pvc
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -527,6 +529,43 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift: Parallel](
     }
   }
```
kyuksel, line 535: I was assuming we're adding these `*InGoogle` methods instead of reusing the previously existing counterparts because we don't want to perform the DB operations for `postgress disk`. If so, I wonder if we should add a comment to the `*InGoogle` methods explaining why they exist and perhaps suggest they be removed when we start persisting `postgress disk` info.

http/src/main/resources/reference.conf
gpcarr: seems like a good addition, will do!

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
gpcarr: I was thinking the same - I planned on picking up the disk work post-alpha but in case that doesn't end up happening then comments for these methods are definitely very useful

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
gpcarr: I think it's on a disk config, I'll look that up

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
gpcarr: I was thinking the same and even had the question of - can two PVCs in the same cluster have the same name? with the current implementation (ie using the values that exist in the chart), there would be multiple PVCs with the same name and I don't think that's ideal or even correct. Going to be reaching out to Alex for advice on this

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
gpcarr: will do!

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/leoPubsubMessageSubscriberModels.scala
gpcarr: ahh okay, I just thought it was weird that it was only on the `Encoder` and not the `Decoder`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
gpcarr: Not sure if disk deletion could run in parallel to nodepool deletion but i think `deleteDisk` and `deletePostgresDisk` can run in parallel and was thinking the same for when they are created

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
kyuksel: PVC is a namespaced resource so I don't think multiple PVCs using the same name will be a problem as long as they belong to distinct namespaces. To clarify, my comment was about not depending on the Galaxy default PVC names.

gpcarr: ah i see, that makes sense - will update!

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
rtitle: Maybe could even name this method `createGalaxyPostgresDiskOnlyInGoogle`?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -701,8 +779,12 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift: Parallel](
           )
       }
 
+      createSecondDisk = if (msg.appType == Galaxy) {
+        createDiskOnlyInGoogle(msg.project, ZoneName("us-central1-a"), msg.appName)
```
rtitle: Maybe we want `adaptError` similar to above, to ensure this falls through to `handleKubernetesError`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -835,11 +918,14 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift: Parallel](
           )
       }
 
+      deletePostgresDisk = deleteDiskOnlyInGoogle(msg.project, ZoneName("us-central1-a"), msg.appName)
```
rtitle: Same as above, about adding `adaptError`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -731,6 +813,7 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift: Parallel](
                     )
                   }
                 }
+                _ <- deleteDiskOnlyInGoogle(msg.project, ZoneName("us-central1-a"), msg.appName)
```
rtitle: Since `deleteApp` deletes both disks, do we need this line here?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -493,6 +504,18 @@ class GKEInterpreter[F[_]: Parallel: ContextShift: Timer](
       else appQuery.updateStatus(dbApp.app.id, AppStatus.Error).transaction
     } yield ()
 
+  def createGalaxyPostresDisk(project: GoogleProject, zone: ZoneName, appName: AppName): Disk =
+    Disk
+      .newBuilder()
+      .setName(s"${appName.value}-${config.galaxyAppConfig.postgresDiskNameSuffix}")
```
rtitle: Could this call `getGalaxyPostresDiskName` so we don't duplicate logic?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -493,6 +504,18 @@ class GKEInterpreter[F[_]: Parallel: ContextShift: Timer](
...
+      .setZone(zone.value)
+      .setSizeGb(config.galaxyAppConfig.postgresDiskSize.gb.toString)
+      .setPhysicalBlockSizeBytes(config.galaxyAppConfig.postgresDiskBlockSize)
+      .build()
+
+  def getGalaxyPostresDiskName(appName: AppName): DiskName =
+    DiskName(s"${appName.value}-${config.galaxyAppConfig.postgresDiskNameSuffix}")
```
rtitle: This will name disks like `saturn-k8s-<uuid>-gxy-postres-disk`

I think this is reasonable, as long as it doesn't exceed name length requirements. :) In the future the UI will name this disk itself.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -787,8 +813,14 @@ class GKEInterpreter[F[_]: Parallel: ContextShift: Timer](
       raw"""galaxy.rbac.enabled=false""",
       raw"""galaxy.rbac.serviceAccount=${ksa.value}""",
       raw"""rbac.serviceAccount=${ksa.value}""",
-      // TODO Update during https://broadworkbench.atlassian.net/browse/IA-2171
-      raw"""persistence={}"""
+      raw"""persistence.nfs.name=${config.galaxyAppConfig.nfsPersistenceName}""",
```
rtitle: Does it work setting `persistence.nfs.name` and `persistence.postgres.name` to a static value when there are multiple apps in a cluster? For some things we need cluster-unique names; for other things names only need to be unique within a namespace. Not sure which category this is :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -787,8 +813,14 @@ class GKEInterpreter[F[_]: Parallel: ContextShift: Timer](
...
+      raw"""persistence.nfs.persistentVolume.extraSpec.gcePersistentDisk.pdName=${nfsDisk.name.value}""",
+      raw"""persistence.nfs.size=${nfsDisk.size.gb.toString}""",
```
rtitle: `nfsDisk.size.gb.toString` will just return the number -- does the helm chart need a unit like `G`?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -787,8 +813,14 @@ class GKEInterpreter[F[_]: Parallel: ContextShift: Timer](
...
+      raw"""persistence.postgres.name=${config.galaxyAppConfig.postgresPersistenceName}""",
+      raw"""persistence.postgres.persistentVolume.extraSpec.gcePersistentDisk.pdName=${appName.value}-${config.galaxyAppConfig.postgresDiskNameSuffix}""",
```
rtitle: Wonder if we could re-use the above method somehow so as to not duplicate this logic again

rtitle:
A few comments above, overall looks good so far.

http/src/main/resources/reference.conf
```diff
@@ -542,6 +546,7 @@ persistentDisk {
   defaultDiskType = "pd-standard"
   defaultBlockSizeBytes = 4096
   zone = "us-central1-a"
+  defaultGalaxyNFSDiskSizeGB = 300
```
kyuksel: Nit: It'd be more consistent to pick the convention either of line 267 or line 549 (unit in comment vs embedded in name) and go with it.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -368,8 +368,10 @@ class GKEInterpreter[F[_]: Parallel: ContextShift: Timer](
         )
       )
 
-      // TODO: is throwing an exception the correct behavior here?
-      nfsDisk = dbApp.app.appResources.disk.getOrElse(throw AppCreationException("NFS Disk not found in DB"))
+      nfsDisk <- F.fromOption(
+        dbApp.app.appResources.disk,
+        AppCreationException(s"NFS Disk not found in DB for app ${app.appName.value} | trace id: ${ctx.traceId}")
```
kyuksel: Nit:
```suggestion
        AppCreationException(s"NFS disk not found in DB for app ${app.appName.value} | trace id: ${ctx.traceId}")
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -504,17 +506,17 @@ class GKEInterpreter[F[_]: Parallel: ContextShift: Timer](
       else appQuery.updateStatus(dbApp.app.id, AppStatus.Error).transaction
     } yield ()
 
-  def createGalaxyPostresDisk(project: GoogleProject, zone: ZoneName, appName: AppName): Disk =
+  def createGalaxyPostresDisk(project: GoogleProject, zone: ZoneName, namespaceName: NamespaceName): Disk =
```
kyuksel: Typo:
```suggestion
  def createGalaxyPostgresDisk(project: GoogleProject, zone: ZoneName, namespaceName: NamespaceName): Disk =
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -504,17 +506,17 @@ class GKEInterpreter[F[_]: Parallel: ContextShift: Timer](
...
+      .setName(getGalaxyPostresDiskName(namespaceName).value)
       .setZone(zone.value)
       .setSizeGb(config.galaxyAppConfig.postgresDiskSize.gb.toString)
-      .setPhysicalBlockSizeBytes("4096")
+      .setPhysicalBlockSizeBytes(config.galaxyAppConfig.postgresDiskBlockSize)
       .build()
 
-  def getGalaxyPostresDiskName(appName: AppName): DiskName =
-    DiskName(s"${appName.value}-${config.galaxyAppConfig.postgresDiskNameSuffix}")
+  def getGalaxyPostresDiskName(namespaceName: NamespaceName): DiskName =
```
kyuksel: Typo:
```suggestion
  def getGalaxyPostgresDiskName(namespaceName: NamespaceName): DiskName =
```

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriberSpec.scala
```diff
@@ -676,56 +677,6 @@ class LeoPubsubMessageSubscriberSpec
...
-      leoSubscriber = makeLeoSubscriber(asyncTaskQueue = queue)
-      asyncTaskProcessor = AsyncTaskProcessor(AsyncTaskProcessor.Config(10, 10), queue)
-      _ <- leoSubscriber.handleCreateAppMessage(msg)
-      _ <- withInfiniteStream(asyncTaskProcessor.process, assertions)
-    } yield ()
-
-    res.unsafeRunSync()
-  }
-
   it should "be able to create an multiple apps in a cluster" in isolatedDbTest {
```
kyuksel: Not part of your changes but might as well fix a typo while at it 😛 
```suggestion
  it should "be able to create multiple apps in a cluster" in isolatedDbTest {
```

kyuksel (commented):
LGTM...just a few typos and nits inline 😊

http/src/main/resources/reference.conf
```diff
@@ -260,6 +260,12 @@ gke {
         kind = "ClusterIP"
       }
     ]
+    nfsPersistenceName = "nfs-disk"
+    postgresPersistenceName = "postgres-disk"
+    # TODO: remove post-alpha once persistence is in place for Galaxy
+    postgresDiskNameSuffix = "gxy-postres-disk"
+    postgresDiskSize = 10 # in GB
+    postgresDiskBlockSize = "4096"
```
Qi77Qi: any reason this is a `String` not `Int`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/GalaxyAppConfig.scala
```diff
@@ -10,7 +10,13 @@ final case class GalaxyAppConfig(releaseNameSuffix: String,
                                  namespaceNameSuffix: String,
                                  services: List[ServiceConfig],
                                  serviceAccount: ServiceAccountName,
-                                 uninstallKeepHistory: Boolean) {
+                                 uninstallKeepHistory: Boolean,
+                                 nfsPersistenceName: String,
+                                 postgresPersistenceName: String,
```
Qi77Qi: can we group these postgresDisk config into a separate data type such as `PostgresDiskConfig`?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -527,6 +531,51 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift: Parallel](
...
+          )
+        ),
+        F.unit
+      )
+    } yield ()
+
+    create.onError {
+      case e =>
+        for {
+          _ <- logger.error(e)(
```
Qi77Qi, line 574: I think this will log the same error twice since the main error handler also logs error https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala#L113...probably better to create a `PostgresDiskError` similar to `PubsubKubernetesError`....that away you can remove `onError`, and it'll be logged in the main error handler

gpcarr: i'm not sure it'll log it twice, we do the same for normal disk creation (in google + in our db): https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala#L520

rtitle (approved):

PR #1639, opened 2020-10-07
[IA-2246] Unable to delete Error'd apps in some cases

rtitle:
https://broadworkbench.atlassian.net/browse/IA-2246

2 behavior changes:
1. In Back Leo, don't delete the Sam resource if the app goes into Error status and we're cleaning up GCP resources. Apps in `Error` status should still have records in Sam.
2. In Front Leo, don't send a `DeleteAppMessage` for deleting Error'd apps (we have the same behavior for runtimes)

Tested on a fiab. Will think about whether this can be unit tested.

---

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesServiceDbQueries.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesServiceDbQueries.scala
@@ -69,8 +69,8 @@ object KubernetesServiceDbQueries {
   def getActiveFullAppByName(googleProject: GoogleProject, appName: AppName, labelFilter: LabelMap = Map())(
     implicit ec: ExecutionContext
   ): DBIO[Option[GetAppResult]] =
-    getActiveFullApp(listClustersByProject(Some(googleProject)),
-                     nodepoolQuery.filter(_.destroyedDate === dummyDate),
+    getActiveFullApp(listClustersByProject(Some(googleProject), true),
+                     nodepoolQuery,
                      appQuery.findActiveByNameQuery(appName),
                      labelFilter)
 
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/KubernetesServiceInterp.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/KubernetesServiceInterp.scala
@@ -255,15 +255,32 @@ final class LeoKubernetesServiceInterp[F[_]: Parallel](
         )(d => F.pure(Some(d.id)))
       else F.pure[Option[DiskId]](None)
 
-      deleteMessage = DeleteAppMessage(
-        appResult.app.id,
-        appResult.app.appName,
-        appResult.nodepool.id,
-        appResult.cluster.googleProject,
-        diskOpt,
-        Some(ctx.traceId)
-      )
-      _ <- publisherQueue.enqueue1(deleteMessage)
+      // If the app status is Error, we can assume that the underlying app/nodepool
+      // has already been deleted. So we just transition the app to Deleted status
+      // without sending a message to Back Leo.
+      //
+      // Note this has the side effect of not deleting the disk if requested to do so. The
+      // caller must manually delete the disk in this situation. We have the same behavior for
+      // runtimes.
+      _ <- if (appResult.app.status == AppStatus.Error) {
+        for {
+          _ <- authProvider.notifyResourceDeleted(appResult.app.samResourceId,
+                                                  appResult.app.auditInfo.creator,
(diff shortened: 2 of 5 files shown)


rtitle:
Currently adding unit tests to this, and re-confirming behavior in a fiab.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/KubernetesServiceInterp.scala
```diff
@@ -255,15 +255,32 @@ final class LeoKubernetesServiceInterp[F[_]: Parallel](
...
-        Some(ctx.traceId)
-      )
-      _ <- publisherQueue.enqueue1(deleteMessage)
+      // If the app status is Error, we can assume that the underlying app/nodepool
+      // has already been deleted. So we just transition the app to Deleted status
+      // without sending a message to Back Leo.
+      //
+      // Note this has the side effect of not deleting the disk if requested to do so. The
+      // caller must manually delete the disk in this situation. We have the same behavior for
+      // runtimes.
```
kyuksel, line 263: For the aforementioned side effect, if the user doesn't delete the disk, would any of the clean-up mechanisms delete it today? If not, would it make sense to either send a `delete disk` message or to create a cron job (as a separate ticket as needed) to detect this case and clean the disk up after the fact. It feels like it may be too easy for users to not be aware that they have to delete the disk separately, which would incur costs silently, which usually makes it important to address.

kyuksel (commented):
LGTM overall. Inline concern and comment about disk potentially being left behind and how we may be able to address it.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/KubernetesServiceInterp.scala
rtitle: I don't think anything would automatically delete the disk if this happens. In generate we never want to automatically delete disks unless the user explicitly requests it.

In the UI the user can view all the active disks they currently have, and the UI warns you if you have >1 disk at a time. So I think the user would be notified, but may be annoyed they have to separately delete the disk when they get an `Error`'d runtime or app. 

Sending a `delete disk` message in this situation may make sense, but I'd kind of prefer to ticket that separately since the bug exists for runtimes and apps and is not directly related to the bug this PR is trying to address.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesServiceDbQueries.scala
```diff
@@ -69,8 +69,8 @@ object KubernetesServiceDbQueries {
   def getActiveFullAppByName(googleProject: GoogleProject, appName: AppName, labelFilter: LabelMap = Map())(
```
rtitle, line 69: Changing this query's behavior slightly: it now returns non-deleted apps even if the underlying cluster/nodepool is deleted. This can happen since we clean up resources when an app enters `Error` status -- we still want the app to be retrievable in this case.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/KubernetesServiceInterp.scala
kyuksel: > I don't think anything would automatically delete the disk if this happens. In generate we never want to automatically delete disks unless the user explicitly requests it.

Certainly. To clarify, I was talking about the case where a user does request the disk to be deleted as part of the `deleteApp` request.

> Sending a delete disk message in this situation may make sense, but I'd kind of prefer to ticket that separately since the bug exists for runtimes and apps and is not directly related to the bug this PR is trying to address.

Creating a separate ticket to handle it for both apps and runtimes sounds good to me 😊

kyuksel (approved):

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesServiceDbQueries.scala
kyuksel: That makes sense to me. The method is used in several places. Haven't checked if there may be an adverse effect in any of the call sites. Hopefully the tests would catch it if there is ™️ .

rtitle: Yeah I'm a bit nervous since this is used in so many places, but I think the new behavior is correct, and I'll check all tests + test manually in the UI :)

rtitle:
Testing in Terra UI in a fiab -- app creation works, and I can successfully delete Error'd and non-Error'd apps. Testing a bit more, but I plan to merge this later today.

PR #1640, opened 2020-10-07
IA-2241 bump sbt

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-2241

TODO: Test in fiab

---

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeServiceInterp.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeServiceInterp.scala
@@ -280,7 +280,7 @@ class RuntimeServiceInterp[F[_]: Parallel](config: RuntimeServiceConfig,
       runtimeConfig <- RuntimeConfigQueries.getRuntimeConfig(runtime.runtimeConfigId).transaction
       persistentDiskToDelete <- runtimeConfig match {
         case x: RuntimeConfig.GceWithPdConfig =>
-          x.persistentDiskId.flatTraverse { diskId =>
+          x.persistentDiskId.flatTraverse[F, DiskId] { diskId =>
             for {
               diskOpt <- persistentDiskQuery.getPersistentDiskRecord(diskId).transaction
               disk <- F.fromEither(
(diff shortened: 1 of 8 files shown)


rtitle (approved):
lgtm

Qi77Qi:
@rtitle I accidentally pushed https://github.com/DataBiosphere/leonardo/pull/1640/commits/94fd6197001650e46d34147ed3f709040d0568aa to this branch before merging the sbt change...let me know if you have any comments to the second commit...I think updating the disk status should be desirable regardless..and I'll try to reproduce what you did...but one question to what you did, did you use the same disk name the second time you create the runtime+disk?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeServiceInterp.scala
```diff
@@ -307,11 +307,10 @@ class RuntimeServiceInterp[F[_]: Parallel](config: RuntimeServiceConfig,
                   F.unit)
               )
               _ <- RuntimeConfigQueries.updatePersistentDiskId(runtime.runtimeConfigId, None, ctx.now).transaction
-            } yield {
-              if (req.deleteDisk)
-                Some(diskId)
-              else None
-            }
+              res <- if (req.deleteDisk)
+                persistentDiskQuery.updateStatus(diskId, DiskStatus.Deleting, ctx.now).transaction.as(Some(diskId))
```
Qi77Qi, line 311: this is the real fix...previously we weren't updating disk status properly, hence UI is reusing the same disk. But UI won't try to re-use the same disk if disk is marked as `Deleting`

Qi77Qi:
building go lib failed..jenkins retest

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeServiceInterp.scala
rtitle: Makes sense -- so the disk deletion will be done by Back Leo, and back Leo will then transition the disk to `Deleted`?

rtitle:
Disk change looks good to me too -- would be good to confirm the fix in Terra UI

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeServiceInterp.scala
Qi77Qi: yep https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala#L222

Qi77Qi: Tested in fiab....with dev code, if I create a runtime immediately after deletion, UI try to use the previously created disk...With this PR (disk status is updated to `Deleting` once delete runtime call finishes), when I create a runtime immediately after deletion, UI will create a new disk name. cc @panentheos  in case I'm misunderstanding UI behavior

PR #1641, opened 2020-10-08
[IA-2251] Do not allow concurrent nodepool operations in front leo

jdcanas:
This disallows concurrent nodepoool operations in front leo, and makes it so we do not mark the app as deleted in the db until both the app and nodepool have been deleted

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/apps/BatchNodepoolCreationSpec.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/apps/BatchNodepoolCreationSpec.scala
@@ -162,6 +162,8 @@ class BatchNodepoolCreationSpec
             s"BatchNodepoolCreationSpec: app ${googleProject.value}/${appName1.value} delete result: $monitorApp1DeletionResult"
           )
 
+          _ = monitorApp1DeletionResult.map(_.status).toSet shouldBe Set(AppStatus.Deleted, AppStatus.Running)
+
           _ <- LeonardoApiClient.deleteApp(googleProject, appName2)
           monitorAppDeletionResult <- testTimer.sleep(30 seconds) >> streamFUntilDone(listApps, 120, 10 seconds)(
             testTimer,
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesServiceDbQueries.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesServiceDbQueries.scala
@@ -15,6 +15,7 @@ import org.broadinstitute.dsde.workbench.leonardo.http.GetAppResult
 import org.broadinstitute.dsde.workbench.leonardo.model.LeoException
 import org.broadinstitute.dsde.workbench.model.google.GoogleProject
 import com.rms.miu.slickcats.DBIOInstances._
+import org.broadinstitute.dsde.workbench.google2.KubernetesClusterNotFoundException
 
 import scala.concurrent.ExecutionContext
 
@@ -84,6 +85,25 @@ object KubernetesServiceDbQueries {
                      nodepoolQuery,
                      appQuery.getByIdQuery(appId),
                      labelFilter)
+
+  def hasClusterOperationInProgress(clusterId: KubernetesClusterLeoId)(
+    implicit ec: ExecutionContext
+  ): DBIO[Boolean] =
+    for {
+      clusterOpt <- kubernetesClusterQuery.getMinimalClusterById(clusterId)
+      cluster <- clusterOpt match {
+        case None =>
+          DBIO.failed(
+            KubernetesClusterNotFoundException(
+              s"did not find cluster with id $clusterId while checking for concurrent cluster operations"
+            )
+          )
+        case Some(c) => DBIO.successful(c)
+      }
+      hasOperationInProgress = cluster.nodepools.exists(nodepool =>
+        NodepoolStatus.nonParellizableStatuses.contains(nodepool.status)
+      )
+    } yield hasOperationInProgress
 
   private[db] def getActiveFullApp(clusterQuery: Query[KubernetesClusterTable, KubernetesClusterRecord, Seq],
                                    nodepoolQuery: Query[NodepoolTable, NodepoolRecord, Seq],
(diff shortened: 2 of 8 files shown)


jdcanas:
jenkins retest

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -958,10 +958,11 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift: Parallel](
 
       deleteDisksInParallel = List(deleteDisk, deletePostgresDisk).parSequence_
 
-      // TODO: can some of this be done in parallel?
+      // The app must be deleted before the nodepool and disk, to future proof against the app potentially flushing the postgres db somewhere
       task = for {
         _ <- deleteApp
         _ <- deleteNodepool
+        _ <- if (!errorAfterDelete) appQuery.markAsDeleted(msg.appId, ctx.now).transaction.void else F.unit
         _ <- deleteDisksInParallel
```
jdcanas, line 970: because this is blocking in terms of cluster actions until the app is marked as deleted, if the user requests a disk delete, we don't wait for that portion.

jdcanas:
jenkins retest

jdcanas:
jenkins retest

jdcanas:
jenkins retest.... not really sure why its failing

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/apps/BatchNodepoolCreationSpec.scala
```diff
@@ -162,6 +162,9 @@ class BatchNodepoolCreationSpec
             s"BatchNodepoolCreationSpec: app ${googleProject.value}/${appName1.value} delete result: $monitorApp1DeletionResult"
           )
 
+          _ = monitorApp1DeletionResult.map(_.status).sortBy(_.toString) shouldBe List(AppStatus.Deleted,
```
rtitle: Nit: `monitorApp1DeletionResult.map(_.status).toSet shouldBe Set(AppStatus.Deleted, AppStatus.Running)`?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesServiceDbQueries.scala
```diff
@@ -85,6 +86,24 @@ object KubernetesServiceDbQueries {
...
+    for {
+      clusterOpt <- kubernetesClusterQuery.getMinimalClusterById(clusterId)
+      cluster <- if (clusterOpt.isEmpty)
+        DBIO.failed(
+          KubernetesClusterNotFoundException(
+            s"did not find cluster with id $clusterId while checking for concurrent cluster operations"
+          )
+        )
+      else DBIO.successful(clusterOpt.get)
+      _ = println(s"cluster statuses ${cluster.nodepools.map(_.status)}")
```
rtitle: Remove println

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesServiceDbQueries.scala
```diff
@@ -85,6 +86,24 @@ object KubernetesServiceDbQueries {
...
+      hasOperationInProgress = cluster.nodepools.map(_.status).map { status =>
```
rtitle: Nit: how about
```
hasOperationInProgress = cluster.nodepools.exists(np =>
  NodepoolStatus.nonParellizableStatuses.contains(np.status)
)
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesServiceDbQueries.scala
```diff
@@ -85,6 +86,24 @@ object KubernetesServiceDbQueries {
                      appQuery.getByIdQuery(appId),
                      labelFilter)
 
+  def hasClusterOperationInProgress(clusterId: KubernetesClusterLeoId)(
+    implicit ec: ExecutionContext
+  ): DBIO[Boolean] =
+    for {
+      clusterOpt <- kubernetesClusterQuery.getMinimalClusterById(clusterId)
+      cluster <- if (clusterOpt.isEmpty)
```
rtitle: Nit: I usually like `match` to unwrap options:
```
cluster <- clusterOpt match {
  case None => DBIO.failed(..)
  case Some(cluster) => cluster
}
```

I guess DBIO doesn't have an analog to cats-effect `F.fromOption(..)`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -503,14 +506,11 @@ class GKEInterpreter[F[_]: Parallel: ContextShift: Timer](
       )
 
       _ <- if (!params.errorAfterDelete) {
-        for {
-          _ <- authProvider.notifyResourceDeleted(dbApp.app.samResourceId,
-                                                  dbApp.app.auditInfo.creator,
-                                                  params.googleProject)
-          _ <- appQuery.markAsDeleted(dbApp.app.id, ctx.now).transaction
-        } yield ()
+        authProvider
```
rtitle: I feel like we should always do `appQuery.markAsDeleted` and `authProvider.notifyResourceDeleted` together.. can we move the authProvider call too to `LeoPubsubMessageSubscriber`?

rtitle (commented):
Overall looks good, some Scala nitpicks above

Qi77Qi:
just for my own understanding...is this before related to https://broadworkbench.atlassian.net/wiki/spaces/IA/pages/850001921/2020-10-01+Concurrent+Nodepool+Operations? and you didn't end up using any of the solution mentioned in the doc, but rather just checking DB to see if there's other operation in progress?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesServiceDbQueries.scala
jdcanas: Yeah, I wish it did :(

rtitle (approved):

rtitle:
@Qi77Qi I think this is Solution 3 in that doc: detect concurrent operations and return an error to the user.

I recall a team meeting a while back where we decided on this approach.

Qi77Qi:
hmm okay....think I saw some messages after I came back about we're going with solution 2, that's why I'm a confused. @jdcanas can you update the doc as to which approach was chosen.

jdcanas:
I updated the doc with a conclusion section.

PR #1648, opened 2020-10-14
[No ticket] Default Galaxy NFS disk to 250

gpcarr:
Tested by creating a new app in my fiab and validating that the two disks are created with the correct names and sizes and attached to galaxy at helm install. also validated that the app successfully transitions from `PROVISIONING` to `RUNNING` and that the PV/PVC pairs are created correct.

---

In all cases:


rtitle (commented):
The change looks good; looks like there's a unit test failure though

gpcarr:
> The change looks good; looks like there's a unit test failure though

just pushed the fix :)

rtitle (approved):

PR #1649, opened 2020-10-15
[IA-2263] Front Leo/proxy performance improvements

rtitle:
https://broadworkbench.atlassian.net/browse/IA-2263

Still a bit experimental. This PR does the following:
* Move the oauth2 `tokeninfo` call (which is called pretty frequently by the proxy) out of `HttpGoogleDataprocDAO` and into its own dedicated class. Also made it use `google2` infrastructure which I think will make it more reliable.
* Periodically log stats (size, hit count, etc) of our various guava caches and also send them to OpenTelemetry. I hope this will help us tune our caches better going forward.

---

In all cases:


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/google/GoogleOAuth2Service.scala
```diff
@@ -0,0 +1,30 @@
...
+import cats.effect.concurrent.Semaphore
+import cats.effect.{Async, Blocker, ContextShift, Resource, Sync, Timer}
+import cats.mtl.ApplicativeAsk
+import com.google.api.client.googleapis.javanet.GoogleNetHttpTransport
+import com.google.api.client.json.jackson2.JacksonFactory
+import com.google.api.services.oauth2.Oauth2
+import io.chrisdavenport.log4cats.StructuredLogger
+import org.broadinstitute.dsde.workbench.model.{TraceId, UserInfo}
+
+trait GoogleOAuth2Service[F[_]] {
```
rtitle, line 14: I thought about putting `GoogleOAuth2Service`/`GoogleOAuth2Interpreter` in `workbench-libs/google2`, but left it in Leo for now. I could move if people feel strongly.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/google/HttpGoogleDataprocDAO.scala
```diff
@@ -210,25 +200,6 @@ class HttpGoogleDataprocDAO(
...
-    val request = oauth2.tokeninfo().setAccessToken(accessToken)
-    retry(retryPredicates: _*)(() => executeGoogleRequest(request)).map { tokenInfo =>
-      (UserInfo(OAuth2BearerToken(accessToken),
-                WorkbenchUserId(tokenInfo.getUserId),
-                WorkbenchEmail(tokenInfo.getEmail),
-                tokenInfo.getExpiresIn.toInt),
-       Instant.now().plusSeconds(tokenInfo.getExpiresIn.toInt))
-    } recover {
-      case e: GoogleJsonResponseException =>
-        val msg = s"Call to Google OAuth API failed. Status: ${e.getStatusCode}. Message: ${e.getDetails.getMessage}"
```
rtitle, line 223: This line is throwing a NPE in prod currently, I think because `e.getDetails` is `null`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/ProxyService.scala
```diff
@@ -93,18 +100,35 @@ class ProxyService(
     .newBuilder()
     .expireAfterWrite(proxyConfig.tokenCacheExpiryTime.toSeconds, TimeUnit.SECONDS)
     .maximumSize(proxyConfig.tokenCacheMaxSize)
+    .recordStats()
     .build(
-      new CacheLoader[String, Future[(UserInfo, Instant)]] {
```
rtitle, line 97: Previously this was caching a `Future` which isn't ideal. One problem with this is it would cache _all_ requests, including invalid tokens which would cause the `Future` to fail. This may have been filling up the cache with junk and causing more requests to `tokeninfo` for legitimate tokens. Now it should only cache _valid_ token lookups, and fail with `AuthenticationError` if we failed to look up the token.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/ProxyService.scala
```diff
@@ -211,18 +240,6 @@ class ProxyService(
       r <- proxyInternal(hostContext, request)
     } yield r
 
-    res.onError {
```
rtitle, line 214: I think this was causing duplicate (noisy) logging. I don't think `equest.entity.discardBytes()` is necessary either; we had added that to address some apache issues but it turned out to be something else.

Qi77Qi: there're a few places where we're doing `onError`, but then also `handleErrorWith` at a later point...would be good to address these slowly

rtitle:
I tried this on a fiab and it seems fine.. going to also run Leo proxy perf tests shortly

rtitle:
`LeoProxySimulation` looks good at 200 users, about the same as in the nightly runs. Although the perf tests don't really simulate reality in that they use the same user/token for all requests, so the caches aren't very full.

I think I'm comfortable merging this though, looking forward to seeing the actual metrics in prod.

http/src/main/resources/reference.conf
```diff
@@ -377,9 +377,9 @@ proxy {
   proxyPort = 443
   dnsPollPeriod = 15 seconds
   tokenCacheExpiryTime = 60 minutes
-  tokenCacheMaxSize = 10000
+  tokenCacheMaxSize = 20000
```
rtitle, line 380: Just a guess really.. the `CacheMetrics` in this PR should help inform us more about tuning these caches.

Was also thinking about analyzing a production heap dump for potential improvements there.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/Boot.scala
```diff
@@ -281,7 +281,15 @@ object Boot extends IOApp {
           )
         }
 
-        val frontLeoOnlyProcesses = List(dateAccessedUpdater.process) //We only need to update dateAccessed in front leo
+        val frontLeoOnlyProcesses = List(
+          dateAccessedUpdater.process, // We only need to update dateAccessed in front leo
+          appDependencies.authProvider.cacheMetrics.process,
```
rtitle: Turns out we have quite a few guava caches in front Leo

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/CacheMetrics.scala
```diff
@@ -0,0 +1,39 @@
...
+import io.chrisdavenport.log4cats.Logger
+import org.broadinstitute.dsde.workbench.openTelemetry.OpenTelemetryMetrics
+
+import scala.concurrent.duration._
+
+class CacheMetrics[F[_]: Timer: Effect] private (name: String, sizeF: F[Long], statsF: F[CacheStats])(
+  implicit metrics: OpenTelemetryMetrics[F],
+  logger: Logger[F]
+) {
+  val process: Stream[F, Unit] =
```
Qi77Qi: how about moving `sizeF` and `statsF` to arguments for `process`? and make both arguments a thunk (`() => F[???]`)?  Just so it's more obvious these 2 are functions that will be run on an interval?

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/dao/google/MockGoogleOAuth2Service$.scala
```diff
@@ -0,0 +1,36 @@
+package org.broadinstitute.dsde.workbench.leonardo.dao.google
```
Qi77Qi: why does this file has `$` in the end 🤔

Qi77Qi (commented):
looks great overall. especially adding metrics to give us some insights how things are doing in prod...one comment about `CallMetrics` class

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/CacheMetrics.scala
rtitle: I could try that.. originally I had `CacheMetrics` as an object and just passed all params to `process`. Maybe that's actually cleaner.

Qi77Qi: you probably won't be able to make it an object because of all the implicits required...another thing I would do is at call site (at each cache load class), just do `def recordCacheMetricsProcess: Stream[F, Unit] = CallMetrics[F].process`

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/dao/google/MockGoogleOAuth2Service$.scala
rtitle: Ah, I think Intellij refactor did it. Will fix

rtitle:
@Qi77Qi applied feedback in above commits

Qi77Qi (approved):
👍

rtitle:
jenkins retest

rtitle:
Galaxy test failed :( jenkins retest

rtitle:
jenkins retest

rtitle:
finally

PR #1651, opened 2020-10-16
[IA-2280] A few Sam-related performance fixes

rtitle:
https://broadworkbench.atlassian.net/browse/IA-2280

I was looking at the prod front Leo logs, and we call Sam a LOT. Basically the UI is polling `listRuntimes`, `listDisks`, and `listApps` for each user logged into Terra, and Leo is making 1-2 Sam calls for each `list*` request it receives. We also log every Sam request/response, and it looks like syslog is getting overwhelmed with the log volume.

This PR makes a few changes:

1. Don't call Sam `getResourcePolicies` if the result from the Leo DB is empty -- just return an empty list. 🤦  I think this should save a lot of Sam calls since many users don't have active runtimes/disks/apps.
2. Don't log every Sam request/response. It seems desirable but most requests just look like `GET https://sam.dsde-prod.broadinstitute.org:443/api/resources/v1/billing-project` so it doesn't really add much value. (We will still log error responses from Sam.)
3. Since we're not logging Sam calls anymore, add an OpenTelemetry counter metric for each type of Sam request (and error). This should hopefully give us some insight into our Sam calls (and may even be better than logging). 

Will test on a fiab tomorrow, would like to try and merge this for next release.

---

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/sam/SamAuthProvider.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/sam/SamAuthProvider.scala
@@ -4,6 +4,7 @@ package auth.sam
 import java.util.UUID
 import java.util.concurrent.TimeUnit
 
+import cats.data.NonEmptyList
 import cats.effect.implicits._
 import cats.effect.{Blocker, ContextShift, Effect, Sync, Timer}
 import cats.implicits._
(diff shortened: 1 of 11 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/sam/SamAuthProvider.scala
```diff
@@ -145,14 +145,17 @@ class SamAuthProvider[F[_]: Effect: Logger: Timer: OpenTelemetryMetrics](samDao:
...
     ev: ApplicativeAsk[F, TraceId]
-  ): F[List[R]] = {
-    val authHeader = Authorization(Credentials.Token(AuthScheme.Bearer, userInfo.accessToken.token))
-    for {
-      resourcePolicies <- samDao
-        .getResourcePolicies[R](authHeader)
-      res = resourcePolicies.filter { case (_, pn) => sr.policyNames.contains(pn) }
-    } yield resources.filter(r => res.exists(_._1 == r))
-  }
+  ): F[List[R]] =
```
Qi77Qi: shall we make the function take `NonEmptyList` instead?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/sam/SamAuthProvider.scala
```diff
@@ -161,21 +164,24 @@ class SamAuthProvider[F[_]: Effect: Logger: Timer: OpenTelemetryMetrics](samDao:
     implicit sr: SamResource[R],
     decoder: Decoder[R],
     ev: ApplicativeAsk[F, TraceId]
-  ): F[List[(GoogleProject, R)]] = {
-    val authHeader = Authorization(Credentials.Token(AuthScheme.Bearer, userInfo.accessToken.token))
-    for {
-      projectPolicies <- samDao.getResourcePolicies[ProjectSamResourceId](authHeader)
-      owningProjects = projectPolicies.collect {
-        case (r, SamPolicyName.Owner) => r.googleProject
+  ): F[List[(GoogleProject, R)]] =
```
Qi77Qi: same, take `NonEmptyList` as argument

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/sam/SamAuthProvider.scala
rtitle: Sure, good idea

rtitle:
@Qi77Qi applied feedback

rtitle:
jenkins retest

rtitle:
I'm having bad luck on tests, jenkins retest

gpcarr:
lgtm
> Will test on a fiab tomorrow

does it work in your fiab?

rtitle:
> does it work in your fiab?

Oh yeah I tested creating runtimes, disks, and apps today and looked at Leo logs, everything looks good.

gpcarr (approved):

Qi77Qi (approved):
thought I thumbed this earlier...oh well

PR #1655, opened 2020-10-20
[no ticket] logback update

Qi77Qi:
Keep 2 days of log on disk...We probably don't really need to keep any log on disk, but I'm not sure exactly how currently log infrastructure works, so just reducing the length of how long we keep them on disk.

probably doesn’t make much difference..but doesn’t hurt.

---

In all cases:


rtitle (approved):
Looks fine.. even though logs are streamed to syslog it's probably good to have `leonardo.log` in case syslog falls behind or has issues.

PR #1656, opened 2020-10-21
Fix for PROD-440: listRuntimes returns 0 runtimes for AoU prod service account

rtitle:
https://broadworkbench.atlassian.net/browse/PROD-440

Scala unchecked `.contains()` strikes again. :(

Root cause of this bug was it was doing `samVisibleRuntimes.contains(a)` where `samVisibleRuntimes` is a `Option[List[A]]`. It's optional because it comes from `NonEmptyList.fromList` (introduced in https://github.com/DataBiosphere/leonardo/pull/1651).

This PR fixes the issue for runtimes, disks, and apps and also adds a unit test for each case. I confirmed that each test fails before the fix, and passes after the fix.

---

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeServiceInterp.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeServiceInterp.scala
@@ -224,9 +224,9 @@ class RuntimeServiceInterp[F[_]: Parallel](config: RuntimeServiceConfig,
       ctx <- as.ask
       paramMap <- F.fromEither(processListParameters(params))
       runtimes <- RuntimeServiceDbQueries.listRuntimes(paramMap._1, paramMap._2, googleProject).transaction
-      runtimesAndProjects = runtimes.map(r => (r.googleProject, r.samResource))
       _ <- ctx.span.traverse(s => F.delay(s.addAnnotation("DB | Done listRuntime db query")))
-      samVisibleRuntimes <- NonEmptyList.fromList(runtimesAndProjects).traverse { rs =>
+      runtimesAndProjects = runtimes.map(r => (r.googleProject, r.samResource))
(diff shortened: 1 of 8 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeServiceInterp.scala
```diff
@@ -224,25 +224,30 @@ class RuntimeServiceInterp[F[_]: Parallel](config: RuntimeServiceConfig,
...
             userInfo
           )
       }
       _ <- ctx.span.traverse(s => F.delay(s.addAnnotation("Sam | Done visible runtimes")))
-    } yield {
-      // Making the assumption that users will always be able to access runtimes that they create
-      // Fix for https://github.com/DataBiosphere/leonardo/issues/821
-      runtimes
-        .filter(c =>
-          c.auditInfo.creator == userInfo.userEmail || samVisibleRuntimes.contains((c.googleProject, c.samResource))
```
rtitle, line 242: This is the bug: `samVisibleRuntimes` is a `Option[List[(GoogleProject, RuntimeSamResourceId)]]` so `.contains((proj, samResource))` always returns false

gpcarr (approved):

gpcarr:
LGTM once checks pass - does this mean that we were returning an empty list only for 4 days?

rtitle:
Thanks @gpcarr. Because we're also checking `c.auditInfo.creator == userInfo.userEmail` the `list` endpoints are still returning resources that the calling user created. That's the main use case for Terra UI.

What's broken is the `list` endpoints returning resources the calling user _didn't_ create. E.g. we allow project owners to list other users' runtimes/disks/etc. AoU happens to rely on this behavior (as do some Terra project owners I think).

Qi77Qi (approved):
🤦‍♀️ ...thanks for the quick fix 👍

PR #1659, opened 2020-10-21
[IA-2244] RStudio session abnormally terminated warning

rtitle:
https://broadworkbench.atlassian.net/browse/IA-2244

Root cause was the shutdown script wasn't working. Plus we now should run the shutdown script at deletion time as well, since we keep the PDs.

Tested on a fiab, my session is preserved when stop/starting or delete/recreating an RStudio runtime.

---

In all cases:

### DIFF
--- a/http/src/main/resources/init-resources/shutdown.sh
+++ b/http/src/main/resources/init-resources/shutdown.sh
@@ -12,5 +12,5 @@ export RSTUDIO_SERVER_NAME=$(rstudioServerName)
 
 # If RStudio is installed, cleanly shut it down
 if [ ! -z "$RSTUDIO_DOCKER_IMAGE" ] ; then
-    docker exec -u root -i $RSTUDIO_SERVER_NAME rstudio-server stop
+    docker exec -u rstudio -i $RSTUDIO_SERVER_NAME rstudio-server stop
 fi
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ZombieRuntimeMonitor.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ZombieRuntimeMonitor.scala
@@ -260,14 +260,16 @@ class ZombieRuntimeMonitor[F[_]: Parallel: ContextShift: Timer](
         s"${traceId.asString} | Deleting inactive zombie runtime: ${zombie.googleProject.value} / ${zombie.runtimeName.asString}"
       )
       _ <- metrics.incrementCounter("numOfInactiveZombieRuntimes")
-      _ <- zombie.cloudService.interpreter
-        .deleteRuntime(
-          DeleteRuntimeParams(
-            zombie.googleProject,
-            zombie.runtimeName,
-            zombie.asyncRuntimeFields.isDefined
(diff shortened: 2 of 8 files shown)


http/src/main/resources/init-resources/shutdown.sh
```diff
@@ -12,5 +12,5 @@ export RSTUDIO_SERVER_NAME=$(rstudioServerName)
 
 # If RStudio is installed, cleanly shut it down
 if [ ! -z "$RSTUDIO_DOCKER_IMAGE" ] ; then
-    docker exec -u root -i $RSTUDIO_SERVER_NAME rstudio-server stop
+    docker exec -u rstudio -i $RSTUDIO_SERVER_NAME rstudio-server stop
```
rtitle, line 15: I tested this manually; apparently it needs to run as `rstudio` in the rocker image we're using

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/ZombieRuntimeMonitor.scala
```diff
@@ -260,14 +260,16 @@ class ZombieRuntimeMonitor[F[_]: Parallel: ContextShift: Timer](
...
       )
       _ <- metrics.incrementCounter("numOfInactiveZombieRuntimes")
-      _ <- zombie.cloudService.interpreter
-        .deleteRuntime(
-          DeleteRuntimeParams(
-            zombie.googleProject,
-            zombie.runtimeName,
-            zombie.asyncRuntimeFields.isDefined
-          ) //TODO: think about this a bit more. We may want to delete disks in certain cases
+      runtimeOpt <- clusterQuery.getClusterById(zombie.id).transaction
```
rtitle, line 263: Had to run an additional query in the zombie monitor since `deleteRuntime` now takes the entire `Runtime` object. Zombie monitor will be going away soon-ish anyway

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreter.scala
```diff
@@ -212,13 +212,17 @@ class DataprocInterpreter[F[_]: Timer: Async: Parallel: ContextShift: Logger](
   override def deleteRuntime(
     params: DeleteRuntimeParams
   )(implicit ev: ApplicativeAsk[F, TraceId]): F[Option[com.google.cloud.compute.v1.Operation]] =
-    for {
-      _ <- if (params.isAsyncRuntimeFields) //check if runtime has been created
-        Async[F].liftIO(
-          IO.fromFuture(IO(gdDAO.deleteCluster(params.googleProject, params.runtimeName)))
+    if (params.runtime.asyncRuntimeFields.isDefined) { //check if runtime has been created
+      for {
+        metadata <- getShutdownScript(params.runtime, blocker)
```
rtitle, line 217: Call shutdown script at deletion time, for Dataproc and GCE. `getShutdownScript` requires the entire `Runtime` object.

gpcarr:
LGTM - good catch!

gpcarr (approved):

PR #1673, opened 2020-10-28
[IA-2279] Latest rstudio rocker versions don't launch

rtitle:
https://broadworkbench.atlassian.net/browse/IA-2279

The root problem seems to be due to missing`SameSite=None` on RStudio cookies. RStudio sets some cookies to disable auth; without `SameSite=None` the browser rejects them (because RStudio is rendered within an iframe) causing it to go into an infinite-redirect loop and not load.

Originally I tried fixing this via an RStudio setting (https://github.com/anvilproject/anvil-docker/pull/18) but it didn't actually work, so doing this in the Apache proxy instead.

I tested the regular image (`us.gcr.io/anvil-gcr-public/anvil-rstudio-bioconductor:0.0.6`) and Nitesh's image based on the latest rocker version (`nitesh1989/anvil-rstudio-base:tidyverse_4.0.2`) and both loaded (the latter doesn't work without this change).

---

In all cases:

### DIFF
--- a/http/src/main/resources/init-resources/cluster-site.conf
+++ b/http/src/main/resources/init-resources/cluster-site.conf
@@ -29,9 +29,10 @@
     # Include a ProxyPassReverse so redirects by RStudio go to the correct server name (e.g. https://notebooks.firecloud.org)
     ProxyPassReverse /proxy/${GOOGLE_PROJECT}/${RUNTIME_NAME}/rstudio/ http://127.0.0.1:8001/
 
-    # Append SameSite=None to cookies set by RStudio. This is needed because we render RStudio in an iframe.
-    # There does not appear to be a way within RStudio to do this.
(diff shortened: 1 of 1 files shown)


Qi77Qi (approved):
👍 ...just for my own understanding, what is `samesite` for jupyter requests?

rtitle:
`SameSite` is a new cookie attribute recently introduced, and is important for cross-domain cookies (which we use since we load tools in an iframe from app.terra.bio).

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite

Leo sets `SameSite=None` for its token cookie; the issue here is that RStudio is setting its own cookies _without_ `SameSite`, causing the browser to block them and bad things happening. (I think the cookies have to do with the setting to disable the RStudio login screen.)

It'd be nice if this could be configured within RStudio but I didn't see a way. I could also do this in the Leo scala proxy I guess, but the Apache proxy seemed cleaner.

gpcarr (approved):

PR #1691, opened 2020-11-09
[no ticket] experiment to see if moving to java 11 works

Qi77Qi:

---

In all cases:


Qi77Qi:
jenkins retest

rtitle:
Maybe want to update https://github.com/DataBiosphere/leonardo/blob/develop/.github/workflows/unit_test.yml#L25 and https://github.com/DataBiosphere/leonardo/blob/develop/docker/build.sh#L150 so it actually compiles with jdk11?

I feel like these classes were removed in >8, so I wonder if compilation would fail: https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dns/JupyterNameService.scala#L9

Good find on `-Djdk.net.hosts.file`, I wonder if there is a way to make that work. The challenge is we need to dynamically update the host -> IP mapping based on runtimes in the DB. We currently do it via an in-memory cache, but would be a bit harder if we need to use an external file.

Qi77Qi:
jenkins retest

rtitle:
Yeah I think compilation is failing because these classes were removed in jdk9:
```
[error] /home/runner/work/leonardo/leonardo/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dns/JupyterNameService.scala:9:20: object nameservice is not a member of package sun.net.spi
[error] import sun.net.spi.nameservice.{NameService, NameServiceDescriptor}
```
So I think we need a new mechanism for the Leo -> Jupyter DNS mapping (maybe dynamically updating the `hosts_file` somehow?)

Qi77Qi:
@rtitle yea these classes are removed. There's `java.net.InetAddress.NameService` looks the same as `sun.net.spi.nameservice.NameService`, but it's private.

Is `JupyterNameService` only used for resolving jupyter runtimes? just wonder if there's another way (say similar to how galaxy is resolved) to do this instead of using a custom nameservice provider

rtitle:
It's used by all proxy connections, so all types of runtimes and Galaxy apps use it (I guess it's a little mis-named).

It's basically used to "spoof" a DNS entry, e.g. make a hostname `1234.jupyter-dev.firecloud.org` resolve to the actual IP of the runtime. We need the spoofed DNS name in order to match the wildcard SSL cert.

It sounds like we could also achieve this by putting these host -> IP mappings in the `hosts_file`, but `NameService` was nice since we could do it all in memory in the JVM (no need to write to an external file).

rtitle:
Going to close in favor of https://github.com/DataBiosphere/leonardo/pull/2002

PR #1703, opened 2020-11-16
IA-697: adding Rstudio smoke test

xuscrbw:
Added a RStudio object and RStudioPage class to help run the test.
Right now, this WIP only has a generic: open the proxy, set a value to a variable, verify the variable shows under the Global Environment section.

Will probably do some more work to add in a more specific CSS Selector and see if there's anything else that we can do.
---

In all cases:


xuscrbw:
jenkins retest

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/rstudio/RStudioMainPage.scala
```diff
@@ -0,0 +1,33 @@
...
+import org.broadinstitute.dsde.workbench.leonardo.notebooks.JupyterPage
+import org.openqa.selenium.WebDriver
+
+import scala.concurrent.duration.FiniteDuration
+import scala.concurrent.duration._
+import scala.util.Try
+
+class RStudioMainPage(override val url: String)(implicit override val authToken: AuthToken,
+                                                implicit override val webDriver: WebDriver)
+    extends JupyterPage
```
rtitle: Just wondering if we should rename `JupyterPage` to `ToolPage` or something if it's meant to be abstract

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/rstudio/RStudioPage.scala
```diff
@@ -0,0 +1,21 @@
+package org.broadinstitute.dsde.workbench.leonardo.rstudio
+
+import com.typesafe.scalalogging.LazyLogging
+import org.broadinstitute.dsde.workbench.auth.AuthToken
+import org.openqa.selenium.WebDriver
+import org.broadinstitute.dsde.workbench.leonardo.notebooks.{JupyterPage, NotebooksListPage}
+
+class RStudioPage(override val url: String)(implicit override val authToken: AuthToken,
```
rtitle, line 12: Curious what the difference is between `RStudioMainPage` and `RStudioPage` -- I thought there was only 1 RStudio page (contrast with Jupyter which has the "list" page and the "notebook" page).

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/rstudio/RStudioSpec.scala
```diff
@@ -1,26 +1,26 @@
...
-        val getResult = Try(RStudio.getApi(runtimeFixture.runtime.googleProject, runtimeFixture.runtime.clusterName))
-        getResult.isSuccess shouldBe true
-        getResult.get should include("unsupported_browser")
-        getResult.get should not include "ProxyException"
+      withWebDriver { implicit driver =>
+        withNewRStudio(runtimeFixture.runtime) { rstudioPage =>
+//          // See this ticket for adding more comprehensive selenium tests for RStudio:
+//          // https://broadworkbench.atlassian.net/browse/IA-697
+          Thread.sleep(10000)
+          rstudioPage.pressKeys("varA <- 1000")
```
rtitle, line 18: Awesome!! does this work?  :)

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/rstudio/RStudioPage.scala
xuscrbw: I initially copied over the Notebook format because I was running into some issues running the test, but I think it was separate. Removing the extra RStudioMainPage

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/rstudio/RStudioSpec.scala
xuscrbw: Yep, looks like it does work. I do want to see if there's a more precise selector for the global environment table, but Keys does seem to work as intended

xuscrbw: More specifically, press keys works here because the console is already active when we get onto the page. There may be some more complexity in typing into other areas of the screen, but for now this should satisfy what we need

xuscrbw:
jenkins retest

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/rstudio/RStudioSpec.scala
```diff
@@ -1,26 +1,26 @@
...
+          rstudioPage.pressKeys(Keys.ENTER.toString)
+          Thread.sleep(10000)
```
rtitle: Curious if the sleeps are needed.. wonder if there is a way to use `Selenium` to `await` conditions instead of hard-coding sleeps.

rtitle (approved):
This looks great! Approving this PR, we can iterate and add more RStudio tests as needed.

xuscrbw:
jenkins retest

xuscrbw:
jenkins retest

PR #1707, opened 2020-11-17
[IA-2336] Migrate testing to github action

jdcanas:
This seems to be working. 

Note that in the check section `continuous-integration/travis-ci Expected — Waiting for status to be reported`  will remain until I make a global change to the repo, at which time I'll also make the github action check required for merging. I plan to do this after approval.


rtitle:
Wow cool! Does it do code coverage too and attach the report to the PR? (I find it useful occasionally)

.github/workflows/unit_test.yml
```diff
@@ -0,0 +1,48 @@
...
+        JVM_OPTS:  -Xmx3G
+      run: sbt -Denv.type=test clean "test:compile" scalafmtCheckAll
+
+    - name: Run tests
+      env:
+        JAVA_OPTS: -Xmx3G
+        JVM_OPTS:  -Xmx3G
+      run: |
+          sbt -Denv.type=test -Dmysql.host=localhost -Dmysql.port=3307 "project http" coverage test coverageReport
+          sbt -Denv.type=test "project core" coverage test coverageReport
```
Qi77Qi: I would run tests for `core` before `http`

Qi77Qi: one thing I'm not sure how things work is how these codecov reports get published in travis...I'm thinking we'll need to do something for that in github workflow? (at least I think codecov needs some token defined?)

jdcanas:
Hmm... it does run code coverage (the output is on  line ~3000 of the test step... lol)

it may fail the entire step if that fails... 

I'm not sure how to get it to print it pretty. Maybe if I split it into a separate step it'll be more useful...

The annoying thing is that I'm pretty sure we generate the coverage as a function of running the tests themselves though, so it may run the tests twice if we want them to be separate checks, which makes it a bit less useful imo

Lemme dig a bit.

jdcanas:
Hmmm... I think I'll have to use something like this github actions plugin https://github.com/codecov/codecov-action. It seems to do what we want... will need to test a bit.

rtitle (approved):
LGTM when it works :)

jdcanas:
jenkins retest

PR #1711, opened 2020-11-18
IA-2339  delete runtime when cryptomining activity happens

Qi77Qi:
This PR currently only deletes GCE VMs. We also don't delete PDs if they're using PDs, which is fine because we probably don't care so much about supporting cryptomining users well.

https://broadworkbench.atlassian.net/browse/IA-2339

---

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
@@ -4,7 +4,7 @@ package db
 import java.time.Instant
 
 import cats.data.Chain
-import cats.implicits._
+import cats.syntax.all._
 import org.broadinstitute.dsde.workbench.model.{IP, WorkbenchEmail}
 import org.broadinstitute.dsde.workbench.model.google.{
   parseGcsPath,
@@ -237,6 +237,16 @@ object clusterQuery extends TableQuery(new ClusterTable(_)) {
       .filter(_.destroyedDate === destroyedDate)
 
     fullClusterQuery(baseQuery)
+  }
+
+  def clusterRecordQueryByUniqueKey(googleProject: GoogleProject,
+                                    clusterName: RuntimeName,
+                                    destroyedDateOpt: Option[Instant]) = {
+    val destroyedDate = destroyedDateOpt.getOrElse(dummyDate)
+    clusterQuery
+      .filter(_.googleProject === googleProject)
+      .filter(_.clusterName === clusterName)
+      .filter(_.destroyedDate === destroyedDate)
   }
 
   private def fullClusterQueryById(id: Long) =
@@ -474,6 +484,17 @@ object clusterQuery extends TableQuery(new ClusterTable(_)) {
       rid <- findByIdQuery(id).result.head.map[RuntimeConfigId](_.runtimeConfigId): DBIO[RuntimeConfigId]
       _ <- RuntimeConfigQueries.updatePersistentDiskId(rid, None, destroyedDate): DBIO[Int]
     } yield ()
+
+  def markDeleted(googleProject: GoogleProject,
+                  runtimeName: RuntimeName,
+                  destroyedDate: Instant,
+                  deletedFrom: Option[String])(
+    implicit ec: ExecutionContext
+  ): DBIO[Unit] =
+    clusterRecordQueryByUniqueKey(googleProject, runtimeName, Some(dummyDate))
+      .map(c => (c.destroyedDate, c.status, c.hostIp, c.dateAccessed, c.deletedFrom))
+      .update((destroyedDate, RuntimeStatus.Deleted, None, destroyedDate, deletedFrom))
+      .map(_ => ())
 
   def updateDeletedFrom(id: Long, deletedFrom: String): DBIO[Int] =
     findByIdQuery(id)
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/NonLeoMessageSubscriber.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/NonLeoMessageSubscriber.scala
@@ -1,0 +1,144 @@
+package org.broadinstitute.dsde.workbench.leonardo
+package monitor
+
+import cats.effect.{Concurrent, Timer}
+import cats.syntax.all._
+import cats.mtl.Ask
+import fs2.Stream
+import io.chrisdavenport.log4cats.StructuredLogger
+import io.circe.Decoder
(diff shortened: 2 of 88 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/NonLeoMessageSubscriber.scala
```diff
@@ -0,0 +1,144 @@
...
+import org.broadinstitute.dsde.workbench.model.google.GoogleProject
+import org.broadinstitute.dsde.workbench.openTelemetry.OpenTelemetryMetrics
+
+import scala.concurrent.ExecutionContext.Implicits.global
+
+/**
+ * This is subscriber for messages that are not published by Leonardo itself.
+ * But rather from cron jobs, or cloud logging sink (triggered from crypto-mining activity)
+ */
+class NonLeoMessageSubscriber[F[_]: Timer](gkeAlg: GKEAlgebra[F],
```
rtitle, line 26: I like the idea of splitting up the `*MessageSubscriber` classes to Leo and non-Leo

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/NonLeoMessageSubscriber.scala
```diff
@@ -0,0 +1,144 @@
...
+        F.unit
+      else
+        for {
+          runtimeFromGoogle <- computeService.getInstance(msg.resource.labels.projectId,
+                                                          msg.resource.labels.zone,
+                                                          InstanceName(msg.resource.labels.instanceId.toString))
+          // We mark the runtime as Deleted, and delete the instance.
+          // If instance deletion fails for some reason, it will be cleaned up by resource-validator
+          _ <- runtimeFromGoogle.traverse { instance =>
+            for {
```
rtitle, line 99: Can we update a metric too? So we can make a stackdriver dashboard for this

rtitle:
At a glance looks good, seems like it'll get the job done :)

We may want to further enhance it to send the runtime `creator` email somewhere so we can exclude these users from data warehouse/mixpanel.. but assuming we can iterate as needed.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/NonLeoMessageSubscriber.scala
Qi77Qi: there's https://github.com/DataBiosphere/leonardo/pull/1711/files/ed6b2882d121c6b0f2f4f1f25764b2c5c3058eb8#diff-1ea743458d40ca65f6c9a7f7d9178ded5967d287a3113158f4e9f713b01d930bR43.... are you thinking something more?

rtitle: Oh missed that.. that's probably fine then

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/NonLeoMessageSubscriber.scala
Qi77Qi: yea..for leo messages, we have to define both encoder and decoder for the messages. but we shoudlnt' need to define encoder for non leo messages

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/NonLeoMessageSubscriber.scala
```diff
@@ -0,0 +1,144 @@
...
+  val process: Stream[F, Unit] = subscriber.messages
+    .parEvalMapUnordered(10)(messageHandler)
+    .handleErrorWith(error => Stream.eval(logger.error(error)("Failed to initialize message processor")))
+
+  private[monitor] def messageHandler(event: Event[NonLeoMessage]): F[Unit] = {
+    val traceId = event.traceId.getOrElse(TraceId("None"))
+    for {
+      now <- nowInstant[F]
+      implicit0(ev: Ask[F, AppContext]) <- F.pure(Ask.const[F, AppContext](AppContext(traceId, now, None)))
+      _ <- metrics.incrementCounter(event.msg.messageType)
```
rtitle: Could we name it like `s"LeoPubSub/${event.msg.messageType}"`? Just to distinguish it when making dashboards.

Qi77Qi:
@rtitle since we have `deletedFrom` field now, we'll be able to know who did these afterwards..I'm thinking punt on BQ in this PR since I'm not sure permission wise if things will work, and this PR already has some permission thing that we need to make sure when it's rolled out to prod, so feels safer to separate the 2.

Qi77Qi:
jenkins retest...seems a flaky unit test

Qi77Qi:
fiab start  failed....jenkins retest

Qi77Qi:
test failed because `leoDeadLetterTopic` doesn't exist in broard-dsde-qa...added now..jenkins retest

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
```diff
@@ -239,6 +239,16 @@ object clusterQuery extends TableQuery(new ClusterTable(_)) {
     fullClusterQuery(baseQuery)
   }
 
+  def fullClusterRecordQueryByUniqueKey(googleProject: GoogleProject,
```
rtitle: MInor: we usually name these methods `full*` when they do all the joins… maybe name this `clusterRecordQueryByUniqueKey`?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -514,8 +514,13 @@ class GKEInterpreter[F[_]: Parallel: ContextShift: Timer](
       )
 
       // delete the namespace only after the helm uninstall completes
-      _ <- kubeService.deleteNamespace(dbApp.cluster.getGkeClusterId,
-                                       KubernetesNamespace(dbApp.app.appResources.namespace.name))
+      _ <- kubeService
+        .deleteNamespace(dbApp.cluster.getGkeClusterId, KubernetesNamespace(dbApp.app.appResources.namespace.name))
+        .recoverWith {
```
rtitle: Is this recovery in wb-libs now, can it be removed here?

rtitle (approved):

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
Qi77Qi: ah yes...forgot to clean up here

Qi77Qi:
jenkins retest

PR #1700, opened 2020-11-12
[IA-2349] add new crypto package

rtitle:
JIRA: https://broadworkbench.atlassian.net/browse/IA-2349

Adds this sidecar container to Leo-created runtimes (GCE and Dataproc): https://github.com/broadinstitute/terra-cryptomining-security-alerts/tree/master/v2

The container logs a message to the project's Stackdriver when a mining event is detected. Currently no action is taken, but when https://broadworkbench.atlassian.net/browse/IA-2339 is done we will update Leo to handle these messages and shutdown the VM.

Automation tests pass, plus did some manual testing in a fiab.

---
Have you read [CONTRIBUTING.md](https://github.com/DataBiosphere/leonardo/blob/develop/CONTRIBUTING.md) lately? If not, do that first.

I, the developer opening this PR, do solemnly pinky swear that:

- [ ] I've documented my API changes in Swagger

In all cases:

- [ ] Get a thumbsworth of review and PO signoff if necessary
- [ ] Verify all tests go green
- [ ] Squash and merge; Delete your branch after this
- [ ] Test this change deployed correctly and works on dev environment after deployment

PR #1712, opened 2020-11-18
[IA-2349] add new crypto package - take 2

rtitle:
Same as https://github.com/DataBiosphere/leonardo/pull/1700 but with one additional commit (https://github.com/DataBiosphere/leonardo/commit/502299607a6d0332792c0d02485914292a65a8df).

Since AoU does not allow GCR images, we need to push the crypto-detector somewhere else. Possibilities include quay.io or ACR. However for now I'm just disabling it for AoU, since cryptomining doesn't seem to happen there anyway.

---

In all cases:


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeServiceInterp.scala
```diff
@@ -542,7 +542,12 @@ class RuntimeServiceInterp[F[_]: Parallel](config: RuntimeServiceConfig,
       )
       // Get the proxy image
       proxyImage = RuntimeImage(Proxy, config.imageConfig.proxyImage.imageUrl, now)
-    } yield Set(toolImage, welderImage, proxyImage)
+      // Crypto detector image - note it's not currently supported on Dockerhub
```
rtitle, line 545: This is where we're disabling crypto-detector for AoU

Qi77Qi (approved):

rtitle:
Just realized I'm probably going to need to regenerate the custom image. However I want other PRs to make it through first. Adding DO NOT MERGE label for now

rtitle:
Tested on a fiab, verified `cryptomining-detector` is started when the registry is GCR but not when it's Dockerhub. Merging.

Qi77Qi:
cool..I'll test this out today

PR #1713, opened 2020-11-18
[no ticket]Fix docker spec due to deleted image

rtitle:
The `broadinstitute/leonardo-notebooks` image was deleted which broke this test. :)

It's probably good to test autodetection against a real DH aou image, although if we migrate off DH we might need to change this test again.

---

In all cases:


Qi77Qi (approved):
thanks for looking into this!

rtitle:
I think I'm going to merge this without Travis -- unit + auto tests passed in Jenkins

PR #1721, opened 2020-11-20
IA-2341 Galaxy stop/start

rtitle:
https://broadworkbench.atlassian.net/browse/IA-2341

This PR adds new `stop` and `start` endpoints for apps, and associated PubSub handling and GKE operations. Note this PR does NOT include autostop detection for Galaxy. Will likely create a separate for PR the autostop piece.

This PR also adds locking functionality to serialize nodepool operations within a cluster. This allows us to remove the "fail-fast" semantics in front Leo and handle queued nodepool operations reliably in back Leo.

Here is a [diagram](https://lucid.app/lucidchart/177a0d3e-63db-4271-90ba-d80eb4609c67/edit?beaconFlowId=68ECC8F6AFB84971&page=0_0#?folder_id=home&browser=icon) of the stop/start flow for Kubernetes apps:

image

Dependent PRs:
* https://github.com/broadinstitute/workbench-libs/pull/374
* https://github.com/broadinstitute/sam/pull/485

Testing:
- Unit tests pass
- Manually tested on fiab
- Existing automation tests pass
- TODO: add new automation tests (will do in a separate PR)

---

In all cases:

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/KeyLock.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/KeyLock.scala
@@ -5,7 +5,7 @@ import java.util.concurrent.TimeUnit
 import cats.effect.concurrent.Semaphore
 import cats.effect.implicits._
 import cats.effect.{Blocker, ConcurrentEffect, ContextShift}
-import cats.implicits._
+import cats.syntax.all._
 import com.google.common.cache.{CacheBuilder, CacheLoader}
 
 import scala.concurrent.duration.FiniteDuration
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/KeyLock.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/KeyLock.scala
@@ -4,28 +4,32 @@ import java.util.concurrent.TimeUnit
 
 import cats.effect.concurrent.Semaphore
 import cats.effect.implicits._
-import cats.effect.{Blocker, Concurrent, ConcurrentEffect, ContextShift, Effect}
+import cats.effect.{Blocker, ConcurrentEffect, ContextShift}
 import cats.implicits._
 import com.google.common.cache.{CacheBuilder, CacheLoader}
 
 import scala.concurrent.duration.FiniteDuration
 
+/**
+ * A functional lock which works on a per-key basis.
+ */
 abstract class KeyLock[F[_], K] {
   def acquire(key: K): F[Unit]
   def release(key: K): F[Unit]
   def withKeyLock[A](key: K)(fa: F[A]): F[A]
 }
 object KeyLock {
-  def apply[F[_]: ConcurrentEffect: ContextShift, K <: AnyRef](expiryTime: FiniteDuration,
-                                                               maxSize: Int,
-                                                               blocker: Blocker): F[KeyLock[F, K]] =
-    Effect[F].delay(new KeyLockImpl(expiryTime, maxSize, blocker))
+  def apply[F[_]: ContextShift, K <: AnyRef](expiryTime: FiniteDuration, maxSize: Int, blocker: Blocker)(
+    implicit F: ConcurrentEffect[F]
+  ): F[KeyLock[F, K]] =
+    F.delay(new KeyLockImpl(expiryTime, maxSize, blocker))
 
   // A KeyLock implementation backed by a guava cache
-  final private class KeyLockImpl[F[_]: ConcurrentEffect: ContextShift, K <: AnyRef](expiryTime: FiniteDuration,
-                                                                                     maxSize: Int,
-                                                                                     blocker: Blocker)
-      extends KeyLock[F, K] {
+  final private class KeyLockImpl[F[_]: ContextShift, K <: AnyRef](expiryTime: FiniteDuration,
+                                                                   maxSize: Int,
+                                                                   blocker: Blocker)(
+    implicit F: ConcurrentEffect[F]
+  ) extends KeyLock[F, K] {
 
     private val cache = CacheBuilder
       .newBuilder()
@@ -41,19 +45,19 @@ object KeyLock {
 
     override def acquire(key: K): F[Unit] =
       for {
-        lock <- blocker.blockOn(Concurrent[F].delay(cache.get(key)))
+        lock <- blocker.blockOn(F.delay(cache.get(key)))
         _ <- lock.acquire
       } yield ()
 
     override def release(key: K): F[Unit] =
       for {
-        lock <- blocker.blockOn(Concurrent[F].delay(cache.get(key)))
+        lock <- blocker.blockOn(F.delay(cache.get(key)))
         _ <- lock.release
       } yield ()
 
     override def withKeyLock[A](key: K)(fa: F[A]): F[A] =
       for {
-        lock <- blocker.blockOn(Concurrent[F].delay(cache.get(key)))
+        lock <- blocker.blockOn(F.delay(cache.get(key)))
         res <- lock.withPermit(fa)
       } yield res
   }
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/KeyLock.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/KeyLock.scala
@@ -2,7 +2,7 @@ package org.broadinstitute.dsde.workbench.leonardo
 
 import java.util.concurrent.TimeUnit
 
-import cats.effect.concurrent.MVar
+import cats.effect.concurrent.{MVar, MVar2}
 import cats.effect.implicits._
 import cats.effect.{Blocker, Concurrent, ConcurrentEffect, ContextShift, Effect}
 import cats.implicits._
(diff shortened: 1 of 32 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -713,22 +717,21 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift: Parallel](
     for {
       ctx <- ev.ask
 
-      // The "create app" flow does a number of things:
```
rtitle, line 714: I decided to change the nodepool creation to be asynchronous. The cluster creation is still synchronous to the message. The reason is that nodepool operations may potentially have long delays due to locks, so probably better to be async. Also the locking code is cleaner if `createNodepool` and `pollNodepool` aren't split up as separate methods.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/KubernetesServiceInterp.scala
```diff
@@ -108,9 +87,6 @@ final class LeoKubernetesServiceInterp[F[_]: Parallel](
 
       clusterId = saveClusterResult.minimalCluster.id
 
-      hasOperationInProgress <- KubernetesServiceDbQueries.hasClusterOperationInProgress(clusterId).transaction
```
rtitle, line 111: Removed "fail-fast" `hasClusterOperationInProgress` functionality because we have nodepool locking now.

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/KeyLock.scala
```diff
@@ -0,0 +1,79 @@
...
+  def release: F[Unit]
+  def withLock[A](fa: F[A]): F[A]
+}
+object Lock {
+  def apply[F[_]: Concurrent]: F[Lock[F]] =
+    MVar.of(()).map(l => new LockImpl(l))
+
+  // A Lock implementation backed by an MVar
+  // Inspired from https://typelevel.org/cats-effect/concurrency/mvar.html#use-case-asynchronous-lock-binary-semaphore-mutex
+  final private class LockImpl[F[_]: Concurrent](lock: MVar2[F, Unit]) extends Lock[F] {
```
Qi77Qi: with Semaphore, you can have permit to require and release, is that not sufficient?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -454,33 +446,38 @@ class GKEInterpreter[F[_]: Parallel: ContextShift: Timer](
...
+                )
             )
-          else
-            logger.error(
-              s"Delete nodepool operation has failed or timed out for nodepool ${params.nodepoolId} | trace id: ${ctx.traceId}"
-            ) >>
-              F.raiseError[Unit](NodepoolDeletionException(params.nodepoolId))
+            .compile
+            .lastOrError
+          _ <- lastOp match {
```
Qi77Qi: any reason not using `traverse`?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -527,7 +524,187 @@ class GKEInterpreter[F[_]: Parallel: ContextShift: Timer](
...
+
+      _ <- logger.info(
+        s"Stopping app ${dbApp.app.appName.value} in cluster ${dbCluster.getGkeClusterId.toString} | trace id: ${ctx.traceId}"
+      )
+
+      _ <- nodepoolQuery.updateStatus(dbNodepool.id, NodepoolStatus.Provisioning).transaction
+
+      // If autoscaling is enabled, disable it first
+      _ <- if (dbNodepool.autoscalingEnabled) {
+        nodepoolLock.withKeyLock(dbCluster.getGkeClusterId) {
```
Qi77Qi, line 539: I think nodepoolLock can be `Ref[IO, Map[gkeClusterId, Semaphore[IO]]]`, and this can be 
```
permitOpt <- nodepoolLock.get(_.get(gkeClusterId))
permit <- permitOpt.fold {
        Semaphore[IO](1L).flatMap(s => nodepoolLock.modify(mp => (mp + (gkeClusterId -> s), s))) //make a new lock if not exist already
      }(p => IO.pure(p))
lock = Resource.make[IO, Unit](permit.acquire)(_ => permit.release)
_ <- lock.use {_ => ???}
```

This way we don't need the custom classes and mvar etc

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/KeyLock.scala
rtitle: Thinking why I didn't use `Semaphore`. :) I think since we just want a mutex then `Semaphore` is a bit too powerful. But yea `Semaphore(1)` would be equivalent -- could try to use that.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
rtitle: Prob not.. copy pasted code

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
rtitle: I think I tried `Ref[IO, Map[..]]` and had some issues.. but your code seems correct. I can experiment. In any case I think we'd want some helper code to remove boilerplate and make the lock easy to use.

Another reason I kinda like using guava cache is it gives us lock timeouts, to prevent potential rogue processes from holding locks forever.

Qi77Qi: ah yea, that makes sense....

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -713,22 +717,21 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift: Parallel](
...
-      //  1. create, poll, and setup cluster if it doesn't exist
-      //  2. create and poll nodepool if it doesn't exist
-      //  3. create and poll disk if it doesn't exist
-      //  4. create and poll app
+      // The "create app" flow potentially does a number of things:
+      //  1. creates a Kubernetes cluster
```
kyuksel: Doesn't the `if it doesn't exist` part still apply?

rtitle: Yeah, will add it back

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/KeyLock.scala
```diff
@@ -0,0 +1,64 @@
...
+import cats.implicits._
+import com.google.common.cache.{CacheBuilder, CacheLoader}
+
+import scala.concurrent.duration.FiniteDuration
+
+/**
+ * A functional lock which works on a per-key basis.
+ */
+abstract class KeyLock[F[_], K] {
+  def acquire(key: K): F[Unit]
```
Qi77Qi: does `acquire` and `release` still needed as public interface?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/leoPubsubMessageSubscriberModels.scala
```diff
@@ -254,6 +262,16 @@ object LeoPubsubMessage {
     val messageType: LeoPubsubMessageType = LeoPubsubMessageType.BatchNodepoolCreate
   }
 
+  final case class StopAppMessage(appId: AppId, appName: AppName, project: GoogleProject, traceId: Option[TraceId])
```
Qi77Qi, line 254: I think `traceId` will always exist

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/KeyLock.scala
rtitle: I guess not, we only use `withKeyLock`. I could remove acquire/release

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/leoPubsubMessageSubscriberModels.scala
rtitle: actually it's optional in the base trait:
```
sealed trait LeoPubsubMessage {
  def traceId: Option[TraceId]
  def messageType: LeoPubsubMessageType
}
```

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/KeyLock.scala
```diff
@@ -0,0 +1,50 @@
...
+import cats.effect.{Blocker, ConcurrentEffect, ContextShift}
+import cats.syntax.all._
+import com.google.common.cache.{CacheBuilder, CacheLoader}
+
+import scala.concurrent.duration.FiniteDuration
+
+/**
+ * A functional lock which works on a per-key basis.
+ */
+abstract class KeyLock[F[_], K] {
```
Qi77Qi, line 16: not necessarily in this PR, but maybe it's worth moving this to wb-libs :)

Qi77Qi (approved):
:shipit:

PR #1727, opened 2020-12-01
[IA-2385] Support public GHCR images

rtitle:
See https://broadworkbench.atlassian.net/browse/IA-2385

Tested with `ghcr.io/lucidtronix/ml4h/ml4h_terra:20201117_123026`, it seems to work:

image

(Verified `ghcr.io/broadinstitute/ml4h/ml4h_terra:20201119_180431` too)

---

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/BaseRuntimeInterpreter.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/BaseRuntimeInterpreter.scala
@@ -126,7 +126,8 @@ abstract private[util] class BaseRuntimeInterpreter[F[_]: Async: ContextShift: L
           .toRight(new Exception(s"Unable to update welder because current welder image is not available"))
           .flatMap(x =>
             x.registry match {
-              case Some(ContainerRegistry.GCR)       => Right(config.imageConfig.welderGcrImage.imageUrl)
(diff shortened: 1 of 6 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/BaseRuntimeInterpreter.scala
```diff
@@ -126,7 +126,8 @@ abstract private[util] class BaseRuntimeInterpreter[F[_]: Async: ContextShift: L
           .toRight(new Exception(s"Unable to update welder because current welder image is not available"))
           .flatMap(x =>
             x.registry match {
-              case Some(ContainerRegistry.GCR)       => Right(config.imageConfig.welderGcrImage.imageUrl)
+              case Some(ContainerRegistry.GCR) | Some(ContainerRegistry.GHCR) =>
```
Qi77Qi, line 129: is this for potentially using welder in ghcr in the future?

Qi77Qi (approved):

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/BaseRuntimeInterpreter.scala
rtitle: Don't think we have a reason to - was just making the match exhaustive. So a Dockerhub container will cause us to use Dockerhub welder, else we'll use the GCR welder.

PR #1763, opened 2020-12-02
IA-2393 config remote properly

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-2393

https://github.com/broadinstitute/dsp-jenkins/pull/471 will add scalaSteward as a remote as well. This PR assumes that if the branch starts with `update/`, it's a PR from `scalaSteward` (a bit hacky, but not sure if there's a better way).

---

In all cases:

### DIFF
--- a/docker/build.sh
+++ b/docker/build.sh
@@ -40,6 +40,7 @@ set -e
 TARGET="${TARGET:-leonardo}"
 DB_CONTAINER="leonardo-mysql"
 GIT_BRANCH="${BRANCH:-$(git rev-parse --abbrev-ref HEAD)}"
+REMOTE=$(if [[ ${GIT_BRANCH} == "update/"* ]]; then echo "scalaSteward"; else echo "origin"; fi)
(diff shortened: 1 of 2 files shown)


docker/build.sh
```diff
@@ -40,6 +40,7 @@ set -e
 TARGET="${TARGET:-leonardo}"
 DB_CONTAINER="leonardo-mysql"
 GIT_BRANCH="${BRANCH:-$(git rev-parse --abbrev-ref HEAD)}"
+REMOTE=$(if [[ ${GIT_BRANCH} == "update/"* ]]; then echo "scalaSteward"; else echo "origin"; fi)
```
rtitle, line 43: I guess this would break non-scalasteward branches that start with `update/`. That's probably fine, just noting it :)

rtitle (approved):
LGTM

docker/build.sh
Qi77Qi: yea...this is assuming no one names their branch `update/xxx`

PR #1766, opened 2020-12-02
(IA-2149) Make Runtime Labels Mutable in Patch endpoint

ehanna4:
Allows for updating the labels on a runtime. Able to add or update a label or delete an existing one. Should not be able to change the default labels.

Added unit tests for the decoder and the for updateRuntimeRequest and updated Swagger

---

- [ X] I've documented my API changes in Swagger

In all cases:


automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/runtimes/RuntimePatchSpec.scala
```diff
@@ -48,7 +48,9 @@ class RuntimePatchSpec extends FixtureAnyFreeSpec with LeonardoTestUtils with Le
                                                     ),
                                                     true,
                                                     None,
-                                                    None)
+                                                    None,
+                                                    Map("" -> ""),
+                                                    List(""))
```
Qi77Qi: if you just need empty Map and empty List, you can do `Map.empty`, and `List.empty`

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/runtimeModels.scala
```diff
@@ -361,6 +361,16 @@ case class DefaultRuntimeLabels(runtimeName: RuntimeName,
     ).filterNot(_._2 == null)
 }
 
+object defaultLabelKeys {
```
Qi77Qi: if this is used for unit tests, I'd move it to https://github.com/DataBiosphere/leonardo/blob/develop/http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/CommonTestData.scala

Qi77Qi (commented):
lgtm mostly...did you test it in a fiab?

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/runtimeRoutesModels.scala
```diff
@@ -81,4 +81,6 @@ object UpdateRuntimeConfigRequest {
 final case class UpdateRuntimeRequest(updatedRuntimeConfig: Option[UpdateRuntimeConfigRequest],
                                       allowStop: Boolean,
                                       updateAutopauseEnabled: Option[Boolean],
-                                      updateAutopauseThreshold: Option[FiniteDuration])
+                                      updateAutopauseThreshold: Option[FiniteDuration],
+                                      labelsToUpsert: LabelMap,
+                                      labelsToDelete: List[String])
```
rtitle: Wonder if this should be a `Set[String]`

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/runtimeModels.scala
rtitle: I think this is intended to be a companion object, so should be `object DefaultRuntimeLabels {..}`

The intent is to return the default label _keys_. Could make that more explicit by:
```
object DefaultRuntimeLabels {
  // Creating a dummy instance to obtain the default label keys
  val defaultLabelKeys = DefaultRuntimeLabels(RuntimeName(""),
                                              GoogleProject(""),
                                              WorkbenchEmail(""),
                                              WorkbenchEmail(""),
                                              None,
                                              None,
                                              None).toMap.keySet
}
```

Unless anyone has a better way..

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -2013,6 +2013,8 @@ components:
               cloudService: gce
               machineType: n1-highmem-16
               diskSize: 500
+            labelsToUpsert: {new_label : value, old_label : new_value}
```
rtitle, line 2103: Is this valid yaml syntax? I usually run any swagger change through https://editor.swagger.io/

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/RuntimeRoutes.scala
```diff
@@ -378,7 +378,16 @@ object RuntimeRoutes {
       as <- x.downField("allowStop").as[Option[Boolean]]
       ap <- x.downField("autopause").as[Option[Boolean]]
       at <- x.downField("autopauseThreshold").as[Option[Int]]
-    } yield UpdateRuntimeRequest(rc, as.getOrElse(false), ap, at.map(_.minutes))
+      ul <- x.downField("labelsToUpsert").as[Option[LabelMap]]
+      dl <- x.downField("labelsToDelete").as[Option[List[String]]]
+    } yield {
+      val labelsToDelete =
```
rtitle: Could this be written as:
```
val labelsToDelete = dl.getOrElse(Set.empty).removedAll(defaultLabelKeys.defaultLabels.keySet)
val labelsToUpdate = ul.getOrElse(Map.empty).removedAll(defaultLabelKeys.defaultLabels.keySet).filter(_._2.nonEmpty)
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeServiceInterp.scala
```diff
@@ -503,6 +503,19 @@ class RuntimeServiceInterp[F[_]: Parallel](config: RuntimeServiceConfig,
       _ <- if (updatedAutopauseThreshold != runtime.autopauseThreshold)
         clusterQuery.updateAutopauseThreshold(runtime.id, updatedAutopauseThreshold, ctx.now).transaction.void
       else Async[F].unit
+
+      // Deleting labels
+      existingLabels <- labelQuery.getAllForResource(runtime.id, LabelResourceType.runtime).transaction
```
rtitle: Looks like `existingLabels` is not used, can we remove this?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeServiceInterp.scala
```diff
@@ -503,6 +503,19 @@ class RuntimeServiceInterp[F[_]: Parallel](config: RuntimeServiceConfig,
...
+
+      _ <- req.labelsToDelete.traverse { label =>
```
rtitle: It might be better to do all the label operations (upserts and deletes) in 1 transaction if possible.

For example we could define deletion by batch in `LabelComponent`:
```
  def deleteForResource(resourceId: Long, resourceType: LabelResourceType, keys: Set[String])(
    ): DBIO[Int] =
    labelQuery
      .filter(_.resourceId === resourceId)
      .filter(_.resourceType === resourceType)
      .filter(_.key inSetBind keys)
      .delete
```

(I think upserts still need to be done one by one.)

Then here we could do:
```
      dbOp = DBIO.seq(
        labelQuery.deleteForResource(runtime.id, LabelResourceType.runtime, req.labelsToDelete),
        req.labelsToUpsert.toList.traverse_ {
          case (k, v) =>
            labelQuery.save(runtime.id, LabelResourceType.runtime, k, v)
        }
      )
      _ <- dbOp.transaction
```

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/api/HttpRoutesSpec.scala
```diff
@@ -226,7 +226,9 @@ class HttpRoutesSpec
       Some(UpdateRuntimeConfigRequest.GceConfig(Some(MachineTypeName("n1-micro-2")), Some(DiskSize(50)))),
       true,
       Some(true),
-      Some(5.minutes)
+      Some(5.minutes),
+      Map("" -> ""),
```
rtitle: Can do `Map.empty` and `List.empty`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/RuntimeRoutes.scala
rtitle: Actually -- I'm thinking it might be better to return `BadRequest` with an error message to the caller if we receive any invalid labels. The caller is probably doing something wrong in this case -- better to fail explicitly than silently ignore the bad data.

http/src/main/resources/swagger/api-docs.yaml
rtitle: Looks like the swagger validator is happy 👍

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/runtimeRoutesModels.scala
ehanna4: oh actually yeah that seems better👍

http/src/main/resources/swagger/api-docs.yaml
ehanna4: yeah I've been using that validator check my changes

Qi77Qi (approved):
lgtm 👍

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/api/RuntimeRoutes.scala
ehanna4: Please see the latest commit - I have updated so that the decoder will now return a `Error: Bad Request` rather than filtering, as you suggested

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodec.scala
```diff
@@ -42,6 +42,9 @@ object JsonCodec {
     "num nodes must be >= 1",
     List.empty
   )
+  val deleteDefaultLabelsDecodingFailure = DecodingFailure("Default labels are not allowed to be deleted", List.empty)
```
rtitle, line 45: 👍 , thanks for adding this

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeServiceInterp.scala
```diff
@@ -503,6 +503,17 @@ class RuntimeServiceInterp[F[_]: Parallel](config: RuntimeServiceConfig,
       _ <- if (updatedAutopauseThreshold != runtime.autopauseThreshold)
         clusterQuery.updateAutopauseThreshold(runtime.id, updatedAutopauseThreshold, ctx.now).transaction.void
       else Async[F].unit
+
+      // Deleting labels
```
rtitle: I think I had a comment about doing the label operations in the same DB transaction -- did that work?

rtitle:
Decoding logic looks good, one comment in `RuntimeServiceInterp`. Also seems like there is a compilation error.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeServiceInterp.scala
ehanna4: Still working on that - I should have called my previous commit "PR feedback 1"

rtitle (approved):
:shipit: ! Nice work

PR #1779, opened 2020-12-07
[no ticket] Fix things needed by DICOM viewer

rtitle:
Just fixing a couple minor issues found when trying to launch the [DICOM viewer](https://github.com/helxplatform/cloudtop-ohif/pull/2) in Terra. Comments inline.

---

In all cases:

### DIFF
--- a/http/src/main/resources/init-resources/rstudio-docker-compose.yaml
+++ b/http/src/main/resources/init-resources/rstudio-docker-compose.yaml
@@ -22,6 +22,7 @@ services:
       - /etc/rstudio-license-file.lic:/etc/rstudio-server/license.lic:ro
     env_file:
       - /etc/google_application_credentials.env
+      - /etc/custom_env_vars.env
     # See https://docs.docker.com/engine/reference/run/#user-memory-constraints
     mem_limit: ${MEM_LIMIT}   # hard limit on memory consumption by the container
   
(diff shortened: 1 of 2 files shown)


http/src/main/resources/init-resources/rstudio-docker-compose.yaml
```diff
@@ -22,6 +22,7 @@ services:
       - /etc/rstudio-license-file.lic:/etc/rstudio-server/license.lic:ro
     env_file:
       - /etc/google_application_credentials.env
+      - /etc/custom_env_vars.env
```
rtitle, line 25: Needed to support custom EVs in the RStudio image. This already exists in the Jupyter image, was just missed here I think.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/ProxyService.scala
```diff
@@ -467,6 +467,7 @@ class ProxyService(
     "Sec-WebSocket-Version",
     "Sec-WebSocket-Key",
     "Sec-WebSocket-Extensions",
+    "Sec-WebSocket-Protocol",
```
rtitle, line 470: Need to filter these `Sec-WebSocket-*` headers otherwise they end up getting set twice and the browser complains.

jdcanas (approved):

Qi77Qi (approved):

PR #1781, opened 2020-12-07
[IA-2350] Upgrade to latest workbench-libs DataprocService

kyuksel:
Pulling in the workbench-libs changes ([Ability to resize and stop Dataproc clusters](https://github.com/broadinstitute/workbench-libs/pull/373)) which included a breaking change in `GoogleDataprocService` constructor.

Some of the diff's are due to `scalafmt` formatting differently than before for some reason 🤷‍♂️

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodec.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodec.scala
@@ -77,16 +77,15 @@ object JsonCodec {
     "numberOfWorkerLocalSSDs",
     "numberOfPreemptibleWorkers",
     "cloudService"
-  )(
-    x =>
-      (x.numberOfWorkers,
-       x.machineType,
-       x.diskSize,
-       x.workerMachineType,
-       x.workerDiskSize,
-       x.numberOfWorkerLocalSSDs,
-       x.numberOfPreemptibleWorkers,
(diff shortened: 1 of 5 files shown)


core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodec.scala
```diff
@@ -194,8 +198,8 @@ object JsonCodec {
   implicit val networkFieldsEncoder: Encoder[NetworkFields] =
     Encoder.forProduct3("networkName", "subNetworkName", "subNetworkIpRange")(x => NetworkFields.unapply(x).get)
   implicit val kubeAsyncFieldEncoder: Encoder[KubernetesClusterAsyncFields] =
-    Encoder.forProduct3("loadBalancerIp", "apiServerIp", "networkInfo")(x =>
-      KubernetesClusterAsyncFields.unapply(x).get
+    Encoder.forProduct3("loadBalancerIp", "apiServerIp", "networkInfo")(
```
Qi77Qi: are these from scalafmt?

kyuksel: I thought I'd run scalafmt but I hadn't. Thanks for pointing it out!

Qi77Qi (approved):

PR #1782, opened 2020-12-07
IA-2369 publish subjectId

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-2369

Publishes
```
{   "userSubjectId": "???" }
```
To leonardo-pubsub when cryptomining activity is detected

Tested in fiab:

Screen Shot 2020-12-10 at 12 28 37 PM

---

In all cases:

### DIFF
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -552,8 +552,11 @@ pubsub {
   }
 
   non-leo-message-subscriber {
-    subscription-name = "nonLeoMessageSubscription"
+    # subscription-name = "nonLeoMessageSubscription"
     dead-letter-topic = "leoDeadLetterTopic"
+    # Topic that we publish cryptomining users to.
+    # Bard and Sam will subscribe to this topic and act upon it accordingly
+    terra-cryptomining-topic = "terra-cryptomining"
   }
 }
 
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/NonLeoMessageSubscriber.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/NonLeoMessageSubscriber.scala
@@ -6,9 +6,17 @@ import cats.syntax.all._
 import cats.mtl.Ask
 import fs2.Stream
 import io.chrisdavenport.log4cats.StructuredLogger
-import io.circe.Decoder
-import org.broadinstitute.dsde.workbench.google2.{Event, GoogleComputeService, GoogleSubscriber, InstanceName, ZoneName}
+import io.circe.{Decoder, DecodingFailure, Encoder}
+import org.broadinstitute.dsde.workbench.google2.{
+  Event,
+  GoogleComputeService,
+  GooglePublisher,
+  GoogleSubscriber,
+  InstanceName,
+  ZoneName
+}
 import org.broadinstitute.dsde.workbench.leonardo.JsonCodec._
+import org.broadinstitute.dsde.workbench.leonardo.dao.{SamDAO, UserSubjectId}
 import org.broadinstitute.dsde.workbench.leonardo.db.{clusterQuery, kubernetesClusterQuery, DbReference}
 import org.broadinstitute.dsde.workbench.leonardo.http.dbioToIO
 import org.broadinstitute.dsde.workbench.leonardo.monitor.NonLeoMessage.{CryptoMining, DeleteKubernetesClusterMessage}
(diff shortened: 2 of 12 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/NonLeoMessageSubscriber.scala
```diff
@@ -90,22 +104,38 @@ class NonLeoMessageSubscriber[F[_]: Timer](gkeAlg: GKEAlgebra[F],
...
+                  _ <- userSubjectId.traverse { sid =>
+                    val byteString = ByteString.copyFromUtf8(CryptominingUserMessage(sid).asJson.noSpaces)
+
+                    val message = PubsubMessage
+                      .newBuilder()
+                      .setData(byteString)
+                      .putAttributes("traceId", ctx.traceId.asString)
+                      .putAttributes("cryptomining", "true")
+                      .build()
+                    publisher.publishNativeOne(message)
```
rtitle, line 135: So this will publish to the `leonardo-pubsub` topic, right? Will it cause any issues in `LeoPubsubMessageSubscriber` to receive these messages which it doesn't know how to handle?

Also will it ok for bard/etc to receive messages from `leonardo-pubsub`? I guess the subscription can filter based on the `cryptomining=true` attribute.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/NonLeoMessageSubscriber.scala
```diff
@@ -117,15 +147,27 @@ object NonLeoMessageSubscriber {
...
   implicit val googleResourceDecoder: Decoder[GoogleResource] = Decoder.forProduct1("labels")(GoogleResource.apply)
-  Decoder.forProduct2("textPayload", "resource")(CryptoMining.apply)
-  implicit val cryptoMiningDecoder: Decoder[CryptoMining] =
-    Decoder.forProduct2("textPayload", "resource")(CryptoMining.apply)
+  implicit val cryptoMiningDecoder: Decoder[CryptoMining] = Decoder.instance { c =>
+    for {
+      textpayload <- c.downField("textPayload").as[String]
+      resource <- c.downField("resource").as[GoogleResource]
+      logName <- c.downField("logName").as[String]
+      splitted = logName.split("/")
```
rtitle, line 161: What is the `logName` format? Maybe could add a comment

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/NonLeoMessageSubscriber.scala
Qi77Qi: yea, that was my original thought...But people preferred we publish to a dedicated topic, so I'm making it a separate topic (didn't realize the change wasn't pushed Tues)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/NonLeoMessageSubscriber.scala
Qi77Qi: looks like `  "logName": "projects/general-dev-billing-account/logs/cryptomining",`, I'll add a comment

http/src/main/resources/reference.conf
```diff
@@ -552,8 +552,11 @@ pubsub {
   }
 
   non-leo-message-subscriber {
-    subscription-name = "nonLeoMessageSubscription"
+    # subscription-name = "nonLeoMessageSubscription"
```
rtitle, line 555: Is this intended to be commented?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/NonLeoMessageSubscriber.scala
rtitle: Cool.. separate topic sounds good 👍

http/src/main/resources/reference.conf
```diff
@@ -552,8 +552,11 @@ pubsub {
...
     dead-letter-topic = "leoDeadLetterTopic"
+    # Topic that we publish cryptomining users to.
+    # Bard and Sam will subscribe to this topic and act upon it accordingly
+    terra-cryptomining-topic = "terra-cryptomining"
```
rtitle, line 559: Was wondering if we'd need a topic-per-fiab like we do for other topics, but it probably doesn't matter here because we won't have automation tests that do cryptomining.

http/src/main/resources/reference.conf
rtitle: Ah guessing so because it's specified in firecloud-develop

rtitle (approved):
Looks good

http/src/main/resources/reference.conf
Qi77Qi: yea...there's no automation test for cryptomining yet...I was thinking about it, but figured we can add it if we run into a case where without automation test it becomes a problem since it'll be a bit of work to make a cryptomining automation test

PR #1792, opened 2020-12-14
[no ticket] Make boot streams resilient to errors

rtitle:
Relates to: https://broadworkbench.atlassian.net/browse/PROD-467

`Stream.handleErrorWith` actually aborts the stream if an error occurs, so this had the effect of an unhandled error in _any_ sub-stream causing the entire app to shut down. This happened on 12/12 when the Leo DB went down for a couple minutes: Leo logged `"Failed to start leonardo"` and was down for ~1 hour until it was manually restarted.

This change immediately tries to restart the app if an unhandled error occurs, which should hopefully make it more resilient to transient errors. 

Gonna try to do a test to see if the behavior is correct, feedback welcome.

---

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/Boot.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/Boot.scala
@@ -332,7 +332,7 @@ object Boot extends IOApp {
       val app = Stream.emits(allStreams).covary[IO].parJoin(allStreams.length)
 
       app
-        .handleErrorWith(error => Stream.eval(logger.error(error)("Failed to start leonardo")))
+        .handleErrorWith(error => Stream.eval(logger.error(error)("Unhandled error occurred")) ++ app)
         .compile
         .drain
     }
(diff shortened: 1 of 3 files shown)


rtitle:
@Qi77Qi curious what you think about this

Qi77Qi:
do you know which Stream is erroring out?...I think it'd be better to find which Stream is terminating and make it so that Stream can never terminate due to error...

I dont think we want to restart the app blindly. For instance, if there's a config error, we should just error and give up

rtitle:
Hm good point.. not sure which particular stream was erroring, will investigate..

rtitle:
Yeah I think there's a couple streams (`AutopauseMonitor` and `DateAccessedUpdater`) which don't gracefully handle DB connection errors. Going to fix those and retest (I am manually killing the DB in my fiab and seeing what happens).

Qi77Qi (approved):
👍

rtitle:
Thanks for the thumb.. I verified in a fiab that Leo now automatically recovers when the DB goes down & up. 🚀

rtitle:
build failed with error connecting to `broadinstitute.jfrog.io`.. jenkins retest

PR #1806, opened 2021-01-06
[IA-2395] Remove nodepool cleanup from app deletion

jdcanas:
This will be accompanied by a PR in leonardo cron job repo to clean up nodepools in the scope of the same ticket. There is no dependency on delaying keeping the nodepool around during app deletion, since the cron job will operate on some time window and clean up any old nodepools when it starts running.


rtitle:
This PR looks good, it should be released close to the new cron job though to ensure we don't leave node pools lying around.

jdcanas:
jenkins retest

rtitle (approved):
Looks good

PR #1809, opened 2021-01-06
Don't include welder for rstudio runtimes IA-2353

ehanna4:
Don't add welder for Rstudio runtimes.

Manual Tests:
Create an rstudio runtime through Terra-UI
- check the Cluster_Image table and see welder is not there
- ssh into the cluster and don’t see welder
- run locally on Terra-UI (after hotswapping) and it’s good to go

Create a jupyter cluster through Terra-UI
- check that cluster_image table and that welder is there
- ssh into the cluster and see welder
- run locally on Terra-UI (after hotswapping) and it’s good to go

Screen Shot 2021-01-06 at 3 17 24 PM

Screen Shot 2021-01-06 at 3 17 29 PM

Screen Shot 2021-01-06 at 3 17 41 PM

Updated the unit test that creates a new Rstudio to make sure that Welder is not there

---

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BaseCloudServiceRuntimeMonitor.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BaseCloudServiceRuntimeMonitor.scala
@@ -459,9 +459,6 @@ abstract class BaseCloudServiceRuntimeMonitor[F[_]] {
       availableTools <- streamFUntilDone(checkTools,
                                          monitorConfig.checkToolsMaxAttempts,
                                          monitorConfig.checkToolsInterval).compile.lastOrError
-      _ <- availableTools.traverse(x =>
-        logger.info("!!!RuntimeImageType:" + x._1.toString + "Boolean:" + x._2.toString)
-      )
       r <- availableTools match {
         case a if a.forall(_._2) =>
           readyRuntime(runtimeAndRuntimeConfig, ip, monitorContext, dataprocInstances)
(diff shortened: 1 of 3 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/RuntimeServiceInterp.scala
```diff
@@ -842,6 +850,13 @@ object RuntimeServiceInterp {
         else req.scopes //default to create gce runtime if runtimeConfig is not specified
     }
 
+    // Do not enable welder for rstudio
+    val welderEnabled =
```
rtitle: Could this just be:
```
val welderEnabled = clusterImages.map(_.imageType).contains(Welder)
```
?

Since we already have the logic to exclude welder for RStudio runtimes in `getRuntimeImages`, we shouldn't need to duplicate it here.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BaseCloudServiceRuntimeMonitor.scala
```diff
@@ -459,6 +459,9 @@ abstract class BaseCloudServiceRuntimeMonitor[F[_]] {
       availableTools <- streamFUntilDone(checkTools,
                                          monitorConfig.checkToolsMaxAttempts,
                                          monitorConfig.checkToolsInterval).compile.lastOrError
+      _ <- availableTools.traverse(x =>
```
rtitle: Reminder to remove before merge

rtitle:
Just took a quick look -- 1 comment above but the approach looks good. I guess we don't need to update the bash scripts after all since all welder stuff is already gated by the `WELDER_ENABLED` boolean.

Qi77Qi (approved):
lgtm

PR #1814, opened 2021-01-11
[IA-2439] Improve docker image auto-detection and error reporting

kyuksel:
- Fixes the issues described in the [JIRA ticket](https://broadworkbench.atlassian.net/browse/IA-2439).
- As part of our docker image auto-detection process, we had been expecting the field `container_config` to be present in the response coming back from a Docker API. In a user's custom image case, that field doesn't exist for some reason. Instead we use the `config` field's `env` variables.
- Since we hadn't been catching the error thrown when JSON decoding fails, we were reporting a very low-level error message to the user. That's now fixed. We do also include the cause for debug purposes, though.

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpDockerDAO.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpDockerDAO.scala
@@ -47,13 +47,18 @@ class HttpDockerDAO[F[_]] private (httpClient: Client[F])(implicit logger: Logge
       traceId <- ev.ask
 
       parsed <- parseImage(image)
+
       tokenOpt <- getToken(parsed, petTokenOpt)
 
       digest <- parsed.imageVersion match {
         case Tag(_)      => getManifestConfig(parsed, tokenOpt).map(_.digest)
         case Sha(digest) => F.pure(digest)
       }
+      _ <- F.delay(println(s"\n\ndigest for ${image.imageUrl} = $digest\n"))
+
       containerConfig <- getContainerConfig(parsed, digest, tokenOpt)
+      _ <- F.delay(println(s"\n\ncontainerConfig for ${image.imageUrl} = $containerConfig\n"))
+
       envSet = containerConfig.env.toSet
       tool = clusterToolEnv
         .find {
@@ -179,7 +184,9 @@ object HttpDockerDAO {
     } yield ManifestConfig(mediaType, size, digest)
   }
   implicit val containerConfigDecoder: Decoder[ContainerConfig] = Decoder.instance { d =>
+    println(s"\n\n d = ${d.value}")
     val cursor = d.downField("container_config")
+    println(s"\n\n cursor = ${cursor.values}")
     for {
       image <- cursor.get[String]("Image")
       env <- cursor.get[List[String]]("Env")
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpDockerDAO.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpDockerDAO.scala
@@ -51,17 +51,21 @@ class HttpDockerDAO[F[_]] private (httpClient: Client[F])(implicit logger: Logge
       tokenOpt <- getToken(parsed, petTokenOpt)
 
       digest <- parsed.imageVersion match {
-        case Tag(_)      => getManifestConfig(parsed, tokenOpt).map(_.digest)
+        case Tag(_) =>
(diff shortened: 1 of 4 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpDockerDAO.scala
```diff
@@ -178,12 +191,25 @@ object HttpDockerDAO {
...
+
+    lazy val configDecoder: Decoder[ContainerEnv] = Decoder.instance { d =>
+      val cursor = d.downField("config")
+
+      for {
+        env <- cursor.get[List[String]]("Env")
+      } yield ContainerEnv(env)
+    }
+
+    containerConfigDecoder or configDecoder
```
kyuksel: Circe selects the first decoder that succeeds.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpDockerDAO.scala
```diff
@@ -210,9 +236,9 @@ final case class ParsedImage(registry: ContainerRegistry,
 }
 
 // API response models
-final case class Token(token: String)
+final case class Token(token: String) extends AnyVal
 final case class ManifestConfig(mediaType: String, size: Int, digest: String)
-final case class ContainerConfig(image: String, env: List[String])
```
kyuksel, line 215: We haven't been using `image` so removed this `case class` and stopped trying to parse that field from the JSON, also because the particular user image doesn't contain that field at all.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeoException.scala
```diff
@@ -7,8 +7,9 @@ import org.broadinstitute.dsde.workbench.leonardo.http.errorReportSource
 class LeoException(val message: String = null,
                    val statusCode: StatusCode = StatusCodes.InternalServerError,
                    val cause: Throwable = null)
-    extends WorkbenchException(message) {
+    extends WorkbenchException(message, cause) {
```
kyuksel, line 10: `LeoException` wasn't overriding `WorkbenchException`'s `cause` field which defaults to `null` and hence swallowing information that is important for this PR's purposes, and probably other use cases' as well.

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpDockerDAOSpec.scala
```diff
@@ -18,6 +18,7 @@ class HttpDockerDAOSpec extends AnyFlatSpec with Matchers with BeforeAndAfterAll
     // TODO this will break if AoU moves off Dockerhub and we delete these images
     // dockerhub with tag
     ContainerImage("broadinstitute/terra-jupyter-aou:1.0.17", DockerHub),
+    ContainerImage("kkimler/r4.0_tca:2020_11_09", DockerHub), // added after a bug fix (see ticket IA-2439)
```
kyuksel, line 28: This image is the problematic user image. I haven't been able to find another DockerHub image with the same characteristics. I'm thinking it's better to have one than none. If the user makes changes and breaks our unit tests, it should be pretty straightforward to remove it then.

rtitle (approved):
Nice fix! Looks good 👍

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpDockerDAO.scala
Qi77Qi: I think decoder can be 
```
Decoder.instance { d =>
      for {
        envOpt <- d.downField("container_config").as[Option[List[String]]]
        env <- envOpt.fold(d.downField("config").as[List[String]])(ls => ls.asRight[DecodingFailure])
      } yield ContainerEnv(env)
    }
```

kyuksel: I don't think that would work because the final field we're looking for is `Env` which may be under `container_config` or `config`. It wasn't clear to me how the for-comprehension in your message can be modified to accommodate for that and since the PR'ed version is easy to read IMO, I'm inclined to not change it. What do you think @Qi77Qi ?

Qi77Qi: I think the current decoder is a bit repetitive..but if you prefer that, I won't insist

Qi77Qi (approved):

kyuksel:
jenkins retest

PR #1821, opened 2021-01-14
[IA-2460] Change R package installation dir

gpcarr:
Tested in a fiab.. created an rstudio runtime using the newest anvil-rstudio-bioconductor image in this PR: https://github.com/anvilproject/anvil-docker/pull/23 and verified that the correct dir is created for R packages and tested that packages install correctly and persist between runtimes

R packages will now be in the directory `/home/rstudio/R/x86_64-pc-linux-gnu-library/{R_VERSION}-${BIOCONDUCTOR_VERSION}` which is more R native package directory. for more questions about this directory, reach out to @nturaga 

The only concern I have is that if a user creates an RStudio VM, installs R packages and then deletes their runtime but keeps their PD then creates a Jupyter runtime they will have the R packages previously installed on their disk in a directory that is bioconductor specific (see screenshot). 
Screen Shot 2021-01-19 at 3 52 54 PM

not sure if this is a blocker, i'm leaning towards leaving it as is because i tested installing the same R package but in jupyter and it works fine and does not even recognize the package in `/home/rstudio/R/x86_64-pc-linux-gnu-library/{R_VERSION}-${BIOCONDUCTOR_VERSION}` (see second screenshot)
Screen Shot 2021-01-19 at 4 13 27 PM

---

In all cases:

### DIFF
--- a/http/src/main/resources/init-resources/gce-init.sh
+++ b/http/src/main/resources/init-resources/gce-init.sh
@@ -480,7 +480,6 @@ if [ ! -z "$RSTUDIO_DOCKER_IMAGE" ] ; then
   if [ $EXIT_CODE -ne 0 ]; then
     echo "RStudio user package installation directory creation failed, creating /packages directory"
     docker exec ${RSTUDIO_SERVER_NAME} /bin/bash -c "mkdir -p ${RSTUDIO_USER_HOME}/packages && chmod a+rwx ${RSTUDIO_USER_HOME}/packages"
-    exit $EXIT_CODE
   fi
 fi
 
--- a/http/src/main/resources/init-resources/gce-init.sh
+++ b/http/src/main/resources/init-resources/gce-init.sh
@@ -113,6 +113,8 @@ IS_GCE_FORMATTED=$(isGceFormatted)
 JUPYTER_HOME=/etc/jupyter
 JUPYTER_SCRIPTS=${JUPYTER_HOME}/scripts
 JUPYTER_USER_HOME=/home/jupyter-user
+RSTUDIO_SCRIPTS=/etc/rstudio/scripts
+RSTUDIO_USER_HOME=/home/rstudio
 KERNELSPEC_HOME=/usr/local/share/jupyter/kernels
 SERVER_CRT=$(proxyServerCrt)
 SERVER_KEY=$(proxyServerKey)
@@ -146,9 +148,6 @@ if [ "$IS_GCE_FORMATTED" == "false" ] ; then
   mkfs.ext4 -m 0 -E lazy_itable_init=0,lazy_journal_init=0,discard /dev/${DISK_DEVICE_ID}
 fi
 mount -o discard,defaults /dev/${DISK_DEVICE_ID} /work
-# Directory to store user installed packages so they persist
-mkdir -p /work/packages
-chmod a+rwx /work/packages
 # Ensure persistent disk re-mounts if runtime stops and restarts
 cp /etc/fstab /etc/fstab.backup
 echo UUID=`blkid -s UUID -o value /dev/${DISK_DEVICE_ID}` /work ext4 discard,defaults,nofail 0 2 | tee -a /etc/fstab
@@ -294,6 +293,10 @@ STEP_TIMINGS+=($(date +%s))
 # Jupyter-specific setup, only do if Jupyter is installed
 if [ ! -z "$JUPYTER_DOCKER_IMAGE" ] ; then
   log 'Installing Jupydocker kernelspecs...'
+
+  # user package installation directory
+  mkdir -p /work/packages
+  chmod a+rwx /work/packages
 
   # Used to pip install packacges
   # TODO: update this if we upgrade python version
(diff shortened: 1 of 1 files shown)


http/src/main/resources/init-resources/gce-init.sh
```diff
@@ -470,6 +473,23 @@ if [ ! -z "$JUPYTER_DOCKER_IMAGE" ] ; then
   STEP_TIMINGS+=($(date +%s))
 fi
 
+# RStudio specific setup; only do if RStudio is installed
+if [ ! -z "$RSTUDIO_DOCKER_IMAGE" ] ; then
+  BIOCONDUCTOR_VERSION=`docker exec rstudio-server R CMD printenv BIOCONDUCTOR_VERSION`
+  # for backwards compatibility (Bioconductor images before version 3.12)
+  # Environment variable $BIOCONDUCTOR_VERSION only exists in images >= 3.12 ie us.gcr.io/anvil-gcr-public/anvil-rstudio-bioconductor:0.0.9 introduces this env var
+  if [ -z "$BIOCONDUCTOR_VERSION" ] ; then
+    BIOCONDUCTOR_VERSION=`docker inspect --format '{{.Config.Labels.version}}' ${RSTUDIO_SERVER_NAME} | sed 's/\(^[0-9].[0-9][0-9]\).*/\1/g'`
```
nturaga: This should work for all the images 0.0.1 - 0.0.9 

```
BIOCONDUCTOR_VERSION=`docker exec rstudio-server R --version | grep 'R version' | sed 's/R version \([0-9].[0-9]\).*/\1/g'`
```

nturaga:
The corresponding PR on the anvil-rstudio-bioconductor side of things is https://github.com/anvilproject/anvil-docker/pull/23.

gpcarr:
Because we moved the logic to create this directory and give permission to run the script to the anvil-docker repo, these changes are not backwards compatible for older bioconductor images

http/src/main/resources/init-resources/gce-init.sh
```diff
@@ -146,9 +147,11 @@ if [ "$IS_GCE_FORMATTED" == "false" ] ; then
   mkfs.ext4 -m 0 -E lazy_itable_init=0,lazy_journal_init=0,discard /dev/${DISK_DEVICE_ID}
 fi
 mount -o discard,defaults /dev/${DISK_DEVICE_ID} /work
-# Directory to store user installed packages so they persist
-mkdir -p /work/packages
-chmod a+rwx /work/packages
+# Directory to store user installed packages so they persist for non RStudio runtimes
+if [ -z "$RSTUDIO_DOCKER_IMAGE" ] ; then
```
rtitle: Wonder if this block could be moved to the jupyter-specific setup below

gpcarr: i like that approach better, will update

http/src/main/resources/init-resources/gce-init.sh
```diff
@@ -470,6 +473,11 @@ if [ ! -z "$JUPYTER_DOCKER_IMAGE" ] ; then
   STEP_TIMINGS+=($(date +%s))
 fi
 
+# RStudio specific setup; only do if RStudio is installed
+if [ ! -z "$RSTUDIO_DOCKER_IMAGE" ] ; then
+  retry 3 docker exec ${RSTUDIO_SERVER_NAME} ${RSTUDIO_SCRIPTS}/set_up_package_dir.sh
```
rtitle: This might not be backwards compatible -- e.g. this script won't exist on older RStudio images which would cause the init script to fail. Could we check if the script exists in the image somehow? Alternatively we've used this pattern for things that might fail:
```
EXIT_CODE=0
docker_cmd_that_might_fail || EXIT_CODE=$?
if [ $EXIT_CODE -ne 0 ]; then
  echo "something failed"
fi
```

rtitle:
A few comments above.. also can you update the PR title, my eyes skip over it :)

gpcarr:
@rtitle I added the logic so that these changes are backwards compatible. I added the `EXIT_CODE` so that if the script does not exist for older images, the init script won't fail. I also added [logic](https://github.com/DataBiosphere/leonardo/pull/1821/files#diff-0ede0263691f9df737e1beecb48f32291556d43f643ecd76f4ebcd14314b6299R482) to create the `/home/rstudio/packages` directory if the script fails so that user installed packages for older images still persist. older images will have `R_LIBS='/home/rstudio/packages` in `/usr/local/lib/R/etc/Renviron.site` so we need to make sure this directory gets created for older images. see https://github.com/anvilproject/anvil-docker/blob/master/anvil-rstudio-bioconductor/Dockerfile#L35 for reference. this PR is ready for final review :) i also retested this in a fiab using jupyter images, older rstudio images and the newest rstudio image which is in my anvil-docker PR

http/src/main/resources/init-resources/gce-init.sh
```diff
@@ -470,6 +473,17 @@ if [ ! -z "$JUPYTER_DOCKER_IMAGE" ] ; then
...
 fi
 
+# RStudio specific setup; only do if RStudio is installed
+if [ ! -z "$RSTUDIO_DOCKER_IMAGE" ] ; then
+  EXIT_CODE=0
+  retry 3 docker exec ${RSTUDIO_SERVER_NAME} ${RSTUDIO_SCRIPTS}/set_up_package_dir.sh || EXIT_CODE=$?
+  if [ $EXIT_CODE -ne 0 ]; then
+    echo "RStudio user package installation directory creation failed, creating /packages directory"
+    docker exec ${RSTUDIO_SERVER_NAME} /bin/bash -c "mkdir -p ${RSTUDIO_USER_HOME}/packages && chmod a+rwx ${RSTUDIO_USER_HOME}/packages"
+    exit $EXIT_CODE
```
rtitle: We probably don't want to exit here, right? E.g. the script should continue on.

gpcarr: ohh yeah probably not - i added it following the other patterns in the script but this is a different case. I'll remove this but quick question - i tested with an older bioconductor image and the script did not fail. i'm guessing `exit` does not mean the script fails but rather just terminates sooner?

rtitle: Yeah looks like this is near the end anyway (only thing left is `docker prune`) so the image probably started up anyway… still better to not `exit` though. :)

rtitle (approved):
:shipit:

nturaga:
Looks good to me too!

PR #1832, opened 2021-01-19
IA-2342: Adding Stopped as a deleteable status for App Pause/Resume UI

xuscrbw:
Since we are implementing Pause/Resume into the UI. It makes sense for us to add stopped as an eligible status to be deleted. 

I think it should just be a change here, but wanted to check to make sure there wasn't anything else that needed to be added

---

In all cases:


rtitle (approved):
LGTM - I think that should be it

xuscrbw:
@rtitle I should be fine to merge despite the codecov/patch failure correct? Or do I need to update a unit test somewhere

rtitle:
I think it's fine, I usually consider codecov diffs < 1% to be noise

PR #1833, opened 2021-01-19
[IA-2472] Turn on debug-level logging when JUPYTER_DEBUG_LOGGING env var is true

gjuggler:
The goal here is to allow clients to conditionally turn on debug-level logs in Leo clusters. Based on a bit of searching around, this seems to be the most appropriate config value to turn on for debug logs. There is also a c.Session value ([ref](https://github.com/aws/amazon-sagemaker-examples/issues/818)), but that's likely overkill for what AoU is looking for.

### DIFF
--- a/http/src/main/resources/init-resources/jupyter_notebook_config.py
+++ b/http/src/main/resources/init-resources/jupyter_notebook_config.py
@@ -5,6 +5,10 @@
 import os
 
 c = get_config()
+
+if os.environ.get('JUPYTER_DEBUG_LOGGING') == 'true':
+  c.Application.log_level = 'DEBUG'
+
 c.NotebookApp.ip = '0.0.0.0'
 c.NotebookApp.port = 8000
 c.NotebookApp.open_browser = False


rtitle (approved):
As mentioned in https://github.com/DataBiosphere/terra-docker/pull/193, this file should be obsolete but shouldn't hurt to merge this.

Qi77Qi:
shall we close this one...I'll just remove these files from leo

rtitle:
See comment in https://github.com/DataBiosphere/leonardo/pull/1834/#discussion_r561050400, this file may actually still be needed (we also bake it in to terra-docker but left it in Leo to support old image versions).

Also I think CI is failing here because there is a `/` in the branch name -- @gjuggler could we re-submit the PR without the slash?

gjuggler:
👍 I'll reopen a new PR. (It may be worth fixing the automation at some point though, as I believe slash-delimited branches are a pretty common pattern.)

rtitle:
Yeah it's bitten us before :(

PR #1834, opened 2021-01-20
[no ticket] Bump a few dependencies

Qi77Qi:
1. bump a few dependencies
2. Remove a jupyter script that's no longer used

---

In all cases:


http/src/main/resources/init-resources/jupyter_notebook_config.py
```diff
@@ -1,47 +0,0 @@
-# adapted from https://github.com/jupyter/docker-stacks/blob/master/base-notebook/jupyter_notebook_config.py
```
rtitle: Hm there might be some Leo code that's still referencing this file..

rtitle: See: 
https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/resources/reference.conf#L337

and
https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/resources/init-resources/gce-init.sh#L308

Hmm maybe the file in Leo _is_ still used, and overwrites the version in terra-docker.

rtitle (changes requested):
Maybe could just add `jupyter_notebook_config.py` back for this PR.. need to figure out how to remove it given we have old images still in use.

rtitle (approved):
:shipit: if tests pass

PR #1836, opened 2021-01-20
[IA-2472] Turn on debug-level logging when JUPYTER_DEBUG_LOGGING env var is true

gjuggler:
The goal here is to allow clients to conditionally turn on debug-level logs in Leo clusters. Based on a bit of searching around, this seems to be the most appropriate config value to turn on for debug logs. There is also a c.Session value (ref), but that's likely overkill for what AoU is looking for.


rtitle (approved):

rtitle:
Will merge when tests pass

PR #1846, opened 2021-01-26
[IA-2486] remove zombie monitor

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-2486

The current zombie monitor is not excluding runtimes with  same project/runtime but still active properly. Hence, it's deleting runtimes that shouldn't be deleted (this mostly affects AOU because they re-use runtime names).

The downside is this check is run every 4 hours, while resource validator only runs once a day, so it increases the max time a "should-be-deleted" runtime can incur cost (we can increase frequence for resource validator if this is a real concern tho). 

---

In all cases:


xuscrbw (approved):
LGTM.

rtitle:
I'm all for removing code -- just double checking, the "inactive" functionality currently exists in Resource Validator for GCE and Dataproc?

rtitle (approved):

Qi77Qi:
@rtitle yea, what's left in zombie monitor is superseded by resource validator, which checks if a "Deleted" runtime still exits in google, if yes, delete in google...the only difference is this code is run more frequently than resource validator

PR #1847, opened 2021-01-26
[no ticket] Add set_parent Sam action to Leo resources

rtitle:
This is needed to allow tests to pass on this PR: https://github.com/broadinstitute/sam/pull/497

… which will then allow us to do https://broadworkbench.atlassian.net/browse/IA-2487

@dvoet fyi

---

In all cases:

### DIFF
--- a/http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpSamDAOSpec.scala
+++ b/http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpSamDAOSpec.scala
@@ -11,11 +11,10 @@ import org.broadinstitute.dsde.workbench.leonardo.model.ServiceAccountProviderCo
 import org.broadinstitute.dsde.workbench.model.WorkbenchEmail
 import org.broadinstitute.dsde.workbench.util.health.Subsystems.{GoogleGroups, GoogleIam, GooglePubSub, OpenDJ}
 import org.broadinstitute.dsde.workbench.util.health.{StatusCheckResponse, SubsystemStatus}
+import org.http4s._
 import org.http4s.circe.CirceEntityEncoder._
 import org.http4s.client.Client
 import org.http4s.client.middleware.{Retry, RetryPolicy}
-import org.http4s.headers.Authorization
-import org.http4s._
 import org.scalatest.BeforeAndAfterAll
 import org.scalatest.flatspec.AnyFlatSpec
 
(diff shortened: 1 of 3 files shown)


http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpSamDAOSpec.scala
```diff
@@ -135,6 +136,105 @@ class HttpSamDAOSpec extends AnyFlatSpec with LeonardoTestSuite with BeforeAndAf
 
     res.unsafeRunSync()
   }
+
+  it should "tolerate unknown runtime actions" in {
```
Qi77Qi: when is it necessary tolerate unknown actions? what would be an example of such unknown action?

rtitle: set_parent was a new action needed to support hierarchical resources. Leo doesn’t need to use it yet; but since we list all actions it broke decoding when Sam added it to its conf. Just thinking this change will help decouple Leo/Sam if Sam needs to add an action again.

Qi77Qi: I'm thinking if sam changes action model, leo should be aware of it? hence not handling the unknown case is more desirable so that things will fail when it happens?

rtitle: I see that logic but I also think it'd be nice to allow _additive_ changes to Sam without needing to update the Leo enum. E.g. without this PR, if we need to add a Sam action we need to merge 3 PRs in sequence:

1. Update Leo enum (if we don't do this then step 2 will break Leo)
2. Add action to Sam conf
3. Update Leo to use new Sam action

After this change we just need to merge 2 PRs:

1. Add action to Sam conf (Leo will ignore it)
2. Update Leo to use that action

Qi77Qi: my thoughts are, when we add action in sam, it's probably a conversation between ci and us, and we agree on what the additional action is. so it'll just be
1. Update Leo to allow new Sam action
2. Add action to sam conf

This does mean things have to happen in this order...but I think we'll have less surprises when we do this

rtitle: I see the concern about tricky surprises if we have `Other` and Leo's knowledge of actions doesn't match Sam's. I think we've had this debate before. :) I'm fine with removing `Other` and just adding the `SetParent` object

rtitle: OK, updated this PR

Qi77Qi (approved):
thanks for fixing this!

PR #1852, opened 2021-01-27
[IA-2441] Use google2 billing

jdcanas:
This PR removes the use of the google1 dao currently used for billing.

It currently uses a snap, pending this PR: https://github.com/broadinstitute/workbench-libs/pull/490

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreter.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreter.scala
@@ -23,6 +23,7 @@ import org.broadinstitute.dsde.workbench.google2.{
   GoogleComputeService,
   GoogleDataprocService,
   GoogleDiskService,
+  GoogleResourceService,
   MachineTypeName,
   ZoneName
 }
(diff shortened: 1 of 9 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreter.scala
```diff
@@ -424,12 +425,10 @@ class DataprocInterpreter[F[_]: Timer: Parallel: ContextShift](
...
-            projectNumberOptIO <- F.liftIO(
-              IO.fromFuture(IO(googleProjectDAO.getProjectNumber(googleProject.value)))
-            )
-            projectNumber <- F.liftIO(
-              IO.fromEither(projectNumberOptIO.toRight(ClusterIamSetupException(imageProject)))
-            )
+            projectNumberOpt <- googleResourceService.getProjectNumber(googleProject)
+
+            //TODO: This can also be from a 404?
+            projectNumber <- F.fromEither(projectNumberOpt.toRight(ClusterIamSetupException(imageProject)))
```
jdcanas: So, I'm not really sure in what circumstances the exception here occurs.

I wanted to see if anyone knew what it was for (@rtitle ?) before I touchedd it, but I added a TODO there with my thoughts. I'm thinking this should be a project not found exception. As in, the library returns `Option[Long]` now, where it is none if the project passed is not found. Any idea if the exception here is still relevant?

Qi77Qi: my guess is this was just for dealing with the case when project doesn't exist...does your `googleResourceService.getProjectNumber` return `None` when project doesn't exist currently?

jdcanas: yeah, I'm thinking I should change the exception here

Qi77Qi: @rtitle just wonder if you have more context around this code...

Qi77Qi (approved):
lgtm as long as the exception refactor is what was intended

PR #1853, opened 2021-01-27
IA-2445 Move NonLeoMessageSubscriber to subscriber module

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-2445

At a high level, this PR moves `NonleoMessageSubscriber` to `subscriber` module, and making `http` depends on `subscriber` module, so that `NonleoMessageSubscriber` will be run the same way as before. The benefit of doing this is, in the future, we'll be able to have `subscriber` as a separate container, but that'll require additional `terra-helm` file change and making multiple containers work in a fiab, so this is a good first step.

1. Created a subscriber module that can be deployed as a separate container. Config reading for this container is using `pureconfig`. Some common `ConfigReader` can be used for cron jobs repo. Hence they live in `core` for now.
2. Moved more code to `core` so that they can be used by `subscriber`
3. Moved a few individual `Config` files from `http` to `configs.scala` in `core`
4. Add `ConfigSpec` in `subscriber`
5. Add `NewConfig` in `http` that has configs for `NonLeomessageSubscriber`....We can expand this config and slowly move existing configs to `NewConfig` over time. 
6. Convert `StatusService` to use `fs2` instead of actors because we aren't cleanly shutdown actors between unit tests and I wasn't able to figure out what else exactly should be shutdown (we can't shutdown actor system becuz it's shared across unit tests)
7. Config files in `subscriber` and `http` are duplicated currently. But only `http` is really being used. And once we do start using `subscriber` module as a separate container, we can figure out a way to not duplicate configs (my expectation is lots of the configs currently exist in `http` can be moved to `subscriber` once we move `LeonardoSubscriber` out of `http`)

TODO:
1. Test subscriber works appropriately 

---

In all cases:


Qi77Qi:
jenkins retest

Qi77Qi:
jenkins retest

PR #1854, opened 2021-01-28
[IA-2469] preventing csrf attacks in Leo proxy

gpcarr:
jira ticket: https://broadworkbench.atlassian.net/browse/IA-2469

accompanying firecloud-develop PR: https://github.com/broadinstitute/firecloud-develop/pull/2408

added unit tests and tested in a fiab manually adding `refererConfig` to [reference.conf](https://github.com/DataBiosphere/leonardo/pull/1854/files#diff-9024d8d3f171d065fa3b61dbc7dc2c56bdc6587644fde470722fdc92e7068f61R445). the config that was manually added is identical to the config added to firecloud-develop in this [PR](https://github.com/broadinstitute/firecloud-develop/pull/2408) for the dev environment. Just kicked off a `fiab-start` job that will test these changes alongside the firecloud-develop changes as well

---

In all cases:

### DIFF
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -443,6 +443,12 @@ contentSecurityPolicy {
   ]
 }
 
+refererConfig {
+  # list of valid hosts and enabled flag is environment specific so it's overriden in firecloud develop
+  validHosts = []
+  enabled = false // false for dev and QA environments, true everywhere else
+}
+
 swagger {
   #googleClientId = "client_id"
   #realm = "broad-dsde-dev"
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/RefererConfig.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/RefererConfig.scala
@@ -1,0 +1,4 @@
+package org.broadinstitute.dsde.workbench.leonardo
+package config
+
+final case class RefererConfig(validHosts: Set[String], enabled: Boolean)
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/ProxyRoutes.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/ProxyRoutes.scala
@@ -17,6 +17,7 @@ import org.broadinstitute.dsde.workbench.leonardo.model.AuthenticationError
 import com.typesafe.scalalogging.LazyLogging
 import io.opencensus.scala.akka.http.TracingDirective.traceRequestForService
 import org.broadinstitute.dsde.workbench.google2.KubernetesSerializableName.ServiceName
+import org.broadinstitute.dsde.workbench.leonardo.config.RefererConfig
 import org.broadinstitute.dsde.workbench.leonardo.{AppContext, AppName, RuntimeName}
 import org.broadinstitute.dsde.workbench.leonardo.dao.TerminalName
 import org.broadinstitute.dsde.workbench.leonardo.http.service.ProxyService
@@ -24,7 +25,7 @@ import org.broadinstitute.dsde.workbench.model.UserInfo
 import org.broadinstitute.dsde.workbench.model.google.GoogleProject
 import org.broadinstitute.dsde.workbench.openTelemetry.OpenTelemetryMetrics
 
-class ProxyRoutes(proxyService: ProxyService, corsSupport: CorsSupport)(
+class ProxyRoutes(proxyService: ProxyService, corsSupport: CorsSupport, refererConfig: RefererConfig)(
   implicit materializer: Materializer,
   cs: ContextShift[IO],
   metrics: OpenTelemetryMetrics[IO]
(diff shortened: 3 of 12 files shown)


gpcarr:
failing tests include:
`should 404 for non-existent runtimes *** FAILED ***`
`should proxy websockets *** FAILED ***`
`should 404 for non-existent apps *** FAILED ***`

gpcarr:
> failing tests include:
> `should 404 for non-existent runtimes *** FAILED ***`
> `should proxy websockets *** FAILED ***`
> `should 404 for non-existent apps *** FAILED ***`

actually only failing test is `should proxy websockets`
not sure if websockets need to be handled differently in terms of referers @rtitle do you have any insight? could just need to add header to this [line](https://github.com/DataBiosphere/leonardo/blob/develop/http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/ProxyRoutesSpec.scala#L231) but want to make sure that there's nothing else i'm missing

http/src/main/resources/reference.conf
```diff
@@ -442,6 +442,16 @@ contentSecurityPolicy {
   ]
 }
 
+refererConfig {
+  # list of valid ports and hosts is environment specific so it's overriden in firecloud develop
+  validPorts = [
```
rtitle: Reading the code, it looks like `validPorts` is only used if the host == `"localhost"`. Wondering if that might be confusing to remember -- could we just represent the allow-list as host:port strings? For example:
```
refererConfig {
  allowedHosts = [
    "localhost:3000",
    "app.terra.bio"
  ]
}
```

Then we could probably check the referer against it with `referer.getUri.authority.toString`:
```
scala> val uri1 = Uri("https://app.terra.bio/")
val uri1: akka.http.scaladsl.model.Uri = https://app.terra.bio/

scala> uri1.authority.toString
val res13: String = app.terra.bio

scala> val uri2 = Uri("http://localhost:3000")
val uri2: akka.http.scaladsl.model.Uri = http://localhost:3000

scala> uri2.authority.toString
val res14: String = localhost:3000
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/RefererConfig.scala
```diff
@@ -0,0 +1,20 @@
+package org.broadinstitute.dsde.workbench.leonardo
+package config
+
+import org.broadinstitute.dsde.workbench.leonardo.config.RefererComponent.{ValidHosts, ValidPorts}
+
+final case class RefererConfig(validHosts: ValidHosts, validPorts: ValidPorts)
+
+sealed trait RefererComponent {
```
rtitle: Not certain we need this ADT -- 1 case class is probably fine:
```
final case class RefererConfig(allowedHosts: List[String])
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/ProxyRoutes.scala
```diff
@@ -248,4 +251,35 @@ class ProxyRoutes(proxyService: ProxyService, corsSupport: CorsSupport)(
...
+          pass
+        } else checkReferer
+      }
+      case None => checkReferer
+    }
+
+  private[api] def checkReferer: Directive0 =
+    optionalHeaderValueByType(Referer) flatMap {
+      case Some(referer) => {
+        logger.info(
```
rtitle: Reminder to remove logging before merge

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/ProxyRoutes.scala
```diff
@@ -248,4 +251,35 @@ class ProxyRoutes(proxyService: ProxyService, corsSupport: CorsSupport)(
       userInfoOpt.traverse(userInfo => proxyService.invalidateAccessToken(userInfo.accessToken.token)) >>
       IO(logger.debug(s"Invalidated access token"))
         .as(StatusCodes.NoContent)
+
+  private[api] def refererHandler: Directive0 =
+    optionalHeaderValueByType(Upgrade) flatMap {
```
rtitle: Nit: could maybe be more concise:
```
optionalHeaderValueByType(Upgrade) flatMap {
  case Some(upgrade) if upgrade.hasWebSocket => pass
  case _ => checkReferer
}
```

http/src/main/resources/reference.conf
dinvlad: Sounds great! Any approach where host:port can be determined in full (without possibility to spoof the check with something like `app.terra.bio.evil.org`) is good.

gpcarr:
@rtitle just pushed up changes to address comments, going to test in a fiab now :)

http/src/main/resources/reference.conf
```diff
@@ -442,6 +442,14 @@ contentSecurityPolicy {
   ]
 }
 
+refererConfig {
+  # list of valid hosts is environment specific so it's overriden in firecloud develop
+  # add dev valid hosts here
+  validHosts = [
+    "none"
```
rtitle: As mentioned in 1-1, can probably put dev or example hosts here instead of "none"

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/RefererConfig.scala
```diff
@@ -0,0 +1,4 @@
+package org.broadinstitute.dsde.workbench.leonardo
+package config
+
+final case class RefererConfig(validHosts: List[String])
```
rtitle: Nit: maybe make this a `Set[String]` since we're checking it with `.contains()`

rtitle:
Couple more nits, overall looks good

http/src/main/resources/reference.conf
gpcarr: will update!

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/RefererConfig.scala
gpcarr: ah yes I also like that better - will update!

http/src/main/resources/reference.conf
```diff
@@ -442,6 +442,21 @@ contentSecurityPolicy {
...
+  # list of valid hosts is environment specific so it's overriden in firecloud develop - these are the valid
+  # hosts for our dev environment (including fiab)
+  validHosts = [
+    "localhost:3000",
+    "localhost:4200",
+    "localhost:443",
+    "bvdp-saturn-dev.appspot.com",
+    "all-of-us-workbench-test.appspot.com",
+    "dev.terra.biodatacatalyst.nhlbi.nih.gov",
+    "leonardo-fiab.dsde-dev.broadinstitute.org",
```
rtitle: I think this should be `leonardo-fiab.dsde-dev.broadinstitute.org:30443`? 

(Since fiabs have a port)

rtitle (approved):
Looks good - one fiab host comment, otherwise should be good to merge once tests pass

rtitle:
Hm.. just thought of something. Do automation tests work? Those use Selenium to hit the proxy. Do those calls have a referrer, and can we allow that somehow?

http/src/main/resources/reference.conf
gpcarr: going to retest in my fiab by running the hotswap script since fiab-start keeps failing

gpcarr:
> Hm.. just thought of something. Do automation tests work? Those use Selenium to hit the proxy. Do those calls have a referrer, and can we allow that somehow?

this is a good point - i'll log some stuff to see what the referer is for automation tests. automation tests run on fiabs correct? so my guess is we should be fine but i'll double check

gpcarr:
unrelated test failure (only 1 test failed `NotebookGCECustomizationSpec` `should execute user-specified start script` which is a currently known flaky test)

jenkins re-test

gpcarr:
jenkins retest

gpcarr:
same unrelated, flakey test failure from before ... jenkins retest

rtitle (approved):
This looks good to me.. 👍  if tests pass

gpcarr:
> This looks good to me.. 👍 if tests pass

tests passed - need to wait for my firecloud-develop PR to be unblocked before merging this PR tho

http/src/main/resources/reference.conf
```diff
@@ -442,6 +442,23 @@ contentSecurityPolicy {
   ]
 }
 
+refererConfig {
+  # list of valid hosts and enabled flag is environment specific so it's overriden in firecloud develop - these are the valid
+  # hosts for our dev environment (including fiab)
+  validHosts = [
+    "localhost:3000",
```
Qi77Qi: this list can be removed too?

Qi77Qi (approved):

gpcarr:
jenkins retest - same failure as before

gpcarr:
unrelated test failure ... jenkins retest

PR #1862, opened 2021-02-09
[IA-2302] Add rshiny automation test

rtitle:
https://broadworkbench.atlassian.net/browse/IA-2302

Opening PR to run tests, don't know if this works yet. 

It tries to open the example app via:
```
library(shiny)
runExample("01_hello")
```
and verify the pop-up window. 

---

In all cases:

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/rstudio/RShinyPage.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/rstudio/RShinyPage.scala
@@ -1,16 +1,22 @@
 package org.broadinstitute.dsde.workbench.leonardo.rstudio
 
-import org.broadinstitute.dsde.workbench.service.test.WebBrowserUtil
+import org.broadinstitute.dsde.workbench.auth.AuthToken
+import org.broadinstitute.dsde.workbench.page.CookieAuthedPage
 import org.openqa.selenium.WebDriver
 
-class RShinyPage(implicit val webDriver: WebDriver) extends WebBrowserUtil {
-  val header: Query = cssSelector("h2")
+import scala.concurrent.duration._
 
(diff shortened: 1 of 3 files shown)


rtitle:

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/rstudio/RShinyPage.scala
```diff
@@ -0,0 +1,22 @@
...
+
+import org.broadinstitute.dsde.workbench.auth.AuthToken
+import org.broadinstitute.dsde.workbench.page.CookieAuthedPage
+import org.openqa.selenium.WebDriver
+
+import scala.concurrent.duration._
+
+class RShinyPage(val url: String)(implicit override val authToken: AuthToken, implicit val webDriver: WebDriver)
+    extends CookieAuthedPage[RShinyPage] {
+  private val exampleHeader: Query = cssSelector("#showcase-app-container > div > h2")
```
jdcanas, line 11: css selectors tend to be pretty brittle... especially since this seems to rely on css in a third party app.

Sometimes it's the best/only way, but definitely worth thinking about for a second (I personally cant think of any appealing alternatives off the top of my head)

Qi77Qi (approved):

jdcanas (approved):

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/rstudio/RShinyPage.scala
rtitle: Yeah - think this works for the "example" apps from the shiny package, but definitely wouldn't work for all shiny apps. Also it would break if shiny changes their examples. Can't really think of a better way either though..

PR #1866, opened 2021-02-10
IA-2397 Use terra-app-setup-chart and persist PVC ids

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-2397
https://broadworkbench.atlassian.net/browse/IA-2533

Automation test failed becuz https://github.com/broadinstitute/firecloud-develop/pull/2426 needs to be merged

* use set up chart
* Persist pvc ids

Validated persisting pvc part working properly in fiab
```
qi@cloudshell:~ (qi-billing-2)$ kubectl get pvc -n ya8b8r-gxy-ns ya8b8r-gxy-rls-galaxy-pvc -o=jsonpath='{.metadata.uid}'
7d5ab7f8-5b75-422d-b4f5-0ee1b93a404e
qi@cloudshell:~ (qi-billing-2)$ kubectl get pvc -n ya8b8r-gxy-ns ya8b8r-gxy-rls-cvmfs-alien-cache-pvc -o=jsonpath='{.metadata.uid}'
939ec32a-6e8e-4a63-b6b5-a6802412bb9f
qi@cloudshell:~ (qi-billing-2)$
```
and this is what's in leo db

```
| id | nodepoolId | appType | appName | status  | samResourceId                        | creator                           | createdDate                | destroyedDate              | dateAccessed               | namespaceId | diskId | customEnvironmentVariables | googleServiceAccount                                               | kubernetesServiceAccount | chart                      | release        | galaxyPvcId                          | cvmfsPvcId                           |
+----+------------+---------+---------+---------+--------------------------------------+-----------------------------------+----------------------------+----------------------------+----------------------------+-------------+--------+----------------------------+--------------------------------------------------------------------+--------------------------+----------------------------+----------------+--------------------------------------+--------------------------------------+
|  1 |          4 | GALAXY  | pvc-0   | RUNNING | 5ca2d63e-59fb-4979-b7b7-10b905a75f5d | hermione.owner@test.firecloud.org | 2021-02-18 17:24:54.945000 | 1970-01-01 00:00:01.000000 | 2021-02-18 17:24:54.945000 |           1 |      1 | NULL                       | b153pet-110530393451290928813@qi-billing-2.iam.gserviceaccount.com | gxy-ksa                  | galaxy/galaxykubeman-0.7.2 | ya8b8r-gxy-rls | 7d5ab7f8-5b75-422d-b4f5-0ee1b93a404e | 939ec32a-6e8e-4a63-b6b5-a6802412bb9f |
```

TODO:
* Move pvc ids to PD table
* Add `lastUsedBy` in PD table
* 
---

In all cases:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -33,7 +33,7 @@ RUN curl -fsSL -o get_helm.sh https://raw.githubusercontent.com/helm/helm/master
 # Add the repos containing nginx and galaxy charts
 RUN helm repo add center https://repo.chartcenter.io && \
     helm repo add galaxy https://raw.githubusercontent.com/cloudve/helm-charts/anvil/ && \
-    helm repo add terra-app-setup-charts https://storage.googleapis.com/terra-app-setup-chart \
+    helm repo add terra-app-setup-charts https://storage.googleapis.com/terra-app-setup-chart && \
     helm repo update
 
 # .Files helm helper can't access files outside a chart. Hence in order to populate cert file properly, we're
--- a/Dockerfile
+++ b/Dockerfile
@@ -33,7 +33,18 @@ RUN curl -fsSL -o get_helm.sh https://raw.githubusercontent.com/helm/helm/master
 # Add the repos containing nginx and galaxy charts
 RUN helm repo add center https://repo.chartcenter.io && \
     helm repo add galaxy https://raw.githubusercontent.com/cloudve/helm-charts/anvil/ && \
+    helm repo add terra-app-setup-charts https://storage.googleapis.com/terra-app-setup-chart \
     helm repo update
+
+# .Files helm helper can't access files outside a chart. Hence in order to populate cert file properly, we're
+# pulling `terra-app-setup` locally and add cert files to the chart.
+# Leonardo will install the chart from local version.
+RUN pushd /leonardo && \
+    helm pull terra-app-setup-charts/terra-app-setup --untar && \
+    cp /etc/rootCA.pem terra-app-setup/ca.crt && \
+    cp /etc/jupyter-server.crt terra-app-setup/tls.crt && \
+    cp /etc/jupyter-server.key terra-app-setup/tls.key && \
+    popd
 
 # Add Leonardo as a service (it will start when the container starts)
 CMD java $JAVA_OPTS -jar $(find /leonardo -name 'leonardo*.jar')
--- a/Dockerfile
+++ b/Dockerfile
@@ -41,9 +41,6 @@ RUN helm repo add center https://repo.chartcenter.io && \
 # Leonardo will install the chart from local version.
 RUN pushd /leonardo && \
     helm pull terra-app-setup-charts/terra-app-setup --untar && \
-    cp /etc/rootCA.pem terra-app-setup/ca.crt && \
-    cp /etc/jupyter-server.crt terra-app-setup/tls.crt && \
-    cp /etc/jupyter-server.key terra-app-setup/tls.key && \
     popd
 
 # Add Leonardo as a service (it will start when the container starts)
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/ConfigImplicits.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/ConfigImplicits.scala
@@ -1,0 +1,17 @@
+package org.broadinstitute.dsde.workbench.leonardo
+
+import pureconfig.ConfigReader
+import cats.syntax.all._
+import org.broadinstitute.dsp.{ChartName, ChartVersion}
+import pureconfig.error.ExceptionThrown
+
+import java.nio.file.{Path, Paths}
+
+object ConfigImplicits {
+  implicit val pathConfigReader: ConfigReader[Path] =
+    ConfigReader.stringConfigReader.emap(s => Either.catchNonFatal(Paths.get(s)).leftMap(err => ExceptionThrown(err)))
+  implicit val chartNameConfigReader: ConfigReader[ChartName] =
+    ConfigReader.stringConfigReader.map(s => ChartName(s))
+  implicit val chartVersionConfigReader: ConfigReader[ChartVersion] =
+    ConfigReader.stringConfigReader.map(s => ChartVersion(s))
+}
--- a/http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20210217_add_app_pvc_id.xml
+++ b/http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20210217_add_app_pvc_id.xml
@@ -12,12 +12,12 @@
                 <constraints nullable="true"/>
             </column>
 
-            <column name="usedBy" type="BIGINT">
+            <column name="lastUsedBy" type="BIGINT">
                 <constraints nullable="true"/>
             </column>
         </addColumn>
 
-        <addForeignKeyConstraint baseColumnNames="usedBy" baseTableName="PERSISTENT_DISK" constraintName="FK_DISK_APP_ID" deferrable="false" initiallyDeferred="false" onDelete="NO ACTION" onUpdate="NO ACTION" referencedColumnNames="id" referencedTableName="APP"/>
+        <addForeignKeyConstraint baseColumnNames="lastUsedBy" baseTableName="PERSISTENT_DISK" constraintName="FK_DISK_APP_ID" deferrable="false" initiallyDeferred="false" onDelete="NO ACTION" onUpdate="NO ACTION" referencedColumnNames="id" referencedTableName="APP"/>
     </changeSet>
 
 </databaseChangeLog>
--- a/http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20210217_add_app_pvc_id.xml
+++ b/http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20210217_add_app_pvc_id.xml
@@ -2,12 +2,21 @@
 <databaseChangeLog logicalFilePath="leonardo" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
 
     <changeSet logicalFilePath="leonardo" author="qi" id="add_app_pvc_id">
-        <addColumn tableName="APP">
+        <addColumn tableName="PERSISTENT_DISK">
             <column name="galaxyPvcId" type="varchar(254)">
                 <constraints nullable="true"/>
             </column>
 
+            <!-- this is a second PVC id. We store this record in the same row for galaxy pvc Id because we don't currently persist disk Id for cvmfs. If we persist cmvfs disk separately in the future, we should move this field to new row -->
             <column name="cvmfsPvcId" type="varchar(254)">
+                <constraints nullable="true"/>
+            </column>
+
+            <column name="chart" type="varchar(254)">
+                <constraints nullable="true"/>
+            </column>
+
+            <column name="release" type="varchar(254)">
                 <constraints nullable="true"/>
             </column>
         </addColumn>
--- a/http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20210217_add_app_pvc_id.xml
+++ b/http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20210217_add_app_pvc_id.xml
@@ -1,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="leonardo" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+
+    <changeSet logicalFilePath="leonardo" author="qi" id="add_app_pvc_id">
+        <addColumn tableName="APP">
+            <column name="galaxyPvcId" type="varchar(254)">
+                <constraints nullable="true"/>
+            </column>
+
+            <column name="cvmfsPvcId" type="varchar(254)">
+                <constraints nullable="true"/>
+            </column>
+        </addColumn>
+    </changeSet>
+
+</databaseChangeLog>
--- a/http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20210217_add_app_pvc_id.xml
+++ b/http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20210217_add_app_pvc_id.xml
@@ -12,14 +12,12 @@
                 <constraints nullable="true"/>
             </column>
 
-            <column name="chart" type="varchar(254)">
-                <constraints nullable="true"/>
-            </column>
-
-            <column name="release" type="varchar(254)">
+            <column name="usedBy" type="BIGINT">
                 <constraints nullable="true"/>
             </column>
         </addColumn>
+
+        <addForeignKeyConstraint baseColumnNames="usedBy" baseTableName="PERSISTENT_DISK" constraintName="FK_DISK_APP_ID" deferrable="false" initiallyDeferred="false" onDelete="NO ACTION" onUpdate="NO ACTION" referencedColumnNames="id" referencedTableName="APP"/>
     </changeSet>
 
 </databaseChangeLog>
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/AppComponent.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/AppComponent.scala
@@ -188,15 +188,10 @@ object appQuery extends TableQuery(new AppTable(_)) {
       .map(_.status)
       .update(status)
 
-  def updatePvcIds(id: AppId, galaxyPvcId: PvcId, cvmfsPvcId: PvcId)(implicit ec: ExecutionContext): DBIO[Unit] =
-    for {
-      _ <- getByIdQuery(id)
-        .map(_.galaxyPvcId)
-        .update(Some(galaxyPvcId))
-      _ <- getByIdQuery(id)
-        .map(_.cvmfsPvcId)
-        .update(Some(cvmfsPvcId))
-    } yield ()
+  def updatePvcIds(id: AppId, galaxyPvcId: PvcId, cvmfsPvcId: PvcId): DBIO[Int] =
+    getByIdQuery(id)
+      .map(x => (x.galaxyPvcId, x.cvmfsPvcId))
+      .update((Some(galaxyPvcId), Some(cvmfsPvcId)))
 
   def updateChart(id: AppId, chart: Chart): DBIO[Int] =
     getByIdQuery(id)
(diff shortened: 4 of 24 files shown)


jdcanas:
This all looks like it will play well with the custom app code im working on, with a lot of conflicts though :(

Dockerfile
```diff
@@ -33,7 +33,15 @@ RUN curl -fsSL -o get_helm.sh https://raw.githubusercontent.com/helm/helm/master
 # Add the repos containing nginx and galaxy charts
 RUN helm repo add center https://repo.chartcenter.io && \
     helm repo add galaxy https://raw.githubusercontent.com/cloudve/helm-charts/anvil/ && \
+    helm repo add terra-app-setup-charts https://storage.googleapis.com/terra-app-setup-chart && \
```
rtitle, line 36: Did you push this manually?

I had created a `terra-app-charts` bucket, wondering if we could use that bucket for both the terra-app and setup charts.

Dockerfile
```diff
@@ -33,7 +33,15 @@ RUN curl -fsSL -o get_helm.sh https://raw.githubusercontent.com/helm/helm/master
...
     helm repo update
 
+# .Files helm helper can't access files outside a chart. Hence in order to populate cert file properly, we're
+# pulling `terra-app-setup` locally and add cert files to the chart.
+# Leonardo will install the chart from local version.
+RUN pushd /leonardo && \
```
rtitle, line 42: Could also `cp` the certs here instead of mounting them separately in the docker-compose file

rtitle: Er actually maybe not.. the cp would run before the mount I guess

Dockerfile
Qi77Qi: yea..I manually pushed to this chart....I wasn't sure if it's supposed to be in the same bucket and was afraid to break the existing chart so I created a separate one

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/ConfigImplicits.scala
```diff
@@ -0,0 +1,17 @@
+package org.broadinstitute.dsde.workbench.leonardo
+
+import pureconfig.ConfigReader
+import cats.syntax.all._
+import org.broadinstitute.dsp.{ChartName, ChartVersion}
+import pureconfig.error.ExceptionThrown
+
+import java.nio.file.{Path, Paths}
+
+object ConfigImplicits {
```
rtitle, line 10: Just added because `Config.scala` is getting large? Should we move all the implicits here?

Dockerfile
Qi77Qi: yea, `cp` is what I did the first time...it doesn't work becuz the files don't exist yet

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20210217_add_app_pvc_id.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="leonardo" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+
+    <changeSet logicalFilePath="leonardo" author="qi" id="add_app_pvc_id">
+        <addColumn tableName="APP">
+            <column name="galaxyPvcId" type="varchar(254)">
```
rtitle, line 6: This is probably fine for now, but we should think about data model for `CUSTOM` app type. Maybe we want some different tables containing Galaxy-specific fields.

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/ConfigImplicits.scala
Qi77Qi: I meant to talk about this during walkthrough but we didn't have time...But basically I'm introducing pureconfig in this PR....For new configs, I think we should just use `pureconfig`, which will be less manual. These are implicits for primitive types that can be shared by different projects (when we move multi project structure)

Dockerfile
rtitle: I think the same bucket can have multiple charts

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20210217_add_app_pvc_id.xml
Qi77Qi: I see...we can probably wait until Justin's PR merged and revisit

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/AppComponent.scala
```diff
@@ -181,6 +188,16 @@ object appQuery extends TableQuery(new AppTable(_)) {
       .map(_.status)
       .update(status)
 
+  def updatePvcIds(id: AppId, galaxyPvcId: PvcId, cvmfsPvcId: PvcId)(implicit ec: ExecutionContext): DBIO[Unit] =
+    for {
+      _ <- getByIdQuery(id)
```
rtitle: I think you can update in 1 query instead of 2:
```
getByIdQuery(id).map(x => (x.a, x.b)).update((Some(a), Some(b)))
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -305,32 +296,42 @@ class GKEInterpreter[F[_]: Parallel: ContextShift: Timer](
       gkeClusterId = dbCluster.getGkeClusterId
       googleProject = params.googleProject
 
-      _ <- logger.info(ctx.loggingCtx)(
```
rtitle, line 308: Can we keep this log statement?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -305,32 +296,42 @@ class GKEInterpreter[F[_]: Parallel: ContextShift: Timer](
...
-        s"Beginning app creation for app ${app.appName.value} in cluster ${gkeClusterId.toString}"
-      )
-
       // Create namespace and secrets
       _ <- logger.info(ctx.loggingCtx)(
         s"Creating namespace ${namespaceName.value} and secrets for app ${app.appName.value} in cluster ${gkeClusterId.toString}"
```
rtitle: Maybe remove `and secrets` from this log

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -305,32 +296,42 @@ class GKEInterpreter[F[_]: Parallel: ContextShift: Timer](
...
+          s"Cluster not found in Google: ${gkeClusterId} | trace id: ${ctx.traceId}"
+        )
       )
-      _ <- kubeService.createServiceAccount(gkeClusterId, ksa, KubernetesNamespace(namespaceName))
 
+      nfsDisk <- F.fromOption(
+        dbApp.app.appResources.disk,
+        AppCreationException(s"NFS disk not found in DB for app ${app.appName.value} | trace id: ${ctx.traceId}")
+      )
+
```
rtitle, line 330: Maybe add a log "installing set-up chart"

(There are a lot of steps here and I find logging each step useful)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -387,6 +374,13 @@ class GKEInterpreter[F[_]: Parallel: ContextShift: Timer](
       _ <- logger.info(ctx.loggingCtx)(
         s"Finished app creation for app ${app.appName.value} in cluster ${gkeClusterId.toString}"
       )
+      pvcs <- kubeService.getPersistentVolumeClaims(gkeClusterId, KubernetesNamespace(app.appResources.namespace.name))
+      galaxyPvc = pvcs.find(pvc => pvc.getMetadata.getName == s"${app.release.asString}-galaxy-pvc")
```
rtitle: Are these suffixes defined by the Galaxy chart?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -387,6 +374,13 @@ class GKEInterpreter[F[_]: Parallel: ContextShift: Timer](
...
+      cvmfsPvc = pvcs.find(pvc => pvc.getMetadata.getName == s"${app.release.asString}-cvmfs-alien-cache-pvc")
+      _ <- (galaxyPvc, cvmfsPvc).tupled.traverse {
```
rtitle: Wondering if we should throw an error if we can't retrieve the PVCs

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
Qi77Qi: yea, according to Alex, these are hardcoded

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
Qi77Qi: haha...I can totally do that...was being lazy and punted it lol

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
Qi77Qi: there will be the following log already...do you still want the additional log? I found it a bit too noisy when each step prints out something
```
2021/02/18 17:29:55 creating 3 resource(s)
2021/02/18 17:29:55

Finished installing release:  ya8b8r-gxy-rls-setup-rls
{"@timestamp":"2021-02-18T17:29:55.075Z","@version":"1","message":"The command 'helm installChart' succeeded","logger_name":"org.broadinstitute.dsde.workbench.leonardo.http.Boot","thread_name":"ioapp-compute-4","severity":"INFO","level_value":20000}
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
Qi77Qi: this log was right before logging `Creating namespace` xxx, there's nothing interesting happening between the 2 log lines..so I figured we don't need both of them. I can add `Begin app creation` text to the second line

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -87,11 +79,13 @@ class GKEInterpreter[F[_]: Parallel: ContextShift: Timer](
       nodepools = dbCluster.nodepools
         .filter(n => params.nodepoolsToCreate.contains(n.id))
         .map(buildLegacyGoogleNodepool)
+      _ = println(s"222 ${dbCluster.id} ${dbCluster.nodepools.map(_.id)} ${params.clusterId}")
```
kyuksel: reminder to remove before merging :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
kyuksel: FWIW, when looking through logs having app creation and namespace creation lines separate would make it easier for me to spot things.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -387,6 +378,17 @@ class GKEInterpreter[F[_]: Parallel: ContextShift: Timer](
...
+      galaxyPvc = pvcs.find(pvc => pvc.getMetadata.getName == s"${app.release.asString}-galaxy-pvc")
+      cvmfsPvc = pvcs.find(pvc => pvc.getMetadata.getName == s"${app.release.asString}-cvmfs-alien-cache-pvc")
+      _ <- (galaxyPvc, cvmfsPvc).tupled
+        .fold(F.raiseError[Unit](new LeoException("Fail to retrieve pvc ids", traceId = Some(ctx.traceId)))) {
+          case (gp, cp) =>
+            appQuery
+              .updatePvcIds(params.appId, PvcId(gp.getMetadata.getUid), PvcId(cp.getMetadata.getUid))
+              .transaction
+              .void
+        }
```
kyuksel: Not sure if this logic will work. I think we need to distinguish between two possible cases and take different actions accordingly, i.e., 

1. When creating a Galaxy app **and** a **new** disk, do as above.
2. When creating a Galaxy app **reattaching** an **existing** disk, get the PVC IDs from `PERSISTENT_DISK` table (since we must have executed 1. at some point when first creating the disk)

[link to the related section of the design doc](https://broadworkbench.atlassian.net/wiki/spaces/IA/pages/1184464905/2021-01-05+Galaxy+Persistence#Explanation)

Qi77Qi: we don't have any code for re-attaching disk yet(it'll be in the next PR)...this is PR's goal is really just to retrieve the disk IDs and persist them (prepping for the actual re-attaching implementation)

Qi77Qi: actually, I might just update this to reflect re-attaching in the same PR...

Qi77Qi: I end up making this particular change since it doesn't inflate this PR that much...but don't worry too much about re-attaching logic yet in this PR since it'll be addressed in a follow up PR...

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
rtitle: I think I'd still like an additional `log.info`… the helm client logs tend to be in a different format I think and may be hard to parse (e.g. with `TraceId`).

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/PersistentDiskComponent.scala
```diff
@@ -89,6 +155,23 @@ object persistentDiskQuery extends TableQuery(new PersistentDiskTable(_)) {
...
+    findByIdQuery(id)
+      .map(x => (x.galaxyPvcId, x.cvmfsPvcId, x.chart, x.release))
+      .update(
+        (Some(galaxyDiskRestore.galaxyPvcId),
+         Some(galaxyDiskRestore.cvmfsPvcId),
+         Some(galaxyDiskRestore.chart),
+         Some(galaxyDiskRestore.release))
+      )
+
+  def isUsedByGalaxy(id: DiskId)(implicit ec: ExecutionContext): DBIO[Option[Boolean]] =
```
rtitle, line 164: Wondering if this should return `DBIO[Boolean]`.. is `Option` important?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/PersistentDiskComponent.scala
```diff
@@ -168,7 +251,8 @@ object persistentDiskQuery extends TableQuery(new PersistentDiskTable(_)) {
       disk.size,
       disk.diskType,
       disk.blockSize,
-      disk.formattedBy
+      disk.formattedBy,
+      None //TODO: update this one PersistentDisk model is updated
```
rtitle, line 253: Are you planning to do this? Or should we leave this out of the `PersistentDisk` model?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
rtitle: Am ok with combining the logs below.. mainly just wanted a "begin" signal in the logs.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -387,6 +381,33 @@ class GKEInterpreter[F[_]: Parallel: ContextShift: Timer](
       _ <- logger.info(ctx.loggingCtx)(
         s"Finished app creation for app ${app.appName.value} in cluster ${gkeClusterId.toString}"
       )
+
+      // TODO: this logic will be changed in the next PR
```
rtitle, line 385: How will this change? 

Right now it looks like:
- if the disk already has a PvcId saved (e.g. it was previously attached to a Galaxy) --> do nothing
- if the disk does NOT have a PvcId saved, fetch it from k8s and persist it

rtitle:
I think this looks pretty good so far.. cool that the app tests pass. (RStudio failed but maybe that is transient.)

A bit more needs to be done for persistence - do you prefer to keep adding to this PR, or do you want to try and get this merged as is and open subsequent PRs?

Qi77Qi:
@rtitle if this is mostly good with you, less the restore part (that part will be changed anyway, but I didn't want to pile too many things in this PR), I'd like to have restore logic in a separate PR just so it's easier to review

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
Qi77Qi: it's mostly the criteria for determining whether we need to persist PVCIds could change. Since there's also `createDisk` flag, that could be used as well....I'm trying to not duplicating this info (also not duplicating `chart`, `release` in multiple places) somehow..I don't know exactly what it'll look like yet, hence just left a comment

rtitle:
Sure works for me, as long as app creation works in this PR (the set-up chart stuff looks good).

BTW for persistence, I tagged you in https://github.com/galaxyproject/galaxykubeman-helm/pull/22 because it seemed relevant - we might need to get those `galaxykubeman` PRs merged and tagged to integrate with Leo..

rtitle (approved):
👍  once tests pass

Qi77Qi:
oh yea...forgot about that chart PR....oh well

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/PersistentDiskComponent.scala
Qi77Qi: this function will actually be removed in the follow up PR

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/PersistentDiskComponent.scala
Qi77Qi: I'm not sure yet..but will revisit in the follow up PR

Qi77Qi:
PR should be ready now....Tested in fiab that deleteApp works properly now. The previous bug was because setup chart was deleted before galaxy chart, which relies on a deletion job to run, which I'm guessing depends on ksa to exist. We don't currently catch this in automation tests, I could potentially add a test for this in the follow up PR, but also not sure if this will increase test instability.

When I was trying deleteApp in dev, I ran into the following error from `deleteNamespace` call.
```
{"response":"{\n  \"kind\": \"Status\",\n  \"apiVersion\": \"v1\",\n  \"metadata\": {\n    \n  },\n  \"status\": \"Failure\",\n  \"message\": \"Operation cannot be fulfilled on namespaces \\\"ehawk4-gxy-ns\\\": The system is ensuring all content is removed from this namespace.  Upon completion, this namespace will automatically be purged by the system.\",\n  \"reason\": \"Conflict\",\n  \"details\": {\n    \"name\": \"ehawk4-gxy-ns\",\n    \"kind\": \"namespaces\"\n  },\n  \"code\": 409\n}","result":"Failed"}"
```
vs, with this PR, we get 
```
{"@timestamp":"2021-02-24T01:53:52.406Z","@version":"1","message":"{\"response\":\"Some(())\",\"result\":\"Succeeded\"}","logger_name":"org.broadinstitute.dsde.workbench.leonardo.http.Boot","thread_name":"ioapp-compute-4","severity":"INFO","level_value":20000,"traceId":"b28422af-dd80-4f2c-8a80-0c2ce3e3c0aa","googleCall":"io.kubernetes.client.openapi.apis.CoreV1Api.deleteNamespace(ncenrf-gxy-ns, true, null, null, null, null, null)","duration":"39 milliseconds","result":"Succeeded"}
```


I think this might be due to previously we weren't deleting ksa, secrets explicitly before calling `deleteNamespace`, which I think this PR should make things better

PR #1873, opened 2021-02-17
QA-1385: Debugging Rshiny Test

xuscrbw:
Just setting up the jenkins test, so we can test on the environment.

EDIT: The Original issue with this ticket seems to have gone away (for now). This ticket instead is updated to address the Rshiny issues we have been having. We are noticing that an autocomplete is occurring while typing out the rshiny command. Pauses did not resolve this issue as cleanly, so I found that changing how we are creating the command string seems to have a better effect on consistency. 
---

In all cases:

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/rstudio/RStudioPage.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/rstudio/RStudioPage.scala
@@ -46,7 +46,8 @@ class RStudioPage(override val url: String)(implicit override val authToken: Aut
     // Enter commands to launch the shiny app
     switchToNewTab {
       Thread.sleep(10000)
-      val launchCommand = s"""shiny::runExample("$exampleName", launch.browser = T)"""
+      //val launchCommand = s"""shiny::runExample("$exampleName", launch.browser = T)"""
+      val launchCommand = "shiny::runExample(" + '"' + "01_hello" + '"' + ", launch.browser = T)"
       pressKeys(launchCommand)
       Thread.sleep(10000)
       pressKeys(Keys.ENTER.toString)
--- a/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/rstudio/RStudioPage.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/rstudio/RStudioPage.scala
@@ -47,6 +47,7 @@ class RStudioPage(override val url: String)(implicit override val authToken: Aut
     switchToNewTab {
       val launchCommand = s"""shiny::runExample("$exampleName", launch.browser = T)"""
       pressKeys(launchCommand)
+      Thread.sleep(4000)
       pressKeys(Keys.ENTER.toString)
       await condition windowHandles.size == 2
     }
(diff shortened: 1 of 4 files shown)


xuscrbw:
jenkins retest

xuscrbw:
jenkins retest

xuscrbw:
jenkins retest

xuscrbw:
jenkins retest

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/rstudio/RStudioPage.scala
```diff
@@ -45,7 +45,7 @@ class RStudioPage(override val url: String)(implicit override val authToken: Aut
   def withRShinyExample[T](exampleName: String)(testCode: RShinyPage => T): T = {
     // Enter commands to launch the shiny app
     switchToNewTab {
-      val launchCommand = s"""shiny::runExample("$exampleName", launch.browser = T)"""
+      val launchCommand = "shiny::runExample(" + '"' + exampleName + '"' + ", launch.browser = T)"
```
rtitle: Is this actually any different? Seems this would result in the same string:
```
scala> val exampleName = "foo"
exampleName: String = foo

scala> val launchCommand1 = s"""shiny::runExample("$exampleName", launch.browser = T)"""
launchCommand1: String = shiny::runExample("foo", launch.browser = T)

scala> val launchCommand2 = "shiny::runExample(" + '"' + exampleName + '"' + ", launch.browser = T)"
launchCommand2: String = shiny::runExample("foo", launch.browser = T)

scala> launchCommand1 == launchCommand2
res0: Boolean = true
```

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/rstudio/RStudioPage.scala
```diff
@@ -45,7 +45,7 @@ class RStudioPage(override val url: String)(implicit override val authToken: Aut
...
       pressKeys(launchCommand)
       pressKeys(Keys.ENTER.toString)
```
rtitle, line 54: I was thinking instead of 2 `pressKeys` calls we could try something like:
```
val launchCommand = s"""shiny::runExample("$exampleName", launch.browser = T)${Keys.ENTER.toString}"""
pressKeys(launchCommand)
```
Not sure if that will help or not..

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/rstudio/RStudioPage.scala
xuscrbw: Correct, I don't think the issue was with the string itself. I tried this change to see if maybe how the string was created would result in it being processed different. Looking that this has errored out on the most recent retest, it does seem like it does not avoid the issue.

rtitle (approved):
:shipit: if tests pass

xuscrbw:
jenkins retest

xuscrbw:
jenkins retest

xuscrbw:
jenkins retest

xuscrbw:
jenkins retest

PR #1890, opened 2021-02-24
IA-2052 IA-2535 Support galaxy restore data from disk 

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-2052
https://broadworkbench.atlassian.net/browse/IA-2535

1. Use new version of helm-scala-sdk with `createNamespace` option enabled. This automatically adds `name` label for the namespace created as well, which is required for enabling network policy in the future.
2. for automation test, switch initial runtime to a gce runtime instead of dataproc to reduce overhead. If we see increased dataproc related test failure, we should switch back.
3. Make `deleteDisk` flag work.
4. Enabled logging header in automation tests. This way, we'll be able to see traceId(`X-Cloud-Trace-Context`) used in automation tests, and use that to look for relevant leo logs. Example, 
```
11:24:35 [scala-execution-context-global-207      ] INFO  [o.h.c.m.RequestLogger    :$anonfun$impl$3] - HTTP/1.1 GET https://leonardo-fiab.dsde-dev.broadinstitute.org:30443/api/google/v1/apps/gpalloc-dev-master-e45nldi/automation-test-app-aj1jfzkkz Headers(Authorization: <REDACTED>, X-Cloud-Trace-Context: c4a6d474-6893-4c8c-a361-322b0f49fd25, Accept: application/json) body=""
```
5.  Support restore from disk flow

---

In all cases:

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoApiClient.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoApiClient.scala
@@ -2,12 +2,12 @@ package org.broadinstitute.dsde.workbench.leonardo
 
 import java.util.UUID
 import java.util.concurrent.TimeoutException
-
 import cats.effect.{IO, Resource, Timer}
 import cats.syntax.all._
+import io.chrisdavenport.log4cats.StructuredLogger
 import org.broadinstitute.dsde.workbench.DoneCheckable
 import org.broadinstitute.dsde.workbench.DoneCheckableSyntax._
-import org.broadinstitute.dsde.workbench.google2.{streamFUntilDone, DiskName, MachineTypeName}
+import org.broadinstitute.dsde.workbench.google2.{streamFUntilDone, streamUntilDoneOrTimeout, DiskName, MachineTypeName}
 import org.broadinstitute.dsde.workbench.leonardo.ApiJsonDecoder._
 import org.broadinstitute.dsde.workbench.leonardo.http.AppRoutesTestJsonCodec._
 import org.broadinstitute.dsde.workbench.leonardo.http.DiskRoutesTestJsonCodec._
@@ -43,7 +43,7 @@ object LeonardoApiClient {
   val client: Resource[IO, Client[IO]] = for {
     blockingEc <- ExecutionContexts.cachedThreadPool[IO]
     client <- blaze.BlazeClientBuilder[IO](blockingEc).resource
-  } yield Logger[IO](logHeaders = false, logBody = true)(client)
+  } yield Logger[IO](logHeaders = true, logBody = true)(client)
 
   val rootUri = Uri.unsafeFromString(LeonardoConfig.Leonardo.apiUrl)
   val defaultCreateDiskRequest = CreateDiskRequest(
@@ -230,6 +230,34 @@ object LeonardoApiClient {
               s"create runtime ${googleProject.value}/${runtimeName.asString}. Runtime is still in ${other}"
             )
           )
+      }
+    } yield res
+  }
+
+  def waitUntilAppRunning(googleProject: GoogleProject, appName: AppName, shouldError: Boolean = true)(
+    implicit timer: Timer[IO],
+    client: Client[IO],
+    authHeader: Authorization,
+    logger: StructuredLogger[IO]
+  ): IO[GetAppResponse] = {
+    val ioa = getApp(googleProject, appName)
+    implicit val doneCheckeable: DoneCheckable[GetAppResponse] = x =>
+      x.status == AppStatus.Running || x.status == AppStatus.Error
+    for {
+      res <- timer.sleep(80 seconds) >> streamUntilDoneOrTimeout(
+        ioa,
+        120,
+        10 seconds,
+        s"app ${googleProject.value}/${appName.value} did not finish app creation after 20 minutes."
+      )
+      _ <- res.status match {
+        case AppStatus.Error =>
+          if (shouldError)
+            IO.raiseError(
+              new RuntimeException(s"${googleProject.value}/${appName.value} errored due to ${res.errors}")
+            )
+          else logger.info(s"${googleProject.value}/${appName.value} errored due to ${res.errors}")
+        case _ => IO.unit
       }
     } yield res
   }
@@ -424,8 +452,19 @@ object LeonardoApiClient {
         }
     } yield r
 
-  def deleteApp(googleProject: GoogleProject, appName: AppName)(implicit client: Client[IO],
-                                                                authHeader: Authorization): IO[Unit] =
+  def createAppWithWait(
+    googleProject: GoogleProject,
+    appName: AppName,
+    createAppRequest: CreateAppRequest = defaultCreateAppRequest
+  )(implicit client: Client[IO], authHeader: Authorization, logger: StructuredLogger[IO], timer: Timer[IO]): IO[Unit] =
+    for {
+      _ <- createApp(googleProject, appName, createAppRequest)
+      _ <- waitUntilAppRunning(googleProject, appName, true)
+    } yield ()
+
+  def deleteApp(googleProject: GoogleProject,
+                appName: AppName,
+                deleteDisk: Boolean = true)(implicit client: Client[IO], authHeader: Authorization): IO[Unit] =
     for {
       traceIdHeader <- genTraceIdHeader()
       r <- client
@@ -433,7 +472,9 @@ object LeonardoApiClient {
           Request[IO](
             method = Method.DELETE,
             headers = Headers.of(authHeader, traceIdHeader),
-            uri = rootUri.withPath(s"/api/google/v1/apps/${googleProject.value}/${appName.value}")
+            uri = rootUri
+              .withPath(s"/api/google/v1/apps/${googleProject.value}/${appName.value}")
+              .withQueryParam("deleteDisk", deleteDisk)
           )
         )
         .use { resp =>
@@ -445,13 +486,13 @@ object LeonardoApiClient {
         }
     } yield r
 
-  def deleteAppWithWait(googleProject: GoogleProject, appName: AppName)(
+  def deleteAppWithWait(googleProject: GoogleProject, appName: AppName, deleteDisk: Boolean = true)(
     implicit timer: Timer[IO],
     client: Client[IO],
     authHeader: Authorization
   ): IO[Unit] =
     for {
-      _ <- deleteApp(googleProject, appName)
+      _ <- deleteApp(googleProject, appName, deleteDisk)
       ioa = getApp(googleProject, appName).attempt
       res <- timer.sleep(120 seconds) >> streamFUntilDone(ioa, 30, 30 seconds).compile.lastOrError
       _ <- if (res.isDone) IO.unit
--- a/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/apps/AppCreationSpec.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/apps/AppCreationSpec.scala
@@ -2,7 +2,7 @@ package org.broadinstitute.dsde.workbench.leonardo
 package apps
 
 import org.broadinstitute.dsde.workbench.DoneCheckable
-import org.broadinstitute.dsde.workbench.google2.{streamFUntilDone, streamUntilDoneOrTimeout}
+import org.broadinstitute.dsde.workbench.google2.{streamFUntilDone, streamUntilDoneOrTimeout, Generators}
 import org.broadinstitute.dsde.workbench.leonardo.LeonardoApiClient._
 import org.broadinstitute.dsde.workbench.leonardo.http.{ListAppResponse, PersistentDiskRequest}
 import org.broadinstitute.dsde.workbench.service.util.Tags
@@ -17,19 +17,22 @@ class AppCreationSpec extends GPAllocFixtureSpec with LeonardoTestUtils with GPA
   implicit val auth: Authorization =
     Authorization(Credentials.Token(AuthScheme.Bearer, ronCreds.makeAuthToken().value))
 
-  "create and delete an app" in { _ =>
+  "create, delete an app and re-create an app with same disk" taggedAs Tags.SmokeTest in { _ =>
     withNewProject { googleProject =>
       val appName = randomAppName
+      val restoreAppName = AppName(s"restore-${appName.value}")
+      val diskName = Generators.genDiskName.sample.get
 
       val createAppRequest = defaultCreateAppRequest.copy(
         diskConfig = Some(
           PersistentDiskRequest(
-            randomDiskName,
-            Some(DiskSize(500)),
+            diskName,
+            Some(DiskSize(300)),
             None,
             Map.empty
           )
-        )
+        ),
+        customEnvironmentVariables = Map("WORKSPACE_NAME" -> "Galaxy-Workshop-ASHG_2020_GWAS_Demo")
       )
 
       LeonardoApiClient.client.use { implicit client =>
@@ -45,7 +48,7 @@ class AppCreationSpec extends GPAllocFixtureSpec with LeonardoTestUtils with GPA
           _ = getAppResponse.status should (be(AppStatus.Provisioning) or be(AppStatus.Precreating))
 
           // Verify the app eventually becomes Running
-          _ <- testTimer.sleep(60 seconds)
+          _ <- testTimer.sleep(120 seconds)
           monitorCreateResult <- streamUntilDoneOrTimeout(
             getApp,
             120,
@@ -60,7 +63,7 @@ class AppCreationSpec extends GPAllocFixtureSpec with LeonardoTestUtils with GPA
           _ <- testTimer.sleep(1 minute)
 
           // Delete the app
-          _ <- LeonardoApiClient.deleteApp(googleProject, appName)
+          _ <- LeonardoApiClient.deleteApp(googleProject, appName, false)
 
           // Verify getApp again
           getAppResponse <- getApp
@@ -85,6 +88,8 @@ class AppCreationSpec extends GPAllocFixtureSpec with LeonardoTestUtils with GPA
               s"AppCreationSpec: app ${googleProject.value}/${appName.value} delete result: $monitorDeleteResult"
             )
           }
+          // Verify creating another app with the same disk doesn't error out
+          _ <- LeonardoApiClient.createAppWithWait(googleProject, restoreAppName, createAppRequest)
         } yield ()
       }
     }
@@ -93,11 +98,12 @@ class AppCreationSpec extends GPAllocFixtureSpec with LeonardoTestUtils with GPA
   "stop and start an app" taggedAs Tags.SmokeTest in { _ =>
     withNewProject { googleProject =>
       val appName = randomAppName
+      val diskName = Generators.genDiskName.sample.get
 
       val createAppRequest = defaultCreateAppRequest.copy(
         diskConfig = Some(
           PersistentDiskRequest(
-            randomDiskName,
+            diskName,
             Some(DiskSize(500)),
             None,
             Map.empty
@@ -118,7 +124,7 @@ class AppCreationSpec extends GPAllocFixtureSpec with LeonardoTestUtils with GPA
           _ = getAppResponse.status should (be(AppStatus.Provisioning) or be(AppStatus.Precreating))
 
           // Verify the app eventually becomes Running
-          _ <- testTimer.sleep(60 seconds)
+          _ <- testTimer.sleep(90 seconds)
           monitorCreateResult <- streamUntilDoneOrTimeout(
             getApp,
             120,
@@ -138,7 +144,7 @@ class AppCreationSpec extends GPAllocFixtureSpec with LeonardoTestUtils with GPA
           _ = getAppResponse.status should (be(AppStatus.Stopping) or be(AppStatus.PreStopping))
 
           // Verify the app eventually becomes Stopped
-          _ <- testTimer.sleep(30 seconds)
+          _ <- testTimer.sleep(60 seconds)
           monitorStopResult <- streamUntilDoneOrTimeout(
             getApp,
             180,
@@ -173,7 +179,8 @@ class AppCreationSpec extends GPAllocFixtureSpec with LeonardoTestUtils with GPA
           _ <- testTimer.sleep(1 minute)
 
           // Delete the app
-          _ <- LeonardoApiClient.deleteApp(googleProject, appName)
+          _ <- LeonardoApiClient.deleteApp(googleProject, appName, true)
+          _ <- testTimer.sleep(30 seconds)
 
           // Verify getApp again
           getAppResponse <- getApp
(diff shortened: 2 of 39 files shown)


Qi77Qi:
```
java.sql.SQLTransientConnectionException: mysql.db - Connection is not available, request timed out after 5001ms
```

jenkins retest

Qi77Qi:
```
err is %!(EXTRA *errors.withStack=rendered manifests contain a resource that already exists. Unable to continue with install: could not get information about the resource: an error on the server ("Internal Server Error: \"/apis/apps/v1/namespaces/evfm8i-gxy-ns/deployments/evfm8i-gxy-rls-galaxy-workflow\": the server is currently unable to handle the request") has prevented the request from succeeding (get deployments.apps evfm8i-gxy-rls-galaxy-workflow), string=
```

just to see if this is just transient...jenkins retest

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppService.scala
```diff
@@ -6,7 +6,7 @@ import org.broadinstitute.dsde.workbench.leonardo.{AppContext, AppName}
 import org.broadinstitute.dsde.workbench.model.UserInfo
 import org.broadinstitute.dsde.workbench.model.google.GoogleProject
 
-trait KubernetesService[F[_]] {
+trait AppService[F[_]] {
```
rtitle, line 9: Thanks for renaming

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
```diff
@@ -124,13 +124,56 @@ final class LeoKubernetesServiceInterp[F[_]: Parallel](
         )
       )
 
+      lastUsedApp <- diskResultOpt.flatTraverse { diskResult =>
+        if (diskResult.creationNeeded) F.pure(none[LastUsedApp])
+        else {
+          (diskResult.disk.formattedBy, diskResult.disk.galaxyRestore) match {
```
rtitle, line 132: The check for `formattedBy` happens in `RuntimeServiceInterp.processPersistentDiskRequest` -- not sure it needs to be repeated here? Feels like this code could just match on `diskResult.disk.galaxyRestore` unless I'm missing something..

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
```diff
@@ -509,30 +553,39 @@ final class LeoKubernetesServiceInterp[F[_]: Parallel](
...
+        NamespaceId(-1)
+      )(app => app.namespaceId)
+      namespaceName <- lastUsedApp.fold(
+        KubernetesName.withValidation(
+          s"${uid}-${galaxyConfig.namespaceNameSuffix}",
+          NamespaceName.apply
+        )
+      )(app => app.namespaceName.asRight[Throwable])
+
+      chart = lastUsedApp.map(_.chart).getOrElse(Chart(galaxyConfig.chartName, galaxyConfig.chartVersion))
```
rtitle: This would mean if we update the `galaxykubeman` chart version in Leo conf, it would not take effect for users who have an existing disk. 

Am a little concerned about this -- e.g. what if we need to roll out a gkm bugfix that does not affect disks in any way? Feels like we should have a way to do that..

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -806,7 +814,11 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift: Parallel](
         // create and monitor app
         _ <- gkeInterp
           .createAndPollApp(CreateAppParams(msg.appId, msg.project, msg.appName))
-          .onError { case _ => cleanUpAfterCreateAppError(msg.appId, msg.appName, msg.project, msg.createDisk) }
+          .onError {
+            case e =>
+              println(e)
```
rtitle: println here :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -783,7 +791,7 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift: Parallel](
       }
 
       // create second Galaxy disk asynchronously
-      createSecondDiskOp = if (msg.appType == Galaxy) {
+      createSecondDiskOp = if (msg.appType == Galaxy && msg.createDisk.isDefined) {
```
rtitle, line 816: Do we need to check `msg.createDisk.isDefined` above (on line 782) as well?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -931,28 +942,28 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift: Parallel](
...
-                .void
-          }
-        else F.unit
+
         // we now use the detach timestamp recorded prior to helm uninstall so we can observe when galaxy actually 'detaches' the disk from google's perspective
         getDisk = googleDiskService.getDisk(msg.project,
                                             zone,
                                             gkeInterp.getGalaxyPostgresDiskName(dbApp.app.appResources.namespace.name))
-        diskDetachResult <- streamUntilDoneOrTimeout(
+        _ <- streamUntilDoneOrTimeout(
```
rtitle: I guess we only need to do this "wait for detach" logic if we're actually deleting the disk.. probably doesn't hurt to do it either way though.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -365,26 +367,33 @@ class GKEInterpreter[F[_]: Parallel: ContextShift: Timer](
...
       )
 
       _ <- logger.info(ctx.loggingCtx)(
         s"Finished app creation for app ${app.appName.value} in cluster ${gkeClusterId.toString}"
       )
 
-      // TODO: this logic will be changed in the next PR
-      isUsedByGalaxy <- persistentDiskQuery.isUsedByGalaxy(diskId).transaction
-      _ <- if (isUsedByGalaxy.exists(identity)) F.unit
+      _ <- if (galaxyRestore.isDefined)
```
rtitle, line 415: Even if this is _not_ a restore disk we want to update `lastUsedBy`, right?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
Qi77Qi: yea...I asked Alex about versioning..he did mention if it's patch upgrade, it should be non-breaking in theory...we definitely need to think about this a bit more...I'm thinking we can table this for this PR, and get it finalized before UI is in prod...I did create https://broadworkbench.atlassian.net/browse/IA-2558

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -764,18 +773,20 @@ class GKEInterpreter[F[_]: Parallel: ContextShift: Timer](
...
           config.ingressConfig.chartVersion,
-          org.broadinstitute.dsp.Values(config.ingressConfig.values.map(_.value).mkString(","))
+          org.broadinstitute.dsp.Values(config.ingressConfig.values.map(_.value).mkString(",")),
+          true
         )
         .run(helmAuthContext)
 
       // Monitor nginx until public IP is accessible
-      loadBalancerIpOpt <- streamFUntilDone(
+      loadBalancerIpOpt <- streamUntilDoneOrTimeout(
```
rtitle: This would now throw a `StreamTimeoutError` I think if the IP does not become available -- could we make it throw `ClusterCreationException` as before? (I think there is some handling attached to these exception types.)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
rtitle: OK, thanks for making that ticket.. yeah maybe a policy like "patch versions are OK, major versions require a new disk" would make sense. There are also UX considerations as ideally users would be able to opt in to upgrading their Galaxy.

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/apps/AppCreationSpec.scala
```diff
@@ -84,6 +88,8 @@ class AppCreationSpec extends GPAllocFixtureSpec with LeonardoTestUtils with GPA
               s"AppCreationSpec: app ${googleProject.value}/${appName.value} delete result: $monitorDeleteResult"
             )
           }
+          // Verify creating another app with the same disk doesn't errorMockAppServiceInterp
```
kyuksel:
```suggestion
          // Verify creating another app with the same disk doesn't error out
```
?

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/diskModels.scala
```diff
@@ -105,3 +106,6 @@ object FormattedBy extends Enum[FormattedBy] {
     override def asString: String = "GALAXY"
   }
 }
+
+final case class PvcId(asString: String) extends AnyVal
+final case class GalaxyRestore(galaxyPvcId: PvcId, cvmfsPvcId: PvcId, lastUsedBy: AppId)
```
kyuksel, line 116: Consider using a noun since it's a class (e.g. `GalaxyDiskToRestore`)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
Qi77Qi: so upgrading galaxy itself doesn't necessarily require chart upgrade..created https://broadworkbench.atlassian.net/browse/IA-2557 to track

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
Qi77Qi: that makes sense..after a closer look, `streamFUntilDone` is probably better anyway becuz in this case, "doneness" is whether the returned option becomes `Some`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
Qi77Qi: yea...basically `lastUsedBy` is updated during app creation, if app1 is created using a disk, this disk's `lastUsedBy` becomes this app1

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
Qi77Qi: nope..`createDiskForApp` function itself is doing the check already..this is why I prefer having the important logic in for comprehension instead of having it in small functions like this, kinda makes things confusing (having to look at the function to realize it's already doing the check)..I'll do a small refactoring

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
kyuksel: Added this comment to the ticket:

[Echoing Alex](https://the-anvil.slack.com/archives/CUYUCUYCB/p1614356649016500), it feels like we should consider implementing chart/Galaxy upgrading separately from persistence, for at least three reasons:

1. Galaxy is much more likely to come up while restoring an existing disk if we use the same chart and values.

2. UX-wise, it would be more clear if users explicitly upgrade when they want to

3. If something fails, debugging can be more difficult if we’re upgrading while restoring.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -783,7 +791,7 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift: Parallel](
       }
 
       // create second Galaxy disk asynchronously
-      createSecondDiskOp = if (msg.appType == Galaxy) {
```
kyuksel, line 789: I wonder if we should rename to use the terms `data` and `metadata` disk going forward instead of `nfs` and `second/postgres` disk, respectively.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
Qi77Qi: there might be a bit overlapping, but not repeated really...`processPersistentDiskRequest` only cares about `formattedBy` and whether a disk is attached....here, we're checking between `formattedBy` and `galaxyRestore`...I don't want to mix `galaxyRestore` into `processPersistentDiskRequest` since it's used for other flow as well

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
kyuksel: Shouldn't we be calling `persistentDiskQuery.updateLastUsedBy()` in the `else` clause below too (since we want to make `lastUpdatedBy = appX` when `appX` is not doing a disk restoration)?

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/diskModels.scala
Qi77Qi: I moved away from having `disk` in the name because this data type represent all info needed for restoring a galaxy app, not just disk info.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
Qi77Qi: this PR is already pretty big..so I'm inclined to not introduce more renaming

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
Qi77Qi: we are

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/diskModels.scala
kyuksel: Got it. Would you mind adding comments about the role of the class in Galaxy restoration since it's not straightforward?

I'd also still encourage you to use nouns for class names whenever possible.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
kyuksel: This PR is indeed too big as is, and I agree that it'd be nice to not make the diff larger. At the same time, this sort of thing tends to never get done later and becomes confusing during onboarding. I wonder if it's worth creating a tech-debt ticket.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
kyuksel: Where in the `createDiskForApp` function is that check being done again?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
kyuksel: ah I see

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
Qi77Qi: per my previous comment, I refactored `createDiskForApp`. The original function looks like [this](https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala#L1087), which is already doing the check.

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/diskModels.scala
Qi77Qi: I'm happy to rename it if you have a suggestion here

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
kyuksel: oh okay, thanks…I was afraid I was losing my mind 🤯

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/diskModels.scala
kyuksel: Some options:

`GalaxyPersistence`
`GalaxyRestoration`
`GalaxyRecovery`

Could also append to those above `Info` (e.g. `GalaxyRestorationInfo`) for further clarity (at the expense of verbosity).

kyuksel: > I moved away from having disk in the name because this data type represent all info needed for restoring a galaxy app, not just disk info.

On a second thought - I'm not sure if that's accurate. We also need `chart` and `release` to restore a Galaxy app, which are not included in the class.

Qi77Qi: hmmm..not sure I like any of the names...

GalaxyPersistence: we've been using persistent disk...so sayng "persistence" just adds more confusion.
GalaxyRestoration: no specific reason. but it's longer and `GalaxyRestore` reads a lot better to me
GalaxyRecovery: we've been using `restore` everywhere else. I don't want to introduce yet another name for the same thing

note, we have plenty other data types that are named with verbs, which I think are perfectly fine.

> On a second thought - I'm not sure if that's accurate. We also need chart and release to restore a Galaxy app, which are not included in the class.

that's what `lastUsedApp` is for

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoApiClient.scala
```diff
@@ -43,7 +43,7 @@ object LeonardoApiClient {
   val client: Resource[IO, Client[IO]] = for {
     blockingEc <- ExecutionContexts.cachedThreadPool[IO]
     client <- blaze.BlazeClientBuilder[IO](blockingEc).resource
-  } yield Logger[IO](logHeaders = false, logBody = true)(client)
+  } yield Logger[IO](logHeaders = true, logBody = true)(client)
```
rtitle, line 46: Does this log the token? Might be good to redact it if we can.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
```diff
@@ -510,30 +558,47 @@ final class LeoKubernetesServiceInterp[F[_]: Parallel](
...
+      namespaceName <- lastUsedApp.fold(
+        KubernetesName.withValidation(
+          s"${uid}-${galaxyConfig.namespaceNameSuffix}",
+          NamespaceName.apply
+        )
+      )(app => app.namespaceName.asRight[Throwable])
+
+      chart = lastUsedApp
+        .map { lastUsed =>
+          // if there's a patch version bump, then we use the later version defined in leo config; otherwise, we use lastUsed chart definition
```
rtitle, line 573: Cool, thanks for adding this

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
rtitle: ok - yeah looks like `msg.createDisk` is being checked in both places.

rtitle (approved):
Looks good, looks like my earlier comments were all addressed 👍 

Seems there is a Galaxy test failure though..

rtitle:
@Qi77Qi just confirming -- does the UI currently pass `deleteDisk=true`? In other words, this PR should have no effect in the UI currently, right?

kyuksel:
> does the UI currently pass deleteDisk=true? In other words, this PR should have no effect in the UI currently, right?

Seems like it:
https://github.com/DataBiosphere/terra-ui/blob/76e96d263204d6660bda567d6d8811b349af5626/src/libs/ajax.js#L1213

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoApiClient.scala
Qi77Qi: it is redacted automatically! I think it might be new thing in newer library version cuz I remember I didn't enable this previously because of that...

This is really nice now since we can see `traceId` from these requests

kyuksel (approved):

Qi77Qi:
jenkins retest

Qi77Qi:
```
Request to https://leonardo-fiab.dsde-qa.broadinstitute.org:30443 timed out after 45000 ms
```

jenkins retest

Qi77Qi:
```
{"@timestamp":"2021-03-11T16:53:49.872Z","@version":"1","message":"Failure checking status for subsystem [Sam]: TickWheelExecutor is shutdown","logger_name":"org.broadinstitute.dsde.workbench.leonardo.http.Boot","thread_name":"leonardo-akka.actor.default-dispatcher-16","severity":"WARN","level_value":30000}
```

jenkins retest

rtitle:
`TickWheelExecutor is shutdown` might indicate an unhandled Leo exception crashing the app.. @jdcanas hit this recently too.

Qi77Qi:
jenkins retest

Qi77Qi:
tests passed...run automation tests one more time...jenkins retest

Qi77Qi:
flaky test in `org.broadinstitute.dsde.workbench.leonardo.apps.CustomAppCreationSpec.create and delete a custom app`

jenkins retest

PR #1903, opened 2021-03-02
IA-2566 Log referers when we reject requests

Qi77Qi:

---

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/ProxyRoutes.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/ProxyRoutes.scala
@@ -262,18 +262,21 @@ class ProxyRoutes(proxyService: ProxyService, corsSupport: CorsSupport, refererC
       pass
     }
 
+  private def requestPath(req: HttpRequest): String = req.uri.toString
+  private val logRequestPath = DebuggingDirectives.logRequest(requestPath _)
+
   private[api] def checkReferer: Directive0 =
     optionalHeaderValueByType(Referer) flatMap {
       case Some(referer) => {
         if (refererConfig.validHosts.contains(referer.uri.authority.toString())) pass
         else {
-          logger.info(s"${referer} is not allowed")
-          logRequest(failWith(AuthenticationError()))
+          logger.info(s"Referer ${referer} is not allowed")
+          logRequestPath.tflatMap(_ => reject)
         }
       }
       case None => {
         logger.info(s"Referer header is missing")
-        logRequest(failWith(AuthenticationError()))
+        logRequestPath.tflatMap(_ => reject)
       }
     }
 }
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/ProxyRoutes.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/ProxyRoutes.scala
@@ -266,9 +266,13 @@ class ProxyRoutes(proxyService: ProxyService, corsSupport: CorsSupport, refererC
     optionalHeaderValueByType(Referer) flatMap {
       case Some(referer) => {
         if (refererConfig.validHosts.contains(referer.uri.authority.toString())) pass
-        else failWith(AuthenticationError())
+        else {
+          logger.info(s"${referer} is not allowed")
+          failWith(AuthenticationError())
+        }
       }
       case None => {
+        logger.info(s"Referer header is missing")
         failWith(AuthenticationError())
       }
     }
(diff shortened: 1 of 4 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/ProxyRoutes.scala
```diff
@@ -266,9 +266,13 @@ class ProxyRoutes(proxyService: ProxyService, corsSupport: CorsSupport, refererC
     optionalHeaderValueByType(Referer) flatMap {
       case Some(referer) => {
         if (refererConfig.validHosts.contains(referer.uri.authority.toString())) pass
-        else failWith(AuthenticationError())
+        else {
+          logger.info(s"${referer} is not allowed")
```
gpcarr: these logs are only accessible by us internally correct?

Qi77Qi: yea...users can see the error we're returning..but they can never see leonardo logs

rtitle:
Any way we could log the request path too? Might be interesting to see which exact calls have missing/invalid referers.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/ProxyRoutes.scala
gpcarr: maybe we could change this to `logger.info(s"${referer} is not a valid Referer header")` so that it's easier to search the logs for the string "Referer header" and find all referer logs with the same grep

rtitle: +1, would be good to include the string "Referer" in this log msg

Qi77Qi:
@rtitle @gpcarr let me know if there's any other comments

gpcarr:
> @rtitle @gpcarr let me know if there's any other comments

lgtm

EDIT one more question

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/ProxyRoutes.scala
```diff
@@ -262,14 +262,21 @@ class ProxyRoutes(proxyService: ProxyService, corsSupport: CorsSupport, refererC
...
-        else failWith(AuthenticationError())
+        else {
+          logger.info(s"Referer ${referer} is not allowed")
+          logRequestPath.tflatMap(_ => reject)
+        }
       }
       case None => {
-        failWith(AuthenticationError())
+        logger.info(s"Referer header is missing")
+        logRequestPath.tflatMap(_ => reject)
```
gpcarr: does `reject` fail with 401?

Qi77Qi: it's probably 40x...I'm assuming this shouldn't matter...also didn't really test out the logging part since you'll definitely be testing this out :)

gpcarr: sounds good although I would run unit test `ProxyRoutesSpec` locally since the last test in there validates that `StatusCodes.Unauthorized` happens to make sure it still passes

gpcarr: actually looks like the test is failing, once it's fixed we should know what status code gets returned

gpcarr: it's 405:
- should 401 when Referer is not a valid URI *** FAILED ***
16:15:00 [info]   405 Method Not Allowed did not equal 401 Unauthorized (ProxyRoutesSpec.scala:557)

rtitle: I think we want 401 in this case.. is there a reason to use `reject`? Can we change back to `failWith(AuthenticationError())`?

gpcarr: +1 for failing with 401

Qi77Qi: types don't match for `failWith(AuthenticationError())`...there might be a way to make it work..but my IJ is really messed up today, so I can't really figure things out easily...if you guys want, feel free to update the PR....

What's the reason for 401 tho? it's not really an auth error either

gpcarr: i'll push up the change

gpcarr: the types matched up for me and i checked to make sure everything compiles and test passes - maybe it's an issue with your IJ

gpcarr (approved):

rtitle (approved):
:shipit:

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/ProxyRoutes.scala
Qi77Qi: interesting...okay...

still not convinced why 401 is preferred over `reject` tho 🤷‍♀️

gpcarr:
jenkins retest - fiab start failure

Qi77Qi:
jenkins retest

PR #1904, opened 2021-03-03
[IA-2400] Jc custom app

jdcanas:
What this PR does:

- Adds the `Custom` `appType` as a valid value for the createApp endpoint 
  - also adds various fields to this endpoint, details in swagger

This is tested by using the terra-app helm chart, and referencing the `app.yaml` located in that same repo, (e.g. https://raw.githubusercontent.com/DataBiosphere/terra-app/main/apps/ucsc_genome_browser/app.yaml) 

This was tested and verified with the ucsc, rstudio, and cirrocumulus `app.yaml` files in this manner. 

For example, 
```
curl -X POST --header 'Content-Type: application/json' --header "Authorization: Bearer `gcloud auth print-access-token`" -d @custom-app.json  https://leonardo-fiab.dsde-dev.broadinstitute.org:30443/api/google/v1/apps/jc-custom3/uscs2
```
where `custom-app.json` contains
```
{
 "diskConfig": {"name": "disk8"},
  "appType": "CUSTOM",
  "descriptorPath": "https://raw.githubusercontent.com/DataBiosphere/terra-app/main/apps/ucsc_genome_browser/app.yaml"
}
```

Things this PR does not address:

- Do custom app disks need different formatting that leo should persist?
  - i.e., does an app with `appType` `CUSTOM` and the `descriptor` for `Jupyter` need a different `formattedBy` than an app with `appType` `CUSTOM` and the `descriptor` for `rstudio`
- Automation testing. I plan to do it as part of this ticket with a separate PR. 
- CI - how the terra-app helm chart or app.yaml files will be published
- Any sort of validation of the custom app descriptor passed in (besides basic URI validation)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/AppDescriptorDAO.scala
```diff
@@ -0,0 +1,29 @@
...
+import cats.mtl.Ask
+import org.broadinstitute.dsde.workbench.leonardo.ContainerImage
+import org.broadinstitute.dsde.workbench.leonardo.model.LeoException
+import org.broadinstitute.dsde.workbench.model.TraceId
+
+trait AppDescriptorDAO[F[_]] {
+  def getDescriptor(path: String)(implicit ev: Ask[F, TraceId]): F[AppDescriptor]
+}
+
+final case class AppDescriptor(name: String,
```
rtitle, line 16: Maybe add a comment that this matches the schema here: https://github.com/DataBiosphere/terra-app#app-schema

I expect the schema will evolve going forward, so we'll need to keep the models in sync.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HTTPAppDescriptorDAO.scala
```diff
@@ -0,0 +1,78 @@
+package org.broadinstitute.dsde.workbench.leonardo
```
rtitle: Just a nit on naming: we usually lower-case `Http`, so `HttpAppDescriptorDAO`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HTTPAppDescriptorDAO.scala
```diff
@@ -0,0 +1,78 @@
...
+import org.broadinstitute.dsde.workbench.leonardo.JsonCodec._
+import org.broadinstitute.dsde.workbench.leonardo.dao.HttpAppDescriptorDAO._
+import org.http4s.{Method, Request, Response, Uri}
+import org.http4s.client.Client
+
+class HttpAppDescriptorDAO[F[_]](httpClient: Client[F])(implicit logger: StructuredLogger[F], F: Effect[F])
+    extends AppDescriptorDAO[F] {
+  override def getDescriptor(path: String)(implicit ev: Ask[F, TraceId]): F[AppDescriptor] =
+    for {
+      _ <- logger.info(s"Attempting to retrieve descriptor $path")
```
rtitle: I think some (all?) of these logs can be removed

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HTTPAppDescriptorDAO.scala
```diff
@@ -0,0 +1,78 @@
...
+          uri = Uri.unsafeFromString(path)
+        )
+      )(onError(path))
+    } yield resp
+
+  private def onError(path: String)(response: Response[F])(implicit ev: Ask[F, TraceId]): F[Throwable] =
+    for {
+      traceId <- ev.ask
+      body <- response.bodyText.compile.foldMonoid
+      _ <- logger.error(s"${traceId} | call failed: $body")
```
rtitle: Instead of embedding the trace id in the string, do this instead:

`logger.error(ctx.loggingCtx)(s"call failed: $body")`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HTTPAppDescriptorDAO.scala
```diff
@@ -0,0 +1,78 @@
...
+      description <- d.downField("description").as[String]
+      version <- d.downField("version").as[String]
+      services <- d.downField("services").as[Map[String, CustomAppService]]
+    } yield AppDescriptor(name, author, description, version, services)
+  }
+
+  implicit val customAppServiceDecoder: Decoder[CustomAppService] = Decoder.instance { d =>
+    for {
+      image <- d.downField("image").as[ContainerImage]
+      port <- d.downField("port").as[Option[Int]]
```
rtitle: I think `port` is required (prob shouldn't default to 80 -- 80 may not work anyway due to ingress)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HTTPAppDescriptorDAO.scala
```diff
@@ -0,0 +1,78 @@
...
+      baseUrl <- d.downField("baseUrl").as[Option[String]]
+      command <- d.downField("command").as[Option[List[String]]]
+      args <- d.downField("args").as[Option[List[String]]]
+      pdMountPath <- d.downField("pdMountPath").as[String]
```
rtitle: Wondering if we can should make `pdMountPath` and `pdAccessMode` optional in the yaml schema and default to `/data` and `ReadWriteOnce` in Leo

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpAppDAO.scala
```diff
@@ -3,26 +3,27 @@ package dao
...
 import org.broadinstitute.dsde.workbench.leonardo.dao.HostStatus.HostReady
 import org.broadinstitute.dsde.workbench.leonardo.dns.KubernetesDnsCache
 import org.broadinstitute.dsde.workbench.model.google.GoogleProject
 import org.http4s.client.Client
 import org.http4s.{Method, Request, Uri}
 
-class HttpGalaxyDAO[F[_]: Timer: ContextShift: Concurrent](val kubernetesDnsCache: KubernetesDnsCache[F],
-                                                           client: Client[F])
-    extends GalaxyDAO[F] {
+class HttpAppDAO[F[_]: Timer: ContextShift: Concurrent](val kubernetesDnsCache: KubernetesDnsCache[F],
```
rtitle, line 14: 👍   renaming this class

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/PersistentDiskComponent.scala
```diff
@@ -229,6 +229,8 @@ object persistentDiskQuery {
           appQuery.isDiskAttached(diskId)
         case Some(FormattedBy.GCE) =>
           RuntimeConfigQueries.isDiskAttached(diskId)
+        case Some(FormattedBy.Custom) =>
```
rtitle: Nit: in scala can do:
```
case Some(FormattedBy.Galaxy | FormattedBy.Custom) =>
  appQuery.isDiskAttached(diskId)
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HTTPAppDescriptorDAO.scala
jdcanas: Yeah... They probably can, but its very helpful as circe can sometimes eat logs if somewhere along the line decoding fails. There are tests to verify it, but it could be annoying for future developers when adding new fields that aren't necessarily in the hard-coded yaml string in the test, or if we move the files (which we currently reference directly via github link) and forget to update the code.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/KubernetesServiceInterp.scala
```diff
@@ -535,15 +536,20 @@ final class LeoKubernetesServiceInterp[F[_]: Parallel](
             NamespaceId(-1),
             namespaceName
           ),
-          disk,
+          Some(disk),
           req.appType match {
             case Galaxy =>
               galaxyConfig.services.map(config => KubernetesService(ServiceId(-1), config))
+            case Custom =>
+              // Back Leo will populate services when it parses the descriptor
```
rtitle, line 544: Interesting. I wonder if it would be good for front Leo to parse & validate the descriptor up front (e.g. before creating k8s resources). I can see pros and cons either way.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -848,6 +871,98 @@ class GKEInterpreter[F[_]: Parallel: ContextShift: Timer](
...
+                                     disk: PersistentDisk,
+                                     descriptorOpt: Option[String],
+                                     extraArgs: List[String],
+                                     customEnvironmentVariables: Map[String, String])(
+    implicit ev: Ask[F, AppContext]
+  ): F[Unit] =
+    for {
+      ctx <- ev.ask
+
+      _ <- logger.info(
```
rtitle: Same comment about doing `logger.info(ctx.loggingCtx)(msg)` going forward

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HTTPAppDescriptorDAO.scala
```diff
@@ -0,0 +1,78 @@
+package org.broadinstitute.dsde.workbench.leonardo
+package dao
+
+import cats.effect.Effect
+import cats.implicits._
+import cats.mtl.Ask
+import io.chrisdavenport.log4cats.StructuredLogger
+import io.circe.Decoder
+import io.circe.yaml.parser
```
rtitle: `io.circe.yaml` works pretty well :)

rtitle:
Mostly nits above, overall looks good! I agree we should follow up with the TODOs you listed in the PR description, but I think the scope of this PR looks good. 👍

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HTTPAppDescriptorDAO.scala
jdcanas: Makes sense, ill leave a TODO in there linking to the disk mode SPIKE in case we end up wanting to support multiple disk modes

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/KubernetesServiceInterp.scala
jdcanas: Yeah, it seems simpler to bundle all the logic in back-leo, since it needs to use various information about the parsed service in the helm commands, and otherwise we'd have to persist a lot of the information in the app.yaml. (though a good bit is persisted...)

There's definitely value in parsing it up front, in that the user would see right away if their descriptor is invalid. I'll do a bit of testing and make sure a good message is bubbled up to the user in the case of a non-existent/unparsable descriptor.

jdcanas: hmm.. this does bring up another concern that we'll have to address at some point: Versioning of these app.yamls. I expect that will be captured with the descriptor, but ill make sure to not it in the CI ticket I'll add

jdcanas: Here is what the errorMessage will be. Everything after `Original Message` is exception specific. Its a bit verbose... but I think it better to include the whole thing for accessibility as opposed to obfuscating. 

```"errorMessage": "Failed to process descriptor: https://rawow.githubusercontent.com/DataBiosphere/terra-app/main/apps/ucsc_genome_browser/app.yaml. Please ensure it is a valid descriptor, and that the remote file is valid yaml following the schema detailed here: https://github.com/DataBiosphere/terra-app#app-schema. \n\tOriginal message: Error occurred fetching app descriptor from path https://rawow.githubusercontent.com/DataBiosphere/terra-app/main/apps/ucsc_genome_browser/app.yaml: \n<html>\n<head>\n<title>Fastly error: unknown domain rawow.githubusercontent.com</title>\n</head>\n<body>\n<p>Fastly error: unknown domain: rawow.githubusercontent.com. Please check that this domain has been added to a service.</p>\n<p>Details: cache-pwk4966-PWK</p></body></html>",```

File: core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/diskModels.scala [hunk 12]
```diff
@@ -104,4 +104,8 @@ object FormattedBy extends Enum[FormattedBy] {
   final case object Galaxy extends FormattedBy {
     override def asString: String = "GALAXY"
   }
+
+  final case object Custom extends FormattedBy {
```
Qi77Qi, line 108: I wonder if it makes sense to have each supported app to be its own data type...cuz a disk formated by one app can't be re-used by another

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/appRoutesModels.scala
```diff
@@ -27,7 +27,9 @@ final case class CreateAppRequest(kubernetesRuntimeConfig: Option[KubernetesRunt
                                   appType: AppType,
                                   diskConfig: Option[PersistentDiskRequest],
                                   labels: LabelMap = Map.empty,
-                                  customEnvironmentVariables: Map[String, String])
+                                  customEnvironmentVariables: Map[String, String],
+                                  descriptorPath: Option[String],
```
Qi77Qi: if this path is a uri, can we use a `org.http4s.Uri` for its type

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/appRoutesModels.scala
```diff
@@ -27,7 +27,9 @@ final case class CreateAppRequest(kubernetesRuntimeConfig: Option[KubernetesRunt
...
+                                  extraArgs: List[String])
```
Qi77Qi, line 31: what does `extraArgs` will look like?

http/src/main/resources/reference.conf
```diff
@@ -251,6 +251,11 @@ gke {
     postgresDiskSizeGB = 10
     postgresDiskBlockSize = 4096
   }
+  customApp {
```
Qi77Qi, line 254: you can also make these into kebab case and use pureconfig instead...and then just need to add it [here](https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/ConfigReader.scala#L18)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/LeoProfile.scala
```diff
@@ -95,6 +95,17 @@ private[leonardo] object LeoProfile extends MySQLProfile {
             res.fold(e => throw e, identity)
           }
         )
+    implicit val mapListColumnType: BaseColumnType[List[String]] =
+      MappedColumnType.base[List[String], String](
+        _.asJson.printWith(Printer.noSpaces),
```
Qi77Qi, line 101: trying to understand why does this need to be in json format.
do you have an example of how this looks like when persisted to DB?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeServiceInterp.scala
```diff
@@ -896,6 +896,10 @@ object RuntimeServiceInterp {
         case Some(pd) =>
           for {
             isAttached <- pd.formattedBy match {
```
Qi77Qi, line 898: might be easier to read if we do
```
(pd.formattedBy, willByUsedBy) match {
???
}
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -364,27 +370,44 @@ class GKEInterpreter[F[_]: Parallel: ContextShift: Timer](
...
+      }
 
       _ <- logger.info(ctx.loggingCtx)(
         s"Finished app creation for app ${app.appName.value} in cluster ${gkeClusterId.toString}"
       )
 
       // TODO: this logic will be changed in the next PR
       isUsedByGalaxy <- persistentDiskQuery.isUsedByGalaxy(diskId).transaction
-      _ <- if (isUsedByGalaxy.exists(identity)) F.unit
+      _ <- if (isUsedByGalaxy.exists(identity) || app.appType != AppType.Galaxy) F.unit
```
Qi77Qi, line 411: should this be `app.appType == AppType.Galaxy`?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -848,6 +871,103 @@ class GKEInterpreter[F[_]: Parallel: ContextShift: Timer](
...
+        .installChart(
+          release,
+          config.customAppConfig.chartName,
+          config.customAppConfig.chartVersion,
+          org.broadinstitute.dsp.Values(chartValues)
+        )
+        .run(helmAuthContext)
+
+      // Poll app until it starts up
+      last <- streamFUntilDone(
```
Qi77Qi, line 966: can probably just use `streamUntilDoneOrTimeout`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GceInterpreter.scala
```diff
@@ -132,10 +132,10 @@ class GceInterpreter[F[_]: Parallel: ContextShift](
               )
             )
             isFormatted <- persistentDisk.formattedBy match {
-              case Some(FormattedBy.Galaxy) =>
+              case Some(FormattedBy.Galaxy) | Some(FormattedBy.Custom) =>
                 F.raiseError[Boolean](
                   new RuntimeException(
-                    "Trying to use a Galaxy formatted disk for creating GCE runtime. This should never happen."
+                    "Trying to use an app formatted disk for creating GCE runtime. This should never happen."
```
Qi77Qi: can we also include the actual app in the message

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/diskModels.scala
rtitle: Hm.. the point of this is to not have to pre-define apps in Leo. I wouldn't want to have to make a code change for each new app (unless infosec says there needs to be an allow-list of apps, of course… but this doesn't seem to be the case).

I don't think it's necessarily true that a disk can't be re-used across app types. A disk with data on it populated with Jupyter can be plugged into cirrocumulus for example. I think it's just Galaxy disks that we want to ensure only get used with Galaxy, because they do some special NFS formatting.

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/appRoutesModels.scala
rtitle: Yeah +1.. for now we are only supporting Uris, in the future might allow things like Dockstore references but that's TBD.

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/appRoutesModels.scala
rtitle: It's kind of like https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/

For example:
```
    command: ["printenv"]
    args: ["HOSTNAME", "KUBERNETES_PORT"]
```

Basically the app.yaml defines a `command` and `args` to startup the app. `extraArgs` exists in case the app accepts extra dynamic arguments that don't necessarily belong in the app descriptor (a path to a data file, for example) but the user might want to specify at app launch time.

 I did this originally mostly for `cirrocumulus` which requires a data path at app launch time.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/LeoProfile.scala
rtitle: I'm guessing it'll be stored in the DB like `["foo","bar","baz"]`

I think this probably makes sense for `APP.extraArgs`. But we shouldn't overuse this.. most 1..N DB relationships should use a join table.

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/diskModels.scala
jdcanas: Yeah, I didn't want to enumerate the apps here for this reason. Thats why I talk a bit about how formattedBy could be derived from a descriptor in the PR description, and mention some other issues with it. I plan to make a spike to plan out how exactly formattedBy will work in the custom-apps world, [link incoming]

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/KubernetesServiceInterp.scala
rtitle: Yeah I think it's slightly better for the API to return a `Bad Request` up front than transition the app from `Creating` -> `Error` (creating k8s resources along the way) with the above error message. OK with making a separate ticket for validation though.

Re versioning, the app descriptors are already stored in GitHub. And the ones in `terra-app` are really just meant to be examples, I imagine users storing app yamls in other places as well. So Leo isn't really managing the app descriptors yet so we probably don't need to worry about versioning. (We may once we have an app "library" but that's a ways down the road.)

rtitle (approved):

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/LeoProfile.scala
jdcanas: Yeah, it seemed limitting to model it another way
Screen Shot 2021-03-05 at 10 07 06 AM

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
rtitle: I kinda like `streamFUntilDone` here so we can check the `Option` below and throw `AppCreationException`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/KubernetesServiceInterp.scala
jdcanas: Yeah, I plan to make a ticket for validation. I know we can theoretically take any app descriptor and run with it, but there is the product angle that I think should be discussed in terms of 'approved' descriptors, and whether leo should enforce that

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
jdcanas: No, we want to not do any of the below logic for any apps besides galaxy (as its custom disk logic)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeServiceInterp.scala
jdcanas: Yeah, I really don't like this block.
It seems like it can be further simplified to something like 
```
              //case Some(x) => if x is willBeUsedBy, check if its attached
              //                  else already formatted exception
              // case None => isAttached
```
I left a TODO there with the above comments, but I was wary to touch this since there is a lot of disk-related work going on, and I plan on making a follow-up ticket that may completely change this logic anyways.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GceInterpreter.scala
jdcanas: I'll include the disk id. Should allow debugging with one extra step on devs part instead of a potential app lookup in the gceinterp, which seems wrong.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
Qi77Qi: nvm..I misread the condition the first time...

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/KubernetesServiceInterp.scala
jdcanas: Though, worth mentioning, we get basic URI validation for free in front leo by nature of the decoder for the `Uri` class based on Qi's feedback (see JsonCodec)

jdcanas:
jenkins retest

PR #1906, opened 2021-03-04
[IA-2507] Better handle Galaxy app resume time-out

rtitle:
https://broadworkbench.atlassian.net/browse/IA-2507

Made the following fixes:
1. Make `getApp` and `listApp` endpoints still return apps in `ERROR` status, even if the underlying cluster/nodepool is `DELETED`.
2. If starting an app fails in the backend, transition the status back to `STOPPING` -> `STOPPED` instead of immediately moving to `ERROR`. Persist an `APP_ERROR` record for traceability.
3. Label more GCP resources with the `leonardo` label, and improve some logging.
4. Update `galaxykubeman` chart to `0.7.3` to get https://github.com/galaxyproject/galaxykubeman-helm/pull/23
5. Made apps in `STOPPED` status non-deleteable.

Verified the above in unit tests and local Terra UI.

A couple UI improvements would be nice TODO:
1. Display the `APP_ERROR` message when an app is in `ERROR` status in Terra UI (it looks like this doesn't happen currently so the user can't see the error msg).
2. If an app start fails and the app moves back to `STOPPED`, display the error via toast notification.

I may try to do the UI stuff if it's easy; but I think this Leo PR can go first regardless. 

---

In all cases:


rtitle:
jenkins retest

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/appRoutesModels.scala
```diff
@@ -73,7 +67,7 @@ object ListAppResponse {
             n.autoscalingEnabled
           ),
           a.errors,
-          if (hasError) AppStatus.Error else a.status,
+          a.status,
```
Qi77Qi, line 70: how does the error status get aggregated now?

rtitle:
jenkins retest

http/src/test/resources/reference.conf
```diff
@@ -71,6 +71,13 @@ pubsub {
   non-leo-message-subscriber {
     subscription-name = "nonLeoMessageSubscription"
   }
+
+  kubernetes-monitor {
```
Qi77Qi, line 75: is this used?

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/appRoutesModels.scala
rtitle: I was thinking we should return the exact app status in the API response, not try to infer the app status from cluster/nodepool tables. I think any error during a user operation (create/start/stop/delete) moves the app to `ERROR` status.. so not sure we need to aggregate all the statuses together.

rtitle: (The immediate reason for this change is I'm persisting an `APP_ERROR` record if `startApp` fails, but moving the app back to `STOPPED` status. So I want the API response to say `STOPPED` not `ERROR`.)

http/src/test/resources/reference.conf
rtitle: Yeah I added a unit test to verify `startApp` timeout, so I wanted a test config with a shorter timeout (think the main one is like 10 mins).

Qi77Qi: oh I see....I thought this is new config you added, and didn't see it being used anywhere in this PR

Qi77Qi (approved):

rtitle:
jenkins retest.. looks like flake

PR #1908, opened 2021-03-04
[IA-2487] Use v2 createResource api 

jdcanas:
Tested in my fiab, SAM seems happy with the request and the app gets into running status. 

new api we are using: 

https://sam.dsde-prod.broadinstitute.org/#/Resources/createResourceV2

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/samModels.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/samModels.scala
@@ -4,17 +4,30 @@ import ca.mrvisser.sealerate
 import org.broadinstitute.dsde.workbench.model.WorkbenchEmail
 import org.broadinstitute.dsde.workbench.model.google.GoogleProject
 
-final case class RuntimeSamResourceId(resourceId: String) extends AnyVal
-final case class PersistentDiskSamResourceId(resourceId: String) extends AnyVal
-final case class ProjectSamResourceId(googleProject: GoogleProject) extends AnyVal
-final case class AppSamResourceId(resourceId: String) extends AnyVal
+sealed trait SamResourceId {
+  def resourceId: String
+  def asString: String = resourceId
+}
+
+object SamResourceId {
+  final case class RuntimeSamResourceId(resourceId: String) extends SamResourceId
+
+  final case class PersistentDiskSamResourceId(resourceId: String) extends SamResourceId
+
+  final case class ProjectSamResourceId(googleProject: GoogleProject) extends SamResourceId {
+    override def resourceId: String = googleProject.value
+  }
+
+  final case class AppSamResourceId(resourceId: String) extends SamResourceId
+
+}
 
 sealed trait SamResourceType extends Product with Serializable {
   def asString: String
 }
 object SamResourceType {
   final case object Project extends SamResourceType {
-    val asString = "billing-project"
+    val asString = "google-project"
   }
   final case object Runtime extends SamResourceType {
     val asString = "notebook-cluster"
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/SamAuthProvider.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/SamAuthProvider.scala
@@ -14,6 +14,7 @@ import fs2.Stream
 import io.chrisdavenport.log4cats.Logger
 import io.circe.{Decoder, Encoder}
 import org.broadinstitute.dsde.workbench.leonardo.JsonCodec._
+import org.broadinstitute.dsde.workbench.leonardo.SamResourceId._
 import org.broadinstitute.dsde.workbench.leonardo.dao.SamDAO
 import org.broadinstitute.dsde.workbench.leonardo.model._
 import org.broadinstitute.dsde.workbench.leonardo.util.CacheMetrics
@@ -181,10 +182,12 @@ class SamAuthProvider[F[_]: Effect: Logger: Timer: OpenTelemetryMetrics](samDao:
     creatorEmail: WorkbenchEmail,
     googleProject: GoogleProject
   )(implicit sr: SamResource[R], encoder: Encoder[R], ev: Ask[F, TraceId]): F[Unit] =
-    if (sr.policyNames == Set(SamPolicyName.Creator))
+    // TODO: consider using v2 for all existing entities if this works out for apps https://broadworkbench.atlassian.net/browse/IA-2569
+    // Apps are modeled different in SAM than other leo resources.
+    if (sr.resourceType != SamResourceType.App)
       samDao.createResource(samResource, creatorEmail, googleProject)
     else
-      samDao.createResourceWithManagerPolicy(samResource, creatorEmail, googleProject)
+      samDao.createResourceV2(samResource, creatorEmail, googleProject)
 
   override def notifyResourceDeleted[R](
     samResource: R,
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/SamAuthProvider.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/SamAuthProvider.scala
@@ -187,7 +187,7 @@ class SamAuthProvider[F[_]: Effect: Logger: Timer: OpenTelemetryMetrics](samDao:
     if (sr.resourceType != SamResourceType.App)
       samDao.createResource(samResource, creatorEmail, googleProject)
     else
-      samDao.createResourceV2(samResource, creatorEmail, googleProject)
+      samDao.createResourceWithParent(samResource, creatorEmail, googleProject)
 
   override def notifyResourceDeleted[R](
     samResource: R,
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpSamDAO.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpSamDAO.scala
@@ -166,7 +166,7 @@ class HttpSamDAO[F[_]: Effect](httpClient: Client[F], config: HttpSamDaoConfig, 
         }
     } yield ()
 
-  def createResourceV2[R](resource: R, creatorEmail: WorkbenchEmail, googleProject: GoogleProject)(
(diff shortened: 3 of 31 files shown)


core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/samModels.scala
```diff
@@ -4,17 +4,24 @@ import ca.mrvisser.sealerate
...
+}
+final case class AppSamResourceId(resourceId: String) extends SamResourceId
 
 sealed trait SamResourceType extends Product with Serializable {
   def asString: String
 }
 object SamResourceType {
   final case object Project extends SamResourceType {
-    val asString = "billing-project"
+    val asString = "google-project" //TODO: verify, is this correct?
```
jdcanas: This is what the ticket says. I dug through where this gets plugged into the request to sam, and verified against a valid SAM payload. Just wanted additional verification on this

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/samModels.scala
```diff
@@ -4,17 +4,24 @@ import ca.mrvisser.sealerate
...
-final case class RuntimeSamResourceId(resourceId: String) extends AnyVal
-final case class PersistentDiskSamResourceId(resourceId: String) extends AnyVal
-final case class ProjectSamResourceId(googleProject: GoogleProject) extends AnyVal
-final case class AppSamResourceId(resourceId: String) extends AnyVal
+sealed trait SamResourceId {
+  def resourceId: String
+  def asString: String = resourceId
+}
+
+final case class RuntimeSamResourceId(resourceId: String) extends SamResourceId
```
rtitle: Normally for ADTs we put the classes in the trait's companion object:
```
sealed SamResourceId { 
  ..
}
object SamResourceId {
  final case class RuntimeSamResourceId(…) extends SamResourceId
  ..
}
```

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/samModels.scala
rtitle: Yup, this is correct.. you can remove the TODO

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/SamAuthProvider.scala
```diff
@@ -181,10 +181,12 @@ class SamAuthProvider[F[_]: Effect: Logger: Timer: OpenTelemetryMetrics](samDao:
     creatorEmail: WorkbenchEmail,
     googleProject: GoogleProject
   )(implicit sr: SamResource[R], encoder: Encoder[R], ev: Ask[F, TraceId]): F[Unit] =
-    if (sr.policyNames == Set(SamPolicyName.Creator))
+    // At the time this comment was written, the else is only executed for apps
```
rtitle: Any way this comment could be made clearer for someone looking at this code in 3 months with no context? :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpSamDAO.scala
```diff
@@ -183,22 +183,21 @@ class HttpSamDAO[F[_]: Effect](httpClient: Client[F], config: HttpSamDaoConfig,
       )
       authHeader = Authorization(Credentials.Token(AuthScheme.Bearer, token))
       _ <- logger.info(
-        s"${traceId} | creating ${sr.resourceType.asString} resource in sam for ${googleProject}/${sr.resourceIdAsString(resource)}"
+        s"${traceId} | creating ${sr.resourceType.asString} resource in sam v2 for ${googleProject}/${sr.resourceIdAsString(resource)}"
       )
       _ <- metrics.incrementCounter(s"sam/createResource/${sr.resourceType.asString}")
-      projectOwnerEmail <- getProjectOwnerPolicyEmail(authHeader, googleProject)
```
rtitle, line 189: Can you remove the `getProjectOwnerPolicyEmail` method if we're no longer using it?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/SamAuthProvider.scala
rtitle: Also: technically we could start using the [createResourceWithDefaultsV2](https://sam.dsde-prod.broadinstitute.org/#/Resources/createResourceWithDefaultsV2) for runtimes and disks today, which is still the `v2` API but without the inherited policies from `google-project`. Maybe as part of IA-2569 we should decide whether to just use `createResourceWithDefaultsV2` or do some larger refactoring to model runtimes/disks similar to apps.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpSamDAO.scala
```diff
@@ -455,6 +465,7 @@ final case class HttpSamDaoConfig(samUri: Uri,
                                   serviceAccountProviderConfig: ServiceAccountProviderConfig)
 
 final case class UserEmailAndProject(userEmail: WorkbenchEmail, googleProject: GoogleProject)
+final case class SerializableSamResource(resourceTypeName: SamResourceType, resourceId: SamResourceId)
```
rtitle, line 459: I see why this is needed in the current model… but I'm wondering if the `SamResource` typeclass could be used for specifying the parent as well. Not sure.. may play around with code.

rtitle: Nevermind, not sure `parent` fits very well in `SamResource` either. I think this is fine. One alternative would be to assume that projects are the only parent type and put this in `CreateSamResourceRequest`: 
```
parent: ProjectSamResourceId,
```
That way we wouldn't need `SerializableSamResource `. But I don't feel strongly.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpSamDAO.scala
```diff
@@ -183,22 +183,21 @@ class HttpSamDAO[F[_]: Effect](httpClient: Client[F], config: HttpSamDaoConfig,
...
       authHeader = Authorization(Credentials.Token(AuthScheme.Bearer, token))
       _ <- logger.info(
-        s"${traceId} | creating ${sr.resourceType.asString} resource in sam for ${googleProject}/${sr.resourceIdAsString(resource)}"
+        s"${traceId} | creating ${sr.resourceType.asString} resource in sam v2 for ${googleProject}/${sr.resourceIdAsString(resource)}"
       )
       _ <- metrics.incrementCounter(s"sam/createResource/${sr.resourceType.asString}")
-      projectOwnerEmail <- getProjectOwnerPolicyEmail(authHeader, googleProject)
-      policies = Map(
-        SamPolicyName.Creator -> SamPolicyData(List(creatorEmail), List(SamRole.Creator)),
-        SamPolicyName.Manager -> SamPolicyData(List(projectOwnerEmail.email), List(SamRole.Manager))
```
rtitle, line 192: I thought above removing `SamPolicyName.Manager` since we're not using that either.. but it may be good to leave it in due to existing Sam app resources in the DB.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/SamAuthProvider.scala
jdcanas: I updated the comment, and added notes to the ticket

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpSamDAO.scala
jdcanas: It is in `CreateSamResourceRequest`, but technically `parent` field is not restricted to project resources according to SAM swagger. We could limit it from leonardo's perspective, but I think this makes it clearer what is actually being sent in the payload.

jdcanas:
its broken atm, need to refactor some references. Tomorrow Justin's problem

rtitle (approved):
This looks good, it's good automation tests pass. 

Because we've had issues (and post-mortems) with Leo/Sam changes in the past, some extra diligent testing would be good here I think.

1. Does all resource creation (runtime, disk, app) work from Terra UI?
2. Can project owners still list (runtimes, disks, apps) made by other members of the project? (note: AoU relies on this functionality)
3. Ensure perf tests look good after this change
4. Ensure AoU doesn't break in dev after this change

OK if some of this verification happens post-merge, and also OK if this merges next week too. Just want to be extra careful around Sam integration changes.

jdcanas:
jenkins retest

jdcanas:
jenkins retest - not even hitting our tests at failure point. Still needs a rebase so they will be run again anyways

jdcanas:
jenkins retest

jdcanas:
Updated with v2, have tested the above scenarios, they passed!

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/SamAuthProvider.scala
```diff
@@ -181,10 +182,12 @@ class SamAuthProvider[F[_]: Effect: Logger: Timer: OpenTelemetryMetrics](samDao:
     creatorEmail: WorkbenchEmail,
     googleProject: GoogleProject
   )(implicit sr: SamResource[R], encoder: Encoder[R], ev: Ask[F, TraceId]): F[Unit] =
-    if (sr.policyNames == Set(SamPolicyName.Creator))
+    // TODO: consider using v2 for all existing entities if this works out for apps https://broadworkbench.atlassian.net/browse/IA-2569
```
rtitle, line 185: I feel like `createResource` vs `createResourceV2` is confusing because we're using Sam v2 APIs in both cases.

How about `createResource` and `createResourceWithParent`?

You can keep the if statement, just want to make naming more clear.

rtitle:
Latest changes look good - just a comment above on method naming.

rtitle (approved):

jdcanas:
jenkins retest

jdcanas:
jenkins retest

PR #1913, opened 2021-03-08
[no ticket] Removing predeleting status update for nodepools

xuscrbw:
Found while doing some Perf test debugging. We shouldn't have a pre-deleting status for nodepools since only cronjobs should be deleting it
---

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesServiceDbQueries.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesServiceDbQueries.scala
@@ -187,7 +187,6 @@ object KubernetesServiceDbQueries {
 
   def markPreDeleting(nodepoolId: NodepoolLeoId, appId: AppId)(implicit ec: ExecutionContext): DBIO[Unit] =
     for {
-      _ <- nodepoolQuery.updateStatus(nodepoolId, NodepoolStatus.Predeleting)
       _ <- appQuery.updateStatus(appId, AppStatus.Predeleting)
     } yield ()
 
--- a/http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/KubernetesServiceInterpSpec.scala
+++ b/http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/KubernetesServiceInterpSpec.scala
@@ -250,6 +250,7 @@ final class KubernetesServiceInterpSpec extends AnyFlatSpec with LeonardoTestSui
 
     //we can't delete while its creating, so set it to Running
     dbFutureValue(appQuery.updateStatus(appResultPreStatusUpdate.get.app.id, AppStatus.Running))
+    dbFutureValue(nodepoolQuery.updateStatus(appResultPreStatusUpdate.get.nodepool.id, NodepoolStatus.Running))
 
     val appResultPreDelete = dbFutureValue {
       KubernetesServiceDbQueries.getActiveFullAppByName(project, appName)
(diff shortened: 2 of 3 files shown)


Qi77Qi (approved):
thanks @xuscrbw !

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesServiceDbQueries.scala
```diff
@@ -187,7 +187,6 @@ object KubernetesServiceDbQueries {
 
   def markPreDeleting(nodepoolId: NodepoolLeoId, appId: AppId)(implicit ec: ExecutionContext): DBIO[Unit] =
     for {
-      _ <- nodepoolQuery.updateStatus(nodepoolId, NodepoolStatus.Predeleting)
```
rtitle, line 190: Can we remove the `NodepoolStatus.Predeleting` enum altogether?

Would also be good to update the appropriate tests in `KubernetesServiceInterpSpec` to make sure the nodepool status is still `RUNNING` after an app is deleted.

xuscrbw: I can do this also in this PR

xuscrbw:
jenkins retest

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesServiceDbQueries.scala
xuscrbw: @rtitle [info] - should delete an app and update status appropriately *** FAILED ***
[info]   PRECREATING did not equal RUNNING (KubernetesServiceInterpSpec.scala:268)

Hm, getting this failure on the unit tests (changed the line to Running). This seems really strange to me since this implies that the status is being changed solely with the deleteApp call. I don't think we've had issues with this test recently, maybe there's another bug here.

rtitle: I think this is expected because the test does not include a back leo/pubsub subscriber. The test does:

1. `kubeServiceInterp.createApp`
(app and nodepool are `precreating` at this point)
2. manually update app to `running` 
(this simulates the status transition done by back leo)
3. `kubeServiceInterp.deleteApp` 
(app should now be `predeleting` and nodepool status should not change)

Maybe you can modify step (2) to update _both_ app and nodepool status to Running for the sake of simulating back Leo, and verify that the nodepool status is still `RUNNING` after step 3.

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/KubernetesServiceInterpSpec.scala
```diff
@@ -265,7 +265,7 @@ final class KubernetesServiceInterpSpec extends AnyFlatSpec with LeonardoTestSui
 
     clusterPostDelete.length shouldEqual 1
     val nodepool = clusterPostDelete.head.nodepools.head
-    nodepool.status shouldEqual NodepoolStatus.Predeleting
+    nodepool.status shouldEqual NodepoolStatus.Running
```
rtitle, line 269: As mentioned above, if you do 
```
dbFutureValue(nodepoolQuery.updateStatus(theNodepoolId, NodepoolStatus.Running))
```
near L252 then this check should succeed.

rtitle (approved):

PR #1916, opened 2021-03-10
QA-1360: make flakey test re-use same notebook across stop/start instead of create new one

rtitle:
https://broadworkbench.atlassian.net/browse/QA-1360

This test frequently fails [here](https://github.com/DataBiosphere/leonardo/blob/develop/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookGCECustomizationSpec.scala#L193) when it tries to create a new notebook after a runtime stop/start due to unexpected selenium alerts.

I noticed [this test](https://github.com/DataBiosphere/leonardo/blob/develop/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookGCEClusterMonitoringSpec.scala#L54) (which rarely fails) also stop/starts a runtime but opens an existing notebook.

Trying a similar approach in this test where I re-use the same notebook instead of creating a new one after stop/start.

---

In all cases:


xuscrbw (approved):
LGTM. Hopefully this solves it!

PR #1919, opened 2021-03-10
[IA-2572] Automation test for custom apps

jdcanas:


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -950,10 +950,19 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift: Parallel](
                                             gkeInterp.getGalaxyPostgresDiskName(dbApp.app.appResources.namespace.name))
         diskDetachResult <- streamUntilDoneOrTimeout(
           getDiskDetachStatus(originalDetachTimestampOpt, getDisk),
-          24,
+          30,
           5 seconds,
           "The disk failed to detach within the time limit, cannot proceed with delete disk"
-        )
+        ).adaptError {
+          case e =>
```
jdcanas, line 957: If this failed, it threw an uncaught exception that caused the db state for the app not being properly updated to error status.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -1073,9 +1082,11 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift: Parallel](
             nodepoolQuery.updateStatus(nodepoolId, NodepoolStatus.Error).transaction
           )
         } yield ()
-      case _ =>
-        F.raiseError(
-          new RuntimeException(s"handleKubernetesError should not be used with a non kubernetes error. Error: ${e}")
+      case e =>
+        ev.ask.flatMap(ctx =>
```
jdcanas, line 1086: I don't think we should be throwing an exception within the top-level error handling function... there's not much we can do, as this essentially indicates we failed error handling at some point and an unintended exception bubbled up. 

Currently, this causes leonardo to crash if the block is hit.

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
```diff
@@ -243,19 +243,19 @@ final case class DefaultKubernetesLabels(googleProject: GoogleProject,
 
 sealed abstract class ErrorAction
 object ErrorAction {
-  case object CreateGalaxyApp extends ErrorAction {
+  case object CreateApp extends ErrorAction {
```
Qi77Qi, line 246: think we need to update `toString` as well

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
Qi77Qi: 👍  if this handler is only used at top level

rtitle: 👍 , this is better than crashing the Leo app I think.

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
rtitle: +1

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/apps/CustomAppCreationSpec.scala
```diff
@@ -0,0 +1,95 @@
...
+class CustomAppCreationSpec
+    extends GPAllocFixtureSpec
+    with LeonardoTestUtils
+    with GPAllocUtils
+    with ParallelTestExecution {
+  implicit val auth: Authorization =
+    Authorization(Credentials.Token(AuthScheme.Bearer, ronCreds.makeAuthToken().value))
+
+  "create and delete a custom app" in { _ =>
+    withNewProject { googleProject =>
```
rtitle, line 25: Currently all the app tests run in separate projects.. wondering if (in the future, not necessarily in this PR) we could re-use the same project + kubernetes cluster for >1 test case. This would speed up the tests, and we don't necessarily need to test k8s cluster creation for each app test. This may be more possible than before because we no longer delete the cluster+nodepool when we delete an app.

rtitle (approved):
LGTM if it passes!

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
jdcanas: I didn't want to update the toString because this field is persisted to the database and decoded. It's a fairly high-risk change (though not necessarily complex), so I split it into this ticket https://broadworkbench.atlassian.net/browse/IA-2580

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/apps/CustomAppCreationSpec.scala
jdcanas: Hmm.... the reason we originally did this pre-dated the nodepool lock, since concurrent nodepool operations are not allowed. At the time, each app spun up its own nodepool. Since we use the same user for all tests, we may get re-use of these for free now.  I'll make a ticket

jdcanas:
jenkins retest. It passed the first time, trying to flush out any instability. I have ran this test 5 times locally and all 5 runs passed, just want to test it with the whole suite.

PR #1920, opened 2021-03-10
Add zone to createRuntime requests

wnojopra:
What I expect this PR does:
- Add an optional zone parameter to the createRuntime API
- Updates Swagger docs
- Adds a migration to add zone to the RUNTIME_CONFIG table
- Updates appropriate slick mapping
- I _think_ the zone is already passed to GCE. 

What I still need help with:
- Testing this locally. I think I have the right approach, but I still don't know how to check that what I'm doing will actually work?
- Testing the liquibase migration. How can I test this?
- I'm seeing unit test failures that all look like `java.sql.SQLTransientConnectionException: mysql.db - Connection is not available, request timed out after 5000ms.` I'm not sure if this is expected to not work, or if there is something I need to fix? I ran the docker commands for mysql that is mentioned in the README.

In general I am asking for feedback on if this is the right approach so far, and how I should proceed to make sure this change works. Thanks!
---

In all cases:

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/runtimeModels.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/runtimeModels.scala
@@ -197,7 +197,7 @@ object RuntimeConfig {
     bootDiskSize: Option[
       DiskSize
     ], //This is optional for supporting old runtimes which only have 1 disk. All new runtime will have a boot disk
-    zone: Option[ZoneName]
+    zone: ZoneName
   ) extends RuntimeConfig {
     val cloudService: CloudService = CloudService.GCE
   }
@@ -206,7 +206,7 @@ object RuntimeConfig {
   final case class GceWithPdConfig(machineType: MachineTypeName,
                                    persistentDiskId: Option[DiskId],
                                    bootDiskSize: DiskSize,
-                                   zone: Option[ZoneName])
+                                   zone: ZoneName)
       extends RuntimeConfig {
     val cloudService: CloudService = CloudService.GCE
   }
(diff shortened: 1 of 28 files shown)


core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/runtimeModels.scala
```diff
@@ -196,15 +196,17 @@ object RuntimeConfig {
     diskSize: DiskSize,
     bootDiskSize: Option[
       DiskSize
-    ] //This is optional for supporting old runtimes which only have 1 disk. All new runtime will have a boot disk
+    ], //This is optional for supporting old runtimes which only have 1 disk. All new runtime will have a boot disk
+    zone: Option[ZoneName]
```
Qi77Qi: I would prefer to make this non optional. For exisiting runtimes, we can do a database migration to populate it with the current default zone value

Qi77Qi:
you'll also need to update zone reference in [this file](https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GceInterpreter.scala) to the actual zone instead of just zone from configuration

wnojopra:
Okay! I finally got all the unit tests passing!

The Jenkins Automation Tests aren't passing but the [details](https://fc-jenkins.dsp-techops.broadinstitute.org/job/leonardo-PR-fiab-test/6391/) page isn't loading for me. Is there any way I can see what is failing / run those tests locally?

rtitle:
That’s awesome! I’ve been a bit busy this week but can take a look. I think Jenkins Automation Tests requires Broad VPN to look at — we can probably help with that.

wnojopra:
Thanks @rtitle! Please let me know what I can do to help move forward with this.

wnojopra:
Need to change the branch name, so moving to #1933

PR #1922, opened 2021-03-11
[IA-2587] Have automation tests re-use the same cluster for apps

rtitle:
https://broadworkbench.atlassian.net/browse/IA-2587

Make test cases in `AppSpec` use the same gpalloc project (and hence the same cluster). This should be possible now that we have the nodepool lock and preserve the nodepool after app deletion. This may make the tests more efficient.

I kept `BatchNodepoolCreationSpec` in its own project, because I think that test does explicitly want to test cluster creation. 

---

In all cases:

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoApiClient.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoApiClient.scala
@@ -229,7 +229,7 @@ object LeonardoApiClient {
     } yield res
   }
 
-  def waitUntilAppRunning(googleProject: GoogleProject, appName: AppName, shouldError: Boolean = true)(
+  def waitUntilAppRunning(googleProject: GoogleProject, appName: AppName)(
     implicit timer: Timer[IO],
     client: Client[IO],
     authHeader: Authorization,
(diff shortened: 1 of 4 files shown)


rtitle:
will retest this PR a few times as well

rtitle:
This actually fails because of:
```
Failed to create app gpalloc-qa-master-kitv38a/automation-test-app-a9vdpmkcz, status: 409 Conflict body: {"source":"leonardo","message":"You cannot create an app while a cluster is in Set(PRECREATING, PROVISIONING). Status: PRECREATING","statusCode":409,"exceptionClass":"class org.broadinstitute.dsde.workbench.leonardo.db.KubernetesAppCreationException","traceId":null}
```

I think the nodepool lock is not held during cluster creation -- we just reject the request up front. I think this is probably correct Leo behavior, thinking about a test fix…

jdcanas:
+1 to keeping `withNewProject` around. Seems useful even if its only in one place atm

jdcanas:
> This actually fails because of:
> 
> ```
> Failed to create app gpalloc-qa-master-kitv38a/automation-test-app-a9vdpmkcz, status: 409 Conflict body: {"source":"leonardo","message":"You cannot create an app while a cluster is in Set(PRECREATING, PROVISIONING). Status: PRECREATING","statusCode":409,"exceptionClass":"class org.broadinstitute.dsde.workbench.leonardo.db.KubernetesAppCreationException","traceId":null}
> ```
> 
> I think the nodepool lock is not held during cluster creation -- we just reject the request up front. I think this is probably correct Leo behavior, thinking about a test fix…

A fix may be fairly hard because we offer no way to check the status of a cluster/nodepool external to leo. You could try the batchcreation endpoint and wait to explicitly create the cluster with no apps, and then perhaps pull in the gke lib and check yourself...

The solution could also be some sort of project status endpoint that allows querying of the internal state. I feel like this has come up over and over, and would make testing a lot easier (in perf too). Security issues abound though...

rtitle:
Yeah.. trying now a test approach where we retry app creation 409s with this error msg. Not sure that's ideal though.

Other options:
- poll GKE directly until cluster is available (this is still hard because the Leo API doesn't expose the cluster name so the test would need to figure it out somehow)
- use batch creation endpoint (I'd rather not rely on that in tests though)
- expose cluster status via Leo API (or maybe separate app status code? 🤔 )
- make cluster creation obtain the nodepool lock. I think there is some reason I didn't want to do that though..

jdcanas:
Hmm.. I actually kind of like the idea of using the batch create endpoint fixture-level, and then have apps poll based on whatever status shows up while its not ready

I do agree its non-ideal to rely on a non real-world usage path for test code though,.,

rtitle:
Yeah my main objection to that is it's not real world usage

rtitle:
Summarized findings & ideas in https://broadworkbench.atlassian.net/browse/IA-2587, I'm going to put this aside for the ticket-taker, and not merge a risky test change before a long weekend. 🏖️

rtitle:
Re-opened this -- I just want to see if it works.

rtitle:
jenkins retest

rtitle:

rtitle:
mysql failed, jenkins retest

rtitle:
Hmm
```
  org.broadinstitute.dsde.workbench.StreamTimeoutError: app gpalloc-qa-master-6sx9ydn/restore-automation-test-app-axyulkunz did not finish app creation after 20 minutes.
      at org.broadinstitute.dsde.workbench.google2.package$.$anonfun$streamUntilDoneOrTimeout$1(package.scala:171)
```

jenkins retest

rtitle:
1 pass

jenkins retest

Dockerfile
```diff
@@ -42,6 +42,8 @@ RUN helm repo add ingress-nginx https://kubernetes.github.io/ingress-nginx && \
 # Leonardo will install the chart from local version.
 RUN pushd /leonardo && \
     helm pull terra-app-setup-charts/terra-app-setup --untar && \
+    helm pull galaxy/galaxykubeman --untar && \
```
rtitle: Seeing a lot of this in tests:
```
 failed to download \"galaxy/galaxykubeman\" (hint: running `helm repo update` may help)
```

Thinking it might be a good idea to cache the charts we use on disk instead of downloading them each time.

rtitle:
2 pass

(RShinySpec failed but app tests passed)

jenkins retest

rtitle:
OK an app test failed: https://fc-jenkins.dsp-techops.broadinstitute.org/job/leonardo-fiab-test-runner/34316/
```
app gpalloc-qa-master-tmprwl0/restore-automation-test-app-aci5eqy8z did not finish app creation after 20 minutes.
```

Need to debug.

rtitle:
1 pass 

jenkins retest

rtitle:
Going to go for 10 passes again, jenkins retest

rtitle:
Have rebased some fixes, going to start the counter again up to 10.

1 pass

jenkins retest

rtitle:
Failure: https://fc-jenkins.dsp-techops.broadinstitute.org/job/leonardo-fiab-test-runner/34906/
```
org.broadinstitute.dsde.workbench.leonardo.apps.AppCreationSpec.create, delete an app and re-create an app with same disk

app gpalloc-qa-master-eqwe8wu/restore-automation-test-app-am2vfghtz did not finish app creation after 20 minutes.
```

There must be some issue with this change, not sure what though.

rtitle:
multi-test

rtitle:
multi-test

PR #1925, opened 2021-03-12
[QA-1411] Bugfixes for custom apps

rtitle:
https://broadworkbench.atlassian.net/browse/QA-1411

This can wait till next week, but I wanted to make this PR while fresh.

I noticed `CustomAppCreationSpec` occasionally failing in Jenkins and made the following fixes:

1. increase admission webhook timeout to address this failure at custom app helm install time:
   ```
   custom app failed with: err is %!(EXTRA *errors.StatusError=Internal error occurred: failed calling webhook "validate.nginx.ingress.kubernetes.io": Post https://nginx-ingress-nginx-controller-admission.nginx.svc:443/networking/v1beta1/ingresses?timeout=10s: no endpoints available for service "nginx-ingress-nginx-controller-admission", string=)
   ```
   Googling suggests the 10s timeout may be too small under load. This webhook can also be disabled, but seeing if this resolves it first.

2. Fix bugs in `deleteApp` flow for custom apps. It was helm uninstalling the chart correctly, but still trying to do Galaxy-specific things like delete the postgres disk (which doesn't exist for custom apps). It was also logging "Deleting Galaxy chart" which was confusing. WIP, there may be more to fix here.

 
---

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
@@ -922,34 +922,42 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift: Parallel](
 
       deleteDisksInParallel = List(deleteDisk, deletePostgresDisk).parSequence_
 
-      // For Galaxy apps, wait for the postgres disk to detach before deleting the disks
-      // TODO should we do this for the other disk too?
-      deleteAppAndWaitForDiskDetach = if (dbApp.app.appType == AppType.Galaxy) {
-        val getGalaxyPostgresDisk = googleDiskService.getDisk(
-          msg.project,
-          zone,
-          gkeInterp.getGalaxyPostgresDiskName(dbApp.app.appResources.namespace.name)
-        )
-        for {
-          // we record the last disk detach timestamp here, before it is removed from galaxy
-          originalDiskOpt <- getGalaxyPostgresDisk
-          originalDetachTimestampOpt = originalDiskOpt.map(_.getLastDetachTimestamp)
+      task = for {
+        // For Galaxy apps, wait for the postgres disk to detach before deleting the disks
+        // TODO should we do this for the other disk too?
+        originalDetachTimestampOpt <- if (dbApp.app.appType == AppType.Galaxy) {
+          googleDiskService
+            .getDisk(
+              msg.project,
+              zone,
+              gkeInterp.getGalaxyPostgresDiskName(dbApp.app.appResources.namespace.name)
+            )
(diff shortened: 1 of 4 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -906,33 +906,36 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift: Parallel](
...
       task = for {
-        //we record the last disk detach timestamp here, before it is removed from galaxy
-        originalDiskOpt <- googleDiskService.getDisk(
-          msg.project,
-          zone,
-          gkeInterp.getGalaxyPostgresDiskName(dbApp.app.appResources.namespace.name)
-        )
-        originalDetachTimestampOpt = originalDiskOpt.map(_.getLastDetachTimestamp)
+        // For Galaxy apps, wait for the postgres disk to detach before deleting the disks
+        // TODO should we do this for the other disk too?
```
kyuksel, line 927: I'd think we'd want to do this for the other disk too. Am I missing anything?

kyuksel (approved):
Looks great to me 👍🏼  Thanks for annotating your PR so well. It helps a lot with reviewing 😊 
One inline comment about a `TODO` question.

Qi77Qi (approved):

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
rtitle: The old logic only checked the postgres disk, so I kept that consistent. But yes it feels like we'd need to wait for all disks to detach.

@Qi77Qi's PR has more changes to disk logic - maybe that could be done there if needed?

rtitle:
Gary's creating a ticket for this - once created I'll tag this PR with it and merge.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
Qi77Qi: yea..less change in this PR will be desirable :) it's going to cause merge conflicts as is and these can be tricky to resolve

For the original question, if we've seen instability around this, I'm all for adding the detaching check; otherwise, I'm thinking it might be ok to leave as is 🤷‍♀️

Qi77Qi: just fyi, I did see one occurrence of disk deletion failure in my local test, so I'm going to add the wait for data disk as well

PR #1931, opened 2021-03-17
[no ticket] disable new tests for now

Qi77Qi:
Will look at failures from last night..but disable some tests for now

---

In all cases:

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/apps/AppCreationSpec.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/apps/AppCreationSpec.scala
@@ -201,8 +201,8 @@ class AppCreationSpec extends GPAllocFixtureSpec with LeonardoTestUtils with GPA
               s"AppCreationSpec: app ${googleProject.value}/${appName.value} delete result: $monitorDeleteResult"
             )
           }
-//          // verify disk is also deleted
-//          getDiskResp <- LeonardoApiClient.getDisk(googleProject, diskName).attempt
+          // verify disk is also deleted
+          getDiskResp <- LeonardoApiClient.getDisk(googleProject, diskName).attempt
         } yield {
           getDiskResp.toOption shouldBe (None)
         }
--- a/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/apps/AppCreationSpec.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/apps/AppCreationSpec.scala
@@ -88,8 +88,8 @@ class AppCreationSpec extends GPAllocFixtureSpec with LeonardoTestUtils with GPA
               s"AppCreationSpec: app ${googleProject.value}/${appName.value} delete result: $monitorDeleteResult"
             )
           }
-          // Verify creating another app with the same disk doesn't error out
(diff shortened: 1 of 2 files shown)


automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/apps/AppCreationSpec.scala
```diff
@@ -201,8 +201,8 @@ class AppCreationSpec extends GPAllocFixtureSpec with LeonardoTestUtils with GPA
               s"AppCreationSpec: app ${googleProject.value}/${appName.value} delete result: $monitorDeleteResult"
             )
           }
-          // verify disk is also deleted
-          getDiskResp <- LeonardoApiClient.getDisk(googleProject, diskName).attempt
+//          // verify disk is also deleted
+//          getDiskResp <- LeonardoApiClient.getDisk(googleProject, diskName).attempt
```
rtitle: I think this failure may be because sometimes the disk response from Google is still `Deleting` (Leo doesn't wait for the disk to be deleted)

rtitle:
```
Some(GetPersistentDiskResponse(DiskId(3),gpalloc-qa-master-pelepag,ZoneName(us-central1-a),DiskName(smaptanbrs),Some(GoogleId(4311630024260997798)),b559pet-109021626430559047321@gpalloc-qa-master-pelepag.iam.gserviceaccount.com,PersistentDiskSamResourceId(test),Deleting,AuditInfo(ron.weasley@quality.firecloud.org,2021-03-17T11:47:14.178Z,None,2021-03-17T12:08:40.275Z),DiskSize(500),Standard,BlockSize(4096),Map(diskName -> smaptanbrs, creator -> ron.weasley@quality.firecloud.org, googleProject -> gpalloc-qa-master-pelepag, serviceAccount -> b559pet-109021626430559047321@gpalloc-qa-master-pelepag.iam.gserviceaccount.com))) was not equal to None
```

rtitle:
I also see this error:
```
org.broadinstitute.dsde.workbench.leonardo.apps.CustomAppCreationSpec.create and delete a custom app

Error Message
List(ERROR) was not equal to List(DELETED)
```

Thinking the bug may be that we only wait for disk detach if the appType == Galaxy. Feels like we should do this line for custom apps as well: https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala#L977

rtitle:
This bugfix looks good I think, let's see how tests do.

The "wait for detach" logic is a bit confusing, wondering if there's a way to make it cleaner.

```
if (shouldDeleteDisk) {
  for {
    _ <- F.bracket(getOriginalTimestamp)(deleteApp)(waitForDiskDetach)
    _ <- deleteDisk()
  } yield ()
} else {
  deleteApp
}
```

Dunno, just thinking out loud..

rtitle (approved):

Qi77Qi:
@rtitle  hmmm...not sure...the entire detach/delete block is inside `msg.diskId.traverse`, which is basically the `shouldDeleteDisk` flag check

PR #1932, opened 2021-03-17
[no ticket] Re-enable restore test

Qi77Qi:

---

In all cases:


core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
```diff
@@ -270,7 +270,9 @@ final case class AppError(errorMessage: String,
                           action: ErrorAction,
                           source: ErrorSource,
                           googleErrorCode: Option[Int],
-                          traceId: Option[TraceId] = None)
+                          traceId: Option[TraceId] = None) {
+  val messageToSaveInDb = s"${traceId.map(x => x.asString).getOrElse(" ")} | ${errorMessage}"
```
Qi77Qi: persist `traceId` as well if it exists

rtitle (approved):
+1 enabling tests, 👍 if they pass reliably

Qi77Qi:
werid git, jenkins issue

```
ERROR: Error cloning remote repo 'origin'
hudson.plugins.git.GitException: Command "git fetch --tags --progress git@github.com:DataBiosphere/leonardo.git +refs/heads/*:refs/remotes/origin/*" returned status code 128:
stdout: 
stderr: ssh: connect to host github.com port 22: Connection timed out
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

jenkins retest

Qi77Qi:
flakey ```12:49:10 [error] 	org.broadinstitute.dsde.workbench.leonardo.dao.HttpDockerDAOSpec```...this doesn't happen very often

jenkins retest

Qi77Qi:
random mysql container error..jenkins retest

rtitle:
Think there are some GCP issues today

rtitle:
@Qi77Qi mind if I add a commit? I think I found a couple things to fix in test code which may be causing flakiness.

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/apps/AppCreationSpec.scala
```diff
@@ -79,17 +79,20 @@ class AppCreationSpec extends GPAllocFixtureSpec with LeonardoTestUtils with GPA
...
+            // Verify creating another app with the same disk doesn't error out
+            for {
+              _ <- loggerIO.info(
+                s"AppCreationSpec: app ${googleProject.value}/${appName.value} delete result: $monitorDeleteResult"
+              )
+              _ <- LeonardoApiClient.createAppWithWait(googleProject, restoreAppName, createAppRequest)
+            } yield ()
           }
-//          // Verify creating another app with the same disk doesn't error out
-//          _ <- LeonardoApiClient.createAppWithWait(googleProject, restoreAppName, createAppRequest)
```
rtitle, line 92: I think may be why we were seeing "namespace already exists".. if the app deletion fails due to the flakey Galaxy pre-delete job, then this test would _still_ go on to create the second app while the namespace still exists in k8s.

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/apps/AppCreationSpec.scala
```diff
@@ -192,20 +195,26 @@ class AppCreationSpec extends GPAllocFixtureSpec with LeonardoTestUtils with GPA
...
           } else {
-            loggerIO.info(
-              s"AppCreationSpec: app ${googleProject.value}/${appName.value} delete result: $monitorDeleteResult"
-            )
+            // verify disk is also deleted
+            for {
+              _ <- loggerIO.info(
+                s"AppCreationSpec: app ${googleProject.value}/${appName.value} delete result: $monitorDeleteResult"
+              )
+              getDiskResp <- streamUntilDoneOrTimeout(
```
rtitle: I made this retry the `getDisk` call.. it looks like monitoring `DiskStatus.Deleting -> DiskStatus.Deleted` in back Leo happens asynchronously to moving the app to `Deleted`. So there was a race condition here.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -609,6 +610,12 @@ class GKEInterpreter[F[_]: Parallel: ContextShift: Timer](
       // delete the namespace only after the helm uninstall completes
       _ <- kubeService.deleteNamespace(dbApp.cluster.getGkeClusterId,
                                        KubernetesNamespace(dbApp.app.appResources.namespace.name))
+
+      fa = kubeService
```
rtitle, line 614: Thinking we may not need this if the test is the problem (see my comment above). It probably doesn't hurt though.

rtitle (approved):
Added a commit, will see how tests do. The other stuff in this PR (adding trace id to error tables, etc) seems fine to me.

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/apps/AppCreationSpec.scala
Qi77Qi: yea..that makes sense....should've read the existing test code a bit closer :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
Qi77Qi: I think this will still help...at very least, an error will be persisted to DB when namespace fail to delete

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/apps/AppCreationSpec.scala
Qi77Qi: I don't think the deletion happens async, no? see [here](https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala#L995)

Qi77Qi: think I'm going to remove the wait for now and monitor tests (also need to add back the assertion)

rtitle: Hm yeah originally I thought disk monitoring happened async but I guess not since `sync=true`.

Maybe this failure was test code related too.. if Galaxy deletion fails then the disk would get stuck in `Deleting`. In that case moving this check to the `else` block here should help too.

Qi77Qi: oh yea....I'm definitely moving the check into the branch(thanks for spotting that), just don't think the wait is needed

rtitle (approved):
:shipit:

PR #1935, opened 2021-03-18
[IA-2574] Fix automation tests to work with Referer header

gpcarr:
Update Leo automation tests to pass `Referer` header in calls to the Leo proxy, so we can enable referer validation in all environments (https://broadworkbench.atlassian.net/browse/IA-2574).

This is trickier than one might expect because Selenium does not support this directly. Our approach is to inject some Javascript to cause Selenium to _redirect_ to the page under test, thereby setting the Referer.

To do this, we replace the `CookieAuthedPage` base trait with `ProxyRedirectPage` which does the redirect.

The old `CookieAuthedPage` did:
1. go to the page (gets 401 initially)
2. sets the cookie via Selenium
3. reloads the page, proceeds with test

The new `ProxyRedirectPage` does:
1. go to the page (gets 401 initially)
2. sets the cookie via Selenium
3. loads custom Javascript from local server, which redirects to the page under test
4. proceeds with test

I confirmed all tests pass from Jenkins and locally against a Leo fiab with Referer validation enabled.

---

In all cases:


automation/src/test/resources/redirect-proxy-page.html
```diff
@@ -0,0 +1,20 @@
...
+    <meta charset="UTF-8" name="referrer" content="always"/>
+    <title>Redirect Proxy Page</title>
+    <script type="text/javascript">
+        var proxyUrl = $(proxyUrl);
+
+        function load()
+        {
+            console.log("hiiiiii");
+            window.location.href = proxyUrl;
+            window.history.replaceState(null, '', proxyUrl);
```
rtitle: Is `replaceState` actually needed?

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/Notebook.scala
```diff
@@ -92,8 +92,8 @@ object Notebook extends RestClient with LazyLogging {
     val path = contentsPath(googleProject, clusterName, contentPath) + (if (includeContent) "?content=1" else "")
     logger.info(s"Get notebook contents: GET /$path")
     val cookie = Cookie(HttpCookiePair("LeoToken", token.value))
-
-    handleContentItemResponse(parseResponse(getRequest(url + path, httpHeaders = List(cookie))))
+    val referer = Referer(Uri("https://leonardo.dsde-dev.broadinstitute.org"))
+    handleContentItemResponse(parseResponse(getRequest(url + path, httpHeaders = List(cookie, referer))))
```
rtitle, line 109: I think there are other direct proxy API calls in this class.. not sure how all are used but probably should update all of them with Referer.

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookTestUtils.scala
```diff
@@ -30,8 +32,16 @@ trait NotebookTestUtils extends LeonardoTestUtils {
 
   def withNotebooksListPage[T](cluster: ClusterCopy)(testCode: NotebooksListPage => T)(implicit webDriver: WebDriver,
                                                                                        token: AuthToken): T = {
-    val notebooksListPage = Notebook.get(cluster.googleProject, cluster.clusterName)
-    testCode(notebooksListPage.open)
+    val bindingFuture = ProxyRedirectClient.startServer
+    val testResult = Try {
+      val proxyRedirectPage = ProxyRedirectClient.get(cluster.googleProject, cluster.clusterName, "jupyter")
+      proxyRedirectPage.open
+      testTimer.sleep(1 minutes).unsafeRunSync()
```
rtitle: Just a reminder to remove sleeps before merge

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/ProxyRedirectClient.scala
```diff
@@ -20,16 +19,18 @@ import scala.io.Source
...
 
   private val url = LeonardoConfig.Leonardo.apiUrl
 
-  def get(googleProject: GoogleProject, clusterName: RuntimeName)(implicit token: AuthToken,
-                                                                  webDriver: WebDriver): DummyClientPage = {
+  def get(googleProject: GoogleProject, clusterName: RuntimeName, tool: String)(
+    implicit webDriver: WebDriver,
+    token: AuthToken
+  ): ProxyRedirectPage = {
     val localhost = java.net.InetAddress.getLocalHost.getHostName
```
rtitle: Curious what `java.net.InetAddress.getLocalHost.getHostName` returns in a Jenkins environment.. and whether that works with Leo Referer validation.

automation/src/test/resources/redirect-proxy-page.html
gpcarr: nope, just removed it

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/Notebook.scala
gpcarr: just updated all the API calls, thanks for pointing that out :)

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/ProxyRedirectClient.scala
gpcarr: we may have to see what that returns per environment and add it to our valid referers in firecloud-develop

rtitle:
This isn't quite working because of how we set cookies I think.. doing some more refactoring to `ProxyRedirectPage to make it a supertrait of `Jupyter`/`RStudio``Page` so the redirect happens automatically.

rtitle:
All tests passed!!!

I still need to test with Referer validation on, but I think it'll work

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoSuite.scala
```diff
@@ -112,6 +115,14 @@ trait GPAllocBeforeAndAfterAll extends GPAllocUtils with BeforeAndAfterAll {
         case Right(billingProject) =>
           IO(sys.props.put(gpallocProjectKey, billingProject.value)) >> createInitialRuntime(billingProject)
       }
+      bindAttempt <- IO.fromFuture(IO(ProxyRedirectClient.startServer)).attempt
```
rtitle, line 118: Start a local server for serving the redirect JS. This is done in the test suite `beforeAll`, so it only happens once.

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/ProxyRedirectClient.scala
```diff
@@ -0,0 +1,57 @@
...
+import com.typesafe.scalalogging.LazyLogging
+import org.broadinstitute.dsde.workbench.ResourceFile
+import org.broadinstitute.dsde.workbench.service.RestClient
+
+import scala.concurrent.Future
+import scala.io.Source
+
+object ProxyRedirectClient extends RestClient with LazyLogging {
+  // Note: change to "localhost" if running tests locally
+  val host = java.net.InetAddress.getLocalHost.getHostName
```
rtitle, line 17: Important to note, in Jenkins we need to use `java.net.InetAddress.getLocalHost.getHostName` as the hostname for loading the proxy redirect client. This hostname gets passed to Leo as the `Referer`.

I think this is because the automation tests run in a docker container in Jenkins, and this is how the host is configured in the docker network. Could probably configure the docker container to work with "localhost", but that is a little scary to touch.

automation/src/test/scala/org/broadinstitute/dsde/workbench/page/ProxyRedirectPage.scala
```diff
@@ -0,0 +1,30 @@
...
+import org.broadinstitute.dsde.workbench.auth.AuthToken
+import org.broadinstitute.dsde.workbench.leonardo.ProxyRedirectClient
+import org.broadinstitute.dsde.workbench.service.test.WebBrowserUtil
+import org.openqa.selenium.WebDriver
+import org.scalatestplus.selenium.Page
+
+trait ProxyRedirectPage[P <: Page] extends Page with PageUtil[P] with WebBrowserUtil with LazyLogging { self: P =>
+  implicit val authToken: AuthToken
+
+  override def open(implicit webDriver: WebDriver): P = {
```
rtitle, line 13: This logic took a lot of time to figure out

http/src/test/resources/reference.conf
```diff
@@ -114,7 +114,8 @@ proxy {
 
 refererConfig {
   validHosts = [
-    "example.com"
+    "example.com",
+    "*:9099"
```
rtitle, line 118: Added support for `*:port` in `refererConfig`. 

This is only used in dev/qa, and is needed because the Referer hostname is dynamic based on the Jenkins test run (see above comment).

rtitle:
All tests passed against a fiab with `Referer` validation enabled! 
https://fc-jenkins.dsp-techops.broadinstitute.org/job/leonardo-fiab-test-runner/34391/

http/src/test/resources/reference.conf
xuscrbw: Just have question about the * notation, does this wildcard work for other operations as well? Haven't seen this before so just want to know how it works

xuscrbw (approved):
LGTM

http/src/test/resources/reference.conf
rtitle: The way it's implemented `*` only works for the hostname, not the port. It's to handle the case where automation test referers come from different hosts but are always the same port.

Perhaps I can add a comment to this effect in the fc-develop PR.

PR #1937, opened 2021-03-19
IA-2541 Add logging for gke pollOperation and fix starting timeout bug

Qi77Qi:
1. Use wb-libs published in https://github.com/broadinstitute/workbench-libs/pull/554, which adds logging for gke pollOperation
2. This PR also address a few bugs revealed by [this ticket](https://broadworkbench.atlassian.net/browse/IA-2541)
*  when a VM times out in starting, we should stop runtime, update DB, and that should monitor the new status transition. But we have a bug currently that doesn't monitor the new status transition properly
```
"status transitioned from Starting -> Stopping. This could be caused by a new status transition call!"
```
* When we see a VM is `Terminated` in Google, we shouldn't keep checking if the VM will eventually become `Running`. Instead, we should update status in DB and terminate the monitor stream.

A bit more context for what happened to the VM in the bug ticket (below is a few log entries that demonstrates what happened)
Screen Shot 2021-03-19 at 3 00 42 PM
Screen Shot 2021-03-19 at 3 00 58 PM
Screen Shot 2021-03-19 at 3 01 12 PM

As you can see from the log. Leo received request to start the VM, VM fails to start for some reason (I checked VM's log, and found it was shutdown due to`Instance terminated by guest OS shutdown.`), VM then transitions to `Terminated`, Leo keeps expecting the VM to come up eventually (a bug should be fixed in this PR), which will never happens, and time out kicks in (which didn't terminate the monitor stream as expected due to bug that should be fixed in this PR)

---

In all cases:

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/runtimeModels.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/runtimeModels.scala
@@ -70,29 +70,60 @@ object Runtime {
 }
 
 /** Runtime status enum */
-sealed trait RuntimeStatus extends EnumEntry with Product with Serializable
+sealed trait RuntimeStatus extends EnumEntry with Product with Serializable {
+  // end status if it is a transition status. For instance, terminalStatus for `Starting` is `Running`
+  def terminalStatus: Option[RuntimeStatus]
+}
 object RuntimeStatus extends Enum[RuntimeStatus] {
   val values = findValues
   // Leonardo defined runtime statuses.
 
   // These statuses exist when we save a runtime during an API request, but we haven't published the event to back leo
-  case object PreCreating extends RuntimeStatus
-  case object PreDeleting extends RuntimeStatus
-  case object PreStarting extends RuntimeStatus
-  case object PreStopping extends RuntimeStatus
+  case object PreCreating extends RuntimeStatus {
+    override def terminalStatus: Option[RuntimeStatus] = Some(Running)
+  }
+  case object PreDeleting extends RuntimeStatus {
+    override def terminalStatus: Option[RuntimeStatus] = Some(Deleted)
+  }
+  case object PreStarting extends RuntimeStatus {
+    override def terminalStatus: Option[RuntimeStatus] = Some(Running)
+  }
+  case object PreStopping extends RuntimeStatus {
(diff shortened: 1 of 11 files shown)


Qi77Qi:
fiab-start failure..

jenkins retest

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/GceRuntimeMonitor.scala
```diff
@@ -278,11 +278,19 @@ class GceRuntimeMonitor[F[_]: Parallel](
           GceInstanceStatus.Suspending,
           GceInstanceStatus.Provisioning,
           GceInstanceStatus.Staging,
-          GceInstanceStatus.Terminated,
           GceInstanceStatus.Suspended,
           GceInstanceStatus.Stopping
         )
         r <- gceStatus match {
+          // This happens when a VM fail to start. Potential causes are: start up script failure,
```
jdcanas: `fails`. Maybe change this to something like 'potential known causes' and remove trailing comma?

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/runtimeModels.scala
```diff
@@ -70,29 +70,60 @@ object Runtime {
 }
 
 /** Runtime status enum */
-sealed trait RuntimeStatus extends EnumEntry with Product with Serializable
+sealed trait RuntimeStatus extends EnumEntry with Product with Serializable {
+  // end status if it is a transition status. For instance, terminalStatus for `Starting` is `Running`
+  def terminalStatus: Option[RuntimeStatus]
```
Qi77Qi, line 75: right now, `terminalStatus` is only used to make log message a bit more informative

rtitle (approved):
Had to stare at it a bit but this makes sense. 👍 I guess this handles 2 cases:

1. Google internally shuts down the VM during the starting process. Leo now handles this and correctly moves the rutnime to `Stopped` status.

2. Starting times out (for whatever reason). Leo then explicitly stops the instance and monitors the `Stopping` -> `Stopped` transition.

Qi77Qi:
jenkins retest

Qi77Qi:
random fiab error...jenkins retest

PR #1941, opened 2021-03-22
QA-1385: dismiss rstudio popup

rtitle:

---

In all cases:


xuscrbw (approved):
LTGM. 

One thing I was wondering, but never was able to confirm, was if the problematic workflow ends up being: 

Press Keys
UI takes a second to bring up the popup
The IF block misses this check

Should still be fine to merge if the tests pass. Just haven't been able to figure out if the autocomplete window stays open as we type more, or if it takes a second after we type to appear

rtitle:
Good point @xuscrbw . The test still failed with:
image
Maybe will try to tweak it..

rtitle:
This passed - going to merge and observe dev runs.

PR #1944, opened 2021-03-23
Add&Update metric handling for Leo Proxy [IA-2531]

ehanna4:
1. Update the existing metric counters to follow the name/tag naming convention laid out by Gabriella in the design docs for the Metrics dashboards.
2. Add counters for fail/success on Proxy requests
3. Add gauge on size of notebooks files.

Will create the dashboards outlined in the ticket accordingly.

---

In all cases:


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/ProxyRoutes.scala
```diff
@@ -237,17 +239,28 @@ class ProxyRoutes(proxyService: ProxyService, corsSupport: CorsSupport, refererC
     for {
       ctx <- ev.ask[AppContext]
       apiCall = proxyService.proxyRequest(userInfo, googleProject, runtimeName, request)
-      _ <- metrics.incrementCounter("proxyRuntime")
       resp <- ctx.span.fold(apiCall)(span => spanResource[IO](span, "proxyRuntime").use(_ => apiCall))
+
+      headerMap: Map[String, String] = request.headers.map(header => (header.name(), header.value())).toMap
+      contentLength = headerMap.get("content-length").toString.toDouble
+      _ <- if (request.uri.toString() contains ".ipynb") {
```
kyuksel: Does it make sense to check if `request.uri` not only `contains` `.ipynb` but also ends with it? Along those lines, is there a pattern that `uri`s are expected to start with? If so, could be good to check that as well in hopes of reducing the likelihood of detecting only the desired requests.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/ProxyRoutes.scala
```diff
@@ -237,17 +239,28 @@ class ProxyRoutes(proxyService: ProxyService, corsSupport: CorsSupport, refererC
     for {
       ctx <- ev.ask[AppContext]
       apiCall = proxyService.proxyRequest(userInfo, googleProject, runtimeName, request)
-      _ <- metrics.incrementCounter("proxyRuntime")
       resp <- ctx.span.fold(apiCall)(span => spanResource[IO](span, "proxyRuntime").use(_ => apiCall))
+
+      headerMap: Map[String, String] = request.headers.map(header => (header.name(), header.value())).toMap
+      contentLength = headerMap.get("content-length").toString.toDouble
+      _ <- if (request.uri.toString().endsWith(".ipynb") && request.method.equals(Method.PUT)) {
```
jdcanas: nit: omit `()` for `toString` if possible

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/ProxyRoutes.scala
```diff
@@ -237,17 +239,28 @@ class ProxyRoutes(proxyService: ProxyService, corsSupport: CorsSupport, refererC
     for {
       ctx <- ev.ask[AppContext]
       apiCall = proxyService.proxyRequest(userInfo, googleProject, runtimeName, request)
-      _ <- metrics.incrementCounter("proxyRuntime")
       resp <- ctx.span.fold(apiCall)(span => spanResource[IO](span, "proxyRuntime").use(_ => apiCall))
+
+      headerMap: Map[String, String] = request.headers.map(header => (header.name(), header.value())).toMap
+      contentLength = headerMap.get("content-length").toString.toDouble
```
jdcanas: I don't think this line is quite correct
`Map[String, String]` `get` function returns `Option[String]`. Hence the need for `.toString` to satisfy the compiler (essentially cheating `Option[String]` to `String`), but this won't have the desired effect if the key doesn't exist, and will definitely cause a program crash if you try to call `toDouble` on `None`. 

It should be something like `headerMap.get("content-length").map(_.toDouble)`

This brings us to the fact that `.toDouble` isn't safe, and will cause the program to crash if for some reason `content-length` is not parsbable. Not saying it could happen, but I would prefer to do something like

```
def safeParseDouble(s: String): Option[Double] = try { Some(s.toDouble) } catch { case _ => None }
```
now `contentLength` will have type `Option[Double]` and can be used as such
`_ <- contentLength.traverse(c =>  metrics.gauge("proxy/notebooksSize", c))`

Sorry if this is a lot but I thought it could be helpful since typing is important to pay attention to in scala 😄 

edit: oof, commented on a line that's now changed, i forgot to hit the submit when i typed it up. It still is relevant to remove the third conditional from the if statement though

ehanna4: thank you for the all the details! I think i have made an appropriate fix but please let me know if it doesn't fit with what you were thinking here!

kyuksel: Thanks for the catch, Justin! A follow-up note:

`try-catch` clauses are typically discouraged in Scala and FP in favour of encoding failures using types and Scala provides the handy `Try` we could use here, like so:

```scala
.
.
contentLength = Try(headerMap.get("content-length").map(_.toDouble)).toOption.flatten
_ <- contentLength.traverse(c => metrics.gauge("proxy/notebooksSize", c))
.
.
```

For a simple example using the Scala REPL:
```
scala> Try(Map("A" -> "1").get("A").map(_.toDouble)).toOption.flatten
res16: Option[Double] = Some(1.0)

scala> Try(Map("A" -> "Not a number").get("A").map(_.toDouble)).toOption.flatten
res17: Option[Double] = None
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/ProxyRoutes.scala
```diff
@@ -237,17 +241,29 @@ class ProxyRoutes(proxyService: ProxyService, corsSupport: CorsSupport, refererC
...
       ctx <- ev.ask[AppContext]
       apiCall = proxyService.proxyRequest(userInfo, googleProject, runtimeName, request)
-      _ <- metrics.incrementCounter("proxyRuntime")
       resp <- ctx.span.fold(apiCall)(span => spanResource[IO](span, "proxyRuntime").use(_ => apiCall))
+
+      headerMap: Map[String, String] = request.headers.map(header => (header.name(), header.value())).toMap
+      contentLength = Try(headerMap.get("content-length").map(_.toDouble)).toOption.flatten
+
+      _ <- if (request.uri.toString.endsWith(".ipynb") && request.method.equals(Method.PUT)) {
+        contentLength.traverse(size => metrics.gauge("proxy/notebooksSize", size))
```
kyuksel: `traverse(...)` will be executed only if `contentLength` is a `Some(...)`. For instance, if `content-length` parsed from a request is not anything cast-able to `Double` then nothing will be recorded as a `notebooksSize` metric…which sounds reasonable to me but I wanted to make sure that was the intention.

kyuksel (approved):
LGTM 👍🏼  One last inline comment to be sure 😊

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/ProxyRoutes.scala
ehanna4: yeah i think that's okay, good catch though!

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/ProxyRoutes.scala
```diff
@@ -237,17 +241,29 @@ class ProxyRoutes(proxyService: ProxyService, corsSupport: CorsSupport, refererC
     for {
       ctx <- ev.ask[AppContext]
       apiCall = proxyService.proxyRequest(userInfo, googleProject, runtimeName, request)
-      _ <- metrics.incrementCounter("proxyRuntime")
       resp <- ctx.span.fold(apiCall)(span => spanResource[IO](span, "proxyRuntime").use(_ => apiCall))
+
+      headerMap: Map[String, String] = request.headers.map(header => (header.name(), header.value())).toMap
```
rtitle: I think this could be a bit more concise:
```
_ <- request.header[`Content-Length`].traverse { h => 
  metrics.gauge("proxy/notebooksSize", h.length)
}
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/ProxyRoutes.scala
```diff
@@ -237,17 +241,29 @@ class ProxyRoutes(proxyService: ProxyService, corsSupport: CorsSupport, refererC
...
+      contentLength = Try(headerMap.get("content-length").map(_.toDouble)).toOption.flatten
+
+      _ <- if (request.uri.toString.endsWith(".ipynb") && request.method.equals(Method.PUT)) {
+        contentLength.traverse(size => metrics.gauge("proxy/notebooksSize", size))
+      } else IO.unit
+
+      _ <- if (resp.status.isSuccess()) {
```
rtitle, line 260: Should we use this pattern (count successes and failures separately) for other methods (e.g. apps)?

ehanna4: oo I like that, seems like a useful thing to collect👍

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/ProxyRoutes.scala
ehanna4: IntelliJ is telling me no on using the `.length`, apparently it returns a Long but the `metric.gauge` requires a Double and so it is saying using `toDouble` is necessary (which then is going bring us back to the issue Justin brought up) so I think I'll have to leave it a little more complicated

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/ProxyRoutes.scala
kyuksel: ➕

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/ProxyRoutes.scala
rtitle: Ok. Still might be better to use the built-in header lookup method instead of iterating through the list though. Also I think we can only do this when necessary to save performance:
```
_ <- if (request.uri.path.endsWith("*.ipynb") && request.method == HttpMethods.PUT) {
  val contentLengthOpt = request.header[`Content-Length`].flatMap(h => Try(h.length.toDouble).toOption)
  contentLengthOpt.traverse { c =>
    metrics.gauge("proxy/notebooksSize", c)
  }
} else IO.unit
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/ProxyRoutes.scala
```diff
@@ -237,17 +251,29 @@ class ProxyRoutes(proxyService: ProxyService, corsSupport: CorsSupport, refererC
     for {
       ctx <- ev.ask[AppContext]
       apiCall = proxyService.proxyRequest(userInfo, googleProject, runtimeName, request)
-      _ <- metrics.incrementCounter("proxyRuntime")
       resp <- ctx.span.fold(apiCall)(span => spanResource[IO](span, "proxyRuntime").use(_ => apiCall))
+
+      headerMap: Map[String, String] = request.headers.map(header => (header.name(), header.value())).toMap
+      contentLength = Try(headerMap.get("content-length").map(_.toDouble)).toOption.flatten
+
+      _ <- if (request.uri.toString.endsWith(".ipynb") && request.method.equals(Method.PUT)) {
```
rtitle: I think `Method.PUT` is an http4s object.. probably want `HttpMethods.PUT` which is akka-http.

Also in Scala generally use `==` instead of `.equals` -- unlike Java this tests object equality and is more type-safe.

rtitle (approved):
LGTM 👍 

Can you also put the IA ticket in the PR title?

PR #1961, opened 2021-03-29
Revert "Add&Update metric handling for Leo Proxy [IA-2531]"

ehanna4:
Reverts DataBiosphere/leonardo#1944

Screen Shot 2021-03-29 at 12 41 59 PM

^ something is unhappy with the new proxy metrics need to debug and redeploy. For now want to revert as it is breaking notebooks in dev


rtitle (approved):

rtitle:
@ehanna4 shall we merge this or is there a fix? We probably shouldn't leave dev in this state too long..

PR #1966, opened 2021-03-30
[IA-2455] Update/Add Metrics for Proxy Request Counts

ehanna4:
Adds/updates the counting of proxy requests per the new naming convention.

Name: proxyRequest
- result
- action
- tool

Name: proxyAuthRequest
- type

Also adds a metric for tracking notebook sizes on notebook saves
Name: notebooksSize
- source

Tested through a fiab - please see https://console.cloud.google.com/monitoring/dashboards/builder/7c01affa-ea94-40cd-aa73-33c9e391d6f5?authuser=3&project=broad-dsp-stackdriver&dashboardBuilderState=%257B%2522editModeEnabled%2522:true%257D&timeDomain=1h for WIP graphs showing the metrics coming through

---

In all cases:


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/ProxyRoutes.scala
```diff
@@ -224,8 +232,19 @@ class ProxyRoutes(proxyService: ProxyService, corsSupport: CorsSupport, refererC
     for {
       ctx <- ev.ask[AppContext]
       apiCall = proxyService.openTerminal(userInfo, googleProject, runtimeName, terminalName, request)
-      _ <- metrics.incrementCounter("openTerminal")
       resp <- ctx.span.fold(apiCall)(span => spanResource[IO](span, "openTerminal").use(_ => apiCall))
+      tool = if (request.uri.toString.contains("jupyter/terminals/")) {
```
rtitle: I think this method is only called from the `"jupyter" / "terminals"` route so this `if` is redundant.. but it doesn't hurt.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/ProxyRoutes.scala
```diff
@@ -237,17 +256,38 @@ class ProxyRoutes(proxyService: ProxyService, corsSupport: CorsSupport, refererC
     for {
       ctx <- ev.ask[AppContext]
       apiCall = proxyService.proxyRequest(userInfo, googleProject, runtimeName, request)
-      _ <- metrics.incrementCounter("proxyRuntime")
       resp <- ctx.span.fold(apiCall)(span => spanResource[IO](span, "proxyRuntime").use(_ => apiCall))
+      tool = if (request.uri.toString.contains("jupyter")) {
```
rtitle: I think we may have an ADT for this .. `RuntimeContainerServiceType`. 

I wonder if this would work (I haven't compiled it :)):
```
tool = RuntimeContainerServiceType.values.find(s => request.uri.toString.contains(s.proxySegment)).map(_.imageType.name).getOrElse("other")
```

I don't feel too strongly, but seems nice to not duplicate the proxy segment / tool name strings.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/ProxyRoutes.scala
```diff
@@ -237,17 +256,38 @@ class ProxyRoutes(proxyService: ProxyService, corsSupport: CorsSupport, refererC
...
+      } else {
+        "other"
+      }
+
+      _ <- if (resp.status.isSuccess()) {
+        metrics.incrementCounter("proxyRequest",
+                                 tags = Map("result" -> "success", "action" -> "runtimeRequest", "tool" -> s"${tool}"))
+
+        if (request.uri.toString.endsWith(".ipynb") && request.method == HttpMethods.PUT) {
+          metrics.gauge("proxy/notebooksSize", resp.entity.getContentLengthOption().getAsLong.toDouble)
```
rtitle: I think `.getContentLengthOption().getAsLong` is the akka Java API, and probably throws an exception if the optional is not present.

I would do:
```
resp.entity.contentLengthOption.traverse { c =>
  metrics.gauge("proxy/notebooksSize", c.toDouble)
}
```
which is the Scala API and safer `Option`-handling. (I don't think it's too important to check the `.toDouble` conversion.)

rtitle: Also just realized this is using the `HttpResponse`. Do we want to do it on the `HttpRequest` instead? (Since I think the PUT _request_ would have the right content-length.)

ehanna4: oops thank you yes should be the request

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/ProxyRoutes.scala
ehanna4: oh yes that's much better thanks

rtitle (approved):
👍

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/ProxyRoutes.scala
```diff
@@ -237,17 +256,35 @@ class ProxyRoutes(proxyService: ProxyService, corsSupport: CorsSupport, refererC
...
-      _ <- metrics.incrementCounter("proxyRuntime")
       resp <- ctx.span.fold(apiCall)(span => spanResource[IO](span, "proxyRuntime").use(_ => apiCall))
+
+      tool = RuntimeContainerServiceType.values
+        .find(s => request.uri.toString.contains(s.proxySegment))
+        .map(_.imageType.entryName)
+        .getOrElse("other")
+
+      _ <- if (request.uri.toString.endsWith(".ipynb") && request.method == HttpMethods.PUT) {
+        request.entity.contentLengthOption.traverse(size => metrics.gauge("proxy/notebooksSize", size.toDouble))
```
gpcarr: going forward we want to keep metric names to a single string (with no additional `/`s) after `OpenCensus/leonardo/` so that we have a uniform naming pattern. We can think about extracting as much as possible into tags so maybe for this metric name, we can use `notebookSize` with tag `source: proxy` or something like that. then going forward we can add other notebook attribute metrics with same or different sources

gpcarr (approved):
LGTM 👍 would be curious to know if you're able to create the graphs mentioned in the original ticket under project `broad-dsde-dev` (using your fiab) with the tags added in this PR. could be a good thing to test before merging

ehanna4:
> LGTM 👍 would be curious to know if you're able to create the graphs mentioned in the original ticket under project `broad-dsde-dev` (using your fiab) with the tags added in this PR. could be a good thing to test before merging

Do you have access with this link: https://console.cloud.google.com/monitoring/dashboards/builder/7c01affa-ea94-40cd-aa73-33c9e391d6f5?authuser=3&project=broad-dsp-stackdriver&dashboardBuilderState=%257B%2522editModeEnabled%2522:true%257D&timeDomain=1h

I've got some WIP graphs made for dev/fiab

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/ProxyRoutes.scala
```diff
@@ -237,17 +256,36 @@ class ProxyRoutes(proxyService: ProxyService, corsSupport: CorsSupport, refererC
...
+
+      } else {
+        metrics.incrementCounter("proxyRequest",
+                                 tags = Map("result" -> "failure", "action" -> "runtimeRequest", "tool" -> s"${tool}"))
+      }
     } yield resp
 
   private[api] def setCookieHandler(userInfoOpt: Option[UserInfo]): IO[ToResponseMarshallable] =
-    metrics.incrementCounter("setCookie") >>
+    metrics.incrementCounter("proxyAuthRequest", tags = Map("type" -> "setCookie")) >>
```
Qi77Qi: maybe use `api` instead of `type`? `type` sounds super vague to me

PR #1996, opened 2021-04-02
[IA-2619] Re-use a nodepool only if it has the requested config

kyuksel:
[JIRA ticket](https://broadworkbench.atlassian.net/browse/IA-2619)


kyuksel:
`leonardo-build`:
```
java.sql.SQLSyntaxErrorException: Table 'leotestdb.LABEL' doesn't exist
```
🤔 

jenkins retest

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/NodepoolComponent.scala
```diff
@@ -131,6 +131,21 @@ object nodepoolQuery extends TableQuery(new NodepoolTable(_)) {
       claimedNodepoolOpt <- unclaimedNodepoolOpt.flatTraverse(n => getMinimalById(n.id))
     } yield claimedNodepoolOpt
 
+  def claimNodepoolWithConfig(clusterId: KubernetesClusterLeoId,
```
rtitle: An aside: @jdcanas discussed in our 1-1 removing the "claim nodepool" functionality. It's not used in production and we've never actually used it in a workshop. So it really just adds complexity with no real benefit.

(You don't need to remove it in this PR though.)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
```diff
@@ -90,23 +89,36 @@ final class LeoAppServiceInterp[F[_]: Parallel](
 
       clusterId = saveClusterResult.minimalCluster.id
 
-      //We want to know if the user already has an app to re-use that nodepool
-      userNodepoolOpt <- nodepoolQuery.getMinimalByUser(userInfo.userEmail, googleProject).transaction
+      machineConfig = req.kubernetesRuntimeConfig.getOrElse(
```
rtitle, line 92: I just realized a difference between the app and runtime APIs. For runtimes, we have a `RuntimeConfigRequest` type where each field is optional so we can have defaults for each individual field:
https://github.com/DataBiosphere/leonardo/blob/dfe49dfcf0291070ff04634ed760a09ceefe7dfb/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/runtimeRoutesModels.scala#L34-L39

But for apps, the create request just takes a `Option[KubernetesRuntimeConfig]` where the individual fields are non-optional:
https://github.com/DataBiosphere/leonardo/blob/develop/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala#L440-L441

I don't think this needs to be changed right now (and the code in this PR is correct) -- it just made me pause for a minute while reviewing. :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
```diff
@@ -90,23 +89,36 @@ final class LeoAppServiceInterp[F[_]: Parallel](
...
 
-      //A claimed nodepool is either the user's existing nodepool, or a pre-created one in that order of precedence
+      // A claimed nodepool is either the user's existing nodepool or a pre-created one, in that order of precedence
       claimedNodepoolOpt <- if (userNodepoolOpt.isDefined) F.pure(userNodepoolOpt)
-      else nodepoolQuery.claimNodepool(clusterId, userInfo.userEmail).transaction
+      else nodepoolQuery.claimNodepoolWithConfig(clusterId, userInfo.userEmail, machineConfig).transaction
 
       nodepool <- claimedNodepoolOpt match {
+        case Some(n) =>
+          log.info(
```
rtitle: Can we put traceId in logs like:
```
log.info(ctx.loggingCtx)(msg)
```

rtitle:
Overall looks good.. could we add a unit test in `AppServiceInterpSpec`? Maybe it could create 2 apps with different compute configs, and verify 2 nodepools get created.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/NodepoolComponent.scala
kyuksel: created a [ticket](https://broadworkbench.atlassian.net/browse/IA-2644)

kyuksel:
@rtitle I've addressed your comments and also removed some duplicate code. Could you please take another look?

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterpSpec.scala
```diff
@@ -127,6 +127,69 @@ final class AppServiceInterpSpec extends AnyFlatSpec with LeonardoTestSuite with
     app1.nodepool.id shouldBe app2.nodepool.id
   }
 
+  it should "result in new nodepool creations when apps request distinct nodepool configurations" in isolatedDbTest {
```
rtitle, line 130: 👍  nice test

rtitle (approved):
LGTM!

kyuksel:
leo build failure

jenkins retest

kyuksel:
jenkins retest

kyuksel:
jenkins retest

kyuksel:
Merging since
- 5 of 7 integration test runs succeeded where the remaining two failed due to `leonardo-build` failures that seemed to be due to DB issues.
- the missing diff pointed out by `codecov/path` is for the `log.info` lines I refactored to embed Trace IDs, and the lines around defining a default `kubernetesRuntimeConfig` so I'm fine with unit tests not hitting those

PR #1997, opened 2021-04-02
[IA-2543] Explicitly set helm versions in dockerfile

jdcanas:
We specify these versions in the Dockerfile now as well as in reference.conf.

The default behavior of helm pull is to pull the latest, so there's no guarantee that versions specified in `.conf` will even be pulled.

We specify these versions in the `.conf` as well because it allows easier development and hotfixing. You can overwrite these values without rebuilding the docker image this way.


Dockerfile
```diff
@@ -19,7 +19,6 @@ EXPOSE 8080
 EXPOSE 5050
 
 ENV GIT_HASH $GIT_HASH
-# TODO: remove this
 ENV HELM_DEBUG 1
```
jdcanas, line 22: I decided this is useful enough to keep, and only is noisy during failures, which is very helpful to always have on.

Dockerfile
```diff
@@ -43,11 +42,12 @@ RUN helm repo add ingress-nginx https://kubernetes.github.io/ingress-nginx && \
 # pulling `terra-app-setup` locally and add cert files to the chart.
 # Leonardo will install the chart from local version.
 # We are also cacheing charts so they are not downloaded with every helm-install
+# If you are changing any versions here, update it in the reference.conf
 RUN pushd /leonardo && \
-    helm pull terra-app-setup-charts/terra-app-setup --untar && \
-    helm pull galaxy/galaxykubeman --untar && \
-    helm pull terra/terra-app --untar && \
-    helm pull ingress-nginx/ingress-nginx --untar && \
+    helm pull terra-app-setup-charts/terra-app-setup --version 0.0.2 --untar && \
```
Qi77Qi: wonder if it's better to keep these versions as `ENV` on top of the file

Qi77Qi (approved):

rtitle (approved):
:shipit: if tests pass

Dockerfile
```diff
@@ -43,11 +42,12 @@ RUN helm repo add ingress-nginx https://kubernetes.github.io/ingress-nginx && \
 # pulling `terra-app-setup` locally and add cert files to the chart.
 # Leonardo will install the chart from local version.
 # We are also cacheing charts so they are not downloaded with every helm-install
+# If you are changing any versions here, update it in the reference.conf
```
Qi77Qi: how about prefix `WARNING:` cuz it's kinda buried in a blob of text

rtitle: +1

Dockerfile
jdcanas: I don't feel strongly either way. Could make it easier to find, but I personally search for `helm pull` when looking for these versions

kyuksel: +1 on keeping the versions as ENV on top of the file for better visibility

jdcanas:
Fairly suspicious failure in stop/start test... https://fc-jenkins.dsp-techops.broadinstitute.org/job/leonardo-fiab-test-runner/34807/console

I want to run it a few more times here and run it by itself locally

jenkins retest

rtitle:
Same `start app` test failure..

This PR is the first time Leo tests are using the network policy from https://github.com/DataBiosphere/terra-app/pull/10, right? I wonder if there some issue in the `start` flow (e.g. network policy is blocking something it shouldn't..)

If this needs further debugging it might be good to remove `0.0.2` from the helm repo before Monday's release. Alternatively: maybe update this PR to use `TERRA_APP_SETUP_VERSION=0.0.1` and see if it works?

rtitle:
fiab-start failed, looks like `consent`. jenkins retest

PR #2002, opened 2021-04-03
[IA-2360] Migrate Leo to Java 11

rtitle:
https://broadworkbench.atlassian.net/browse/IA-2360

Note: depends on https://github.com/broadinstitute/firecloud-develop/pull/2501

Made changes to use JDK 11. Mostly involves updating Dockerfiles, build scripts, etc. The biggest change here is removing use of `sun.net.spi.nameservice.*` classes, which were [removed in JDK 9](https://bugs.openjdk.java.net/browse/JDK-8134577). Explanation below. 

### Background
Leonardo connects via `https` to many different VMs with different IPs. We install the same certificate on each VM with domain `jupyter-$env.firecloud.org`. The hostname we use for the SSL connection needs to match the domain of the certificate.

### Old way
We were using a custom `sun.net.spi.nameservice.NameService` class to "spoof" DNS. So Leo would connect to `https://<uuid>.jupyter-$env.firecloud.org` and the JVM would consult our custom `NameService` to resolve the actual IP of the runtime. Because the hostname matches the cert, things were happy.

### New way
`sun.net.spi.nameservice.NameService` no longer exists, but luckily akka-http and http4s-blaze have implemented similar functionality:
- akka-http [custom host name resolution](https://doc.akka.io/docs/akka-http/current/client-side/client-transport.html#custom-host-name-resolution-transport)
- blaze [custom dns resolver](https://github.com/http4s/http4s/pull/4699) (just [released](https://github.com/http4s/http4s/releases/tag/v0.21.22) on 4/6/21!)

Each of these allows us to continue to make requests to `https://<uuid>.jupyter-$env.firecloud.org` and have the http client resolve to the correct IP, which makes cert verification happy.

_Big shout-out to @Qi77Qi for discovering this approach!_

### Alternatives Considered

1. Connect to `https://<ip>` and pass a header `Host: https://<uuid>.jupyter-$env.firecloud.org`
   * This works for server-side SSL validation in Apache, which is what Jupyter/RStudio runtimes use.
   * It did NOT work for GKE + nginx ingress, where it seems that using the `Host` header for TLS is not supported.
   * It also broke client-side hostname verification, which is done to prevent [Man in the Middle](https://sites.google.com/site/cse825maninthemiddle/) attacks. Akka-http and blaze had some limited functionality to get around this, but it was a little hacky and involved using insecure/deprecated code.
2. Use `jdk.net.hosts.file`
   * This is a JVM-native way to support custom host name resolution. It was introduced in JDK9 as a replacement for `sun.net.spi.nameservice.NameService`. The difference is that the host->ip mapping is contained in a file (e.g. analogous to `/etc/hosts`).
   * We could probably get this to work using concurrency/file-io primitives in cats-effect and fs2. But I _really_ don't like the idea of dynamically updating a file in the Leo proxy.  I predict performance bottlenecks and unexpected issues (what if the file gets corrupted?). This is even less appealing in the k8s world where pods/containers should be stateless.
3. Use real DNS
   * Instead of trying to spoof DNS in Java, actually do it for real. E.g. when a runtime is created, create a Cloud DNS entry for its IP, and have Leo connect to that. This seems to be what AI Platform Notebooks does: when you create a notebook instance you get a minted domain like `https://2131bf5b9a780536-dot-us-central1.notebooks.googleusercontent.com`. We'd still need to have the Leo proxy to do authorization, but we could remove the custom DNS stuff entirely. This is maybe a good long-term solution, but involves some additional Google setup (and potential migration of old runtimes). So I'm thinking we could have a backlog ticket for this.

---

In all cases:


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/Boot.scala
```diff
@@ -332,34 +326,52 @@ object Boot extends IOApp {
       blockingEc <- ExecutionContexts.cachedThreadPool[F]
       semaphore <- Resource.eval(Semaphore[F](255L))
       blocker = Blocker.liftExecutionContext(blockingEc)
-      storage <- GoogleStorageService.resource[F](pathToCredentialJson, blocker, Some(semaphore))
```
rtitle, line 335: Re-organized this method a little to make it easier to read

project/Settings.scala
```diff
@@ -191,9 +191,7 @@ object Settings {
               s"-Dlogback.configurationFile=${(Test / baseDirectory).value.getAbsolutePath}/src/test/resources/logback-test.xml",
               s"-Djava.util.logging.config.file=${(Test / baseDirectory).value.getAbsolutePath}/src/test/resources/logback-test.xml",
               s"-Dtest.name=${test.name}",
-              s"-Ddir.name=${(Test / baseDirectory).value}",
-              s"-Dheadless=${Option(System.getProperty("headless")).getOrElse("false")}",
-              s"-Djsse.enableSNIExtension=${Option(System.getProperty("jsse.enableSNIExtension")).getOrElse("false")}"))
```
rtitle, line 196: `-Dheadless` and `-Djsse.enableSNIExtention` were breaking automation tests on Java 11. Removing them seems to work as intended.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/Boot.scala
```diff
@@ -332,34 +333,52 @@ object Boot extends IOApp {
...
+                                                  kubernetesDnsCacheConfig,
+                                                  hostToIpMapping,
+                                                  blocker)
+
+      // Set up SSL context and http clients
+      retryPolicy = RetryPolicy[F](RetryPolicy.exponentialBackoff(30 seconds, 5))
+      sslContext <- Resource.eval(SslContextReader.getSSLContext())
+      httpClient <- blaze
+        .BlazeClientBuilder[F](blockingEc, Some(sslContext))
+        .withCustomDnsResolver(proxyResolver.resolveHttp4s)
```
rtitle, line 360: Here's where the custom dns resolver is set in the blaze client

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/ProxyService.scala
```diff
@@ -99,7 +100,10 @@ class ProxyService(
   metrics: OpenTelemetryMetrics[IO],
   loggerIO: StructuredLogger[IO])
     extends LazyLogging {
+
   val httpsConnectionContext = ConnectionContext.httpsClient(sslContext)
+  val clientConnectionSettings =
+    ClientConnectionSettings(system).withTransport(ClientTransport.withCustomResolver(proxyResolver.resolveAkka))
```
rtitle, line 106: Here's where the custom dns resolver is set in akka-http

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/Boot.scala
Qi77Qi, line 360: we only need to use custom dns resolver for proxy requests right?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dns/ProxyResolver.scala
```diff
@@ -0,0 +1,54 @@
...
+
+  // akka-http API
+  // See https://doc.akka.io/docs/akka-http/current/client-side/client-transport.html#custom-host-name-resolution-transport
+  def resolveAkka(host: String, port: Int): Future[InetSocketAddress]
+}
+
+/**
+ * Implementation of ProxyResolver using a Map[Host, IP] stored in a Ref.
+ */
+class ProxyResolverInterp[F[_]](hostToIpMapping: Ref[F, Map[Host, IP]])(
```
Qi77Qi: Make this private?(just to make sure this is not being used as arguments in any class definition directly in other places)  and make a constructor for this in `ProxyResolver`'s companion object?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/Boot.scala
```diff
@@ -332,34 +333,52 @@ object Boot extends IOApp {
...
       implicit0(dbRef: DbReference[F]) <- DbReference.init(liquibaseConfig, concurrentDbAccessPermits, blocker)
-      runtimeDnsCache = new RuntimeDnsCache(proxyConfig, dbRef, runtimeDnsCacheConfig, blocker)
-      kubernetesDnsCache = new KubernetesDnsCache(proxyConfig, dbRef, kubernetesDnsCacheConfig, blocker)
-      welderDao = new HttpWelderDAO[F](runtimeDnsCache, clientWithRetryAndLogging)
-      dockerDao = HttpDockerDAO[F](clientWithRetryAndLogging)
-      jupyterDao = new HttpJupyterDAO[F](runtimeDnsCache, clientWithRetryAndLogging)
-      rstudioDAO = new HttpRStudioDAO(runtimeDnsCache, clientWithRetryAndLogging)
+
+      // Set up DNS caches
+      hostToIpMapping <- Resource.eval(Ref.of(Map.empty[Host, IP]))
```
Qi77Qi, line 343: just to understand the whole flow, `RuntimeDnsCache` and `KubernetesDnsCache` will update the `Ref` whenever we call `getHostStatus`....this does 3 things 1. populate cache (a mapping from basicaly runtime name -> `https://<uuid>.jupyter-$env.firecloud.org`) 2. return host name ( `https://<uuid>.jupyter-$env.firecloud.org`) 3. update `Ref`, which will be used by customDnsResolver to find the IP for `https://<uuid>.jupyter-$env.firecloud.org`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/Boot.scala
rtitle: Hm that's true I guess, though I made the resolver fall back to default resolution if the host is not in the dns cache. But maybe cleaner would be a separate client for proxy requests.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dns/ProxyResolver.scala
rtitle: Sure 👍

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/Boot.scala
Qi77Qi: ah I see...that makes sense...maybe just add a comment about this?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/Boot.scala
rtitle: Yes exactly :)

1. `ProxyService` wants to make a request
2. `ProxyService` calls `RuntimeDnsCache` to get the hostname based on the runtime project/name
3. `RuntimeDnsCache` updates the `Ref` with the host->IP mapping and returns the hostname
4. `ProxyService` makes the call via akka-http
5. akka-http calls custom resolver which looks up the IP from the hostname

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/Boot.scala
rtitle: Sure sounds good

rtitle:
Let's see if multi-test works

rtitle:
multi-test

rtitle:
fiab-start failed on the "regular" test, jenkins retest

rtitle:
3/4 automation runs passed, the 1 failure is in `NotebookRKernelSpec`: https://fc-jenkins.dsp-techops.broadinstitute.org/job/leonardo-fiab-test-runner/35049/artifact/failure_screenshots/NotebookRKernelSpec_15-03-51-826.png

I haven't seen that before, but I can't see how it's Java-related, going to try another multi test.

rtitle:
multi-test

rtitle:
2/3 multi tests again, this time the failure in `leonardo-build`.

rtitle:
multi-test

rtitle:
multi-test

rtitle:
@Qi77Qi applied feedback, and all the multi-tests passed.. :)

Qi77Qi (approved):
:shipit: 🎉

PR #2007, opened 2021-04-05
Removing subversions from GKE version

almahmoud:
Trying to not hardcode specific versions to avoid having to bump GKE subversions within a Kubernetes version.

(This is not tested, just a suggestion as a starting point based on the fact that on the command-line I can do `--cluster-version=1.16`)

Alternatively, can use something like `gcloud container get-server-config 2>&1 | grep -o "1.16.*" | head -1` to get the latest available GKE version of a K8S version.

---

In all cases:

### DIFF
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -189,7 +189,7 @@ gke {
     ]
     # See https://cloud.google.com/kubernetes-engine/docs/release-notes
     # As of 2021-02-22, 1.17.x is the default but the Galaxy chart expects 1.16.x.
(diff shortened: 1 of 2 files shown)


rtitle:
Thanks for making this! Unfortunately the `Jenkins Automation Tests` hook currently won't run on forks, I opened this as a branch instead (and trying 1.18): https://github.com/DataBiosphere/leonardo/pull/2008

rtitle:
Going to close this in favor of https://github.com/DataBiosphere/leonardo/pull/2008

PR #2018, opened 2021-04-07
IA-2601 Let users delete app when no attached disk is found

rtitle:
https://broadworkbench.atlassian.net/browse/IA-2601

Fix this error when I tried to delete an `Error`'d app:

image

I think what happens is:
1. App creation fails
2. [cleanUpAfterCreateAppError](https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala#L856) deletes the disk if it was a newly-created disk in the `createApp` request (which I think is reasonable).
3. End state is app is in `Error` status with no disk

I simply made the code in `AppServiceInterp` tolerate this case and updated a unit test to cover it.

---

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
@@ -674,13 +674,6 @@ case class AppCannotBeDeletedException(googleProject: GoogleProject,
       traceId = Some(traceId)
     )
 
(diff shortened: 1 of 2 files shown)


jdcanas:
Even simplifies the code, nice :shipit:

jdcanas (approved):

Qi77Qi (approved):
👍

kyuksel (approved):
👍🏼

rtitle:
jenkins retest (fiab-start)

PR #2024, opened 2021-04-07
[IA-2644] Remove batch nodepool create

jdcanas:


jdcanas:
jenkins retest, fiab failure

Qi77Qi (approved):
👍

rtitle (approved):
image

Love it, :shipit:

kyuksel (commented):
Thanks for taking care of this, Justin!

I noticed that Swagger also contains related pieces around [batchNodepoolCreate endpoint](https://github.com/DataBiosphere/leonardo/blob/706a7504420ea4bec686d4f761455e8502b2ddf1/http/src/main/resources/swagger/api-docs.yaml#L820) which we should probably remove as well.

Also, is there going to be a cron-jobs PR to remove related code from there as well?

PR #2028, opened 2021-04-08
IA-2640 persist disk zone properly

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-2640

Turns out we're already persisting zone for disks, so it's actually less change than I expected.

* Persist disk's zone properly in `createRuntime` API call (this should fix the bug that @wnojopra ran into)
* Reject a request if createRuntime's zone is not the same as existing PD's zone
* Add `zone` to `createDisk` API

Tested in fiab

---

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/Boot.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/Boot.scala
@@ -147,7 +147,7 @@ object Boot extends IOApp {
       )
       val runtimeService = RuntimeService(
         runtimeServiceConfig,
-        persistentDiskConfig,
+        ConfigReader.appConfig.persistentDiskConfig,
         appDependencies.authProvider,
         appDependencies.serviceAccountProvider,
         appDependencies.dockerDAO,
@@ -157,7 +157,7 @@ object Boot extends IOApp {
         appDependencies.publisherQueue
       )
       val diskService = new DiskServiceInterp[IO](
-        persistentDiskConfig,
+        ConfigReader.appConfig.persistentDiskConfig,
         appDependencies.authProvider,
         appDependencies.serviceAccountProvider,
         appDependencies.publisherQueue
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/Boot.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/Boot.scala
@@ -147,7 +147,7 @@ object Boot extends IOApp {
       )
       val runtimeService = RuntimeService(
         runtimeServiceConfig,
-        ConfigReader.appConfig.persistentDiskConfig,
+        ConfigReader.appConfig.persistentDisk,
         appDependencies.authProvider,
         appDependencies.serviceAccountProvider,
         appDependencies.dockerDAO,
@@ -157,7 +157,7 @@ object Boot extends IOApp {
         appDependencies.publisherQueue
       )
       val diskService = new DiskServiceInterp[IO](
-        ConfigReader.appConfig.persistentDiskConfig,
+        ConfigReader.appConfig.persistentDisk,
         appDependencies.authProvider,
         appDependencies.serviceAccountProvider,
         appDependencies.publisherQueue
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/Boot.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/Boot.scala
@@ -145,7 +145,7 @@ object Boot extends IOApp {
         dataprocConfig,
         gceConfig
       )
-      val runtimeService = new RuntimeServiceInterp[IO](
+      val runtimeService = RuntimeService(
         runtimeServiceConfig,
         persistentDiskConfig,
         appDependencies.authProvider,
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskServiceInterp.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskServiceInterp.scala
@@ -254,7 +254,7 @@ object DiskServiceInterp {
     } yield PersistentDisk(
       DiskId(0),
       googleProject,
-      req.zone.getOrElse(config.zone),
+      req.zone.getOrElse(config.defaultZone),
       diskName,
       None,
       serviceAccount,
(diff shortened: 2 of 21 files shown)


http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeServiceInterpSpec.scala
```diff
@@ -1482,13 +1482,65 @@ class RuntimeServiceInterpSpec extends AnyFlatSpec with LeonardoTestSuite with T
...
+                                      userInfo,
+                                      serviceAccount,
+                                      FormattedBy.GCE,
+                                      whitelistAuthProvider,
+                                      Config.persistentDiskConfig)
+        .attempt
+    } yield {
+      diskResult shouldBe (Left(
+        BadRequestException(
+          s"requested disk ${diskName.value} is not in the same zone as the requested zone ${targetZone.value}",
```
wnojopra: Does the error string need to match exactly?

wnojopra (approved):
Asides from fixing tests, code lgtm

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeServiceInterpSpec.scala
Qi77Qi: ah yea....I updated the message but forgot to update test

Qi77Qi:
leonardo-build failed..jenkins retest

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/Boot.scala
```diff
@@ -145,7 +145,7 @@ object Boot extends IOApp {
         dataprocConfig,
         gceConfig
       )
-      val runtimeService = new RuntimeServiceInterp[IO](
+      val runtimeService = RuntimeService(
```
rtitle, line 148: Should `RuntimeServiceInterp` be made private given we have this constructor?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskServiceInterp.scala
```diff
@@ -254,7 +254,7 @@ object DiskServiceInterp {
     } yield PersistentDisk(
       DiskId(0),
       googleProject,
-      config.zone,
+      req.zone.getOrElse(config.zone),
```
rtitle: Wondering if we should rename `zone` to `defaultZone` in `reference.conf`..

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/Boot.scala
Qi77Qi: yea, I tried..but we're testing some members of `RuntimeServiceInterp` directly in unit tests, it's not as easy to make it private

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskServiceInterp.scala
Qi77Qi: will do

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeServiceInterp.scala
```diff
@@ -903,6 +907,14 @@ object RuntimeServiceInterp {
       disk <- diskOpt match {
         case Some(pd) =>
           for {
+            _ <- if (pd.zone == targetZone) F.unit
+            else
+              F.raiseError(
```
rtitle, line 912: Should this be an error? Wondering if it should create a new disk in this case.

E.g. instead of `persistentDiskQuery.getActiveByName` we could have `persistentDiskQuery.getActiveByNameAndZone` maybe, and if it returns `None` we create a new disk..

rtitle:
Few comments above… also it feels like there should be a DB migration to add `zone` to the `PERSISTENT_DISK` table, no?

Qi77Qi:
@rtitle `zone` was already added to the table...

rtitle:
Ohhh 🤯

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeServiceInterp.scala
rtitle: Not sure which is better, need to think about it..

gpcarr: in terms of persistence, if the UI sends a create runtime request and we try to attach an existing disk to a new runtime but the disk and runtime are in different zones - if we create a new disk then we lose persistence, I think? So maybe error-ing out is best if the zones don't match

rtitle: Yeah I can see that.. also `zone` is not in the `PERSISTENT_DISK` unique key so it wouldn't work anyway to create a new disk with the same project/name in a different zone. So erroring feels right.

How about just tweaking the error message:

```
"Existing disk ${googleProject.value}/${pd.name.value} is in zone ${pd.zone.value}, and cannot be attached to a runtime in zone ${targetZone.value}. Please create a new disk."
```

The UI might need a way to actually allow users to create a new disk if they get in this situation..

gpcarr: That's a good point - we could also add to the error message `... Please create a new disk or create your runtime in zone ${pd.zone.value} if you want to keep your old disk.`

Qi77Qi: Updated it to
```
                  s"existing disk ${pd.projectNameString} is in zone ${pd.zone.value}, and cannot be attached to a runtime in zone ${targetZone.value}. Please create a your runtime in zone ${pd.zone.value} if you'd like to use this disk; or opt to use a new disk",
```

Qi77Qi:
jenkins retest...

leonardo-build failed
```
15:25:02 java.sql.SQLTransientConnectionException: mysql.db - Connection is not available, request timed out after 5000ms.
```

rtitle (approved):
LGTM

PR #2032, opened 2021-04-09
[No ticket] Debug KubernetesClusterNotFound in dev

jdcanas:


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/NonLeoMessageSubscriber.scala
```diff
@@ -86,6 +86,7 @@ class NonLeoMessageSubscriber[F[_]: Timer](gkeAlg: GKEAlgebra[F],
     for {
       ctx <- ev.ask
       clusterId = msg.clusterId
+      _ <- logger.info(s"Beginning to delete kubernetes cluster ${clusterId} in project ${msg.project} from NonLeoMessageSubscriber | trace id: ${ctx.traceId}")
```
rtitle: We are trying to do `logger.info(ctx.loggingCtx)(msg)` now, so the trace id doesn't need to be embedded in the string.

kyuksel:
```suggestion
      _ <- logger.info(ctx.loggingCtx)(s"Beginning to delete kubernetes cluster ${clusterId} in project ${msg.project} from NonLeoMessageSubscriber")
```

kyuksel (approved):
👍🏼  with one change request inline

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/NonLeoMessageSubscriber.scala
```diff
@@ -86,6 +86,7 @@ class NonLeoMessageSubscriber[F[_]: Timer](gkeAlg: GKEAlgebra[F],
     for {
       ctx <- ev.ask
       clusterId = msg.clusterId
+      _ <- logger.info(ctx.loggingCtx)(s"Beginning to delete kubernetes cluster ${clusterId} in project ${msg.project} from NonLeoMessageSubscriber")
```
Qi77Qi: not sure if this is necessary. If a cluster is deleted from `zombieMonitor`, we should have `deletedFrom` field indicating that (if cron job is not doing that currently, we should update there)

Qi77Qi: oh nvm, cron job just sends message to leo, so I guess we can really differentiate

jdcanas:
multi-test

PR #2034, opened 2021-04-09
[no ticket] remove some extra trace id

Qi77Qi:

---

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
@@ -37,7 +37,7 @@ import org.broadinstitute.dsde.workbench.leonardo.dao.{AppDAO, AppDescriptorDAO,
 import org.broadinstitute.dsde.workbench.leonardo.db.{DbReference, kubernetesClusterQuery, nodepoolQuery, _}
 import org.broadinstitute.dsde.workbench.leonardo.http._
 import org.broadinstitute.dsde.workbench.leonardo.http.service.AppNotFoundException
-import org.broadinstitute.dsde.workbench.leonardo.model.LeoException
+import org.broadinstitute.dsde.workbench.leonardo.model.{BadRequestException, LeoException}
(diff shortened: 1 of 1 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -245,8 +246,9 @@ class GKEInterpreter[F[_]: Parallel: ContextShift: Timer](
       dbClusterOpt <- kubernetesClusterQuery.getMinimalClusterById(dbNodepool.clusterId).transaction
       dbCluster <- F.fromOption(
         dbClusterOpt,
-        KubernetesClusterNotFoundException(
-          s"Cluster with id ${dbNodepool.clusterId} not found in database | trace id: ${ctx.traceId}"
+        BadRequestException(
```
ehanna4, line 249: Isn't `KubernetesClusterNotFoundException` more accurate of an exception in this case? Could we update that sort of exception to handle the traceId the way a `BadRequestException` type does instead of switching the types here?

Qi77Qi: it is...two reasons why switching to `BadRequestException`

1. `KubernetesClusterNotFoundException` is an exception defined in wb-libs, and it doesn't have `traceId` field.
2. I don't think it's super helpful to have a unique type for various "bad requests". The message itself should be unique so that we can easily identify where the error happens.

kyuksel: The name `KubernetesClusterNotFoundException` in this context threw me off in the past because it sounds like the caller requests some info about a cluster (like in a `GET cluster` request) that doesn't exist whereas what's happening here is specific to our implementation (i.e. registering a record in Leo DB before doing anything else on the Google side). I don't think the name `BadRequestException` accurately describes the situation either. How about `CreateClusterException`?

jdcanas:
Very old

PR #2041, opened 2021-04-13
[IA-2661] Perf fixes: Increase dns cache expiry and reduce app monitoring timeouts

rtitle:
2 fixes here:

### 1. DNS cache expiry
Perf tests showed a couple app errors like:
```
MySQL [leonardo]> select * from APP_ERROR where appId in (12308, 12367);
+------+-------+--------------+----------------------------+-----------------+--------+-----------------+------------------------------------------------------+
| id   | appId | errorMessage | timestamp                  | action          | source | googleErrorCode | traceId                                              |
+------+-------+--------------+----------------------------+-----------------+--------+-----------------+------------------------------------------------------+
| 6692 | 12308 | 5 seconds    | 2021-04-12 16:34:28.051000 | createGalaxyApp | app    |            NULL | 96b6077099bd235b82ff55e19d301148/3485759192482900616 |
| 6691 | 12367 | 5 seconds    | 2021-04-12 16:34:32.139000 | createGalaxyApp | app    |            NULL | a689e094ab503f225822acf91158c319/8959465316832708319 |
+------+-------+--------------+----------------------------+-----------------+--------+-----------------+------------------------------------------------------+
```
I suspect the `5 seconds` error is likely from the `KubernetesDnsCache` guava cache [expiry](https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/resources/reference.conf#L466). I don't think there is a good reason it needs to be so low for k8s. See comment in `reference.conf`, I'm hoping increasing it to 10 minutes improves things.

### 2. App monitoring timeouts
Other perf test errors show app creation timeouts:
```
MySQL [leonardo]> select * from APP_ERROR where appId = 12329;
+------+-------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+----------------------------+-----------------+--------+-----------------+-------------------------------------------------------+
| id   | appId | errorMessage                                                                                                                                                                                                               | timestamp                  | action          | source | googleErrorCode | traceId                                               |
+------+-------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+----------------------------+-----------------+--------+-----------------+-------------------------------------------------------+
| 6695 | 12329 | Galaxy installation has failed or timed out for app leo-perf-test-4c10bac6-3857-4a33-bf9f-fd2ac1777159 in cluster projects/perf-env-test-project-992/locations/us-central1-a/clusters/k51fd49e-7280-487d-a91a-990683311c53 | 2021-04-12 16:34:29.404000 | createGalaxyApp | app    |            NULL | a0a6dd6fe533f1738a9a8a19f50070f0/10167083808395204781 |
+------+-------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+----------------------------+-----------------+--------+-----------------+-------------------------------------------------------+
```
I noticed that the timeout takes much longer than intended because we are actually retrying the `http4s` call (which itself retries) with `streamFUntilDone`. This is annoying because it makes the gatling tests time out. So I added `InterruptableMonitorConfig` and use fs2 `Stream.interruptAfter` to `createApp` and `startApp` monitoring. (Note this isn't needed for runtimes because of how `BaseCloudServiceRuntimeMonitor` works.)


---

In all cases:

### DIFF
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -465,7 +465,7 @@ runtimeDnsCache {
 }
 
 # Kubernetes expiration can be higher because the IP is at the cluster level, which is
-# consistent across app creations. Clusters are garbage collected ~1 hour after app deletion.
+# consistent across app pause/resume. Clusters are garbage collected ~1 hour after app deletion.
 kubernetesDnsCache {
   cacheExpiryTime = 10 minutes
   cacheMaxSize = 10000
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -457,8 +457,8 @@ clusterFiles {
   rstudioLicenseFile = "/etc/rstudio-license-file.lic"
 }
 
-# Expiration is low because runtimes can be recreated with the same project/name but with
-# a different IP, so we need to ensure we don't use stale cache entries.
+# The expiration is low because the IP changes when a runtime is pause/resumed,
+# so we need to ensure we don't use stale cache entries.
 runtimeDnsCache {
   cacheExpiryTime = 5 seconds
   cacheMaxSize = 10000
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -631,7 +631,8 @@ pubsub {
     }
     createApp {
       interval = 10 seconds
-      max-attempts = 60 # (10 seconds + ~20 second http4s retry) * 60 =~ 30 min
+      max-attempts = 120 # 10 seconds * 120 = 20 min
+      interruptAfter = 20 minutes
     }
     deleteApp {
       interval = 10 seconds
@@ -646,8 +647,9 @@ pubsub {
       interval = 10 seconds
     }
     startApp {
-      max-attempts = 24 # (5 seconds + ~20 second http4s retry) * 24 = ~10 min
-      interval = 5 seconds
+      max-attempts = 200 # 3 seconds * 200 is 10 min
+      interval = 3 seconds
+      interruptAfter = 10 minutes
     }
   }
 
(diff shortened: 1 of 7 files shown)


http/src/main/resources/reference.conf
```diff
@@ -627,7 +631,7 @@ pubsub {
     }
     createApp {
       interval = 10 seconds
-      max-attempts = 120 # 10 seconds * 120 = 20 min
+      max-attempts = 60 # (10 seconds + ~20 second http4s retry) * 60 =~ 30 min
```
rtitle: Thinking another option may be to call `interruptAfter(x minutes)` on the Stream 🤔

http/src/main/resources/reference.conf
```diff
@@ -642,8 +647,9 @@ pubsub {
       interval = 10 seconds
     }
     startApp {
-      max-attempts = 100 # 3 seconds * 100 is 5 min
+      max-attempts = 200 # 3 seconds * 200 is 10 min
```
rtitle, line 650: 5 min felt a little short

Qi77Qi (approved):
lgtm...wonder what the error will look like when interruption happens

rtitle:
> wonder what the error will look like when interruption happens

Yeah good question, I can check.. although we should be calling `adaptError` for errors thrown here.

rtitle:
Looks like fs2 `interruptAfter` doesn't throw an exception, it just cancels the stream.
```
scala> val x = (Stream.sleep(1 second) ++ Stream.eval(IO(println("tick")))).repeat.interruptAfter(10 seconds).compile.lastOrError.unsafeRunSync()
tick
tick
tick
tick
tick
tick
tick
tick
tick
val x: Unit = ()
```

Which is what we want I think, since we check `.isDone()` after the stream terminates, and throw an error if it's not done.

rtitle:

PR #2063, opened 2021-04-20
[IA-2675] make application concurrency configurable

rtitle:
https://broadworkbench.atlassian.net/browse/IA-2675

---

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/Boot.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/Boot.scala
@@ -451,7 +451,7 @@ object Boot extends IOApp {
         .resource(Paths.get(pathToCredentialJson), gkeService, blocker, semaphore)
       // Use a low concurrency for helm because it can generate very chatty network traffic
       // (especially for Galaxy) and cause issues at high concurrency.
-      helmConcurrency <- Resource.eval(Semaphore[F](10L))
+      helmConcurrency <- Resource.eval(Semaphore[F](20L))
       helmClient = new HelmInterpreter[F](blocker, helmConcurrency)
 
       googleDependencies = GoogleDependencies(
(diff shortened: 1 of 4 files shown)


Qi77Qi (approved):
👍

rtitle:
Perf test running now..

rtitle:
Don't think this really helped.. reverted back to previous value. Still think it's nice to make configurable.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/Boot.scala
```diff
@@ -449,7 +449,8 @@ object Boot extends IOApp {
       // Set up k8s and helm clients
       kubeService <- org.broadinstitute.dsde.workbench.google2.KubernetesService
         .resource(Paths.get(pathToCredentialJson), gkeService, blocker, semaphore)
-      helmClient = new HelmInterpreter[F](blocker, semaphore)
+      helmConcurrency <- Resource.eval(Semaphore[F](50L))
```
rtitle: Wondering if we'd get better throughput if helm had its own permits

rtitle:
Actually seeing some different results at smaller concurrency levels for helm. I'm starting to think that the Galaxy `helm install` is just really chatty and we're limited by network speed on the back Leo node.

- concurrency of 255+: all 100 apps get installed at the same time, wait time ~45 mins for all helm installs to complete
- concurrency of 50: app installs are a bit more staggered, avg wait time ~10-15 mins per helm install
- concurrency of 10: app installs even more staggered, wait time ~2 mins per helm install

So I think a lower concurrency for helm probably increases throughput overall, since it staggers the `helm install`s and allows Galaxy to start up on some apps while helm is running on others (letting more things happen in parallel).

rtitle:
100% test pass with concurrency of 10: http://34.70.158.48:8666/bvdp-daily/. The dev run last night was about 81% passing. Gonna try concurrency of 20 next.

PR #2069, opened 2021-04-21
Update naming convention of galaxy postgres disk [IA-2740]

ehanna4:
Updated the naming convention for the postgres/metadata disk for Galaxy Apps to use the datadisk name so that the Disk table in the Leo DB will have access to the info needed for resource validator to delete both disks from Google.

I tested with a fiab and Gabriela's UI changes the following successfully:
- Create a Galaxy App with develop branch then switch to my changes and successfully - Delete the App and Disks (checked Google and DB), Delete just the App and keep the Disks and then attach said Disks to a new App
- Create a fully new Galaxy App with the new styled named Disk and successfully deleted this as well.

---

In all cases:


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/MonitorAtBoot.scala
```diff
@@ -144,11 +144,13 @@ class MonitorAtBoot[F[_]: Timer](publisherQueue: fs2.concurrent.Queue[F, LeoPubs
               )
           }
           diskIdOpt = app.appResources.disk.flatMap(d => if (d.status == DiskStatus.Creating) Some(d.id) else None)
+          diskNameOpt = app.appResources.disk.flatMap(d => if (d.status == DiskStatus.Creating) Some(d.name) else None)
           msg = CreateAppMessage(
             cluster.googleProject,
             action,
             app.id,
             app.appName,
+            diskNameOpt,
```
Qi77Qi: why is disk `DiskName` needed to be passed in the message? I think `diskId` is already available from the message, from which you can get the diskName

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -1234,7 +1279,7 @@ class GKEInterpreter[F[_]: Parallel: ContextShift: Timer](
       raw"""persistence.postgres.name=${namespaceName.value}-${config.galaxyDiskConfig.postgresPersistenceName}""",
       raw"""galaxy.postgresql.galaxyDatabasePassword=${config.galaxyAppConfig.postgresPassword}""",
       raw"""persistence.postgres.persistentVolume.extraSpec.gcePersistentDisk.pdName=${getGalaxyPostgresDiskName(
-        namespaceName
+        nfsDisk.name
```
Qi77Qi: we might need to the disk name depending on which style disk actually exists in google if we're worried about users might have an older style disk....but at the same time, Gabriela's UI restore PR is not merged yet, so we probably don't really need to worry about this much

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/MonitorAtBoot.scala
ehanna4: I'm not sure i understand how to get `diskName` from `diskId` - wouldn't that require querying the DB?

Qi77Qi: yea..I think in subscriber, we're probably already querying for the disk anyway

ehanna4: alright, looking into it!

ehanna4: i think i've made the right change! please let me know what you think!

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -896,6 +941,14 @@ class GKEInterpreter[F[_]: Parallel: ContextShift: Timer](
       _ <- logger.info(ctx.loggingCtx)(
         s"Installing helm chart ${config.galaxyAppConfig.chart} for app ${appName.value} in cluster ${dbCluster.getGkeClusterId.toString}"
       )
+      postgresDiskNameOpt <- for {
+        disk <- getGalaxyPostgresDisk(nfsDisk.name, namespaceName, nfsDisk.googleProject, nfsDisk.zone)
```
rtitle, line 945: Do we need to call `getGalaxyPostgresDisk` (which makes a Google call) here? Since we just need the name could we do:
```
postgresDiskName = getGalaxyPostgresDiskName(nfsDisk.name)
```

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriberSpec.scala
```diff
@@ -1280,7 +1281,7 @@ class LeoPubsubMessageSubscriberSpec
       //The non-default nodepool should still be there, as it is not deleted on app deletion
       getCluster.nodepools.size shouldBe 2
       getCluster.nodepools.filter(_.isDefault).head.status shouldBe NodepoolStatus.Running
-      getApp.app.errors.size shouldBe 1
+//      getApp.app.errors.size shouldBe 1 // <- emily error here gets 0
```
rtitle: Noting TODO, should revert before merge (or understand why it's failing and fix)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
Qi77Qi: this is to support disks that're created in old naming convention...altho I'm not sure how important that is given restore from a disk is not available in UI yet

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -819,19 +823,60 @@ class GKEInterpreter[F[_]: Parallel: ContextShift: Timer](
...
       .setSizeGb(config.galaxyDiskConfig.postgresDiskSizeGB.gb.toString)
       .setPhysicalBlockSizeBytes(config.galaxyDiskConfig.postgresDiskBlockSize.bytes.toString)
       .putAllLabels(Map("leonardo" -> "true").asJava)
       .build()
 
-  private[leonardo] def getGalaxyPostgresDiskName(namespaceName: NamespaceName): DiskName =
+  private[leonardo] def getGalaxyPostgresDiskName(dataDiskName: DiskName): DiskName =
+    DiskName(s"${dataDiskName.value}-${config.galaxyDiskConfig.postgresDiskNameSuffix}")
+
+  private[leonardo] def getOldStyleGalaxyPostgresDiskName(namespaceName: NamespaceName): DiskName =
```
rtitle, line 839: Maybe should add a TODO to delete the `oldStyle` method once there are no legacy Galaxy disks in the system?

Qi77Qi: yea...I think that'd be a good ticket to do

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -894,13 +899,9 @@ class LeoPubsubMessageSubscriber[F[_]: Timer: ContextShift: Parallel](
       task = for {
```
rtitle, line 899: Out of scope for this PR, but I recall thinking this method could be cleaned up

Qi77Qi (approved):
lgtm

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
rtitle: Right, I see… yeah technially disk support is not in the UI yet but probably safe to have this here I suppose.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
rtitle: Ticket sounds good 👍

rtitle (approved):
Few comments above but overall looks good. The testing performed sounds good too. 👍

It would be good to have a follow-up ticket to remove support for legacy disk names, and also cleaning up some of the disk-related code (which existed before this PR).

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
ehanna4: yeah I wasn't sure if my change or Gabriela's was going to get out first so figured better safe than sorry lol and will protect against any folks who maybe made a PD and stopped using it awhile will be able to start up again no problem once those UI changes do go out.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
ehanna4: Should I make a note in the code like a TODO or a ticket?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
ehanna4: I'll make a ticket👍

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
rtitle: Eh, not sure what concretely to suggest so don't worry about it :)

PR #2088, opened 2021-04-30
IA-2637 Use cos custom image and enable gpu

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-2637

* Switches to use disk snapshots
* Add GPU support in back leo..but front leo's message always passes `None` for `gpuConfig` to back leo as of this PR.

Justin's PR will update messages properly to reflect user input, and I'll add automation tests once his PR is merged.

Validate that GPUs can be access from within jupyter:
Screen Shot 2021-05-10 at 2 57 52 PM

---

In all cases:


core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/gceModels.scala
```diff
@@ -18,3 +19,29 @@ object GceInstanceStatus extends Enum[GceInstanceStatus] {
   case object Suspended extends GceInstanceStatus
   case object Terminated extends GceInstanceStatus
 }
+
+sealed abstract class GpuType extends Product with Serializable {
+  def asString: String
+}
+object GpuType {
```
rtitle, line 26: Is this from https://cloud.google.com/compute/docs/gpus?

`accelerator-optimized` machine type sounds interesting, maybe could be a future enhancement

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodec.scala
```diff
@@ -343,6 +349,13 @@ object JsonCodec {
     Decoder.decodeString.emap(x => DiskStatus.withNameOption(x).toRight(s"Invalid disk status: $x"))
   implicit val diskTypeDecoder: Decoder[DiskType] =
     Decoder.decodeString.emap(x => DiskType.stringToObject.get(x).toRight(s"Invalid disk type: $x"))
+  implicit val gpuTypeDecoder: Decoder[GpuType] =
```
rtitle, line 352: Reminder to update swagger :)

Also maybe we could name it `gpuType` and `gpuCount`? Looks like GCP names it `acceleratorType` and `acceleratorCount`.

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/gceModels.scala
Qi77Qi: yea....so the `accelerator-optimized` are machine types, which we can consider adding in terra-ui in the future...I'm thinking we won't offer those in the beginning

http/src/main/resources/init-resources/cloud-init.yml
```diff
@@ -0,0 +1,6 @@
+#cloud-config
+
+bootcmd:
+- fsck.ext4 -tvy /dev/sdb
+- mkdir -p /mnt/disks/work
+- mount -t ext4 -O ... /dev/sdb /mnt/disks/work
```
rtitle: Is this `…` supposed to be there?

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodec.scala
Qi77Qi: @jdcanas 's front leo PR will update API doc...I think `gpuType` is better since it's widely known, and I remember reading somewhere that seems like there's another type of accelerator (non gpu), altho I can't find the doc at the moment

http/src/main/resources/init-resources/cluster-site-gce.conf
```diff
@@ -0,0 +1,77 @@
+<VirtualHost _default_:${SSL_HTTPD_PORT}>
```
rtitle, line 1: I guess we need a separate version of this file because of the docker networking change, right?

http/src/main/resources/init-resources/cloud-init.yml
Qi77Qi: yea...I'm not sure exactly what it is...but it's from [this doc](https://cloud.google.com/container-optimized-os/docs/concepts/disks-and-filesystem#mounting_and_formatting_disks)

http/src/main/resources/init-resources/gce-init.sh
```diff
@@ -188,116 +192,73 @@ USER_SCRIPT_OUTPUT_URI=$(userScriptOutputUri)
...
+mkdir -p ${WORK_DIRECTORY}
+mkdir -p ${CERT_DIRECTORY}
+mkdir -p ${DOCKER_COMPOSE_FILES_DIRECTORY}
 
 log 'Formatting and mounting persistent disk...'
 
 # Format and mount persistent disk
-export DISK_DEVICE_ID=$(lsblk -o name,serial | grep 'user-disk' | awk '{print $1}')
-# Only format disk is it hasn't already been formatted
+# Fix this to `sdb`. We've never seen a device name that's not `sdb`,
```
rtitle, line 209: Fair enough :)

http/src/main/resources/init-resources/cluster-site-gce.conf
Qi77Qi: yep..the hostnames are different...

http/src/main/resources/init-resources/gce-init.sh
Qi77Qi: more important reason is becuz the original cmd doesn't work anymore :) we can improve this if we realize there's a need

http/src/main/resources/init-resources/gce-init.sh
```diff
@@ -188,116 +192,73 @@ USER_SCRIPT_OUTPUT_URI=$(userScriptOutputUri)
...
-gsutil cp ${ROOT_CA} /certs
-gsutil cp ${PROXY_SITE_CONF} /etc
-gsutil cp ${JUPYTER_DOCKER_COMPOSE_GCE} /etc
-gsutil cp ${RSTUDIO_DOCKER_COMPOSE} /etc
-gsutil cp ${PROXY_DOCKER_COMPOSE} /etc
-gsutil cp ${WELDER_DOCKER_COMPOSE} /etc
-gsutil cp ${CRYPTO_DETECTOR_DOCKER_COMPOSE} /etc
-
-echo "" > /etc/google_application_credentials.env
+$GSUTIL_CMD cp ${SERVER_CRT} ${CERT_DIRECTORY}
```
rtitle, line 233: (Minor) since all these files are in the same dir in the bucket, I wonder if we could do something like:
```
gsutil cp gs://init-bucket/* /var/
```

That way we wouldn't have to spin up the `gsutil` container multiple times, which might be faster.

I don't feel too strongly to do it in this PR, just an idea.

http/src/main/resources/init-resources/crypto-detector-docker-compose.yaml
```diff
@@ -1,7 +1,6 @@
-version: '2'
+version: '2.4'
 services:
   crypto-detector:
     container_name: "${CRYPTO_DETECTOR_SERVER_NAME}"
     image: "${CRYPTO_DETECTOR_DOCKER_IMAGE}"
-    network_mode: host
```
rtitle, line 6: Does this need to reference `app_network` somehow?

http/src/main/resources/init-resources/jupyter-docker-compose-gce.yaml
```diff
@@ -2,19 +2,23 @@
...
   jupyter:
     container_name: "${JUPYTER_SERVER_NAME}"
     image: "${JUPYTER_DOCKER_IMAGE}"
     # Override entrypoint with a placeholder to keep the container running indefinitely.
     # The cluster init script will run some scripts as root and then start pyspark as
     # jupyter-user via docker exec.
-    entrypoint: "tail -f /dev/null"
-    network_mode: host
+    # -F will follow the log when the log is created.
+    entrypoint: "tail -F /home/jupyter/notebooks/jupyter.log"
```
rtitle: Ahh.. nice this is better

http/src/main/resources/init-resources/jupyter-docker-compose.yaml
```diff
@@ -2,19 +2,19 @@
 # configuring memory options in container mode. See discussion in:
```
rtitle, line 2: Wonder if we should name this file `jupyter-docker-compose-dataproc.yaml`

http/src/main/resources/init-resources/crypto-detector-docker-compose.yaml
Qi77Qi: I added `network.yaml` and `network_gce.yaml` because otherwise I'll have to have 2 copies of each docker-compose file

http/src/main/resources/init-resources/gce-init.sh
```diff
@@ -306,48 +267,76 @@ log 'Starting up the Jupyter...'
...
+END
 
-# If Welder is installed, start the service.
-# See https://broadworkbench.atlassian.net/browse/IA-1026
-if [ ! -z "$WELDER_DOCKER_IMAGE" ] && [ "$WELDER_ENABLED" == "true" ] ; then
-  log 'Starting Welder (file synchronization service)...'
-  retry 3 docker exec -d ${WELDER_SERVER_NAME} /opt/docker/bin/entrypoint.sh
-fi
+# Create a network that allows containers to talk to each other via exposed ports
+docker network create -d bridge app_network
```
rtitle, line 318: Wonder if it's possible to create the network in `network-compose.yaml`? (thinking out loud, not a big deal either way)

http/src/main/resources/init-resources/rstudio-docker-compose.yaml
```diff
@@ -20,11 +19,9 @@ services:
...
     volumes:
-      - /work:/home/rstudio
-      - /etc/rstudio-license-file.lic:/etc/rstudio-server/license.lic:ro
+      - ${WORK_DIRECTORY}:/home/rstudio
     env_file:
-      - /etc/google_application_credentials.env
-      - /etc/custom_env_vars.env
+      - /var/custom_env_vars.env
     # See https://docs.docker.com/engine/reference/run/#user-memory-constraints
     mem_limit: ${MEM_LIMIT}   # hard limit on memory consumption by the container
```
rtitle, line 27: Need to add
```
networks:
  app_network:
    external: true
```
?

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodec.scala
rtitle: yeah the `gpu` naming sounds fine.. mainly was thinking `s/numOfGpus/gpuCount`

rtitle:
Pausing for now, will take another pass (so much bash/yaml). Overall looks great, lots of nice improvements.

http/src/main/resources/init-resources/jupyter-docker-compose.yaml
Qi77Qi: there's so much yml, bash changes in this PR and it's a bit time consuming to debug...so I'm inclined to save this renaming to another time :)

http/src/main/resources/init-resources/gce-init.sh
Qi77Qi: the network needs to exist before other services can reference it...but `network.yaml` needs to be appended last because it references other services..so I think we'll have to create it beforehand.

http/src/main/resources/init-resources/rstudio-docker-compose.yaml
Qi77Qi: I think it's not needed since `docker -f xxx -f xxx ... config` will merge all the config files, and this block is already added in `network.yaml`

Qi77Qi:
```
[info] org.broadinstitute.dsde.workbench.leonardo.LeonardoTerraDockerSuite *** ABORTED *** (26 milliseconds)
[info]   java.lang.ExceptionInInitializerError
```

jenkins retest

kyuksel:
@Qi77Qi Were you planning on updating GCE image hash in reference.conf to point to the COS-based image as part of this PR?

**Edit**
Never mind. I see the change is already there 😊

Qi77Qi:
@kyuksel that's part of the PR https://github.com/DataBiosphere/leonardo/pull/2088/files#diff-9024d8d3f171d065fa3b61dbc7dc2c56bdc6587644fde470722fdc92e7068f61R88

Qi77Qi:
multi-test

Qi77Qi:
leonardo-build failed in one of the swatomation tests..the rest all passed

multi-test

Qi77Qi:
multi-test

Qi77Qi:
multi-test

http/src/main/resources/init-resources/cloud-init.yml
rtitle: I feel like that's a placeholder for a `mount` option list…

> Refer to the Linux documentation for a full set of options supported by the fsck.ext4 and mount commands.

https://man7.org/linux/man-pages/man8/mount.8.html

Previously we used `-o discard,defaults`: https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/resources/init-resources/gce-init.sh#L203

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodec.scala
rtitle: still think `gpuCount` sounds cleaner :)

Qi77Qi: just saw this...I think `numOfGpus` is a bit more consistent with existing code, since we have `numOfxxx` and `numberOfxx` in most places

http/src/main/resources/init-resources/jupyter-docker-compose-gce.yaml
```diff
@@ -29,7 +33,22 @@ services:
...
+      - /var/custom_env_vars.env
     # See https://docs.docker.com/engine/reference/run/#user-memory-constraints
     mem_limit: ${MEM_LIMIT}   # hard limit on memory consumption by the container
+  welder:
+    ports:
+      - "8080:8080"
+    networks:
+      - app_network
+  crypto-detector:
+    network_mode: "service:jupyter"
```
rtitle, line 46: TIL about `network_mode: service:xxx`. I guess this means it shares the network stack with another container?

http/src/main/resources/init-resources/rstudio-docker-compose.yaml
rtitle: Gotcha.. also missed that we now have `rstudio-docker-compose-gce.yaml`

http/src/main/resources/init-resources/welder-docker-compose.yaml
```diff
@@ -11,8 +11,7 @@ services:
     # The cluster init script will start welder via docker exec,
     # when Application Default Credentials are available.
     # See https://broadworkbench.atlassian.net/browse/IA-1026
-    entrypoint: "tail -f /dev/null"
```
rtitle, line 14: Do we not need to override the welder entrypoint anymore, and start it with `docker exec`? I can't recall why exactly we did that..

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/BucketHelper.scala
```diff
@@ -142,13 +145,44 @@ class BucketHelper[F[_]: Concurrent: ContextShift: Parallel](config: BucketHelpe
           )).compile.drain
         }
 
+    val uploadGpuDockerCompose = gpuConfig.traverse { gC =>
+      val additionalGpuConfigString = (1 to gC.numOfGpus)
```
rtitle, line 149: Thought about using `circe-yaml` to modify the docker-compose file but I guess this is fine 🙈

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodec.scala
rtitle: ok, fair enough 👍

rtitle:
Lot of bash/yaml, but this seems close! I was actually thinking of doing a little manual testing of this branch, to get familiar with the compose files/networking.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/BucketHelper.scala
Qi77Qi: ah good to know....maybe worth using `circe-yaml` if the use case becomes more complicated

http/src/main/resources/init-resources/welder-docker-compose.yaml
Qi77Qi: the comment above seems to suggest the original reason is to wait until application default credential becomes available. But I think it works just fine with the default entrypoint from what I can tell

rtitle: I see.. well if it works fine this way then :shipit:

http/src/main/resources/init-resources/jupyter-docker-compose-gce.yaml
Qi77Qi: yea, this is for the cryto-detector container to share network with the app container

http/src/main/resources/init-resources/cloud-init.yml
Qi77Qi: I'll change it to `discard.defaults` and see

http/src/main/resources/init-resources/jupyter-docker-compose-gce.yaml
rtitle: Could this be:
```
entrypoint: "tail -F ${NOTEBOOKS_DIR}/jupyter.log"
```
?

That way it would still work on old and new home directories. (Right now it says "file does not exist" on current runtimes.)

Qi77Qi: oh yea...definitely...thanks for the catch!

rtitle:
Tested a bunch of things and mostly all looks good. However I found 1 bug I think. When I stop/start an RStudio runtime (or delete/recreate with the same disk) I get this error:

image

This has happened in the past, and the fix is to run [shutdown.sh](https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/resources/init-resources/shutdown.sh) as a GCE [shutdown script](https://cloud.google.com/compute/docs/shutdownscript) to cleanly shut down `rstudio-server` before the instance stops. I verified it still works in prod, so something changed in this PR. Will think more..

Qi77Qi:
@rtitle ah thanks! will look into that

http/src/main/resources/init-resources/cloud-init.yml
```diff
@@ -0,0 +1,17 @@
+#cloud-config
+
+bootcmd:
+- fsck.ext4 -tvy /dev/sdb
+- mkdir -p /mnt/disks/work
+- mount -t ext4 -O discard,defaults /dev/sdb /mnt/disks/work
+
+write_files:
```
rtitle, line 8: I am not sure if this is really needed but maybe we should do the same thing for `startup` so the startup script is guaranteed to run after docker?
```
- path: /etc/systemd/system/google-startup-scripts.service.d/override.conf 
  permissions: 0644
  owner: root    
  content:
    [Unit]      
    After=docker.service
```

rtitle (approved):
:shipit:

http/src/main/resources/init-resources/cloud-init.yml
Qi77Qi: this file exists across vm restarts..so I think we're good :)

Qi77Qi:
jenkins retest

Qi77Qi:
going to merge this as is....but [this](https://github.com/DataBiosphere/leonardo/blob/develop/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/runtimes/RuntimeCreationDiskSpec.scala#L175) failed twice among all automation test runs (including multi test runs)...not entirely sure if it's an existing issue...but think we can comment out this particular validation if it turns out to be flaky....

PR #2108, opened 2021-05-10
[IA-2633] COS-based snapshots for GCE VMs

kyuksel:
**Note for PR reviewers:** Please consider adding `?w=1` to the end of the GitHub URL (or click [here](https://github.com/DataBiosphere/leonardo/pull/2108/files?w=1)) to not be distracted by white space changes due to pipeline stage indentation fixes in `Jenkinsfile`.

[JIRA ticket](https://broadworkbench.atlassian.net/browse/IA-2633)

For the reasons described on the aforementioned ticket, this PR enables automatic generation of snapshots based on container-optimized and Google-maintained OS images. 

- Thanks to COS images' supporting several tools we previously manually installed plus the enhanced security that comes out-of-the-box (and hence rendering our custom image hardening redundant), the new `prepare…` script is reduced to caching docker images only 😊 
- The new Jenkins job created off of `Jenkinsfile` updated in this PR is [leonardo-gce-snapshot-and-dataproc-image](https://fc-jenkins.dsp-techops.broadinstitute.org/job/leonardo-gce-snapshot-and-dataproc-image)
- We're keeping the GCE custom image build stuff and the corresponding (`leonardo-custom-image`) Jenkins job in for the time being. We will remove those after we incorporate the GPU epic's image stack and feel confident we won't have to create another custom GCE image.
- We continue leveraging Daisy to further automate the snapshot creation step.

**TODOs before merging:**
- [x] Rename the Jenkins job as `leonardo-gce-snapshot-and-dataproc-image`.
- [x] Revert the local development related pointers in Jenkinsfile before merging.


jenkins/Jenkinsfile
```diff
@@ -39,7 +47,7 @@ boolean hasHashChanges(Set<String> images, String[] configPaths) {
     for (String configFile: configPaths) {
         println("Checking config file '$configFile' for any hash changes...")
         def gitOutput = sh(
-            script: "git diff HEAD~1..HEAD $configFile",
+            script: "git diff --unified=0 HEAD~1..HEAD $configFile",
```
kyuksel, line 52: Without the option above, `git diff` outputs a few of the previous and subsequent lines as context to the lines that changed, which can mess up the change detection logic in `Jenkinsfile`. Not sure how this hadn't caused issues.

jenkins/Jenkinsfile
```diff
@@ -126,45 +134,48 @@ pipeline {
...
-        string(name: "terra_jupyter_aou", defaultValue: "1.0.14")
-        string(name: "anvil_rstudio_bioconductor", defaultValue: "0.0.8")
-        string(name: "welder_server", defaultValue: "59be71a",
+        string(name: "terra_jupyter_base", defaultValue: "0.0.19")
+        string(name: "terra_jupyter_hail", defaultValue: "0.1.2")
+        string(name: "terra_jupyter_r", defaultValue: "1.0.13")
+        string(name: "terra_jupyter_python", defaultValue: "0.1.1")
+        string(name: "terra_jupyter_bioconductor", defaultValue: "1.0.13")
+        string(name: "terra_jupyter_gatk", defaultValue: "1.1.1")
+        string(name: "terra_jupyter_aou", defaultValue: "1.0.23")
```
kyuksel: Didn't pick the latest possible for `terra-jupyter-aou` as per a [Slack thread](https://terra-bio.slack.com/archives/C9C4KKU65/p1619034101059900) with @calbach.

jenkins/Jenkinsfile
```diff
@@ -190,343 +203,336 @@ pipeline {
...
-                    if (shouldUpdate) {
-                        for (String image: images) {
-                            // The version could be empty if the image from terra-docker is not in a leo conf yet.
-                            if (image in versionMap.keySet() && !versionMap[image].isEmpty()) {
-                                version = versionMap[image]
+        stage ('Detect Hash Changes') {
+            when { expression { USE_CUSTOM_DOCKER_IMAGE_VERSIONS == "false" } }
+                steps {
+                    script {
+                        // TODO: Don't forget to revert before merging
```
Qi77Qi: just point out the todo

Qi77Qi (commented):
lgtm..there're a few TODOs need to be fixed before this can be merged tho...also I think the prepare custom gce image script can be removed soon

jenkins/Jenkinsfile
rtitle: These are just the default params for the Jenkins job, and IMO don't add much value:

1. if `USE_CUSTOM_DOCKER_IMAGE_VERSIONS==false` (as is default), these values aren't used at all
2. if `USE_CUSTOM_DOCKER_IMAGE_VERSIONS==true`, then we are usually explicitly populating the versions anyway, and should not be relying on the Jenkins defaults

So I'm wondering if we should just remove all these `defaultValue`s?

Qi77Qi: @rtitle  do you mean remove the defaults and always read the versions from automation's reference.conf if no version specified?

rtitle: Yeah that's what it does now if `USE_CUSTOM_DOCKER_IMAGE_VERSIONS==false` in this job (it reads from automation reference.conf). These fields are only used if we need to manually explicitly set the versions, but the Jenkinsfile defaults are in practice always out of date.

http/src/main/resources/reference.conf
```diff
@@ -86,6 +86,7 @@ dataproc {
 
 gce {
   customGceImage = "projects/broad-dsp-gcr-public/global/images/custom-leo-image-gce-debian9-5e2206c"
+  customGceSnapshot = "projects/broad-dsp-gcr-public/global/snapshots/custom-leo-gce-cos-snapshot-d1a330f"
```
kyuksel, line 89: FYI @Qi77Qi that the snapshot referenced above is the first Jenkins-created one. I've verified it has what we want for the stack downstream :)

jenkins/Jenkinsfile
Qi77Qi: I was wondering if we can make it so that if there's explicit versions specified, they can still get prioritized over automation's referece.conf (basically automation's reference.conf is the "defaults")....

Either way, I think removing these manual defaults is a good idea

jenkins/Jenkinsfile
```diff
@@ -1,15 +1,25 @@
-// Update the custom data proc image used in leo.  The name of the job contains the name of the dataproc image created. This job will fail if master of terra-docker has not been updated since the last run, as the git hash has not changed. This is a happy side-effect of the fact image names must be unique.
```
rtitle, line 1: Is this `Jenkinsfile` intended to support GCE images _and_ snapshots or just snapshots? 

If it just supports snapshots then maybe it should be a separate `Jenkinsfile` so it does not break the existing https://fc-jenkins.dsp-techops.broadinstitute.org/job/leonardo-custom-image/ job (which reads the `Jenkinsfile` from `develop` branch I think).

jenkins/Jenkinsfile
rtitle: I see, some merging of the 2 modes.. I agree, that may be useful. As is today, it's either or: use all reference.conf versions, or all explicit versions.

jenkins/dataproc-custom-images/prepare-custom-leonardo-jupyter-dataproc-image.sh
```diff
@@ -17,6 +17,7 @@ set -e -x
 #
 
 # The versions below don't matter; they are replaced by the Jenkins job
+# TODO Would be nice to read the image URLs below from automation/src/test/resources/reference.conf
```
rtitle: This comment is a bit misleading because it already does this if "use explicit versions" is unchecked in the Jenkins job: it reads from reference.conf and replaces these strings with `sed`. 

@Qi77Qi's comment above would be a nice future enhancement: to all _specific_ image overrides, while using `reference.conf` as a default. But that's out of scope here, and I don't think it warrants a TODO.

jenkins/gce-custom-images/prepare_custom_leonardo_gce_image.sh
```diff
@@ -20,15 +20,15 @@ ansible_version="2.7.0.0"
...
-terra_jupyter_bioconductor="us.gcr.io/broad-dsp-gcr-public/terra-jupyter-bioconductor:0.0.12"
-terra_jupyter_gatk="us.gcr.io/broad-dsp-gcr-public/terra-jupyter-gatk:0.0.13"
+terra_jupyter_base="us.gcr.io/broad-dsp-gcr-public/terra-jupyter-base:0.0.19"
+terra_jupyter_python="us.gcr.io/broad-dsp-gcr-public/terra-jupyter-python:0.1.1"
+terra_jupyter_r="us.gcr.io/broad-dsp-gcr-public/terra-jupyter-r:1.0.13"
+terra_jupyter_bioconductor="us.gcr.io/broad-dsp-gcr-public/terra-jupyter-bioconductor:1.0.13"
+terra_jupyter_gatk="us.gcr.io/broad-dsp-gcr-public/terra-jupyter-gatk:1.1.1"
 terra_jupyter_aou_old="us.gcr.io/broad-dsp-gcr-public/terra-jupyter-aou:1.0.23"
-terra_jupyter_aou="us.gcr.io/broad-dsp-gcr-public/terra-jupyter-aou:1.1.2"
+terra_jupyter_aou="us.gcr.io/broad-dsp-gcr-public/terra-jupyter-aou:1.0.23"
```
rtitle: I think `terra_jupyter_aou` should be `1.1.2` and `terra_jupyter_aou_old` should be `1.0.23`.

jenkins/gce-custom-images/prepare_gce_snapshot.sh
```diff
@@ -15,6 +17,7 @@ set -e -x
 #
 
 # The versions below don't matter; they are replaced by the Jenkins job
+# TODO Would be nice to read the image URLs below from automation/src/test/resources/reference.conf
```
rtitle: Same as above re this comment being misleading

jenkins/gce-custom-images/prepare_gce_snapshot.sh
```diff
@@ -15,6 +17,7 @@ set -e -x
...
 terra_jupyter_base="us.gcr.io/broad-dsp-gcr-public/terra-jupyter-base:0.0.19"
 terra_jupyter_python="us.gcr.io/broad-dsp-gcr-public/terra-jupyter-python:0.1.1"
 terra_jupyter_r="us.gcr.io/broad-dsp-gcr-public/terra-jupyter-r:1.0.13"
```
rtitle: Also same as above re the aou image versions

rtitle (changes requested):
Looks good, and cool to see it work at sprint review! 🎉 

Just a few nitpicky comments above

jenkins/dataproc-custom-images/prepare-custom-leonardo-jupyter-dataproc-image.sh
kyuksel: Thanks. Yea, I realize I'd missed the fact that automation `reference.conf` is already where the versions are extracted from and used when manual overrides aren't set.

jenkins/gce-custom-images/prepare_custom_leonardo_gce_image.sh
kyuksel: thanks for the catch

jenkins/Jenkinsfile
kyuksel: This `Jenkinsfile` is intended to eventually support both GCE snapshot and custom Dataproc image generation. Since we'll be able to cut over to snapshots for GCE pretty soon, I was hoping to not have two `Jenkinsfile`s. We could still run the previously existing Jenkins job (`leonardo-custom-image`) to generate any GCE custom images in case something comes up in the meantime. Once we cut over, we can just remove that Jenkins job.

jenkins/Jenkinsfile
kyuksel: That's all good thinking. I've updated the PR to only remove the defaults for now.

jenkins/Jenkinsfile
rtitle: Makes sense.. however I think the existing Jenkins job (`leonardo-custom-image`) reads the `Jenkinsfile` in Leonardo `develop` branch at job launch time, so it might break once this PR is merged.

I agree about ideally not having 2 `Jenkinsfiles` checked in though, as we'll not be in this state for very long.

Perhaps a simple way to solve this would be to have the `leonardo-custom-image` job not look at Leo `develop` branch anymore (we could configure it with a different branch/hash).

kyuksel: Ah you're right, @rtitle. Thanks for pointing that out! 

I've created `branch-for-legacy-leonardo-custom-image-jenkins-build` off of current `dev`. The [old Jenkins job](https://fc-jenkins.dsp-techops.broadinstitute.org/job/leonardo-custom-image/configure) now points to it. 

This way hopefully we'll be able to just remove that job altogether when ready without having to do anything else.

http/src/main/resources/reference.conf
Qi77Qi: I wonder if it'd be better to just use date as the suffix for the snapshot just so it's super obvious how old the snapshot is...also for dataproc...just so we can easily notice when they need to be updated

kyuksel: I could see the benefit but given how easy it is to see the creation time on GCP and how much time I've already spent on this ticket, I'm inclined to not add that change unless you're feeling strongly.

image

Qi77Qi: I think the benefit is if someone see this in config, they're probably not gonna open console and check creation time, vs if the creation time is there in the name, then we're more likely to notice if the image is super old.

Anyways, up to you...

rtitle (approved):
🚀

Qi77Qi (approved):

PR #2112, opened 2021-05-10
IA-2603: Adding Retryable tags to certain tests

xuscrbw:
Adding retryable tag to allow for a retry in case one of these flaky tests fails. 

I think I might need to intentionally fail a test to make sure it works properly

---

In all cases:

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/apps/AppCreationSpec.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/apps/AppCreationSpec.scala
@@ -8,7 +8,6 @@ import org.broadinstitute.dsde.workbench.leonardo.http.{ListAppResponse, Persist
 import org.broadinstitute.dsde.workbench.service.util.Tags
 import org.http4s.headers.Authorization
 import org.http4s.{AuthScheme, Credentials}
-import org.scalatest.tagobjects.Retryable
 import org.scalatest.{DoNotDiscover, ParallelTestExecution}
 
 import scala.concurrent.duration._
@@ -18,7 +17,7 @@ class AppCreationSpec extends GPAllocFixtureSpec with LeonardoTestUtils with GPA
   implicit val auth: Authorization =
     Authorization(Credentials.Token(AuthScheme.Bearer, ronCreds.makeAuthToken().value))
 
-  "create, delete an app and re-create an app with same disk" taggedAs (Tags.SmokeTest, Retryable) in { _ =>
+  "create, delete an app and re-create an app with same disk" taggedAs Tags.SmokeTest in { _ =>
     withNewProject { googleProject =>
       val appName = randomAppName
       val restoreAppName = AppName(s"restore-${appName.value}")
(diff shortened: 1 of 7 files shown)


automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/apps/AppCreationSpec.scala
```diff
@@ -17,7 +18,7 @@ class AppCreationSpec extends GPAllocFixtureSpec with LeonardoTestUtils with GPA
   implicit val auth: Authorization =
     Authorization(Credentials.Token(AuthScheme.Bearer, ronCreds.makeAuthToken().value))
 
-  "create, delete an app and re-create an app with same disk" taggedAs Tags.SmokeTest in { _ =>
+  "create, delete an app and re-create an app with same disk" taggedAs (Tags.SmokeTest, Retryable) in { _ =>
```
rtitle, line 27: Do we need to also add a `withRetry {}` block somewhere?

See: https://www.scalatest.org/scaladoc/3.2.0/org/scalatest/tagobjects/Retryable$.html

(I haven't used this Scalatest feature before, but that seems to be the case.)

xuscrbw: I think this block exists in two places already:
RuntimeFixtureSpec: in the withFixture function
and 
LeonardoSuite within the GPAllocFixtureSpec Trait.

Not sure if this is enough, but I will try testing this out on a non-selenium test to see if it is enough

rtitle: Oh yeah! Cool

Yeah maybe force a failure and see if it gets retried :)

xuscrbw:
jenkins retest

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/apps/AppCreationSpec.scala
xuscrbw: Tested this out. Had to overload the withFixture function, but I think I was able to get this to work.

Is that safe to do?

rtitle: Weird.. we have this in the superclass:
```
 override def withFixture(test: OneArgTest): Outcome = …
```

I wonder if the `withRetry` call needs to be with `NoArgTest`? ScalaTest is so mysterious sometimes.

I would say this seems fine if it works..

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookGCEDataSyncingSpec.scala
```diff
@@ -21,6 +22,12 @@ import scala.concurrent.duration._
 class NotebookGCEDataSyncingSpec extends RuntimeFixtureSpec with NotebookTestUtils {
   override def enableWelder: Boolean = true
 
+  override def withFixture(test: NoArgTest) =
```
rtitle: Maybe this could be moved to `RuntimeFixtureSpec` to avoid duplication (not all tests extend `RuntimeFixtureSpec`, but many do)

rtitle (approved):
Couple comments, but looks good to me if it works 👍

PR #2128, opened 2021-05-17
Don't throw error if trying to stop an already stopping runtime [IA-2772]

ehanna4:
Update Leo to not throw an error if a user attempts to stop an already stopping runtime. This can be reproduced by having two windows or tabs open on the same runtime.

---

In all cases:


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeServiceInterp.scala
```diff
@@ -403,7 +403,8 @@ class RuntimeServiceInterp[F[_]: Parallel](config: RuntimeServiceConfig,
 
       _ <- if (hasStopPermission) F.unit else F.raiseError[Unit](ForbiddenError(userInfo.userEmail))
       // throw 409 if the cluster is not stoppable
-      _ <- if (runtime.status.isStoppable) F.unit
+      alreadyStopping = (runtime.status == RuntimeStatus.Stopping || runtime.status == RuntimeStatus.PreStopping)
```
Qi77Qi: shall we also add `RuntimeStatus.Stopped` as well

Qi77Qi (approved):
👍

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeServiceInterp.scala
```diff
@@ -403,7 +403,7 @@ class RuntimeServiceInterp[F[_]: Parallel](config: RuntimeServiceConfig,
 
       _ <- if (hasStopPermission) F.unit else F.raiseError[Unit](ForbiddenError(userInfo.userEmail))
       // throw 409 if the cluster is not stoppable
-      _ <- if (runtime.status.isStoppable) F.unit
+      _ <- if (runtime.status.isStoppable || runtime.status.isStopping) F.unit
```
kyuksel: It looks like if `status` is one of `stoppingStatuses`, this would go on to publish a `StopRuntimeMessage`. I'd think what we want is then to neither throw an error nor take any actions since there's no need to attempt to stop a runtime that is/will be stopped. Am I missing something?

Qi77Qi: oh good point..maybe we should take out `PreStopping, Stopping`..It shouldn't hurt to republish a `StopRuntime` message, but if for some reason back leo drops the message, republishing the message will fix it

kyuksel: Not sure if I understand completely. 

1. To clarify the desired behaviour…when a user attempts to stop compute on the UI that is stopped/stopping/prestopping, we want Leo to return success, no?
2. If we have a problem of a back Leo dropping PubSub messages, wouldn't it warrant a more holistic solution than a one-off solution for a very particular case?

ehanna4: hmm okay I think I understand what you're saying. I've moved the check to wrap the enqueue at the end so if it's already stopping it won't enqueue another message - testing it out now but is that more what you're thinking?

Qi77Qi: > To clarify the desired behaviour…when a user attempts to stop compute on the UI that is stopped/stopping/prestopping, we want Leo to return success, no?
 
yes

> If we have a problem of a back Leo dropping PubSub messages, wouldn't it warrant a more holistic solution than a one-off solution for a very particular case?

Not sure what more holistic solution you're thinking..But the point is more of in back leo, if it gets a second 
`StopRuntime` message, it should be able to deal with it gracefully.

Qi77Qi (commented):
would be nice to add a unit test btw

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeServiceInterp.scala
kyuksel: > hmm okay I think I understand what you're saying. I've moved the check to wrap the enqueue at the end so if it's already stopping it won't enqueue another message - testing it out now but is that more what you're thinking?

exactly :)

> Not sure what more holistic solution you're thinking..But the point is more of in back leo, if it gets a second
StopRuntime message, it should be able to deal with it gracefully

I was trying to say that I don't think we should publish a message redundantly :)

kyuksel (approved):
:shipit:

Qi77Qi:
> I was trying to say that I don't think we should publish a message redundantly :)

haha, I was trying to say exactly the opposite that we should in this case re-send the message, and we should make back leo handle these messages properly (this should be a requirement for back leo)...altho it's true that I don't know for sure if leo currently handles this properly

kyuksel:
> we should make back leo handle these messages properly (this should be a requirement for back leo)

I agree with that one 😊

ehanna4:
@Qi77Qi I know you approved before the most recent changes so I want to double check if you are okay with the current approach?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeServiceInterp.scala
```diff
@@ -403,12 +403,17 @@ class RuntimeServiceInterp[F[_]: Parallel](config: RuntimeServiceConfig,
...
       else
         F.raiseError[Unit](RuntimeCannotBeStoppedException(runtime.googleProject, runtime.runtimeName, runtime.status))
-      // stop the runtime
-      _ <- clusterQuery.updateClusterStatus(runtime.id, RuntimeStatus.PreStopping, ctx.now).transaction
-      _ <- publisherQueue.enqueue1(StopRuntimeMessage(runtime.id, Some(ctx.traceId)))
+
+      _ <- if (runtime.status.isStopping) F.unit
+      else {
+        // stop the runtime
+        clusterQuery.updateClusterStatus(runtime.id, RuntimeStatus.PreStopping, ctx.now).transaction
```
rtitle: This should be in a for-comp.. as is now, these are 2 un-chained `IO`s so will never run

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeServiceInterp.scala
```diff
@@ -403,12 +403,17 @@ class RuntimeServiceInterp[F[_]: Parallel](config: RuntimeServiceConfig,
...
       // throw 409 if the cluster is not stoppable
-      _ <- if (runtime.status.isStoppable) F.unit
+      _ <- if (runtime.status.isStoppable || runtime.status.isStopping) F.unit
       else
         F.raiseError[Unit](RuntimeCannotBeStoppedException(runtime.googleProject, runtime.runtimeName, runtime.status))
-      // stop the runtime
-      _ <- clusterQuery.updateClusterStatus(runtime.id, RuntimeStatus.PreStopping, ctx.now).transaction
-      _ <- publisherQueue.enqueue1(StopRuntimeMessage(runtime.id, Some(ctx.traceId)))
+
+      _ <- if (runtime.status.isStopping) F.unit
```
rtitle, line 407: Maybe possible to combine the logic into 1 if-statement?
```
if (runtime.status.isStopping) F.unit
else if (runtime.status.isStoppable) {
  for {
    _ <- clusterQuery.updateClusterStatus(..)
   _ <- publisherQueue.enqueue1(..)
  } yield ()
} else F.raiseError[Unit](RuntimeCannotBeStoppedException(..)
```

rtitle (changes requested):
See comment above.. actually a bit confused how tests passed without the for-comp in `stopRuntime`, unless I'm missing something.. 🤔

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeServiceInterp.scala
```diff
@@ -403,12 +403,16 @@ class RuntimeServiceInterp[F[_]: Parallel](config: RuntimeServiceConfig,
 
       _ <- if (hasStopPermission) F.unit else F.raiseError[Unit](ForbiddenError(userInfo.userEmail))
       // throw 409 if the cluster is not stoppable
-      _ <- if (runtime.status.isStoppable) F.unit
-      else
+
+      _ <- if (runtime.status.isStopping) F.unit
```
Qi77Qi, line 407: shouldn't `Stopped` status in this branch as well?

rtitle: Oh yeah, +1.. so maybe `stoppingStatuses` should be `Set(Prestopping, Stopping, Stopped)`

ehanna4: oh i think i misunderstood your previous comment, will add it back

rtitle (approved):

Qi77Qi (approved):

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeServiceInterp.scala
kyuksel: good catch!

PR #2130, opened 2021-05-18
[PROD-549] Fix Galaxy serviceAccount configs

rtitle:
https://broadworkbench.atlassian.net/browse/PROD-549

Actually it looks like we can roll forward instead of revert. 

Basically Leo expects to pass the KSA (which we annotate with the pet GSA) to the Galaxy helm chart. The helm fields changed however, so Galaxy was not using this KSA. We think this is the root cause for the filepicker isses (no credentials for fiss) and job failures (no permissions to create jobs in the namespace).

---

In all cases:


Qi77Qi (approved):

rtitle:
Github actions is down (https://www.githubstatus.com/), but unit/automation tests passed in Jenkins. Merging this.

PR #2135, opened 2021-05-24
[IA-2630] Poll group updates in DataprocInterpreter

rtitle:
JIRA: https://broadworkbench.atlassian.net/browse/IA-2630
Depends on: https://github.com/broadinstitute/workbench-libs/pull/630

Attempt to fix errors like:
```
Required 'compute.images.useReadOnly' permission for 'projects/broad-dsp-gcr-public/global/images/custom-leo-image-dataproc-1-4-51-debian10-d26876f'
```

I think the cause is delays on the GCP side for group additions to the Dataproc custom image group to take effect. I added some polling after the group addition in `DataprocInterpreter`. 

I decided not to retry the Dataproc cluster creation as specified in the ticket because it's an asynchronous error, and we don't currently have a pattern for `create -> error -> delete -> recreate -> etc` during monitoring. If this PR doesn't fix it, we could revisit this. 

---

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/Boot.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/Boot.scala
@@ -396,9 +396,9 @@ object Boot extends IOApp {
       googlePublisher <- GooglePublisher.resource[F](publisherConfig)
       cryptoMiningUserPublisher <- GooglePublisher.resource[F](cryptominingTopicPublisherConfig)
       gkeService <- GKEService.resource(Paths.get(pathToCredentialJson), blocker, semaphore)
+
       // Retry 400 responses from Google, as those can occur when resources aren't ready yet
       // (e.g. if the subnet isn't ready when creating an instance).
-
       googleComputeRetryPolicy = RetryPredicates.retryConfigWithPredicates(RetryPredicates.standardGoogleRetryPredicate,
                                                                            RetryPredicates.whenStatusCode(400))
 
@@ -409,13 +409,19 @@ object Boot extends IOApp {
         googleComputeRetryPolicy
       )
 
+      googleDataprocRetryPolicy = RetryPredicates.retryConfigWithPredicates(
+        RetryPredicates.standardGoogleRetryPredicate,
+        RetryPredicates.whenStatusCode(400)
+      )
+
       dataprocService <- GoogleDataprocService
         .resource(
           googleComputeService,
           pathToCredentialJson,
           blocker,
           semaphore,
-          dataprocConfig.supportedRegions
+          dataprocConfig.supportedRegions,
+          googleDataprocRetryPolicy
         )
 
       _ <- OpenTelemetryMetrics.registerTracing[F](Paths.get(pathToCredentialJson), blocker)
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreter.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreter.scala
@@ -122,15 +122,12 @@ class DataprocInterpreter[F[_]: Timer: Parallel: ContextShift](
         _ <- vpcAlg.setUpProjectFirewalls(
           SetUpProjectFirewallsParams(params.runtimeProjectAndName.googleProject, network, machineConfig.region)
         )
-        resourceConstraints <- getClusterResourceContraints(params.runtimeProjectAndName,
-                                                            machineConfig.masterMachineType,
-                                                            machineConfig.region)
-
-        // Set up IAM roles necessary to create a cluster.
-        _ <- createClusterIamRoles(params.runtimeProjectAndName.googleProject, params.serviceAccountInfo)
 
         // Add member to the Google Group that has the IAM role to pull the Dataproc image
         _ <- updateDataprocImageGroupMembership(params.runtimeProjectAndName.googleProject, createCluster = true)
+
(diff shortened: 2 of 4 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/Boot.scala
```diff
@@ -410,13 +410,19 @@ object Boot extends IOApp {
         googleComputeRetryPolicy
       )
 
+      googleDataprocRetryPolicy = RetryPredicates.retryConfigWithPredicates(
```
rtitle, line 412: Not directly related to this ticket, but added a `RetryConfig` to Dataproc. See: https://github.com/broadinstitute/workbench-libs/pull/630

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreter.scala
```diff
@@ -435,9 +442,65 @@ class DataprocInterpreter[F[_]: Timer: Parallel: ContextShift](
   def removeClusterIamRoles(googleProject: GoogleProject, serviceAccountInfo: WorkbenchEmail): F[Unit] =
     updateClusterIamRoles(googleProject, serviceAccountInfo, createCluster = false)
 
+  // Called at Leo boot time. Sets up the Dataproc image user group if it does not already exist.
+  // Group membership is modified at Dataproc cluster creation time.
   def setupDataprocImageGoogleGroup(implicit ev: Ask[F, AppContext]): F[Unit] =
```
rtitle, line 450: Just refactoring: previously this logic was split across multiple methods, I feel it's easier to follow as a single for-comp.

project/Dependencies.scala
```diff
@@ -16,7 +16,7 @@ object Dependencies {
   val opencensusV = "0.28.3"
 
 
-  private val workbenchLibsHash = "197294d"
+  private val workbenchLibsHash = "db20f46-SNAP"
```
rtitle: TODO update before merge

rtitle:
Darn, I deployed this branch to a fiab and still repro'd the error:
image

So maybe more investigation needed (luckily it seems fairly easy to reproduce).

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreter.scala
```diff
@@ -439,9 +446,65 @@ class DataprocInterpreter[F[_]: Timer: Parallel: ContextShift](
...
+            s"Dataproc image user Google group '${config.groupsConfig.dataprocImageProjectGroupEmail}' already exists: $group \n Won't attempt to create it."
+          )
+      }
+      // Add compute.imageUser role to the group in the custom image's project
+      imageProject <- F.fromOption(parseImageProject(config.dataprocConfig.customDataprocImage),
+                                   ImageProjectNotFoundException)
+      _ <- logger.info(ctx.loggingCtx)(
+        s"Attempting to grant 'compute.imageUser' permissions to '${config.groupsConfig.dataprocImageProjectGroupEmail}' on project '$imageProject' ..."
+      )
+      _ <- retry(
```
Qi77Qi, line 495: does this call need to be inside `retry`?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreter.scala
```diff
@@ -686,15 +679,28 @@ class DataprocInterpreter[F[_]: Timer: Parallel: ContextShift](
       F.liftIO(IO.fromFuture(IO(googleDirectoryDAO.removeMemberFromGroup(groupEmail, memberEmail)))),
       when409
     )
+    val poll = for {
+      ctx <- ev.ask
+      d: DoneCheckable[Boolean] = (_: Boolean) == addToGroup
+      isMember <- Timer[F].sleep(5 seconds) >>
```
Qi77Qi: is it necessary to wait 5 seconds? just wonder if the group addition usually happens relatively fast

Qi77Qi (approved):

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreter.scala
rtitle: Probably not.. actually it doesn't seem like polling the group addition fixes the problem so I probably will remove it. I am not sure where the delay is actually happening causing the Dataproc SA to sometimes not have permission to the compute image.

The only other thing I can think of is retrying the entire Dataproc cluster creation, but that's tricky because it's an asynchronous error (meaning we'd have to delete, recreate, and restart the monitoring flow). But maybe it's doable in this case..

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreter.scala
rtitle: This call happens at boot time so probably not.. also probably doesn't hurt though

rtitle:
Thanks for reviewing @Qi77Qi! As noted above the group polling doesn't seem to fix the issue, but this PR is probably good to merge anyway. I think I'll remove the group polling and fix a couple other things

Qi77Qi:
@rtitle  yea, these async errors are tricky...I'm planning to show something in log in next walkthrough, which is an example of a similar async error that's hard for leo to capture.

rtitle:
Did another pass on this.. going to merge

PR #2144, opened 2021-05-26
[IA-2792] [QA-1487] Runtime creation failures after change to GPU/COS images

rtitle:
Problem statement: as part of the COS PR, [jupyter-docker-compose.yaml](https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/resources/init-resources/jupyter-docker-compose-gce.yaml#L45) references external services like `crypto-detector`, `proxy`, and `welder` in order to set up networking. This is problematic because we don't deploy `crypto-detector` on AoU runtimes, so the `jupyter-docker-compose` crashes. (Note we don't install `crypto-detector` on AoU because AoU requires Dockerhub images, and cryptomining isn't really a problem anyway on AoU.)

I think it would be better if all the docker-compose files were independent of each other, so we could then apply them conditionally in `gce-init.sh` (this is how it was before the COS PR). So in this PR I tried to make each docker-compose file independent. This means we need GCE and Dataproc versions for each of them (because GCE uses a network and Dataproc doesn't); but the changes aren't too bad.

I tested this in a fiab, using Terra UI to create GCE and Dataproc runtimes, and verified all the images launch as expected. I also verified that launching the AoU Dockerhub image works (via Leo swagger), and the cryptomining container does NOT get launched as expected.

Also running automation tests, but hopefully they will pass.

---

In all cases:

### DIFF
--- a/http/src/main/resources/init-resources/gce-init.sh
+++ b/http/src/main/resources/init-resources/gce-init.sh
@@ -465,7 +465,7 @@ if [ ! -z "$JUPYTER_DOCKER_IMAGE" ] ; then
   docker exec $JUPYTER_SERVER_NAME /bin/bash -c "sed -i 's/127.0.0.1/welder/g' /etc/jupyter/custom/jupyter_delocalize.py"
 
   # make sure home directory is owned by jupyter-user
-  docker exec -u root $JUPYTER_SERVER_NAME /bin/bash -c "chown -R jupyter-user:users ${JUPYTER_USER_HOME}"
+  docker exec -u root $JUPYTER_SERVER_NAME /bin/bash -c "chown -R jupyter-user:users ${NOTEBOOKS_DIR}"
 
   log 'Starting Jupyter Notebook...'
   retry 3 docker exec -d $JUPYTER_SERVER_NAME /bin/bash -c "${JUPYTER_SCRIPTS}/run-jupyter.sh ${NOTEBOOKS_DIR}"
--- a/http/src/main/resources/init-resources/gce-init.sh
+++ b/http/src/main/resources/init-resources/gce-init.sh
@@ -332,6 +332,9 @@ if [ ! -z "$JUPYTER_DOCKER_IMAGE" ] ; then
   mkdir -p ${WORK_DIRECTORY}/packages
   chmod a+rwx ${WORK_DIRECTORY}/packages
 
+  # make sure home directory is owned by jupyter-user
+  docker exec -u root ${JUPYTER_SERVER_NAME} chown -R jupyter-user:users $JUPYTER_HOME
+
   # Used to pip install packacges
   JUPYTER_USER_PIP_DIR=$JUPYTER_USER_HOME/.local/lib/python3.7/site-packages
 
--- a/http/src/main/resources/init-resources/gce-init.sh
+++ b/http/src/main/resources/init-resources/gce-init.sh
@@ -332,9 +332,6 @@ if [ ! -z "$JUPYTER_DOCKER_IMAGE" ] ; then
   mkdir -p ${WORK_DIRECTORY}/packages
   chmod a+rwx ${WORK_DIRECTORY}/packages
 
-  # make sure home directory is owned by jupyter-user
-  docker exec -u root ${JUPYTER_SERVER_NAME} chown -R jupyter-user:users $JUPYTER_HOME
-
   # Used to pip install packacges
   JUPYTER_USER_PIP_DIR=$JUPYTER_USER_HOME/.local/lib/python3.7/site-packages
 
@@ -467,8 +464,12 @@ if [ ! -z "$JUPYTER_DOCKER_IMAGE" ] ; then
   # For older jupyter images, jupyter_delocalize.py is using 127.0.0.1 as welder's url, which won't work now that we're no longer using `network_mode: host` for GCE VMs
   docker exec $JUPYTER_SERVER_NAME sed -i  "s/127.0.0.1/welder/g" /etc/jupyter/custom/jupyter_delocalize.py
 
+  # make sure home directory is owned by jupyter-user
+  docker exec -u root ${JUPYTER_SERVER_NAME} chown -R jupyter-user:users $JUPYTER_HOME
+
   log 'Starting Jupyter Notebook...'
-  retry 3 docker exec -d ${JUPYTER_SERVER_NAME} ${JUPYTER_SCRIPTS}/run-jupyter.sh ${NOTEBOOKS_DIR}
+  sleep 5
+  retry 3 docker exec -d ${JUPYTER_SERVER_NAME} /bin/bash -c "${JUPYTER_SCRIPTS}/run-jupyter.sh ${NOTEBOOKS_DIR}"
 
   # done start Jupyter
   STEP_TIMINGS+=($(date +%s))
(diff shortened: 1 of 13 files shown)


http/src/main/resources/init-resources/gce-init.sh
```diff
@@ -321,8 +321,6 @@ ${DOCKER_COMPOSE} --env-file=/var/variables.env "${COMPOSE_FILES[@]}" config
 
 retry 5 ${DOCKER_COMPOSE} --env-file=/var/variables.env "${COMPOSE_FILES[@]}" pull
 
-chmod a+rwx ${WORK_DIRECTORY}
```
rtitle, line 324: This was added in a prior PR, but not really needed

http/src/main/resources/init-resources/gce-init.sh
```diff
@@ -466,8 +464,12 @@ if [ ! -z "$JUPYTER_DOCKER_IMAGE" ] ; then
   # For older jupyter images, jupyter_delocalize.py is using 127.0.0.1 as welder's url, which won't work now that we're no longer using `network_mode: host` for GCE VMs
   docker exec $JUPYTER_SERVER_NAME sed -i  "s/127.0.0.1/welder/g" /etc/jupyter/custom/jupyter_delocalize.py
 
+  # make sure home directory is owned by jupyter-user
+  docker exec -u root ${JUPYTER_SERVER_NAME} chown -R jupyter-user:users $JUPYTER_HOME
```
rtitle: I noticed some directories owned by root; not exactly sure why, but figured this is a good idea regardless

http/src/main/resources/init-resources/gce-init.sh
```diff
@@ -466,8 +464,12 @@ if [ ! -z "$JUPYTER_DOCKER_IMAGE" ] ; then
...
+
   log 'Starting Jupyter Notebook...'
-  retry 3 docker exec -d ${JUPYTER_SERVER_NAME} ${JUPYTER_SCRIPTS}/run-jupyter.sh ${NOTEBOOKS_DIR}
+  sleep 5
```
rtitle: I'm not sure why this sleep is needed, and it makes me nervous. Without it, the `run-jupyter.sh` script was not actually taking effect, making me think there's some race condition. Will keep investigating this, but wanted to get a PR out

http/src/main/resources/reference.conf
```diff
@@ -442,14 +442,13 @@ gceClusterResources {
...
-  proxyDockerCompose = "proxy-docker-compose.yaml"
-  welderDockerCompose = "welder-docker-compose.yaml"
+  proxyDockerCompose = "proxy-docker-compose-gce.yaml"
+  welderDockerCompose = "welder-docker-compose-gce.yaml"
   proxySiteConf = "cluster-site-gce.conf"
   jupyterNotebookConfigUri = "jupyter_notebook_config.py"
   jupyterNotebookFrontendConfigUri = "notebook.json"
   customEnvVarsConfigUri = "custom_env_vars.env"
-  cryptoDetectorDockerCompose = "crypto-detector-docker-compose.yaml"
-  networkDockerCompose = "network-gce.yaml"
```
rtitle, line 452: `network-gce.yaml` does not exist in Leo and is not referenced elsewhere, so removed it

kyuksel (approved):
LGTM  👍🏼   pending integration tests passing and fixing Trivy vulnerabilities.
It may also be good for @Qi77Qi to take a look when she's back.

http/src/main/resources/init-resources/gce-init.sh
rtitle: Think I figured this out, removed the sleep

rtitle:
Thanks @kyuksel! Yeah I feel pretty good based on manual testing -- will confirm automation tests and `multi-test` as well. (I also was able to remove the `sleep` 🎉 )

Yeah weird trivy is failing, the Leo Dockerfile hasn't changed. Will look more..

rtitle:
jenkins retest

rtitle:
jenkins retest, sbt breakage

rtitle:
multi-test

rtitle:
Yay trivy resolved

rtitle:
multi-test

rtitle:
multi-test

rtitle:
multi-test

rtitle:
I think this PR is good but I'm trying to address some test flakiness before merge. In the most recent run 1/4 tests failed with:
```
org.broadinstitute.dsde.workbench.leonardo.runtimes.RuntimeCreationDiskSpec.create runtime and attach an existing persistent disk
```

rtitle:
multi-test

rtitle:
4/4 passes after my last commit! Doing some final manual testing of the aou image, then I will merge this I think and re-enable Argo syncs in dev.

http/src/main/resources/init-resources/gce-init.sh
```diff
@@ -464,10 +462,14 @@ if [ ! -z "$JUPYTER_DOCKER_IMAGE" ] ; then
...
-  docker exec $JUPYTER_SERVER_NAME sed -i  "s/127.0.0.1/welder/g" /etc/jupyter/custom/jupyter_delocalize.py
+  docker exec $JUPYTER_SERVER_NAME /bin/bash -c "sed -i 's/127.0.0.1/welder/g' /etc/jupyter/custom/jupyter_delocalize.py"
+
+  # make sure home directory is owned by jupyter-user
+  docker exec -u root $JUPYTER_SERVER_NAME /bin/bash -c "chown -R jupyter-user:users ${NOTEBOOKS_DIR}" || true
+  docker exec -u root $JUPYTER_SERVER_NAME /bin/bash -c "chown -R jupyter-user:users ${JUPYTER_USER_HOME}/.local/share/jupyter/" || true
 
   log 'Starting Jupyter Notebook...'
-  retry 3 docker exec -d ${JUPYTER_SERVER_NAME} ${JUPYTER_SCRIPTS}/run-jupyter.sh ${NOTEBOOKS_DIR}
+  retry 3 docker exec -u jupyter-user -d $JUPYTER_SERVER_NAME /bin/bash -c "${JUPYTER_SCRIPTS}/run-jupyter.sh ${NOTEBOOKS_DIR}"
```
Qi77Qi, line 472: hmm...this will not work for new jupyter images since the user won't be the same...what was the reason adding `-u`?

http/src/main/resources/init-resources/jupyter-docker-compose-gce.yaml
```diff
@@ -37,18 +37,6 @@ services:
       - /var/custom_env_vars.env
     # See https://docs.docker.com/engine/reference/run/#user-memory-constraints
     mem_limit: ${MEM_LIMIT}   # hard limit on memory consumption by the container
-  welder:
```
Qi77Qi, line 40: hmm...I did start with duplicating `welder-docker-compose` and `proxy-docker-compose`, but figured it'd be nicer if we can just have one `-gce` and non gce compose file, and have separate network seting appropriately. Since AOU complicates things a bit, I was thinking we can just have a separate `crypto-detector-docker-compose.yaml` that we attach conditionally

http/src/main/resources/init-resources/gce-init.sh
rtitle: Ah I forgot the user is different in the new images. I don't think `-u` is really needed, I was trying to fix test flakiness (thought it might be starting as root somehow)

http/src/main/resources/init-resources/jupyter-docker-compose-gce.yaml
rtitle: I thought it was better if each `*-docker-compose.yaml` file was independent and didn't reference each other. That way we can apply them conditionally without worrying about dependencies. We already have separate GCE and non-GCE `clusterResources` in `reference.conf` so there isn't much extra templating needed.

http/src/main/resources/init-resources/gce-init.sh
rtitle: Probably also need to fix the `chown` commands above (again, not sure that is really needed, was trying to fix test flakiness but wasn't really successful).

PR #2177, opened 2021-06-15
IA-2820 try fixing starting runtime bug

Qi77Qi:

---

In all cases:


http/src/main/resources/init-resources/startup.sh
```diff
@@ -95,6 +95,13 @@ else
     GSUTIL_CMD='gsutil'
     GCLOUD_CMD='gcloud'
     DOCKER_COMPOSE='docker-compose'
+
+    # Update old welder docker-compose file's entrypoint
+    sed -i  "s/tail -f \/dev\/null/\/opt\/docker\/bin\/entrypoint.sh/g" /etc/welder-docker-compose.yaml
```
Qi77Qi: sed is tested against a prod instance

rtitle: Looks like this is Dataproc `else` block? Should it be done for GCE and Dataproc?

Qi77Qi: yea, so the condition is to check if `/var/certs/jupyter-server.crt` exists (the comment is a bit misleading since it's only talking about new images), which will be `false` if the gce runtime is old (created before the gpu image change)

rtitle: Ahh I see

rtitle (approved):

Qi77Qi:
jenkins retest

PR #2183, opened 2021-06-21
Make terra remote file source writeable in Galaxy

almahmoud:
Friends with https://github.com/galaxyproject/galaxykubeman-helm/pull/31

Change was added to chart, but also needed here in order not to overwrite it


Qi77Qi:
will this require a galaxy chart version upgrade?

almahmoud:
It works independent of it, but yes it's meant to go with the new GKM version that will have Galaxy `21.05`.

Qi77Qi (approved):
can you add a comment next to this line about this is available since galaxy 21.05...

approving assuming there will be a followup PR bumping the chart version

rtitle:
Closing in favor of https://github.com/DataBiosphere/leonardo/pull/2187

PR #2208, opened 2021-07-14
[IA-2804] update db entry when nodepool errors during delete

xuscrbw:
https://broadworkbench.atlassian.net/browse/IA-2804

We have been running a bug in Janitor where nodepool remover fails to transition errored nodepols to a deleted state. We think this happens when Leo issues a deleteNodepool call to GKE and the nodepool doesn't exist.

This change marks the db entry as deleted even if it errors out, which should solve the issue of janitor not taking care of errored entries.

The only place that this deleteNodepool gets called should be from the cron job msg. Since we don't really expose nodepool status to the user, it should be safe to raise the error but also turn the DB entry to deleted
---

In all cases:


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -541,7 +541,7 @@ class GKEInterpreter[F[_]: Parallel: ContextShift: Timer](
         } yield operationOpt
       }
 
-      _ <- operationOpt.traverse(_ => nodepoolQuery.markAsDeleted(params.nodepoolId, ctx.now).transaction)
+      _ <- nodepoolQuery.markAsDeleted(params.nodepoolId, ctx.now).transaction
```
jdcanas, line 544: So, if I remember correctly, the issue stems from the NonLeo pubsub subscriber calling this function, and the nodepool not terminating in a deleted status. (feel free to provide more context if this isn't the case)

It seem like, if the nodepool is not actually deleted, we'd want to mark it as error and record the error in some way, not just mark it as deleted anyways. Unfortunately, we currently have an app-centric error recording model... 

The LeoPubsubMessageSubscriber has infrastructure where you can raise an error at any stage of message processing, and the source is at least recorded (i.e., `nodepool` being the source here) and logged in the app error table. The solution I'd imagine is 
1) we want to introduce something like this error bubbling up and being logged in `NonLeoMessageSubscriber`
2) we'd want to `.raiseError` in a `fold` on the operationOpt, essentially raising an error if the delete failed and logging it, so we can investigate later. Right now, and with the solution above, I think the error is snarfed and never seen again?

Qi77Qi (approved):
would be nice to add a unit test...but otherwise looks good to me

rtitle (approved):
I think this makes sense.. if `operationOpt` is `None` then the nodepool doesn't exist in the first place, so it makes sense to just mark it as Deleted.

(If the nodepool _did_ exist and the deletion timed out, then the method would raise an error I think.)

kyuksel:
> (If the nodepool did exist and the deletion timed out, then the method would raise an error I think.)

Yea, the recovery logic in `workbench-libs` is specific to the status code of `404`.

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreterSpec.scala
```diff
@@ -159,4 +165,40 @@ class GKEInterpreterSpec extends AnyFlatSpecLike with TestComponent with Leonard
 
     res.unsafeRunSync()
   }
+
+  it should "deleteAndPollNodepool handles errored nodepool deletion" in isolatedDbTest {
```
kyuksel: I wonder if something like

```scala
it should "mark a nodepool as Deleted in DB when it doesn't exist in Google"  in isolatedDbTest {
```

be easier to understand?

kyuksel (approved):
LGTM 👍🏼  with a minor comment inline

xuscrbw:
jenkins retest

PR #2213, opened 2021-07-20
[IA-2868] Memory updates not reflected in RStudio

rtitle:
https://broadworkbench.atlassian.net/browse/IA-2868

This came from UL. When a runtime PATCH call went through a stop/start to update machine type, the _old_ machine type was still being used to determine docker memory limits. Changed it to use the new value instead.

Also fixed an error with the `docker update` call (see comment below) with large memory sizes, potentially due to a newer version of docker-compose.

Tested in a fiab:

1. Create an RStudio runtime with 4 CPUs and 15G memory
2. Open RStudio, go to `Tools -> Memory -> Memory Usage Report`, see ~13G
3. In Terra UI, update the machine type to 8 CPUs and 52G memory
4. The VM goes through stop/start cycle
5. Open RStudio again, go to `Tools -> Memory -> Memory Usage Report`, see ~50G

I tried a few other updates to go larger & smaller, seems to work correctly now.

---

In all cases:

### DIFF
--- a/http/src/main/resources/init-resources/rstudio-docker-compose-gce.yaml
+++ b/http/src/main/resources/init-resources/rstudio-docker-compose-gce.yaml
@@ -28,6 +28,7 @@ services:
       - /var/custom_env_vars.env
     # See https://docs.docker.com/engine/reference/run/#user-memory-constraints
     mem_limit: ${MEM_LIMIT}   # hard limit on memory consumption by the container
+    memswap_limit: ${MEM_LIMIT}
 networks:
   app_network:
     external: true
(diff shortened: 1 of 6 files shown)


http/src/main/resources/init-resources/rstudio-docker-compose-gce.yaml
```diff
@@ -28,6 +28,7 @@ services:
       - /var/custom_env_vars.env
     # See https://docs.docker.com/engine/reference/run/#user-memory-constraints
     mem_limit: ${MEM_LIMIT}   # hard limit on memory consumption by the container
+    memswap_limit: ${MEM_LIMIT}
```
rtitle, line 31: I was seeing this error when increasing the memory to ~60G:
```
Jul 20 00:31:07 saturn-222c020b-94b4-4f75-9cca-9203affc86c3 GCEMetadataScripts[451]: 2021/07/20 00:31:07 GCEMetadataScripts: startup-script: Error response from daemon: Cannot update container fedd86e891f95a351657188e3ef09c31818895a74b013c0674533f6be159646e: Memory limit should be smaller than already set memoryswap limit, update the memoryswap at the same time
```
Reading docs it feels reasonable to have mem limit == swap limit.

Qi77Qi (approved):
lgtm...if test looks good too

PR #2218, opened 2021-07-21
[No Ticket] Bump dependencies

broadbot:
Consolidate scala-steward PRs


rtitle:
ScalaFix did some real work this time

Qi77Qi:
probably needs to bump https://github.com/broadinstitute/workbench-libs/blob/develop/project/Dependencies.scala#L11 first...or directly bump to 0.23  and bump cats-effect to 3 with a minor version bump in wb-libs

rtitle:
Yeah that sounds right

PR #2220, opened 2021-07-22
IA-2853 check whether user is register and enabled for proxy api calls

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-2853

---

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpSamDAO.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpSamDAO.scala
@@ -340,24 +340,23 @@ class HttpSamDAO[F[_]: Effect](httpClient: Client[F], config: HttpSamDaoConfig, 
         )(s => Effect[F].pure(s))
       )
       resp <- getUserSubjectIdFromToken(token)
+    } yield resp
+
+  override def getUserSubjectIdFromToken(token: String)(implicit ev: Ask[F, TraceId]): F[Option[UserSubjectId]] = {
+    val authHeader = Authorization(Credentials.Token(AuthScheme.Bearer, token))
+
+    for {
+      resp <- httpClient.expectOptionOr[GetGoogleSubjectIdResponse](
+        Request[F](
+          method = Method.GET,
+          uri = config.samUri.withPath(
+            s"/register/user/v2/self/info"
+          ),
+          headers = Headers.of(authHeader)
+        )
+      )(onError)
     } yield resp.map(_.userSubjectId)
-
-  override def getUserSubjectIdFromToken(token: String)(implicit ev: Ask[F, TraceId]): F[Option[UserSubjectId]] =
-   {
-     val authHeader = Authorization(Credentials.Token(AuthScheme.Bearer, token))
-
-     for {
-       resp <- httpClient.expectOptionOr[GetGoogleSubjectIdResponse](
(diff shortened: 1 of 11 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpSamDAO.scala
```diff
@@ -339,7 +339,13 @@ class HttpSamDAO[F[_]: Effect](httpClient: Client[F], config: HttpSamDaoConfig,
           )
         )(s => Effect[F].pure(s))
       )
-      authHeader = Authorization(Credentials.Token(AuthScheme.Bearer, token))
+      resp <- getUserSubjectIdFromToken(token)
+    } yield resp
+
+  override def getUserSubjectIdFromToken(token: String)(implicit ev: Ask[F, TraceId]): F[Option[UserSubjectId]] = {
```
rtitle, line 345: Given an invalid token will this return `F[None]` or an error?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/ProxyService.scala
```diff
@@ -124,6 +125,8 @@ class ProxyService(
           val res = for {
             now <- nowInstant
             userInfo <- googleOauth2Service.getUserInfoFromToken(key)
+            userOpt <- samDAO.getUserSubjectIdFromToken(key)
+            _ <- IO.fromOption(userOpt)(AuthenticationError(Some(userInfo.userEmail)))
```
rtitle, line 129: Trying to remember what happens if an exception is thrown inside the `CacheLoader`. I think it gets re-thrown as `ExecutionException` in the `cache.get()` call?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/ProxyService.scala
```diff
@@ -138,7 +141,8 @@ class ProxyService(
   def getCachedUserInfoFromToken(token: String): IO[UserInfo] =
     for {
       cache <- blocker.blockOn(IO(googleTokenCache.get(token))).adaptError {
-        case _ =>
+        case e: AuthenticationError => e
+        case _                      =>
```
rtitle, line 145: Ah I see - this will map `cache.get` exceptions to `AuthenticationError`

rtitle (approved):
👍 this seems correct to me. cc @zbedo , this resolves the ToS vulnerability you had raised [here](https://docs.google.com/document/d/1FLcK_Go24qr8GMm34-SlZws04FePP4bxNTK0daRhQVs/edit?ts=60ef4f71#heading=h.7r2bv4i6ss7l)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpSamDAO.scala
Qi77Qi: So I think if the token is invalid, the previous google auth call will fail. This will return `None` (sam call returns 404) if the user is not found

rtitle: Ah, ok

zbedo:
This looks good to me 👍🏻

PR #2223, opened 2021-07-23
[IA-2836] Proxy Spark web UI through Leonardo

rtitle:
https://broadworkbench.atlassian.net/browse/IA-2836

See design doc: https://docs.google.com/document/d/1Zr6qzICStrrJwYkfoIE7fTf9Ziu3f6JmhdA9Ao3Y8Bo/edit

Depends on https://github.com/broadinstitute/workbench-libs/pull/697

This does the following:
1. Adds `DataprocConfig.componentGatewayEnabled` field to DB, models, JSON decoders. Defaults to `false` for now.
2. If that field is true, enables [Component Gateway](https://cloud.google.com/dataproc/docs/concepts/accessing/dataproc-gateways) when creating Dataproc cluster
3. Updates proxy rules to proxy Spark web UIs (see doc for details)
4. Updates swagger
5. Adds an automation test

It seems to work in a fiab, still doing some more detailed testing of the Spark UIs themselves.

UPDATE: I ran through the [Hail GWAS tutorial](https://hail.is/docs/0.2/tutorials/01-genome-wide-association-study.html) and clicked through all the UIs and they all seem functional. 🎉 

---

In all cases:

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/runtimeRoutesModels.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/runtimeRoutesModels.scala
@@ -59,7 +59,8 @@ object RuntimeConfigRequest {
                                   numberOfWorkerLocalSSDs: Option[Int] = None, //min 0 max 8
                                   numberOfPreemptibleWorkers: Option[Int] = None,
                                   properties: Map[String, String],
-                                  region: Option[RegionName] = None)
+                                  region: Option[RegionName],
+                                  componentGatewayEnabled: Option[Boolean])
       extends RuntimeConfigRequest {
     val cloudService: CloudService = CloudService.Dataproc
   }
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/runtimeRoutesModels.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/runtimeRoutesModels.scala
@@ -60,7 +60,7 @@ object RuntimeConfigRequest {
                                   numberOfPreemptibleWorkers: Option[Int] = None,
                                   properties: Map[String, String],
                                   region: Option[RegionName],
-                                  componentGatewayEnabled: Option[Boolean])
+                                  componentGatewayEnabled: Boolean)
       extends RuntimeConfigRequest {
     val cloudService: CloudService = CloudService.Dataproc
   }
--- a/http/src/main/resources/swagger/api-docs.yaml
+++ b/http/src/main/resources/swagger/api-docs.yaml
@@ -1613,7 +1613,7 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/ErrorReport"
-  "/proxy/{googleProject}/{runtimeName}/spark-yarn":
+  "/proxy/{googleProject}/{runtimeName}/yarn":
     get:
       summary: YARN Resource Manager
       description: >
@@ -1668,7 +1668,7 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/ErrorReport"
-  "/proxy/{googleProject}/{runtimeName}/spark-jobhistory":
+  "/proxy/{googleProject}/{runtimeName}/jobhistory":
     get:
       summary: MapReduce Job History
       description: >
@@ -1723,7 +1723,7 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/ErrorReport"
-  "/proxy/{googleProject}/{runtimeName}/spark-apphistory":
+  "/proxy/{googleProject}/{runtimeName}/apphistory":
     get:
       summary: YARN Application Timeline
       description: >
@@ -1778,7 +1778,7 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/ErrorReport"
-  "/proxy/{googleProject}/{runtimeName}/spark-sparkhistory":
+  "/proxy/{googleProject}/{runtimeName}/sparkhistory":
     get:
       summary: Spark History Server
       description: >
@@ -1833,7 +1833,7 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/ErrorReport"
-  "/proxy/{googleProject}/{runtimeName}/spark-hdfs":
+  "/proxy/{googleProject}/{runtimeName}/hdfs":
     get:
       summary: HDFS NameNode
       description: >
--- a/http/src/main/resources/swagger/api-docs.yaml
+++ b/http/src/main/resources/swagger/api-docs.yaml
@@ -1320,14 +1320,14 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/ErrorReport"
-        "420":
-          description: Runtime not ready
+        "422":
+          description: Runtime is stopped
           content:
             application/json:
               schema:
                 $ref: "#/components/schemas/ErrorReport"
-        "422":
-          description: Runtime is stopped
+        "423":
+          description: Runtime not ready
           content:
             application/json:
               schema:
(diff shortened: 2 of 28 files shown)


project/Dependencies.scala
```diff
@@ -20,7 +20,7 @@ object Dependencies {
   val serviceTestV = s"0.19-$workbenchLibsHash"
   val workbenchModelV = s"0.14-$workbenchLibsHash"
   val workbenchGoogleV = s"0.21-$workbenchLibsHash"
-  val workbenchGoogle2V = s"0.21-$workbenchLibsHash"
+  val workbenchGoogle2V = "0.21-30caedc3-SNAP"
```
rtitle: TODO: update to non SNAP

rtitle:
jenkins retest

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -1613,6 +1613,281 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/ErrorReport"
+  "/proxy/{googleProject}/{runtimeName}/yarn":
```
rtitle, line 1616: Looks like Swagger-UI alphabetizes endpoints so these get kind of mixed together:

image


I wonder if they should have a prefix like `/proxy/../spark/yarn`, `/proxy/../spark/hdfs`, etc?

rtitle:
I'll do a multi-test to be safe

rtitle:
multi-test

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/runtimeRoutesModels.scala
```diff
@@ -59,7 +59,8 @@ object RuntimeConfigRequest {
                                   numberOfWorkerLocalSSDs: Option[Int] = None, //min 0 max 8
                                   numberOfPreemptibleWorkers: Option[Int] = None,
                                   properties: Map[String, String],
-                                  region: Option[RegionName] = None)
+                                  region: Option[RegionName],
+                                  componentGatewayEnabled: Option[Boolean])
```
Qi77Qi: should it just be `Boolean`? just so we need one less layer of default? (default to `None`, and when it's `None`, default to `false`...so ultimately the default is `false` unless user says `true`

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -1320,14 +1320,14 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/ErrorReport"
-        "420":
-          description: Runtime not ready
+        "422":
```
Qi77Qi, line 1323: think we should bump the swagger version

project/Dependencies.scala
```diff
@@ -20,7 +20,7 @@ object Dependencies {
   val serviceTestV = s"0.19-$workbenchLibsHash"
   val workbenchModelV = s"0.14-$workbenchLibsHash"
   val workbenchGoogleV = s"0.21-$workbenchLibsHash"
-  val workbenchGoogle2V = s"0.21-$workbenchLibsHash"
+  val workbenchGoogle2V = "0.21-d89d5792-SNAP"
```
Qi77Qi: we can probably use the snap version for now...and once we have the other wb-libs ugprade PR ready, we can then update the lib

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/runtimeRoutesModels.scala
kyuksel: This makes sense to me too. Also wondered why we haven't done the same for a lot of the other parameters 🤔

http/src/main/resources/swagger/api-docs.yaml
kyuksel: Another option may be to hyphenate like in `/proxy/../spark-yarn`, `/proxy/../spark-hdfs`, etc.

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -1497,14 +1497,14 @@ paths:
...
+          description: App is stopped
           content:
             application/json:
               schema:
                 $ref: "#/components/schemas/ErrorReport"
-        "422":
-          description: App is stopped
+        "423":
+          description: App not ready
```
kyuksel, line 1507: What was the reason for the `42*` response code changes?

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -1613,6 +1613,281 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/ErrorReport"
+  "/proxy/{googleProject}/{runtimeName}/yarn":
+    get:
+      summary: YARN Resource Manager
+      description: >
+        Only available for Dataproc runtimes with Component Gateway enabled.
```
kyuksel, line 1620: On the UX, UI, User Ed and Front-line realms, we've been trying to deprecate the term `runtime` in favour of `cloud compute` (or just `compute`) where a `cloud environment` is composed of `application config`, `cloud compute` and optionally `persistent disk`. I wonder if we should try to do the same on the backend, at least on the user-facing portions. Having a consistent terminology makes communication and onboarding easier.

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -1613,6 +1613,281 @@ paths:
...
+            application/json:
+              schema:
+                $ref: "#/components/schemas/ErrorReport"
+        "500":
+          description: Internal Error
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/ErrorReport"
+  "/proxy/{googleProject}/{runtimeName}/jobhistory":
```
kyuksel, line 1671: I wonder if we should hyphenate URI components for easier reading as in `job-history`. That's what I mostly commonly see in API naming conventions.

kyuksel (commented):
Looks great! Mostly questions inline.

http/src/main/resources/swagger/api-docs.yaml
rtitle: I kind of like that convention

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/runtimeRoutesModels.scala
rtitle: Yeah that makes sense -- I think I made it `Option` here because it's optional in the request json (for backwards compatibility). But we can achieve that in the decoder too and make the class field `Boolean`.

http/src/main/resources/swagger/api-docs.yaml
rtitle: Will do!

http/src/main/resources/swagger/api-docs.yaml
rtitle: I just noticed that the status code returned by Leo didn't match the status code in swagger, so I updated Swagger docs.

http/src/main/resources/swagger/api-docs.yaml
rtitle: Hm, I stole the URL paths directly from Dataproc Component Gateway: https://cloud.google.com/dataproc/docs/concepts/accessing/dataproc-gateways

http/src/main/resources/swagger/api-docs.yaml
rtitle: I hear you but all the Swagger routes/models/etc are named `runtimes` so I think this is consistent from a Swagger perspective..

rtitle:
Thanks for the comments @kyuksel and @Qi77Qi! I'll apply feedback, rebase, and retest. Then I think this can be merged :)

kyuksel (approved):

Qi77Qi (approved):
👍

PR #2243, opened 2021-08-12
[IA-2437] Updating Leo Automation tests to use PPW

xuscrbw:
This PR was created in joint effort with CI to get Leonardo using PPW when claiming projects and creating workspace/unclaiming  project and deleting workspace.

We increased the timeout on some polling when we were running into some timeout issues with runtimes/apps. We also implemented some Sam resource clean up when deleteApp fails to ensure we can properly clean up the workspace after the tests finish
---

In all cases:

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoSuite.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoSuite.scala
@@ -11,7 +11,7 @@ import org.broadinstitute.dsde.workbench.leonardo.notebooks._
 import org.broadinstitute.dsde.workbench.leonardo.rstudio.RStudioSpec
 import org.broadinstitute.dsde.workbench.leonardo.runtimes._
 import org.broadinstitute.dsde.workbench.model.google.GoogleProject
-import org.broadinstitute.dsde.workbench.service.{BillingProject, Orchestration}
+import org.broadinstitute.dsde.workbench.service.{BillingProject, Orchestration, Rawls}
 import org.http4s.AuthScheme
 import org.http4s.Credentials.Token
 import org.http4s.headers.Authorization
@@ -112,13 +112,16 @@ trait GPAllocUtils extends BillingFixtures with LeonardoTestUtils {
                                                       ronEmail,
                                                       BillingProject.BillingProjectRole.User)(hermioneAuthToken)
       )
+      workspaceName = UUID.randomUUID().toString
       _ <- IO(
-        Orchestration.workspaces.create(claimedBillingProject.projectName, UUID.randomUUID().toString)(ronAuthToken)
+        Orchestration.workspaces.create(claimedBillingProject.projectName, workspaceName)(ronAuthToken)
       )
 
-      workspaceDetails <- IO(Orchestration.workspaces.)
+      workspaceDetails = Rawls.workspaces.getWorkspaceDetails(claimedBillingProject.projectName, workspaceName)(ronAuthToken)
+      googleProjectId <- IO(workspaceDetails.parseJson.asJsObject.getFields("googleProjectId"))
+      // _ <- loggerIO.info(s"Workspace details: ${googleProjectId}")
       _ <- loggerIO.info(s"Billing project claimed: ${claimedBillingProject.projectName}")
-    } yield GoogleProject(workspaceDetails.googleProject)
+    } yield GoogleProject(googleProjectId)
 
   /**
    * Unclaiming billing project claim by Hermione
--- a/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoSuite.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoSuite.scala
@@ -4,7 +4,6 @@ import cats.effect.IO
 import cats.syntax.all._
 import com.typesafe.scalalogging.LazyLogging
 import org.broadinstitute.dsde.rawls.model.WorkspaceName
-//import org.broadinstitute.dsde.rawls.model.Workspace
 import org.broadinstitute.dsde.workbench.fixture.{BillingFixtures}
 import org.broadinstitute.dsde.workbench.leonardo.GPAllocFixtureSpec.{shouldUnclaimProjectsKey, _}
 import org.broadinstitute.dsde.workbench.leonardo.apps.{AppCreationSpec, CustomAppCreationSpec}
@@ -42,7 +41,6 @@ trait GPAllocFixtureSpec extends FixtureAnyFreeSpecLike with Retries with LazyLo
 
     sys.props.get(googleProjectKey) match {
       case None => throw new RuntimeException("leonardo.googleProject system property is not set")
-      // case Some(msg) if msg.startsWith(gpallocErrorPrefix) => throw new RuntimeException(msg)
       case Some(googleProjectId) =>
         if (isRetryable(test))
           withRetry(runTestAndCheckOutcome(GoogleProject(googleProjectId)))
@@ -84,7 +82,6 @@ trait GPAllocUtils extends BillingFixtures with LeonardoTestUtils {
       _ <- IO(
         Orchestration.workspaces.create(claimedBillingProject.projectName, workspaceName)(ronAuthToken)
       )
-      // Testing with: sbt "testOnly *RuntimeAutopauseSpec"
       workspaceDetails <- IO(
         Rawls.workspaces.getWorkspaceDetails(claimedBillingProject.projectName, workspaceName)(ronAuthToken)
       )
--- a/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoSuite.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoSuite.scala
@@ -97,7 +97,7 @@ trait GPAllocUtils extends BillingFixtures with LeonardoTestUtils {
     for {
       _ <- IO(
         Orchestration.workspaces.delete(workspaceName.namespace, workspaceName.name)(ronAuthToken)
-      )
+      ).attempt
       _ <- IO(
         Orchestration.billing
           .removeUserFromBillingProject(workspaceName.namespace, ronEmail, BillingProject.BillingProjectRole.User)(
--- a/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoSuite.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoSuite.scala
@@ -35,13 +35,9 @@ trait GPAllocFixtureSpec extends FixtureAnyFreeSpecLike with Retries with LazyLo
       outcome
     }
 
-    sys.props.get(workspaceNamespaceKey) match {
-      case Some(msg) if msg.startsWith(gpallocErrorPrefix) => throw new RuntimeException(msg)
-      case x => logger.info(s"Workspace namespace is : ${x}")
-    }
-
     sys.props.get(googleProjectKey) match {
       case None => throw new RuntimeException("leonardo.googleProject system property is not set")
+      case Some(msg) if msg.startsWith(gpallocErrorPrefix) => throw new RuntimeException(msg)
       case Some(googleProjectId) =>
         if (isRetryable(test))
           withRetry(runTestAndCheckOutcome(GoogleProject(googleProjectId)))
@@ -138,7 +134,7 @@ trait GPAllocBeforeAndAfterAll extends GPAllocUtils with BeforeAndAfterAll {
       _ <- loggerIO.info(s"Running GPAllocBeforeAndAfterAll beforeAll")
       claimAttempt <- claimGPAllocProjectAndCreateWorkspace().attempt
       _ <- claimAttempt match {
-        case Left(e) => IO(sys.props.put(workspaceNamespaceKey, gpallocErrorPrefix + e.getMessage))
+        case Left(e) => IO(sys.props.put(googleProjectKey, gpallocErrorPrefix + e.getMessage))
         case Right(googleProjectAndWorkspaceName) =>
           IO(
             sys.props.addAll(
--- a/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoSuite.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoSuite.scala
@@ -34,14 +34,18 @@ trait GPAllocFixtureSpec extends FixtureAnyFreeSpecLike with Retries {
       outcome
     }
 
+    sys.props.get(workspaceNamespaceKey) match {
+      case Some(msg) if msg.startsWith(gpallocErrorPrefix) => throw new RuntimeException(msg)
+    }
+
     sys.props.get(googleProjectKey) match {
       case None                                            => throw new RuntimeException("leonardo.googleProject system property is not set")
       case Some(msg) if msg.startsWith(gpallocErrorPrefix) => throw new RuntimeException(msg)
-      case Some(billingProject) =>
+      case Some(googleProjectId) =>
         if (isRetryable(test))
-          withRetry(runTestAndCheckOutcome(GoogleProject(billingProject)))
+          withRetry(runTestAndCheckOutcome(GoogleProject(googleProjectId)))
         else
-          runTestAndCheckOutcome(GoogleProject(billingProject))
+          runTestAndCheckOutcome(GoogleProject(googleProjectId))
     }
   }
 }
@@ -111,10 +115,10 @@ trait GPAllocUtils extends BillingFixtures with LeonardoTestUtils {
     val test = for {
       _ <- loggerIO.info("Allocating a new single-test project")
       googleProjectAndWorkspaceName <- claimGPAllocProjectAndCreateWorkspace()
-      _ <- loggerIO.info(s"Single test project $googleProjectAndWorkspaceName claimed")
+      _ <- loggerIO.info(s"Single test project ${googleProjectAndWorkspaceName.workspaceName.namespace} claimed")
       t <- testCode(googleProjectAndWorkspaceName.googleProject)
-      _ <- loggerIO.info(s"Releasing single-test project: ${googleProjectAndWorkspaceName.googleProject.value}")
-      _ <- unclaimProject(googleProjectAndWorkspaceName)
+      _ <- loggerIO.info(s"Releasing single-test project: ${googleProjectAndWorkspaceName.workspaceName.namespace}")
+      _ <- unclaimProject(googleProjectAndWorkspaceName.workspaceName)
     } yield t
 
     test.unsafeRunSync()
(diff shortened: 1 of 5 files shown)


automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoSuite.scala
```diff
@@ -2,22 +2,29 @@ package org.broadinstitute.dsde.workbench.leonardo
 
 import cats.effect.IO
 import cats.syntax.all._
-import org.broadinstitute.dsde.workbench.fixture.BillingFixtures
+import com.typesafe.scalalogging.LazyLogging
+import org.broadinstitute.dsde.rawls.model.WorkspaceName
+//import org.broadinstitute.dsde.rawls.model.Workspace
```
jacmrob: can get rid of the commented out import

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoSuite.scala
```diff
@@ -61,34 +80,54 @@ trait GPAllocUtils extends BillingFixtures with LeonardoTestUtils {
...
+        Orchestration.workspaces.create(claimedBillingProject.projectName, workspaceName)(ronAuthToken)
+      )
+      // Testing with: sbt "testOnly *RuntimeAutopauseSpec"
+      workspaceDetails <- IO(
+        Rawls.workspaces.getWorkspaceDetails(claimedBillingProject.projectName, workspaceName)(ronAuthToken)
+      )
+      googleProjectId <- IO(
+        workspaceDetails.parseJson.asJsObject
+          .getFields("workspace")
+          .flatMap(workspace => workspace.asJsObject.getFields("googleProjectId"))
```
andy7i: on our v2_workspaces feature branches, we're reverting `googleProjectId` back to `googleProject` so that it is backwards compatible with the existing APIs. I think we can replace the `googleProjectId` here with `googleProject` and leave it as is in other places to disambiguate it (as we're doing in our PR https://github.com/broadinstitute/rawls/pull/1496/files#diff-77ed35dcd112268da988b8799c4aeab677725f2acd7c59b01942d38a511b5306R580)

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoSuite.scala
```diff
@@ -28,31 +35,43 @@ trait GPAllocFixtureSpec extends FixtureAnyFreeSpecLike with Retries {
       outcome
     }
 
-    sys.props.get(gpallocProjectKey) match {
-      case None                                            => throw new RuntimeException("leonardo.billingProject system property is not set")
+    sys.props.get(workspaceNamespaceKey) match {
       case Some(msg) if msg.startsWith(gpallocErrorPrefix) => throw new RuntimeException(msg)
-      case Some(billingProject) =>
+      case x                                               => logger.info(s"Workspace namespace is: ${x}")
```
jacmrob: can we get rid of this logging line? as well as the lazy logging import above?

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoSuite.scala
```diff
@@ -28,31 +35,43 @@ trait GPAllocFixtureSpec extends FixtureAnyFreeSpecLike with Retries {
...
+    }
+
+    sys.props.get(googleProjectKey) match {
+      case None => throw new RuntimeException("leonardo.googleProject system property is not set")
+      // case Some(msg) if msg.startsWith(gpallocErrorPrefix) => throw new RuntimeException(msg)
```
jacmrob: remove comment

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoSuite.scala
```diff
@@ -61,34 +80,54 @@ trait GPAllocUtils extends BillingFixtures with LeonardoTestUtils {
                                                       BillingProject.BillingProjectRole.User)(hermioneAuthToken)
       )
       _ <- loggerIO.info(s"Billing project claimed: ${claimedBillingProject.projectName}")
-    } yield GoogleProject(claimedBillingProject.projectName)
+      workspaceName = UUID.randomUUID().toString
+      _ <- IO(
+        Orchestration.workspaces.create(claimedBillingProject.projectName, workspaceName)(ronAuthToken)
+      )
+      // Testing with: sbt "testOnly *RuntimeAutopauseSpec"
```
jacmrob: remove comment

andy7i (commented):
not entirely sure if this will work in dev, but it's certainly worth a shot to reduce the # of feature branches we have. In v1 workspaces, the workspace's googleProject should just be the billing project's googleProject so I think it could work. 

We should update googleProjectId -> googleProject and see if tests pass with that.

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoSuite.scala
```diff
@@ -61,34 +80,54 @@ trait GPAllocUtils extends BillingFixtures with LeonardoTestUtils {
...
+        Rawls.workspaces.getWorkspaceDetails(claimedBillingProject.projectName, workspaceName)(ronAuthToken)
+      )
+      googleProjectId <- IO(
+        workspaceDetails.parseJson.asJsObject
+          .getFields("workspace")
+          .flatMap(workspace => workspace.asJsObject.getFields("googleProjectId"))
+          .head
+          .convertTo[String]
+      )
+      _ <- loggerIO.info(s"Workspace details: ${workspaceDetails}")
```
jacmrob: do we need to log this?

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/RuntimeFixtureSpec.scala
```diff
@@ -119,26 +119,31 @@ abstract class RuntimeFixtureSpec
     super.beforeAll()
     logger.info("beforeAll")
 
-    sys.props.get(gpallocProjectKey) match {
+    sys.props.get(workspaceNamespaceKey) match {
       case Some(msg) if msg.startsWith(gpallocErrorPrefix) =>
         clusterCreationFailureMsg = msg
-      case Some(billingProject) =>
-        Either.catchNonFatal(createRonRuntime(GoogleProject(billingProject))).handleError { e =>
+      case x => logger.info(s"Workspace namespace is: ${x}")
```
jacmrob: can we remove this log line?

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoSuite.scala
```diff
@@ -28,31 +33,41 @@ trait GPAllocFixtureSpec extends FixtureAnyFreeSpecLike with Retries {
       outcome
     }
 
-    sys.props.get(gpallocProjectKey) match {
-      case None                                            => throw new RuntimeException("leonardo.billingProject system property is not set")
+    sys.props.get(workspaceNamespaceKey) match {
```
rtitle: This `match` statement should account for all cases. 

Actually maybe we should just remove this `match` and put the `gpallocErrorPrefix` in the `googleProjectKey` property, and check for it down below.

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoSuite.scala
```diff
@@ -61,34 +76,52 @@ trait GPAllocUtils extends BillingFixtures with LeonardoTestUtils {
                                                       BillingProject.BillingProjectRole.User)(hermioneAuthToken)
       )
       _ <- loggerIO.info(s"Billing project claimed: ${claimedBillingProject.projectName}")
-    } yield GoogleProject(claimedBillingProject.projectName)
+      workspaceName = UUID.randomUUID().toString
```
rtitle: Nit: maybe `workspaceName <- IO(UUID.randomUUID().toString)` since it's technically a side effect

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoSuite.scala
```diff
@@ -61,34 +76,52 @@ trait GPAllocUtils extends BillingFixtures with LeonardoTestUtils {
...
+      _ <- IO(
+        Orchestration.workspaces.create(claimedBillingProject.projectName, workspaceName)(ronAuthToken)
+      )
+      workspaceDetails <- IO(
+        Rawls.workspaces.getWorkspaceDetails(claimedBillingProject.projectName, workspaceName)(ronAuthToken)
+      )
+      googleProjectId <- IO(
+        workspaceDetails.parseJson.asJsObject
```
rtitle: I wonder if we could use circe instead of `spray-json` for this…

rtitle: Example using circe:
```
      workspaceDetails <- IO(
        Rawls.workspaces.getWorkspaceDetails(claimedBillingProject.projectName, workspaceName)(ronAuthToken)
      )
      json <- IO.fromEither(parse(workspaceDetails))
      googleProjectOpt = json.hcursor.downField("workspace").get[String]("googleProject").toOption
      googleProject <- IO.fromOption(googleProjectOpt)(
        new Exception(s"Could not get googleProject from workspace $workspaceName")
      )
```

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoSuite.scala
```diff
@@ -102,11 +135,15 @@ trait GPAllocBeforeAndAfterAll extends GPAllocUtils with BeforeAndAfterAll {
...
       _ <- loggerIO.info(s"Running GPAllocBeforeAndAfterAll beforeAll")
-      claimAttempt <- claimProject().attempt
+      claimAttempt <- claimGPAllocProjectAndCreateWorkspace().attempt
       _ <- claimAttempt match {
-        case Left(e) => IO(sys.props.put(gpallocProjectKey, gpallocErrorPrefix + e.getMessage))
-        case Right(billingProject) =>
-          IO(sys.props.put(gpallocProjectKey, billingProject.value)) >> createInitialRuntime(billingProject)
+        case Left(e) => IO(sys.props.put(workspaceNamespaceKey, gpallocErrorPrefix + e.getMessage))
+        case Right(googleProjectAndWorkspaceName) =>
+          IO(sys.props.put(googleProjectKey, googleProjectAndWorkspaceName.googleProject.value)) >> IO(
```
rtitle: How about:
```
IO(
  sys.props.addAll(
    Map(
      googleProjectKey -> googleProjectAndWorkspaceName.googleProject.value,
      workspaceNamespaceKey -> googleProjectAndWorkspaceName.workspaceName.namespace,
      workspaceNameKey -> googleProjectAndWorkspaceName.workspaceName.name
    )
  )
)
```

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoSuite.scala
```diff
@@ -119,12 +156,21 @@ trait GPAllocBeforeAndAfterAll extends GPAllocUtils with BeforeAndAfterAll {
...
+      projectProp <- IO(sys.props.get(googleProjectKey))
+      workspaceNamespaceProp <- IO(sys.props.get(workspaceNamespaceKey))
+      workspaceNameProp <- IO(sys.props.get(workspaceNameKey))
       project = projectProp.filterNot(_.startsWith(gpallocErrorPrefix)).map(GoogleProject)
-      _ <- if (shouldUnclaimProp != Some("false")) {
-        project.traverse(p => deleteInitialRuntime(p) >> unclaimProject(p))
-      } else loggerIO.info(s"Not going to release project: ${projectProp} due to error happened")
-      _ <- IO(sys.props.remove(gpallocProjectKey))
+      _ <- if (!shouldUnclaimProp.contains("false")) {
+        project.traverse(p => deleteInitialRuntime(p)).flatMap { _ =>
```
rtitle: Maybe a nice case for `traverseN`:
```
(project, workspaceNamespaceProp, workspaceNameProp).traverseN { 
  case (p, n, w) =>
    deleteInitialRuntime(p) >> unclaimProject(WorkspaceName(n, w))
}
```

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoSuite.scala
```diff
@@ -119,12 +156,21 @@ trait GPAllocBeforeAndAfterAll extends GPAllocUtils with BeforeAndAfterAll {
...
+          workspaceNamespaceProp.traverse(workspaceNamespace =>
+            workspaceNameProp
+              .traverse(workspaceName => unclaimProject(WorkspaceName(workspaceNamespace, workspaceName)))
+          )
+        }
+      } else loggerIO.info(s"Not going to release project: ${workspaceNamespaceProp} due to error happened")
+      _ <- IO(sys.props.remove(googleProjectKey))
```
rtitle:
```
IO(sys.props.subtractAll(List(googleProjectKey, workspaceNamespaceKey, workspaceNameKey)))
```

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/RuntimeFixtureSpec.scala
```diff
@@ -119,26 +119,29 @@ abstract class RuntimeFixtureSpec
     super.beforeAll()
     logger.info("beforeAll")
 
-    sys.props.get(gpallocProjectKey) match {
+    sys.props.get(workspaceNamespaceKey) match {
```
rtitle: Also looks like an incomplete match -- I'd probably again suggest removing this and putting the `gpallocErrorPrefix ` in the `googleProjectKey` prop

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/apps/AppCreationSpec.scala
```diff
@@ -91,13 +93,18 @@ class AppCreationSpec extends GPAllocFixtureSpec with LeonardoTestUtils with GPA
             loggerIO.warn(
               s"AppCreationSpec: app ${googleProject.value}/${appName.value} did not finish deleting after 20 minutes. Result: $monitorDeleteResult"
             )
+            IO(Sam.user.deleteResource("kubernetes-app", appName.value)(ronCreds.makeAuthToken()))
```
rtitle: This doesn't seem correct -- maybe should be removed? Leo is supposed to delete the Sam resource when deletion finishes; also I think the Sam resource has a different name than the app name (a UUID).

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/apps/AppCreationSpec.scala
```diff
@@ -201,12 +208,13 @@ class AppCreationSpec extends GPAllocFixtureSpec with LeonardoTestUtils with GPA
...
           implicit0(doneCheckable: DoneCheckable[List[ListAppResponse]]) = appDeleted(appName)
-          monitorDeleteResult <- streamFUntilDone(listApps, 120, 10 seconds).compile.lastOrError
+          monitorDeleteResult <- streamFUntilDone(listApps, 180, 10 seconds).compile.lastOrError
           // TODO remove first case in below if statement when Galaxy deletion is reliable
           _ <- if (!doneCheckable.isDone(monitorDeleteResult)) {
             loggerIO.warn(
-              s"AppCreationSpec: app ${googleProject.value}/${appName.value} did not finish deleting after 20 minutes. Result: $monitorDeleteResult"
+              s"AppCreationSpec: app ${googleProject.value}/${appName.value} did not finish deleting after 30 minutes. Result: $monitorDeleteResult"
             )
+            IO(Sam.user.deleteResource("kubernetes-app", appName.value)(ronCreds.makeAuthToken()))
```
rtitle: Same comment here - not sure this will do anything, probably remove

rtitle (changes requested):
A few comments above. Overall looks close though, and the approach looks good. 

This PR reminded me of how much we're using system props to pass data across tests which maybe is a bit strange… but I think there were some good reasons related to how ScalaTest suites/fixtures work.

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/apps/AppCreationSpec.scala
xuscrbw: I feel like we added this because of some clean up we were having issues with. It may have been resolved by a time out extension we added. Will try a test run without these to verify that these aren't needed anymore.

I think the original intention was because we are now having to clean up a workspace, we may have issues in the instance where the workspace still has a child resource and cannot be cleaned up.

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoSuite.scala
xuscrbw: I think googleProjectKey here is just a string variable. Does it still make sense to turn it into a property?

rtitle: Yeah `googleProjectKey` is a string key for a Java system property. We're overloading the property a bit -- the property value with either contain a google project name, _or_ an error message if we encountered an error generating the project. 

My comment is just to remove the `sys.props.get(workspaceNamespaceKey) match …` and just do:
```
sys.props.get(googleProjectKey) match {
      case None => throw new RuntimeException("leonardo.googleProject system property is not set")
      case Some(msg) if msg.startsWith(gpallocErrorPrefix) => throw new RuntimeException(msg)
      case Some(googleProjectId) => // run the test
    }
```

Then in the `beforeAll` we put the error message in the `googleProjectKey` property value instead.

xuscrbw: Thanks for explaining, made the changes

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoSuite.scala
```diff
@@ -61,34 +75,50 @@ trait GPAllocUtils extends BillingFixtures with LeonardoTestUtils {
...
 
   /**
    * Unclaiming billing project claim by Hermione
    */
-  protected def unclaimProject(project: GoogleProject): IO[Unit] =
+  protected def unclaimProject(workspaceName: WorkspaceName): IO[Unit] =
     for {
+      _ <- IO(
+        Orchestration.workspaces.delete(workspaceName.namespace, workspaceName.name)(ronAuthToken)
+      )
```
rtitle: I'm wondering if we should do `.attempt` here (and log if it failed similar to the call to `releaseGPAllocProject` a few lines down). This way it does not fail the test in `afterAll` if cleanup fails, which can happen if some individual tests fail to delete stuff.

rtitle (approved):
This looks good to me if tests pass reliably. Maybe a multi-test would be a good idea

xuscrbw:
multi-test

PR #2298, opened 2021-09-22
Cromwell App. Swagger hosted via ingress, API needs manually `curl`ing [BW-816]

cjllanwarne:
Notes to help reviewers:

- The dockerfile changes mean that the specified version of the Cromwell app is included in the Leonardo `docker` image when it gets created.
- The `appType` in `kubernetesModels` allows us to specify the app via `appType` in the leonardo API
- I'm not sure whether the values in the `reference.conf` are unused copy/paste from other apps, or vital to the app operating successfully.
- `Config.scala` and `CromwellLocalAppConfig` represent reading in values from leo config. Same comment about whether or not they're useful...
- `installCromwellLocal` in `GKEInterpreter.scala` is where the magic happens. It's mostly copy/paste from other functions, so feel free to ask whether particular lines are needed.... it's quite possible that they are not!

Please co-review this PR alongside:
- https://github.com/broadinstitute/cromwhelm/pull/8


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
```diff
@@ -563,6 +564,9 @@ final class LeoAppServiceInterp[F[_]: Parallel](
             galaxyConfig.chart
           else lastUsed.chart
         }
+        // TODO fix this for the cromwell case (BW-867)
```
rtitle, line 567: OK to fix this in a followup PR, as long as it's relatively soon I think.

This method also hard-codes `galaxyConfig.namespaceNameSuffix` (L554) and `galaxyConfig.releaseNameSuffix` (L573) which should be generalized to support multiple app types.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -965,6 +974,62 @@ class GKEInterpreter[F[_]](
 
     } yield ()
 
+  private[util] def installCromwellApp(helmAuthContext: AuthContext,
+                                       appName: AppName,
+                                       release: Release,
+                                       cluster: KubernetesCluster,
+                                       nodepoolName: NodepoolName,
+                                       namespaceName: NamespaceName)(implicit ev: Ask[F, AppContext]): F[Unit] = {
+    val chart = config.cromwellAppConfig.chart
```
rtitle, line 984: Maybe add a comment to use the `chart` from the DB instead of config when BW-860 is fixed.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -965,6 +974,62 @@ class GKEInterpreter[F[_]](
...
+          false
+        )
+        .run(helmAuthContext)
+
+      // Currently we always retry.
+      // The main failure mode here is helm install, which does not have easily interpretable error codes
+      retryConfig = RetryPredicates.retryAllConfig
+      _ <- tracedRetryF(retryConfig)(
+        helmInstall,
+        s"helm install for app ${appName.value} in project ${cluster.googleProject.value}"
```
rtitle: Now that we have more app types, maybe we can log the app type as well as name for better debugging.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -1161,6 +1226,38 @@ class GKEInterpreter[F[_]](
...
+      raw"""ingress.annotations.nginx\.ingress\.kubernetes\.io/auth-tls-secret=${namespaceName.value}/ca-secret""",
+      raw"""ingress.port=8000""", // TODO: Move port to .Values.service.port in chart
+      raw"""ingress.path=${ingressPath}""",
+      raw"""ingress.hosts[0].host=${k8sProxyHost}""",
+      raw"""ingress.hosts[0].paths[0]=${ingressPath}${"(/|$)(.*)"}""",
+      raw"""ingress.tls[0].secretName=tls-secret""",
+      raw"""ingress.tls[0].hosts[0]=${k8sProxyHost}"""
+    )
+
+    List(
```
rtitle, line 1252: For the future: will Cromwell want the workspace/project/bucket injected as configs or EVs? For Galaxy we do that so it can read/write from the workspace data model/bucket. Something to think about.

Also for PD support there are probably some `persistence.` values to add, but that can be done later I think.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -1161,6 +1226,38 @@ class GKEInterpreter[F[_]](
...
+    // These nginx an ingress rules are condition.
+    // Some apps do not like behind behind a reverse proxy in this way, and require routing specified via this baseUrl
+    // The two methods are mutually exclusive
+    val ingress = List(
+      raw"""ingress.enabled=true""",
+      raw"""ingress.annotations.nginx\.ingress\.kubernetes\.io/proxy-redirect-from=https://${k8sProxyHost}""",
+      raw"""ingress.annotations.nginx\.ingress\.kubernetes\.io/proxy-redirect-to=${leoProxyhost}""",
+      raw"""ingress.annotations.nginx\.ingress\.kubernetes\.io/rewrite-target=/${rewriteTarget}""",
+      raw"""ingress.annotations.nginx\.ingress\.kubernetes\.io/auth-tls-secret=${namespaceName.value}/ca-secret""",
+      raw"""ingress.port=8000""", // TODO: Move port to .Values.service.port in chart
```
rtitle: Is this needed anymore since it looks like it uses `.Vaules.service.port` in https://github.com/broadinstitute/cromwhelm/pull/8/?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -1161,6 +1226,38 @@ class GKEInterpreter[F[_]](
...
+  private[util] def buildCromwellAppChartOverrideValuesString(appName: AppName,
+                                                              cluster: KubernetesCluster,
+                                                              nodepoolName: NodepoolName,
+                                                              namespaceName: NamespaceName): List[String] = {
+    val ingressPath = s"/proxy/google/v1/apps/${cluster.googleProject.value}/${appName.value}/cromwell-service"
+    val k8sProxyHost = kubernetesProxyHost(cluster, config.proxyConfig.proxyDomain).address
+    val leoProxyhost = config.proxyConfig.getProxyServerHostName
+
+    val rewriteTarget = "$2"
+    // These nginx an ingress rules are condition.
```
rtitle: This comment is probably not relevant for Cromwell.

rtitle (commented):
Some nits above but overall looks great! Excited to have this working.

Couple questions:

1. Is [BW-860](https://broadworkbench.atlassian.net/browse/BW-860) expected to come soon? This PR technically works as is but it's a bit ugly that the Leo DB is in an inconsistent state (e.g. Cromwell apps having Galaxy chart/release names).
2. Any unit testing we can add to this PR? Maybe look into adding a case to `GKEInterpreterSpec` to validate the Cromwell override values string.
3. Is there a ticket for adding an automation test similar to [CustomAppCreationSpec](https://github.com/DataBiosphere/leonardo/blob/develop/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/apps/CustomAppCreationSpec.scala) or [AppCreationSpec](https://github.com/DataBiosphere/leonardo/blob/develop/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/apps/AppCreationSpec.scala)? I'm probably okay with doing that in a separate PR as well, as long as it's tracked.

cjllanwarne:
Ticket to enhance test coverage: https://broadworkbench.atlassian.net/browse/BW-864

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
cjllanwarne: Almost certainly. I suspect we'll be amending this very soon (in https://broadworkbench.atlassian.net/browse/BW-863, for example)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
cjllanwarne: Used here: https://github.com/broadinstitute/cromwhelm/pull/8/files#diff-e86d1783dc09c19bf650c45181ec012b6ddfc3052bb96635777a3a56776c6e57R30

Although if I'm updating the helm chart anyway it might be better to just do this TODO now

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/Config.scala
```diff
@@ -592,6 +592,17 @@ object Config {
     )
   }
 
+  implicit private val cromwellAppConfigReader: ValueReader[CromwellAppConfig] = ValueReader.relative { config =>
```
Qi77Qi, line 595: would be nice if you could add it to [here](https://github.com/DataBiosphere/leonardo/blob/7e2f6d734d1205c723c47fa792ceaea1337caa75/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/ConfigReader.scala#L21). It uses pureconfig, so that you don't have to manually write the config reader

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -965,6 +974,62 @@ class GKEInterpreter[F[_]](
...
+      // Currently we always retry.
+      // The main failure mode here is helm install, which does not have easily interpretable error codes
+      retryConfig = RetryPredicates.retryAllConfig
+      _ <- tracedRetryF(retryConfig)(
+        helmInstall,
+        s"helm install for app ${appName.value} in project ${cluster.googleProject.value}"
+      ).compile.lastOrError
+
+      // Poll the app until it starts up
+      last <- streamFUntilDone(
```
Qi77Qi, line 1016: you could use [streamUntilDoneOrTimeout](https://github.com/broadinstitute/workbench-libs/blob/develop/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/package.scala#L164) instead, and then `adaptError` to proper error

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -965,6 +974,62 @@ class GKEInterpreter[F[_]](
 
     } yield ()
 
+  private[util] def installCromwellApp(helmAuthContext: AuthContext,
```
Qi77Qi, line 977: would be nice to make these `installX` methods duplicate less...but I guess that's sth leo team can work on

Qi77Qi (approved):
lgtm in general....more tests would be nice...since it's not used anywhere yet and assuming this will be iterated on actively, I guess it's ok for now

Qi77Qi:
oh, I think this needs some swagger updates as well? at least the app type enum in the createApp payload

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
cjllanwarne: Similar to your comment below about generalizing the `installX` methods, this is in analogy with what happens in those methods. FWIW there is a very clear error path if the service doesn't come up in time, tested thoroughly by my bad attempts to write the Cromwell app ingress rules properly...

cjllanwarne:
@rtitle -

1. Is BW-860 expected to come soon? This PR technically works as is but it's a bit ugly that the Leo DB is in an inconsistent state (e.g. Cromwell apps having Galaxy chart/release names).
    - Yes. I wanted to get this merged to allow all these little tasks to start happening in parallel, but the intention is definitely to get them started ASAP
2. Any unit testing we can add to this PR? Maybe look into adding a case to GKEInterpreterSpec to validate the Cromwell override values string.
    - I added a unit test for the values.  
3. Is there a ticket for adding an automation test similar to CustomAppCreationSpec or AppCreationSpec? I'm probably okay with doing that in a separate PR as well, as long as it's tracked.
    - On its way in BW-864. Same comment as for (1) w.r.t. parallelizing rather than waiting for everything to be done and submitting all at once, but this is the top of our backlog.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/Config.scala
```diff
@@ -754,6 +767,7 @@ object Config {
       org.broadinstitute.dsde.workbench.leonardo.http.ConfigReader.appConfig.terraAppSetupChart,
       gkeIngressConfig,
       gkeGalaxyAppConfig,
```
cahrens, line 769: Way out of scope for this merge request, but do folks know if there is a framework in Scala for doing plugins? Ideally there should not need to be code changes when a new app is registered.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
cahrens: I agree that seeing all the duplication is painful. I've had positive experiences in the past with using a plugin architecture (it was in Python) to allow registration of "apps" without having to add custom code for each one.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -1161,6 +1227,34 @@ class GKEInterpreter[F[_]](
     )
   }
 
+  private[util] def buildCromwellAppChartOverrideValuesString(appName: AppName,
+                                                              cluster: KubernetesCluster,
+                                                              nodepoolName: NodepoolName,
+                                                              namespaceName: NamespaceName): List[String] = {
+    val ingressPath = s"/proxy/google/v1/apps/${cluster.googleProject.value}/${appName.value}/cromwell-service"
```
cahrens, line 1234: Isn't "cromwell-service" the only thing in here that is specific to Cromwell? Can this code be generalized across apps?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/Config.scala
rtitle: That is the intent with the `CUSTOM` app type (which is what the UCSC genome browser uses): to register new apps without any Leo code changes at all. We decided not to go that route for Cromwell because we thought Cromwell has requirements that the custom app helm chart doesn't/won't support. I agree there are probably opportunities for code deduplication, but in general first class app integrations like Cromwell are probably rare compared with custom app integrations, which can be plugged in easily.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
rtitle: This method is specific to the cromwell helm chart.. it is probably similar to the other methods but I feel like it's good to have a dedicated `buildOverrideValues` method per helm chart we support.

(fwiw I think as tech debt Leo should move ingress setup outside the application charts and into its internal setup chart, which should remove the need for ingress duplication across all these methods.)

rtitle (approved):
Thanks for applying changes, looks good.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
rtitle: +1, the install methods have a bunch of boilerplate which could probably be deduplicated.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
cjllanwarne: Yeah, I suspect we'll need to add back in persistent disk setup, swagger overrides, etc, pretty quickly, even if today this is relatively uninspiring as a separate method

cjllanwarne:
jenkins retest

cjllanwarne:
jenkins retest

cahrens (approved):

PR #2315, opened 2021-10-01
Add northamerica-northeast2 to list of supported regions

wnojopra:
Recently the Toronto datacenter was announced by Google as the new northamerica-northeast2 region. 

Currently, I believe this and the terra-ui to be the only places we currently need to keep updated. But we should document if there are others. 

Please note that I did not test this change.

### DIFF
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -14,7 +14,9 @@ application {
 
 # Google Cloud dataproc configuration
 dataproc {
-  # add more regions as necessary
+  # Add more regions as necessary.
+  # This list should match the list of regions in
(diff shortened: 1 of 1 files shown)


rtitle (commented):
Looks good to me.. unfortunately however our CI does not like branch names with a `/`. I think it's because we make a docker tag with the branch name:
```
17:00:17 Error parsing reference: "gcr.io/broad-dsp-gcr-public/leonardo:wn/region" is not a valid repository/tag: invalid reference format
```
@wnojopra would you be able to make a PR from a different branch so tests run? (Sorry)

wnojopra:
Thanks @rtitle ! No problem. Opened #2319

PR #2319, opened 2021-10-04
[no ticket] Add northamerica-northeast2 to list of supported regions

wnojopra:
(From #2315 )

Recently the Toronto datacenter was announced by Google as the new northamerica-northeast2 region.

Currently, I believe this and the terra-ui to be the only places we currently need to keep updated. But we should document if there are others.

Please note that I did not test this change.


rtitle (approved):
LGTM if tests pass 👍

kyuksel (approved):

PR #2322, opened 2021-10-04
[IA-2886] Runtime creation timeout not being honored

rtitle:
https://broadworkbench.atlassian.net/browse/IA-2886

Fixed runtime monitoring timeout handling. We now have an `interruptAfter` field for tool monitoring so it does not retry beyond the intended timeout. Comments inline.

Tested in a fiab by forcing Jupyter to not start up, it seems to work as intended.

---

In all cases:

### DIFF
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -71,9 +71,10 @@ dataproc {
   monitor {
     initialDelay = 30 seconds
     # Defines polling for runtime status transitions. Used commonly for all statuses.
+    # Includes user script execution, if present.
     # Does not include tool checking (see below).
     pollStatus {
-      max-attempts = 80 # 15 seconds * 80 is 20 min
+      max-attempts = 120 # 15 seconds * 120 is 30 min
       interval = 15 seconds
     }
     # Defines a timeout per status transition. If a status is not listed there is no timeout.
@@ -115,9 +116,10 @@ gce {
   monitor {
     initialDelay = 20 seconds
     # Defines polling for runtime status transitions. Used commonly for all statuses.
+    # Includes user script execution, if present.
     # Does not include tool checking (see below).
     pollStatus {
-      max-attempts = 80 # 15 seconds * 80 is 20 min
+      max-attempts = 120 # 15 seconds * 120 is 30 min
       interval = 15 seconds
     }
     # Defines a timeout per status transition. If a status is not listed there is no timeout.
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -70,17 +70,27 @@ dataproc {
 
   monitor {
     initialDelay = 30 seconds
-    pollingInterval = 15 seconds
-    pollCheckMaxAttempts = 120 # 15 seconds * 120 is 30 min
-    checkToolsInterval = 8 seconds
-    checkToolsMaxAttempts = 75 # 8 seconds * 75 = 10 min
-    # Defines timeouts for cluster status transitions. If a status is not listed there is no timeout.
+    # Defines polling for runtime status transitions. Used commonly for all statuses.
+    # Does not include tool checking (see below).
+    pollStatus {
+      max-attempts = 80 # 15 seconds * 80 is 20 min
+      interval = 15 seconds
+    }
+    # Defines a timeout per status transition. If a status is not listed there is no timeout.
     # In the case of a Starting cluster, a timeout will transition it back to Stopped. Otherwise,
     # a timeout will transition it to Error status.
     statusTimeouts {
       creating = 30 minutes
       starting = 20 minutes
       deleting = 30 minutes
+    }
+    # Defines polling for tool checking. This is only done for Creating and Starting status transitions.
+    # Tool checking is not included in the statusTimeouts above, therefore it defines its own
+    # interruptAfter.
+    checkTools {
+      max-attempts = 75 # 8 seconds * 75 = 10 min
+      interval = 8 seconds
+      interruptAfter = 10 minutes
     }
   }
 }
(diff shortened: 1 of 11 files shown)


http/src/main/resources/reference.conf
```diff
@@ -70,18 +70,28 @@ dataproc {
 
   monitor {
     initialDelay = 30 seconds
-    pollingInterval = 15 seconds
-    pollCheckMaxAttempts = 120 # 15 seconds * 120 is 30 min
-    checkToolsInterval = 8 seconds
-    checkToolsMaxAttempts = 75 # 8 seconds * 75 = 10 min
-    # Defines timeouts for cluster status transitions. If a status is not listed there is no timeout.
+    # Defines polling for runtime status transitions. Used commonly for all statuses.
```
rtitle, line 73: Changed some of these fields to use `PollMonitor` and `InterruptablePollMonitor` case classes. Added some comments as well.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpDockerDAO.scala
```diff
@@ -210,7 +210,7 @@ object HttpDockerDAO {
     val res = for {
       splitted <- Either.catchNonFatal(s.split("="))
       first <- Either.catchNonFatal(splitted(0))
-      second <- Either.catchNonFatal(splitted(1))
+      second <- Either.catchNonFatal(if (splitted.length > 1) splitted(1) else "")
```
rtitle, line 213: Not related to this ticket, but a user found this so thought I'd fix. It's caused by EVs in the Docker image with a blank value (e.g. `FOO=`).

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BaseCloudServiceRuntimeMonitor.scala
```diff
@@ -479,7 +481,11 @@ abstract class BaseCloudServiceRuntimeMonitor[F[_]] {
           failedRuntime(
             monitorContext,
             runtimeAndRuntimeConfig,
-            RuntimeErrorDetails(s"${toolsStillNotAvailable} didn't start up properly", None, Some("tool_start_up")),
+            RuntimeErrorDetails(
+              s"${toolsStillNotAvailable.map(_.entryName).mkString(", ")} failed to start after ${monitorConfig.checkTools.interruptAfter.toMinutes} minutes.",
```
rtitle, line 485: Tried to improve this error message. It looks like this now:
image

cc @alessanrdo

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BaseCloudServiceRuntimeMonitor.scala
```diff
@@ -468,9 +468,11 @@ abstract class BaseCloudServiceRuntimeMonitor[F[_]] {
...
       }
       // wait for 10 minutes for tools to start up before time out.
-      availableTools <- streamFUntilDone(checkTools,
-                                         monitorConfig.checkToolsMaxAttempts,
-                                         monitorConfig.checkToolsInterval).compile.lastOrError
+      availableTools <- streamFUntilDone(
+        checkTools,
+        monitorConfig.checkTools.maxAttempts,
+        monitorConfig.checkTools.interval
+      ).interruptAfter(monitorConfig.checkTools.interruptAfter).compile.lastOrError
```
rtitle, line 475: This is where it interrupts tool polling (fs2 `interuptAfter` also requires `ConcurrentEffect` typeclass).

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BaseCloudServiceRuntimeMonitor.scala
Qi77Qi: so this is 10 min starts from when the VM becomes `Running` in GCP? do we need a bit more time just case if user's startup script takes a bit longer to run?

rtitle: I think if there is a user script, it waits for the user script to be done before moving on to `handleCheckTools`. See: https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/GceRuntimeMonitor.scala#L254. So user script execution would fall into the 20min bucket for VM creation.

Qi77Qi: ah right...forgot about that...so this means if user's startup script is stuck, user will still need to wait....think this is still improvement, but just trying to understand the impact

Qi77Qi: oh nvm, I think `checkAgain` has a timeout check as well...so this compliments that then

Qi77Qi (approved):

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BaseCloudServiceRuntimeMonitor.scala
rtitle: Yeah it's tricky. I think we have the following situation:

* If there is a startup script:
   - 20m timeout for GCP VM creation, docker image download, startup script
   - 10m timeout for Jupyter startup

* If there is no startup script:
   - 20m timeout for GCP VM creation
   - 10m timeout for docker image download, Jupyter startup

For the second case it might be good to put `docker image download` in the 20m bucket, and try to use the 10m bucket _only_ for Jupyter startup. wdyt?

Qi77Qi: it makes sense.. but when there's no startup script, how do you make sure `docker image download` happens before we check if tools are up?

rtitle: Yeah I'm not sure.. we'd need to design how that works. Maybe the init script could write a marker file to the staging bucket when `docker image download` is done.

Anyway maybe don't need to do that for this PR, but could ticket separately in case there are timeouts with custom images.

Qi77Qi: I wonder if it makes sense to adjust `10m` bucket timeout based on whether there's startup script...if without, we make it 20m or sth to give it more time to `docker pull`?

rtitle: Hmm that could work. Basically we try and allocate `10m` for "customization" (e.g. docker download and user script), and put that 10m in either bucket depending on whether there's a user script present. Will see how this looks..

rtitle: I feel like this idea makes the logic a bit more complicated -- e.g. we're trying to infer where docker image pull happens depending on other fields. For Dataproc it's part of `Creating` GCP status; for GCE it happens after VM is `Running` but before the user script which may or may not be present, etc.

Leaning toward leaving this PR as is and adding explicit tracking of image pull if we need to (in a separate PR).

Qi77Qi: sounds good...it might make sense to do what you mentioned in the previous comment about writing a marker file too

PR #2323, opened 2021-10-05
[No ticket] disable Rmd file syncing

gpcarr:
tested in my fiab
validated that for gce runtimes (both jupyter and rstudio) we set env var `IS_RSTUDIO_RUNTIME` to false and tested jupyter in dataproc as well

---

In all cases:

### DIFF
--- a/http/src/main/resources/init-resources/gce-init.sh
+++ b/http/src/main/resources/init-resources/gce-init.sh
@@ -105,11 +105,12 @@ export MEM_LIMIT=$(memLimit)
 export WELDER_MEM_LIMIT=$(welderMemLimit)
 export PROXY_SERVER_HOST_NAME=$(proxyServerHostName)
 export WELDER_ENABLED=$(welderEnabled)
-if [ ! -z "$RSTUDIO_DOCKER_IMAGE" ] ; then
-  export IS_RSTUDIO_RUNTIME="true"
-else
-  export IS_RSTUDIO_RUNTIME="false"
-fi
+export IS_RSTUDIO_RUNTIME="false" # TODO: update to commented out code once we release Rmd file syncing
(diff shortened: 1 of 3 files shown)


http/src/main/resources/init-resources/gce-init.sh
```diff
@@ -105,11 +105,12 @@ export MEM_LIMIT=$(memLimit)
 export WELDER_MEM_LIMIT=$(welderMemLimit)
 export PROXY_SERVER_HOST_NAME=$(proxyServerHostName)
 export WELDER_ENABLED=$(welderEnabled)
-if [ ! -z "$RSTUDIO_DOCKER_IMAGE" ] ; then
-  export IS_RSTUDIO_RUNTIME="true"
-else
-  export IS_RSTUDIO_RUNTIME="false"
-fi
+export IS_RSTUDIO_RUNTIME="false" # TODO: update to commented out code once we release Rmd file syncing
```
jdcanas, line 108: I'm assuming that since this is a VM-scoped env var set at start-up that it doesn't have any affect on Leonardo's labelling of the runtime as an RStudio runtime in the DB?

Qi77Qi: didn't realize there's labeling in DB for this as well..is that being used anywhere?

Qi77Qi (approved):

rtitle (approved):
lgtm

http/src/main/resources/init-resources/gce-init.sh
gpcarr: it doesn't affect the db - just double checked that an rstudio runtime still has the label `tool: RStudio`. not sure where we use this though

PR #2330, opened 2021-10-12
IA-2994 use built-in stopCluster

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-2994

Related PR https://github.com/broadinstitute/workbench-libs/pull/783

Dataproc started supporting "stop" a cluster a while ago. This means we can use sdk's stopCluster API directly to stop a cluster, which should save on dataproc premium when users intend to stop a cluster, because today, when user stops a cluster, we manually stops all instances of the cluster, but the cluster itself is still in `RUNNING` status.

However, during a patch update, it is still useful to not have a full cluster stop, hence I introduced a `isDataprocFullStop` flag to indicate whether we intend to also stop the cluster in addition to shutting down all VM instances.

Also made some optimizations in `DataprocMonitor` so that we're not querying instance's status all the time when not needed (for instance, when we create a cluster, if the cluster is still in `Creating`, we don't need to poll individual VMs at that point, but we poll all instances to get VM status regardless of cluster status today)

Separately upgrade http4s to 0.23.6, which has a retry bug fix.

Created https://broadworkbench.atlassian.net/browse/IA-2999 to reflect the cost reduction in UI

Manual tests:
* Create a dataproc cluster via local terra UI. Create/stop/start works as expected
* Add pre-emptibles, and start/stop cluster works as expected

---

In all cases:

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/runtimes/RuntimeDataprocSpec.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/runtimes/RuntimeDataprocSpec.scala
@@ -238,7 +238,13 @@ class RuntimeDataprocSpec
         _ <- IO(stopAndMonitorRuntime(runtime.googleProject, runtime.clusterName))
 
         // preemptibles should be removed in Dataproc
-        _ <- verifyDataproc(project, runtime.clusterName, dep.dataproc, 2, 0, RegionName("us-central1"))
+        _ <- verifyDataproc(project,
+                            runtime.clusterName,
+                            dep.dataproc,
+                            2,
+                            0,
+                            RegionName("us-central1"),
+                            RuntimeStatus.Stopped)
 
         // start the cluster
         _ <- IO(startAndMonitorRuntime(runtime.googleProject, runtime.clusterName))
@@ -283,12 +289,15 @@ class RuntimeDataprocSpec
     res.unsafeRunSync()(cats.effect.unsafe.IORuntime.global)
   }
 
-  private def verifyDataproc(project: GoogleProject,
-                             runtimeName: RuntimeName,
-                             dataproc: GoogleDataprocService[IO],
-                             expectedNumWorkers: Int,
-                             expectedPreemptibles: Int,
-                             expectedRegion: RegionName)(implicit httpClient: Client[IO]): IO[Unit] =
+  private def verifyDataproc(
+    project: GoogleProject,
+    runtimeName: RuntimeName,
+    dataproc: GoogleDataprocService[IO],
+    expectedNumWorkers: Int,
+    expectedPreemptibles: Int,
+    expectedRegion: RegionName,
+    expectedStatus: RuntimeStatus = RuntimeStatus.Running
+  )(implicit httpClient: Client[IO]): IO[Unit] =
     for {
       // check cluster status in Dataproc
       clusterOpt <- dataproc.getCluster(project, expectedRegion, DataprocClusterName(runtimeName.asString))
@@ -298,7 +307,7 @@ class RuntimeDataprocSpec
       status <- IO.fromOption(DataprocClusterStatus.withNameInsensitiveOption(cluster.getStatus.getState.name))(
         fail(s"Unknown Dataproc status ${cluster.getStatus.getState.name}")
       )
-      _ <- IO(List(DataprocClusterStatus.Running, DataprocClusterStatus.Stopped).contains(status) shouldBe true)
+      _ <- IO(status shouldBe expectedStatus)
 
       // check cluster instances in Dataproc
       instances = GoogleDataprocInterpreter.getAllInstanceNames(cluster)
--- a/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/runtimes/RuntimeDataprocSpec.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/runtimes/RuntimeDataprocSpec.scala
@@ -298,7 +298,7 @@ class RuntimeDataprocSpec
       status <- IO.fromOption(DataprocClusterStatus.withNameInsensitiveOption(cluster.getStatus.getState.name))(
         fail(s"Unknown Dataproc status ${cluster.getStatus.getState.name}")
       )
-      _ <- IO(status shouldBe DataprocClusterStatus.Running)
+      _ <- IO(List(DataprocClusterStatus.Running, DataprocClusterStatus.Stopped).contains(status) shouldBe true)
 
       // check cluster instances in Dataproc
       instances = GoogleDataprocInterpreter.getAllInstanceNames(cluster)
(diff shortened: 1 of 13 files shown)


automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/runtimes/RuntimeDataprocSpec.scala
```diff
@@ -298,7 +298,7 @@ class RuntimeDataprocSpec
       status <- IO.fromOption(DataprocClusterStatus.withNameInsensitiveOption(cluster.getStatus.getState.name))(
         fail(s"Unknown Dataproc status ${cluster.getStatus.getState.name}")
       )
-      _ <- IO(status shouldBe DataprocClusterStatus.Running)
+      _ <- IO(List(DataprocClusterStatus.Running, DataprocClusterStatus.Stopped).contains(status) shouldBe true)
```
Qi77Qi: previous dataproc cluster is always in `RUNNING` status even when stopped, but this is no longer true now that we're inducing "full" stop of a cluster

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BaseCloudServiceRuntimeMonitor.scala
```diff
@@ -214,7 +214,7 @@ abstract class BaseCloudServiceRuntimeMonitor[F[_]] {
   def checkAgain(
     monitorContext: MonitorContext,
     runtimeAndRuntimeConfig: RuntimeAndRuntimeConfig,
-    dataprocInstances: Set[DataprocInstance], //only applies to dataproc
+    fetchDataprocInstances: Option[F[Set[DataprocInstance]]], //only applies to dataproc
```
Qi77Qi, line 217: made this into an `Option[F[Set[DataprocInstance]]]` so that we don't have to eagerly fetch instances at call site

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/runtimes/RuntimeDataprocSpec.scala
rtitle: Can we pass in `expectedStatus` to this method so we can verify the specific status?

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dataprocModels.scala
```diff
@@ -103,4 +103,6 @@ object DataprocClusterStatus extends Enum[DataprocClusterStatus] {
   case object Unknown extends DataprocClusterStatus // The cluster state is unknown.
 
   case object Updating extends DataprocClusterStatus // The cluster is being updated. It continues to accept and process jobs.
+
+  case object Stopped extends DataprocClusterStatus // The cluster is being updated. It continues to accept and process jobs.
```
rtitle: Update comment

rtitle: Is there a `Stopping` / `Starting` from Dataproc that we should map to our status as well?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/DataprocRuntimeMonitor.scala
```diff
@@ -300,19 +323,25 @@ class DataprocRuntimeMonitor[F[_]: Parallel](
...
         } else {
-          checkAgain(
-            monitorContext,
-            runtimeAndRuntimeConfig,
-            instances,
-            Some(s"not all instances has been terminated yet. ${instances}")
-          )
+          for {
+            instances <- fetchInstances
+            r <- if (instances
```
rtitle, line 334: Looks like we handle `Dataproc status == Stopped OR all instance status == Stopped`. Wonder if we need the latter case still? Or should we be checking `Dataproc status == Stopped AND all instance status == Stopped`?

rtitle: Oh I see, we still handle both styles (looking at `isDataprocFullStop`).. this probably makes sense then.

project/Dependencies.scala
```diff
@@ -10,15 +10,15 @@ object Dependencies {
...
   val slickV = "3.3.3"
-  val http4sVersion = "0.23.5"
+  val http4sVersion = "0.23.6"
   val guavaV = "31.0.1-jre"
   val monocleV = "2.1.0"
   val opencensusV = "0.28.3"
 
 
-  private val workbenchLibsHash = "4cb37a0"
+  private val workbenchLibsHash = "0c87ddaa-SNAP"
```
rtitle: Update to non SNAP?

rtitle (approved):
Few nits.. overall looks good.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/DataprocRuntimeMonitor.scala
Qi77Qi: yea exactly....it's useful during patch where we don't need to fully stop a dataproc cluster

project/Dependencies.scala
Qi77Qi: yea, it's pending on broadinstitute/workbench-libs#783

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dataprocModels.scala
Qi77Qi: good point...I haven't noticed it during test....will double check

Qi77Qi:
randome fiab-start failure... jenkins retest

Qi77Qi:
another fiab start failure..jenkins retest

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreter.scala
```diff
@@ -341,15 +340,15 @@ class DataprocInterpreter[F[_]: Parallel](
...
+        Some(metadata),
+        params.isDataprocFullStop
       )
     } yield None
 
   override protected def startGoogleRuntime(params: StartGoogleRuntime)(
     implicit ev: Ask[F, AppContext]
   ): F[Unit] =
     for {
-      ctx <- ev.ask
```
rtitle, line 352: Just curious, does Dataproc `startCluster` handle both cases (e.g. if Dataproc `stop` was used or not)?

And does our `startup.sh` script get applied to the master node either way?

Qi77Qi: if cluster was `Running` previously, we can't `startCluster`; so at [here](https://github.com/broadinstitute/workbench-libs/blob/0c87ddaa20eba14eb0e6dfc13fa20c4326e51feb/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleDataprocInterpreter.scala#L222), it checks the status of the cluster before calling `startCluster`....in both cases, `startup.sh` will be applied when the master instances is started up

PR #2339, opened 2021-10-20
IA-2977 Refactoring some database fields to accommodate Azure VM 

Qi77Qi:
Part of https://broadworkbench.atlassian.net/browse/IA-2977

- Mostly look at `20211015_support_proxy_to_azure_vm.xml` to see what's being refactored. The majority of the diff is due to renaming `googleProject` to `cloudContext`, which is term used workspace manager. It's also documented in [here](https://docs.google.com/document/d/1qkte2Y1SB6RR34mFhVNkpqb377Ta5y44yeJh6GUS43g/edit#heading=h.lsjdzdaj8yo7). Not all intended refactoring is done yet to keep the size of the PR.
- Added `proxy/az/v1/runtimes` route, but this is not anywhere close to working yet. This will be actively worked on tgt with more refactoring in a follow up PR

Verified the migration works against dev DB

---

In all cases:

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodec.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodec.scala
@@ -53,7 +53,7 @@ object JsonCodec {
   val upsertEmptyLabelDecodingFailure = DecodingFailure("Label values are not allowed to be empty", List.empty)
 
   implicit val operationNameEncoder: Encoder[OperationName] = Encoder.encodeString.contramap(_.value)
-  implicit val googleIdEncoder: Encoder[GoogleId] = Encoder.encodeString.contramap(_.value)
+  implicit val googleIdEncoder: Encoder[ProxyHostName] = Encoder.encodeString.contramap(_.value)
   implicit val ipEncoder: Encoder[IP] = Encoder.encodeString.contramap(_.asString)
   implicit val gcsBucketNameEncoder: Encoder[GcsBucketName] = Encoder.encodeString.contramap(_.value)
   implicit val workbenchEmailEncoder: Encoder[WorkbenchEmail] = Encoder.encodeString.contramap(_.value)
@@ -77,6 +77,12 @@ object JsonCodec {
   implicit val diskStatusEncoder: Encoder[DiskStatus] = Encoder.encodeString.contramap(_.entryName)
   implicit val diskTypeEncoder: Encoder[DiskType] = Encoder.encodeString.contramap(_.asString)
   implicit val gpuTypeEncoder: Encoder[GpuType] = Encoder.encodeString.contramap(_.asString)
+  implicit val cloudProviderEncoder: Encoder[CloudProvider] = Encoder.encodeString.contramap(_.asString)
+
+  implicit val cloudContextEncoder: Encoder[CloudContext] = Encoder.forProduct2(
+    "cloudProvider",
+    "cloudResource"
+  )(x => (x.cloudProvider, x.asString))
 
   implicit val gpuConfigEncoder: Encoder[GpuConfig] = Encoder.forProduct2(
     "gpuType",
@@ -171,10 +177,11 @@ object JsonCodec {
   )(x => DefaultRuntimeLabels.unapply(x).get)
   implicit val asyncRuntimeFieldsEncoder: Encoder[AsyncRuntimeFields] =
     Encoder.forProduct4("googleId", "operationName", "stagingBucket", "hostIp")(x => AsyncRuntimeFields.unapply(x).get)
-  implicit val clusterProjectAndNameEncoder: Encoder[RuntimeProjectAndName] = Encoder.forProduct2(
-    "googleProject",
-    "clusterName"
-  )(x => RuntimeProjectAndName.unapply(x).get)
+  implicit val clusterProjectAndNameEncoder: Encoder[RuntimeProjectAndName] = Encoder.forProduct3(
+    "cloudProvider",
+    "cloudContext",
+    "runtimeName"
+  )(x => (x.cloudContext.cloudProvider, x.cloudContext.asString, x.runtimeName))
   implicit val runtimeErrorEncoder: Encoder[RuntimeError] = Encoder.forProduct4(
     "errorMessage",
     "errorCode",
@@ -244,7 +251,11 @@ object JsonCodec {
   implicit val gceInstanceStatusDecoder: Decoder[GceInstanceStatus] =
     Decoder.decodeString.emap(s => GceInstanceStatus.withNameInsensitiveOption(s).toRight(s"invalid gce status ${s}"))
   implicit val operationNameDecoder: Decoder[OperationName] = Decoder.decodeString.map(OperationName)
-  implicit val googleIdDecoder: Decoder[GoogleId] = Decoder.decodeString.map(GoogleId)
+  implicit val managedResourceGroupDecoder: Decoder[ManagedResourceGroup] =
+    Decoder.decodeString.map(ManagedResourceGroup)
+  implicit val cloudProviderDecoder: Decoder[CloudProvider] =
+    Decoder.decodeString.emap(s => CloudProvider.stringToCloudProvider.get(s).toRight(s"invalid cloud provider ${s}"))
+  implicit val googleIdDecoder: Decoder[ProxyHostName] = Decoder.decodeString.map(ProxyHostName)
   implicit val ipDecoder: Decoder[IP] = Decoder.decodeString.map(IP)
   implicit val containerImageDecoder: Decoder[ContainerImage] =
     Decoder.decodeString.emap(s => ContainerImage.fromImageUrl(s).toRight(s"invalid container image ${s}"))
@@ -274,6 +285,18 @@ object JsonCodec {
   implicit val runtimeImageTypeDecoder: Decoder[RuntimeImageType] = Decoder.decodeString.emap(s =>
     RuntimeImageType.stringToRuntimeImageType.get(s).toRight(s"invalid RuntimeImageType ${s}")
   )
+
+  implicit val cloudContextDecoder: Decoder[CloudContext] = Decoder.instance { x =>
+    for {
+      cloudProvider <- x.downField("cloudProvider").as[CloudProvider]
+      context <- cloudProvider match {
+        case CloudProvider.Gcp =>
+          x.downField("cloudResource").as[GoogleProject].map(p => CloudContext.Gcp(p))
+        case CloudProvider.Azure =>
+          x.downField("cloudResource").as[ManagedResourceGroup].map(p => CloudContext.Azure(p))
+      }
+    } yield context
+  }
 
   implicit val runtimeImageDecoder: Decoder[RuntimeImage] = Decoder.forProduct4(
     "imageType",
@@ -350,8 +373,28 @@ object JsonCodec {
                                        le.getOrElse(Map.empty))
   }
 
+  // Used only in pubsub messages
   implicit val clusterProjectAndNameDecoder: Decoder[RuntimeProjectAndName] =
-    Decoder.forProduct2("googleProject", "clusterName")(RuntimeProjectAndName.apply)
+    Decoder.instance(x =>
+      // support old messages where `googleProject` and `clusterName` are still being used. Remove once this code is released
+      for {
+        cloudProviderOpt <- x.downField("cloudProvider").as[Option[CloudProvider]]
+        cloudContext <- cloudProviderOpt match {
+          case Some(value) =>
+            value match {
+              case CloudProvider.Gcp =>
+                x.downField("cloudContext").as[GoogleProject].map(p => CloudContext.Gcp(p))
+              case CloudProvider.Azure =>
+                x.downField("cloudContext").as[ManagedResourceGroup].map(mrg => CloudContext.Azure(mrg))
+            }
+          case None =>
+            x.downField("googleProject")
+              .as[GoogleProject]
+              .map(p => CloudContext.Gcp(p)) //TODO: remove this case since this is just for backwards compatibility once this change is released to prod
+        }
+        runtimeName <- x.downField("clusterName").as[RuntimeName].orElse(x.downField("runtimeName").as[RuntimeName])
+      } yield RuntimeProjectAndName(cloudContext, runtimeName)
+    )
 
   implicit val asyncRuntimeFieldsDecoder: Decoder[AsyncRuntimeFields] =
     Decoder.forProduct4("googleId", "operationName", "stagingBucket", "hostIp")(AsyncRuntimeFields.apply)
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodec.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodec.scala
@@ -251,8 +251,8 @@ object JsonCodec {
   implicit val gceInstanceStatusDecoder: Decoder[GceInstanceStatus] =
     Decoder.decodeString.emap(s => GceInstanceStatus.withNameInsensitiveOption(s).toRight(s"invalid gce status ${s}"))
   implicit val operationNameDecoder: Decoder[OperationName] = Decoder.decodeString.map(OperationName)
-  implicit val managedResourceGroupDecoder: Decoder[ManagedResourceGroup] =
-    Decoder.decodeString.map(ManagedResourceGroup)
+  implicit val managedResourceGroupDecoder: Decoder[AzureCloudContext] =
+    Decoder.decodeString.emap(s => AzureCloudContext.fromString(s))
   implicit val cloudProviderDecoder: Decoder[CloudProvider] =
     Decoder.decodeString.emap(s => CloudProvider.stringToCloudProvider.get(s).toRight(s"invalid cloud provider ${s}"))
   implicit val googleIdDecoder: Decoder[ProxyHostName] = Decoder.decodeString.map(ProxyHostName)
@@ -293,7 +293,7 @@ object JsonCodec {
         case CloudProvider.Gcp =>
           x.downField("cloudResource").as[GoogleProject].map(p => CloudContext.Gcp(p))
         case CloudProvider.Azure =>
-          x.downField("cloudResource").as[ManagedResourceGroup].map(p => CloudContext.Azure(p))
+          x.downField("cloudResource").as[AzureCloudContext].map(p => CloudContext.Azure(p))
       }
     } yield context
   }
@@ -385,7 +385,7 @@ object JsonCodec {
               case CloudProvider.Gcp =>
                 x.downField("cloudContext").as[GoogleProject].map(p => CloudContext.Gcp(p))
               case CloudProvider.Azure =>
-                x.downField("cloudContext").as[ManagedResourceGroup].map(mrg => CloudContext.Azure(mrg))
+                x.downField("cloudContext").as[AzureCloudContext].map(mrg => CloudContext.Azure(mrg))
             }
           case None =>
             x.downField("googleProject")
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/azureModels.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/azureModels.scala
@@ -1,0 +1,3 @@
+package org.broadinstitute.dsde.workbench.leonardo
+
+final case class ManagedResourceGroup(value: String) extends AnyVal
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/azureModels.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/azureModels.scala
@@ -1,3 +1,25 @@
 package org.broadinstitute.dsde.workbench.leonardo
 
-final case class ManagedResourceGroup(value: String) extends AnyVal
+import cats.implicits._
+
+final case class TenantId(value: String) extends AnyVal
+final case class SubscriptionId(value: String) extends AnyVal
+final case class ManagedResourceGroupName(value: String) extends AnyVal
+
+final case class AzureCloudContext(tenantId: TenantId,
+                                   subscriptionId: SubscriptionId,
+                                   managedResourceGroupName: ManagedResourceGroupName) {
+  val asString = s"${tenantId.value}/${subscriptionId.value}/${managedResourceGroupName.value}"
+}
+
+object AzureCloudContext {
+  def fromString(s: String): Either[String, AzureCloudContext] = {
+    val res = for {
+      splitted <- Either.catchNonFatal(s.split("/"))
+      tenantId <- Either.catchNonFatal(splitted(0)).map(TenantId)
+      subscriptionId <- Either.catchNonFatal(splitted(1)).map(SubscriptionId)
+      mrgName <- Either.catchNonFatal(splitted(2)).map(ManagedResourceGroupName)
+    } yield AzureCloudContext(tenantId, subscriptionId, mrgName)
+    res.leftMap(t => s"Fail to decode $s as Azure Cloud Context due to ${t.getMessage}")
+  }
+}
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/models.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/models.scala
@@ -1,0 +1,41 @@
+package org.broadinstitute.dsde.workbench.leonardo
+
+import ca.mrvisser.sealerate
+import org.broadinstitute.dsde.workbench.model.google.GoogleProject
+
+final case class WorkspaceId(value: String) extends AnyVal
+
+final case class CloudContextDb(value: String) extends AnyVal
+
+sealed abstract class CloudContext extends Product with Serializable {
+  def asString: String
+  def asStringWithProvider: String
+  def cloudProvider: CloudProvider
+  def asCloudContextDb: CloudContextDb = CloudContextDb(asString)
+}
+object CloudContext {
+  final case class Gcp(value: GoogleProject) extends CloudContext {
+    override val asString = value.value
+    override val asStringWithProvider = s"Gcp/${value.value}"
+    override def cloudProvider: CloudProvider = CloudProvider.Gcp
+  }
+  final case class Azure(value: ManagedResourceGroup) extends CloudContext {
+    override val asString = value.value
+    override val asStringWithProvider = s"Azure/${value.value}"
+    override def cloudProvider: CloudProvider = CloudProvider.Azure
+  }
+}
+
+sealed abstract class CloudProvider extends Product with Serializable {
+  def asString: String
+}
+object CloudProvider {
+  final case object Gcp extends CloudProvider {
+    override val asString = "GCP"
+  }
+  final case object Azure extends CloudProvider {
+    override val asString = "AZURE"
+  }
+
+  val stringToCloudProvider = sealerate.values[CloudProvider].map(p => (p.asString, p)).toMap
+}
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/models.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/models.scala
@@ -19,9 +19,9 @@ object CloudContext {
     override val asStringWithProvider = s"Gcp/${value.value}"
     override def cloudProvider: CloudProvider = CloudProvider.Gcp
   }
-  final case class Azure(value: ManagedResourceGroup) extends CloudContext {
-    override val asString = value.value
-    override val asStringWithProvider = s"Azure/${value.value}"
+  final case class Azure(value: AzureCloudContext) extends CloudContext {
+    override val asString = value.asString
+    override val asStringWithProvider = s"Azure/${value.asString}"
     override def cloudProvider: CloudProvider = CloudProvider.Azure
   }
 }
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/runtimeModels.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/runtimeModels.scala
@@ -22,7 +22,7 @@ import scala.collection.immutable
 final case class Runtime(id: Long,
                          samResource: RuntimeSamResourceId,
                          runtimeName: RuntimeName,
-                         googleProject: GoogleProject,
+                         cloudContext: CloudContext,
                          serviceAccount: WorkbenchEmail,
                          asyncRuntimeFields: Option[AsyncRuntimeFields],
                          auditInfo: AuditInfo,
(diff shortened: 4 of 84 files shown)


Qi77Qi:
jenkins retest

Qi77Qi:
jenkins retest

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/models.scala
```diff
@@ -32,3 +32,7 @@ object AppContext {
     )
 
 }
+
+final case class WorkspaceId(value: String) extends AnyVal
+
+final case class CloudContext(value: String) extends AnyVal
```
jdcanas: Almost feels like this should be slightly more intelligent than a simple value class. It seems a little pointless, if the intent is to wrap a GoogleProject AND an Azure string identifier, to represent them with more-or-less a string wrapper.

Namely, I think the data structure should express if its a googleProject or [meaningfulAzureIdentifierName] to the consumer at the call site.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/PersistentDiskComponent.scala
```diff
@@ -1,24 +1,22 @@
...
+import java.time.Instant
 import scala.concurrent.ExecutionContext
 
 final case class PersistentDiskRecord(id: DiskId,
-                                      googleProject: GoogleProject,
+                                      cloudContext: CloudContext,
                                       zone: ZoneName,
                                       name: DiskName,
-                                      googleId: Option[GoogleId],
+                                      googleId: Option[ProxyHostName],
```
jdcanas: should this still be called googleId?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/PersistentDiskComponent.scala
```diff
@@ -137,15 +135,17 @@ object persistentDiskQuery {
...
       .filter(_.name === name)
       .filter(_.destroyedDate === dummyDate)
 
-  private[db] def findByNameQuery(googleProject: GoogleProject, name: DiskName) =
+  private[db] def findByNameQuery(cloudContext: CloudContext, name: DiskName) =
     tableQuery
-      .filter(_.googleProject === googleProject)
+      .filter(
+        _.cloudContext === cloudContext
+      ) //TODO: rename googleProject in PD table to CloudContext
```
jdcanas: this TODO seems done

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/DiskRoutes.scala
```diff
@@ -194,9 +197,10 @@ object DiskRoutes {
     } yield UpdateDiskRequest(l.getOrElse(Map.empty), us)
   }
 
-  implicit val getDiskResponseEncoder: Encoder[GetPersistentDiskResponse] = Encoder.forProduct12(
+  implicit val getDiskResponseEncoder: Encoder[GetPersistentDiskResponse] = Encoder.forProduct13(
     "id",
     "googleProject",
+    "cloudContext",
```
jdcanas, line 203: I think there's a bug here based on how the signature of `GetPersistentDiskResponse` changes here (the number of fields don't seem to change in the linked signature but the for product number does?)

https://github.com/DataBiosphere/leonardo/pull/2339/files#diff-69288d2ed640980de319fedf8f45eb9f03c794e90cda4239dac24cb1ebe0e9d0R40

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/models.scala
Qi77Qi: yea, I'm imagining we'll have such an ADT you're talking about in addition to this simple value wrapper...this simple value structure is still needed becuz we need to serielize/deserielize it with DB...from `CLUSTER` table, we can't really tell which context it is when reading from DB.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/PersistentDiskComponent.scala
Qi77Qi: probably not...there're probably still cases here and there that Type and name don't exactly match, but I think we can fix over time....I'll fix this one

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/DiskRoutes.scala
Qi77Qi: yea...for backwards compatiblity, both `cloudContext` and `googleProject` are populated with `cloudContext` value in the actual response json...but the model only has one field, which is `cloudContext`...

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/PersistentDiskComponent.scala
Qi77Qi: after a closer look..it seems like this `googleId` isn't really used...so I'm going to attempt to remove it

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20211015_support_proxy_to_azure_vm.xml
```diff
@@ -0,0 +1,57 @@
...
+<databaseChangeLog logicalFilePath="leonardo" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+
+<changeSet logicalFilePath="leonardo" author="qi" id="support_proxy_to_azure_vm">
+    <addColumn tableName="CLUSTER">
+        <column name="workspaceId" type="varchar(254)">
+            <constraints nullable="true"/>
+        </column>
+    </addColumn>
+
+    <dropNotNullConstraint columnName="runtimeConfigId" tableName="CLUSTER" columnDataType="bigint(20)"/>
```
rtitle: Curious why we need to drop the not-null and foreign key on `runtimeConfigId`?

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20211015_support_proxy_to_azure_vm.xml
```diff
@@ -0,0 +1,57 @@
...
+                  columnDataType="varchar(255)"
+                  oldColumnName="googleId"
+                  tableName="CLUSTER"/>
+
+    <renameColumn newColumnName="runtimeName"
+                  columnDataType="varchar(255)"
+                  oldColumnName="clusterName"
+                  tableName="CLUSTER"/>
+
+    <renameColumn newColumnName="cloudContext"
```
rtitle: There is an index on `(googleProject, runtimeName, destroyedDate)`… will the liquibase `renameColumn` update the index automatically w/ the new names?

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20211015_support_proxy_to_azure_vm.xml
```diff
@@ -0,0 +1,57 @@
...
+                  columnDataType="varchar(254)"
+                  oldColumnName="googleProject"
+                  tableName="PERSISTENT_DISK"/>
+
+    <dropForeignKeyConstraint baseTableName="CLUSTER" constraintName="FK_CLUSTER_RUNTIME_CONFIG_ID"/>
+
+    <createTable tableName="WSM_TO_CLUSTER">
```
rtitle: Maybe `CLUSTER_WORKSPACE` would fit our naming pattern better?

Similar to:
```
MySQL [leonardo]> DESCRIBE CLUSTER_IMAGE;
+---------------+---------------+------+-----+----------------------+-------+
| Field         | Type          | Null | Key | Default              | Extra |
+---------------+---------------+------+-----+----------------------+-------+
| clusterId     | bigint(20)    | NO   | MUL | NULL                 |       |
| imageType     | varchar(254)  | YES  |     | NULL                 |       |
| imageUrl      | varchar(1024) | YES  |     | NULL                 |       |
| timestamp     | timestamp(6)  | NO   |     | CURRENT_TIMESTAMP(6) |       |
| homeDirectory | varchar(254)  | YES  |     | NULL                 |       |
+---------------+---------------+------+-----+----------------------+-------+
5 rows in set (0.037 sec)
```

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20211015_support_proxy_to_azure_vm.xml
```diff
@@ -0,0 +1,57 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="leonardo" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
```
rtitle: Would be good to test this on some real dev snapshots :)

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20211015_support_proxy_to_azure_vm.xml
Qi77Qi: ah good call....didn't notice that..will check

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20211015_support_proxy_to_azure_vm.xml
Qi77Qi: oh, so I initially wasn't planning on refactoring runtimeConfig table to account for Azure data as well...forgot to remove this after our convo last time

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20211015_support_proxy_to_azure_vm.xml
Qi77Qi: yea..I guess I can try running a local leo that points to a dev snapshots....lmk if you have a better way to do this

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20211015_support_proxy_to_azure_vm.xml
rtitle: Thinking more, maybe `RUNTIME_CONTROLLED_RESOURCE`?

Idk, that's what I put in this doc though: https://docs.google.com/document/d/1aZPZkzIqJ5P-HBSxcu65uwRPUaBRKodqdZ59W9ayw_I/edit#heading=h.i8n1rfgcgf3o

Qi77Qi: yea think I like `RUNTIME_CONTROLLED_RESOURCE`

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20211015_support_proxy_to_azure_vm.xml
rtitle: Yeah that's what I've done in the past too: cloned the dev DB here: https://console.cloud.google.com/sql/instances?project=broad-dsde-dev and pointed a Leo to it

Qi77Qi:
weird fiab error...jenkins retest

Qi77Qi:
```
20:31:48 [ERROR] Ran out of retries without successful config render for thurloe.
```

jenkins retest

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20211015_support_proxy_to_azure_vm.xml
```diff
@@ -0,0 +1,75 @@
...
+        </column>
+    </addColumn>
+
+    <addColumn tableName="PERSISTENT_DISK">
+        <column name="cloudProvider" type="ENUM('GCP', 'AZURE')" defaultValue="GCP">
+            <constraints nullable="false"/>
+        </column>
+    </addColumn>
+
+    <dropForeignKeyConstraint baseTableName="CLUSTER" constraintName="FK_CLUSTER_RUNTIME_CONFIG_ID"/>
```
rtitle: Do we need to drop `FK_CLUSTER_RUNTIME_CONFIG_ID`?

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20211015_support_proxy_to_azure_vm.xml
```diff
@@ -0,0 +1,75 @@
...
+                  columnDataType="varchar(255)"
+                  oldColumnName="googleId"
+                  tableName="CLUSTER"/>
+
+    <renameColumn newColumnName="runtimeName"
+                  columnDataType="varchar(255)"
+                  oldColumnName="clusterName"
+                  tableName="CLUSTER"/>
+
+    <renameColumn newColumnName="cloudContext"
```
rtitle: Have been thinking… the managed resource group by itself does not fully define the cloud context. The Azure cloud context is really the combination of `(mrg, subscription, tenant)`. Those 3 pieces of information are needed to make cloud calls in a user's cloud context.

By contrast for GCP only the `googleProject` is needed because the organization is implied (firecloud.org) and the billing account is contained within the googleProject.

So I think it's fine to have `cloudContext: varchar` in the DB tables for uniqueness checks… just noting that to get the _actual_ Azure cloud context (those 3 fields) Leo will have to make a [query](https://workspace.dsde-dev.broadinstitute.org/swagger-ui.html#/Workspace/getWorkspace) to WorkspaceManager.

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/models.scala
```diff
@@ -32,3 +34,40 @@ object AppContext {
...
+}
+object CloudContext {
+  final case class Gcp(value: GoogleProject) extends CloudContext {
+    override val asString = value.value
+    override val asStringWithProvider = s"GCP/${value.value}"
+    override def cloudProvider: CloudProvider = CloudProvider.Gcp
+  }
+  final case class Azure(value: ManagedResourceGroup) extends CloudContext {
+    override val asString = value.value
+    override val asStringWithProvider = s"AZURE/${value.value}"
```
rtitle: Nit: maybe `Azure` capitalization if this string will be user-facing.

Qi77Qi: this string is mostly used in exception messages...so it might bubble up to user...

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20211015_support_proxy_to_azure_vm.xml
```diff
@@ -0,0 +1,75 @@
...
+        <column name="resourceType" type="VARCHAR(254)">
+            <constraints nullable="false"/>
+        </column>
+    </createTable>
+
+    <createIndex indexName="FK_RUNTIME_CONTROLLED_RESOURCE" tableName="RUNTIME_CONTROLLED_RESOURCE">
+        <column name="runtimeId"/>
+    </createIndex>
+    <addForeignKeyConstraint baseColumnNames="runtimeId" baseTableName="RUNTIME_CONTROLLED_RESOURCE" constraintName="FK_RUNTIME_CONTROLLED_RESOURCE" deferrable="false" initiallyDeferred="false" onDelete="NO ACTION" onUpdate="NO ACTION" referencedColumnNames="id" referencedTableName="CLUSTER"/>
+    <addUniqueConstraint columnNames="runtimeId, workspaceId, resourceId" constraintName="IDX_RUNTIME_CONTROLLED_RESOURCE_UNIQUE" tableName="RUNTIME_CONTROLLED_RESOURCE"/>
```
rtitle: Should this be unique on `(runtimeId, workspaceId, resourceId, resourceType)`?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/package.scala
```diff
@@ -22,7 +22,9 @@ package object api {
     Marshaller(implicit ec => (x => m(x.unsafeToFuture()(cats.effect.unsafe.IORuntime.global))))
 
   val googleProjectSegment = Segment.map(GoogleProject)
+  val managedResourceGroupSegment = Segment.map(ManagedResourceGroup)
```
rtitle: Not sure we need `managedResourceGroupSegment` -- were we going to put `workspaceId` in the v2 routes?

Qi77Qi: ah yea..forgot to update this after that discussion

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/ProxyRoutes.scala
```diff
@@ -52,6 +51,18 @@ class ProxyRoutes(proxyService: ProxyService, corsSupport: CorsSupport, refererC
                         }
                       }
                   }
+                } ~ pathPrefix("v2" / "runtimes" / "azure") {
+                  pathPrefix(managedResourceGroupSegment / runtimeNameSegment) { (mrg, runtimeName) =>
```
rtitle: I didn't fully update the [doc](https://docs.google.com/document/d/1aZPZkzIqJ5P-HBSxcu65uwRPUaBRKodqdZ59W9ayw_I/edit#), but perhaps we want:

```
/proxy/v2/runtimes/{workspaceId}/gcp/{name}
/proxy/v2/runtimes/{workspaceId}/azure/{name}
```
similar to runtimes routes.

Or maybe leave out proxy route changes from this PR since it's descoped from the demo?

rtitle (commented):
The refactoring looks good.. couple minor comments.

One comment above that `managedResourceGroup: String` does not fully represent the Azure cloud context (you need tenant and subscription too). However not sure if Leo really needs to store this information since it can look it up from WSM when needed. Anyway I think this is a good first pass and we can iterate if we need to change more fields once we start adding the new routes/WSM calls.

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20211015_support_proxy_to_azure_vm.xml
```diff
@@ -0,0 +1,75 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="leonardo" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+
+<changeSet logicalFilePath="leonardo" author="qi" id="support_proxy_to_azure_vm">
+    <addColumn tableName="CLUSTER">
```
kyuksel: Would it be good to rename the table `CLUSTER` since everywhere else related we've shifted to using the term `runtime`?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/ProxyRoutes.scala
Qi77Qi: I updated the route to `/proxy/v2/runtimes/{workspaceId}/azure/{name}`, but left it as not implemented

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20211015_support_proxy_to_azure_vm.xml
Qi77Qi: definitely..probably not for this PR tho since I'm just trying to wrap this up at this point... we won't continue this work until later after we feel good about demo

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20211015_support_proxy_to_azure_vm.xml
Qi77Qi: oops...this was leftoever of when I thought we need to make `runtimeConfigId` nullable...thanks for the catch!

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20211015_support_proxy_to_azure_vm.xml
Qi77Qi: talked to Rob over slack...what we'll do for now is having `mrg,subscription,tenant` in the case of Azure...if this turns out to be not easy to work with down the line we can revisit this

jdcanas (approved):
Looks good as a first pass at a leo cloud context. I imagine the 'cloudResource' field may need to be a more complex object in the future, otherwise i'd call it 'cloudResourceName'.  

Seems like one of those 'trust the tests' changes. Might run a multi-test too.

Qi77Qi:
@jdcanas yea, we'll start with `mrg,subscription,tenant` for `cloudContext` field for azure. Altho there're also some conversations about relationships between workspaces and subscriptions etc, which might change. But I think having it as `mrg,subscription,tenant` is a good first pass. This will be the combo we need for retrieving an azure resource

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodec.scala
```diff
@@ -275,6 +286,18 @@ object JsonCodec {
...
   )
 
+  implicit val cloudContextDecoder: Decoder[CloudContext] = Decoder.instance { x =>
+    for {
+      cloudProvider <- x.downField("cloudProvider").as[CloudProvider]
+      context <- cloudProvider match {
+        case CloudProvider.Gcp =>
+          x.downField("cloudResource").as[GoogleProject].map(p => CloudContext.Gcp(p))
+        case CloudProvider.Azure =>
+          x.downField("cloudResource").as[ManagedResourceGroup].map(p => CloudContext.Azure(p))
```
rtitle: Are we still thinking of embedding `"tenantId,subscriptionId,managedResourceGroupName"` as the Azure cloud context? I'm wondering If the decoder should split that string into separate fields..

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/models.scala
```diff
@@ -0,0 +1,41 @@
...
+  def cloudProvider: CloudProvider
+  def asCloudContextDb: CloudContextDb = CloudContextDb(asString)
+}
+object CloudContext {
+  final case class Gcp(value: GoogleProject) extends CloudContext {
+    override val asString = value.value
+    override val asStringWithProvider = s"Gcp/${value.value}"
+    override def cloudProvider: CloudProvider = CloudProvider.Gcp
+  }
+  final case class Azure(value: ManagedResourceGroup) extends CloudContext {
```
rtitle: Similar to above comment, wondering if `CloudContext.Azure` should have `(tenantId: TenantId, subscriptionId: SubscriptionId, mrg: ManagedResourceGroupName)`

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/azureModels.scala
```diff
@@ -0,0 +1,3 @@
+package org.broadinstitute.dsde.workbench.leonardo
+
+final case class ManagedResourceGroup(value: String) extends AnyVal
```
rtitle: Maybe `ManagedResourceGroupName`?

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/runtimeModels.scala
```diff
@@ -64,7 +64,7 @@ object Runtime {
       .getOrElse(JupyterService)
 
     new URL(
-      urlBase + googleProject.value + "/" + runtimeName.asString + "/" + tool.proxySegment
+      urlBase + cloudContext.asString + "/" + runtimeName.asString + "/" + tool.proxySegment
```
rtitle: I think this only works for the Google case, right? Just wondering if it should fail if it's an Azure cloud context for now.

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20211015_support_proxy_to_azure_vm.xml
```diff
@@ -0,0 +1,74 @@
...
+        </column>
+    </addColumn>
+
+    <addColumn tableName="CLUSTER">
+        <column name="cloudProvider" type="ENUM('GCP', 'AZURE')" defaultValue="GCP">
+            <constraints nullable="false"/>
+        </column>
+    </addColumn>
+
+    <addColumn tableName="PERSISTENT_DISK">
```
rtitle: Maybe would be nice to break up updates to different tables to different change sets? Just so if 1 fails it doesn't leave a table partially updated..

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20211015_support_proxy_to_azure_vm.xml
```diff
@@ -0,0 +1,74 @@
...
+                  oldColumnName="googleProject"
+                  tableName="CLUSTER"/>
+
+    <renameColumn newColumnName="cloudContext"
+                  columnDataType="varchar(254)"
+                  oldColumnName="googleProject"
+                  tableName="PERSISTENT_DISK"/>
+
+    <addUniqueConstraint
+            columnNames="cloudContext,runtimeName,destroyedDate"
```
rtitle: Should this be `cloudContext,cloudProvider,runtimeName,destroyedDate`?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
```diff
@@ -75,8 +77,6 @@ class ClusterTable(tag: Tag) extends Table[ClusterRecord](tag, "CLUSTER") {
   def customClusterEnvironmentVariables = column[Option[Map[String, String]]]("customClusterEnvironmentVariables")
   def deletedFrom = column[Option[String]]("deletedFrom")
 
-  def uniqueKey = index("IDX_CLUSTER_UNIQUE", (googleProject, clusterName, destroyedDate), unique = true)
```
rtitle, line 78: Can we keep this `def`? I'm not sure what it actually does though..

rtitle:
The refactoring looks good overall. Main comment is whether we can model `CloudContext.Azure` as containing tenantId/subscriptionId/managedResourceGroupName. We will need those 3 fields in order to make Azure cloud calls using the Java client.

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodec.scala
Qi77Qi: ah yea...forgot this needs to be refactored for that

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/runtimeModels.scala
Qi77Qi: will do

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
Qi77Qi: based on slick doc in source code, I think this just adds the unique index, which is duplicated by [this](https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20171010_cluster_unique_index.xml#L5)....

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20211015_support_proxy_to_azure_vm.xml
Qi77Qi: I separated PD update and creating the new wsm resource table into a separate xml file.

Qi77Qi:
@rtitle  thanks for re-review...addressed comments in [this](https://github.com/DataBiosphere/leonardo/pull/2339/commits/fbd13eb8fabc5c0a83d9b8b2f5036ead35cbc471)

rtitle (approved):
Thanks for addressing.. lgtm if tests pass

Qi77Qi:
```
message: Failed to create runtime terra-quality-cf3a5b27/automation-test-afy8ukx7z, status: 500 Internal Server Error body: {"source":"leonardo","message":"HTTP connection closed: https://sam-fiab.dsde-qa.broadinstitute.org:443","statusCode":500,"exceptionClass":"class java.net.SocketException","traceId":null}
```

jenkins retest

Qi77Qi:
multi-test

Qi77Qi:
multi-test

Qi77Qi:
one failure from multi-test that saw happened in dev branch today as well...

multi-test

Qi77Qi:
multi-test

Qi77Qi:
multi-test

PR #2340, opened 2021-10-20
[IA-2812] [IA-2958] Add labels to list disks and list apps response

gpcarr:
This PR adds a query parameter to the `listDisks` and `listApps` endpoints called `includeLabels` which is a comma separated list of label keys. Those key, value pairs will be returned as a `LabelMap` in the response for each endpoint. We want to be able to use labels in the UI for disks and apps

Tested in a fiab with the following test cases for both disks and apps:

```
disk1

{
  "size": 50,
  "diskType": "pd-standard",
  "blockSize": 4096,
  "labels": {
  	"key1": "val1",
  	"key2": "val2",
  	"key3": "val3"
    }
}

disk2

{
  "size": 50,
  "diskType": "pd-standard",
  "blockSize": 4096,
  "labels": {
  	"key1": "val1",
  	"key2": "val2"
    }
}

disk3

{
  "size": 50,
  "diskType": "pd-standard",
  "blockSize": 4096,
  "labels": {
  	"key1": "val1"
    }
}


test cases

1)
_labels=key1=val1,key2=val2
includeLabels=key1,key2,key5
should return disk1 and disk2 with labels key1=val1 and key2=val2

2) (curl)
_labels=key1=val1
&key2=val2
includeLabels=key1,key5
should return disk1 and disk2 with labels key1=val1

3)
includeLabels=key1,key2,key3
should return disk1 disk2 and disk3 with labels key1=val1, key2=val2, key3=val3 for disk1 (1 and 2 for disk2, 1 for disk3)

4)
includeLabels=key1,key2
should return disk1 disk2 disk3 with labels key1=val1, key2=val2

5)
should return disk1 disk2 disk3 with no labels TODO: this returns labels: {} - which does not mean there are no labels on the disk but rather includeLabels was empty. i think this behavior makes sense

now delete disk1

1)
_labels=key1=val1
&key2=val2
includeLabels=key1,key5
includeDeleted = true
should return disk1 and disk2 with labels key1=val1
```

---

In all cases:


http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -461,6 +461,15 @@ paths:
           schema:
             type: boolean
             default: false
```
Qi77Qi, line 463: think the `listRuntime` and `listDisk` responses need to be udpated as well

core/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/AppRoutesTestJsonCodec.scala
```diff
@@ -45,12 +45,13 @@ object AppRoutesTestJsonCodec {
...
+    Decoder.forProduct9("googleProject",
                         "kubernetesRuntimeConfig",
                         "errors",
                         "status",
                         "proxyUrls",
                         "appName",
                         "diskName",
-                        "auditInfo")(ListAppResponse.apply)
+                        "auditInfo",
+                        "includeLabels")(ListAppResponse.apply)
```
Qi77Qi: think you're just returning them as `labels` in response

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/PackageSpec.scala
```diff
@@ -16,4 +16,11 @@ class PackageSpec extends AnyFlatSpec with Matchers {
     val duplicateLabel = Map("_labels" -> "foo=bar,foo=biz")
     processLabelMap(duplicateLabel) shouldBe (Right(Map("foo" -> "biz")))
   }
+
+  it should "process labels to return properly" in {
+    val input1 = Map("includeLabels" -> "foo,bar")
+    processLabelsToReturn(input1, None) shouldBe Right(List("foo", "bar"))
+
+    // TODO: what would a bad request look like?
```
Qi77Qi: one example, val input1 = Map("includeLabels" -> "foo;bar")

gpcarr: that would process as `["foo;bar"]` and not fail - i think i need more logic in `processLabelsToReturn` so that we catch that kind of exception. what else would be considered an exception? happy to jump on a call for this if it's easier

http/src/main/resources/swagger/api-docs.yaml
gpcarr: i updated `listDisks` that's this section of code. I didn't change anything in `listRuntime` - it already returns labels. i updated `listApps`

core/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/AppRoutesTestJsonCodec.scala
gpcarr: ah good catch

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/PackageSpec.scala
Qi77Qi: it's probably better to use regex instead of just simple `split` then...but since the API is most consumed internally....so I guess it's probably not a big deal if we're a bit relaxed here

http/src/main/resources/swagger/api-docs.yaml
Qi77Qi: sry `listApp`...I only see requests are udpated in diff...which line is for updating response for listDisk or listApp?

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/PackageSpec.scala
gpcarr: i was thinking to use regex as well - i'll update

http/src/main/resources/swagger/api-docs.yaml
gpcarr: OH i see what you mean - i still need to update the responses. will do!

Qi77Qi (approved):

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -461,6 +461,15 @@ paths:
...
+        - in: query
```
rtitle, line 464: Can you update the swagger patch version at the top of this file?

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -516,6 +525,15 @@ paths:
...
+          name: includeLabels
```
rtitle, line 529: Wonder if it makes sense to update the other `_labels` query string param to `filterLabels` to make it more clear the semantics of that param vs `includeLabels`.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/package.scala
```diff
@@ -42,12 +45,48 @@ package object service {
...
+   * Top-level query string parameter for apps and disks: GET /api/apps?includeLabels=foo,bar
+   * where foo,bar are label keys for which this endpoint returns the LabelMap for each key
+   *
+   * @param params raw query string params
+   */
+  private[service] def processLabelsToReturn(
+    labelsToReturn: String
+  ): Either[ParseLabelsException, List[String]] =
+    Either
+      .catchNonFatal(labelsToReturn.split(',').toList.filter(l => !l.isEmpty))
```
rtitle, line 85: Was wondering if this might be URL-encoded but I think it gets decoded prior to this method. Might be good to test this via Swagger just to be sure but it's probably ok.

rtitle (approved):
Looks good!

http/src/main/resources/swagger/api-docs.yaml
gpcarr: updated!

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -498,7 +507,7 @@ paths:
           schema:
             type: string
         - in: query
-          name: _labels
+          name: filterLabels
```
Qi77Qi: before we make this change..can we check with AoU and bioconductor team that they're not using the `_labels`...

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/package.scala
gpcarr: tested it in postman and it looks good!

Screen Shot 2021-10-22 at 2 37 26 PM

http/src/main/resources/swagger/api-docs.yaml
gpcarr: decided to punt on changing to `filterLabels` since bioconductor hasn't responded and aou requested that we support both. i want to get this PR merged so i can test my UI changes against dev

gpcarr:
jenkins retest - fiab start failure

gpcarr:
jenkins retest

PR #2341, opened 2021-10-20
Update proxyUrls returned for Cromwell App to link to Cromwell Swagger page [BW-869]

cjllanwarne:
This PR updates the response from `getApp` to reference the correct path to the Cromwell swagger page, with correct parameters to set up the swagger service with the correct yaml.

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
@@ -3,22 +3,13 @@ package org.broadinstitute.dsde.workbench.leonardo
 import java.net.URL
 import java.time.Instant
 import java.util.UUID
+
 import ca.mrvisser.sealerate
 import org.broadinstitute.dsde.workbench.leonardo.SamResourceId._
 import org.broadinstitute.dsde.workbench.google2.GKEModels.{KubernetesClusterId, KubernetesClusterName, NodepoolName}
-import org.broadinstitute.dsde.workbench.google2.KubernetesSerializableName.{
-  NamespaceName,
-  ServiceAccountName,
-  ServiceName
-}
-import org.broadinstitute.dsde.workbench.google2.{
-  KubernetesName,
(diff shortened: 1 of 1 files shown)


kyuksel (approved):
LGTM 👍🏼 
Haven't tested it.

Qi77Qi (commented):
For other cases, the proxy url is where the app can be accessed(think UI uses this for serving the page). Does Cromwell have a UI similar to that? Or users are supposed access via swagger?

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
```diff
@@ -357,8 +359,14 @@ final case class App(id: AppId,
                      extraArgs: List[String]) {
   def getProxyUrls(project: GoogleProject, proxyUrlBase: String): Map[ServiceName, URL] =
     appResources.services.map { service =>
-      (service.config.name,
-       new URL(s"${proxyUrlBase}google/v1/apps/${project.value}/${appName.value}/${service.config.name.value}"))
+      val proxyPath = s"google/v1/apps/${project.value}/${appName.value}/${service.config.name.value}"
+      appType match {
```
rtitle, line 363: I guess we kind of made the assumption previously that `service.name` (e.g. "cromwell-service", defined [here](https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/resources/reference.conf#L420)) is the leaf element of the proxy URL. But this is not the case for accessing Swagger, I guess.

I wonder if this is a one-off for Cromwell or if other custom apps will have this issue. If it's common, we might want to make `proxyExtraPath` or something configurable in the app definition.

rtitle (approved):
This seems fine for now… also echoing Qi whether eventually Cromwell-as-an-app will have a different entry point rather than its swagger page, in which case this will need to be changed.

cjllanwarne:
jenkins retest

cjllanwarne:
jenkins retest

aednichols (approved):

PR #2344, opened 2021-10-25
Change name of cromwell helm chart to allow the name to be used as part of the templatized names (BW-859, BW-865)

cahrens:
This renames the Cromwell helm chart. The related cromwhelm MR is https://github.com/broadinstitute/cromwhelm/pull/9.

---

In all cases:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -46,7 +46,7 @@ RUN helm repo add ingress-nginx https://kubernetes.github.io/ingress-nginx && \
     helm repo add galaxy https://raw.githubusercontent.com/cloudve/helm-charts/anvil/ && \
     helm repo add terra-app-setup-charts https://storage.googleapis.com/terra-app-setup-chart && \
     helm repo add terra https://terra-app-charts.storage.googleapis.com && \
-    helm repo add cromwell-local https://broadinstitute.github.io/cromwhelm/charts/ && \
+    helm repo add cromwell-helm https://broadinstitute.github.io/cromwhelm/charts/ && \
     helm repo update
 
 # .Files helm helper can't access files outside a chart. Hence in order to populate cert file properly, we're
@@ -59,7 +59,7 @@ RUN cd /leonardo && \
     helm pull galaxy/galaxykubeman --version $GALAXY_VERSION --untar && \
     helm pull terra/terra-app --version $TERRA_APP_VERSION --untar  && \
     helm pull ingress-nginx/ingress-nginx --version $NGINX_VERSION --untar && \
-    helm pull cromwell-local/CromwellHsqldbLocalMinikube --version $CROMWELL_CHART_VERSION --untar && \
+    helm pull cromwell-helm/cromwell --version $CROMWELL_CHART_VERSION --untar && \
     cd /
 
 # Add Leonardo as a service (it will start when the container starts)
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,7 +29,7 @@ ENV TERRA_APP_VERSION 0.3.0
 ENV GALAXY_VERSION 1.1.0
 ENV NGINX_VERSION 3.23.0
 # If you update this here, make sure to also update reference.conf:
-ENV CROMWELL_CHART_VERSION 0.1.2
+ENV CROMWELL_CHART_VERSION 0.1.3
 
 RUN mkdir /leonardo
 COPY ./leonardo*.jar /leonardo
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -413,7 +413,7 @@ gke {
   }
   cromwellApp {
     # If you update the chart name or version here, make sure to also update it in the dockerfile:
-    chartName = "/leonardo/CromwellHsqldbLocalMinikube"
+    chartName = "/leonardo/cromwell"
     chartVersion = "0.1.2"
     services = [
       {
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -414,7 +414,7 @@ gke {
   cromwellApp {
     # If you update the chart name or version here, make sure to also update it in the dockerfile:
     chartName = "/leonardo/cromwell"
-    chartVersion = "0.1.2"
+    chartVersion = "0.1.3"
     services = [
       {
         name = "cromwell-service"


http/src/main/resources/reference.conf
```diff
@@ -413,8 +413,8 @@ gke {
   }
   cromwellApp {
     # If you update the chart name or version here, make sure to also update it in the dockerfile:
-    chartName = "/leonardo/CromwellHsqldbLocalMinikube"
-    chartVersion = "0.1.2"
+    chartName = "/leonardo/cromwell"
+    chartVersion = "0.1.3"
     services = [
       {
         name = "cromwell-service"
```
cahrens, line 420: I do not believe that these services values are referenced anywhere (the Cromwell service is no longer named "cromwell-service", and in fact the name cannot be stated here because it starts with a randomly generated prefix). I would recommend that we remove this services block to reduce confusion.

Dockerfile
```diff
@@ -46,7 +46,7 @@ RUN helm repo add ingress-nginx https://kubernetes.github.io/ingress-nginx && \
     helm repo add galaxy https://raw.githubusercontent.com/cloudve/helm-charts/anvil/ && \
     helm repo add terra-app-setup-charts https://storage.googleapis.com/terra-app-setup-chart && \
     helm repo add terra https://terra-app-charts.storage.googleapis.com && \
-    helm repo add cromwell-local https://broadinstitute.github.io/cromwhelm/charts/ && \
+    helm repo add cromwell-helm https://broadinstitute.github.io/cromwhelm/charts/ && \
```
cjllanwarne, line 49: I see you stopped short of `cromwhelm` here... 😢

cjllanwarne (approved):

Dockerfile
cahrens: You know, this morning I was thinking that I should have just used cromwhelm. :)

http/src/main/resources/reference.conf
cahrens: @cjllanwarne what did you think of this ^? I'm wondering if you could remove them in your MR.

cahrens:
jenkins retest

cahrens:
Since this MR requires the change to the cromwhelm repo in order for the leo build to pass (example failure: https://fc-jenkins.dsp-techops.broadinstitute.org/job/leonardo-build/12004/console), I will have to rerun automation tests after broadinstitute/cromwhelm#9 merges.

http/src/main/resources/reference.conf
cjllanwarne: Did you test bringing the app up? It looks like we're relying on these values for the "did Cromwell come up yet" check in Leo

cahrens: I did test bringing the app up and was able to load the Cromwell Swagger page.

cjllanwarne: It looks like we also use this value for generating the proxy path. It's possible that this is a "Leonardo service" name and has to _exist_ but doesn't need to match any underlying k8s service. I wonder whether @rtitle knows?

cahrens: I was wondering if it is supposed to match a label? Here are the labels on the cromwell service.

image

cahrens: I noticed that the Galaxy app has "galaxy" specified for services.name, but there isn't a service named galaxy.

cahrens: Ah, I understand now what you are saying @cjllanwarne. It's here in the URL, but doesn't seem to need to match any of the k8s service names. https://leonardo-fiab.dsde-dev.broadinstitute.org:30443/proxy/google/v1/apps/terra-test-d180e87d/cromwell-1/cromwell-service/swagger/index.html?url=/proxy/google/v1/apps/terra-test-d180e87d/cromwell-1/cromwell-service/swagger/cromwell.yaml#/Engine/engineVersion

Dockerfile
```diff
@@ -59,7 +59,7 @@ RUN cd /leonardo && \
     helm pull galaxy/galaxykubeman --version $GALAXY_VERSION --untar && \
     helm pull terra/terra-app --version $TERRA_APP_VERSION --untar  && \
     helm pull ingress-nginx/ingress-nginx --version $NGINX_VERSION --untar && \
-    helm pull cromwell-local/CromwellHsqldbLocalMinikube --version $CROMWELL_CHART_VERSION --untar && \
+    helm pull cromwell-helm/cromwell --version $CROMWELL_CHART_VERSION --untar && \
```
rtitle, line 62: I see this in the [build](https://fc-jenkins.dsp-techops.broadinstitute.org/job/leonardo-build/12004/console):
```
11:35:18 Error: chart "cromwell" matching 0.1.3 not found in cromwell-helm index. (try 'helm repo update'): no chart name found
```

Does the chart exist yet?

cahrens: The cromwhelm merge request has been approved but not yet merged. Once this ticket has thumbs I was going to merge cromwhelm, do jenkins retest here, and then merge this one after tests pass.

cahrens: FWIW, I did deploy this branch to a fiab and the build passed when cromwhelm's pages value was set to my branch instead of main.

http/src/main/resources/reference.conf
rtitle: The idea is that the service name in the Leo config is used for setting the ingress rule and building the proxy URL to the app. It can actually be any string. The config is used as follows:

1. In front Leo, when we receive a `createApp` request, we read the serviceName from config and persist it to the DB. Looks like the cromwell case is currently handled, so we are persisting `cromwell-service` to the DB: https://github.com/DataBiosphere/leonardo/blob/f9007d5b47e9a48be254aced5ae9dc45315b6732/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala#L596-L603

2. In Back Leo, when we process a `createApp` request, we read the serviceName from the DB, and pass it to the ingress config via helm. Looks like we are currently hard-coding `"cromwell-service"`: https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala#L1234. I think I would recommend reading the service from the DB instead (e.g. adding a `ServiceName` parameter to this method and passing it from the DB object).

3. After the app is created, when a `getApp` request is issued, Leo builds the proxyUrl to return to the caller. Looks like today we are reading the `serviceName` from the DB which is good, and appending some extra stuff to deal with Cromwell swagger.

So TLDR I think that config is used and we should keep it -- however it might be good to update `GKEInterepreter` to not hard-code `cromwell-service` and read it from the DB instead.

Dockerfile
rtitle: 👍

http/src/main/resources/reference.conf
cahrens: Thanks @rtitle. But to be clear, there is no need for "cromwell-service" to be the name that is in the helm chart for the Cromwell k8s service, correct? That used to be the name in our helm chart, but I am changing it to be based off of the app name, similar to galaxy.

rtitle (approved):
Looks good once helm chart is pushed and tests pass

cahrens:
jenkins retest

http/src/main/resources/reference.conf
rtitle: Ah - yeah I think that's correct, this does _not_ need to be the name of the k8s service, it's just used for generating the ingress path.

PR #2351, opened 2021-10-28
[QA-1564] Fix token expiration for AppCreationSpec

kyuksel:
https://broadworkbench.atlassian.net/browse/QA-1564

Attempting to fix `AppCreationSpec` that seems to be failing at times due to token expiring 60 mins after it's first created, and tests occassionally taking longer than that.


kyuksel:
multi-test

kyuksel:
jenkins-retest

kyuksel:
multi-test

kyuksel:
multi-test

kyuksel:
multi-test

kyuksel:
multi-test

kyuksel:
multi-test

kyuksel:
multi-test

Qi77Qi:
fwiw, I saw one failure in my PR run today https://fc-jenkins.dsp-techops.broadinstitute.org/job/leonardo-fiab-test-runner/41680/testReport/junit/org.broadinstitute.dsde.workbench.leonardo.apps/AppCreationSpec/stop_and_start_an_app/

kyuksel:
> fwiw, I saw one failure in my PR run today https://fc-jenkins.dsp-techops.broadinstitute.org/job/leonardo-fiab-test-runner/41680/testReport/junit/org.broadinstitute.dsde.workbench.leonardo.apps/AppCreationSpec/stop_and_start_an_app/

😔

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/apps/AppCreationSpec.scala
```diff
@@ -16,7 +16,8 @@ import scala.concurrent.duration._
 
 @DoNotDiscover
 class AppCreationSpec extends GPAllocFixtureSpec with LeonardoTestUtils with GPAllocUtils with ParallelTestExecution {
-  implicit val auth: Authorization =
+  // Using def instead of val to prevent test flakiness due to token expiration when tests take long
+  implicit def auth: Authorization =
     Authorization(Credentials.Token(AuthScheme.Bearer, ronCreds.makeAuthToken().value))
```
kyuksel, line 21: `Credentials` in `workbench-libs` caches tokens for 60 mins so I don't expect a performance penalty per se.

kyuksel:
multi-test

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/apps/AppCreationSpec.scala
jdcanas: Whats the Mechanism by which this would be refreshed? I thought specs were more or less singletons created at the time of test suite launch (which would mean that is when this token is initialized as well). Maybe make it a lazy val to defer its initialization (the 60min countdown) to the point where it is used?

Qi77Qi: I'm a bit less concerned about performance penalty but more introducing flakiness in a different way since getting auth token is a good call as well...maybe just add a line for grabbing new token before the line where token is likely to expire?

making it lazy val might not be sufficient if the test is taking too long....

Qi77Qi (approved):
👍  if multi-test look all good

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/apps/AppCreationSpec.scala
rtitle: I think `def` is correct given there's already a [60m cache in wb-libs](https://github.com/broadinstitute/workbench-libs/blob/develop/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/config/Credentials.scala#L16). This method is called for every call to `LeonardoApiClient`. It's better I think to have a guaranteed fresh token and not add any additional caching to Leo via `val` or `lazy val`.

rtitle (approved):
Looks good.. do you want to update any other tests? Looks like they all do this. [AppLifecycleSpec](https://github.com/DataBiosphere/leonardo/blob/develop/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/apps/AppLifecycleSpec.scala) might be another good candidate that could take a long time.

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/apps/AppCreationSpec.scala
Qi77Qi: oh...in that case, definitely makes sense to have `def`

kyuksel: > Whats the Mechanism by which this would be refreshed? I thought specs were more or less singletons created at the time of test suite launch (which would mean that is when this token is initialized as well).

The API call methods defined in `LeonardoApiClient` like [this](https://github.com/DataBiosphere/leonardo/blob/f2bea48050dc0a391b8bc0741ce564f9500ad630/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoApiClient.scala#L541) takes an `authHeader: Authorization` instance as an implicit input. Changing `val` into `def` should force `ronCreds.makeAuthToken()` to be re-evaluated at every API method call. Re-evaluation means, if a cached token has been generated less than 60 mins ago, it'll be re-used. After 60 mins, cache should be invalidated and a new token should be generated. This should fit into our use case nicely.

kyuksel:
> do you want to update any other tests? Looks like they all do this. AppLifecycleSpec might be another good candidate that could take a long time.

Good thinking. Will do.

PR #2352, opened 2021-10-29
Remove "galaxy" from Cromwell App's k8s resource names [BW-860]

cjllanwarne:
Cromwell's k8s resources are now functions of the cromwell prefixes in the database, rather than being hard-coded to use "galaxy"s configuration.

Note that this involved a little bit of refactoring of the app config code and identification of some common fields.


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/GkeAppConfig.scala
```diff
@@ -0,0 +1,48 @@
...
+                                 chartVersion: ChartVersion,
+                                 namespaceNameSuffix: String,
+                                 services: List[ServiceConfig],
+                                 serviceAccount: ServiceAccountName,
+                                 uninstallKeepHistory: Boolean,
+                                 postgresPassword: String,
+                                 orchUrl: String,
+                                 drsUrl: String)
+    extends GkeAppConfig {
+  override lazy val kubernetesServices: List[KubernetesService] = services.map(s => KubernetesService(ServiceId(-1), s))
```
cahrens, line 38: What does this do?

cahrens (approved):
Just the one question (because I'm not understanding the syntax… not because it's wrong).

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/GkeAppConfig.scala
cjllanwarne: I guess the simple answer is, brings something that was happening somewhere else into this class.

Syntax wise, it's a lazy val (so gets evaluated the first time it's asked for, then remembered), which takes the services section from config and turns each of them into a KubernetesService instance, with an ID of `-1`. 

In terms of why, I'd defer to an IA member, but I guess implicitly it assumes that Galaxy and Cromwell configs only have a single service (at least in the leo config), otherwise the IDs would clash. And later on, this is used to calculate the proxy URLs

cahrens: The ID of -1 was particularly confusing to me-- I'd love to hear more about that from IA folks.

Qi77Qi: yea a bit confusing, but this is just a helper function for persisting `KubernetesService` to DB...`id` is an auto-incrementing key, and the way slick works is you just stub out the id when saving it to DB

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/Config.scala
```diff
@@ -606,7 +606,8 @@ object Config {
     CustomAppConfig(
       config.as[ChartName]("chartName"),
       config.as[ChartVersion]("chartVersion"),
-      config.as[String]("releaseNameSuffix")
+      config.as[String]("releaseNameSuffix"),
+      config.as[String]("namespaceNameSuffix")
```
Qi77Qi: you might have some merge conflicts with https://github.com/DataBiosphere/leonardo/pull/2354

Qi77Qi (approved):

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/Config.scala
cahrens: Definitely will have merge conflicts! I figured I would do the rebase/conflict resolution of this branch after mine merges (@cjllanwarne is out until Wednesday).

cahrens:
@Qi77Qi @cjllanwarne FYI that I have rebased this branch on develop and resolved conflicts introduced by https://github.com/DataBiosphere/leonardo/pull/2354. The branch has been force pushed.

Testing notes: I first pushed up the changes on a new branch (https://github.com/DataBiosphere/leonardo/pull/2361) and verified that tests passed. I then pushed the updated JAR to my FIAB and tested deploying both Galaxy and Cromwell to make sure that the names look correct.

image

cjllanwarne:
Subsequent changes to the PR look good to me 👍

PR #2354, opened 2021-10-29
Pass database password to Cromwell app (BW-863).

cahrens:
This allows the password for Postgres to be passed to the Cromwell app, so that it can be overridden using vault secrets in the manner of https://github.com/broadinstitute/firecloud-develop/blob/1476aef978767e57d4918e25d9e643ee33025609/base-configs/leonardo/leonardo.conf.ctmpl#L95.

---

In all cases:


cahrens:
jenkins retest

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/Config.scala
```diff
@@ -598,7 +598,8 @@ object Config {
       namespaceNameSuffix = config.as[String]("namespaceNameSuffix"),
       releaseNameSuffix = config.as[String]("releaseNameSuffix"),
       services = config.as[List[ServiceConfig]]("services"),
-      serviceAccountName = config.as[ServiceAccountName]("serviceAccountName")
+      serviceAccountName = config.as[ServiceAccountName]("serviceAccountName"),
+      dbPassword = config.as[String]("dbPassword")
```
Qi77Qi: we don't typically use `String` directly. Can you add sth like
```
final case class CromwellDbPassword(value: String) extends Anyval
```
in [this](https://github.com/DataBiosphere/leonardo/blob/develop/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala) file, and use this data type for all the places where `dbPassword` is used

cahrens: Why don't you use String directly? I followed the pattern of the Galaxy db password, which is also using String (https://github.com/DataBiosphere/leonardo/pull/2354/files#diff-ba07c699cdc1e4408e16e9141ff02c3f6e447e380219a6868382409f6121a476R577).

I think it would be odd if the database password is the only String field in the config object that is using a final case class instead of a String. If you guys really don't want the Strings, all the String fields in CromwellConfig (and I would advocate the other config objects as well) should be updated. I would prefer that to be done as a separate ticket, if you are OK with that.

Qi77Qi: > Why don't you use String directly? I followed the pattern of the Galaxy db password

it's a general good practice for preventing bugs like say you have `def foo(userName: String, password: String): ???`, at call site you do sth like `foo(passwordString, userNameString)`


>  If you guys really don't want the Strings, all the String fields in CromwellConfig (and I would advocate the other config objects as well) should be updated. I would prefer that to be done as a separate ticket, if you are OK with that.

This PR feels small enough to add that refactoring..but up to you

cahrens: It feels like a lot of overhead to me… but take my opinion with a grain of salt, as I've been writing Python for the last decade (and thus am biased towards simplicity).

To reduce the proliferation of these declared types, would it be OK to introduce "DbPassword" and use it for both the Cromwell and Galaxy db parameters instead of defining both "CromwellDbPassword" and "GalaxyDbPassword"? Similarly, would there be a single "ReleaseNameSuffix" instead of "CromwellReleaseNameSuffix", "GalaxyReleaseNameSuffix", and "CustomAppReleaseNameSuffix"?

Qi77Qi: >To reduce the proliferation of these declared types, would it be OK to introduce "DbPassword" and use it for both the Cromwell and Galaxy db parameters instead of defining both "CromwellDbPassword" and "GalaxyDbPassword"? Similarly, would there be a single "ReleaseNameSuffix" instead of "CromwellReleaseNameSuffix", "GalaxyReleaseNameSuffix", and "CustomAppReleaseNameSuffix"?

I'm fine with that...it's probably not super likely that we'll mix up a galaxy db password for a cromwell db password?

cahrens: @Qi77Qi I pushed up the changes. Though TBH, I'm still not seeing how this is really helpful given how the config objects are constructed.

Qi77Qi: thanks! think there're a few unit test and automation test failures..I haven't looked at the failures, but likely at call site, you need to do sth like `password.value` now because it's a custom type instead of `String`

Qi77Qi (approved):

salonishah11 (approved):
LGTM 👍

PR #2360, opened 2021-11-02
BW-890 Add appType to getApp and listApps

cahrens:
Adds app Type information so that it can be displayed in the UI (previously the code in terra-ui was hardcoded to always display "Galaxy").

See https://github.com/DataBiosphere/terra-ui/pull/2675

---

In all cases:

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/appRoutesModels.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/appRoutesModels.scala
@@ -39,7 +39,9 @@ final case class GetAppResponse(kubernetesRuntimeConfig: KubernetesRuntimeConfig
                                 proxyUrls: Map[ServiceName, URL],
                                 diskName: Option[DiskName],
                                 customEnvironmentVariables: Map[String, String],
-                                auditInfo: AuditInfo)
+                                auditInfo: AuditInfo,
+                                appType: AppType
+                               )
 
 final case class ListAppResponse(googleProject: GoogleProject,
                                  kubernetesRuntimeConfig: KubernetesRuntimeConfig,
@@ -47,6 +49,7 @@ final case class ListAppResponse(googleProject: GoogleProject,
                                  status: AppStatus, //TODO: do we need some sort of aggregate status?
                                  proxyUrls: Map[ServiceName, URL],
                                  appName: AppName,
+                                 appType: AppType,
                                  diskName: Option[DiskName],
                                  auditInfo: AuditInfo,
                                  labels: LabelMap)
@@ -68,6 +71,7 @@ object ListAppResponse {
           a.status,
           a.getProxyUrls(c.googleProject, proxyUrlBase),
           a.appName,
+          a.appType,
           a.appResources.disk.map(_.name),
           a.auditInfo,
           a.labels.filter(l => labelsToReturn.contains(l._1))
@@ -90,6 +94,7 @@ object GetAppResponse {
       appResult.app.getProxyUrls(appResult.cluster.googleProject, proxyUrlBase),
       appResult.app.appResources.disk.map(_.name),
       appResult.app.customEnvironmentVariables,
-      appResult.app.auditInfo
+      appResult.app.auditInfo,
+      appResult.app.appType
     )
 }
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/appRoutesModels.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/appRoutesModels.scala
@@ -40,8 +40,7 @@ final case class GetAppResponse(kubernetesRuntimeConfig: KubernetesRuntimeConfig
                                 diskName: Option[DiskName],
                                 customEnvironmentVariables: Map[String, String],
                                 auditInfo: AuditInfo,
-                                appType: AppType
-                               )
+                                appType: AppType)
 
 final case class ListAppResponse(googleProject: GoogleProject,
                                  kubernetesRuntimeConfig: KubernetesRuntimeConfig,
--- a/http/src/main/resources/swagger/api-docs.yaml
+++ b/http/src/main/resources/swagger/api-docs.yaml
@@ -1990,6 +1990,12 @@ components:
             email: email authorization
             profile: profile authorization
   schemas:
+    AppType:
+      type: string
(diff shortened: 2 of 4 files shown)


http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -2909,6 +2909,9 @@ components:
           description: Optional environment variables to be set on the app
         auditInfo:
```
cahrens, line 2915: I have not actually tested the Swagger changes, as pushing over the JAR to my fiab doesn't impact update this file. I can update my fiab with the branch if there are any concerns.

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/appRoutesModels.scala
```diff
@@ -39,14 +39,16 @@ final case class GetAppResponse(kubernetesRuntimeConfig: KubernetesRuntimeConfig
                                 proxyUrls: Map[ServiceName, URL],
                                 diskName: Option[DiskName],
```
cahrens, line 40: I couldn't find any unit tests for these models (to add verification of AppType). Is there a test I should update to cover these changes?

http/src/main/resources/swagger/api-docs.yaml
Qi77Qi: updating jar should update swagger page

can you also list out the enum...appType isn't arbitrary string

cahrens: Hmm, it definitely didn't update the Swagger page.

appType isn't an arbitrary string, but if we list the values that it can be at the moment, we will have to keep updating it as new apps are added. Is that valuable enough to the user that it's worth the overhead and the possibility of being out of date? In my opinion, when you are looking at the what the types of apps in creation are, you don't really care about what other types they could have been.

cahrens: Maybe I could put something like "for example, GALAXY, CROMWELL, or CUSTOM" so that we aren't committed to updating it every time a new app is added?

dinvlad: I agree with @Qi77Qi it'd be better to expose them as enums, and keep adding to the list as we go (since it doesn't happen often). This helps with request validation in Swagger.

dinvlad: Btw I don't think we added that validation here https://github.com/DataBiosphere/leonardo/blob/b647bfe08699f697a5a2cd3a1c43f2ffebda7419/http/src/main/resources/swagger/api-docs.yaml#L2866 - adding enum to requests is much more important than to responses.

dinvlad: Just checked with @Qi77Qi , I think it's ok to keep as-is. The app creation request is getting validated on the backend, so there's no harm in keeping the response schema a little loose, from security perspective.

Qi77Qi: huh...I think it's still helpful to use enum here regardless where the validation happens...If we were to add another apptype, that's sth should be made visiable, which updating swagger kinda servces the purpose as well

cahrens: That is exposed at the add App endpoint, because that is where you need to know the available options. I do not see why you need to know all the possible App options at the point where you are examining already created Apps.

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/appRoutesModels.scala
gpcarr: you could update [HttpRoutesSpec](https://github.com/DataBiosphere/leonardo/blob/develop/http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/HttpRoutesSpec.scala) to include a test case for appType in listApp and getApp endpoints

http/src/main/resources/swagger/api-docs.yaml
dinvlad: @cahrens I think @Qi77Qi is right on that from the API robustness principle perspective: https://en.wikipedia.org/wiki/Robustness_principle

dinvlad: Basically, if our API response is lax and there're clients who start depending on the particular `appType` without us documenting them explicitly, that could mean if we later change the values of those `appType` (because they're not "in contract" if we don't declare them as such), then this could break existing clients.

That being said, not sure if folks are _actually_ using Leo APIs for any third-party integrations :-)

cahrens: I will look into creating a reference block can be referenced from addApp, listApp, and getApp. As long as I can avoid duplicating the list of app types I have no objections.

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/appRoutesModels.scala
cahrens: Thanks @gpcarr. Looking at those tests, I think perhaps it already covers this change because of this sort of test. `getAppResponse` will include the new field. But it is very much not clear to me how these tests work (with all the mocking).

````
it should "get app" in {
    Get("/api/google/v1/apps/googleProject1/app1") ~> routes.route ~> check {
      status shouldEqual StatusCodes.OK
      validateRawCookie(header("Set-Cookie"))
      responseAs[GetAppResponse] shouldBe getAppResponse
    }
  }

http/src/main/resources/swagger/api-docs.yaml
cahrens: Updated and tested on my fiab (I think I was experiencing caching issues previously).

salonishah11 (approved):
LGTM 👍  If any tests are added please feel free to hit the recycle button and I can re-review the PR.

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/appRoutesModels.scala
Qi77Qi: this is covered...the test is using the same `ListAppResponse.fromCluster` helper, which is already updated

Qi77Qi (approved):

kyuksel (approved):

PR #2365, opened 2021-11-03
[QA-1564] Update to latest workbench-service-test to get the token caching changes

kyuksel:
https://broadworkbench.atlassian.net/browse/QA-1564


project/Dependencies.scala
```diff
@@ -17,7 +17,7 @@ object Dependencies {
 
 
   private val workbenchLibsHash = "d619ea2a-SNAP"
```
Qi77Qi: update `workbenchLibsHash` instead....it's easier when all moduels are on the same hash

project/Dependencies.scala
```diff
@@ -17,7 +17,7 @@ object Dependencies {
...
-  val serviceTestV = s"0.20-$workbenchLibsHash"
```
kyuksel, line 20: @Qi77Qi I didn't want to risk messing with anything you were trying to do by using the `SNAP` version so I directly pointed to the particular `serviceTest` version.

Qi77Qi: the reason why this was `SNAP` was becuz I didn't want to wait for https://github.com/broadinstitute/workbench-libs/pull/797 to merge Mon night before I merge leo fix (otherwise auto -tests would've failed all night Mon)....

in other words, we should use latest wb-libs hash...(I did update has in the kernel issue PR, but that was a PR is held up, and haven't had the time to just updating this hash)

kyuksel: Gotcha 👍🏼  Updated accordingly.

Qi77Qi (approved):
:shipit:

kyuksel:
@Qi77Qi Thanks for merging this 🙏🏼

PR #2367, opened 2021-11-04
IA-3035 Support setting different network tags for dataproc worker nodes

rtitle:
https://broadworkbench.atlassian.net/browse/IA-3035

Depends on https://github.com/broadinstitute/workbench-libs/pull/800/

Adds a new API param `DataprocConfig.workerPrivateAccess`. If true then sets an additional `leonardo-private` network tag on the worker nodes. Defaults to false, meaning no change in behavior. This setting supports AoU.

Added an automation test, plus manually tested in a fiab. It seems to work:

Master:
image

Worker:
image

---

In all cases:


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/DataprocRuntimeMonitor.scala
```diff
@@ -137,14 +137,32 @@ class DataprocRuntimeMonitor[F[_]: Parallel](
...
+                    case Some((dataprocInstance, computeInstance)) =>
+                      dataprocInstance.ip match {
                         case Some(ip) =>
-                          // It takes a bit for jupyter to startup, hence wait 5 seconds before we check jupyter
-                          F.sleep(8 seconds) >> handleCheckTools(monitorContext, runtimeAndRuntimeConfig, ip, instances)
+                          for {
+                            // If the cluster is configured with worker private access, then
+                            // remove the workerPrivateAccessNetworkTag from the master node
+                            // once it is running.
+                            _ <- if (dataprocConfig.workerPrivateAccess) {
```
rtitle, line 150: Note this technically returns an `Operation` so we could poll for it.. but in practice it seems pretty quick and we are polling Jupyter immediately after this anyway.

rtitle:
jenkins retest

http/src/main/resources/reference.conf
```diff
@@ -152,6 +152,7 @@ vpc {
   highSecurityProjectSubnetworkLabel = "vpc-subnetwork-name"
   networkName = "leonardo-network"
   networkTag = "leonardo"
+  privateAccessNetworkTag = "leonardo-private"
```
Qi77Qi, line 155: where is this tag being used?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/DataprocRuntimeMonitor.scala
```diff
@@ -137,14 +137,32 @@ class DataprocRuntimeMonitor[F[_]: Parallel](
...
+                              googleComputeService.setInstanceTags(
+                                dataprocInstance.key.project,
+                                dataprocInstance.key.zone,
+                                dataprocInstance.key.name,
+                                Tags
+                                  .newBuilder()
+                                  .addItems(config.vpcConfig.networkTag.value)
```
Qi77Qi, line 157: this will add additional tag?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreter.scala
```diff
@@ -178,10 +178,19 @@ class DataprocInterpreter[F[_]: Parallel](
         // If we need to support 2 version of dataproc custom image, we'll update this
         dataprocImage = config.dataprocConfig.customDataprocImage
 
+        // If the cluster is configured with worker private access, then specify the
+        // `leonardo-private` network tag. This tag will be removed from the master node
+        // once the cluster is running.
+        tags = if (machineConfig.workerPrivateAccess) {
```
Qi77Qi, line 184: where is the code that removes the tag? I only see code that we're adding additional tag on master

http/src/main/resources/reference.conf
rtitle: This tag is set when we create the cluster in `DataprocInterpreter`. It matches a [firewall rule](https://github.com/DataBiosphere/terra-resource-buffer/pull/193/files#diff-dd28bc9005c6828e422edf6d2f9cf50ea66531685d27be1f36fc46478af308c6R190) defined in RBS to block outgoing traffic.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/DataprocRuntimeMonitor.scala
rtitle: This call _replaces_ the set of tags on the instance. So since the cluster was created with `[leonardo, leonardo-private]` tags, here we are setting the master to have just `[leonardo]`.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreter.scala
rtitle: It is in `DataprocRuntimeMonitor`, mentioned above

rtitle:
Note: I'm going to review https://github.com/DataBiosphere/terra-resource-buffer/pull/193 before merging to make sure this seems correct

http/src/main/resources/reference.conf
Qi77Qi: ok..is the screenshot still valid?

rtitle: Ah - the screenshot is valid if you `s/leonardo-worker/leonardo-private`

Qi77Qi (approved):

PR #2372, opened 2021-11-08
[QA-1564] Third attempt to fix token expiration for AppCreationSpec

kyuksel:
- Our test suite isn't quite robust against tests running for longer than 60 mins. When a test runs long, Google tokens start expiring because tests generate a token only once at the beginning. This PR attempts to fix that for `listApps` and `project` and `workspace` API calls used by`AppCreationSpec` since that's the only known test that can run for particularly long.
- Tested locally by reproducing the automation failure.
- I `git stash`ed a lot of changes from when I set out to make the rest of the testing codebase robust against token expiration in order to keep the scope of this PR more manageable, and also due to discovering that `implicit` `by-name` parameters don't work as I thought they would, which causes a ton of diff. Will create a Leo tech debt ticket to follow up.


automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoTestUtils.scala
```diff
@@ -724,3 +714,19 @@ trait LeonardoTestUtils
     x => x.status == status || x.status == AppStatus.Error
 
 }
+
+// Ron and Hermione are on the dev Leo's allowed list, and Hermione is a Project Owner
+sealed trait TestUser extends EnumEntry with Product with Serializable {
```
xuscrbw: What are the properties of sealed trait that make this the approach we are using?

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoApiClient.scala
```diff
@@ -538,7 +538,7 @@ object LeonardoApiClient {
   def listApps(
     googleProject: GoogleProject,
     includeDeleted: Boolean = false
-  )(implicit client: Client[IO], authHeader: Authorization): IO[List[ListAppResponse]] = {
+  )(implicit client: Client[IO], authHeader: => Authorization): IO[List[ListAppResponse]] = {
```
rtitle: I thought we talked about making this an `IO[Authorization]`? Curious why "by name" was chosen instead of IO.

I _think_ by-name still works but I had to test it:

```
// Make a side effecting function
scala> def sideEffect(): String = {
     | println("side effect")
     | "side effect"
     | }
def sideEffect(): String

// Wraps a String in an IO by value
scala> def byValue(s: String): IO[String] = IO(s)
def byValue(s: String): cats.effect.IO[String]

// Side effect is done only once when the IO is built
scala> val io = byValue(sideEffect())
side effect
val io: cats.effect.IO[String] = IO(...)

scala> io.unsafeRunSync()
val res56: String = side effect

scala> io.unsafeRunSync()
val res57: String = side effect

scala> io.unsafeRunSync()
val res58: String = side effect

// Wrap a string in an IO by name
scala> def byName(s: => String): IO[String] = IO(s)
def byName(s: => String): cats.effect.IO[String]

scala> val io = byName(sideEffect())
val io: cats.effect.IO[String] = IO(...)

// Side effect is done everytime the IO is _run_
scala> io.unsafeRunSync()
side effect
val res59: String = side effect

scala> io.unsafeRunSync()
side effect
val res60: String = side effect

scala> io.unsafeRunSync()
side effect
val res61: String = side effect
```

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoSuite.scala
```diff
@@ -124,13 +124,13 @@ trait GPAllocUtils extends BillingFixtures with LeonardoTestUtils {
   }
 }
 
-trait GPAllocBeforeAndAfterAll extends GPAllocUtils with BeforeAndAfterAll {
+trait GPAllocBeforeAndAfterAll extends GPAllocUtils with BeforeAndAfterAll with LeonardoTestUtils {
```
rtitle: I'm not sure `GPAllocBeforeAndAfterAll` should extend `LeonardoTestUtils`? I think `GPAllocBeforeAndAfterAll` is intended to be mixed into the _suite_ to provide `beforeAll` and `afterAll` -- while `LeonardoTestUtils` is intended to be mixed into actual spec classes.

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/RuntimeFixtureSpec.scala
```diff
@@ -23,7 +24,7 @@ abstract class RuntimeFixtureSpec
     with LeonardoTestUtils
     with Retries {
 
-  implicit val ronToken: AuthToken = ronAuthToken
+  implicit def ronToken: AuthToken = Ron.authToken()
```
rtitle: Do you need this `def`? Could just call `Ron.authToken()` at each call site.

rtitle:
Few comments above -- also I know we're just having trouble with `listApps` but should we update all the `LeonardoApiClient` methods which take a token just to future proof this a little bit?

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoTestUtils.scala
```diff
@@ -724,3 +714,19 @@ trait LeonardoTestUtils
...
+  val name: String
+  val creds: Credentials = LeonardoConfig.Users.NotebooksWhitelisted.getUserCredential(name)
+  val email: String = creds.email
+  def authToken(): AuthToken = creds.makeAuthToken()
```
Qi77Qi: this function should return `IO[AuthToken]`

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoApiClient.scala
Qi77Qi: think I like `IO[Authorization]` better 🤔

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoTestUtils.scala
kyuksel: The Scala compiler can check for exhaustiveness when a `trait` is `sealed` because `sealed trait`s can be `extend`ed only in the same file (see [here](https://underscore.io/blog/posts/2015/06/02/everything-about-sealed.html) for details). The exhaustiveness checking is probably not very important for this case but I think locality is desirable.

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/RuntimeFixtureSpec.scala
kyuksel: Not needed but removing it would increase the diff in this PR quite a bit since that `implicit` is relied on indirectly in a lot of places where `RuntimeFixtureSpec` is mixed in. I was hoping to contain the diffs in this PR and have a follow-up tech debt ticket do more of the refactoring so we can get the fix in sooner and with less chance of introducing errors. What do you think?

kyuksel:
@rtitle @Qi77Qi

Thanks for reviewing the PR!

Regarding the comments on usage of `IO` (or the lack thereof) and future-proofing the rest of the test code, let me try to explain the evolution of the PR:

I first locally tested and verified that using a `by-name` parameter for `auth` was fixing the issue without having to use `IO`. It looked nicely simple and `IO` felt like an overkill. Then I started applying the solution to other API calls and their various callers. Unlike `listApps()` however, a lot of the API methods are in turn calling other methods that also defines `implicit` `auth`s, in which case, `by-name` parameters aren't propagated down implicitly (it looks like there is a pending [proposal](https://docs.scala-lang.org/sips/byname-implicits.html) to support that). Explicitly passing `auth` around was causing a lot of diffs. In the spirit of getting the fix in sooner rather than later and not risking introducing other errors, I stashed my changes and was planning on creating a follow-up ticket, which could also re-evaluate the `IO` decision.

What do you think?

kyuksel:
jenkins retest

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoTestUtils.scala
rtitle: I like this ADT.. but does it need to be an enum? If not using any enumeratum functionality like `withName` then maybe this should just be a regular trait.

rtitle:
I think using `IO` for the token is more correct because the token generation is performing a side effect (it's updating a cache). "by-name" _probably_ works as well but it's not super clear to me that it actually works when called implicitly and combined with our use of `IO`. 

Also I wouldn't worry about the diff size too much, as long as the code intention is clear. How I would approach this:

1. Update all the type signatures in `LeonardoApiClient` to take an `implicit IO[Authorization]`
2. Use the compiler to update each call site of `LeonardoApiClient` -- probably adding an `implicit val token: IO[AuthToken] = Ron.authToken` or similar to various classes

I like the approach to have singleton objects for each test user as opposed to `lazy val`s everywhere -- so long as the token generation returns `IO`. :)

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoTestUtils.scala
kyuksel: Doesn't need to be an enum. Thanks for the catch!

kyuksel:
@rtitle @Qi77Qi @xuscrbw 

I believe I've addressed your comments. Could you please take another look? I'd recommend the [unified view with whitespace hidden](https://github.com/DataBiosphere/leonardo/pull/2372/files?diff=unified&w=1) :)

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/runtimes/RuntimeCreationDiskSpec.scala
```diff
@@ -58,6 +58,8 @@ class RuntimeCreationDiskSpec
       for {
         getRuntimeResponse <- LeonardoApiClient.createRuntimeWithWait(googleProject, runtimeName, createRuntimeRequest)
         clusterCopy = ClusterCopy.fromGetRuntimeResponseCopy(getRuntimeResponse)
+        rat <- Ron.authToken()
+        implicit0(authToken: AuthToken) = rat
```
Qi77Qi: you can do `implicit0(authToken: AuthToken) <- Ron.authRoken()`

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/runtimes/RuntimeCreationDiskSpec.scala
```diff
@@ -189,6 +191,8 @@ class RuntimeCreationDiskSpec
         _ <- LeonardoApiClient.createDiskWithWait(googleProject,
                                                   diskName,
                                                   defaultCreateDiskRequest.copy(size = Some(diskSize)))
+        rat <- Ron.authToken()
+        implicit0(authToken: AuthToken) = rat
```
Qi77Qi: same comment

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/runtimes/RuntimeCreationDiskSpec.scala
kyuksel: Awesome! Pushed a commit changing all occurrences accordingly.

kyuksel:
multi-test

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/apps/AppCreationSpec.scala
```diff
@@ -48,6 +47,9 @@ class AppCreationSpec extends GPAllocFixtureSpec with LeonardoTestUtils with GPA
         for {
           _ <- loggerIO.info(s"AppCreationSpec: About to create app ${googleProject.value}/${appName.value}")
 
+          rat <- Ron.authToken()
```
rtitle, line 50: 🐀

rtitle (approved):
Looks good to me. I feel like this actually reads cleaner. :shipit: if tests pass -- there might be a compile error though in auto test code: https://fc-jenkins.dsp-techops.broadinstitute.org/job/leonardo-fiab-test-runner/42086/console

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/apps/AppCreationSpec.scala
kyuksel: I had to think for a bit to understand your emoji  😂

rtitle: Ha maybe I was inspired by [this comment](https://github.com/DataBiosphere/terra-resource-buffer/pull/193#discussion_r744023823)

Qi77Qi (approved):

kyuksel:
@Qi77Qi I was going to `multi-test` it a couple of times but oh well 😀

Qi77Qi:
I think this PR is pretty safe to merge (gcp also seems has recovered from previous issues)...but if you still feel uneasy, you can open a dummy PR and trigger multi-test that way (or maybe just trigger the [job](https://fc-jenkins.dsp-techops.broadinstitute.org/job/Leonardo-PR-fiab-multi-test/) directly)

kyuksel:
It's okay. I'll monitor `dsde-workbench-notify`.

PR #2380, opened 2021-11-12
IA-3035 take 2: remove the leonardo-private tag sooner

rtitle:
cc @Qi77Qi I haven't tested this yet but here is what I'm thinking: we remove the `leonardo-private` tag from the master as soon as possible in `DataprocInterpreter` instead of waiting for the status in Dataproc to be `Running`. Hopefully that's quick enough so the init script can run..

---

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreter.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreter.scala
@@ -309,23 +309,26 @@ class DataprocInterpreter[F[_]: Parallel](
                 machineConfig.region,
                 DataprocClusterName(params.runtimeProjectAndName.runtimeName.asString)
               )
-            masterComputeInstance <- dataprocInstances.find(_._1.role == DataprocRole.Master).flatTraverse {
-              case (DataprocRoleZonePreemptibility(_, z, _), is) =>
-                is.headOption.flatTraverse { i =>
-                  googleComputeService.getInstance(params.runtimeProjectAndName.googleProject, z, i)
-                }
+            masterComputeInstanceAndZone <- dataprocInstances.find(_._1.role == DataprocRole.Master).flatTraverse {
+              case (DataprocRoleZonePreemptibility(_, z, _), instances) =>
+                instances.headOption
+                  .flatTraverse { i =>
+                    googleComputeService.getInstance(params.runtimeProjectAndName.googleProject, z, i)
+                  }
+                  .map(_.map(i => (i, z)))
             }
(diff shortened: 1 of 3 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreter.scala
```diff
@@ -289,6 +295,42 @@ class DataprocInterpreter[F[_]: Parallel](
...
+                machineConfig.region,
+                DataprocClusterName(params.runtimeProjectAndName.runtimeName.asString)
+              )
+            masterComputeInstance <- dataprocInstances.find(_._1.role == DataprocRole.Master).flatTraverse {
+              case (DataprocRoleZonePreemptibility(_, z, _), is) =>
+                is.headOption.flatTraverse { i =>
+                  googleComputeService.getInstance(params.runtimeProjectAndName.googleProject, z, i)
+                }
+            }
+            op <- masterComputeInstance.traverse { m =>
```
Qi77Qi: I wonder if we can just assmue master instance name to be `<clusterName>-m` 🤔

Qi77Qi (approved):
remove at the earliest time sounds good to me 👍

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreter.scala
rtitle: Maybe.. although we need to call `getInstance` anyway to get the fingerprint to pass to `setInstanceTags`

Qi77Qi: oh I see...

rtitle:
I _think_ this is working. I successfully made a Dataproc cluster in a dev fiab with `workerPrivateAccess: true`.

The resulting network tags look correct. Master has `leonardo`, worker has `[leonardo, leonardo-private]`. 

I compared the timestamps of the `setInstanceTags` call vs when the init script runs. And it seems like `setInstanceTags` happens first:
* setInstanceTags: 15:55:25
* init script: 15:58:02

I also tested patching the cluster to add workers + preemptibles, and the new nodes all have `[leonardo, leonardo-private]` as intended.

I don't LOVE how racy this is. E.g. if the init script runs before `setInstanceTags` then things might break. But I think it's probably fairly safe, since (a) we call `setInstanceTags` immediately after cluster creation (before polling even starts); and (b) the init script retries these operations with exponential backoff anyway.

cc @Qi77Qi @yonghaoy

PR #2394, opened 2021-11-17
IA-3040 Don't create firewall if certain project labels exist

Qi77Qi:
The previous PR was reverted in dev since there was some issues when I merged it (somehow the checks were all green when I merged it, but checking again later, it was red)

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/VPCInterpreter.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/VPCInterpreter.scala
@@ -59,7 +59,7 @@ final class VPCInterpreter[F[_]: StructuredLogger: Parallel](
     RetryPredicates.whenStatusCode(409)
   )
 
-  override def setUpProjectNetwork(
+  override def setUpProjectNetworkAndFirewalls(
     params: SetUpProjectNetworkParams
   )(implicit ev: Ask[F, TraceId]): F[(NetworkName, SubnetworkName)] =
     for {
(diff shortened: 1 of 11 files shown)


Qi77Qi:
multi-test

Qi77Qi:
fiab-start failed

jenkins retest

Qi77Qi:
multi-test

kyuksel:
Trying to understand the flow: I understand that Leo will create firewall rules based whether certain project labels exist or not. Where exactly are those labels set, and by whom?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/VPCInterpreter.scala
```diff
@@ -109,15 +109,19 @@ final class VPCInterpreter[F[_]: StructuredLogger: Parallel](
         case _ =>
           F.raiseError[(NetworkName, SubnetworkName)](InvalidVPCSetupException(params.project))
       }
+      _ <- setUpProjectFirewalls(
+        SetUpProjectFirewallsParams(params.project, network, params.region, projectLabels.getOrElse(Map.empty))
```
kyuksel, line 113: Nit: I like it better when class names are `noun`s (vs. `verb`s) since method names are typically `verb`s.

kyuksel (approved):
LGTM with a nit inline

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/VPCInterpreter.scala
Qi77Qi: my preference is actually kinda opposite cuz nouns usually make things longer (tbh, verb reads a lot better to me as well..it's typically clearer what the case class is intended for...like in this case, `SetUpProjectFirewallsParams` is super obvious what it's for)...prefers shorter names personally

PR #2401, opened 2021-11-19
[BW-892] Allow a disk to be reused when re-creating Cromwell app

salonishah11:
Snapshots from testing using FIAB:

Cromwell app `cromwell-app-1` created with `disk1` attached (note: the snapshot from db was taken after app was deleted). A workflow was successfully run in Cromwell and attached are metadata details.

Screen Shot 2021-11-19 at 11 32 43 AM

Shows `disk1` was last used by `cromwell-app-1`

Screen Shot 2021-11-19 at 11 33 28 AM

Metadata details for workflow id `407c3f1e-881b-4abf-a3ad-1645aab34ab0`

-----

After `cromwell-app-1` (not the disk) was deleted, a new Cromwell app `cromwell-restore-app-1` was created and successfully attached to `disk1` and retrieval of metadata for workflow id `407c3f1e-881b-4abf-a3ad-1645aab34ab0` was successful

Screen Shot 2021-11-19 at 11 51 44 AM

Shows `disk1` with last used by `cromwell-restore-app-1`

Screen Shot 2021-11-19 at 11 52 29 AM

Metadata details for workflow id `407c3f1e-881b-4abf-a3ad-1645aab34ab0` from new Cromwell app

Closes https://broadworkbench.atlassian.net/browse/BW-892

---

In all cases:


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeServiceInterp.scala
```diff
@@ -923,36 +923,20 @@ object RuntimeServiceInterp {
...
                   isAttached <- if (isAttachedToRuntime) F.pure(true)
                   else appQuery.isDiskAttached(pd.id).transaction
                 } yield isAttached
-              case Some(FormattedBy.Galaxy) =>
-                if (willBeUsedBy == FormattedBy.Galaxy) //TODO: If we support more apps, we need to update this check
-                  appQuery.isDiskAttached(pd.id).transaction
-                else
+              case Some(formattedBy) =>
+                if (willBeUsedBy == formattedBy) {
+                  if (formattedBy == FormattedBy.Galaxy || formattedBy == FormattedBy.Cromwell)
```
aednichols: I think factoring this as a `match` will enable the compiler to check for exhaustiveness of cases. This way, when adding a new case down the road it won't be forgotten to update here.

aednichols (commented):
Mostly me learning about Leonardo - what is `lastUsedBy` used for as distinct from `FormattedBy`?

Is there a case where the `lastUsedBy` identity can change from one app to another, despite having the same fixed `FormattedBy`?

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/apps/AppCreationSpec.scala
```diff
@@ -25,92 +31,112 @@ class AppCreationSpec extends GPAllocFixtureSpec with LeonardoTestUtils with GPA
...
-            listApps,
-            120,
-            10 seconds
-          ).compile.lastOrError
-          // TODO remove first case in below if statement when Galaxy deletion is reliable
-          _ <- if (!deletedDoneCheckable.isDone(monitorDeleteResult)) {
-            loggerIO.warn(
-              s"AppCreationSpec: app ${googleProject.value}/${appName.value} did not finish deleting after 20 minutes. Result: $monitorDeleteResult"
+    ),
+    customEnvironmentVariables = Map("WORKSPACE_NAME" -> "Galaxy-Workshop-ASHG_2020_GWAS_Demo")
```
cahrens: Should the appType of Galaxy be specified? Perhaps that is the default, but it would be more explicit if it was specified (and I would hope we keep moving away from the Galaxy default assumption).

cahrens: Nit: I would also suggest creating a helper method for createAppRequest since it looks like the only difference between the Galaxy one and the Cromwell one is the appType and the workspace name (which could be provided as input arguments).

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/apps/AppCreationSpec.scala
```diff
@@ -25,92 +31,112 @@ class AppCreationSpec extends GPAllocFixtureSpec with LeonardoTestUtils with GPA
...
-          }
-        } yield ()
+              //IO(Sam.user.deleteResource("kubernetes-app", appName.value)(ronCreds.makeAuthToken()))
+            } else {
+              // Verify creating another app with the same disk doesn't error out
+              for {
+                _ <- loggerIO.info(
+                  s"AppCreationSpec: app ${googleProject.value}/${appName.value} delete result: $monitorDeleteResult"
+                )
+                _ <- LeonardoApiClient.createAppWithWait(googleProject, restoreAppName, createAppRequest)(
```
cahrens, line 120: I'm having trouble understanding-- is the disk name to reuse passed in, or does the backend code automatically reuse a disk if one exists for that workspace/app combination?

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/diskModels.scala
```diff
@@ -111,8 +112,12 @@ object FormattedBy extends Enum[FormattedBy] {
   final case object Custom extends FormattedBy {
     override def asString: String = "CUSTOM"
   }
+
+  final case object Cromwell extends FormattedBy {
```
cahrens, line 115: Nit: It's interesting that we have both FormattedBy and AppType, even though they are the same strings. IMO it would be simpler if there was just a single enum.

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
```diff
@@ -312,6 +312,13 @@ object AppType {
 
   def values: Set[AppType] = sealerate.values[AppType]
   def stringToObject: Map[String, AppType] = values.map(v => v.toString -> v).toMap
+
+  def appTypeToFormattedByType(appType: AppType): FormattedBy =
+    appType match {
```
cahrens, line 317: The need for this is related to my previous nit.

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20211116_update_persistent_disk_formattedBy_enum.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="leonardo" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+
+    <changeSet logicalFilePath="leonardo" author="sshah" id="update_persistent_disk_formattedBy_enum">
+        <modifyDataType tableName="PERSISTENT_DISK" columnName="formattedBy" newDataType="ENUM('GALAXY', 'GCE', 'CROMWELL', 'CUSTOM')"/>
```
cahrens, line 5: Question for IA: So does this mean that all CUSTOM apps have to use the same persistent disk instance?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
```diff
@@ -156,16 +156,19 @@ final class LeoAppServiceInterp[F[_]: Parallel](
...
-                lastUsedOpt <- appQuery.getLastUsedApp(galaxyRestore.lastUsedBy, Some(ctx.traceId)).transaction
+                lastUsedBy <- F.fromOption(
+                  diskResult.disk.lastUsedBy,
+                  new LeoException("Existing disk found, but no restore info found in DB", traceId = Some(ctx.traceId))
+                )
+                lastUsedOpt <- appQuery.getLastUsedApp(lastUsedBy, Some(ctx.traceId)).transaction
                 lastUsed <- F.fromOption(
                   lastUsedOpt,
-                  new LeoException(s"last used app(${galaxyRestore.lastUsedBy}) not found", traceId = Some(ctx.traceId))
+                  new LeoException(s"last used app($lastUsedBy) not found", traceId = Some(ctx.traceId))
```
cahrens, line 168: Why is this not diskResult.disk.lastUsedBy?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
```diff
@@ -156,16 +156,19 @@ final class LeoAppServiceInterp[F[_]: Parallel](
         if (diskResult.creationNeeded) F.pure(none[LastUsedApp])
         else {
           (diskResult.disk.formattedBy, diskResult.disk.galaxyRestore) match {
-            case (Some(FormattedBy.Galaxy), Some(galaxyRestore)) =>
+            case (Some(FormattedBy.Galaxy), Some(_)) | (Some(FormattedBy.Cromwell), None) =>
```
cahrens: What is the second argument in this case statement?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
```diff
@@ -179,7 +182,7 @@ final class LeoAppServiceInterp[F[_]: Parallel](
               F.raiseError[Option[LastUsedApp]](
                 new LeoException("Existing disk found, but no restore info found in DB", traceId = Some(ctx.traceId))
               )
-            case (Some(FormattedBy.GCE), _) | (Some(FormattedBy.Custom), _) =>
+            case (Some(FormattedBy.GCE), _) | (Some(FormattedBy.Custom), _) | (Some(FormattedBy.Cromwell), Some(_)) =>
               F.raiseError[Option[LastUsedApp]](
                 new LeoException(
                   s"Disk is formatted by ${diskResult.disk.formattedBy.get} already, cannot be used for galaxy app",
```
cahrens, line 188: This error message shouldn't refer to Galaxy, should it? And what is this error case?

cahrens: Or maybe Galaxy is correct here because this code is specifically a Galaxy pathway?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GceInterpreter.scala
```diff
@@ -150,7 +150,7 @@ class GceInterpreter[F[_]](
               )
             )
             isFormatted <- persistentDisk.formattedBy match {
-              case Some(FormattedBy.Galaxy) | Some(FormattedBy.Custom) =>
+              case Some(FormattedBy.Galaxy) | Some(FormattedBy.Custom) | Some(FormattedBy.Cromwell) =>
```
cahrens, line 151: I wish there was some way that we could avoid having to explicitly list the different app types to make it easier to add new ones. Could we have a property about App vs. Runtime so we could iterate through the Apps?

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/apps/AppCreationSpec.scala
salonishah11: The name of the disk to be reused is passed in `createAppRequest` (the same request used to create the original instance of app). The new name of the app is passed through `restoreAppName`.

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/diskModels.scala
salonishah11: Yes. I am not sure in which case one would be different from another. Maybe the IA team can answer this. I didn't have enough knowledge on it's background hence didn't collapse it into single enum.

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20211116_update_persistent_disk_formattedBy_enum.xml
salonishah11: The reason I added `CUSTOM` to this enum was so that custom apps can be marked `CUSTOM` in DB. Up until know I think it was default to `GALAXY`. If that was on purpose, I can remove this change.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
salonishah11: `disk.galaxyRestore` stores details (PVC ids) on how to restore a Galaxy app specifically and reconnect to an existing disk that was formatted by Galaxy previously.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeServiceInterp.scala
```diff
@@ -931,9 +931,10 @@ object RuntimeServiceInterp {
                 } yield isAttached
               case Some(formattedBy) =>
                 if (willBeUsedBy == formattedBy) {
-                  if (formattedBy == FormattedBy.Galaxy || formattedBy == FormattedBy.Cromwell)
-                    appQuery.isDiskAttached(pd.id).transaction
-                  else RuntimeConfigQueries.isDiskAttached(pd.id).transaction
+                  formattedBy match {
+                    case FormattedBy.Galaxy | FormattedBy.Cromwell => appQuery.isDiskAttached(pd.id).transaction
+                    case FormattedBy.GCE | FormattedBy.Custom      => RuntimeConfigQueries.isDiskAttached(pd.id).transaction
```
cahrens: I think that Custom should actually be in the same case statement as Galaxy and Cromwell (I *think* Custom would be a k8s App), but IA should confirm.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
cahrens: FYI, apparently "restore" is used for the Pause/Resume functionality.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/PersistentDiskComponent.scala
```diff
@@ -70,7 +71,8 @@ class PersistentDiskTable(tag: Tag) extends Table[PersistentDiskRecord](tag, "PE
      diskType,
      blockSize,
      formattedBy,
-     (galaxyPvcId, cvmfsPvcId, lastUsedBy)) <> ({
+     (galaxyPvcId, cvmfsPvcId),
+     lastUsedBy) <> ({
```
xuscrbw: Clarification question, what was the reason for breaking this out? Was it to allow easier look up in the DB for lastUsedBy?

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/diskModels.scala
xuscrbw: In the case that the formattedBy is "GCE" there will not be an AppType, so I do not think we can collapse these values just yet.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/PersistentDiskComponent.scala
salonishah11: I extracted this out since Cromwell app only needs the `lastUsedBy` and not `galaxyPvcId, cvmfsPvcId` to reconnect to previously used disk when recreating the app.

cahrens: If I understand correctly, the Galaxy "recreate" scenario is actually pause/resume. We are not actually "recreating" Cromwell, we are just creating a new instance using a disk that already exists from a previously deleted instance.

salonishah11: Yes that is correct. For Cromwell, `lastUsedBy` is being used to make sure that the Cromwell app that last used the disk and the new Cromwell app that is requesting to reconnect to the disk are in same workspace.
Links:

- [get last used App](https://github.com/DataBiosphere/leonardo/blob/5636dffa3ac58e8a1b1f69ecdcf18fc4a981ceb8/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/AppComponent.scala#L230-L247)
- [check that workspace is same](https://github.com/DataBiosphere/leonardo/blob/cdd26de1b56a92fa3005f28835b1a2d0edfa4596/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala#L170-L178)

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/diskModels.scala
```diff
@@ -111,8 +112,12 @@ object FormattedBy extends Enum[FormattedBy] {
...
+    override def asString: String = "CROMWELL"
+  }
 }
 
 final case class PvcId(asString: String) extends AnyVal
 // information needed for restoring a galaxy app
-final case class GalaxyRestore(galaxyPvcId: PvcId, cvmfsPvcId: PvcId, lastUsedBy: AppId)
+final case class GalaxyRestore(galaxyPvcId: PvcId, cvmfsPvcId: PvcId)
```
Qi77Qi: I'd prefer the data model to be sth like
```
sealed trait AppRestore extends Product with Serializable {
  def lastUsedBy: AppId
}
object AppRestore {
  final case class GalaxyRestore(galaxyPvcId: PvcId, cvmfsPvcId: PvcId, lastUsedBy: AppId) extends AppRestore
  final case class CromwellRestore(lastUsedBy: AppId) extends AppRestore
}
```

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
```diff
@@ -312,6 +312,13 @@ object AppType {
...
+      case Galaxy   => FormattedBy.Galaxy
+      case Cromwell => FormattedBy.Cromwell
+      case _        => FormattedBy.Custom
```
Qi77Qi: `_` makes me a bit nervous..can you do `case Custom => FormattedBy.Custom`

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/diskModels.scala
salonishah11: So with this is the idea that [PersistentDiskRecord](https://github.com/DataBiosphere/leonardo/blob/cdd26de1b56a92fa3005f28835b1a2d0edfa4596/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/PersistentDiskComponent.scala#L33-L34) with store both GalaxyRestore and CromwellRestore details, and those details will be populated based on which app is present in `formattedBy` field?

Qi77Qi: yea exactly...can jump on a call if you need some help with that

salonishah11: I see. I will try and refactor it. If I come across any hurdles I can let you know. Thank you :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -361,7 +364,11 @@ class GKEInterpreter[F[_]](
       ).compile.lastOrError
 
       //TODO: validate app release is the same as restore release
-      galaxyRestore <- persistentDiskQuery.getGalaxyDiskRestore(diskId).transaction
+      appRestore: Option[AppRestore] <- persistentDiskQuery.getAppDiskRestore(diskId).transaction
+      galaxyRestore: Option[GalaxyRestore] = appRestore.flatMap {
```
cahrens, line 368: Can you explain this code? It seems odd to have CromwellRestore be a case within something a block of code that seems specific to Galaxy.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -616,7 +624,11 @@ class GKEInterpreter[F[_]](
         s"Delete app operation has finished for app ${app.appName.value} in cluster ${gkeClusterId.toString}"
       )
 
-      _ <- dbApp.app.appResources.disk.flatMap(_.galaxyRestore).traverse { restore =>
+      appRestore: Option[GalaxyRestore] = dbApp.app.appResources.disk.flatMap(_.appRestore).flatMap {
```
cahrens, line 638: Same question as the previous.

cahrens:
@salonishah11 I updated my FIAB with your branch, and now I am getting "formattedBy" as a field on the disk. I'm a bit surprised though by the casing-- I am seeing "GCE", "Cromwell", and "Galaxy". I would have expected CROMWELL and GALAXY, which would match the AppType. It's not a huge deal because I can do a case-insensitive match, but if you have control over this, can you make it upper case?

(Perhaps it has something to do with the parsing of the JSON response, as the disk statuses are also camel-cased instead of upper-case)

Also, the first time I tried creating a Galaxy instance I got an error that resources weren't available. I doubt it's related to your change, but then I ended up with a disk with null as provisionedBy. I was able to delete the Galaxy instance that errored, but I can't delete the disk which is stuck in Creating.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
salonishah11: Since `getAppDiskRestore` now returns an `AppRestore`, it is trying to convert it into `GalaxyRestore`. So if it encounters a `CromwellRestore` for the disk, it returns `None`. But now I am thinking if it actually returned `CromwellRestore`, instead of returning `None` it show throw an error? 🤔

cahrens: I don't understand the syntax/architecture-- perhaps you can explain it to me during the mob time.

salonishah11:
Edit: Please ignore this, I didn't see your previous message

@cahrens The list disks API should now return `formattedBy` in it's response

Screen Shot 2021-11-30 at 9 48 28 AM

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
```diff
@@ -33,6 +33,9 @@ import org.typelevel.log4cats.StructuredLogger
 
 import java.time.Instant
 import java.util.UUID
+
+import org.broadinstitute.dsde.workbench.leonardo.AppRestore.{CromwellRestore, GalaxyRestore}
```
cahrens: Shouldn't this import be up with the other org.broadinstitute imports?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
```diff
@@ -167,17 +171,15 @@ final class LeoAppServiceInterp[F[_]: Parallel](
                     )
                 }
               } yield lastUsed.some
-            case (Some(FormattedBy.Galaxy), None) =>
+            case (Some(FormattedBy.Galaxy), Some(CromwellRestore(_))) =>
+              getAlreadyFormattedError(FormattedBy.Galaxy, "Cromwell", ctx.traceId)
```
cahrens: Can you use a constant for "Cromwell" (and "Galaxy" on line 178)?

So this code is saying that if you are trying to use a disk for a type that is different from the restore type, throw an error? It would be so nice if we could find a way to express this without explicitly listing every combination.

I'm also confused about the cases on line 176 that are mapping to Galaxy having formatted the disk. Shouldn't the second argument in both of those cases be GalaxyRestore?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeServiceInterp.scala
cahrens: @Qi77Qi should Custom be in the case with Galaxy and Cromwell?

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreterSpec.scala
```diff
@@ -31,6 +31,9 @@ import org.scalatest.flatspec.AnyFlatSpecLike
 
 import java.nio.file.Files
 import java.util.Base64
+
+import org.broadinstitute.dsde.workbench.leonardo.AppRestore.GalaxyRestore
```
cahrens: Why was this import needed?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
salonishah11: It seems IntelliJ auto import put it there, but I can move it up with other org.broadinstitute imports

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreterSpec.scala
salonishah11: Without it, [this](https://github.com/DataBiosphere/leonardo/blob/54a6a9d7b83282d604b36407cba6a12251260281/http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreterSpec.scala#L143) line of code fails

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
salonishah11: The case conditions on line 176 are actually same as before (see the red area lines from 174 to 180) with a new Cromwell condition to it. I am thinking that disks previously formatted by Custom or GCE apps can't be reused by any other app and hence the second argument doesn't matter. @Qi77Qi can you clarify the reasoning behind this case statement?

salonishah11:
I tested manually in FIAB below scenarios for Cromwell and Galaxy app and they worked as expected:

Cromwell (through Swagger):
- Create a Cromwell app
- Once it is running successfully, run a workflow
- Delete the current app. Don't delete the disk
- Create a new Cromwell app and pass in the same disk name as before
- Once the new app is running, retrieve the metadata for workflow run in the first app

Galaxy (through UI):
- Create a Galaxy app from Notebooks tab
- Launch Galaxy app once provisioned
- Upload a small file
- Delete the app. Don't delete the disk
- Create a new Galaxy app through UI. UI should pass the name of previously used Galaxy disk
- Launch new app once provisioned and verify that the previously uploaded file still exists 

@cahrens - the `formattedBy` in `listDisks()` api is fixed and is now returned in uppercase

Screen Shot 2021-11-30 at 11 44 14 PM

cahrens:
@salonishah11 I updated my fiab against your branch. I then created a Workspace, started creating a Galaxy instance, then immediately tried to create a Jupyter instance. I got this error:

{
  "source": "leonardo",
  "message": "Persistent disk terra-test-f828b4cd/saturn-pd-2f869dee-8da9-42d3-8c89-f60b9e27db79 is already attached to another runtime",
  "statusCode": 409,
  "exceptionClass": "class org.broadinstitute.dsde.workbench.leonardo.http.service.DiskAlreadyAttachedException",
  "traceId": "f9d29383-99e7-4046-a0c7-72df9bf38c8c"
}

I then went to the Cloud Environments page to see if anything got created for Jupyter (it did not). When I returned to my workspace I was able to create a Jupyter instance (which created a new disk).

The persistent disk name that I got an error for is the name of the disk associated with the Galaxy app, so for some reason Jupyter tried to use that disk.

It may be that there is a lag between when a disk is created and when its provisionedBy status is set. I saw that when I created a Cromwell instance, I got the warning icon about 2 disks in the same workspace (where it is looking specifically for 2 disks that are both "not apps"). When I refreshed the page, the warning had gone away. This is consistent with the issue I saw yesterday that when Galaxy failed to create properly (it showed on the Cloud Environments page, but with an error icon), the formattedBy for its disk was null. If there is a way that provisionedBy can be set at the beginning, it will be much better.

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/apps/AppLifecycleSpec.scala
```diff
@@ -45,22 +45,9 @@ class AppLifecycleSpec
...
-        Map.empty
-      )
-    ),
-    appType = AppType.Cromwell
-  )
-
   private val appTestCases = Table(
     ("description", "createAppRequest"),
-    ("create and delete a CUSTOM app", customCreateAppRequest),
-    ("create and delete a CROMWELL app", cromwellCreateAppRequest)
```
Qi77Qi, line 63: the test that was initially here moved to AppCreationSpec?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/LeoProfile.scala
```diff
@@ -159,7 +159,7 @@ private[leonardo] object LeoProfile extends MySQLProfile {
       )
     implicit val formattedByMappedColumnType: BaseColumnType[FormattedBy] =
       MappedColumnType.base[FormattedBy, String](
-        _.asString,
+        _.toString,
```
Qi77Qi: why are we doing `toString` instead of `asString`? `toString` is a default function, so it's hard to tell if the value is really intentional..`asString` is a common function we use to make it obvious that the returned value is an intentional string representation that we need to be care about backawrds compatibility etc

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -421,46 +428,46 @@ class GKEInterpreter[F[_]](
...
+                      app.id
+                    )
+                    persistentDiskQuery
+                      .updateGalaxyDiskRestore(diskId, galaxyDiskRestore)
+                      .transaction
+                      .void
+                }
+            } yield ()
+        case AppType.Cromwell => persistentDiskQuery.updateLastUsedBy(diskId, app.id).transaction
+        case _                => F.unit
```
Qi77Qi: would prefer to have `case AppType.Custom => F.unit`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeServiceInterp.scala
Qi77Qi: yea. for Custom app, it should use `appQuery` for checking if the disk is attached

Qi77Qi:
can you also run `multi-test` once you're ready

cahrens:
> It may be that there is a lag between when a disk is created and when its provisionedBy status is set. I saw that when I created a Cromwell instance, I got the warning icon about 2 disks in the same workspace (where it is looking specifically for 2 disks that are both "not apps"). When I refreshed the page, the warning had gone away. This is consistent with the issue I saw yesterday that when Galaxy failed to create properly (it showed on the Cloud Environments page, but with an error icon), the formattedBy for its disk was null. If there is a way that provisionedBy can be set at the beginning, it will be much better.

Following up on this, I'm seeing that while the disk status is "Creating", formattedBy is null. Therefore I'm changing the UI code to ignore disks in the "Creating" status for the purposes of warning about multiple disks of a given type in the same project.

However, I still think it's worth seeing if it is possible to set formattedBy earlier, as the app already has a link to its disk during that time.

kyuksel:
> However, I still think it's worth seeing if it is possible to set formattedBy earlier, as the app already has a link to its disk during that time.

Makes sense to me and I don't see why we couldn't do that.

cahrens:
> > However, I still think it's worth seeing if it is possible to set formattedBy earlier, as the app already has a link to its disk during that time.
> 
> Makes sense to me and I don't see why we couldn't do that.

I have gone back to using the saturnApplication label to determine if a disk is associated with an app. The possibility of creating a notebook with the wrong disk (because the disk was still in a "Creating" status with null `formattedBy`) was too big of a risk, an I found out yesterday that I could request the saturnApplication label information to be returned from the server. Therefore you can back out the changes to include `formattedBy` in the API if you wish.

kyuksel:
> I have gone back to using the saturnApplication label to determine if a disk is associated with an app. The possibility of creating a notebook with the wrong disk (because the disk was still in a "Creating" status with null formattedBy) was too big of a risk, an I found out yesterday that I could request the saturnApplication label information to be returned from the server. Therefore you can back out the changes to include formattedBy in the API if you wish.

+1 on backing out the changes to include `formattedBy` in the Leo API since there's no longer a use case.

salonishah11:
@cahrens sounds good. I will remove `formattedBy` from the api since it is no longer going to be used. Once the PR is ready and I have incorporated feedback from Qi, I will ping for re-reviews.

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/apps/AppLifecycleSpec.scala
salonishah11: Yes. AppLifecycleSpec was only creating an app with disk attached, monitoring it and then deleting it. While AppCreationSpec is additionally re-creating the app and connecting to same disk which was better for Cromwell app's test case. Hence I moved it there.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/LeoProfile.scala
salonishah11: I had made that change so that `listDisks()` API could return `formattedBy` value in uppercase. But I have reverted that change as we no longer need `formattedBy` in the response.

salonishah11:
multi-test

salonishah11:
@cahrens @Qi77Qi this is now ready for re-review

cahrens:
@salonishah11 I think you need to retry the build.

cahrens (approved):

salonishah11:
jenkins retest

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
Qi77Qi: Nit, but I honestly would prefer there's no `getAlreadyFormattedError`, but instead just a custom exception

`final case class AlreadyFomattedError(alreadyFormattedBy: FormattedBy, targetAppName: String, traceId: TraceId)` 

And have `F.raiseError(???)` in where `getAlreadyFormattedError` is currently

Qi77Qi (approved):

Qi77Qi:
lgtm....thanks for addressing the PR comments

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
salonishah11: Yeah that can be done. It would be cleaner as well

PR #2402, opened 2021-11-19
[BW-919] - Leonardo update for basic Cromwell PAPI config

kpierre13:

---

In all cases:


Qi77Qi (approved):

Qi77Qi:
run `sbt scalafmtAll`

cahrens (approved):

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -1229,14 +1238,18 @@ class GKEInterpreter[F[_]](
...
+    cluster: KubernetesCluster,
+    nodepoolName: NodepoolName,
+    namespaceName: NamespaceName,
+    disk: PersistentDisk,
+    customEnvironmentVariables: Map[String, String]
+  ): List[String] = {
     val proxyPath = s"/proxy/google/v1/apps/${cluster.googleProject.value}/${appName.value}/cromwell-service"
     val k8sProxyHost = kubernetesProxyHost(cluster, config.proxyConfig.proxyDomain).address
     val leoProxyhost = config.proxyConfig.getProxyServerHostName
+    val gcsBucket = customEnvironmentVariables.getOrElse("WORKSPACE_BUCKET", "")
```
cjllanwarne: Feels like a slightly weird way of accessing the bucket (via environment variables that might be passed to the pod, but aren't, and we fish a specific value out of them instead and throw the rest away)...

I'm not saying it's bad, just surprising. Can we look at the code that puts the bucket _into_ the environment variables list in the first place and look up the bucket in the same way?

aednichols: I don't know whether it makes sense in context, but sometimes I like to do something like
```
customEnvironmentVariables.getOrElse("WORKSPACE_BUCKET", "<no workspace bucket defined>")
```
so that error messages look like
```
cannot access 'gs://<no bucket in config>/blah.txt'
```
instead of
```
cannot access 'gs:///blah.txt'
```

aednichols (approved):

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
cjllanwarne: A missing bucket value should probably be a failure to start the instance rather than passing along an empty value and letting Cromwell fail later on

cjllanwarne: Or even better, can we fail here instead of letting cromwell get half-started and then fail with an awkward error message that needs interpretation?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -1259,7 +1272,10 @@ class GKEInterpreter[F[_]](
       // Persistence
       raw"""persistence.size=${disk.size.gb.toString}G""",
       raw"""persistence.gcePersistentDisk=${disk.name.value}""",
-      raw"""env.swaggerBasePath=$proxyPath"""
+      raw"""env.swaggerBasePath=$proxyPath""",
+      // cromwellConfig
+      raw"""cromwellConfig.gcsProject=${cluster.googleProject.value}""",
+      raw"""cromwellConfig.gcsBucket=$gcsBucket"""
```
cjllanwarne: During the PR process for https://github.com/broadinstitute/cromwhelm/pull/15 this got simplified to just `config`, so these are now mismatching what the helm chart expects

```suggestion
      raw"""config.gcsProject=${cluster.googleProject.value}""",
      raw"""config.gcsBucket=$gcsBucket/cromwell-execution"""
```

cjllanwarne (changes requested):

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
cjllanwarne: I also had to add `/cromwell-execution` onto the end. Despite being called "bucket", this actually wants a path within the bucket... 🤦

cjllanwarne: (or alternatively, I guess the helm chart could always append something like `/cromwell-execution` to the end of the bucket that Leo provides it with... that's probably better in the long term but would require an additional helm change... 😬 )

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
kpierre13: So are you asking that we find the customEnvironmentVariables function in the code and see if there is an alternative? @cjllanwarne

cjllanwarne: That's kind of where I was going... it wouldn't be the end of the world to merge as is, so don't spend too long on it, this way just seemed a little odd to me

cjllanwarne (approved):

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
kpierre13: I see. The function `buildGalaxyChartOverrideValuesString` did a similar thing for the workspace name and workspace namespace in the same file (line 1293 & 1294), but it would be interesting to see this done in the way you suggested.

PR #1940, opened 2021-03-22
[QA-1411] cache charts

jdcanas:
Hoping to address some automation test stability issues along the lines  of

```2021/03/22 10:52:41 failed to download "galaxy/galaxykubeman" (hint: running `helm repo update` may help)```
```{"kubernetesRuntimeConfig":{"numNodes":1,"machineType":"n1-highmem-8","autoscalingEnabled":false},"errors":[{"errorMessage":"failed to download \"ingress-nginx/ingress-nginx\" (hint: running `helm repo update` may help)",...```

PR #2430, opened 2021-11-30
Stop caching Galaxy's chart

almahmoud:
Reverts https://github.com/DataBiosphere/leonardo/pull/1940 for Galaxy which will allow us to make small within-version updates/bugfixes/tool updates/etc to Galaxy without having to ask or wait for a Leo rebuild, as was the case before the PR above.
Also changes the helm repository to a GCS repository, to remove any Github server issues when pulling the repo at helm install time.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -43,7 +43,7 @@ RUN curl -fsSL -o get_helm.sh https://raw.githubusercontent.com/helm/helm/master
 
 # Add the repos containing nginx and galaxy charts
 RUN helm repo add ingress-nginx https://kubernetes.github.io/ingress-nginx && \
-    helm repo add galaxy https://raw.githubusercontent.com/cloudve/helm-charts/anvil/ && \
+    helm repo add galaxy https://storage.cloud.google.com/galaxy-anvil/helm && \
     helm repo add terra-app-setup-charts https://storage.googleapis.com/terra-app-setup-chart && \
     helm repo add terra https://terra-app-charts.storage.googleapis.com && \
     helm repo add cromwell-helm https://broadinstitute.github.io/cromwhelm/charts/ && \
--- a/Dockerfile
+++ b/Dockerfile
@@ -56,7 +56,6 @@ RUN helm repo add ingress-nginx https://kubernetes.github.io/ingress-nginx && \
 
 RUN cd /leonardo && \
     helm pull terra-app-setup-charts/terra-app-setup --version $TERRA_APP_SETUP_VERSION --untar && \
-    helm pull galaxy/galaxykubeman --version $GALAXY_VERSION --untar && \
     helm pull terra/terra-app --version $TERRA_APP_VERSION --untar  && \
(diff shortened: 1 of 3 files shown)


Qi77Qi:
@almahmoud you might wanna try deleting your local leonardo repo and re-clone...Just so you don't have to rely someone from our team to re-open the same PR you have just to trigger tests

almahmoud:
I repulled and rebased on top of the latest develop.

Qi77Qi:
taking a closer look... https://github.com/DataBiosphere/leonardo/pull/1940 is trying to address test flakiness, which is a big issue for us. I'd prefer we don't regress on that.

There's also some concerns from @xuscrbw that surprising chart updates that might affect our testing or prod.

kyuksel:
Closing the PR as it seems stale. Please re-open if/when needed.

almahmoud:
On the Galaxy side, the GCS chart repository was made for the sole purpose of addressing flakiness of github which was the original stated problem leading to the caching. Conceptually, caching seemed undesirable in terms of slowing down the process of updates, and taking away a lot of control from the chart maintainers.
I understand anxiety about production changes, but my personal opinion was that if Galaxy team is tasked with maintaining/configuring the chart and deciding what should be in production, there should be some level of trust/control over that production chart given to the maintainers, rather than always having to go through Terra team for any changes, especially since it seems like it just creates unnecessarily long back and forths but all the debugging and bug fixing to the chart happens on the Galaxy side anyway. This situation seems to me that it creates a model where the people with all the responsibility over the chart have none of the control over modifying it in production, while the people who have the control to modify it are not maintainers at all.
More concretely, if someone reports a bug in production that should be fixed in the chart, and someone on Galaxy team has a quick fix that is backwards compatible and critical for production, having the ability to roll it into production without passing through another team seems normal and desirable, since ultimately the responsibility for that bug being fixed would conceptually fall on the Galaxy folks. In the event that there is a bug that will accidentally be rolled into production, if it is caught by automatic tests it will be caught either way, if it is not, then I doubt it will be caught by Terra team even with the caching, unless there is extensive in-Galaxy testing being done that I don't know of. So ultimately, the extra layer seems like it adds more complexity but no real advantage, and from personal experience it was unnecessarily frustrating and time consuming (for engineering time on both sides) when I had to do it. In the cached model, no bug can be fixed without both teams getting involved: Galaxy folks to actually fix the problem, and Broad folks just to move it to production, which doesn't seem to conceptually make sense to me unless Broad folks are also committing to taking some chart maintenance/debugging responsibility, cause otherwise it seems like maintaining control over something that will only ever be changed by another team which seems like somewhat unnecessary gatekeeping. More long term, this mode also seems to conceptually become less and less efficient, if the idea is to have more partner projects be able to similarly contribute charts in the future.

In any case, I have since moved away from the Galaxy project so this won't be my problem to deal with anymore, and I think I've made all of these points multiple times on Slack before, but just figured at least for the record I'd make it one last time on this PR as well.

PR #2474, opened 2021-12-17
[IA-3142] Bump Dataproc image version to 2.0.26

kyuksel:
The upgrade is to include Dataproc's addressing of log4j security vulnerabilities as of 12/16/2021.


Qi77Qi (approved):
thanks @kyuksel !!!

PR #2490, opened 2022-01-03
[DDO-1730] Expose metrics to Prometheus

jack-r-warren:
Adds the OpenCensus-to-Prometheus exporter added and tested in https://github.com/broadinstitute/workbench-libs/pull/851 to Leonardo. 

This PR combined with the helm chart addition in https://github.com/broadinstitute/terra-helmfile/pull/2163 makes all existing OpenCensus metrics accessible through our new Grafana server (in addition to the current Stackdriver). Neither PR will break anything if the other isn't present.

~~I am not sure how to address test coverage: I don't know how to write a meaningful test for the added lines in the huge Boot.scala method (for what it is worth, the library method being called there _is_ tested over in the library). I've tested all of this manually end-to-end in dev, it is just the in-code tests I'm unsure about.~~ Edit: addressed below

---

> No API changes here, app behavior on existing ports is exactly the same as before. Just adds a new internal-use port with metric data.

In all cases:

- [x] Verify all tests go green (you can re-run automation tests with a comment saying `jenkins retest`
- [x] Run the automation tests multiple times in parallel to weed out instability if applicable via a comment saying `multi-test`
- [x] Squash and merge; Delete your branch after this


gpcarr:
@jack-r-warren thank you for adding this!

Manual testing for changes to `Boot.scala` is okay - we don't have unit tests or automation test coverage for that. We don't always have `codecov/patch` pass on PRs. Since you added to `Config.scala`, you could add a test case to [ConfigSpec](https://github.com/DataBiosphere/leonardo/blob/develop/http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/config/ConfigSpec.scala) and that would be enough for unit testing! And I'm guessing the new workbench libs hash is published already since manual testing is working?

jack-r-warren:
Done @gpcarr, thanks! And you're correct, the library updates have already been published.

gpcarr (approved):
lgtm

mikeflinn (approved):
Lgtm, thanks for doing this Jack!

jyang-broad (approved):
Look good!

PR #2497, opened 2022-01-10
PROD-628 install jattach

Qi77Qi:

---

In all cases:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -66,4 +66,5 @@ RUN cd /leonardo && \
 CMD java $JAVA_OPTS -jar $(find /leonardo -name 'leonardo*.jar')
 
 # Install https://github.com/apangin/jattach to get access to JDK tools
-RUN apt install jattach
+RUN apt update && \
+    apt install jattach
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,7 +15,7 @@ RUN mkdir /helm-go-lib-build && \
 
 # Use this graalvm image if we need to use jstack etc
 # FROM ghcr.io/graalvm/graalvm-ce:ol8-java11-21.0.0.2
-FROM us.gcr.io/broad-dsp-gcr-public/base/jre:11-debian
+FROM ghcr.io/graalvm/graalvm-ce:java11-21.3.0
 
 EXPOSE 8080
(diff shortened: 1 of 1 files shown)


xuscrbw (approved):

rtitle:
I'm not sure we should be moving away from the AppSec blessed images: https://github.com/broadinstitute/dsp-appsec-blessed-images

gpcarr:
> https://github.com/broadinstitute/dsp-appsec-blessed-images

@rtitle i don't see a blessed image that has jdk instead of jre (since we want jstack). do you know if any blessed image comes with jdk and jstack? if not then should i reach out to appsec asking for one to be added?

rtitle:
Looks like they are all JREs. FWIW you can take a thread dump with `kill -3 <pid>` on a JRE instead of jstack. Heap dump may require JDK though.

I think AppSec wanted to keep the base images as minimal as possible runtime, so JREs. Worth asking whether we could have a blessed JDK image, I think that makes sense for troubleshooting production issues. Another idea is to keep the base image as is, but install something like https://github.com/apangin/jattach or even a full JDK in the Leo Dockerfile (just to get access to JDK tools).

rtitle:
In fact if you can SSH into a Leo pod you could try just installing https://github.com/apangin/jattach manually

gpcarr:
> https://github.com/apangin/jattach

I tried installing what you suggested in case the team needs to use JDK tools in the future!

xuscrbw (approved):

Dockerfile
```diff
@@ -64,3 +64,7 @@ RUN cd /leonardo && \
 
 # Add Leonardo as a service (it will start when the container starts)
 CMD java $JAVA_OPTS -jar $(find /leonardo -name 'leonardo*.jar')
+
+# Install https://github.com/apangin/jattach to get access to JDK tools
```
rtitle: Nits:

* I'd put this before `CMD`, usually CMD is last
* Does `apt-get` work? Think that's slightly preferred in Dockerfiles

rtitle (approved):
Couple nits but seems like it works -- if tests pass :shipit:

PR #2505, opened 2022-01-18
[No Ticket] Bump dependencies (from scala-steward PRs 2406, 2487, 2502, 2504)

kyuksel:
- `dependency-bot` run failed so taking the manual route.
- Decided not to subsume #2501 as it doesn't even seem to be a patch update.
- `sbt-scalafix` upgrade was to fix `[error] sbt.librarymanagement.ResolveException: Error downloading org.scalameta:semanticdb-scalac_2.13.8:4.4.30` which seemed to be due to the `scalac` upgrade causing a conflict when `sbt-scalafix` isn't also upgraded 🤷🏼 
- Similarly for the `sbt-scoverage` upgrade

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -114,7 +114,7 @@ object Dependencies {
 
   val slick: ModuleID =           "com.typesafe.slick"  %% "slick"                % slickV excludeAll (excludeTypesafeConfig, excludeReactiveStream)
   val hikariCP: ModuleID =        "com.typesafe.slick"  %% "slick-hikaricp"       % slickV excludeAll (excludeSlf4j)
-  val mysql: ModuleID =           "mysql"               % "mysql-connector-java"  % "8.0.22"
+  val mysql: ModuleID =           "mysql"               % "mysql-connector-java"  % "8.0.28"
   val liquibase: ModuleID =       "org.liquibase"       % "liquibase-core"        % "4.6.2"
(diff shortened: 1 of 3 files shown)


project/Dependencies.scala
```diff
@@ -114,7 +114,7 @@ object Dependencies {
 
   val slick: ModuleID =           "com.typesafe.slick"  %% "slick"                % slickV excludeAll (excludeTypesafeConfig, excludeReactiveStream)
   val hikariCP: ModuleID =        "com.typesafe.slick"  %% "slick-hikaricp"       % slickV excludeAll (excludeSlf4j)
-  val mysql: ModuleID =           "mysql"               % "mysql-connector-java"  % "8.0.22"
+  val mysql: ModuleID =           "mysql"               % "mysql-connector-java"  % "8.0.28"
```
Qi77Qi, line 117: This might cause issue with a database that has existing data. Validating that is a bit of work. Maybe we can just try it out, but keep an eye on dev deploy after this PR merge. If there's issue, we can easily revert dev and make a new PR for reverting the upgrade afterwards

Qi77Qi (approved):

project/Dependencies.scala
kyuksel: Ah good to know. What kind of issues with existing DB have you experienced due to a minor upgrade like this?
Sounds good to merge and monitor.

PR #2514, opened 2022-01-26
IA-2965 Fix a bug where files under previous home dir will be lost

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-2965

The fix essentially copies everything under `$HOME` to `notebooks` so that those files will live on PD and re-mapped into the juptyer container the we make a new one, and we shouldn't copy again if this has already happened when next time the VM restarts.

Test case:
1. Create a runtime with pre home dir change branch
2. Create a test file under $HOME
3. Stop and restart runtime with this PR's change
4. Veirfy that the test file still exists under new $HOME dir, and there's no more `notebooks` directory

The above tests is done on both GCE VM and Dataproc cluster. Start/stop runtimes a few times and things still work as expected and `rsync` is only run the first time it starts up with new directory change

---

In all cases:

### DIFF
--- a/http/src/main/resources/init-resources/startup.sh
+++ b/http/src/main/resources/init-resources/startup.sh
@@ -108,7 +108,7 @@ then
     if [ ! -z "$JUPYTER_DOCKER_IMAGE" ] ; then
         echo "Restarting Jupyter Container $GOOGLE_PROJECT / $CLUSTER_NAME..."
 
-        docker exec $JUPYTER_SERVER_NAME /bin/bash -c "[ -d $JUPYTER_USER_HOME/notebooks ] && rsync -av --progress . $JUPYTER_USER_HOME/notebooks --exclude $JUPYTER_USER_HOME/notebooks || true"
+        docker exec $JUPYTER_SERVER_NAME /bin/bash -c "[ -d $JUPYTER_USER_HOME/notebooks ] && [ ! -d $JUPYTER_USER_HOME/notebooks/.jupyter ] && rsync -av --progress . $JUPYTER_USER_HOME/notebooks --exclude $JUPYTER_USER_HOME/notebooks || true"
 
         # Make sure when runtimes restarts, they'll get a new version of jupyter docker compose file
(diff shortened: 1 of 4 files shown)


http/src/main/resources/init-resources/startup.sh
```diff
@@ -108,6 +108,8 @@ then
     if [ ! -z "$JUPYTER_DOCKER_IMAGE" ] ; then
         echo "Restarting Jupyter Container $GOOGLE_PROJECT / $CLUSTER_NAME..."
 
+        docker exec $JUPYTER_SERVER_NAME /bin/bash -c "[ -d $JUPYTER_USER_HOME/notebooks ] && [ ! -d $JUPYTER_USER_HOME/notebooks/.jupyter ] && rsync -av --progress --exclude notebooks . $JUPYTER_USER_HOME/notebooks || true"
```
calbach, line 117: Thanks for the fix. Unfortunately I don't really understand it - why copy HOME (presumably, that's the cwd) -> notebooks ? I thought the notebooks dir went away with your changes?

Qi77Qi: so this line is only for migration. Say you have an existing runtime where jupyter container's PD is mapped at `$HOME/notebooks`, then all jupyter related files (.jupyter, .local) and things like bash history etc all lives under `$HOME`. The home diretory change will make it so that next time this runtime starts up, PD will be mapped to `$HOME`, but this means that the previous  files under $HOME (`.jupyter`, `.local` etc) will be lost....So this one line is to before we restart jupyter container with updated home directory mapping, we will copy all files under `$HOME` to `$HOME/notebooks` first, which will live on PD...So when it starts up, what was previously under $HOME will now appear in new $HOME as well

calbach: I see - thanks! May be worth a comment here

calbach (approved):

jdcanas (approved):

http/src/main/resources/init-resources/startup.sh
kyuksel: +1 on comment :)

Qi77Qi: it's already added

PR #2515, opened 2022-01-26
[IA-3113] Azure backend 

jdcanas:
I'm getting this up here to get some initial eyes on the changes

There's still a few TODOs, but for the most part the code is feature-complete

What's left:
- Any handling of auth (see subtasks for IA-3113)


core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/azureModels.scala
```diff
@@ -12,6 +12,16 @@ final case class AzureCloudContext(tenantId: TenantId,
...
 
+final case class AzureUnimplementedException(message: String) extends Exception {
+  override def getMessage: String = message
+}
+
+final case class ClientId(value: String) extends AnyVal
+final case class ClientSecret(value: String) extends AnyVal
+final case class ManagedAppTenantId(value: String) extends AnyVal
+
+final case class AzureConfig(clientId: ClientId, clientSecret: ClientSecret, managedAppTenantId: ManagedAppTenantId)
```
Qi77Qi: can we call this `AzureAppRegistrationConfig` or sth....`AzureConfig` is pretty vague

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodec.scala
```diff
@@ -435,6 +456,12 @@ object JsonCodec {
     "gpuConfig"
   )((mt, ds, bds, z, gpu) => RuntimeConfig.GceConfig(mt, ds, bds, z, gpu))
 
+  implicit val azureVmConfigDecoder: Decoder[RuntimeConfig.AzureVmConfig] = Decoder.forProduct3(
+    "machineType",
```
Qi77Qi, line 460: do we need to pass vm image as well

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20220118_update_runtime_controlled_resource.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="leonardo" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+
+    <changeSet logicalFilePath="leonardo" author="jdcanas" id="update_runtime_controlled_resource">
+        <addColumn tableName="RUNTIME_CONTROLLED_RESOURCE">
+            <column name="azureName" type="VARCHAR(254)">
```
Qi77Qi: can you remind me why we need this?

http/src/main/resources/reference.conf
```diff
@@ -144,6 +144,43 @@ gce {
...
+     networkNamePrefix = "network"
+     subnetNamePrefix = "subnet"
+     addressSpaceCidr = "192.168.0.0/16"
+     subnetAddressCidr = "192.168.0.0/24"
+     diskControlledResourceDesc = "Azure Disk"
+     vmControlledResourceDesc = "Azure Vm"
+   }
+
+   wsm {
+      uri = "https://localhost:8000"
```
Qi77Qi: this will need to live in firecloud-develop...also the next two config blocks

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/azureModels.scala
```diff
@@ -12,6 +12,16 @@ final case class AzureCloudContext(tenantId: TenantId,
   val asString = s"${tenantId.value}/${subscriptionId.value}/${managedResourceGroupName.value}"
 }
 
+final case class AzureUnimplementedException(message: String) extends Exception {
```
Qi77Qi, line 17: think we can just use the general java `UnsupportedOperationException`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/Config.scala
```diff
@@ -806,4 +810,56 @@ object Config {
       proxyConfig,
       gkeGalaxyDiskConfig
     )
+
+  implicit private val azureInterpConfigReader: ValueReader[AzureInterpretorConfig] = ValueReader.relative { config =>
```
Qi77Qi: would be nice to use [this](https://github.com/DataBiosphere/leonardo/blob/f9007d5b47e9a48be254aced5ae9dc45315b6732/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/ConfigReader.scala) instead of the old config reader style....you won't need to write these readers explicitly with that except for some custom types

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/WsmDao.scala
```diff
@@ -0,0 +1,356 @@
...
+                                     vmSize: VirtualMachineSizeTypes,
+                                     vmImageUri: RuntimeImage,
+                                     ipId: WsmControlledResourceId,
+                                     diskId: WsmControlledResourceId,
+                                     networkId: WsmControlledResourceId)
+
+final case class AzureVmName(value: String) extends AnyVal
+final case class AzureImageUri(value: String) extends AnyVal
+
+final case class WsmVm(resourceId: WsmControlledResourceId)
```
Qi77Qi, line 63: some of these more primitive types probably can be moved to `core`'s models.scala

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/WsmDao.scala
```diff
@@ -0,0 +1,356 @@
...
+      azureContext <- c.downField("azureContext").as[AzureCloudContext]
+    } yield WorkspaceDescription(WorkspaceId(id), displayName, azureContext)
+  }
+
+  implicit val wsmJobStatusDecoder: Decoder[WsmJobStatus] =
+    Decoder.decodeString.emap(s => WsmJobStatus.stringToObject.get(s).toRight(s"Invalid WsmJobStatus found: $s"))
+
+  implicit val wsmJobReportDecoder: Decoder[WsmJobReport] = Decoder.instance { c =>
+    for {
+      id <- c.downField("id").as[String]
```
Qi77Qi: we should create a `Decoder[UUID]` so that you don't have to later doing `UUID.fromString(id)`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BaseCloudServiceRuntimeMonitor.scala
```diff
@@ -566,6 +566,7 @@ abstract class BaseCloudServiceRuntimeMonitor[F[_]] {
         dbRef.inTransaction {
           clusterQuery.mergeInstances(runtimeAndRuntimeConfig.runtime.copy(dataprocInstances = dataprocInstances))
         }.void
+      case CloudService.AzureVm => F.unit
```
Qi77Qi: throw exception here suggesting this should never happen instead of doing nothing silently

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/MonitorAtBoot.scala
```diff
@@ -243,6 +243,8 @@ class MonitorAtBoot[F[_]](publisherQueue: Queue[F, LeoPubsubMessage], computeSer
               Right(
                 LeoLenses.runtimeConfigPrism.getOption(rtConfig).get: RuntimeConfigInCreateRuntimeMessage
               )
+            case _: RuntimeConfig.AzureVmConfig =>
+              throw AzureUnimplementedException("Azure vms should not be handled with existing monitor at boot code")
```
Qi77Qi: I think it's more like "not handled yet" instead of "shoud not be"

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpWsmDao.scala
```diff
@@ -0,0 +1,145 @@
...
+          ),
+        headers = Headers(fakeAuth())
+      )
+    )(onError)
+
+  private def onError(response: Response[F])(implicit ev: Ask[F, AppContext]): F[Throwable] =
+    for {
+      context <- ev.ask
+      body <- response.bodyText.compile.foldMonoid
+      _ <- logger.error(s"${context.traceId} | WSM call failed: $body")
```
Qi77Qi: use `logger.error(context)("???")` so that `traceId` is consistently as a separate field

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzureInterpreter.scala
```diff
@@ -0,0 +1,394 @@
...
+      getWsmJobResult = wsmDao.getCreateVmJobResult(GetJobResultRequest(params.workspaceId, params.jobId))
+      getRuntimeOpt <- azureComputeManager.getAzureVm(params.runtime.runtimeName, params.resourceGroup)
+
+      getRuntime = F.fromOption(
+        getRuntimeOpt,
+        PubsubHandleMessageError.AzureRuntimeError(params.runtime.id,
+                                                   ctx.traceId,
+                                                   None,
+                                                   "Could not retrieve vm controlled resource for runtime from DB")
+      )
```
Qi77Qi: I don't think this is what you want. line 170 will be executed immediately and only that once...it won't be executed repeated in the stream below becuz `getRuntime` will be a static value instead of the action of retrieving the result

it should be
```
getRuntime = azureComputeManager.getAzureVm(params.runtime.runtimeName, params.resourceGroup).flatMap(op => F.fromOptoin(op))
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzureInterpreter.scala
```diff
@@ -0,0 +1,394 @@
...
+
+      taskToRun = for {
+        // first poll the WSM createVm job for completion
```
Qi77Qi: like I mentioned previously...I would put t `F.sleep(30 seconds)` here since the VM creation takes a while and there's no point polling too early

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzureInterpreter.scala
```diff
@@ -0,0 +1,394 @@
...
+
+      _ <- wsmDao.deleteVm(
+        DeleteVmRequest(
+          params.workspaceId,
+          params.vmControlledResource.resourceId,
+          WsmJobControl(WsmJobId(UUID.fromString(uniqueName("delete-job"))))
+        )
+      )
+
+      wsmWorkspace <- wsmDao.getWorkspace(msg.workspaceId)
```
Qi77Qi: same error as mentioned above

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodec.scala
jdcanas: that's stored in the CLUSTER_IMAGES table

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20220118_update_runtime_controlled_resource.xml
jdcanas: I think its important since 
   1) its required for any querying of a controlled resource in azure itself (the ComputeManager `get` calls
   2) to provide readable logging down the line that can be easily mapped to azure entities as opposed to needing to look up IDs from Leo in WSM first

It's not super useful as of this PR (especially since for VM and PDs we have the name stored in their respective tables already), but I think it's a good idea to store. I could see an argument for removing it for now though, not too attached.

http/src/main/resources/reference.conf
jdcanas: Yeah, this is mainly here for reference in terms of all the config parts needed for azure (i.e. for copying into the test reference.conf when I get there). 

I'll make a subtask to add at least placeholders for all the relevant pieces in firecloud-develop

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20220118_update_runtime_controlled_resource.xml
Qi77Qi: Is there an example of this where we don't already store the info? cuz say for VM, the vm name should already exist in `Runtime` table; for disk name, it should already be in `PersistentDisk` table, and for these 2, it'll be duplicate info if we add this column here...so I'm inclined to not have it unless we have an example that we do need this field

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/azureModels.scala
jdcanas: We could, but I like having a custom exception to easily be able to track for future completeness. I could make this exception extend that one. 

Some will be unsupported and some are simply unimplemented for now

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzureInterpreter.scala
jdcanas: Ah thanks, the proper logic was lost when I refactored the dao to return an option

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzureInterpreter.scala
jdcanas: So I looked into `streamUntilDoneOrTimeout` implementation, it actually sleeps for `interval` before the first check so it already does what we want!

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzureInterpreter.scala
jdcanas: This one should be correct as I understand it. 

We're getting the workspace once from WSM simply because we need the static managedResourceGroupName (which should exist from a prior call creating the `azureContext`) to query azure directly about the resource.

Then, we 
```
      getDeleteResult = azureComputeManager.getAzureVm(runtime.runtimeName,
                                                       wsmWorkspace.azureContext.managedResourceGroupName)
```

And poll on that `F[Option[VirtualMachine]]`. Slightly different here because we don't have to flatmap from the option.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/WsmDao.scala
jdcanas: any suggestions? I feel likes the appropriate ones will surface once I start to work on the unit test code. wary to pollute core before the models are needed in other modules

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/Config.scala
jdcanas: How exactly does this method specify the path in the config file its reading from?

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20220118_update_runtime_controlled_resource.xml
jdcanas: For Network and IP we don't store this info anywhere else. That being said, we don't actually need this field for anything yet. I'lll remove it.

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/runtimeModels.scala
```diff
@@ -69,7 +69,8 @@ object Runtime {
           urlBase + cloudContext.asString + "/" + runtimeName.asString + "/" + tool.proxySegment
         )
       case _: CloudContext.Azure =>
-        throw new NotImplementedError("Proxying Azure runtime is not supported yet")
+        //TODO: this is not the correct proxy URL
+        new URL(urlBase)
```
Qi77Qi, line 73: think we should leave this as `NotImplementedError`

http/src/main/resources/reference.conf
Qi77Qi: we should move these to `src/test/resources/reference.conf`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/WsmDao.scala
Qi77Qi: I would put all value classes in [core](https://github.com/DataBiosphere/leonardo/blob/develop/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/azureModels.scala)

This will make later adding automation tests a bit easier

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/WsmDao.scala
```diff
@@ -0,0 +1,351 @@
...
+
+  implicit val vmRequestDataEncoder: Encoder[CreateVmRequestData] =
+    Encoder.forProduct7("name", "region", "vmSize", "vmImageUri", "ipId", "diskId", "networkId")(x =>
+      (x.name.asString,
+       x.region.toString,
+       x.vmSize.toString,
+       x.vmImageUri.imageUrl,
+       x.ipId.value.toString,
+       x.diskId.value.toString,
+       x.networkId.value.toString)
```
Qi77Qi: if you have encoder for these value classes, you shouldn't need to call `toString`..encoders for value classes should be in https://github.com/DataBiosphere/leonardo/blob/develop/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodec.scala

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzureInterpreter.scala
```diff
@@ -0,0 +1,389 @@
+package org.broadinstitute.dsde.workbench
```
Qi77Qi: maybe name the file `AzurePubsubHandlerInterp` or sth? `AzureInterpreter` seems too generic. Same for `AzureAlgebra`

http/src/test/resources/reference.conf
```diff
@@ -290,4 +290,36 @@ akka.ssl-config {
...
 
+azure {
+   monitor {
+    poll-status {
+      max-attempts = 120 # 15 seconds * 120 is 30 min
+      interval = 1 seconds
+    }
+   }
+
+   runtime-defaults {
```
Qi77Qi: for stuff already exists in `reference.conf`, we don't need them here

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriberSpec.scala
```diff
@@ -1606,6 +1622,296 @@ class LeoPubsubMessageSubscriberSpec
     res.unsafeRunSync()(cats.effect.unsafe.IORuntime.global)
   }
 
+  it should "create azure vm properly" in isolatedDbTest {
```
Qi77Qi: shall we have most of these tests in `AzureInterpreterSpec`...the logic in pubsub is pretty straightforward...

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/runtimeModels.scala
jdcanas: I don't think its a good idea to simply throw an exception here because it is used in two places that are invoked prior to proxy URL being implemented where runtimes are created (in DB and test code), i.e. this exception is hit whenever an Azure Runtime is created.

https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala#L790
https://github.com/DataBiosphere/leonardo/blob/develop/http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/CommonTestData.scala#L267

Given the use case of this field is via leo proxy and consumers, I see little risk in including a dummy value with a future TODO.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/WsmDao.scala
jdcanas: I'm a bit wary to move them to core, as I see WSM DAO encoders/decoders as mainly a back-leo implementation details;  I don't see us hitting WSM directly in any integration tests. These are not the request models that automation tests will use when they hit the front-end service.

We can always move them if its relevant

edit: I addressed the above comment, this is in response to your comment [here](https://github.com/DataBiosphere/leonardo/pull/2515#discussion_r797043658) , which I can't directly respond to

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzureInterpreter.scala
jdcanas: I'm just following the `util` package naming convention.

The corresponding GKE, VPC, and Gce equivalents are called `[X]Interpreter` and `[X]Algebra`. Are you suggestion we rename these too? I could definitely get behind that

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriberSpec.scala
jdcanas: Sure. There are two tests that aim to verify a top-level error results in pubsub correctly updating the cluster to error that I'll leave in this spec, but the rest can just use the interp directly.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzureInterpreter.scala
Qi77Qi: I think those are fine for now cuz they're not directly "messageHandlers"...If down the line, we make those more of `messageHandler`, then we can rename them...but yea, naming for some those files are a bit confusing as well

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriberSpec.scala
Qi77Qi: cool..yea..mainly I think this spec is getting a bit out of hand. And unit tests should be closer to what's being tested if possible (unless the top level is the concern)

Qi77Qi (approved):
think there's still one last comment https://github.com/DataBiosphere/leonardo/pull/2515/files#r797052778 I'd prefer addressed before merging...otherwise looks great overall...

👍  assuming tests will pass

jdcanas:
jenkins retest

jdcanas:
jenkins retest, maven errors

jdcanas:
jenkins retest

jdcanas:
jenkins retest

jdcanas:
jenkins retest

jdcanas:
jenkins retest

jdcanas:
jenkins retest, fiab-start failure

Qi77Qi (changes requested):
reference.conf needs to have some of those new dummy Azure configs since `firecloud-develop` will only populate it for dev env

Qi77Qi (approved):
one last comment https://github.com/DataBiosphere/leonardo/pull/2515/files#r797055920

PR #2525, opened 2022-02-04
PROD-635 add a wait before start vm

Qi77Qi:
https://broadworkbench.atlassian.net/browse/PROD-635

We're seeing lots of `RESOURCE_NOT_READY` error when starting VM. Google support says this is due to the setMetadata API and start VM API is called too close to each other. It's odd this starts happening jsut now since the code hasn't changed for a while, but hopefully adding the wait will make things better

---

In all cases:

### DIFF
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -116,6 +116,8 @@ gce {
   }
   gceReservedMemory = 1g
 
+  waitBeforeStartVm = 5 seconds
+
   monitor {
     initialDelay = 20 seconds
     # Defines polling for runtime status transitions. Used commonly for all statuses.
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -116,7 +116,7 @@ gce {
   }
   gceReservedMemory = 1g
 
-  waitBeforeStartVm = 5 seconds
+  waitBeforeStartingVM = 5 seconds
 
   monitor {
     initialDelay = 20 seconds
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GceInterpreter.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GceInterpreter.scala
@@ -30,6 +30,7 @@ import org.typelevel.log4cats.StructuredLogger
 import org.broadinstitute.dsde.workbench.leonardo.http.ctxConversion
 import scala.concurrent.ExecutionContext
 import scala.jdk.CollectionConverters._
+import scala.concurrent.duration._
 
 final case class InstanceResourceConstraintsException(project: GoogleProject, machineType: MachineTypeName)
     extends LeoException(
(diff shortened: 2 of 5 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GceInterpreter.scala
```diff
@@ -370,6 +370,9 @@ class GceInterpreter[F[_]](
         metadataToAdd = metadata,
         metadataToRemove = Set("startup-script-url")
       )
+      // This sleep is added here because as of 2/1/2022, we started seeing increased amount of `RESOURCE_NOT_READY` error.
+      // Google support says this is due to the setMetadata API and start VM API is called too close to each other.
+      _ <- F.sleep(config.gceConfig.waitBeforeStartVM)
       _ <- googleComputeService.startInstance(googleProject,
```
calbach, line 377: Should this also (or instead) have some kind of retry / backoff for the RESOURCE_NOT_READY error?

http/src/main/resources/reference.conf
```diff
@@ -116,6 +116,8 @@ gce {
   }
   gceReservedMemory = 1g
 
+  waitBeforeStartVm = 5 seconds
```
kyuksel: Nit: `waitBeforeStartingVm` may read better

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GceInterpreter.scala
```diff
@@ -370,6 +370,9 @@ class GceInterpreter[F[_]](
         metadataToAdd = metadata,
         metadataToRemove = Set("startup-script-url")
       )
+      // This sleep is added here because as of 2/1/2022, we started seeing increased amount of `RESOURCE_NOT_READY` error.
+      // Google support says this is due to the setMetadata API and start VM API is called too close to each other.
```
kyuksel, line 374: It may be good to reference the Jira ticket and/or the Google ticket for posterity.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GceInterpreter.scala
Qi77Qi: `RESOURCE_NOT_READY` happens without an error to client. The API response will be sth like success, and operation pending. A few min later, on GCP side, `RESOURCE_NOT_READY` will happen, but still when leo pings GCP about "what's the status of this VM", GCP will just respond with the VM status being `TERMINATED`

So there's no way for leo to know that `RESOURCE_NOT_READY` happened

kyuksel (approved):
👍🏼  with minor comments inline

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GceInterpreter.scala
calbach: Ah, I see - thanks.

PR #2534, opened 2022-02-10
[IA-2903] use default SA for GKE nodepools

gpcarr:
Tested by adding unit test and testing in a Fiab. All nodepools that are created when creating galaxy get the newly expected serviceAccount: 
Screen Shot 2022-02-10 at 5 41 29 PM

There are two methods of interest in nodepool creation: [buildGoogleNodepool](https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala#L1219) and [buildLegacyGoodNodepool](https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala#L1253)

`buildGoogleNodepool` is called by [createAndPollNodepool](https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala#L255) for new node pools
`buildLegacyNodepool` is called when we create a k8s cluster. The reason we need to do this is because we use an older google client library to create the GKE cluster because the newer client does not offer `.setWorkloadIdentity` on a k8s cluster. So we need to pass the older Nodepool object to the create cluster request.

Both methods are used. One for cluster creation, the other for nodepool creation. So we update both methods in this PR

---

In all cases:

### DIFF
--- a/automation/hotswap.sh
+++ b/automation/hotswap.sh
@@ -16,7 +16,7 @@ ENV=${2:-dev}
 
 printf "Generating the Leo jar...\n\n"
 # Example: /Users/qi/workspace/leonardo/http/target/scala-2.12/http-assembly-0.1-437ee4a9-SNAPSHOT.jar
-sbt -Dsbt.log.noformat=true "project http" clean assembly
+sbt -J-Xmx4g -J-Xss2M -J-Xms4G -Dsbt.log.noformat=true "project http" clean assembly
 LEO_JAR_PATH=$(ls http/target/scala-2.13/http*)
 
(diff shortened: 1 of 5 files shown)


Qi77Qi (approved):

Qi77Qi:
it does seem like things will work without my commit as well...but I'll leave it in there just so it's more obvious

PR #2540, opened 2022-02-11
IA-3215 add resize2fs in GCE init script

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-3215

In AoU, UI allows users to update size of an existing disk, while also enabling GPU.
image

In this flow, AoU will need
1. call updateDisk API to increase disk size
2. Delete old VM, and recreate a new VM with GPU

In this flow, resize2fs needs to be called when during VM creation. `resize2fs` is pretty low cost cmd, so running it regardless should be fine.

---

In all cases:

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoApiClient.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoApiClient.scala
@@ -359,6 +359,33 @@ object LeonardoApiClient {
         }
     } yield r
 
+  def patchDisk(
+    googleProject: GoogleProject,
+    diskName: DiskName,
+    req: UpdateDiskRequest
+  )(implicit client: Client[IO], authorization: IO[Authorization]): IO[Unit] =
+    for {
+      traceIdHeader <- genTraceIdHeader()
+      authHeader <- authorization
+      r <- client
+        .run(
+          Request[IO](
+            method = Method.PATCH,
+            headers = Headers(authHeader, defaultMediaType, traceIdHeader),
+            uri = rootUri
+              .withPath(Uri.Path.unsafeFromString(s"/api/google/v1/disks/${googleProject.value}/${diskName.value}")),
+            body = req
+          )
+        )
+        .use { resp =>
+          if (!resp.status.isSuccess) {
+            onError(s"Failed to create disk ${googleProject.value}/${diskName.value}")(resp)
+              .flatMap(IO.raiseError)
+          } else
+            IO.unit
+        }
+    } yield r
+
   def createDiskWithWait(googleProject: GoogleProject, diskName: DiskName, createDiskRequest: CreateDiskRequest)(
     implicit client: Client[IO],
     authorization: IO[Authorization]
--- a/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoApiClient.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoApiClient.scala
@@ -379,7 +379,7 @@ object LeonardoApiClient {
         )
         .use { resp =>
           if (!resp.status.isSuccess) {
-            onError(s"Failed to create disk ${googleProject.value}/${diskName.value}")(resp)
+            onError(s"Failed to patch disk ${googleProject.value}/${diskName.value}")(resp)
               .flatMap(IO.raiseError)
           } else
             IO.unit
--- a/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/runtimes/RuntimeCreationDiskSpec.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/runtimes/RuntimeCreationDiskSpec.scala
@@ -10,7 +10,7 @@ import org.broadinstitute.dsde.workbench.google2.{streamFUntilDone, DiskName, Go
 import org.broadinstitute.dsde.workbench.leonardo.DiskModelGenerators._
 import org.broadinstitute.dsde.workbench.leonardo.LeonardoApiClient._
 import org.broadinstitute.dsde.workbench.leonardo.TestUser.{getAuthTokenAndAuthorization, Ron}
-import org.broadinstitute.dsde.workbench.leonardo.http.{PersistentDiskRequest, RuntimeConfigRequest}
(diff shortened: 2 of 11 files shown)


automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/runtimes/RuntimeCreationDiskSpec.scala
```diff
@@ -219,19 +220,29 @@ class RuntimeCreationDiskSpec
...
+                """! cat /home/jupyter/test.txt""".stripMargin
+              notebookPage.executeCell(persistedData).get should include("this should save")
+              val persistedPackage = "! pip show beautifulSoup4"
+              notebookPage.executeCell(persistedPackage).get should include(
+                "/home/jupyter/.local/lib/python3.7/site-packages"
+              )
+
+              val res = notebookPage
+                .executeCell(
+                  "! df -H |grep sdb"
```
calbach: nit: I landed on this command to get a more precise value (size only, home dir only):

```
df -h --output=size $HOME
```

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/runtimes/RuntimeCreationDiskSpec.scala
```diff
@@ -219,19 +220,29 @@ class RuntimeCreationDiskSpec
...
+                )
+                .get
+              res should include("158G")
```
calbach: Nice.. thanks for the integration test.

For this check, I wouldn't be confident this value is stable, since some of the partition is not available after formatting. I imagine this could change slightly across GCE image updates... but it's possible it won't. Allowing some error margin here might be better (e.g. parse numeric value, allow +/- 5)

calbach (approved):

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoApiClient.scala
```diff
@@ -359,6 +359,33 @@ object LeonardoApiClient {
...
+            method = Method.PATCH,
+            headers = Headers(authHeader, defaultMediaType, traceIdHeader),
+            uri = rootUri
+              .withPath(Uri.Path.unsafeFromString(s"/api/google/v1/disks/${googleProject.value}/${diskName.value}")),
+            body = req
+          )
+        )
+        .use { resp =>
+          if (!resp.status.isSuccess) {
+            onError(s"Failed to create disk ${googleProject.value}/${diskName.value}")(resp)
```
kyuksel: Should this be

```suggestion
            onError(s"Failed to patch disk ${googleProject.value}/${diskName.value}")(resp)
```

?

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/runtimes/RuntimeCreationDiskSpec.scala
kyuksel: If `158` is expected to grow, I wonder if it should be derived from `newDiskSize` so we don't have to update another place in the test.

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/runtimes/RuntimeCreationDiskSpec.scala
```diff
@@ -165,7 +165,8 @@ class RuntimeCreationDiskSpec
     val runtimeName = randomeName.copy(asString = randomeName.asString + "pd-spec") // just to make sure the test runtime name is unique
     val runtimeWithDataName = randomeName.copy(asString = randomeName.asString + "pd-spec-data-persist")
     val diskName = genDiskName.sample.get
-    val diskSize = genDiskSize.sample.get
+    val diskSize = DiskSize(110)
```
kyuksel, line 168: Why can't we continue `genDiskSize.sample.get`ing where we could set `newDiskSize` with some delta?

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/runtimes/RuntimeCreationDiskSpec.scala
```diff
@@ -219,19 +220,29 @@ class RuntimeCreationDiskSpec
           }
         })
         _ <- deleteRuntimeWithWait(googleProject, runtimeName, false)
+        _ <- LeonardoApiClient.patchDisk(googleProject, diskName, UpdateDiskRequest(Map.empty, newDiskSize))
+        _ <- IO.sleep(5 seconds)
```
kyuksel, line 224: I wonder if there's another way like monitoring/polling something instead of explicit `sleep`ing.

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/runtimes/RuntimeCreationDiskSpec.scala
Qi77Qi: we have similar check in a few other places. Things seem pretty consistent

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/runtimes/RuntimeCreationDiskSpec.scala
Qi77Qi: the exact disk size from `df -H` will always a bit different from what we set. So having the starting VM size vary will make the validation impossible

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/runtimes/RuntimeCreationDiskSpec.scala
Qi77Qi: `IO.sleep` doesn't block any threads. So I think this is fine....Introducing polling here will add a lot more complexility, which I don't see much benefit in this particular case

kyuksel (approved):

Qi77Qi:
```
Failed to create cluster 20 due to 503 Service Unavailable&#010;POST https://storage.googleapis.com/upload/storage/v1/b/leoinit-automation-test-ap-1e1dbe70-e1aa-4cba-aaeb-1beb1b3b134c/o?name=cluster-site-gce.conf&uploadType=resumable&#010;Service Unavailable,None,2022-02-14T21:45:56Z,None))
```

jenkins retest

PR #2546, opened 2022-02-15
IA-3219 Fix GPU enabled runtime start up issue

Qi77Qi:
The bug is that in startup script, we assume jupyter container is already up and running, which is usually the case. But for GPU, jupyter container will fail to start up until certain volumes are mounted, hence we need to move that code up before we do anything with jupyter container

Tested manually in fiab

---

In all cases:

### DIFF
--- a/http/src/main/resources/init-resources/startup.sh
+++ b/http/src/main/resources/init-resources/startup.sh
@@ -119,12 +119,6 @@ then
 
     # (1/6/22) Restart Jupyter Container to reset `NOTEBOOKS_DIR` for existing runtimes. This code can probably be removed after a year
     if [ ! -z "$JUPYTER_DOCKER_IMAGE" ] ; then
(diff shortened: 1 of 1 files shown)


http/src/main/resources/init-resources/startup.sh
```diff
@@ -85,6 +85,19 @@ FILE=/var/certs/jupyter-server.crt
 USER_DISK_DEVICE_ID=$(lsblk -o name,serial | grep 'user-disk' | awk '{print $1}')
 DISK_DEVICE_ID=${USER_DISK_DEVICE_ID:-sdb}
 
+# Make this run conditionally
```
kyuksel: Nit: The comment above seems unnecessary.

kyuksel (approved):
LGTM 👍🏼 
Haven't tested it.

PR #2548, opened 2022-02-15
[IA-2967] Adding autopauseEnabled value to runtime

cstrandw:
RELATED TO: https://github.com/DataBiosphere/terra-ui/pull/2815

- Changed logic in calculateAutopauseThreshold to accommodate no autopauseEnabled value being passed in on update

Testing:
- Ran tests locally
- Tested manually through the UI and checked corresponding database values for validity
- See https://github.com/DataBiosphere/terra-ui/pull/2815 for other testing there

---

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeServiceInterp.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeServiceInterp.scala
@@ -1014,12 +1014,13 @@ object RuntimeServiceInterp {
                                                    autoFreezeConfig: AutoFreezeConfig): Int =
     autopause match {
       case None =>
-        autoFreezeConfig.autoFreezeAfter.toMinutes.toInt
+        if (autopauseThreshold.isEmpty) autoFreezeConfig.autoFreezeAfter.toMinutes.toInt
+        else autopauseThreshold.get
       case Some(false) =>
         autoPauseOffValue
       case _ =>
         if (autopauseThreshold.isEmpty) autoFreezeConfig.autoFreezeAfter.toMinutes.toInt
-        else Math.max(autoPauseOffValue, autopauseThreshold.get)
+        else autopauseThreshold.get
     }
 }
 
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeServiceInterp.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeServiceInterp.scala
@@ -1013,14 +1013,13 @@ object RuntimeServiceInterp {
                                                    autopauseThreshold: Option[Int],
                                                    autoFreezeConfig: AutoFreezeConfig): Int =
     autopause match {
-      case None =>
-        if (autopauseThreshold.isEmpty) autoFreezeConfig.autoFreezeAfter.toMinutes.toInt
-        else autopauseThreshold.get
       case Some(false) =>
         autoPauseOffValue
       case _ =>
-        if (autopauseThreshold.isEmpty) autoFreezeConfig.autoFreezeAfter.toMinutes.toInt
-        else autopauseThreshold.get
+        autopauseThreshold match {
+          case Some(v) => v
+          case None    => autoFreezeConfig.autoFreezeAfter.toMinutes.toInt
+        }
     }
 }
 
--- a/http/src/main/resources/swagger/api-docs.yaml
+++ b/http/src/main/resources/swagger/api-docs.yaml
@@ -2404,8 +2404,8 @@ components:
         autopause:
           type: boolean
           description: Whether autopause feature is enabled for this specific cluster. If
-            unset, autopause will be enabled and the threshold value will be set to the one 
-            specified or a system default if unset.
+            unset, autopause will be enabled and the threshold value will be set to either the autopauseThreshold value
+            or the system default if autopauseThreshold is also unset.
         autopauseThreshold:
           type: integer
           description: The number of minutes of idle time to elapse before the cluster is
--- a/http/src/main/resources/swagger/api-docs.yaml
+++ b/http/src/main/resources/swagger/api-docs.yaml
@@ -2404,8 +2404,8 @@ components:
         autopause:
           type: boolean
           description: Whether autopause feature is enabled for this specific cluster. If
-            unset, autopause will be enabled and a system default threshold will
-            be used.
+            unset, autopause will be enabled and the threshold value will be set to the one 
+            specified or a system default if unset.
         autopauseThreshold:
           type: integer
           description: The number of minutes of idle time to elapse before the cluster is
--- a/http/src/main/resources/swagger/api-docs.yaml
+++ b/http/src/main/resources/swagger/api-docs.yaml
@@ -4,7 +4,7 @@ info:
   description: |
     Workbench notebooks service.
   # Follow semantic versioning https://semver.org/
-  version: "1.3.2"
+  version: "1.3.3"
   license:
     name: BSD
     url: http://opensource.org/licenses/BSD-3-Clause


core/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/RuntimeRoutesTestJsonCodec.scala
```diff
@@ -81,7 +81,7 @@ object RuntimeRoutesTestJsonCodec {
     "startUserScriptUri",
     "runtimeConfig",
     "userJupyterExtensionConfig",
-    "autopause",
+    "autopauseEnabled",
```
Qi77Qi: this will be breaking change for clients. Terra UI isn't the only client for Leonardo. So I'd prefer we keep the previous name

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20220207_autopause_enabled.xml
```diff
@@ -0,0 +1,19 @@
...
+<databaseChangeLog logicalFilePath="leonardo" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+    <changeSet logicalFilePath="leonardo" author="cstrandw" id="autopause_enabled">
+        <addColumn tableName="CLUSTER">
+            <column name="autopauseEnabled" type="boolean" defaultValue="false">
+            </column>
+        </addColumn>
+    </changeSet>
+
+    <changeSet id="set autopauseEnabled to false when autopauseThreshold is 0, else set to true" author="cstrandw">
+        <sql>UPDATE CLUSTER SET autopauseEnabled = TRUE WHERE autopauseThreshold != 0</sql>
```
Qi77Qi: I don't think we need this field. Like I said in slack, `autopauseThreshold` essentially represents all cases already
* None --> autopause disable
* Some(v) --> autopause enabled, and its threshold is “v”

cstrandw: I was talking with @gpcarr and @jdcanas about whether we should keep the current implementation or add the enabled value to the database, and we agreed that it would best practice to add it instead of relying on deriving the value every time

Qi77Qi: > deriving the value every time

where will be be "deriving" the value?

The downside of having both `autopauseEnabled` and `autopauseThreshold` is you'll like have invalid states. For instance, you have `autopauseEnabled` being `None`, and `autopauseThreshold` being `Some(v)`....this will be invalid. The two fields aren't independent of each other

cstrandw:
> Seeing this formatting change, I just wanted to check if you've been running `sbt scalafmtAll` on your Terminal :)

Yes, that's what I did

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/runtimeRoutesModels.scala
```diff
@@ -103,6 +103,7 @@ final case class GetRuntimeResponse(id: Long,
                                     startUserScriptUri: Option[UserScriptPath],
                                     errors: List[RuntimeError],
                                     userJupyterExtensionConfig: Option[UserJupyterExtensionConfig],
+                                    autopause: Boolean,
```
Qi77Qi: what is this boolean for?

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/runtimeModels.scala
```diff
@@ -35,6 +35,7 @@ final case class Runtime(id: Long,
                          errors: List[RuntimeError],
                          dataprocInstances: Set[DataprocInstance],
                          userJupyterExtensionConfig: Option[UserJupyterExtensionConfig],
+                         autopause: Boolean,
```
Qi77Qi: don't think this is needed?

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/runtimeRoutesModels.scala
cstrandw: I removed the autopause column from the database, but kept it in the models so that it would keep the logic I previously implemented.

Qi77Qi:
used `git reset --merge HEAD~1` to remove the bad merge

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
```diff
@@ -29,6 +29,7 @@ import java.sql.SQLDataException
 import java.time.Instant
 import scala.concurrent.ExecutionContext
 
+// TODO decide if we need to add autopause?
```
kyuksel: Is this comment needed?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeServiceInterp.scala
```diff
@@ -1014,12 +1022,13 @@ object RuntimeServiceInterp {
...
     autopause match {
       case None =>
-        autoFreezeConfig.autoFreezeAfter.toMinutes.toInt
+        autoPauseOffValue
       case Some(false) =>
         autoPauseOffValue
       case _ =>
-        if (autopauseThreshold.isEmpty) autoFreezeConfig.autoFreezeAfter.toMinutes.toInt
-        else Math.max(autoPauseOffValue, autopauseThreshold.get)
+        if (autopauseThreshold.isEmpty) 30 //TODO what behavior do we want when enabled but no threshold
```
kyuksel: Magic number alert :) If we default to a value (like `30` here), it should probably come from a config, or at the least it should be defined somewhere higher.

cstrandw: Planning to change! Talked to Qi about it earlier

core/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/RuntimeRoutesTestJsonCodec.scala
```diff
@@ -95,7 +95,7 @@ object RuntimeRoutesTestJsonCodec {
       x.startUserScriptUri,
       x.runtimeConfig,
       x.userJupyterExtensionConfig,
-      x.autopause,
+      x.autopause.getOrElse(false),
```
Qi77Qi: you don't need `.getOrElse(false)` here

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/runtimeRoutesModels.scala
```diff
@@ -82,7 +82,7 @@ object UpdateRuntimeConfigRequest {
 
 final case class UpdateRuntimeRequest(updatedRuntimeConfig: Option[UpdateRuntimeConfigRequest],
                                       allowStop: Boolean,
-                                      updateAutopauseEnabled: Option[Boolean],
+                                      updateAutopause: Option[Boolean],
```
Qi77Qi: is this rename intended?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
```diff
@@ -758,6 +758,7 @@ object clusterQuery extends TableQuery(new ClusterTable(_)) {
       startUserScriptUri = clusterRecord.startUserScriptUri,
       errors = errors map clusterErrorQuery.unmarshallClusterErrorRecord,
       userJupyterExtensionConfig = extensionQuery.unmarshallExtensions(userJupyterExtensionConfig),
+      autopause = clusterRecord.autopauseThreshold != autoPauseOffValue,
```
Qi77Qi: think we can remove `autopause` from `Runtime` model?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/RuntimeRoutes.scala
```diff
@@ -547,6 +547,7 @@ object RuntimeRoutes {
       ("jupyterStartUserScriptUri", x.startUserScriptUri.asJson),
       ("errors", x.errors.asJson),
       ("userJupyterExtensionConfig", x.userJupyterExtensionConfig.asJson),
+      ("autopause", x.autopause.asJson),
```
Qi77Qi: so I think we can probably remove `autopause` from `GetRuntimeResponse` as well, and just adding this field in the encoder to make UI easier.

you can have
```
val autoPauseEnabled = x.autopauseThreshold != autoPauseOffValue

Json.obj(
...
("autopause", autoPauseEnabled.asJson)
...
```

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/runtimeRoutesModels.scala
cstrandw: @Qi77Qi I want the autopause value to be returned in the response for the UI, so shouldn't I keep this value here and not remove it? Or do you want me to refactor the UI side to work only with the threshold value?

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/runtimeRoutesModels.scala
cstrandw: The rename was intended to help this match the corresponding variable, which is simply autopause.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/RuntimeRoutes.scala
cstrandw: Where is the encoder? Still getting familiar with the codebase

Qi77Qi (approved):
the syntax can be updated a bit...but otherwise looks good

```
autopause match {
      case Some(false) =>
        autoPauseOffValue
      case _ =>
        autopauseThreshold match {
          case Some(v) => v
          case None    => autoFreezeConfig.autoFreezeAfter.toMinutes.toInt
        }
    }
```

PR #2549, opened 2022-02-16
[IA-3112][IA-3115][IA-3116] Azure front end CREATE/GET/DELETE impl

jdcanas:
https://broadworkbench.atlassian.net/browse/IA-3112

This PR implements the front leo logic for the listed azure runtime operations, and includes Sam logic and initial DB record saving.

For reference, this sam config was used heavily for writing the AuthProvider leo code. Hoping to get feedback on that implementation especially since I lack a good way to test... cc @rtitle 
 if you could take a look at the Sam impl 😄 
 https://github.com/broadinstitute/sam/blob/develop/src/main/resources/reference.conf#L351

There are a few TODO's added that are mainly there as conversation points for reviewers, so I'm leaving them for now. 

Additionally, I've added unit tests for the service, but not the routes yet. Trying to get some feedback ahead of time while I work on those tests since its a larger PR


core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodec.scala
```diff
@@ -134,11 +135,26 @@ object JsonCodec {
...
-  implicit val azureRuntimeConfigEncoder: Encoder[RuntimeConfig.AzureVmConfig] = Encoder.forProduct3(
+  implicit val azureRuntimeConfigEncoder: Encoder[RuntimeConfig.AzureConfig] = Encoder.forProduct3(
     "machineType",
     "persistentDiskId",
     "region"
   )(x => (x.machineType, x.persistentDiskId, x.region))
+
+  implicit val azureMachineTypeDecoder: Decoder[VirtualMachineSizeTypes] = Decoder.decodeString.emap { s =>
+    val machineSizeOpt: Option[VirtualMachineSizeTypes] =
+      if (VirtualMachineSizeTypes.values.stream
```
Qi77Qi, line 147: love the validation here

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/azureRoutesModels.scala
```diff
@@ -0,0 +1,26 @@
+package org.broadinstitute.dsde.workbench.leonardo
+
+import com.azure.core.management.Region
+import com.azure.resourcemanager.compute.models.VirtualMachineSizeTypes
+
+final case class CreateAzureRuntimeRequest(labels: LabelMap,
+                                           region: Region,
+                                           machineSize: VirtualMachineSizeTypes,
+                                           imageUri: Option[AzureImageUri],
+                                           customEnvironmentVariables: Option[Map[String, String]],
```
Qi77Qi: I'd prefer this is `Map[String, String]` just so call site doesn't have care about `Option`

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/azureRoutesModels.scala
```diff
@@ -0,0 +1,26 @@
...
+                                           azureDiskConfig: CreateAzureDiskRequest)
+
+final case class CreateAzureDiskRequest(labels: LabelMap,
+                                        name: AzureDiskName,
+                                        size: Option[DiskSize],
+                                        diskType: Option[DiskType],
+                                        //TODO: should we keep blockSize for azureDisks?
+                                        blockSize: Option[BlockSize])
```
Qi77Qi: feels like we never used `blockSize`...maybe jsut remove it?

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/runtimeModels.scala
```diff
@@ -433,9 +439,10 @@ case class DefaultRuntimeLabels(runtimeName: RuntimeName,
 object DefaultRuntimeLabels {
   // Creating a dummy instance to obtain the default label keys
   val defaultLabelKeys = DefaultRuntimeLabels(RuntimeName(""),
-                                              GoogleProject(""),
-                                              WorkbenchEmail(""),
+                                              Some(GoogleProject("")),
```
Qi77Qi, line 439: think we can just use `None` for optional values?

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/samModels.scala
```diff
@@ -175,6 +186,31 @@ object AppAction {
...
+  val stringToAction: Map[String, WorkspaceAction] =
+    sealerate.collect[WorkspaceAction].map(a => (a.asString, a)).toMap
+}
+
+sealed trait WsmResourceAction extends Product with Serializable {
+  def asString: String
+}
+object WsmResourceAction {
+  final case object Write extends WsmResourceAction {
+    val asString = "write"
```
Qi77Qi, line 207: will there be `read` as well?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/Boot.scala
```diff
@@ -180,13 +181,22 @@ object Boot extends IOApp {
                                 appDependencies.publisherQueue,
                                 appDependencies.googleDependencies.googleComputeService)
 
+      val azureService = new AzureServiceInterp[IO](
+        runtimeServiceConfig,
+        ConfigReader.appConfig.persistentDisk,
```
Qi77Qi: maybe create a `AzureServiceConfig` case class?
`final case class AzureServiceConfig(runtimeServiceConfig: ???, diskConfig:???)`

That way, if in the future we need to add more configs, it'll be one config object instead of multiple config objects in the constructor of `AzureServiceInterp`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/RuntimeV2Routes.scala
```diff
@@ -0,0 +1,178 @@
...
+                            updateAzureRuntimeHandler(userInfo, workspaceId, runtimeName, req)
+                          )
+                        }
+                      } ~ delete {
+                        complete(
+                          deleteAzureRuntimeHandler(userInfo, workspaceId, runtimeName)
+                        )
+                      }
+                    } //~
+//                      path("stop") {
```
Qi77Qi: maybe just have `NonImplemented` exception for these unimplemented routes?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AzureServiceInterp.scala
```diff
@@ -0,0 +1,351 @@
...
+                             runtimeName: RuntimeName,
+                             workspaceId: WorkspaceId,
+                             req: CreateAzureRuntimeRequest)(implicit as: Ask[F, AppContext]): F[Unit] =
+    for {
+      context <- as.ask
+
+      azureContext <- wsmDao.getWorkspace(workspaceId).map(_.azureContext)
+      cloudContext = CloudContext.Azure(azureContext)
+
+      samResource = WorkspaceResourceSamResourceId(UUID.randomUUID().toString)
```
Qi77Qi: minor, but typically `UUID.randomUUID()` should be in a `F` context as well

so
```
uuid <- F.delay(UUID.randomUUID())
samResource = WorkspaceResourceSamResourceId(uuid.toString)
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AzureServiceInterp.scala
```diff
@@ -0,0 +1,351 @@
...
+        case None =>
+          for {
+            diskToSave <- F.fromEither(
+              convertToDisk(
+                userInfo,
+                cloudContext,
+                DiskName(req.azureDiskConfig.name.value),
+                azureDiskConfig,
+                req,
+                Instant.now()
```
Qi77Qi: use 
```
instant <- F.nowInstant
```
instead of `Instant.now()`, or maybe just use `context.now`?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AzureServiceInterp.scala
```diff
@@ -0,0 +1,351 @@
...
+              now = context.now,
+              workspaceId = Some(workspaceId)
+            )
+            savedRuntime <- clusterQuery.save(runtimeToSave).transaction
+            _ <- publisherQueue.offer(
+              CreateAzureRuntimeMessage(savedRuntime.id, workspaceId, runtimeImage, Some(context.traceId))
+            )
+          } yield ()
+      }
+      //TODO: I don't  think we notify auth provider, that's WSM's job?
```
Qi77Qi: re TODO, I think so

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AzureServiceInterp.scala
```diff
@@ -0,0 +1,351 @@
...
+
+      hasPermission <- authProvider.hasPermission(WsmResourceSamResourceId(azureRuntimeControlledResource.resourceId),
+                                                  WsmResourceAction.Write,
+                                                  userInfo)
+
+      _ <- context.span.traverse(s => F.delay(s.addAnnotation("Done auth call for delete azure runtime permission")))
+      _ <- if (hasPermission) F.unit
+      else F.raiseError[Unit](RuntimeNotFoundException(cloudContext, runtimeName, "permission denied"))
+
+      _ <- if (runtime.status.isDeletable) F.unit
```
Qi77Qi: minor, but these can be written this way too
```
      _ <- F
        .raiseError[Unit](RuntimeCannotBeDeletedException(cloudContext, runtime.clusterName, runtime.status))
        .whenA(!runtime.status.isDeletable)
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AzureServiceInterp.scala
```diff
@@ -0,0 +1,351 @@
...
+      else
+        Right(allLabels)
+    } yield PersistentDisk(
+      DiskId(0),
+      cloudContext,
+      ZoneName(req.region.toString),
+      diskName,
+      userInfo.userEmail,
+      //TODO: should this be something else?
+      PersistentDiskSamResourceId("fakeUUID"),
```
Qi77Qi, line 271: I think WSM probably creates a sam resource for this. Not sure if this info is returned from WSM, but I think we probably do need this for `getDisk` API

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/leoPubsubMessageSubscriberModels.scala
```diff
@@ -320,6 +320,7 @@ object LeoPubsubMessage {
   }
 
   final case class DeleteAzureRuntimeMessage(runtimeId: Long,
+                                             diskId: DiskId,
```
Qi77Qi: how about we make this optional? I think eventually do want to make it possible to keep disk

Qi77Qi (commented):
looks great overall!

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/samModels.scala
jdcanas: So when we create resources, we give the `creator` the wsm `editor` permission, which automatically gives them `read` and `write` roles in sam. So, if we assume leo is creating them, users should have both. I'll add read here and use it as the permission checked in `getRuntime` though since it seems more semantically correct

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AzureServiceInterp.scala
jdcanas: Ah, I do like how this reads better

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AzureServiceInterp.scala
jdcanas: We talked about punting on disk interoperability because of various sam/service account concerns. Tracked in this ticket https://broadworkbench.atlassian.net/browse/IA-3213

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/leoPubsubMessageSubscriberModels.scala
jdcanas: Hmm... well, currently, WSM doesn't support this as calling delete on the azure runtime deletes all associated child resources. I think we should keep this as non-optional since it wouldn't make much sense to allow keeping the leo record when the WSM record and azure resource are always deleted

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/runtimeModels.scala
jdcanas: We can't here, we use the default labels to validate which keys are illegal for labels in the create/update requests

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/leoPubsubMessageSubscriberModels.scala
Qi77Qi: I don't mean this needs to be cascading down to WSM call...It's just having this as `Option` in the message so that down the line, it'll be easier to support the case where we'd like to keep the disk. For WSM call, we can just do `diskOpt.traverse(x => ???)`

jdcanas:
jenkins retest

Qi77Qi (approved):
👍 :shipit: once tests pass

jdcanas:
jenkins retest

jdcanas:
jenkins retest - was aborted

PR #2554, opened 2022-02-18
IA-3223 more getting rid of fullClusterQuery

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-3223

* Moved some queries using `fullClusterQuery` usage to test since they're only used in test. Slowly I think we should deprecate use of `fullClusterQuery` if possible
* Removed `clusterQuery.getClusterById(runtime.id).transaction` from `MonitorAtBoot`, which uses `fullClusterQuery`
* Remove `dataprocInstances` field from `Runtime`. This allows us to avoid joining Instance table in retrieving runtime in all the places I'm aware of. This in theory should make leo much more stable when users are using large dataproc clusters...We might be able to just avoid persisting non master instances entirely, but I'll leave that to another time

---

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
@@ -10,7 +10,11 @@ import org.broadinstitute.dsde.workbench.leonardo.db.LeoProfile.api._
 import org.broadinstitute.dsde.workbench.leonardo.db.LeoProfile.dummyDate
 import org.broadinstitute.dsde.workbench.leonardo.db.LeoProfile.mappedColumnImplicits._
 import org.broadinstitute.dsde.workbench.leonardo.db.RuntimeConfigQueries.runtimeConfigs
-import org.broadinstitute.dsde.workbench.leonardo.monitor.{RuntimeToAutoPause, RuntimeToMonitor}
+import org.broadinstitute.dsde.workbench.leonardo.monitor.{
+  ExtraInfoForCreateRuntime,
+  RuntimeToAutoPause,
+  RuntimeToMonitor
+}
 import org.broadinstitute.dsde.workbench.model.google.{
   parseGcsPath,
   GcsBucketName,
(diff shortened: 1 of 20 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/MonitorAtBoot.scala
```diff
@@ -210,46 +210,56 @@ class MonitorAtBoot[F[_]](publisherQueue: Queue[F, LeoPubsubMessage], computeSer
...
-                x.gpuConfig
-              ): RuntimeConfigInCreateRuntimeMessage
-            case _: RuntimeConfig.DataprocConfig =>
-              Right(
-                LeoLenses.runtimeConfigPrism.getOption(rtConfig).get: RuntimeConfigInCreateRuntimeMessage
+        val message: Either[String, RuntimeConfigInCreateRuntimeMessage] = runtime.runtimeConfig match {
+          case x: RuntimeConfig.GceConfig =>
+            for {
+              bootDiskSize <- x.bootDiskSize.toRight(
+                s"disk Size field not found for ${runtime.id}. This should never happen"
```
kyuksel:
```suggestion
                s"Disk size field was not found for ${runtime.id}. This should never happen."
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
```diff
@@ -356,13 +348,39 @@ object clusterQuery extends TableQuery(new ClusterTable(_)) {
       .on(_._1.id === _.clusterId)
       .result map { recs =>
       recs.map { rec =>
-        RuntimeToMonitor(rec._1._1.id,
-                         rec._1._2.runtimeConfig.cloudService,
-                         rec._1._1.status,
-                         rec._2.map(_.inProgress).getOrElse(false))
+        val asyncFileds = (rec._1._1.googleId, rec._1._1.operationName, rec._1._1.stagingBucket).mapN {
```
kyuksel: asyncFi**el**ds

kyuksel (approved):
LGTM 👍🏼 A couple of typo comments inline.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/DataprocRuntimeMonitor.scala
```diff
@@ -472,10 +472,6 @@ class DataprocRuntimeMonitor[F[_]: Parallel](
           _ <- setStagingBucketLifecycle(runtimeAndRuntimeConfig.runtime,
                                          config.runtimeBucketConfig.stagingBucketExpiration)
 
-          _ <- dbRef.inTransaction {
```
Qi77Qi, line 475: this is when we delete a runtime. I don't think we need to update instances when we delete cluster. Hence deleting this

rtitle (approved):
Nice change… this looks good to me. 👍 Yeah makes a lot of sense to remove `dataprocInstances` from `Runtime` and just query it when we need to.

PR #2559, opened 2022-02-24
IA-3222 fix overwrite bug

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-3222

Huge thanks to @jdcanas and @xuscrbw for debugging help.

The direct cause to the error message is becuz lock expiration, which is caused becuz sometimes UI fails to call `/objects/lock` api periodically.

We notice that this happens when a new runtime is restarted, and `/etc/jupyter/nbconfig/notebook.json` disappears after runtime restarts.

So the root cause is in home directory change PR, we're removing old jupyter container and restarting new one. In the new jupyter container, `/etc/jupyter/nbconfig/notebook.json` will no longer be there.

So the fix is
1. Don't remove the existing jupyter container if there's no home directory migration needed
2. When home directory migration is needed, then we'll recreate the `notebook.json` file

Tests:
1. Tested out in fiab that a new runtime works as expected after restart

---

In all cases:


xuscrbw (approved):
LGTM

PR #2612, opened 2022-04-06
[IA-3268] list azure runtime

jdcanas:
https://broadworkbench.atlassian.net/browse/IA-3268

Implement listRuntimeV2.

For the most part, this list runtime is fully functional and non-azure specific. As such, a lot of the lines here are duplicated runtimeV1 test code so that when we start switching over we have the same level of test coverage.


http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -388,7 +388,7 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/ErrorReport"
-  /api/google/v1/runtimes/{googleProject}/{name}/start:
+  "/api/google/v1/runtimes/{googleProject}/{name}/start":
```
Qi77Qi, line 391: I don't think the quotes are necessary?

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -432,7 +432,192 @@ paths:
...
+            type: string
+      responses:
+        "200":
+          description: List of runtimes
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: "#/components/schemas/ListRuntimeResponse"
```
Qi77Qi, line 543: should `ListRuntimeResponse` be a different one?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/WsmDao.scala
```diff
@@ -277,12 +285,16 @@ object WsmDecoders {
...
+
   implicit val getWorkspaceResponseDecoder: Decoder[WorkspaceDescription] = Decoder.instance { c =>
     for {
       id <- c.downField("id").as[UUID]
       displayName <- c.downField("displayName").as[String]
-      azureContext <- c.downField("azureContext").as[AzureCloudContext]
-    } yield WorkspaceDescription(WorkspaceId(id), displayName, azureContext)
+      azureContext <- c.downField("azureContext").as[Option[AzureCloudContext]]
+      gcpContext <- c.downField("gcpContext").as[Option[WsmGcpContext]]
+    } yield WorkspaceDescription(WorkspaceId(id), displayName, azureContext, gcpContext.map(_.projectId))
```
Qi77Qi: can we have a `Decoder[WorkspaceId]`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala
```diff
@@ -179,27 +204,75 @@ class AzureServiceInterp[F[_]: Parallel](config: RuntimeServiceConfig,
...
+          for {
+            uuid <- Either.catchNonFatal(UUID.fromString(r.samResource.resourceId))
+          } yield WsmResourceSamResourceId(WsmControlledResourceId(uuid))
+        }
+        .flatMap(either =>
+          either match {
+            case Right(id) => List(id)
+            case Left(_)   => List.empty
+          }
+        )
```
Qi77Qi: this means /api/v2/runtimes won't show runtimes whose sam resource Id is not UUID. Can we make sure that in prod, all active runtimes do have UUID as Sam resource ID?

jdcanas:
jenkis retest, fiab-start failure

jdcanas:
jenkins retest, fiab-start failure due to sam

http/src/main/resources/swagger/api-docs.yaml
jdcanas: They aren't but the majority of our routes have them in this file, so i added for consistency

http/src/main/resources/swagger/api-docs.yaml
jdcanas: It is the same json entity. Trying to keep it consistent (the underlying runtime model is the same and all) so its easier for the consumer.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala
jdcanas: Hmm... that doesn't seem to be the case.

I'm actually realizing this logic is pretty bad as it assumes all runtimes in the future will be registered with WSM, which isn't the case. Unfortunately, I think we will need 2 sam calls for this to:
1.  support azure runtimes registered as a WSM controlled resource
2. a fallback to check all non-creator-is-caller in sam the way the current runtime APIs do (as we have no plans to migrate google runtimes to be registered in WSM) 

I'll push up the code shortly

Qi77Qi:
one thing I notice is `checkSamPermission` is used in `deleteRuntime` as well...we should probably check `writer` permission for delete, but `checkSamPermission` is only checking `reader`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/RuntimeServiceDbQueries.scala
```diff
@@ -167,7 +166,37 @@ object RuntimeServiceDbQueries {
         .filter(_.cloudContextDb === p.asCloudContextDb)
         .filter(_.cloudProvider === p.cloudProvider)
     )
-    val runtimeQueryJoinedWithLabel = runtimeLabelQuery(clusterQueryFilteredByProject)
+
+    joinAndFilterByLabelForList(labelMap, clusterQueryFilteredByProject)
+  }
+
+  def listRuntimes(labelMap: LabelMap,
```
Qi77Qi: can we give this one a different name...overloading functions is generally discouraged in Scala

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala
```diff
@@ -179,27 +202,79 @@ class AzureServiceInterp[F[_]: Parallel](config: RuntimeServiceConfig,
...
+              case Left(_)   => List.empty
+            }
+          } yield WsmResourceSamResourceId(WsmControlledResourceId(uuid))
+        }
+
+      samVisibleWsmControlledResourceSamIds <- NonEmptyList
+        .fromList(wsmControlledResourceSamIds)
+        .fold[F[List[WsmResourceSamResourceId]]](F.pure(List.empty)) { ids =>
+          authProvider.filterUserVisible(ids, userInfo)
+        }
```
Qi77Qi: can do this instead

```
NonEmptyList
        .fromList(wsmControlledResourceSamIds)
        .traverse { ids =>
          authProvider.filterUserVisible(ids, userInfo)
        }
```

Qi77Qi: can do this instead

```
NonEmptyList
        .fromList(wsmControlledResourceSamIds)
        .traverse { ids =>
          authProvider.filterUserVisible(ids, userInfo)
        }
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala
```diff
@@ -179,27 +202,79 @@ class AzureServiceInterp[F[_]: Parallel](config: RuntimeServiceConfig,
...
+
+      // We must also check the RuntimeSamResourceId in sam to support already existing and newly created google runtimes
+      samVisibleRuntimeSamResourceIds <- NonEmptyList
+        .fromList(runtimesUserIsNotCreator.map(_.samResource))
+        .fold[F[List[RuntimeSamResourceId]]](F.pure(List.empty)) { ids =>
+          authProvider.filterUserVisible(ids, userInfo)
```
Qi77Qi: does this still have equivalent of project level fallback? It doesn't like this logic do anything special if the caller is workspace owner?

Qi77Qi: does this still have equivalent of project level fallback? It doesn't like this logic do anything special if the caller is workspace owner?

jdcanas: We do check if the caller is the creator of the runtime, but I'm not sure exactly what a 'project fallback' entails from the code. Does it check the owner of the project? 

So the corollary would be to check if the user is he owner of the workspace in addition to the sam runtime permission we are checking now for this use case?

Qi77Qi: sry just saw the comment,

>So the corollary would be to check if the user is he owner of the workspace in addition to the sam runtime permission we are checking now for this use case?

yes!

jdcanas: cc @Qi77Qi

Qi77Qi (approved):
lgtm..would be good to verify things on Dev

PR #2618, opened 2022-04-11
[IA-3320] create individual http client

Qi77Qi:
Today we're sharing one http client for all DAOs. It has been fine, but we don't have to use the same one. This PR makes it so that each DAO has its own http client, and depends on the DAO, we'll generate metrics for each of them. Another benefit is that if Sam is down, and it takes up all the connection pool, Leo will still be able to make requests to other services. Having metrics for each DAO will also help us identify more easily where things might be wrong.

I added http4s client metrics for the shared client recently. Here's what it'll look like

Screen Shot 2022-04-11 at 10 15 11 AM

---

In all cases:


xuscrbw (approved):
lgtm

jdcanas (approved):

Qi77Qi:
jenkins retest

.github/workflows/unit_test.yml
```diff
@@ -43,10 +44,9 @@ jobs:
 
     - name: Run tests
       env:
-        JAVA_OPTS: -Xmx4G -Xms4G -Xss2M
+        JAVA_OPTS: -Xmx4G -Xms4G -Xss2M -Duser.timezone=UTC -Denv.type=test -Dmysql.host=localhost -Dmysql.port=3307
```
jdcanas, line 47: Hmm... all the test failures have to do with the created date being off from the expected value it seems. Could it be this time zone flag?

Qi77Qi: it's not timezone...for some reason, one timestamp is more precise, see 
```
org.scalatest.exceptions.TestFailedException: Some(DataprocInstance(DataprocInstanceKey(dsp-leo-test,ZoneName(my-zone),InstanceName(master-instance)),12345,Running,Some(IP(1.2.3.4)),Master,2022-04-12T16:12:06.553181Z)) did not equal Some(DataprocInstance(DataprocInstanceKey(dsp-leo-test,ZoneName(my-zone),InstanceName(master-instance)),12345,Running,Some(IP(1.2.3.4)),Master,2022-04-12T16:12:06.553180524Z))
```

PR #2630, opened 2022-04-20
[No ticket] bump image version

jdcanas:


Qi77Qi (approved):
we typically pre-run the job https://fc-jenkins.dsp-techops.broadinstitute.org/job/leonardo-custom-image-rtitle/... but approving assuming image will be updated in a follow up PR

PR #2634, opened 2022-04-21
TOAZ-83 add param to docker

sergiygetlin:
Please disregard this PR. Take a look at this one - https://github.com/DataBiosphere/leonardo/pull/2651

https://broadworkbench.atlassian.net/browse/TOAZ-83

1)Add network parameter to docker
2) Jyputer is running on behalf of non-root user. User doesn't have membership in sudo and docker groups. So, it can't manage docker and run administrative command on behalf of root user.
3) Shutdown command is no longer needed.

---

In all cases:


http/src/main/resources/init-resources/msdsvmcontent/ms_dsvm_script_v5_cat.sh
```diff
@@ -23,7 +23,7 @@ LISTENER_DOCKER_IMAGE=$8
 
 # Define environment variables for Jupyter Server customization
 
-SERVER_APP_PORT=8080
+SERVER_APP_PORT=8888
```
Qi77Qi, line 39: any particular reason why we need to change port?

http/src/main/resources/reference.conf
```diff
@@ -67,7 +67,7 @@ dataproc {
     region = "us-central1"
   }
 
-  customDataprocImage = "projects/broad-dsp-gcr-public/global/images/leo-0322-2-image-dataproc-2-0-27-debian10-52fca9d"
+  customDataprocImage = "projects/broad-dsp-gcr-public/global/images/custom-leo-image-dataproc-2-0-27-debian10-aa373a8"
```
Qi77Qi: can you rebase properly to get rid of non related commits

http/src/main/resources/init-resources/msdsvmcontent/ms_dsvm_script_v5_cat.sh
sergiygetlin: Yes. RelayListener works on 8080. Since we introduce --network host it interferes with the the Jupyter port which is also 8080.

http/src/main/resources/reference.conf
sergiygetlin: Done.

http/src/main/resources/init-resources/msdsvmcontent/ms_dsvm_script_v5_cat.sh
```diff
@@ -49,29 +65,25 @@ RELAY_CONNECTIONSTRING="Endpoint=sb://${RELAY_NAME}.servicebus.windows.net/;Shar
...
 
-docker run -d --restart always --name RelayListener \
+docker run -d --restart always --network host --name RelayListener \
 --env LISTENER_RELAYCONNECTIONSTRING=$RELAY_CONNECTIONSTRING \
 --env LISTENER_RELAYCONNECTIONNAME=$RELAY_CONNECTION_NAME \
---env LISTENER_TARGETPROPERTIES_TARGETHOST="http://${RELAY_TARGET_HOST}:8080" \
+--env LISTENER_TARGETPROPERTIES_TARGETHOST="http://${RELAY_TARGET_HOST}:8888" \
 $LISTENER_DOCKER_IMAGE
+
+sudo shutdown -r 1
```
Qi77Qi: what does this do?

sergiygetlin: This command schedules restart of the vm. @erdalcosgun could you please add more context why do we need it?

Qi77Qi: more importantly, can we avoid the restart?

erdalcosgun: @Qi77Qi  We have changed the ownership of the Jupyter engine. Therefore, we should keep this command. We configured the VM with custom script and re-start. There is no lag on provisioning the Jupyter service. This re-start will not affect the end-user experience.

sergiygetlin: So, shutdown command for restarting VM is no longer needed. It has been removed. Initial assumption was that ownership is not fully propagated that's why we need restart. But it turns out that the real problem was with the "runuser" command on line 66. It actually failed to run jupyter, but after restart it worked because of correct crontab configuration.

PR #2635, opened 2022-04-22
IA-3337 Fix deletion bug

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-3337

The bug is that when users runtime creation fail, and then they try to delete runtime, `detach` disk will fail becuz the runtime doesn't exist, hence throwing exception right away. A second similar bug is that in back leo's delete runtime handler, we're erroring out when `setMetadata` call fails becuz the runtime doesn't exist. In both cases, we should ignore the exception to let delete flow to go through

---

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreter.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreter.scala
@@ -351,6 +351,7 @@ class DataprocInterpreter[F[_]: Parallel](
   ): F[Option[OperationFuture[Operation, Operation]]] =
     if (params.runtimeAndRuntimeConfig.runtime.asyncRuntimeFields.isDefined) { //check if runtime has been created
       for {
+        ctx <- ev.ask
         region <- F.fromOption(
           LeoLenses.dataprocRegion.getOption(params.runtimeAndRuntimeConfig.runtimeConfig),
           new RuntimeException("DataprocInterpreter shouldn't get a GCE request")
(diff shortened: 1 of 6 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreter.scala
```diff
@@ -357,18 +357,41 @@ class DataprocInterpreter[F[_]: Parallel](
...
         metadata <- getShutdownScript(params.runtimeAndRuntimeConfig, false)
         _ <- params.masterInstance.traverse { instance =>
-          googleComputeService
-            .addInstanceMetadata(instance.key.project, instance.key.zone, instance.key.name, metadata)
+          for {
+            opFutureAttempt <- googleComputeService
+              .addInstanceMetadata(instance.key.project, instance.key.zone, instance.key.name, metadata)
+              .attempt
+            _ <- opFutureAttempt match {
+              case Left(e) if e.getMessage.contains("Instance not found") =>
```
jdcanas, line 366: We might want to log something here in addition to no-op. Might be worth adding to the googleComputeService itself as well, but don't feel super strong about it since it would require a wb-libs PR too, and users are actively experiencing the issue.

jdcanas (approved):

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreter.scala
Qi77Qi: I think from library perspective, if a VM doesn't exist, and client tries to update metadata, then it should error out. Do nothing might not be applicable to all cases..For instance, during starting runtime flow, we do want to error out if `addInstanceMetadata` fails becuz it can't find the runtime

Qi77Qi:
```
16:34:59 ERROR: for martha-proxy  Cannot start service martha-proxy: driver failed programming external connectivity on endpoint firecloud_martha-proxy_1 (0714bdc6309e41842b92b41c136a82494d942131c83f1f94afe68357d06e3935): Error starting userland proxy: listen tcp4 0.0.0.0:32800: bind: address already in use
```

jenkins retest

Qi77Qi:
jenkins retest

PR #2641, opened 2022-04-27
[IA-3260] Fix deletion and misc changes

jdcanas:
Currently, deletion is over-eager to error-out when something unexpected happens instead of deleting as much as possible given the system state. 

The changes here seem desirable to prevent users from getting stuck with Errored runtimes. I have gotten stuck a few times during UI development due to partial resource creation, which results in the deletion code erroring out and not actually letting me delete my runtime. This requires manual editting the database to recover from.

Another odd thing I ran that this PR addresses is the frontend code hitting and error at some point in the async flow, and then the backend code hitting a somewhat related error. I addressed this by including the back leo message sending sequentially in the async task, such that it won't be sent if front leo encounters an error in the proceeding steps.  Otherwise, what happens in the case of an error in the async task and an error in back leo is the runtime ends up with two errors in the database, and its very confusing to debug without digging deep into stackdriver with timestamps with frontend/backend log labels; especially since the UI truncates and only displays the first error currently.

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala
@@ -246,11 +246,9 @@ class RuntimeV2ServiceInterp[F[_]: Parallel](config: RuntimeServiceConfig,
       // For now, azure disk life cycle is tied to vm life cycle and incompatible with disk routes
       _ <- persistentDiskQuery.markPendingDeletion(diskId, ctx.now).transaction
 
-      _ <- wsmResourceIdOpt.traverse { wsmResourceId =>
-        publisherQueue.offer(
-          DeleteAzureRuntimeMessage(runtime.id, Some(diskId), workspaceId, wsmResourceId, Some(ctx.traceId))
-        )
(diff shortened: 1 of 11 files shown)


Qi77Qi (approved):
thanks @jdcanas ! lgtm

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala
```diff
@@ -118,18 +118,22 @@ class RuntimeV2ServiceInterp[F[_]: Parallel](config: RuntimeServiceConfig,
             runtimeToSave = SaveCluster(cluster = runtime, runtimeConfig = runtimeConfig, now = ctx.now)
             savedRuntime <- clusterQuery.save(runtimeToSave).transaction
 
-            task = Task(
+            task = for {
+              _ <- createRuntime(CreateAzureRuntimeParams(workspaceId, savedRuntime, runtimeConfig, disk, runtimeImage),
+                                 WsmJobControl(createVmJobId))
+              _ <- publisherQueue.offer(
```
Qi77Qi: if this is the change that's causing unit tests to hang, I would just leave things as is cuz it'll be fixed by a PR I'll have soon

PR #2642, opened 2022-04-27
BW-1220 Changes to allow pets to create apps

cjllanwarne:
~In draft form while I test that this works and get early feedback.~
With this PR, pet SA will be able to create apps. Getting app details and deleting app using pet SA already existed. Adding support for pet SA for remaining APIs - list app, start and stop app, will be done in separate ticket.

https://broadworkbench.atlassian.net/browse/BW-1220


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
```diff
@@ -71,8 +71,11 @@ final class LeoAppServiceInterp[F[_]: Parallel](
       )
 
       samResourceId <- F.delay(AppSamResourceId(UUID.randomUUID().toString))
+
+      // Look up the original email in case this API was called by a pet SA:
+      originatingUserEmail <- authProvider.lookupOriginatingUserEmail(userInfo)
```
rtitle, line 76: ~We might want to do this further up the stack in the route layer, so it applies to all `/api` requests. Maybe in https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/StandardUserInfoDirectives.scala ?~

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/SamAuthProvider.scala
```diff
@@ -207,6 +208,24 @@ class SamAuthProvider[F[_]: OpenTelemetryMetrics](
   )(implicit sr: SamResource[R], ev: Ask[F, TraceId]): F[Unit] =
     samDao.deleteResource(samResource, creatorEmail, googleProject)
 
+  override def lookupOriginatingUserEmail[R](petOrUserInfo: UserInfo)(implicit ev: Ask[F, TraceId]): F[WorkbenchEmail] =
```
rtitle, line 211: This sam call also returns an `enabled` flag -- maybe Leo should read that flag and 401 if the user is disabled (this is essentially replacing the OpenDJ check today).

Maybe we could do some refactoring in this DAO -- instead of separate methods to get subject/email maybe we could have 1 method that calls Sam `/register/user` and returns a `case class SamUserInfo(email, subject, enabled)`, and callers can just use that object.

rtitle:
I think the change seems correct, just a couple suggestions above.

It might also be good to quantify how many additional Sam calls this adds (hopefully it should be ok because this API is pretty cheap).

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
rtitle: Actually thinking more… this is probably not needed for every request, because we already make Sam checks for those. Maybe we should just do this in places where we create a pet -- e.g. createApp and createRuntime. (So this PR is correct, suggest just adding it to createRuntime as well :))

rtitle:
Also cross-posting the OpenDJ deprecation plan which touches on this at the bottom: https://docs.google.com/document/d/1XTSne7Yd3hUsjK3yG3bNwaq9LzuDxxjMvmVEsAyYEP8/edit#

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
salonishah11: ~So just before we [create a UserInfo object on line 20 here](https://github.com/DataBiosphere/leonardo/blob/8972cd97e1df768eb9f07ab03c4afb97e4bda383/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/StandardUserInfoDirectives.scala#L20), we want to check that (1) the user is enabled and (2) if a pet SA was used to submit the request, we want to map to the user's email address and use that for all the requests moving forward?~
Update: I just saw the second message. For some reason it hadn't loaded when I wrote ^ comment. I will add this to `createRuntime` as well. And I think we might need this in `deleteApp` and maybe `deleteRuntime` as well?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/SamAuthProvider.scala
salonishah11: Yeah I can look into refactoring this. What is `subject` in terms of user info?

salonishah11:
jenkins retest

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
rtitle: Yeah sorry I backtracked a bit there. :)

Sounds good re `createRuntime`. Technically deletes don't request pets so I don't _think_ it's needed there. But it also doesn't hurt so I don't feel too strongly.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/SamAuthProvider.scala
rtitle: Subject is like the internal userid in Sam (e.g. primary key). Ex:
```
GET https://sam.dsde-dev.broadinstitute.org/register/user/v2/self/info
{
  "adminEnabled": true,
  "enabled": true,
  "userEmail": "ron.weasley@test.firecloud.org",
  "userSubjectId": "114763077412354570085"
}
```
Email is typically not a good identifier for a user since it can change.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
salonishah11: Oh I see. If I run delete app request through notebook, won't it pass the pet SA token instead of user token? 🤔

rtitle: deleteApp won't request a new pet from Sam though. So it might just work as-is with the pet token. However I haven't tested it so not sure.

salonishah11: Ah I see. I will test it out 👍

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/SamAuthProvider.scala
salonishah11: Ah so its basically user ID. Thank you!

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/SamAuthProvider.scala
```diff
@@ -207,6 +208,23 @@ class SamAuthProvider[F[_]: OpenTelemetryMetrics](
...
+        .flatMap(
+          _.fold(
+            F.raiseError[SamUserInfo](
+              AuthProviderException(
+                traceId,
+                s"[SamAuthProvider.lookupOriginatingUserEmail] Subject info not found for ${petOrUserInfo.userEmail}",
+                StatusCodes.Unauthorized
+              )
+            )
+          )(s => F.pure(s))
```
Qi77Qi: syntax wise, this might be a bit easier to read

```
      samUserInfoOpt <- samDao.getSamUserInfo(petOrUserInfo.accessToken.token)
      samUserInfo <- F.fromOption(
        samUserInfoOpt,
        AuthProviderException(
          traceId,
          s"[SamAuthProvider.lookupOriginatingUserEmail] Subject info not found for ${petOrUserInfo.userEmail}",
          StatusCodes.Unauthorized
        )
      )
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
Qi77Qi: yea, this doesn't feel right...in most cases, users are still calling the API as their original email

salonishah11:
Testing strategy:
I tested the changes in FIAB by opening a Notebook, fetching token from the Notebook (which would be from pet SA) and then using the token with curl calls to APIs in laptop terminal. Some snapshots:
`CURL_URL = "https://leonardo-fiab.dsde-dev.broadinstitute.org:30443/api/google/v1/apps/terra-test-81218270/terra-app-135cabf1-5203-43ee-9f68-c660a6c8905c"`
Create app:
```
curl -X 'POST' $CURL_URL \
 -H "Authorization: Bearer $TOKEN" \
 -d $REQUEST_BODY \
 -H "Content-Type:application/json"
The request has been accepted for processing, but the processing has not been completed.
```
The app did show up in Cloud Environments page:
Screen Shot 2022-05-03 at 1 52 34 PM
Get app details. The app's creator is the user and not pet SA:
```
curl -X 'GET' $CURL_URL \
 -H "Authorization: Bearer $TOKEN" \
 -H "Content-Type:application/json" | jq .
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   938    0   938    0     0   2726      0 --:--:-- --:--:-- --:--:--  2726
{
  ...
  "errors": [],
  "status": "PROVISIONING",
  "proxyUrls": {
    "cromwell-service": "https://leonardo-fiab.dsde-dev.broadinstitute.org:30443/proxy/google/v1/apps/terra-test-81218270/terra-app-135cabf1-5203-43ee-9f68-c660a6c8905c/cromwell-service/swagger/index.html?url=/proxy/google/v1/apps/terra-test-81218270/terra-app-135cabf1-5203-43ee-9f68-c660a6c8905c/cromwell-service/swagger/cromwell.yaml"
  },
  ....
  "auditInfo": {
    "creator": "groot.testerson@gmail.com",
    "createdDate": "2022-05-03T17:45:47.143382Z",
    "destroyedDate": null,
    "dateAccessed": "2022-05-03T17:45:47.143382Z"
  },
  "appType": "CROMWELL"
}
```
Delete app:
```
curl -X 'DELETE' $DELETE_URL \
  -H "Authorization: Bearer $TOKEN" \
  -H "Content-Type:application/json"
The request has been accepted for processing, but the processing has not been completed
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/SamAuthProvider.scala
```diff
@@ -207,6 +208,19 @@ class SamAuthProvider[F[_]: OpenTelemetryMetrics](
   )(implicit sr: SamResource[R], ev: Ask[F, TraceId]): F[Unit] =
     samDao.deleteResource(samResource, creatorEmail, googleProject)
 
+  override def lookupOriginatingUserEmail[R](petOrUserInfo: UserInfo)(implicit ev: Ask[F, TraceId]): F[WorkbenchEmail] =
+    for {
+      traceId <- ev.ask
+      samUserInfoOpt <- samDao.getSamUserInfo(petOrUserInfo.accessToken.token)
+      samUserInfo <- F.fromOption(
```
rtitle, line 215: Could this also check the `enabled` flag on the user for safety? Something like:
```
_ <- if (samUserInfo.enabled) F.unit else F.raiseError(AuthproviderException(…))
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/SamDAO.scala
```diff
@@ -72,9 +72,10 @@ trait SamDAO[F[_]] {
     implicit ev: Ask[F, TraceId]
   ): F[Option[UserSubjectId]]
 
-  def getUserSubjectIdFromToken(token: String)(implicit ev: Ask[F, TraceId]): F[Option[UserSubjectId]]
+  def getSamUserInfo(token: String)(implicit ev: Ask[F, TraceId]): F[Option[SamUserInfo]]
```
rtitle, line 75: Nit: the pattern in this class seems to use the `Authorization` type for the token instead of `String`, but I don't feel that strongly.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/SamDAO.scala
```diff
@@ -72,9 +72,10 @@ trait SamDAO[F[_]] {
...
 
   def getLeoAuthToken: F[Authorization]
 }
 
 final case class UserSubjectId(asString: String) extends AnyVal
+final case class UserEmail(asString: String) extends AnyVal
```
rtitle: Could probably just use `WorkbenchEmail` directly in `SamUserInfo`, and eliminate this type + decoders.

rtitle (approved):
A few nits above, overall looks good!

aednichols:
Bit off topic but is there any way to hide these Codecov annotations while reviewing?

( "Add tests until the code is covered" is not feasible to me as a reviewer 😄 )

Screen Shot 2022-05-04 at 1 06 45 PM

aednichols (approved):

PR #2648, opened 2022-04-29
BT-620 Add proxy path configuration

breilly2:
https://broadworkbench.atlassian.net/browse/BT-620

~In addition,~ I added service path to config to allow for different proxy endpoints for different services. A side effect of this is that the code for generating the proxy URL is back to being general for all apps, not different for Cromwell.

Update: While this no longer adds support for Job Manager (which we've pivoting away from), this code is still useful for adding a proxy for C-BAS (see https://broadworkbench.atlassian.net/browse/BW-1189 in epic https://broadworkbench.atlassian.net/browse/BW-1108)

---

In all cases:

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
@@ -367,7 +367,7 @@ final case class App(id: AppId,
     appResources.services.map { service =>
       val proxyPath = s"google/v1/apps/${project.value}/${appName.value}/${service.config.name.value}"
       val servicePath = service.config.path match {
-        case Some(path) => path.replace("{proxyPath}", proxyPath)
+        case Some(path) => path.value.replace("{proxyPath}", proxyPath)
         case None       => ""
       }
       (service.config.name, new URL(s"${proxyUrlBase}${proxyPath}${servicePath}"))
@@ -446,8 +446,9 @@ object AppStatus {
 
 final case class KubernetesService(id: ServiceId, config: ServiceConfig)
 final case class ServiceId(id: Long) extends AnyVal
-final case class ServiceConfig(name: ServiceName, kind: KubernetesServiceKindName, path: Option[String] = None)
+final case class ServiceConfig(name: ServiceName, kind: KubernetesServiceKindName, path: Option[ServicePath] = None)
 final case class KubernetesServiceKindName(value: String) extends AnyVal
+final case class ServicePath(value: String) extends AnyVal
 
 final case class KubernetesRuntimeConfig(numNodes: NumNodes, machineType: MachineTypeName, autoscalingEnabled: Boolean)
 final case class NumNodepools(value: Int) extends AnyVal
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
@@ -20,7 +20,6 @@ import org.broadinstitute.dsde.workbench.google2.{
   RegionName,
   SubnetworkName
 }
-import org.broadinstitute.dsde.workbench.leonardo.AppType.Cromwell
 import org.broadinstitute.dsde.workbench.model.{IP, TraceId, WorkbenchEmail}
 import org.broadinstitute.dsde.workbench.model.google.GoogleProject
 import org.broadinstitute.dsp.{ChartName, ChartVersion, Release}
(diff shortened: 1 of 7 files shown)


core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
```diff
@@ -364,17 +363,17 @@ final case class App(id: AppId,
                      customEnvironmentVariables: Map[String, String],
                      descriptorPath: Option[Uri],
                      extraArgs: List[String]) {
-  def getProxyUrls(project: GoogleProject, proxyUrlBase: String): Map[ServiceName, URL] =
+  def getProxyUrls(project: GoogleProject, proxyUrlBase: String): Map[ServiceName, URL] = {
```
breilly2: I can remove the surrounding `{ }` if you want. I added them during development to break the statement apart, but then inlined everything again at the end and forgot that I had added the `{ }`.

http/src/main/resources/reference.conf
```diff
@@ -478,6 +478,12 @@ gke {
       {
         name = "cromwell-service"
         kind = "ClusterIP"
+        path = "/swagger/index.html?url=/proxy/{proxyPath}/swagger/cromwell.yaml"
+      },
+      {
+        name = "job-manager"
+        kind = "ClusterIP"
+        path = "/jobs"
```
breilly2: This currently doesn't work because CromwellApp doesn't yet have a reachable JobManager. That work is in https://broadworkbench.atlassian.net/browse/BT-619 which I am starting work on now. Assume that BT-619 will merge before this.

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
Qi77Qi: I think scalafmt will auto remove these...can you run `sbt scalafmtAll`..CI will fail if it's not run

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/Config.scala
```diff
@@ -655,7 +655,8 @@ object Config {
   implicit private val serviceReader: ValueReader[ServiceConfig] = ValueReader.relative { config =>
     ServiceConfig(
       config.as[ServiceName]("name"),
-      config.as[KubernetesServiceKindName]("kind")
+      config.as[KubernetesServiceKindName]("kind"),
+      config.as[Option[String]]("path")
```
Qi77Qi: can you create a custom type for `ServicePath`

http/src/main/resources/reference.conf
Qi77Qi: how will the 2 URLs returned to user? looks like `getProxyUrls` is still returning one URL

Also in APIs, we're only returning one proxy URL for a given app

breilly2: `getProxyUrls` returns `Map[ServiceName, URL]`. Since each `App` can have multiple services and each service has its own URL, apps can have multiple proxy URLs.

The UI is using the `apps` API, specifically https://leonardo.dsde-dev.broadinstitute.org/#/apps/listAppByProject which returns a `proxyUrls` object that contains the service name to URL mapping returned by `getProxyUrls`.

Qi77Qi: ah I see...didn't realize we're returning a Map already for apps.

Qi77Qi (approved):

Qi77Qi:
PR still needs to be rebased...only up-to-date PRs are mergeable

http/src/main/resources/reference.conf
jgainerdewar: It sounds like this is the first time the `proxyUrls` map will contain more than one item. Should we be concerned about unexpected behavior from the UI, which will need to be prepared to select one of several URLs? I can imagine something on the frontend checking that the map contains a single item and throwing an error if not.

breilly2: There are 3 non-test reads of `proxyUrls` and they all ask for specific keys (`galaxy` or `cromwell-service`) so adding an additional entry should not break anything.

jgainerdewar (commented):
These changes look good to me, will thumb once rebased with passing tests.

breilly2:
I think it was up-to-date when I opened it. 🤷‍♂️  Anyway, I added an "on hold" label pending [BT-619](https://broadworkbench.atlassian.net/browse/BT-619). Once that ticket is done, I will rebase this and re-request Janet's review.

breilly2:
jenkins retest

Qi77Qi (approved):

salonishah11 (approved):
LGTM 👍

PR #2667, opened 2022-05-09
IA-3359 increase gce boot disk size and dataproc disk size

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-3359

---

In all cases:

### DIFF
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -59,7 +59,7 @@ dataproc {
   runtimeDefaults {
     numberOfWorkers = 0
     masterMachineType = "n1-standard-4"
-    masterDiskSize = 100
+    masterDiskSize = 120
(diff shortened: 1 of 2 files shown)


calbach (approved):
For future reference.. I'm not sure AoU needs to increase our min disk first.

For GCE: AoU increasing the default disk should have no impact, since we don't control the boot disk size.
For Dataproc: is masterDisk the same as the boot disk? If so, we don't necessarily need a Leo change to push an increase there. The only blocker would be if Leo is pushing changes that require a min disk increase.

Qi77Qi:
@calbach for dataproc, there's only one disk, so I guess you can say it's the boot disk size? when user sets disk size for dataproc, this is the size it's setting.

gpcarr (approved):

PR #2668, opened 2022-05-09
BW-1236: Remove no-longer-appropriate proxy path suffix

cjllanwarne:
I think this semi-duplicates work going on in #2648 but in a more targeted way, specifically only to un-break the proxy URL for Cromwell.


Qi77Qi (approved):

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
```diff
@@ -367,13 +367,7 @@ final case class App(id: AppId,
...
     appResources.services.map { service =>
       val proxyPath = s"google/v1/apps/${project.value}/${appName.value}/${service.config.name.value}"
-      appType match {
-        case Galaxy | Custom =>
-          (service.config.name, new URL(s"${proxyUrlBase}${proxyPath}"))
-        case Cromwell =>
-          (service.config.name,
-           new URL(s"${proxyUrlBase}${proxyPath}/swagger/index.html?url=/proxy/${proxyPath}/swagger/cromwell.yaml"))
-      }
+      (service.config.name, new URL(s"${proxyUrlBase}${proxyPath}"))
```
breilly2: Does this work without the trailing `/`? I think I remember seeing one last 301 to add the `/`, but maybe that one works.

breilly2 (approved):
👍  as long as it works for you without the trailing `/`.

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
cjllanwarne: Testing locally, the paths seem to work equivalently in my browser. However, for the benefit of my `workspace-cromwell` script which probably assumes a trailing `/`, I I will re-add it

PR #2671, opened 2022-05-10
IA-3164 Support B2C

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-3164

Depends on https://github.com/broadinstitute/firecloud-develop/pull/2912, https://github.com/broadinstitute/firecloud-develop/pull/2916

Similar to https://github.com/broadinstitute/sam/pull/720

Tested locally that using Hermione's google b2c token can see Hermione's runtimes

---

In all cases:


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/Boot.scala
```diff
@@ -154,7 +155,7 @@ object Boot extends IOApp {
       )
 
       val httpRoutes = new HttpRoutes(
-        swaggerConfig,
```
rtitle, line 157: Can we delete the `SwaggerConfig` class too?

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/ConfigReaderSpec.scala
```diff
@@ -44,6 +44,12 @@ class ConfigReaderSpec extends AnyFlatSpec with Matchers {
           ),
           Set.empty
         )
+      ),
+      OidcAuthConfig(
+        Uri.unsafeFromString("https://fake"),
```
rtitle, line 49: This might need to be a real URL like `"https://accounts.google.com"` if you're actually instantiating a` OpenIDConnectConfiguration`. Or you could use `FakeOpenIDConnectConfiguration`

rtitle: This is probably why unit tests are failing

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/CommonTestData.scala
```diff
@@ -150,6 +150,14 @@ object CommonTestData {
   val contentSecurityPolicy = Config.contentSecurityPolicy
   val refererConfig = Config.refererConfig
   val leoKubernetesConfig = Config.leoKubernetesConfig
+  lazy val openIdConnectionConfiguration = org.broadinstitute.dsde.workbench.oauth2
```
rtitle: Could use `FakeOpenIDConnectConfiguration` here

rtitle (approved):
Didn't test it, but the change LGTM

Qi77Qi:
support b2c token in proxy will be in a separate PR so that diff will be a bit more clear

PR #2689, opened 2022-05-13
IA-3164 Update proxyservice to handle b2c token

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-3164

Tested in fiab with b2c token

---

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/ProxyService.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/ProxyService.scala
@@ -5,7 +5,7 @@ package service
 import akka.actor.ActorSystem
 import akka.http.scaladsl.model.Uri.Host
 import akka.http.scaladsl.model._
-import akka.http.scaladsl.model.headers.`Content-Disposition`
+import akka.http.scaladsl.model.headers.{`Content-Disposition`, OAuth2BearerToken}
 import akka.http.scaladsl.model.ws._
 import akka.http.scaladsl.settings.ClientConnectionSettings
 import akka.http.scaladsl.unmarshalling.Unmarshal
@@ -15,12 +15,14 @@ import cats.effect.IO
 import cats.effect.std.Queue
 import cats.mtl.Ask
 import cats.syntax.all._
+import com.auth0.jwt.JWT
+import com.auth0.jwt.exceptions.JWTDecodeException
 import com.typesafe.scalalogging.LazyLogging
 import org.broadinstitute.dsde.workbench.google2.KubernetesSerializableName.ServiceName
 import org.broadinstitute.dsde.workbench.leonardo.SamResourceId._
 import org.broadinstitute.dsde.workbench.leonardo.config.ProxyConfig
 import org.broadinstitute.dsde.workbench.leonardo.dao.HostStatus._
-import org.broadinstitute.dsde.workbench.leonardo.dao.google.GoogleOAuth2Service
+import org.broadinstitute.dsde.workbench.leonardo.dao.google.{GoogleOAuth2Service, OAuth2Service}
 import org.broadinstitute.dsde.workbench.leonardo.dao.{HostStatus, JupyterDAO, Proxy, SamDAO, TerminalName}
 import org.broadinstitute.dsde.workbench.leonardo.db.{appQuery, clusterQuery, DbReference, KubernetesServiceDbQueries}
 import org.broadinstitute.dsde.workbench.leonardo.dns.{KubernetesDnsCache, ProxyResolver, RuntimeDnsCache}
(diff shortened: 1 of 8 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/ProxyService.scala
```diff
@@ -497,4 +510,22 @@ object ProxyService {
       }
       case _ => path
     }
+
+  def decodeB2cToken(token: String, now: Instant): Either[Throwable, (UserInfo, Instant)] =
```
rtitle, line 519: Nice

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeoException.scala
```diff
@@ -24,10 +24,12 @@ class LeoException(val message: String = null,
 final case class BadRequestException(msg: String, traceId: Option[TraceId])
     extends LeoException(msg, StatusCodes.BadRequest, traceId = traceId)
 
-final case class AuthenticationError(email: Option[WorkbenchEmail] = None)
-    extends LeoException(s"${email.map(e => s"'${e.value}'").getOrElse("Your account")} is not authenticated",
-                         StatusCodes.Unauthorized,
-                         traceId = None)
+final case class AuthenticationError(email: Option[WorkbenchEmail] = None, extraMessage: String = "")
```
rtitle, line 27: Is `extraMessage` used?

Qi77Qi: yea, I'm using it in https://github.com/DataBiosphere/leonardo/pull/2689/files#diff-4bbc2634fc8c5c714a0b57a58781db11b31911b90e711cfd6051f5572edb1c6eR133 to log a bit more info...but it's not returned to user

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/ProxyService.scala
```diff
@@ -105,28 +107,39 @@ class ProxyService(
...
-      _ <- IO.fromOption(samUserInfo.map(_.userSubjectId))(AuthenticationError(Some(userInfo.userEmail)))
-    } yield (userInfo, now.plusSeconds(userInfo.tokenExpiresIn.toInt))
+  private[leonardo] def getUserInfo(token: String,
+                                    now: Instant)(implicit ev: Ask[IO, TraceId]): IO[(UserInfo, Instant)] =
+    decodeB2cToken(token, now) match {
+      case Left(_: JWTDecodeException) =>
+        for {
+          userInfo <- googleOauth2Service.getUserInfoFromToken(token)
+          samUserInfo <- samDAO.getSamUserInfo(token)
+          _ <- IO.fromOption(samUserInfo.map(_.userSubjectId))(AuthenticationError(Some(userInfo.userEmail)))
```
rtitle: Is the point of calling Sam here to check user enablement? If so then maybe should check `samUserInfo.enabled` instead..

rtitle: Then again I don't think we really need to check user enablement explicitly in `ProxyService`, since the proxy calls Sam `/api` routes to check resource permissions, which does the user enablement check as part of it. 

So could we just remove the `samDAO.getSamUserInfo` call?

Qi77Qi: that sounds good to me..the sam check was just there...I assume it was for checking if a user is enabled

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/ProxyService.scala
```diff
@@ -105,28 +107,39 @@ class ProxyService(
...
+        } yield (userInfo, now.plusSeconds(userInfo.tokenExpiresIn.toInt))
+      case Left(e) =>
+        IO.raiseError(e)
+      case Right(value) =>
+        IO.pure(value)
```
rtitle, line 126: I love how the B2C case doesn't require a network hop to decode the token..

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeoException.scala
rtitle: I see 👍

rtitle (approved):
Looks good overall once tests pass.. as mentioned above I think we can remove the `Sam.getUserInfo` call (and noting we're not doing it anyway for the B2C case).

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/ProxyService.scala
Qi77Qi: So we can't really remove `samDAO.getSamUserInfo` from everywhere becuz `invalidateToken` and `setCookie` APIs also rely on this. I added a `checkUserEnabled` flag instead and we're skipping the Sam check only for proxy requests

PR #2701, opened 2022-05-20
IA-3432 IA-3440 Improve metrics and deletion bug fix

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-3416

* Previously we only have metrics for measuring from pubsub message that was published to when the message is enqueued into `AsyncTaskProcessor`. Now we're also adding metrics for measruring pubsub message published time to when the task completes in `AsyncTaskProcessor`
* Fix a bug in deletion where `deleteDisk` and `deleteNetworks` apis is called immediately after `deleteVM` wsm api, this can cause race condition where if `deleteVM` job doesn't finish by the time we try to delete disk and networks, the job will fail because disk and network would've still been associated the VM.
* When we delete a VM, previously we're deleting all the info in `RUNTIME_CONTROLLED_RESOURCE` table, but I think those are useful info to keep in case of error and it'll be useful for debugging purpose as well.

---

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/leoPubsubMessageSubscriberModels.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/leoPubsubMessageSubscriberModels.scala
@@ -898,6 +898,13 @@ object PubsubHandleMessageError {
       s"\n\truntimeId: ${runtimeId}, \n\tmsg: ${errorMsg})"
     val isRetryable: Boolean = false
   }
+
+  final case class AzureRuntimeDeletioinError(runtimeId: Long, workspaceId: WorkspaceId, errorMsg: String)
+      extends PubsubHandleMessageError {
+    override def getMessage: String =
+      s"\n\truntimeId: ${runtimeId}, \n\tmsg: ${errorMsg})"
+    val isRetryable: Boolean = false
+  }
 }
 
 final case class PersistentDiskMonitor(maxAttempts: Int, interval: FiniteDuration)
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/leoPubsubMessageSubscriberModels.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/leoPubsubMessageSubscriberModels.scala
@@ -899,7 +899,7 @@ object PubsubHandleMessageError {
     val isRetryable: Boolean = false
   }
 
-  final case class AzureRuntimeDeletioinError(runtimeId: Long, workspaceId: WorkspaceId, errorMsg: String)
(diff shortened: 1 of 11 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/leoPubsubMessageSubscriberModels.scala
```diff
@@ -898,6 +898,13 @@ object PubsubHandleMessageError {
       s"\n\truntimeId: ${runtimeId}, \n\tmsg: ${errorMsg})"
     val isRetryable: Boolean = false
   }
+
+  final case class AzureRuntimeDeletioinError(runtimeId: Long, workspaceId: WorkspaceId, errorMsg: String)
```
jdcanas: Typo in class name, `Deletion`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
```diff
@@ -326,106 +330,123 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
...
+              _ <- msg.diskId.traverse(diskId =>
+                dbRef.inTransaction(persistentDiskQuery.updateStatus(diskId, DiskStatus.Deleted, ctx.now))
+              )
+            } yield ()
+          case WsmJobStatus.Failed =>
+            F.raiseError[Unit](
+              AzureRuntimeDeletioinError(
+                msg.runtimeId,
+                msg.workspaceId,
+                s"WSM delete VM job failed due due to ${resp.errorReport.map(_.message).getOrElse("unknown")}"
```
jdcanas: `due` is in the string twice

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandlerSpec.scala
```diff
@@ -129,7 +129,7 @@ class AzurePubsubHandlerSpec
           controlledResources <- controlledResourceQuery.getAllForRuntime(runtime.id).transaction
         } yield {
           getRuntime.status shouldBe RuntimeStatus.Deleted
-          controlledResources.length shouldBe 0
+          (controlledResources.length > 0) shouldBe (true)
```
jdcanas: As I understand it the amount of controlled resources should be static since we don't delete the records anymore, we may want to check for the exact number instead of >0. Don't feel too strongly about it though.

jdcanas (approved):

jdcanas:
few small nits, looks good though ✅

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandlerSpec.scala
Qi77Qi: ah yea, I can make it equal to a specific number

PR #2723, opened 2022-06-02
IA-3463 Add group check for createApp API

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-3463

Security team wants to make it so that creating CUSTOM app is only allowed for a specific group

I tested locally with both a user in the group and a user isn't in the group with local leo against dev everything else.

Screen Shot 2022-06-02 at 3 29 22 PM

---

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/Config.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/Config.scala
@@ -327,7 +327,9 @@ object Config {
         config.getOrElse("notebookAuthCacheEnabled", true),
         config.getAs[Int]("notebookAuthCacheMaxSize").getOrElse(1000),
         config.getAs[FiniteDuration]("notebookAuthCacheExpiryTime").getOrElse(15 minutes),
-        config.getAs[GroupName]("SamAuthProviderConfig").getOrElse(GroupName("custom_app_user"))
+        config
+          .getAs[GroupName]("customAppCreationAllowedGroup")
+          .getOrElse(throw new Exception("No customAppCreationAllowedGroup key found"))
       )
   }
 
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/Config.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/Config.scala
@@ -328,8 +328,8 @@ object Config {
         config.getAs[Int]("notebookAuthCacheMaxSize").getOrElse(1000),
         config.getAs[FiniteDuration]("notebookAuthCacheExpiryTime").getOrElse(15 minutes),
         config
-          .getAs[GroupName]("customAppCreationAllowedGroup")
-          .getOrElse(throw new Exception("No customAppCreationAllowedGroup key found"))
+          .getOrElse[GroupName]("customAppCreationAllowedGroup",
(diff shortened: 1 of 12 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/Config.scala
```diff
@@ -325,7 +326,10 @@ object Config {
       SamAuthProviderConfig(
         config.getOrElse("notebookAuthCacheEnabled", true),
         config.getAs[Int]("notebookAuthCacheMaxSize").getOrElse(1000),
-        config.getAs[FiniteDuration]("notebookAuthCacheExpiryTime").getOrElse(15 minutes)
+        config.getAs[FiniteDuration]("notebookAuthCacheExpiryTime").getOrElse(15 minutes),
+        config
```
rtitle, line 330: Minor - could do `config.as[GroupName]("customAppCreationAllowedGroup")` to not get an Option?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpSamDAO.scala
```diff
@@ -383,6 +383,31 @@ class HttpSamDAO[F[_]](httpClient: Client[F],
...
+          method = Method.GET,
+          uri = config.samUri.withPath(Uri.Path.unsafeFromString(s"/api/groups/v1/${groupName.asString}/member")),
+          headers = Headers(leoAuth)
+        )
+      )(onError)
+      res <- if (members.contains(workbenchEmail))
+        F.pure(true)
+      else
+        for {
+          admins <- httpClient.expectOr[List[WorkbenchEmail]](
```
rtitle, line 402: Would be nice if we didn't have to make a Sam call for each policy but I don't see a way in the Sam API

Qi77Qi: I know...wish Sam api supports that

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeoAuthProvider.scala
```diff
@@ -201,4 +201,6 @@ trait LeoAuthProvider[F[_]] {
 
   // Get user info from Sam. If petOrUserInfo is a pet SA, Sam will return it's associated user account info
   def lookupOriginatingUserEmail[R](petOrUserInfo: UserInfo)(implicit ev: Ask[F, TraceId]): F[WorkbenchEmail]
+
+  def isAllowed(userEmail: WorkbenchEmail)(implicit ev: Ask[F, TraceId]): F[Boolean]
```
rtitle: A slightly more descriptive name would be nice? :) Maybe `isCustomAppAllowed`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/Config.scala
Qi77Qi: best I can see is `getOrElse`...guess I can do that

rtitle (approved):
Looks correct to me. Remember that the group needs to be created in each env (and Leo SA added)

PR #2754, opened 2022-06-24
IA-3522 fix bug that's preventing non cached images being created

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-3522

From support
```
I got an update from our Compute Engine team, they were able to reproduce this issue and the issue happens only when the startup script produces more than 64k bytes of output without a newline. 

The maximum length for a token in Go's bufio scanner is 64K. Is it possible to change the startup script to limit/redirect the output from docker pull command? 

Another possible workaround is using systemd units[1] instead of startup scripts. 

I hope this solution works for you. If you have further questions, I'd be glad to assist you. 
```

---

In all cases:


xuscrbw (approved):
Lgtm

PR #2757, opened 2022-06-24
PPWM-59 clone persistent disks

dvoet:
https://broadworkbench.atlassian.net/browse/PPWM-59

This adds:
- `sourceDisk: Option[SourceDiskRequest]` to requests to create persistent disks
- `sourceDisk: Option[DiskLink]` to `PersistentDisk`
- `sourceDisk: Option[DiskLink]` to `PersistentDiskComponent` and associated db column

related firecloud develop PR https://github.com/broadinstitute/firecloud-develop/pull/3029

---

In all cases:

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/runtimes/RuntimeCreationDiskSpec.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/runtimes/RuntimeCreationDiskSpec.scala
@@ -245,7 +245,7 @@ class RuntimeCreationDiskSpec
 
       val createDiskCloneRequest =
         CreateDiskRequest(Map.empty,
-                          None,
+                          Some(newDiskSize),
                           None,
                           None,
                           defaultCreateDiskRequest.zone,
@@ -299,7 +299,7 @@ class RuntimeCreationDiskSpec
         // Creating new runtime with existing disk should have test.txt file and user installed package
         runtimeWithData <- createRuntimeWithWait(googleProject, runtimeWithDataName, createRuntime2Request)
         _ <- verifyDisk(ClusterCopy.fromGetRuntimeResponseCopy(runtimeWithData))
-        _ <- deleteRuntimeWithWait(googleProject, runtimeWithDataName, deleteDisk = true)
+//        _ <- deleteRuntimeWithWait(googleProject, runtimeWithDataName, deleteDisk = true)
 
         // verify clone
         runtimeWithClone <- waitUntilRunning(googleProject, runtimeWithCloneName)
--- a/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/runtimes/RuntimeCreationDiskSpec.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/runtimes/RuntimeCreationDiskSpec.scala
@@ -10,7 +10,12 @@ import org.broadinstitute.dsde.workbench.google2.{streamFUntilDone, DiskName, Go
 import org.broadinstitute.dsde.workbench.leonardo.DiskModelGenerators._
 import org.broadinstitute.dsde.workbench.leonardo.LeonardoApiClient._
 import org.broadinstitute.dsde.workbench.leonardo.TestUser.{getAuthTokenAndAuthorization, Ron}
-import org.broadinstitute.dsde.workbench.leonardo.http.{PersistentDiskRequest, RuntimeConfigRequest, UpdateDiskRequest}
+import org.broadinstitute.dsde.workbench.leonardo.http.{
+  PersistentDiskRequest,
+  RuntimeConfigRequest,
+  SourceDiskRequest,
+  UpdateDiskRequest
+}
 import org.broadinstitute.dsde.workbench.leonardo.notebooks.{NotebookTestUtils, Python3}
 import org.http4s.client.Client
 import org.http4s.Status
@@ -23,6 +28,7 @@ import scala.concurrent.duration._
 @DoNotDiscover
 class RuntimeCreationDiskSpec
     extends BillingProjectFixtureSpec
+    with NewBillingProjectAndWorkspaceBeforeAndAfterAll
     with ParallelTestExecution
     with LeonardoTestUtils
     with NotebookTestUtils {
@@ -186,12 +192,14 @@ class RuntimeCreationDiskSpec
     res.unsafeRunSync()(cats.effect.unsafe.IORuntime.global)
   }
 
-  "create runtime and attach an existing persistent disk" taggedAs Retryable in { googleProject =>
+  "create runtime and attach an existing persistent disk or clone" taggedAs Retryable in { googleProject =>
     val randomeName = randomClusterName
     val runtimeName =
       randomeName.copy(asString = randomeName.asString + "pd-spec") // just to make sure the test runtime name is unique
     val runtimeWithDataName = randomeName.copy(asString = randomeName.asString + "pd-spec-data-persist")
+    val runtimeWithCloneName = randomeName.copy(asString = randomeName.asString + "pd-spec-clone")
     val diskName = genDiskName.sample.get
+    val diskCloneName = DiskName(diskName.value + "_clone")
     val diskSize = DiskSize(110)
     val newDiskSize = DiskSize(150)
 
@@ -217,6 +225,23 @@ class RuntimeCreationDiskSpec
       val createRuntime2Request = createRuntimeRequest.copy(toolDockerImage =
         Some(ContainerImage(LeonardoConfig.Leonardo.pythonImageUrl, ContainerRegistry.GCR))
       ) // this just needs to be a different image from default image Leonardo uses, which is gatk
+
+      val createRuntimeCloneRequest = createRuntime2Request.copy(
+        runtimeConfig = Some(
+          RuntimeConfigRequest.GceWithPdConfig(
+            None,
+            PersistentDiskRequest(
+              diskCloneName,
+              Some(DiskSize(500)),
+              None,
+              Map.empty,
+              Some(SourceDiskRequest(googleProject, diskName))
+            ),
+            defaultCreateDiskRequest.zone,
+            None
+          )
+        )
+      )
 
       for {
         _ <- LeonardoApiClient.createDiskWithWait(googleProject,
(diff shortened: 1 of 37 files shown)


http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20220626_add_disk_source.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="leonardo" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+
+    <changeSet logicalFilePath="leonardo" author="dvoet" id="add_source_disk">
+        <addColumn tableName="PERSISTENT_DISK">
+            <column name="sourceDisk" type="varchar(1024)">
```
dvoet, line 6: I added this to the database for consistency and so that round-tripping a `PersistentDisk` object to the database would not lose information. I don't think the feature requires this though. Comment if you think it should be removed.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/Boot.scala
```diff
@@ -128,21 +135,17 @@ object Boot extends IOApp {
...
         googleDependencies.googleComputeService,
-        appDependencies.publisherQueue
-      )
-      val diskService = new DiskServiceInterp[IO](
-        ConfigReader.appConfig.persistentDisk,
-        appDependencies.authProvider,
-        appDependencies.serviceAccountProvider,
-        appDependencies.publisherQueue
+        appDependencies.publisherQueue,
+        diskService
```
dvoet: I added `diskService` as a dependency for `RuntimeService` in order to reuse the code to lookup and verify access to a source disk. Let me know if there is a better way to do this with Leo.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskServiceInterp.scala
```diff
@@ -90,6 +101,22 @@ class DiskServiceInterp[F[_]: Parallel](config: PersistentDiskConfig,
...
 
+  override def lookupSourceDiskLink(userInfo: UserInfo, ctx: AppContext)(
+    sourceDiskReq: SourceDiskRequest
+  )(implicit as: Ask[F, AppContext]): F[DiskLink] =
+    for {
+      sourceDisk <- getDisk(userInfo, CloudContext.Gcp(sourceDiskReq.googleProject), sourceDiskReq.name).recoverWith {
+        case _: DiskNotFoundException =>
+          F.raiseError(BadRequestException("source disk does not exist", Option(ctx.traceId)))
+      }
+      maybeGoogleDisk <- googleDiskService.getDisk(sourceDiskReq.googleProject, sourceDisk.zone, sourceDisk.name)
```
dvoet, line 162: Note that this is actually verifying the disk exists in googleland. This is not strictly necessary. My main desire is to be able to call `googleDisk.getSelfLink` below instead of constructing the link via string concatenation.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala
```diff
@@ -358,7 +358,8 @@ class RuntimeV2ServiceInterp[F[_]: Parallel](config: RuntimeServiceConfig,
       config.defaultBlockSizeBytes,
       None,
       None,
-      labels
+      labels,
+      None
```
dvoet, line 362: Is there something else you want me to do here?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeServiceInterp.scala
```diff
@@ -1008,7 +1012,8 @@ object RuntimeServiceInterp {
                 diskConfig,
                 CreateDiskRequest.fromDiskConfigRequest(req, Some(targetZone)),
                 ctx.now,
-                if (willBeUsedBy == FormattedBy.Galaxy) true else false
+                if (willBeUsedBy == FormattedBy.Galaxy) true else false,
```
ehigham: oportunity to clean this up:
```suggestion
                willBeUsedBy == FormattedBy.Galaxy,
```

dvoet:
jenkins retest

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -2932,6 +2932,20 @@ components:
         blockSize:
           type: integer
           description: The block size of the persistent disk to be created in bytes.
+        sourceDisk:
+          $ref: "#/components/schemas/SourceDisk"
+
+    SourceDisk:
+      description: opitonal location of a disk to clone
```
Qi77Qi, line 2947: can we also add a comment saying is only applicable to GCP?

dvoet:
jenkins retest

dvoet:
jenkins retest

dvoet:
jenkins retest

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/runtimes/RuntimeCreationDiskSpec.scala
```diff
@@ -257,29 +290,19 @@ class RuntimeCreationDiskSpec
...
-              .executeCell(
-                "! df -h --output=size $HOME"
-              )
-              .get
-            res should include("148G")
-          }
-        })
         _ <- deleteRuntimeWithWait(googleProject, runtimeWithDataName, deleteDisk = true)
-        getDiskAttempt = getDisk(googleProject, diskName).attempt
+        _ <- deleteRuntimeWithWait(googleProject, runtimeWithCloneName, deleteDisk = true)
```
Qi77Qi, line 302: line 301 is already deleting runtime...

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -582,51 +582,55 @@ class LeoPubsubMessageSubscriber[F[_]](
...
+    val create = {
+      val diskBuilder = Disk
+        .newBuilder()
+        .setName(msg.name.value)
+        .setSizeGb(msg.size.gb)
+        .setZone(msg.zone.value)
+        .setType(msg.diskType.googleString(msg.googleProject, msg.zone))
+        .setPhysicalBlockSizeBytes(msg.blockSize.bytes)
+        .putAllLabels(Map("leonardo" -> "true").asJava)
+      msg.sourceDisk.foreach(sourceDisk => diskBuilder.setSourceDisk(sourceDisk.asString))
```
Qi77Qi, line 594: functionally it's the same here..but can we use `traverse` instead of `foreach`

Also, what does the string format of `DiskLink` look like? it'd be nice to add at `DiskLink`'s definition site

http/src/test/resources/reference.conf
```diff
@@ -329,3 +329,5 @@ azure {
 }
 
 opencensus-scala.trace.sampling-probability = 1.0
+
+persistent-disk.dont-clone-from-these-google-folders = ["bogus"]
```
Qi77Qi, line 333: if we add the real ones in the main `reference.conf`, then we don't need this

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala
Qi77Qi: this is fine since we're not supporting creating a runtime from a source disk directly

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskServiceInterp.scala
```diff
@@ -90,6 +105,68 @@ class DiskServiceInterp[F[_]: Parallel](config: PersistentDiskConfig,
...
+    for {
+      ctx <- as.ask
+      sourceAncestry <- F.fromFuture(F.delay(googleProjectDAO.getAncestry(sourceGoogleProject.value)))
+      sourceAncestor <- immediateAncestor(sourceAncestry)
+
+      targetAncestry <- F.fromFuture(F.delay(googleProjectDAO.getAncestry(targetGoogleProject.value)))
+      targetAncestor <- immediateAncestor(targetAncestry)
+
+      _ <- F.raiseWhen(
+        sourceAncestor != targetAncestor &&
```
Qi77Qi, line 131: should it be `||`?

Qi77Qi (commented):
lgtm in general...but tests are still failing..

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/runtimes/RuntimeCreationDiskSpec.scala
dvoet: this is deleting a different runtime

dvoet:
> lgtm in general...but tests are still failing..

I will rerun. All Leo test runs were failing up until about an hour ago

http/src/test/resources/reference.conf
dvoet: the real ones are there... this line makes the tests easier

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
dvoet: If I use `traverse` instead I need to move it down into the for comp and do
```
        _ <- msg.sourceDisk.traverse(sourceDisk => F.pure(diskBuilder.setSourceDisk(sourceDisk.asString))) 
```
`foreach` is more straight forward to me but whatever you want.

I don't really know what the format of `DiskLink` looks like. It comes from google and we send it to google. 🙈

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskServiceInterp.scala
dvoet: no, we raise the error when the source is different from the target AND the source is on the don't clone list

dvoet: in other words
- clone is allowed if the source and target are the same (cloning within the same folder is fine)
- clone is allowed if the source is not on the list regardless of the target

Qi77Qi: so if source and target are the same, and source is from a dontCloneFromTheseGoogleFolders, we will still allow the copy?

dvoet: yes

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskServiceInterp.scala
```diff
@@ -90,6 +105,68 @@ class DiskServiceInterp[F[_]: Parallel](config: PersistentDiskConfig,
...
+      _ <- F.raiseWhen(ancestry.size < 2)(
+        LeoInternalServerError(s"expected at least 2 ancestors but got ${ancestry.mkString(",")}", Option(ctx.traceId))
+      )
+    } yield ancestry.tail.head
+
+  private def lookupSourceDisk(userInfo: UserInfo, ctx: AppContext)(
+    sourceDiskReq: SourceDiskRequest
+  )(implicit as: Ask[F, AppContext]): F[SourceDisk] =
+    for {
+      sourceDisk <- getDisk(userInfo, CloudContext.Gcp(sourceDiskReq.googleProject), sourceDiskReq.name).recoverWith {
```
Qi77Qi, line 158: `getDisk` function does a sam call to check if the caller has right permission. I think this is already done at call site for `lookupSourceDisk`...

you can probably just use `DiskServiceDbQueries.getGetPersistentDiskResponse(cloudContext, diskName, ctx.traceId).transaction` instead

dvoet: The code is correct as written. This is doing a sam call on the _source_ disk. The call site does a sam call for the disk being created.

Qi77Qi (approved):

PR #2768, opened 2022-07-07
[IA-3497] readiness and liveliness implemented

jdcanas:
Note that there is a firecloud develop PR that exists as this endpoint is exposed on a different port

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/Boot.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/Boot.scala
@@ -28,7 +28,18 @@ import org.broadinstitute.dsde.workbench.google.{
 }
 import org.broadinstitute.dsde.workbench.google2.GKEModels.KubernetesClusterId
 import org.broadinstitute.dsde.workbench.google2.util.RetryPredicates
-import org.broadinstitute.dsde.workbench.google2.{GoogleComputeService, Event, GoogleSubscriber, GooglePublisher, GoogleDataprocService, GKEService, GoogleDiskService, GoogleResourceService, GoogleStorageService, credentialResource}
+import org.broadinstitute.dsde.workbench.google2.{
+  credentialResource,
+  Event,
+  GKEService,
+  GoogleComputeService,
+  GoogleDataprocService,
+  GoogleDiskService,
+  GooglePublisher,
+  GoogleResourceService,
+  GoogleStorageService,
+  GoogleSubscriber
+}
 import org.broadinstitute.dsde.workbench.leonardo.AsyncTaskProcessor.Task
 import org.broadinstitute.dsde.workbench.leonardo.auth.{AuthCacheKey, PetClusterServiceAccountProvider, SamAuthProvider}
 import org.broadinstitute.dsde.workbench.leonardo.config.Config._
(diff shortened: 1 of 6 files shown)


jdcanas:
jenkins retest

n8landolt (approved):
It all looks good to me!

I'd love an explanation of the changes from StatusRoutes to package.scala so I can get a better idea of what exactly those lines of code do.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/Boot.scala
```diff
@@ -85,6 +92,33 @@ object Boot extends IOApp {
...
+              logger
+                .error(t)("FATAL - failure starting liveliness http server")
+                .unsafeToFuture()(cats.effect.unsafe.IORuntime.global)
+            }
+        }
+      }
+      _ <- logger.info("Liveliness server has been created, starting...")
+    } yield serverBinding
+
+    val livelinessApp = Stream.emits(List(Stream.eval[IO, Unit](livelinessServer.void))).covary[IO].parJoin(1)
```
Qi77Qi: you don't need `parJoin` here...we should just add this stream to the main one

jdcanas: hmm... how would I start it here then add it to the main one? The fs2.stream lib has really bad documentation :(

Qi77Qi: so you can remove `livelinessApp` and add `livelinessServer` after line 280...sth like

```
 List(
          appDependencies.leoPublisher.process, // start the publisher queue .dequeue
          Stream.eval[IO, Unit](httpServer), // start http server
          Stream.eval[IO, Unit](livelinessServer)
        )
```

jdcanas: the whole point is that it starts as soon as possible though, before the setup of all the dependences though from what I understood?

Qi77Qi: oh, you're right...forgot about that..

in that case, you can just do this
```
    val livelinessRoutes = new LivelinessRoutes

    logger.info("Liveliness server has been created, starting...").unsafeToFuture() >> Http()
      .newServerAt("0.0.0.0", 9000)
      .bindFlow(livelinessRoutes.route)
      .onError { case t: Throwable =>
        logger
          .error(t)("FATAL - failure starting liveliness http server")
          .unsafeToFuture()(cats.effect.unsafe.IORuntime.global)
      }
```

jdcanas: ok, I modified the code a bit. I figured we wanted to block on the Future here, so i added an `Await` to the snippet you posted here.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/Boot.scala
```diff
@@ -91,6 +82,21 @@ object Boot extends IOApp {
...
+    val liveliness = logger.info("Liveliness server has been created, starting...").unsafeToFuture()(cats.effect.unsafe.IORuntime.global) >> Http()
+      .newServerAt("0.0.0.0", 9000)
+      .bindFlow(livelinessRoutes.route)
+      .onError { case t: Throwable =>
+        logger
+          .error(t)("FATAL - failure starting liveliness http server")
+          .unsafeToFuture()(cats.effect.unsafe.IORuntime.global)
+      }
+
+    Await.ready(liveliness, Duration.Inf)
```
Qi77Qi: I don't think we need ot await here...we can just let it start asynchrounously

jdcanas: I guess it doesn't really matter, but the thought is it would be awkward if readiness was up before liveliness due to async thread scheduling nonsense...

We can remove it though and see

Qi77Qi: that shouldn't happen...the `liveness` app is really simple, the start up should almost be instant...if liveness takes that long to start up, I think we will have a problem regardless

Qi77Qi (approved):

PR #2779, opened 2022-07-18
[PROD-698] disable galaxy tests

gpcarr:
https://broadworkbench.atlassian.net/browse/PROD-698

We are disabling app creation specs until galaxy upgrades the version of GKE that it uses

image (3)


Qi77Qi (approved):
could've used `ignore`...but this works too :)

PR #2798, opened 2022-07-27
IA-3509 Refactor app table

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-3509

During the mobbing, we went down the path updating some GKE models in wb-libs towards the end, as I rethink about this, I think it's better to hold off on that until we have a better understanding of what these models for Azure will look like (I'm also slightly inclined to keep things separate at this point. But knowing what Azure needs will help us make better decisions).

---

In all cases:


xuscrbw (approved):
LGTM

PR #2804, opened 2022-08-04
IA-3512 Run welder on Azure VM

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-3512

* This sets up welder on Azure VM. The updated welder version should also be compatible with GCP, so it should be safe to merge
* Fix a bug where in [RuntimeV2ServiceInterp.scala](https://github.com/DataBiosphere/leonardo/pull/2804/files#diff-b6c8ddc1824e19ffe2fc9a0ab8b0e372d1cfc4944dc67c23b6f1a5411da904d0), we're using leo's token for WSM calls. But we should use user's access token
* Create a leo staging storage container during VM creation
---

In all cases:


core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/azureModels.scala
```diff
@@ -3,9 +3,12 @@ package org.broadinstitute.dsde.workbench.leonardo
 import java.util.UUID
 
 final case class WsmControlledResourceId(value: UUID) extends AnyVal
+final case class StorageContainerName(value: String) extends AnyVal
```
jdcanas: maybe use `ContainerName` from the wb-libs AzureStorage here to cut down on duplicate models

Qi77Qi: you mean `org.broadinstitute.dsde.workbench.google2.KubernetesSerializableName.ContainerName`? that one doesn't seem right...this is an Azure storage container

http/src/main/resources/reference.conf
```diff
@@ -148,8 +148,15 @@ gce {
...
 azure {
+   wsm {
+     uri = "https://localhost:8000"
+   }
+
    pubsub-handler {
     sam-url = ${sam.server}
+    wsm-url = ${azure.wsm.uri}
+    welder-acr-uri = "terradevacrpublic.azurecr.io/welder-server"
+    welder-image-hash = ${image.welderHash}
```
jdcanas, line 159: hmm.. don't we have this in a specific place already?

I remember our CI that publishes welder hash changes to `reference.conf` had trouble last time we added another welder hash to the config. Could be worth looking into compatibility so we don't have to manually update this too...

Qi77Qi: yea, it's referecing that exact config you're thinking... which is why value here is ${image.welderHash}

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
```diff
@@ -465,14 +465,29 @@ object clusterQuery extends TableQuery(new ClusterTable(_)) {
...
       // staging bucket is saved as a bucket name rather than a path
-      .map(recs => recs.headOption.flatten.flatMap(head => parseGcsPath("gs://" + head + "/").toOption))
+      .map(recs =>
+        recs.headOption.flatMap { head =>
+          head._1 match {
+            case CloudProvider.Gcp => head._2.map(s => StagingBucket.Gcp(GcsBucketName(s)))
+            case CloudProvider.Azure =>
+              head._2.map { s =>
+                val res = for {
+                  splitted <- Either.catchNonFatal(s.split("/"))
```
jdcanas, line 482: nit: `splitted` sounds weird (its tense doesn't match the usual words on the left hand side of for comprehensions) can we have it say `split`

Qi77Qi: how about `splittedString` just to be more specific? I think it's fairly common to call it this way

jdcanas (approved):
Looks good to me pending manual test confirmation....

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
jdcanas: meh, not super important. just don't like the `ed` which makes it a verb in the past tense, and the other vars are nouns

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/azureModels.scala
jdcanas: This one https://github.com/broadinstitute/workbench-libs/blob/develop/azure/src/main/scala/org/broadinstitute/dsde/workbench/azure/AzureStorageService.scala#L66

Qi77Qi: ah cool..I'll pull in the latest wb-libs

Qi77Qi:
jenkins retest

Qi77Qi:
jenkins retest

Qi77Qi:
```
19:14:49 ERROR: for martha-app  Cannot create container for service martha-app: mkdir /var/lib/docker/overlay2/d76d6394b94c4277ff753997825f1e8ae39b0a277ab8747bcda11416162e6393-init: no space left on device
```

jenkins retest

PR #2813, opened 2022-08-10
PROD-718 IA-3630 Fix wrong googleProject encoding

Qi77Qi:

---

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/AppRoutes.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/AppRoutes.scala
@@ -247,7 +247,7 @@ object AppRoutes {
       "labels"
     )(x =>
       (x.cloudContext,
-       x.cloudContext,
+       x.cloudContext.asString,
        x.kubernetesRuntimeConfig,
        x.errors,
        x.status,
--- a/core/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/AppRoutesTestJsonCodec.scala
+++ b/core/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/AppRoutesTestJsonCodec.scala
@@ -47,7 +47,7 @@ object AppRoutesTestJsonCodec {
     )(GetAppResponse.apply)
 
   implicit val listAppResponseDecoder: Decoder[ListAppResponse] =
-    Decoder.forProduct10("googleProject",
+    Decoder.forProduct10("cloudContext",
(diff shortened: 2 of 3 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/AppRoutes.scala
```diff
@@ -247,7 +247,7 @@ object AppRoutes {
       "labels"
     )(x =>
       (x.cloudContext,
-       x.cloudContext,
```
gpcarr, line 250: cloudContext is an object, i don't think this will work

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/AppRoutes.scala
```diff
@@ -247,7 +247,7 @@ object AppRoutes {
...
+       x.cloudContext.asString,
```
n8landolt, line 250: Will this return the property "cloudResource" from cloudContext?

We're expecting the property "cloudResource" here, at least for galaxy apps.

Qi77Qi, line 250: it's a class that has `asString`....when it's CloudContext.Gcp, this is equivelant of the previous value

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/AppRoutes.scala
Qi77Qi: it's a class that has `asString`....when it's CloudContext.Gcp, this is equivelant of the previous value we had for googleProject

Qi77Qi: see similar usage https://github.com/DataBiosphere/leonardo/blob/develop/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/runtimeRoutesModels.scala#L182

n8landolt (approved):

gpcarr (approved):

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/AppRoutes.scala
Qi77Qi: @n8landolt yep...the `cloudContext` field will have what both `cloudResource` and `cloudProvider`

PR #2829, opened 2022-08-26
[IA-3555] Use application allowlist + project label when creating applications.

n8landolt:
https://broadworkbench.atlassian.net/browse/IA-3555

Firecloud develop PR: https://github.com/broadinstitute/firecloud-develop/pull/3086

Note - in the user story, it mentioned "Remove user allowlist check"

---

In all cases:


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
```diff
@@ -68,6 +78,22 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
 
       _ <- checkIfUserAllowed(req.appType, userInfo.userEmail)
 
+      projectLabels <- googleResourceService.getLabels(googleProject)
+      _ <-
+        if (req.appType == Custom && customAppSecurityConfig.enableCustomAppCheck) {
```
jdcanas: we usually prefer to use the `match` keyword to simplify complex branches. (Similar to `Utils.cond`/`Utils.switchcase` in the UI)

I believe, simplified, it would look something like this

```
appAllowList =
            if (projectLabels.getOrElse("security-group", "").canEqual("high"))
              customAppSecurityConfig.customApplicationAllowList.highSecurity
            else customAppSecurityConfig.customApplicationAllowList.default
_ <- req.appType match {
   case Custom if customAppSecurityConfig.enableCustomAppCheck and  !appAllowList.contains(req.descriptorPath.getOrElse("None").toString) => F.raiseError(...)
   case _ => F.unit
}
```

I think you can simplify this even further though, something like

```
appAllowList =
            if (projectLabels.getOrElse("security-group", "").canEqual("high"))
              customAppSecurityConfig.customApplicationAllowList.highSecurity
            else customAppSecurityConfig.customApplicationAllowList.default
_ <- F.raiseError[Unit](AppCreationException(s"App is not in app allow list."))
        .whenA(!appAllowList.contains(req.descriptorPath.getOrElse("None").toString) && req.appType == Custom && customAppSecurityConfig.enableCustomAppCheck)
```

You can find other uses of `whenA` in the code, check `RuntimeServiceInterp` for the equivalent of what you are doing

jdcanas (dismissed):
Overall, semantically, the business logic looks good and the tests are thorough. Seems like you adapted well to our style of having mocks where you can override specific methods, instead of using something like mockito mocks/verifies.

I don't like the syntax in the AppServiceInterp, though I'd say that seems to be unfamiliarity with scala.

http/src/main/resources/reference.conf
```diff
@@ -906,3 +906,11 @@ app-service {
   # Defaults to true, but disabled for fiab runs
   enable-custom-app-group-permission-check = true
 }
+
+customAppSecurityConfig {
+    enableCustomAppCheck = true
+    customApplicationAllowList = {
+      default: [],
+      highSecurity: []
```
Qi77Qi: this list should go to the `firecloud-develop` repo

Qi77Qi: also, can we move the config to https://github.com/DataBiosphere/leonardo/blob/91d54dc6d3abb12693ee29b6166a7c4ad9e03061/http/src/main/resources/reference.conf#L507

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
```diff
@@ -529,8 +553,7 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
       for {
         ctx <- ev.ask
         isUserAllowed <- appType match {
```
n8landolt: Should I remove the isUserAllowed check entirely?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
```diff
@@ -68,6 +78,20 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
 
       _ <- checkIfUserAllowed(req.appType, userInfo.userEmail)
 
+      projectLabels <- googleResourceService.getLabels(googleProject)
+      appAllowList =
+        if (projectLabels.getOrElse("security-group", "").canEqual("high"))
```
Qi77Qi: can we make the string `security-group` and `high` a constant at the top of this class..

Also can we move the code into `checkIfUserAllowed` instead.

The check here isn't quite right. So if there's no project label for `security-group`, we want to keep using the previous check; only when there is `security-group` label, we'll enforce this check.

Another thing is before we check both, we should check `enableCustomAppCheck` flag first

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
```diff
@@ -529,8 +553,7 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
       for {
         ctx <- ev.ask
         isUserAllowed <- appType match {
-          case AppType.Custom => authProvider.isCustomAppAllowed(userEmail)
-          case _              => F.pure(true)
+          case _ => F.pure(true)
```
Qi77Qi: we still want to keep the current check...

See previous comment

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
n8landolt: > can we make the string `security-group` and `high` a constant at the top of this class..

Done!

> Also can we move the code into `checkIfUserAllowed` instead.

I saw that method and I was wondering about it. This update isn't really checking if the _user_ is allowed, so I'm not sure that specific method is where we want to implement it. Thoughts?

> The check here isn't quite right. So if there's no project label for `security-group`, we want to keep using the previous check; only when there is `security-group` label, we'll enforce this check.


What is this previous check? Is it the checkIfUserAllowed?

> Another thing is before we check both, we should check `enableCustomAppCheck` flag first


Do you mind giving example syntax for this? As a scala newbie, this would be helpful.

http/src/main/resources/reference.conf
```diff
@@ -906,3 +910,4 @@ app-service {
   # Defaults to true, but disabled for fiab runs
   enable-custom-app-group-permission-check = true
```
Qi77Qi: think this needs to be renamed?

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterpSpec.scala
```diff
@@ -108,34 +152,19 @@ final class AppServiceInterpSpec extends AnyFlatSpec with LeonardoTestSuite with
     }
   }
 
-  it should "fail request if user is not in custom_app_users group" in {
```
Qi77Qi: we should add back the test..

n8landolt: *facepalm* I must have ctrl+x not ctrl+c

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
```diff
@@ -66,7 +78,24 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
...
 
-      _ <- checkIfUserAllowed(req.appType, userInfo.userEmail)
+      _ <- req.appType match {
+        case AppType.Custom =>
+          req.descriptorPath match {
+            case Some(descriptorPath) =>
+              checkIfAppCreationIsAllowed(userInfo.userEmail, googleProject, descriptorPath)
+            case None =>
+              F.raiseError(
+                new LeoException(
```
Qi77Qi: use `BadRequestException` instead?

Qi77Qi (commented):
lgtm! and tests are passing...

have you tested things out locally to make sure the logic is working as expected?

Qi77Qi (approved):

n8landolt:
jenkins retest

n8landolt:
jenkins retest

PR #2860, opened 2022-09-15
[IA-3427] Add start and stop functionality to Leo Backend.

n8landolt:
https://broadworkbench.atlassian.net/browse/IA-3427

---

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
@@ -11,6 +11,7 @@ import com.azure.resourcemanager.compute.models.{VirtualMachine, VirtualMachineS
 import org.broadinstitute.dsde.workbench.azure.{
   AzureCloudContext,
   AzureRelayService,
+  AzureVmServiceInterp,
   ContainerName,
   RelayHybridConnectionName
 }
@@ -20,25 +21,25 @@ import org.broadinstitute.dsde.workbench.leonardo.SamResourceId.WsmResourceSamRe
 import org.broadinstitute.dsde.workbench.leonardo.config.ContentSecurityPolicyConfig
 import org.broadinstitute.dsde.workbench.leonardo.dao._
 import org.broadinstitute.dsde.workbench.leonardo.db._
-import org.broadinstitute.dsde.workbench.leonardo.http.dbioToIO
+import org.broadinstitute.dsde.workbench.leonardo.http.{cloudServiceOps, ctxConversion, dbioToIO}
 import org.broadinstitute.dsde.workbench.leonardo.monitor.LeoPubsubMessage.{
   CreateAzureRuntimeMessage,
   DeleteAzureRuntimeMessage
 }
-import org.broadinstitute.dsde.workbench.leonardo.monitor.PubsubHandleMessageError
+import org.broadinstitute.dsde.workbench.leonardo.monitor.{LeoPubsubMessage, PubsubHandleMessageError}
 import org.broadinstitute.dsde.workbench.leonardo.monitor.PubsubHandleMessageError.{
   AzureRuntimeCreationError,
   AzureRuntimeDeletionError,
   ClusterError
 }
 import org.broadinstitute.dsde.workbench.model.{IP, WorkbenchEmail}
+import org.broadinstitute.dsde.workbench.util2.InstanceName
 import org.http4s.headers.Authorization
 import org.typelevel.log4cats.StructuredLogger
 
 import java.time.Instant
 import java.util.UUID
 import scala.concurrent.ExecutionContext
-import org.broadinstitute.dsde.workbench.leonardo.http.ctxConversion
 
 class AzurePubsubHandlerInterp[F[_]: Parallel](
   config: AzurePubsubHandlerConfig,
@@ -47,7 +48,8 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
   wsmDao: WsmDao[F],
   samDAO: SamDAO[F],
   jupyterDAO: JupyterDAO[F],
-  azureRelay: AzureRelayService[F]
+  azureRelay: AzureRelayService[F],
+  azureVmServiceInterp: AzureVmServiceInterp[F]
 )(implicit val executionContext: ExecutionContext, dbRef: DbReference[F], logger: StructuredLogger[F], F: Async[F])
     extends AzurePubsubHandlerAlgebra[F] {
   implicit val wsmDeleteVmDoneCheckable: DoneCheckable[Option[GetDeleteJobResult]] = (v: Option[GetDeleteJobResult]) =>
@@ -167,6 +169,28 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
       }
       _ <- wsmDao.createVm(createVmRequest, auth)
     } yield ()
+
+  override def startAndPollRuntime(msg: LeoPubsubMessage.StartRuntimeMessage, runtime: Runtime)(implicit
+    ev: Ask[F, AppContext]
+  ): F[Unit] = for {
+    ctx <- ev.ask
+    runtimeOpt <- clusterQuery.getClusterById(msg.runtimeId).transaction
+    runtime <- F.fromOption(runtimeOpt, PubsubHandleMessageError.ClusterNotFound(msg.runtimeId, msg))
+    runtimeConfig <- RuntimeConfigQueries.getRuntimeConfig(runtime.runtimeConfigId).transaction
+
+    op <- azureVmServiceInterp.startAzureVm(InstanceName(runtime.runtimeName.asString), runtime.cloudContext)
+
+    // TODO: Polling?
+    //    _ <- monitorCreateRuntime(
+    //      PollRuntimeParams(msg.workspaceId, runtime, createVmJobId, msg.relayNamespace)
+    //    )
+  } yield ()
+
+  override def stopAndPollRuntime(msg: LeoPubsubMessage.StopRuntimeMessage)(implicit
+    ev: Ask[F, AppContext]
+  ): F[Unit] = for {
+    ctx <- ev.ask
+  } yield ()
 
   private def createStorageContainer(params: CreateAzureRuntimeParams, auth: Authorization)(implicit
     ev: Ask[F, AppContext]
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
@@ -22,7 +22,6 @@ import org.broadinstitute.dsde.workbench.leonardo.config.ContentSecurityPolicyCo
 import org.broadinstitute.dsde.workbench.leonardo.dao._
 import org.broadinstitute.dsde.workbench.leonardo.db._
 import org.broadinstitute.dsde.workbench.leonardo.http.{ctxConversion, dbioToIO}
-import org.broadinstitute.dsde.workbench.leonardo.model.RuntimeNotFoundException
 import org.broadinstitute.dsde.workbench.leonardo.monitor.LeoPubsubMessage.{
   CreateAzureRuntimeMessage,
   DeleteAzureRuntimeMessage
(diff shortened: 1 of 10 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
```diff
@@ -168,6 +173,103 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
...
+      case None =>
+        logger.error(ctx.loggingCtx)(
+          s"Cannot find runtime ${runtime.projectNameString}."
+        ) // Error? For now, log a message
+      case Some(mono) =>
+        val task = for {
+          _ <- F.blocking(mono.block()) // TODO: Add timeout
+          _ <- clusterQuery.updateClusterStatus(runtime.id, RuntimeStatus.Stopped, ctx.now).transaction
+          // TODO: Do we NONE the IP address like GCP?
+          // _ <- clusterQuery.updateClusterHostIp(params.runtime.id, None, ctx.now).transaction
```
n8landolt: GCP logic performs this query. Should we do this for Azure as well?
If yes, we'll need to update the Azure start logic.

Qi77Qi: we don't need to...the reason why we need to do this for GCE is becuz everytime a VM starts/stops, the host IP will change....but for Azure, we're accessing it via a relay URL, which won't change during start/stop cycle

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
```diff
@@ -168,6 +176,104 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
...
+    ctx <- ev.ask
+    monoOpt <- azureVmServiceInterp.startAzureVm(InstanceName(runtime.runtimeName.asString), azureCloudContext)
+
+    _ <- monoOpt match {
+      case None =>
+        logger.error(ctx.loggingCtx)(
+          s"Cannot find runtime ${runtime.projectNameString}."
+        ) // Error? For now, log a message
+        F.raiseError(
+          RuntimeNotFoundException(
```
Qi77Qi: we should raise `AzureRuntimeStartingError` so that it'll be handled properly in the handler. And you can get rid of the log statement above

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
```diff
@@ -168,6 +176,104 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
...
+            s"Jupyter was not running within ${config.createVmPollConfig.maxAttempts} attempts with ${config.createVmPollConfig.interval} delay"
+          )
+          _ <- clusterQuery.updateClusterStatus(runtime.id, RuntimeStatus.Running, ctx.now).transaction
+          _ <- logger.info(ctx.loggingCtx)("runtime is ready")
+        } yield ()
+        asyncTasks.offer(
+          Task(
+            ctx.traceId,
+            task,
+            // Set status to error? Probably not, will get user stuck.
```
Qi77Qi: remove comments

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
```diff
@@ -168,6 +176,104 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
...
+    ev: Ask[F, AppContext]
+  ): F[Unit] = for {
+    ctx <- ev.ask
+    monoOpt <- azureVmServiceInterp.stopAzureVm(InstanceName(runtime.runtimeName.asString), azureCloudContext)
+    _ <- monoOpt match {
+      case None =>
+        logger.error(ctx.loggingCtx)(
+          s"Cannot find runtime ${runtime.projectNameString}."
+        ) // Error? For now, log a message
+        F.raiseError(
```
Qi77Qi, line 234: same thing..raise the properly error and remove the log statement above

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriberSpec.scala
```diff
@@ -1781,6 +1781,60 @@ class LeoPubsubMessageSubscriberSpec
...
+        makeCluster(1)
+          .copy(status = RuntimeStatus.Starting, cloudContext = CloudContext.Azure(azureCloudContext))
+          .saveWithRuntimeConfig(azureRuntimeConfig)
+      )
+      tr <- traceId.ask[TraceId]
+      _ <- leoSubscriber.messageResponder(StartRuntimeMessage(runtime.id, Some(tr)))
+      updatedRuntime <- clusterQuery.getClusterById(runtime.id).transaction
+    } yield {
+      updatedRuntime shouldBe defined
+      updatedRuntime.get.status shouldBe RuntimeStatus.Starting
```
Qi77Qi: I'm not sure if this validating anything...Runtime's status was initially `Starting` to begin with

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriberSpec.scala
```diff
@@ -1781,6 +1781,60 @@ class LeoPubsubMessageSubscriberSpec
...
+      runtime <- IO(
+        makeCluster(1)
+          .copy(status = RuntimeStatus.Stopping, cloudContext = CloudContext.Azure(azureCloudContext))
+          .saveWithRuntimeConfig(azureRuntimeConfig)
+      )
+      tr <- traceId.ask[TraceId]
+      _ <- leoSubscriber.messageResponder(StopRuntimeMessage(runtime.id, Some(tr)))
+      updatedRuntime <- clusterQuery.getClusterById(runtime.id).transaction
+    } yield {
+      updatedRuntime shouldBe defined
```
Qi77Qi: same comment

Qi77Qi: I think most of the logic lives in `AzurePubsubHandler`, so I think the unit tests `AzurePubsubHandlerSpec` is mostly enough...unless you have some specific logic in `leoPubsubMessageSubscriber` that you'd like to test

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandlerSpec.scala
```diff
@@ -375,19 +379,184 @@ class AzurePubsubHandlerSpec
...
+      _ <- azureInterp.stopAndMonitorRuntime(runtime, azureCloudContext)
+
+      _ <- withInfiniteStream(asyncTaskProcessor.process, assertions)
+
+    } yield ()
+
+    res.unsafeRunSync()(cats.effect.unsafe.IORuntime.global)
+  }
+
+  it should "fail on stopAzureVm - no runtime found" in isolatedDbTest {
```
Qi77Qi: it's more valuable to test that we're handling a `AzureRuntimeStopError` properly than to verify an arbitrary exception is thrown

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriberSpec.scala
n8landolt: I mostly wanted the coverage on the branching logic, and the test looks similar to the existing tests.

Also, the test just verifies that it's still in "starting" status, so in a way it is verifying that it hasn't been changed.

I do think that it's not "testing much" but I still think we should keep them

Qi77Qi: if you wanna test out branching logic, use mock of the AzurePubsubHandler and confirm the mock is being called will be much better test than just validating a DB record never changed (which to me is pretty meaningless)

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriberSpec.scala
```diff
@@ -1828,56 +1831,48 @@ class LeoPubsubMessageSubscriberSpec
...
-
       runtime <- IO(
         makeCluster(1)
           .copy(status = RuntimeStatus.Starting, cloudContext = CloudContext.Azure(azureCloudContext))
           .saveWithRuntimeConfig(azureRuntimeConfig)
       )
       tr <- traceId.ask[TraceId]
-      _ <- leoSubscriber.messageResponder(StartRuntimeMessage(runtime.id, Some(tr)))
+      _ <- leoSubscriber.messageResponder(StartRuntimeMessage(runtime.id, Some(tr))).attempt
       updatedRuntime <- clusterQuery.getClusterById(runtime.id).transaction
```
Qi77Qi: think we don't need `updatedRuntime <- clusterQuery.getClusterById(runtime.id).transaction` anymore? also what is `attempt` for?

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriberSpec.scala
```diff
@@ -1828,56 +1831,48 @@ class LeoPubsubMessageSubscriberSpec
...
-
       runtime <- IO(
         makeCluster(1)
           .copy(status = RuntimeStatus.Stopping, cloudContext = CloudContext.Azure(azureCloudContext))
           .saveWithRuntimeConfig(azureRuntimeConfig)
       )
       tr <- traceId.ask[TraceId]
-      _ <- leoSubscriber.messageResponder(StopRuntimeMessage(runtime.id, Some(tr)))
+      _ <- leoSubscriber.messageResponder(StopRuntimeMessage(runtime.id, Some(tr))).attempt
       updatedRuntime <- clusterQuery.getClusterById(runtime.id).transaction
```
Qi77Qi: same

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriberSpec.scala
n8landolt: So I'm not entirely sure why, but when trying to verify startAndMonitorRuntime using the original runtime as the first argument, the values are different.

The values that change are proxyUrl and scopes. The rest is the same.

n8landolt: For the attempt portion, without using attempt, I get an NPE on "fa"

 should handle Azure StartRuntimeMessage and start runtime *** FAILED ***
[info]   java.lang.NullPointerException: Cannot invoke "cats.effect.IO.handleErrorWith(scala.Function1)" because "fa" is null

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandlerSpec.scala
```diff
@@ -375,19 +382,184 @@ class AzurePubsubHandlerSpec
...
+        .copy(
+          cloudContext = CloudContext.Azure(azureCloudContext)
+        )
+        .saveWithRuntimeConfig(azureRuntimeConfig)
+
+      _ <- azureInterp.startAndMonitorRuntime(runtime, azureCloudContext)
+
+    } yield ()
+    a[AzureRuntimeStartingError] should be thrownBy {
+      res.unsafeRunSync()(cats.effect.unsafe.IORuntime.global)
```
Qi77Qi: this will never be thrown...

`res` is an `IO`, so you need to run it with `res.unsafeRunSync()(cats.effect.unsafe.IORuntime.global)`


And this is where I'd like the `handleError` function being tested instead of validating an exception is thrown

n8landolt: When I run it without the `a[AzureRuntimeStartingError] should be thrownBy` wrapper I get a test failed.

Qi77Qi (approved):
once tests pass 🎉

Qi77Qi:
fiab-start failure

jenkins retest

Qi77Qi:
fiab-start failed..jenkins retest

n8landolt:
jenkins retest

project/Dependencies.scala
```diff
@@ -15,7 +15,7 @@ object Dependencies {
   val monocleV = "2.1.0"
   val opencensusV = "0.29.0"
 
-  private val workbenchLibsHash = "982e088b-SNAP"
```
Qi77Qi: use what's in main branch for now..that has the fix

n8landolt: Will do!

PR #2884, opened 2022-10-03
[IA-3704] Remove startup script vulnerabilities.

n8landolt:
https://broadworkbench.atlassian.net/browse/IA-3704

There's two main parts to this PR.

The first is adjustments to the startup script to prevent user input to be run from the command line when starting a VM.

The second is standard scheme URL validation. The google storage scheme isn't available to be validated by UrlSchemeValidator, and so must be caught by the startup scripts.

# Testing

Ensured current functionality stayed the same by passing a valid user script, both through a google storage link (gs://) and through https://

Ensured a (potentially) malicious user input would not allow the user to create an environment for both gs:// and https:// See:

Screen Shot 2022-10-06 at 3 58 02 PM

---

In all cases:

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/runtimeModels.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/runtimeModels.scala
@@ -1,6 +1,7 @@
 package org.broadinstitute.dsde.workbench.leonardo
 
-import java.net.{MalformedURLException, URL}
+import cats.implicits._
+import java.net.URL
 import java.time.Instant
 import enumeratum.{Enum, EnumEntry}
 import monocle.Prism
@@ -12,7 +13,8 @@ import org.broadinstitute.dsde.workbench.model.google.{parseGcsPath, GcsBucketNa
 import org.broadinstitute.dsde.workbench.model.{IP, TraceId, ValueObject, WorkbenchEmail}
 import java.nio.file.Path
 import scala.collection.immutable
-import org.apache.commons.validator.routines.UrlValidator
+
+import org.http4s.Uri
 
 /**
  * This file contains models for Leonardo runtimes.
@@ -308,16 +310,16 @@ object UserScriptPath {
     case UserScriptPath.Http(_) => None
   }(identity)
 
-  def stringToUserScriptPath(string: String): Either[Throwable, UserScriptPath] =
+  def stringToUserScriptPath(string: String): Either[Throwable, UserScriptPath] = {
     parseGcsPath(string) match {
       case Right(value) => Right(Gcs(value))
       case Left(_) =>
-        if (UrlValidator.getInstance().isValid(string)) {
-          Right(Http(new URL(string)))
-        } else {
-          Left(new MalformedURLException("Startup script has invalid content."))
-        }
+        for {
+          _ <- Uri.fromString(string)
+          res <- Either.catchNonFatal(new URL(string))
+        } yield UserScriptPath.Http(res)
     }
+  }
 }
 
 /** Jupyter extension configuration */
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/runtimeModels.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/runtimeModels.scala
@@ -2,7 +2,6 @@ package org.broadinstitute.dsde.workbench.leonardo
 
 import java.net.{MalformedURLException, URL}
 import java.time.Instant
-import cats.syntax.all._
 import enumeratum.{Enum, EnumEntry}
 import monocle.Prism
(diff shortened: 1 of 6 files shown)


http/src/main/resources/init-resources/gce-init.sh
```diff
@@ -161,19 +161,19 @@ function apply_user_script() {
 
   log "Running user script $USER_SCRIPT_URI in $CONTAINER_NAME container..."
   USER_SCRIPT=`basename ${USER_SCRIPT_URI}`
-  if [[ "$USER_SCRIPT_URI" == 'gs://'* ]]; then
-    $GSUTIL_CMD cp ${USER_SCRIPT_URI} /var &> /var/user_script_copy_output.txt
+  if [[ `${USER_SCRIPT_URI}` == 'gs://'* ]]; then
```
Qi77Qi: are these quotes necessary?

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/runtimeModels.scala
```diff
@@ -311,7 +312,12 @@ object UserScriptPath {
   def stringToUserScriptPath(string: String): Either[Throwable, UserScriptPath] =
     parseGcsPath(string) match {
       case Right(value) => Right(Gcs(value))
-      case Left(_)      => Either.catchNonFatal(new URL(string)).map(url => Http(url))
+      case Left(_) =>
+        if (UrlValidator.getInstance().isValid(string)) {
```
Qi77Qi: you can use `org.http4s.Uri` instead since the project already has it and we don't have to bring in additional dependency

```
@  org.http4s.Uri.fromString("http://127.0.0.1") 
res2: org.http4s.package.ParseResult[org.http4s.Uri] = Right(
  value = Uri(scheme = Some(value = Scheme(http)), authority = Some(value = Authority(userInfo = None, host = Ipv4Address(address = 127.0.0.1), port = None)), path = , query = , fragment = None)
)

@ org.http4s.Uri.fromString("userscript.sh | nslookup http://another-url.com") 
res3: org.http4s.package.ParseResult[org.http4s.Uri] = Left(value = ParseFailure(sanitized = "Invalid URI", details = "Error(13,NonEmptyList(EndOfString(13,47)))"))
```

http/src/main/resources/init-resources/gce-init.sh
n8landolt: It's part of the suggested fixes in the document, here:

https://docs.google.com/document/d/1u0Bz2LG8DB-prJ2FxmSDqBayHR-ON0IuGQHYhgt-YZY/edit#heading=h.cukwnduvnuig

Let me know if I misunderstood the suggestions.

n8landolt: [Lack of proper quotation/sanitization of the URL in gce-init.sh script](https://github.com/DataBiosphere/leonardo/blob/6fe7502ffe064b939b25897624931b0d7203d7b1/http/src/main/resources/init-resources/gce-init.sh#L132): Leonardo code renders user URL in double-quotes ([””](https://storage.googleapis.com/dsp-appsec-test123op214h21oh-public/hello.sh$(docker)), which makes it possible to run arbitrary code in [$(](https://storage.googleapis.com/dsp-appsec-test123op214h21oh-public/hello.sh$(docker)...) when the URL is assigned. In this case, the URL should be rendered in single-quotes ([’ ’](https://storage.googleapis.com/dsp-appsec-test123op214h21oh-public/hello.sh$(docker)) to avoid shell interpolation. Other places that may be affected (although we haven’t been able to find an exploit) and should be quoted with double-quotes (which would prevent multiple parameter expansion) are where USER_SCRIPT_URI and START_USER_SCRIPT_URI are used without quotes, e.g. [here](https://github.com/DataBiosphere/leonardo/blob/6fe7502ffe064b939b25897624931b0d7203d7b1/http/src/main/resources/init-resources/gce-init.sh#L163-L167).

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeonardoModelSpec.scala
```diff
@@ -148,14 +151,22 @@ class LeonardoModelSpec extends LeonardoTestSuite with AnyFlatSpecLike {
 
   it should "create UserScriptPath objects according to provided path" in {
     val gcsPath = "gs://userscript_bucket/userscript.sh"
-    val httpPath = "https://userscript_path"
+    val httpPath = "https://w3.org"
```
Qi77Qi: just curious why the change :)

n8landolt: The other validator didn't like the previous input, I think in part because it's not truly a valid URL

Qi77Qi (approved):
🎉

PR #2901, opened 2022-10-13
[IA-3765] Add V2 workspace based app routes

n8landolt:
https://broadworkbench.atlassian.net/browse/IA-3765
---

In all cases:

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/appRoutesModels.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/appRoutesModels.scala
@@ -10,10 +10,12 @@ import org.broadinstitute.dsde.workbench.leonardo.{
   AppType,
   AuditInfo,
   CloudContext,
+  CloudProvider,
   KubernetesCluster,
   KubernetesRuntimeConfig,
   LabelMap,
-  Nodepool
+  Nodepool,
+  WorkspaceId
 }
 import org.broadinstitute.dsde.workbench.model.UserInfo
 import org.http4s.Uri
@@ -56,6 +58,19 @@ final case class ListAppResponse(cloudContext: CloudContext,
                                  labels: LabelMap
 )
 
+final case class ListAppV2Response(cloudProvider: Option[CloudProvider],
+                                   workspaceId: Option[WorkspaceId],
+                                   kubernetesRuntimeConfig: KubernetesRuntimeConfig,
+                                   errors: List[AppError],
+                                   status: AppStatus, // TODO: do we need some sort of aggregate status?
+                                   proxyUrls: Map[ServiceName, URL],
+                                   appName: AppName,
+                                   appType: AppType,
+                                   diskName: Option[DiskName],
+                                   auditInfo: AuditInfo,
+                                   labels: LabelMap
+)
+
 final case class GetAppResult(cluster: KubernetesCluster, nodepool: Nodepool, app: App)
 
 object ListAppResponse {
@@ -64,6 +79,35 @@ object ListAppResponse {
       n.apps.map { a =>
         ListAppResponse(
           c.cloudContext,
+          KubernetesRuntimeConfig(
+            n.numNodes,
+            n.machineType,
+            n.autoscalingEnabled
+          ),
+          a.errors,
+          a.status,
+          a.getProxyUrls(c.cloudContext.asInstanceOf[CloudContext.Gcp].value,
+                         proxyUrlBase
+          ), // TODO: refactor once we support proxying azure app
+          a.appName,
+          a.appType,
+          a.appResources.disk.map(_.name),
+          a.auditInfo,
+          a.labels.filter(l => labelsToReturn.contains(l._1))
+        )
+      }
+    )
+
+}
+
+object ListAppV2Response {
+  def fromCluster(c: KubernetesCluster, proxyUrlBase: String, labelsToReturn: List[String]): List[ListAppV2Response] =
+    c.nodepools.flatMap(n =>
+      n.apps.map { a =>
+        ListAppV2Response(
+          // TODO: WorkspaceId
+          None,
+          None,
           KubernetesRuntimeConfig(
             n.numNodes,
             n.machineType,
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/appRoutesModels.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/appRoutesModels.scala
@@ -81,11 +81,7 @@ object ListAppResponse {
           ),
           a.errors,
           a.status,
-          a.getProxyUrls(c.cloudContext,
-                         c.workspaceId,
-                         proxyUrlBase,
-                         apiVersion
-          ), // TODO: refactor once we support proxying azure app
+          a.getProxyUrls(c.cloudContext, c.workspaceId, proxyUrlBase, apiVersion),
           a.appName,
           a.appType,
           a.appResources.disk.map(_.name),
@@ -112,7 +108,7 @@ object GetAppResponse {
                                  appResult.cluster.workspaceId,
                                  proxyUrlBase,
                                  apiVersion
-      ), // TODO: refactor once we support proxying azure app
+      ),
       appResult.app.appResources.disk.map(_.name),
       appResult.app.customEnvironmentVariables,
       appResult.app.auditInfo,
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/appRoutesModels.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/appRoutesModels.scala
@@ -63,7 +63,11 @@ final case class ListAppResponse(cloudProvider: CloudProvider,
 final case class GetAppResult(cluster: KubernetesCluster, nodepool: Nodepool, app: App)
 
 object ListAppResponse {
-  def fromCluster(c: KubernetesCluster, proxyUrlBase: String, labelsToReturn: List[String]): List[ListAppResponse] =
+  def fromCluster(c: KubernetesCluster,
+                  proxyUrlBase: String,
+                  labelsToReturn: List[String],
+                  apiVersion: String
+  ): List[ListAppResponse] =
     c.nodepools.flatMap(n =>
       n.apps.map { a =>
         ListAppResponse(
@@ -77,8 +81,10 @@ object ListAppResponse {
           ),
           a.errors,
           a.status,
-          a.getProxyUrls(c.cloudContext.asInstanceOf[CloudContext.Gcp].value,
-                         proxyUrlBase
+          a.getProxyUrls(c.cloudContext,
+                         c.workspaceId,
+                         proxyUrlBase,
+                         apiVersion
           ), // TODO: refactor once we support proxying azure app
           a.appName,
           a.appType,
@@ -88,11 +94,10 @@ object ListAppResponse {
         )
       }
     )
-
 }
 
 object GetAppResponse {
-  def fromDbResult(appResult: GetAppResult, proxyUrlBase: String): GetAppResponse =
+  def fromDbResult(appResult: GetAppResult, proxyUrlBase: String, apiVersion: String): GetAppResponse =
     GetAppResponse(
       appResult.app.appName,
       appResult.cluster.cloudContext,
@@ -103,8 +108,10 @@ object GetAppResponse {
       ),
       appResult.app.errors,
       appResult.app.status,
-      appResult.app.getProxyUrls(appResult.cluster.cloudContext.asInstanceOf[CloudContext.Gcp].value,
-                                 proxyUrlBase
+      appResult.app.getProxyUrls(appResult.cluster.cloudContext,
+                                 appResult.cluster.workspaceId,
+                                 proxyUrlBase,
+                                 apiVersion
       ), // TODO: refactor once we support proxying azure app
       appResult.app.appResources.disk.map(_.name),
       appResult.app.customEnvironmentVariables,
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/appRoutesModels.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/appRoutesModels.scala
@@ -46,7 +46,9 @@ final case class GetAppResponse(appName: AppName,
                                 labels: LabelMap
 )
 
-final case class ListAppResponse(cloudContext: CloudContext,
+final case class ListAppResponse(cloudProvider: CloudProvider,
+                                 workspaceId: Option[WorkspaceId],
+                                 cloudContext: CloudContext,
                                  kubernetesRuntimeConfig: KubernetesRuntimeConfig,
                                  errors: List[AppError],
                                  status: AppStatus, // TODO: do we need some sort of aggregate status?
@@ -58,19 +60,6 @@ final case class ListAppResponse(cloudContext: CloudContext,
                                  labels: LabelMap
 )
 
-final case class ListAppV2Response(cloudProvider: Option[CloudProvider],
-                                   workspaceId: Option[WorkspaceId],
-                                   kubernetesRuntimeConfig: KubernetesRuntimeConfig,
-                                   errors: List[AppError],
-                                   status: AppStatus, // TODO: do we need some sort of aggregate status?
-                                   proxyUrls: Map[ServiceName, URL],
-                                   appName: AppName,
-                                   appType: AppType,
-                                   diskName: Option[DiskName],
-                                   auditInfo: AuditInfo,
-                                   labels: LabelMap
-)
-
 final case class GetAppResult(cluster: KubernetesCluster, nodepool: Nodepool, app: App)
 
 object ListAppResponse {
@@ -78,36 +67,9 @@ object ListAppResponse {
     c.nodepools.flatMap(n =>
       n.apps.map { a =>
         ListAppResponse(
+          c.cloudContext.cloudProvider,
+          c.workspaceId,
           c.cloudContext,
-          KubernetesRuntimeConfig(
-            n.numNodes,
-            n.machineType,
-            n.autoscalingEnabled
-          ),
-          a.errors,
-          a.status,
-          a.getProxyUrls(c.cloudContext.asInstanceOf[CloudContext.Gcp].value,
-                         proxyUrlBase
-          ), // TODO: refactor once we support proxying azure app
-          a.appName,
-          a.appType,
-          a.appResources.disk.map(_.name),
-          a.auditInfo,
-          a.labels.filter(l => labelsToReturn.contains(l._1))
-        )
-      }
-    )
-
-}
-
-object ListAppV2Response {
-  def fromCluster(c: KubernetesCluster, proxyUrlBase: String, labelsToReturn: List[String]): List[ListAppV2Response] =
-    c.nodepools.flatMap(n =>
-      n.apps.map { a =>
-        ListAppV2Response(
-          // TODO: WorkspaceId
-          None,
-          None,
           KubernetesRuntimeConfig(
             n.numNodes,
             n.machineType,
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
@@ -234,18 +234,21 @@ final case class NumNodes(amount: Int) extends AnyVal
 final case class AutoscalingMin(amount: Int) extends AnyVal
 final case class AutoscalingMax(amount: Int) extends AnyVal
 
-final case class DefaultKubernetesLabels(googleProject: GoogleProject,
+final case class DefaultKubernetesLabels(cloudContext: CloudContext,
                                          appName: AppName,
                                          creator: WorkbenchEmail,
                                          serviceAccount: WorkbenchEmail
 ) {
+  val cloudContextList = cloudContext match {
+    case CloudContext.Gcp(value)   => List("googleProject" -> value.value)
+    case CloudContext.Azure(value) => List("cloudContext" -> value.asString)
+  }
   val toMap: LabelMap =
     Map(
       "appName" -> appName.value,
-      "googleProject" -> googleProject.value,
       "creator" -> creator.value,
       "clusterServiceAccount" -> serviceAccount.value
-    )
+    ) ++ cloudContextList
 }
 
 sealed abstract class ErrorAction
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
@@ -39,7 +39,8 @@ case class KubernetesCluster(id: KubernetesClusterLeoId,
                              auditInfo: AuditInfo,
                              asyncFields: Option[KubernetesClusterAsyncFields],
                              namespaces: List[Namespace],
-                             nodepools: List[Nodepool]
+                             nodepools: List[Nodepool],
+                             workspaceId: Option[WorkspaceId]
 ) {
 
   // TODO consider renaming this method and the KubernetesClusterId class
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
@@ -379,15 +379,50 @@ final case class App(id: AppId,
                      descriptorPath: Option[Uri],
                      extraArgs: List[String]
 ) {
-  def getProxyUrls(project: GoogleProject, proxyUrlBase: String): Map[ServiceName, URL] =
+
+  def getProxyUrls(cloudContext: CloudContext,
+                   workspaceId: Option[WorkspaceId],
+                   proxyUrlBase: String,
+                   apiVersion: String
+  ): Map[ServiceName, URL] =
     appResources.services.map { service =>
-      val proxyPath = s"google/v1/apps/${project.value}/${appName.value}/${service.config.name.value}"
-      val servicePath = service.config.path match {
-        case Some(path) => path.value.replace("{proxyPath}", proxyPath)
-        case None       => ""
+      apiVersion match {
+        case "v1" =>
+          cloudContext match {
+            case CloudContext.Gcp(googleProject) => getProxyUrlsV1(googleProject, proxyUrlBase, service)
+            case CloudContext.Azure(_)           => ??? // TODO: Error
+          }
+        case "v2" =>
+          workspaceId match {
+            case Some(workspace) => getProxyUrlsV2(workspace, proxyUrlBase, service)
+            case None            => ??? // TODO: Error
+          }
       }
-      (service.config.name, new URL(s"${proxyUrlBase}${proxyPath}${servicePath}"))
     }.toMap
+
+  def getProxyUrlsV1(project: GoogleProject,
+                     proxyUrlBase: String,
+                     kubernetesService: KubernetesService
+  ): (ServiceName, URL) = {
+    val proxyPath = s"google/v1/apps/${project.value}/${appName.value}/${kubernetesService.config.name.value}"
+    val servicePath = kubernetesService.config.path match {
+      case Some(path) => path.value.replace("{proxyPath}", proxyPath)
+      case None       => ""
+    }
+    (kubernetesService.config.name, new URL(s"${proxyUrlBase}${proxyPath}${servicePath}"))
+  }
+
+  def getProxyUrlsV2(workspaceId: WorkspaceId,
+                     proxyUrlBase: String,
+                     kubernetesService: KubernetesService
+  ): (ServiceName, URL) = {
+    val proxyPath = s"apps/v2/${workspaceId.value}/${appName.value}/${kubernetesService.config.name.value}"
+    val servicePath = kubernetesService.config.path match {
+      case Some(path) => path.value.replace("{proxyPath}", proxyPath)
+      case None       => ""
+    }
+    (kubernetesService.config.name, new URL(s"${proxyUrlBase}${proxyPath}${servicePath}"))
+  }
 }
 
 sealed abstract class AppStatus
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
@@ -21,7 +21,6 @@ import org.broadinstitute.dsde.workbench.google2.{
   SubnetworkName
 }
 import org.broadinstitute.dsde.workbench.model.{IP, TraceId, WorkbenchEmail}
-import org.broadinstitute.dsde.workbench.model.google.GoogleProject
 import org.broadinstitute.dsp.{ChartName, ChartVersion, Release}
 import org.http4s.Uri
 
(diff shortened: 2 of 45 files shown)


http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -78,6 +80,60 @@ paths:
                 $ref: "#/components/schemas/LeonardoVersion"
       security: []
 
+  ## Appv2 Paths ##
```
Qi77Qi: add doc for all apis

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -121,6 +121,8 @@ class LeoPubsubMessageSubscriber[F[_]](
               e.getMessage
             )
           }
+        case _: CreateAppV2Message => F.unit
```
Qi77Qi: add TODO here

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/leoPubsubMessageSubscriberModels.scala
```diff
@@ -260,6 +260,22 @@ object LeoPubsubMessage {
...
 
+  final case class CreateAppV2Message(
+    workspaceId: WorkspaceId,
+    appId: AppId,
+    appName: AppName,
+    createDisk: Option[DiskId],
+    customEnvironmentVariables: Map[String, String],
+    appType: AppType,
+    namespaceName: NamespaceName,
+    machineType: Option[
```
Qi77Qi: is the comment accurate? think we can remove `machineType` for now

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
```diff
@@ -505,12 +513,217 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
...
+  ): F[Vector[ListAppResponse]] = for {
+    paramMap <- F.fromEither(processListParameters(params))
+    allClusters <- KubernetesServiceDbQueries
+      .listAppsByWorkspaceId(Some(workspaceId), paramMap._1, paramMap._2)
+      .transaction
+
+    res = allClusters
+      .flatMap(c => ListAppResponse.fromCluster(c, Config.proxyConfig.proxyUrlBase, paramMap._3))
+      .toVector
+    // TODO: Figure out permissions
```
Qi77Qi: this permission part should be mostly the same as v1 implementation. might make sense to have a helper function to share the code

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -78,6 +80,60 @@ paths:
...
+
+  "/api/apps/v2/{workspaceId}":
```
rtitle: Should we add the other routes (getApp, createApp, deleteApp) to swagger as well?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
```diff
@@ -505,12 +513,217 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
...
+      ctx <- as.ask
+      userToken = org.http4s.headers.Authorization(
+        org.http4s.Credentials.Token(AuthScheme.Bearer, userInfo.accessToken.token)
+      )
+      workspaceDescOpt <- wsmDao.getWorkspace(workspaceId, userToken)
+      workspaceDesc <- F.fromOption(workspaceDescOpt, WorkspaceNotFoundException(workspaceId, ctx.traceId))
+
+      hasPermission <- authProvider.hasPermission[WorkspaceResourceSamResourceId, WorkspaceAction](
+        WorkspaceResourceSamResourceId(workspaceId),
+        WorkspaceAction.CreateControlledUserResource,
```
rtitle, line 527: `WorkspaceAction.CreateControlledUserResource` seems like the right thing to check here 👍 (even though we're not actually making a controlled resource ;))

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
```diff
@@ -505,12 +513,217 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
...
+
+      _ <- F.raiseUnless(hasPermission)(ForbiddenError(userInfo.userEmail))
+
+      appOpt <- KubernetesServiceDbQueries.getActiveFullAppByWorkspaceIdAndAppName(workspaceId, appName).transaction
+      _ <- appOpt.fold(F.unit)(c =>
+        F.raiseError[Unit](AppAlreadyExistsInWorkspaceException(workspaceId, appName, c.app.status, ctx.traceId))
+      )
+
+      samResourceId <- F.delay(AppSamResourceId(UUID.randomUUID().toString))
+      // TODO Verify managed identiy is covered by lookupOriginatingUserEmail
```
rtitle: It should, I think.. the commented code here looks mostly correct (we just don't want to create apps with a `googleProject` parent resource in this case I think).

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
```diff
@@ -505,12 +513,217 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
...
+        if (saveClusterResult.minimalCluster.status == KubernetesClusterStatus.Error)
+          F.raiseError[Unit](
+            KubernetesAppCreationException(
+              s"You cannot create an app while a cluster${saveClusterResult.minimalCluster.clusterName.value} is in status ${saveClusterResult.minimalCluster.status}",
+              Some(ctx.traceId)
+            )
+          )
+        else F.unit
+
+      runtimeServiceAccountOpt = none[WorkbenchEmail] // TODO: Figure out how to get PET managed identity
```
rtitle: Yup we'll need to get a pet managed identity for the azure case.. it's a separate Sam call: https://sam.dsde-dev.broadinstitute.org/#/Azure

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
```diff
@@ -505,12 +513,217 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
...
+//        serviceAccountProvider.getClusterServiceAccount(userInfo, googleProject)
+      _ <- ctx.span.traverse(s => F.delay(s.addAnnotation("Done Sam call for getClusterServiceAccount")))
+      petSA <- F.fromEither(
+        runtimeServiceAccountOpt.toRight(new Exception(s"user ${userInfo.userEmail.value} doesn't have a PET SA"))
+      )
+      // TODO: Adapt the following to support Azure disk
```
rtitle: I think short term we can just not support diskConfig for Azure apps. (We should continue to support it for GCP though.)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
```diff
@@ -505,12 +513,217 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
...
+//          userInfo,
+//          petSA,
+//          appTypeToFormattedByType(req.appType),
+//          authProvider,
+//          config.leoKubernetesConfig.diskConfig
+//        )
+//      )
+
+      saveApp <- F.fromEither(
+        getSavableApp(
```
rtitle, line 623: We should confirm `getSavableApp` populates the right helm chart on both clouds. On GCP we should continue to support our existing charts. For Azure we should only support the `Cromwell` appType and use the chart defined here: https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/resources/reference.conf#L209

rtitle: Thinking out loud: should we have a new `AppType` enum for Cromwell-on-Azure?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
rtitle: Great to see PubSub support started! (even if TODO :)).

Eventually the azure pubsub handling should invoke methods on https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSAlgebra.scala

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
```diff
@@ -505,12 +513,217 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
...
+        )
+
+    // Get the disk to delete if specified
+    diskOpt = if (deleteDisk) appResult.app.appResources.disk.map(_.id) else None
+
+    googleProjectOpt = LeoLenses.cloudContextToGoogleProject.get(appResult.cluster.cloudContext)
+    _ <-
+      if (appResult.app.status == AppStatus.Error) {
+        for {
+          // we only need to delete Sam record for clusters in Google. Sam record for Azure is managed by WSM
```
rtitle: Not sure this is the case -- even for Azure apps Leo is creating the Sam resource I think (so we need to delete it)

rtitle (commented):
Made some comments (hope you don't mind) but looking good so far! We can always iterate on TODOs in separate PRs too.

http/src/main/resources/swagger/api-docs.yaml
n8landolt: Added!

http/src/main/resources/swagger/api-docs.yaml
n8landolt: Done! Could use a once over from another set of eyes

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
n8landolt: Should I create a more generic version of this in this ticket?

n8landolt: Could you review my updated implementation?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
n8landolt: Could you review my updated implementation?

rtitle:
I see the new changes, will look again tomorrow. The Sam stuff is tricky, want to think about it a bit

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpSamDAO.scala
```diff
@@ -179,6 +179,37 @@ class HttpSamDAO[F[_]](httpClient: Client[F],
         }
     } yield ()
 
+  override def createResourceV2[R](resource: R,
+                                   creatorEmail: WorkbenchEmail,
```
rtitle: It looks like `creatorEmail` is not used at all in this method and `cloudContext` is just used for logging. Thinking maybe we log `"${sr.resourceType.asString}/${sr.resourceIdAsString(resource)}"` instead and just remove these params?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpSamDAO.scala
```diff
@@ -179,6 +179,37 @@ class HttpSamDAO[F[_]](httpClient: Client[F],
...
+                                   cloudContext: CloudContext,
+                                   userInfo: UserInfo
+  )(implicit sr: SamResource[R], ev: Ask[F, TraceId]): F[Unit] = for {
+    traceId <- ev.ask
+    authHeader = Authorization(Credentials.Token(AuthScheme.Bearer, userInfo.accessToken.token))
```
rtitle, line 187: Noting there is a slight change compared to `createResource`. In `createResource` we request the pet key based on the `creatorEmail`, and make the Sam calls as the pet. In this method we call Sam with the user token directly.

I think using the token directly is probably fine for `createResourceV2`, since it comes from Front Leo (where we always have the token). However we might need to adjust this when we call `deleteResourceV2` from Back Leo (where we don't have the user token).

I think we can probably start with this though, and revisit as part of [adding deleteApp](https://broadworkbench.atlassian.net/browse/TOAZ-230) (I think there are a couple workable solutions..)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpSamDAO.scala
```diff
@@ -223,6 +254,40 @@ class HttpSamDAO[F[_]](httpClient: Client[F],
...
+    _ <- logger.info(
+      s"${traceId} | creating ${sr.resourceType.asString} resource in sam v2 for ${cloudContext}/${sr.resourceIdAsString(resource)}"
+    )
+    _ <- metrics.incrementCounter(s"sam/createResource/${sr.resourceType.asString}")
+    policies = Map[SamPolicyName, SamPolicyData](
+      SamPolicyName.Creator -> SamPolicyData(List(creatorEmail), List(SamRole.Creator))
+    )
+    parent = SerializableSamResource(SamResourceType.Project, WorkspaceResourceSamResourceId(workspaceId))
```
rtitle: I don't think this is right -- it's using `Project` as the resource type and the workspace id. I think the best thing to do for now is continue to use `SamResourceType.Project` and `ProjectSamResourceId(googleProject)` if the cloudContext is GCP -- and just not set a parent if the cloudContext is Azure. 

(We lose the inherited project-owner permissions in that case -- but honestly we'll be revisiting the Sam stuff anyway when we implement multi-tenancy.)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/SamAuthProvider.scala
```diff
@@ -201,13 +201,33 @@ class SamAuthProvider[F[_]: OpenTelemetryMetrics](
...
       samDao.createResourceWithParent(samResource, creatorEmail, googleProject)
 
+  override def notifyResourceCreatedV2[R](
+    samResource: R,
+    creatorEmail: WorkbenchEmail,
+    cloudContext: CloudContext,
+    workspaceId: WorkspaceId,
+    userInfo: UserInfo
+  )(implicit sr: SamResource[R], encoder: Encoder[R], ev: Ask[F, TraceId]): F[Unit] =
+    if (sr.resourceType != SamResourceType.App)
```
rtitle: Re my other comment, maybe we only call `createResourceWithParentV2` if the resourceType is App _and_ the cloud is GCP. Otherwise just `createResourceV2`.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpSamDAO.scala
```diff
@@ -265,6 +330,38 @@ class HttpSamDAO[F[_]](httpClient: Client[F],
         }
     } yield ()
 
+  override def deleteResourceV2[R](resource: R,
+                                   creatorEmail: WorkbenchEmail,
```
rtitle: Same as above: the `creatorEmail` and `cloudContext` aren't really used here.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpSamDAO.scala
rtitle: Also, is the only difference between `createResource` and `createResourceV2` the fact that the former uses a GCP pet to make the call?

Maybe we could name the `SamDAO` methods based on that difference in behavior, instead of just V1/V2:

```
def createResourceAsGcpPet(resource: R, creatorEmail: WorkbenchEmail, googleProject: GoogleProject)

def createResourceAsUser(resource: R, userInfo: UserInfo)

// etc
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpSamDAO.scala
```diff
@@ -278,6 +375,19 @@ class HttpSamDAO[F[_]](httpClient: Client[F],
         )
       )(onError)
 
+  def getPetManagedIdentity(authorization: Authorization)(implicit
+    ev: Ask[F, TraceId]
+  ): F[Option[WorkbenchEmail]] =
+    metrics.incrementCounter("sam/getPetServiceAccount") >>
```
rtitle: Maybe change the metric name

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
```diff
@@ -505,12 +485,247 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
...
+          AppCannotBeDeletedByWorkspaceIdException(workspaceId, appName, appResult.app.status, ctx.traceId)
+        )
+
+    // Get the disk to delete if specified
+    diskOpt = if (deleteDisk) appResult.app.appResources.disk.map(_.id) else None
+
+    _ <-
+      if (appResult.app.status == AppStatus.Error) {
+        for {
+          _ <- authProvider.notifyResourceDeletedV2(appResult.app.samResourceId,
```
rtitle: Yeah we call `notifyResourceDeletedV2` here, but only if the status is `Error`.. otherwise the Sam deletion will happen in Back Leo after polling for deletion is done. [example for GKE](https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala#L1196)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
rtitle: Looks good

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpSamDAO.scala
n8landolt: I'll keep the implementation as is!

I like the name adjustments, I'll do that.

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -16,6 +16,8 @@ tags:
     description: Persistent Disks API.
   - name: apps
     description: Experimental App API
+  - name: appsV2
+    description: Apps API v2. Support for Google Dataproc, Google Compute Engine, and Azure.
```
rtitle: Minor nit -- maybe change this text to
```
Apps API v2. Support for Google Kubernetes Engine and Azure Kubernetes Service.
```

(since the apps stuff is all kubernetes.)

Also technically v2 create/delete isn't supported as of this PR (the pubsub messages don't do anything yet ;)). I think it's still fine to add the routes to Swagger, maybe just add `Under active developement.` or something.

rtitle (approved):
Thumbing this! There may be some iteration/bugfixes once we wire things together but I think this is a good first pass at the route support.

Please make sure tests pass before merge -- would probably be good to get @Qi77Qi thumb too.

http/src/main/resources/swagger/api-docs.yaml
Qi77Qi: yea, apps are only on k8s

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
```diff
@@ -505,12 +485,244 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
...
+        getSavableApp(
+          cloudContext,
+          appName,
+          userInfo.userEmail,
+          samResourceId,
+          req,
+          None, // diskResultOpt.map(_.disk),
+          None, // lastUsedApp,
+          petSA,
+          NodepoolLeoId(123123123), // TODO: Fix this //nodepool.id,
```
n8landolt: What do we do about the NodepoolLeoID?
seems like we're missing machineConfig and clusterId from the code block in createApp.

n8landolt: @Qi77Qi @rtitle

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpSamDAO.scala
```diff
@@ -265,6 +329,37 @@ class HttpSamDAO[F[_]](httpClient: Client[F],
         }
     } yield ()
 
+  override def deleteResourceV2[R](resource: R, userInfo: UserInfo)(implicit
```
Qi77Qi: maybe `deleteResourceWithUserInfo` instead of `deleteResourceV2`? `deleteResourceV2` reads to me we're using v2 api or sth

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
```diff
@@ -505,12 +485,244 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
...
+      )
+
+      samResourceId <- F.delay(AppSamResourceId(UUID.randomUUID().toString))
+
+      cloudContext <- (workspaceDesc.azureContext, workspaceDesc.gcpContext) match {
+        case (Some(azureContext), _) => F.pure[CloudContext](CloudContext.Azure(azureContext))
+        case (_, Some(gcpContext))   => F.pure[CloudContext](CloudContext.Gcp(gcpContext))
+        case (None, None) => F.raiseError[CloudContext](CloudContextNotFoundException(workspaceId, ctx.traceId))
+      }
+      // TODO Verify managed identiy is covered by lookupOriginatingUserEmail
```
Qi77Qi, line 545: so we did the TODO?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
Qi77Qi: where will disk be coming from for say cromwell chart without it?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
Qi77Qi: we can leave this TODO as is for now...it'll come from WSM api which doesn't support it yet

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
```diff
@@ -505,12 +485,244 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
...
+          ctx
+        )
+      )
+      app <- appQuery.save(saveApp, Some(ctx.traceId)).transaction
+
+      createAppV2Message = CreateAppV2Message(
+        workspaceId,
+        app.id,
+        app.appName,
+        None, // diskResultOpt.flatMap(d => if (d.creationNeeded) Some(d.disk.id) else None),
```
Qi77Qi: Add TODO


We should add TODO in all places where we're populating temporary values. Otherwise, it'll be pretty hard to find these spots once the PR is merged...For larger items, tickets will be ideal

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
```diff
@@ -505,12 +485,244 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
...
+    ctx <- as.ask
+    appOpt <- KubernetesServiceDbQueries.getActiveFullAppByWorkspaceIdAndAppName(workspaceId, appName).transaction
+    appResult <- F.fromOption(
+      appOpt,
+      AppNotFoundByWorkspaceIdException(workspaceId, appName, ctx.traceId, "No active app found in DB")
+    )
+    listOfPermissions <- authProvider.getActions(appResult.app.samResourceId, userInfo)
+
+    // throw 404 if no GetAppStatus permission
+    hasReadPermission = listOfPermissions.toSet.contains(AppAction.GetAppStatus)
```
Qi77Qi, line 663: as I mentioned in slack, if these are all just copy from v1 implementation, let's use a common function

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
n8landolt: I'm not entirely sure how to address this portion.

Do I add another AppType, create CromwellOnAzureAppConfig with the same fields referenced in your link and add a new case under getSavableApp?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpSamDAO.scala
n8landolt: I like it - I'll actually adjust the createResourceAsUser to be WithUserInfo for consistency.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
n8landolt: I have another commit incoming, the TODO will be moved to a different spot.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
rtitle: Cromwell on Azure chart doesn't have a persistent disk parameter -- it mounts a postgres DB and writes logs/etc to a blob storage container. So this implementation is fine I think..

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
rtitle: What is the resulting data in the DB after `getSavableApp` is called, on both clouds? (Maybe this would be a good unit test).

In particular this logic might need to be changed: https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala#L687-L691

Currently it handles 3 app types: Galaxy, Cromwell, and Custom. Cromwell-on-Azure is a separate helm chart though, so there should be some case which results in populating this information from conf: https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/resources/reference.conf#L209-L216

I wasn't sure if we want to make `CromwellOnAzure` a separate AppType, or have logic like:
```
appType match {
  case Galaxy   => config.leoKubernetesConfig.galaxyAppConfig
  case Cromwell if cloud == GCP => config.leoKubernetesConfig.cromwellAppConfig
  case Cromwell if cloud == Azure => config.leoKubernetesConfig.coaAppConfig
  case Custom   => config.leoKubernetesConfig.customAppConfig
}
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
rtitle: Yeah for Azure we don't actually care about the nodepool and cluster in the Leo DB, since all that comes from the landing zone.

Qi77Qi (approved):
wanna give a thumb before I go on vacation....

the only thing we need to be careful is not to break existing app APIs ideally. Once tests all pass, things should be mostly okay. But keep an eye on any reported issues on app APIs from @yonghaoy

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
```diff
@@ -375,15 +379,50 @@ final case class App(id: AppId,
...
                      extraArgs: List[String]
 ) {
-  def getProxyUrls(project: GoogleProject, proxyUrlBase: String): Map[ServiceName, URL] =
+
+  def getProxyUrls(cloudContext: CloudContext,
+                   workspaceId: Option[WorkspaceId],
+                   proxyUrlBase: String,
+                   apiVersion: String
+  ): Map[ServiceName, URL] =
     appResources.services.map { service =>
```
rtitle: The logic here is kind of tricky actually. :) 

Thinking through the cases, I think we want to generate v1 or v2 style URLs based on the API being called. But we probably shouldn't fail if there are apps in the system that we can't generate a URL for -- e.g. if I call `listAppsV2` and I have some v1 apps in the system.

I came up with this (warning: didn't compile!). Does this cover all the cases? Btw I like using a pattern match to describe this type of thing.

```
def getProxyUrls(cloudContext: CloudContext,
                   workspaceId: Option[WorkspaceId],
                   proxyUrlBase: String,
                   apiVersion: String
): Map[ServiceName, URL] =
  appResources.services.flatMap { service =>
    val proxyPathOpt = (apiVersion, cloudContext, workspaceId) match {
      // v1 apis should always generate urls for google apps
      case ("v1", CloudContext.Gcp(project), _) =>  
        Some(s"google/v1/apps/${project.value}/${appName.value}/${kubernetesService.config.name.value}")
      // v2 apis should generate urls for apps on both clouds, but only if there's a workspace
      case ("v2", _, Some(workspace)) => 
        Some(s"apps/v2/${workspaceId.value}/${appName.value}/${kubernetesService.config.name.value}")
      // otherwise don't generate a proxy url
      case _ =>
        None
    }
    proxyPathOpt match {
      case Some(proxyPath) =>
        Map(kubernetesService.config.name -> new URL(s"${proxyUrlBase}${proxyPath}${kubernetesService.config.path.getOrElse("")}"))
      case None =>
        Map.empty
    }
  }
```

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/appRoutesModels.scala
```diff
@@ -71,8 +81,10 @@ object ListAppResponse {
           ),
           a.errors,
           a.status,
-          a.getProxyUrls(c.cloudContext.asInstanceOf[CloudContext.Gcp].value,
-                         proxyUrlBase
+          a.getProxyUrls(c.cloudContext,
+                         c.workspaceId,
+                         proxyUrlBase,
+                         apiVersion
           ), // TODO: refactor once we support proxying azure app
```
rtitle: Can we remove this TODO comment?

rtitle (commented):
One suggestion above w.r.t. proxyUrl generation.

Overall this is looking good. There may be some things to fix once we start wiring the v2 routes into Back Leo, but I think any issues can be fixed in subsequent PRs. Please just make sure these are no regressions in the v1 app routes (I think there are swatomation tests to cover these?), then I think this is probably mergeable.

yonghaoy:
Hello as AoU starts integrating with Leo APP apis, few questions I have: 

- Does this support GCP workspace? If so, does it support Rawls powered GCP workspace or WSM powered workspace? 
-  Is that something AoU can/should start using?

rtitle:
@yonghaoy 

The end state is the v2 app routes will support GCP and Azure. But the v2 routes do require a WSM-powered workspace. Since only Azure workspaces are WSM-powered at the moment, I think that means in the short term that clients should continue using v1 for GCP and v2 for Azure.

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
n8landolt: This looks great! I fixed up the errors and I like it a lot more than my first solution.
Mostly did it as a first pass, so I appreciate the follow up.

rtitle (approved):

rtitle:
jenkins retest

PR #2911, opened 2022-10-20
[IA-3796] Add appName, cloudContext and lables into GetAppResponse

yonghaoy:
It is helpful to make GetAppResponse returns simialr field as listAppResponse. And it is easier for us to write mapper from LeoResponse to AoU APP object

---

In all cases:

### DIFF
--- a/http/src/main/resources/swagger/api-docs.yaml
+++ b/http/src/main/resources/swagger/api-docs.yaml
@@ -3498,7 +3498,6 @@ components:
         cloudResource:
           type: string
           description: Cloud resource name. Google project if cloud provider is GCP OR Azure's tenantId/subscriptionId/managedResourceGroupId if in Azure
-          example: n1-standard-1
 
     KubernetesRuntimeConfig:
       description: configuration for a kubernetes app runtime
--- a/http/src/main/resources/swagger/api-docs.yaml
+++ b/http/src/main/resources/swagger/api-docs.yaml
@@ -3411,6 +3411,12 @@ components:
         - auditInfo
       type: object
       properties:
+        appName:
+          type: string
+          description: the name of the app
+        googleProject:
+          type: string
+          description: Deprecated. The project this app is associated with
         kubernetesRuntimeConfig:
           $ref: '#/components/schemas/KubernetesRuntimeConfig'
         errors:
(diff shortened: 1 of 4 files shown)


http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -3411,6 +3411,12 @@ components:
         - auditInfo
       type: object
       properties:
+        appName:
+          type: string
+          description: the name of the app
+        cloudContext:
```
Qi77Qi, line 3422: [this](https://github.com/DataBiosphere/leonardo/blob/develop/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodec.scala#L93) is encoder for `cloudContext`, it will be encoded into sth like
```
"cloudProvider": "",
"cloudResource":"",
```

yonghaoy: I was following the cloudContext in ListAppResponse below: https://github.com/DataBiosphere/leonardo/pull/2911/files#diff-6fad074e668b9395b43409265ad159ab303e3e3ff3c3d63bff4595706a3da88fR3457-R3459
Is that piece also wrong?

Qi77Qi: that link is just diff...which line you're referring to?

Qi77Qi: this is what it looks like for listApp

```
    "cloudContext": {
      "cloudProvider": "GCP",
      "cloudResource": "terra-dev-a6dd5d8f"
    },
    "googleProject": "terra-dev-a6dd5d8f",
```

yonghaoy: That link is ListAPP's cloudContext.  Why those are in string format? is that correct? 
https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/resources/swagger/api-docs.yaml#L3448
https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/resources/swagger/api-docs.yaml#L2548

Qi77Qi: I didn't verify the runtime one(most likely it's also wrong)..but listApp is definitely wrong..

yonghaoy: ah, okay. Create a `CloudContext`. PTAL.  One thing I am not sure is the case sensitive enum mapping.
Not sure if [CloudProvider.Azure ](https://github.com/DataBiosphere/leonardo/blob/develop/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodec.scala#L352) can be mapped to `AZURE`

Qi77Qi (approved):

mikenev (approved):

PR #2925, opened 2022-10-24
TOAZ-229 Poll Created Azure App for Status.

Ghost-in-a-Jar:
Leo is in charge of starting up apps (collections of services in this context) in azure. This PR adds polling of cromwell's http status endpoint, at the end marking the app as running once all status endpoints return 200/ok. Subsequent pr's will add the other service daos/logic.

---

In all cases:


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpCromwellDAO.scala
```diff
@@ -0,0 +1,96 @@
...
+                            samDAO: SamDAO[F]
+)(implicit logger: Logger[F], F: Async[F], metrics: OpenTelemetryMetrics[F])
+    extends CromwellDao[F]
+    with Http4sClientDsl[F] {
+
+  private val retryable = (_: Request[F], resp: Either[Throwable, Response[F]]) => resp match {
+    case Left(_) => true
+    case Right(r: Response[F]) => ???
+  }
+  private val retryPolicy = RetryPolicy[F](RetryPolicy.exponentialBackoff(30 seconds, 5), retryable)
```
rtitle: Not sure if we want `exponentialBackoff`, as we're just waiting for the deployed service to come up, not backing off an error.

In fact our other app polling logic doesn't use http4s `Retry` -- it makes a non-retryable https call and polls it with an fs2 stream. Example: https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala#L1168-L1175

I suspect we want something similar here..

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpCromwellDAO.scala
```diff
@@ -0,0 +1,96 @@
...
+          F.pure(Headers.empty)
+      }
+      res <- hostStatus match {
+        case host: HostReady =>
+
+          metrics.incrementCounter("cromwell/status") >>
+            httpClientWithRetry.expectOr[CromwellStatusCheckResponse](
+            Request[F](
+              method = Method.GET,
+              uri = host.toUri.withPath(Uri.Path.unsafeFromString(s"api/engine/v1/status")),
```
rtitle: This calls Cromwell, but I think we'll want to aggregate status for WDS and CBAS status endpoints as well. The Leo "app" should only come online if all 3 services are OK status.

Not sure if we want to do that aggregation in `HttpCromwellDAO`, or make `HttpWdsDAO` and `HttpCbasDAO` classes as well…?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpCromwellDAO.scala
```diff
@@ -0,0 +1,96 @@
...
+    case Left(_) => true
+    case Right(r: Response[F]) => ???
+  }
+  private val retryPolicy = RetryPolicy[F](RetryPolicy.exponentialBackoff(30 seconds, 5), retryable)
+  private val httpClientWithRetry = Retry(retryPolicy)(httpClient)
+
+  import HttpCromwellDAO._
+
+  override def getStatus(cloudContext: CloudContext, runtimeName: RuntimeName)(implicit
+    ev: Ask[F, TraceId]
```
rtitle: Maybe make this take `ev: Ask[F, AppContext]` instead

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpCromwellDAO.scala
```diff
@@ -0,0 +1,96 @@
...
+      }
+    } yield res
+
+
+
+  private def onError(response: Response[F])(implicit ev: Ask[F, TraceId]): F[Throwable] =
+    for {
+      traceId <- ev.ask
+      body <- response.bodyText.compile.foldMonoid
+      _ <- logger.error(s"${traceId} | Cromwell call failed: $body")
```
rtitle: Nit - I think we can do `logger.error(ctx.loggingCtx)(msg)` to use MDC and avoid embedding the traceId in the log string.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/Boot.scala
```diff
@@ -345,6 +324,9 @@ object Boot extends IOApp {
       samDao <- buildHttpClient(sslContext, proxyResolver.resolveHttp4s, Some("leo_sam_client"), true).map(client =>
         HttpSamDAO[F](client, httpSamDaoConfig, petTokenCache)
       )
+      cromwellDao <- buildHttpClient(sslContext, proxyResolver.resolveHttp4s, Some("leo_cromwell_client"), true).map(
```
rtitle: WRT the above comment, can pass `false` here to avoid retrying in the http4s client.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpCromwellDAO.scala
Ghost-in-a-Jar: Yeah I am thinking we do the aggregation in AKSInterpreter

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpCromwellDAO.scala
Ghost-in-a-Jar: Ah yeah I totally know what youre saying but that specifically didnt work

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -106,8 +111,11 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
       _ <- logger.info(ctx.loggingCtx)(
         s"Finished app creation for app ${params.appName.value} in cluster ${landingZoneResources.clusterName.value} in cloud context ${params.cloudContext.asString}"
       )
+      // Poll app status
+      _ <- streamFUntilDone(
+          cromwellDao.getStatus(CloudContext.Azure(params.cloudContext), RuntimeName(params.appName.value)),
+        maxAttempts=6, delay=10 seconds).interruptAfter(1 minute).compile.lastOrError
```
rtitle: I think want to increase this to something like 20 minutes total, polling every 10 seconds. We actually define this in config [here](https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/resources/reference.conf#L793-L797), and read it for GKE [here](https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala#L1173-L1175). We can probably reuse the config objects for AKS, I think.

And as mentioned above, let's get rid of the `RetryPolicy` in `HttpCromwellDAO`, so all polling happens here.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/CromwellDao.scala
```diff
@@ -0,0 +1,10 @@
+package org.broadinstitute.dsde.workbench.leonardo.dao
+
+import cats.mtl.Ask
+import org.broadinstitute.dsde.workbench.leonardo.{AppContext, CloudContext, RuntimeName}
+
+trait CromwellDao[F[_]] {
+  def getStatus(cloudContext: CloudContext, runtimeName: RuntimeName)(implicit
```
Qi77Qi: I think it should be called `appName` instead of `runtimeName`...runtimes in Leo are referring to either VMs or dataproc clusters specifically. Not k8s apps

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpCromwellDAO.scala
```diff
@@ -0,0 +1,91 @@
...
+import org.http4s.circe.CirceEntityDecoder._
+import org.http4s.client.Client
+import org.http4s.client.dsl.Http4sClientDsl
+import org.http4s.client.middleware.{Retry, RetryPolicy}
+import org.typelevel.log4cats.Logger
+
+import scala.concurrent.duration._
+import scala.util.control.NoStackTrace
+
+class HttpCromwellDAO[F[_]](val runtimeDnsCache: RuntimeDnsCache[F], httpClient: Client[F], samDAO: SamDAO[F])(implicit
```
Qi77Qi: I don't think we need `RuntimeDnsCache` anymore...becuz cromwell on Azure's proxy path will be using relay

rtitle: I think we should plan for `HttpCromwellDAO` to support GKE and AKS.

…However `RuntimeDnsCache` is for VMs, this should probably use `KubernetesDnsCache` for GKE.

…And I agree, for Azure we probably don't need a DNS cache at all. Can we just build the relay hybrid connection URL directly (how is that done in other places)?

Qi77Qi: yea you should be able to build the relay url directly..for runtime, we persist the relay url in DB, and the hybridconnection name is created using the runtime name

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
Ghost-in-a-Jar: ok sounds good yeah this was just a temporary measure.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpCromwellDAO.scala
rtitle: OK.. for apps we are also creating the hybridConnection as the app name: https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala#L111

But we're not persisting it in the DB. Perhaps we need to add a call like [this](https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala#L231) in AKSInterpreter..

Qi77Qi: yea, I think that makes sense...all apps on the same cluster will have the same relay namespace..

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -327,5 +354,9 @@ final case class AKSInterpreterConfig(terraAppSetupChartConfig: TerraAppSetupCha
                                       coaAppConfig: CoaAppConfig,
                                       aadPodIdentityConfig: AadPodIdentityConfig,
                                       appRegistrationConfig: AzureAppRegistrationConfig,
-                                      samConfig: SamConfig
+                                      samConfig: SamConfig,
+                                      pollingConfig: PollingConfig =
+                                        PollingConfig(maxAttempts = 120, delay = 10 seconds, interruptAfter = 20 minutes)
```
Ghost-in-a-Jar: Does this seem alright or should I expose these as environment variables/.conf values?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -323,9 +350,13 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
...
-                                      aadPodIdentityConfig: AadPodIdentityConfig,
-                                      appRegistrationConfig: AzureAppRegistrationConfig,
-                                      samConfig: SamConfig
+final case class AKSInterpreterConfig(
+  terraAppSetupChartConfig: TerraAppSetupChartConfig,
+  coaAppConfig: CoaAppConfig,
+  aadPodIdentityConfig: AadPodIdentityConfig,
+  appRegistrationConfig: AzureAppRegistrationConfig,
+  samConfig: SamConfig,
+  pollingConfig: PollingConfig = PollingConfig(maxAttempts = 120, delay = 10 seconds, interruptAfter = 20 minutes)
```
Ghost-in-a-Jar: Does this seem alright or should I have these be environment variables/.conf values?

rtitle (commented):
This looks good I think -- going to try and test it out in my AKS setup

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -142,14 +143,41 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
...
 
+  private def pollAppsForSuccess(relayNamespace: RelayNamespace)(implicit ev: Ask[F, AppContext]): F[Boolean] = {
+    val relayBaseUri = Uri.unsafeFromString(s"https://${relayNamespace.value}.servicebus.windows.net")
+    val cromwellStatusUri = relayBaseUri / "cromwell" / "api" / "engine" / "v1" / "status"
+    for {
+
+      headers <- samDao.getLeoAuthToken.map(x => Headers(x))
+      cromwellOk <- streamFUntilDone(
+        cromwellDao.getStatus(cromwellStatusUri, headers),
+        maxAttempts = config.pollingConfig.maxAttempts,
```
Qi77Qi: we probably should have a dedicated poll config for polling comwell

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -323,9 +351,13 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
...
-                                      aadPodIdentityConfig: AadPodIdentityConfig,
-                                      appRegistrationConfig: AzureAppRegistrationConfig,
-                                      samConfig: SamConfig
+final case class AKSInterpreterConfig(
+  terraAppSetupChartConfig: TerraAppSetupChartConfig,
+  coaAppConfig: CoaAppConfig,
+  aadPodIdentityConfig: AadPodIdentityConfig,
+  appRegistrationConfig: AzureAppRegistrationConfig,
+  samConfig: SamConfig,
+  pollingConfig: PollingConfig = PollingConfig(maxAttempts = 120, delay = 10 seconds, interruptAfter = 20 minutes)
```
Qi77Qi: we should add the actual value https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/resources/reference.conf#L772

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpCromwellDAO.scala
```diff
@@ -0,0 +1,70 @@
...
+
+object HttpCromwellDAO {
+  implicit val statusDecoder: Decoder[CromwellStatusCheckResponse] = Decoder.instance { d =>
+    for {
+      ok <- d.downField("ok").as[Boolean]
+    } yield CromwellStatusCheckResponse(ok)
+  }
+}
+
+class HttpCromwellDAO[F[_]](httpClient: Client[F], samDAO: SamDAO[F])(implicit
```
Qi77Qi: this looks specific to cromwell on azure? if that's true, can we rename the class

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
rtitle: Maybe.. we could start with the values in conf [pubsub.kubernetes-monitor](https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/resources/reference.conf#L793) and make a dedicate cromwell config if needed..

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpCromwellDAO.scala
rtitle: I don't _think_ this class is cloud specific, it is just concerned with checking Cromwell `/status` (which could be on either cloud). 

(The method takes `uri: Uri, headers: Headers` params which will be different depending on cloud.)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -142,14 +143,41 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
...
-      // TODO (TOAZ-229): poll app for completion
-
     } yield ()
 
+  private def pollAppsForSuccess(relayNamespace: RelayNamespace)(implicit ev: Ask[F, AppContext]): F[Boolean] = {
+    val relayBaseUri = Uri.unsafeFromString(s"https://${relayNamespace.value}.servicebus.windows.net")
+    val cromwellStatusUri = relayBaseUri / "cromwell" / "api" / "engine" / "v1" / "status"
+    for {
+
+      headers <- samDao.getLeoAuthToken.map(x => Headers(x))
```
rtitle: I will test but not sure the Leo auth token will work here since it's not a member of the `kubernetes-app` Sam resource.

We may want to use the UAMI token instead (which we also need for calling the LZ API I think)… I can look into this.

http/src/main/resources/reference.conf
```diff
@@ -756,9 +756,9 @@ auth {
   providerClass = "org.broadinstitute.dsde.workbench.leonardo.auth.sam.SamAuthProvider"
   providerConfig {
     samServer = ${sam.server}
-    petTokenCacheEnabled = true
-    petTokenCacheExpiryTime = 60 minutes
-    petTokenCacheMaxSize = 1000
+    petKeyCacheEnabled = true
```
rtitle, line 759: I changed `HttpSamDAO` to cache pet _keys_ (which is what we're actually requesting from Sam) instead of pet tokens. We can generate tokens from keys on-demand. Keys don't expire, so this lets us have a longer cache expiry, which should improve performance I think.

Qi77Qi: ah that's great!..thanks!

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/CbasDAO.scala
```diff
@@ -0,0 +1,12 @@
+package org.broadinstitute.dsde.workbench.leonardo.dao
+
+import cats.mtl.Ask
+import org.broadinstitute.dsde.workbench.leonardo.AppContext
+import org.http4s.Uri
+import org.http4s.headers.Authorization
+
+trait CbasDAO[F[_]] {
```
rtitle, line 8: I made `CbasDAO`, `CromwellDAO`, and `WdsDAO`. They are similar so there is some duplication, but they have slight differences (different status endpoints, payloads).

Each of these DAOs are cloud-agnostic, so can be used on Azure or GCP.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpSamDAO.scala
```diff
@@ -146,15 +146,12 @@ class HttpSamDAO[F[_]](httpClient: Client[F],
...
-      token <- getCachedPetAccessToken(creatorEmail, googleProject).flatMap(
-        _.fold(
-          F.raiseError[String](
-            AuthProviderException(traceId,
-                                  s"No pet SA found for ${creatorEmail} in ${googleProject}",
-                                  StatusCodes.Unauthorized
-            )
-          )
-        )(s => F.pure(s))
+      tokenOpt <- getCachedPetAccessToken(creatorEmail, googleProject)
```
rtitle, line 149: Just some cleanup, use `F.fromOption`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpSamDAO.scala
```diff
@@ -350,36 +341,52 @@ class HttpSamDAO[F[_]](httpClient: Client[F],
...
+    } yield token
+
+  override def getCachedArbitraryPetAccessToken(
+    userEmail: WorkbenchEmail
+  )(implicit ev: Ask[F, TraceId]): F[Option[String]] =
+    for {
+      keyOpt <-
+        if (config.petCacheEnabled) {
+          // Cache only by email in the "arbitrary pet" case
+          petKeyCache.cachingF(UserEmailAndProject(userEmail, GoogleProject("user-shell-project")))(None)(
```
rtitle, line 363: Depends on https://github.com/broadinstitute/sam/pull/889/

rtitle:
@Qi77Qi I made some changes to this, wondering if you could take another look? Hopefully it should be working now. Testing it in a BEE now..

rtitle:
jenkins retest

rtitle:
Not quite working yet. Breadcrumb for next week:

The status check is hitting the app (CBAS in this case):
```
"HTTP/1.1 GET https://relay-ns-f01401d5-fb13-47cc-bc39-9f9ee17ad209.servicebus.windows.net/cbas/status Headers(Authorization: <REDACTED>, Accept: application/json, Accept: application/json)"
```

The response is 200:
```
 "HTTP/1.1 200 OK Headers(Transfer-Encoding: chunked, Content-Type: application/atom+xml;type=feed;charset=utf-8, Server: Microsoft-HTTPAPI/2.0, Strict-Transport-Security: max-age=31536000, Date: Fri, 04 Nov 2022 20:50:11 GMT)"
```

The app creation errors out:
```
"Encountered an error for app Some(AppId(26)), An error occurred with a kubernetes operation from source app during action createApp. 
Original message: Error creating Azure app with id 26 and cloudContext 0cb7a640-45a2-4ed6-be9f-63519f86e04b/3efc5bdf-be0e-44e7-b1d7-c08931e3c16c/mrg-rt-november2-20221102094002: Malformed message body: Invalid JSON"
```

I am not sure why I see `application/atom+xml` as the content-type in the response. When I `curl` it it's `application/json`. Some http4s thing??

Maybe this'll be more clear next week :-D

rtitle:
Oh! I think the URL is wrong -- it needs to have the appName. Fixing…

rtitle:
It works now as of https://github.com/DataBiosphere/leonardo/pull/2925/commits/880412a1d4084996bdbac420f752456c2c5f5135 :-D

Will look to merge next week.

rtitle (approved):

Ghost-in-a-Jar:
I would approve this if i could

rtitle:
jenkins retest

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/CbasDAO.scala
Qi77Qi: doesn't look like this is being used for GCP? or does this not apply to GCP yet?

rtitle: Yeah it's not used for GCP but probably should be (as CBAS is deployed on GCP today). I think `GKEIntepreter` today just uses https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpAppDAO.scala to poll the root endpoint of the app, but not all individual services that might be running on the app..

PR #2976, opened 2022-11-03
IA-3822 wait for welder to come up as well before marking Azure runtime as Running

Qi77Qi:

---

In all cases:

### DIFF
--- a/http/src/test/resources/reference.conf
+++ b/http/src/test/resources/reference.conf
@@ -300,7 +300,7 @@ azure {
    pubsub-handler {
         create-vm-poll-config {
            initial-delay = 1 seconds
-           max-attempts = 120
+           max-attempts = 10
            interval = 1 seconds
        }
        delete-vm-poll-config {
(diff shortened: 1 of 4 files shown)


http/src/test/resources/reference.conf
```diff
@@ -300,7 +300,7 @@ azure {
    pubsub-handler {
         create-vm-poll-config {
            initial-delay = 1 seconds
-           max-attempts = 120
+           max-attempts = 10
            interval = 1 seconds
```
n8landolt, line 304: Do we want to increase the interval to give the service more time to come up?

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/ConfigReaderSpec.scala
```diff
@@ -38,7 +38,7 @@ class ConfigReaderSpec extends AnyFlatSpec with Matchers {
           Uri.unsafeFromString("https://localhost:8000"),
           "terradevacrpublic.azurecr.io/welder-server",
           "ef956b2",
-          PollMonitorConfig(1 seconds, 120, 1 seconds),
+          PollMonitorConfig(1 seconds, 10, 1 seconds),
```
n8landolt, line 41: I made this adjustment for you, test was failing.

n8landolt (approved):

http/src/test/resources/reference.conf
Qi77Qi: nah...this is only for unit tests where we just mock the results typically..so shorter attempts make sense

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/ConfigReaderSpec.scala
Qi77Qi: ah great..thanks!

PR #2984, opened 2022-11-07
TOAZ-259 don't override the path for Cromwell-on-GCP

rtitle:
https://broadworkbench.atlassian.net/browse/TOAZ-259

Should fix it I think. I suspect the reason our tests didn't catch this is we just check that the app reaches Running status, but don't actually hit the proxyUrl.

We are adding Azure tests anyway as a separate ticket, suggest we make that improvement there.

I can verify this quickly in a BEE..

---

In all cases:


cjllanwarne (approved):

http/src/main/resources/reference.conf
```diff
@@ -533,7 +533,6 @@ gke {
       {
         name = "cromwell-service"
         kind = "ClusterIP"
-        path = "/"
```
salonishah11, line 536: I am not sure why this needs to removed now as it has been here for few months now. But if it fixes the issue then 👍

salonishah11 (approved):

n8landolt (approved):

PR #2986, opened 2022-11-07
TOAZ-230 Delete Azure App

Ghost-in-a-Jar:
Adds:
- handling of app delete pubsub message
- app deletion logic (list pod statuses, delete namespace)
- AKSInterpreter integration test

---

In all cases:


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -55,6 +64,14 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
   implicit private def booleanDoneCheckable: DoneCheckable[Boolean] = identity[Boolean]
   implicit private def listDoneCheckable[A: DoneCheckable]: DoneCheckable[List[A]] = as => as.forall(_.isDone)
 
+  private[util] def isPodDone(pod: KubernetesPodStatus): Boolean =
+    pod.podStatus == PodStatus.Failed || pod.podStatus == PodStatus.Succeeded
+  implicit private def podDoneCheckable: DoneCheckable[List[KubernetesPodStatus]] =
+    (ps: List[KubernetesPodStatus]) => ps.forall(isPodDone)
+
+  private def getTerraAppSetupChartReleaseName(appReleaseName: Release): Release =
+    Release(s"${appReleaseName.asString}-setup-rls")
```
Ghost-in-a-Jar, line 79: is this the same in aks?

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreterSpec.scala
```diff
@@ -151,9 +153,22 @@ class AKSInterpreterSpec extends AnyFlatSpecLike with TestComponent with Leonard
...
+
+    for {
+      _ <- aksInterp.deleteApp(DeleteAKSAppParams(app.appName, workspaceId, cloudContext))
+      app <- KubernetesServiceDbQueries
+        .getActiveFullAppByName(CloudContext.Azure(cloudContext), app.appName)
+        .transaction
+    } yield {
+      app shouldBe defined
+      app.get.app.status shouldBe AppStatus.Deleted
+    }
```
Ghost-in-a-Jar: Any other checks/assertions I could do here?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -55,6 +64,14 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
   implicit private def booleanDoneCheckable: DoneCheckable[Boolean] = identity[Boolean]
   implicit private def listDoneCheckable[A: DoneCheckable]: DoneCheckable[List[A]] = as => as.forall(_.isDone)
 
+  private[util] def isPodDone(pod: KubernetesPodStatus): Boolean =
+    pod.podStatus == PodStatus.Failed || pod.podStatus == PodStatus.Succeeded
+  implicit private def podDoneCheckable: DoneCheckable[List[KubernetesPodStatus]] =
+    (ps: List[KubernetesPodStatus]) => ps.forall(isPodDone)
```
Ghost-in-a-Jar: pulled these verbatim from gkeInterpreter

Ghost-in-a-Jar: the above two fns

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSAlgebra.scala
```diff
@@ -9,7 +9,9 @@ trait AKSAlgebra[F[_]] {
   /** Creates an app and polls it for completion */
   def createAndPollApp(params: CreateAKSAppParams)(implicit ev: Ask[F, AppContext]): F[Unit]
 
-  // TODO (TOAZ-230): delete, stop, start
+  def deleteApp(params: DeleteAKSAppParams)(implicit ev: Ask[F, AppContext]): F[Unit]
+
+  // TODO (TOAZ-230): stop, start
```
rtitle: TOAZ-230 is just for delete app -- don't think we have a ticket for stop/start yet, maybe just remove the comment.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -40,6 +48,7 @@ import scala.jdk.CollectionConverters._
 
 class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
                            helmClient: HelmAlgebra[F],
+                           kubeService: org.broadinstitute.dsde.workbench.google2.KubernetesService[F],
```
rtitle: I think it's right to use `KubernetesService`, but I suspect the implementation is GKE-specific right now. It uses a `com.google.container.v1.Cluster` to create a client: https://github.com/broadinstitute/workbench-libs/blob/develop/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesInterpreter.scala#L434

There should be a way to make this cloud agnostic, will think about this..

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
rtitle: Yup - can you update `createAndPollApp` to use this method when deploying the setup chart?

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreterSpec.scala
rtitle: This seems like enough -- Sam/helm/etc are all mocked out but we can assert DB status.

I don't think this last for-comp is running though because there is no `unsafeRunSync`..

rtitle (commented):
The logic in `AKSInterpreter` looks good. I think we need to figure out how to use `KubernetesService` on Azure (I can help with that).

Can you also add the PubSub wiring which actually calls `AKSInterpreter`? 
* [LeoPubSubMessageSubscriber](https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala#L125) which receives the `DeleteAppV2Message`
* [LeoPublisher](https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/LeoPublisher.scala#L128) which sends the message
* [AppServiceInterp](https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala#L653) which builds the message
* By the way we can change [DeleteAppV2Message](https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/leoPubsubMessageSubscriberModels.scala#L285) based on what we actually need -- this message is only used for the v2 routes currently, and therefore only used for Azure.

Thanks!

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -45,7 +66,8 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
                            samDao: SamDAO[F],
                            cromwellDao: CromwellDAO[F],
                            cbasDao: CbasDAO[F],
-                           wdsDao: WdsDAO[F]
+                           wdsDao: WdsDAO[F],
+                           apiClientCache: Cache[F, AKSClusterName, ApiClient]
```
rtitle: Do we really need a cache of k8s clients? I think the code in `AKSInterpreter.create/deleteApp` is already calling `getClusterCredential` every time to auth with helm. Could we just build a k8s client each time, and use it for the duration of the polling?

Just thinking the cache adds some complexity which we don't really need..

Ghost-in-a-Jar: Yeah I was thinking the same thing but this is how it is written in workbench libs -- I agree lets take it out

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/Boot.scala
```diff
@@ -472,9 +477,20 @@ object Boot extends IOApp {
         200,
         2 hours
       )
+
       kubeCache <- Resource.make(F.delay(CaffeineCache[F, KubernetesClusterId, ApiClient](underlyingKubeClientCache)))(
         _.close
       )
+
+      underlyingAksKubeClientCache = buildCache[AKSClusterName, scalacache.Entry[ApiClient]](
```
rtitle: Seems these aren't used anymore - can they be removed?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -1447,4 +1445,37 @@ class LeoPubsubMessageSubscriber[F[_]](
...
+    msg: DeleteAppV2Message
+  )(implicit ev: Ask[F, AppContext]): F[Unit] =
+    for {
+      ctx <- ev.ask
+      _ <- msg.cloudContext match {
+        case CloudContext.Azure(c) =>
+          val task =
+            azurePubsubHandler.deleteApp(msg.appId, msg.appName, msg.workspaceId, msg.landingZoneResourcesOpt, c)
+          asyncTasks.offer(
+            Task(ctx.traceId, task, Some(handleKubernetesError), ctx.now, "createAppV2")
```
rtitle: s/createAppV2/deleteAppV2

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -1447,4 +1445,37 @@ class LeoPubsubMessageSubscriber[F[_]](
...
+          )
+        case CloudContext.Gcp(c) =>
+          F.raiseError(
+            PubsubKubernetesError(
+              AppError(
+                s"Error creating GCP app with id ${msg.appId} and cloudContext ${c.value}: CreateAppV2 not supported for GCP",
```
rtitle: s/create/delete :)

rtitle (approved):
A few nits - but overall looks correct. Giving 👍 pending testing on a real deployed app

rtitle:
Looks like a swatomation test flake :( jenkins retest

rtitle:
Sigh. This did pass once: https://fc-jenkins.dsp-techops.broadinstitute.org/job/leonardo-fiab-test-runner/

jenkins retest

I'm inclined to merge this with admin privileges if we can't get past this

rtitle:
jenkins retest

PR #2987, opened 2022-11-07
DDO-2377: create leonardo-tests in terra-github-workflows

ichengchang:

This PR provides a GHA pipeline for

Building leonardo

- Tag the build with semvar
- Publish the leonardo image to GCR
- ON DEV MERGE: push semver to dev
  - Addl steps to stop jenkins from pushing to DEV required here.
- Spin up a BEE env
- Run leonardo swat tests
- Report Test Results
- Tear down BEE env

See [DDO-2377](https://broadworkbench.atlassian.net/browse/DDO-2377) related to IRWG pipeline pod.

Update: now reports test results on failure:
https://github.com/broadinstitute/terra-github-workflows/actions/runs/6424097850/job/17444001870

---

In all cases:

[DDO-2377]: https://broadworkbench.atlassian.net/browse/DDO-2377?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lucymcnatt:
jenkins retest

lucymcnatt:
jenkins retest

.github/workflows/leo-build-tag-publish-and-run-tests.yml
```diff
@@ -0,0 +1,191 @@
...
+    permissions:
+      contents: 'read'
+      id-token: 'write'
+    outputs:
+      custom-version-json: ${{ steps.render-leo-version.outputs.custom-version-json }}
+      tag: ${{ steps.tag.outputs.tag }}
+    steps:
+      - uses: 'actions/checkout@v4'
+
+      - name: Bump the tag to a new version
```
LizBaldo: Maybe a naive question, but we will now have two sets of GHA, one for GCP and one for Azure that would be tagged the exact same way, would this cause issues?

LizBaldo (commented):
LGTM, the one thing that I am trying to understand is that if this job is supposed to just replace our current jenkins test, or also include the suite of Azure automation tests that currently runs in a separate GHA.

@lucymcnatt do you think you could help Justin in making sure that this GHA only runs the GCP suite? From this test failure it looks like this GHA was picking up the Azure tests: https://github.com/broadinstitute/terra-github-workflows/actions/runs/6424885848/job/17446468922

It would be great to see a fully green pipeline on GHA that we know should pass

.github/workflows/leo-build-tag-publish-and-run-tests.yml
jyang-broad: *looks around*
Oh, hey I didn't realize the azure test were running their own builds. Thanks for the heads up! My thoughts here is that this will build on dev merge so it'll always be up to date, I can update the azure tests to use the version on dev, which should be the latest available version.

jyang-broad:
endgame is that these GHA tests run exactly what runs in Jenkins, so if it's currently running Azure tests, that's unintentional.

Right now: 
 - GHA tests not are mandatory to pass since I'm not sure how stable they are and don't have the expertise to troubleshoot tests w/o IA help. So I'd like to just get the tests running to start collecting data and getting eyes on it.
   - This will increase noise in the short term.
 - This mean Jenkins tests will still run
 - Once GHA tests are stable enough to replace jenkins, we can do a cutover for the tests.
 - This does mean in the meantime, we are running 2x tests.

jyang-broad:
This one is (once again) ready for review... changes from last time:
azure-automation-tests:
- leo build has been removed an changed moved explicitly to leo-build-tag-publish-and-run-tests.
  - instead the azure cron will run against current `dev` leo app and `dev` automation tests. This shouldn't change anything in practice.
- the input for `branch` has been broken up into `automation-branch` and `app-version`. This is BREAKING for any current workflow dispatch automation, but I expect that number to be zero so it shouldn't matter. This does not break cron runs as it doesn't use input.
leo-build-tag-publish-job:
- build has been broken up into, `app-version`, `build-branch`, and `automation-branch`. Even though these are all the same string, this has been done to clarify that these a completely different concepts that should be kept distinct.
  - app-version refers to the version of the leo app build artifact that is to be deployed.
  - automation-branch is the leo repo branch to check out and run the `automation` suite, in the case that someone is devving tests against a current leonardo.
  - build-branch refers to the leo branch to check out to make a new build, this only gets used in the build section 

Testing: Azure and swat tests have been run and
- run but don't pass - this is expected(?)
- report stats correctly

Overall: Added more comments and line breaks for readability

Not in this PR but maybe future:
- break up the leonardo build vs test. Technically possible to make dependent flows, but will take more time

jdcanas:
jenkins retest

jyang-broad (approved):
last test "failed" but jcanas confirms both issues are unrelated and known issues.

approving.

https://github.com/broadinstitute/terra-github-workflows/actions/runs/7480952354/job/20361509384#step:9:22589
https://github.com/broadinstitute/terra-github-workflows/actions/runs/7480952354/job/20361509384#step:9:24216

PR #2996, opened 2022-11-10
IA-3814: Implement Pact.IO into Leo and write Sam consumer contract test 

ichengchang:

This PR sets up a project structure within Leo and include the dependencies needed in order to get `pact` running to do contract testing.

We need to add pact into Leonardo. In order to do this, we must first make sure the dependencies can be properly set. 

We would want pact to only run tests relevant to the provider. We have chosen to use

- Sam (Provider) 
- Leo (Consumer) model, 

As a start, this PR provides implementation of the following provider (Sam) endpoints for consumer (Leo) tests:

- GET /status
- GET /api/resources/v2/$resourceType

We would want it to live close to the code here: [leonardo/HttpSamDAO.scala at develop · DataBiosphere/leonardo](https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpSamDAO.scala#L39) Since that is where all the sam code is used.

Furthermore, we want to start writing contract tests against Sam here.

We will be using `pact4s` due to 

- support of pact V3 
- compatibility with libraries used by Leo

Please see the README file under the `pact4s` project folder on how to launch `pact` tests in Scala.

**Note: there's no contract tests written for Leonardo APIs in this PR**

---

In all cases:


Qi77Qi:
thanks @ichengchang ....it'd be nice to remove jenkins tests run in this PR...

And before we make the switch, I'd like us solving the logging problem first

araina2018:
@Qi77Qi I believe 

> 

@Qi77Qi ,There is no jenkins run on pact tests.  I think you are confusing this with PR for switch to GHA where we are still running all the swat tests and BEEs need to support logging. @ichengchang on that note, do we know what's the DevOps status on BEE logging. But we can track that in https://broadworkbench.atlassian.net/browse/DDO-2377

pact4s/src/main/scala/org/broadinstitute/dede/workbench/leonardo/consumer/FakeClient.scala
```diff
@@ -0,0 +1,68 @@
...
+import cats.syntax.all.none
+import io.circe.syntax.EncoderOps
+import io.circe.{Decoder, Encoder, Json}
+import org.http4s.Credentials.Token
+import org.http4s.circe.{jsonEncoderOf, jsonOf}
+import org.http4s.client.Client
+import org.http4s.headers.Authorization
+import org.http4s.{EntityDecoder, EntityEncoder, Method, Request, Status, Uri}
+
+trait FakeClient[F[_]] {
```
mspector: would we be better off with just one client? why have a `FakeClient` alongside a `SamClient`?

pact4s/src/test/scala/org/broadinstitute/dede/workbench/leonardo/consumer/FakeClientSpec.scala
```diff
@@ -0,0 +1,126 @@
...
+      .uponReceiving("Request to fetch extant resource")
+      .method("GET")
+      .path(s"/resource/$testID")
+      .headers("Authorization" -> mockBearerHeader(MockSamDAO.petSA))
+      .willRespondWith()
+      .status(200)
+      .body(
+        Json.obj("id" -> testID.asJson, "value" -> 123.asJson)
+      ) // can use circe json directly for both request and response bodies with `import pact4s.circe.implicits._`
+      .`given`("resource does not exist")
```
mspector: it looks like you're defining multiple states here within the same chain of definitions. clearly this works, but I wonder if it'd be more readable if each state were separate `PactDSLResponse` objects?

then again, I'm no expert, so perhaps this is the best approach. just some food for thought.

pact4s/src/test/scala/org/broadinstitute/dede/workbench/leonardo/consumer/SamClientSpec.scala
```diff
@@ -0,0 +1,213 @@
...
+  /*
+    we can define the folder that the pact contracts get written to upon completion of this test suite.
+   */
+  override val pactTestExecutionContext: PactTestExecutionContext =
+    new PactTestExecutionContext(
+      "./target/pacts"
+    )
+
+  // Uncomment this so that mock server will run on specific port (e.g. 9003) instead of dynamically generated port.
+  // override val mockProviderConfig: MockProviderConfig = MockProviderConfig.httpConfig("localhost", 9003)
```
mspector, line 40: just curious why this is left commented-out. at first glance, it seems worthwhile to set the port explicitly

pact4s/src/test/scala/org/broadinstitute/dede/workbench/leonardo/consumer/SamClientSpec.scala
```diff
@@ -0,0 +1,213 @@
...
+
+  // ---- Dsl for specifying pacts between consumer and provider
+  // Lambda Dsl: required for generating matching rules.
+  // Favored over old-style Pact Dsl.
+  val okSystemStatusDsl: DslPart = newJsonBody { o =>
+    o.booleanType("ok", true)
+    o.`object`("systems",
+               s =>
+                 for (subsystem <- subsystems)
+                   s.`object`(subsystem.value, o => o.booleanType("ok", true))
```
mspector, line 119: I'm new enough at this that I may be mistaken - but are you requiring that each subsystem be present in the response from Sam? if so, I would just suggest that you make sure that Leonardo really does require that all subsystems be present in the response, otherwise the contract might be overly restrictive

pact4s/src/test/scala/org/broadinstitute/dede/workbench/leonardo/consumer/SamClientSpec.scala
```diff
@@ -0,0 +1,213 @@
...
+    o.`object`("public",
+               s => {
+                 s.array("actions", _ => Set())
+                 s.array("roles", _ => Set())
+               }
+    )
+  }.build()
+
+  // Old-style Pact Dsl.
+  // Provided here for reference only.
```
mspector: I could be convinced otherwise, but I might suggest that these tests be treated as "real" tests, not reference implementations -- and so this section should perhaps be removed

pact4s/src/test/scala/org/broadinstitute/dede/workbench/leonardo/consumer/SamClientSpec.scala
```diff
@@ -0,0 +1,213 @@
...
+      .`given`("status is healthy")
+      .uponReceiving("Request to status")
+      .method("GET")
+      .path("/status")
+      .headers("Accept" -> "application/json")
+      .willRespondWith()
+      .status(200)
+      .headers("Content-type" -> "application/json")
+      .body(okSystemStatusDsl)
+      // -------------------------- FETCH WORKSPACE RESOURCES with the caller's roles and actions for each resource. --------------------------
```
mspector: same question here as above -- is it more readable to create a handful of `PactDslResponse` objects that each represent a state? or do you need to chain them all together in a single one?

mspector (commented):
The tests themselves look pretty good -- I would just encourage you to think of this as a first attempt at a real contract test suite, rather than an extended demo. E.g. I would recommend getting rid of the `FakeClient`, and bundle its functionality into the `SamClient` (unless there's something I'm misunderstanding about the setup, of course).

I note that the `SamClient` is new -- how does Leonardo typically communicate with Sam?

Qi77Qi:
this looks to me is for testing Sam API, not Leo API? How will this help Leo's independent release?

pact4s/src/test/scala/org/broadinstitute/dede/workbench/leonardo/consumer/SamClientSpec.scala
ichengchang: Without setting the port explicitly, the client will pick a random port that's available. This becomes handy when the consumer has pacts with multiple service providers. We don't have to assign specific ports to the mock services.

pact4s/src/main/scala/org/broadinstitute/dede/workbench/leonardo/consumer/FakeClient.scala
ichengchang: The `pact4s` library is currently limited to one provider per test spec (_cf._ [issue](https://github.com/jbwheatley/pact4s/issues/118)), it's on their radar to address this shortcoming.

`FakeClient` was intended to be a pattern on dealing with a different provider unrelated to `Sam`. Still I think it's a good practice to separate unrelated pacts into different test specs and limit ourselves to logically related pacts within same spec when that becomes available.

pact4s/src/test/scala/org/broadinstitute/dede/workbench/leonardo/consumer/FakeClientSpec.scala
ichengchang: Totally agreed with making the code more readable. To this end, I added a `PactHelper` object with `buildInteraction` functions to separate each state into its own `PactDSLResponse` object.

pact4s/src/test/scala/org/broadinstitute/dede/workbench/leonardo/consumer/SamClientSpec.scala
ichengchang: I have added more helper methods to support variants for building interactions. We can consolidate the overloaded helpers in a future ticket.

pact4s/src/test/scala/org/broadinstitute/dede/workbench/leonardo/consumer/SamClientSpec.scala
ichengchang: Removed the commented section.

ichengchang:
> this looks to me is for testing Sam API, not Leo API? How will this help Leo's independent release?

The point here is that `Leo` plays the role of a consumer and the service it consumes is `Sam`. The tests written in `pact4s` are used to forge a contractual obligation between `Leo` and `Sam`. It is to ensure that any change in `Sam` does not break what `Leo` expects.

Qi77Qi:
>The point here is that Leo plays the role of a consumer and the service it consumes is Sam. The tests written in pact4s are used to forge a contractual obligation between Leo and Sam. It is to ensure that any change in Sam does not break what Leo expects.

Isn't that still validating Sam change, instead of Leo? I'm not saying this is useless..Just want to point out, I don't think this helps Leo api test coverage

msilva-broad:
> 



> > The point here is that Leo plays the role of a consumer and the service it consumes is Sam. The tests written in pact4s are used to forge a contractual obligation between Leo and Sam. It is to ensure that any change in Sam does not break what Leo expects.
> 
> Isn't that still validating Sam change, instead of Leo? I'm not saying this is useless..Just want to point out, I don't think this helps Leo api test coverage

It's validating (and sending to the Pact.io contract broker) Leo's *expecations" of how the SAM API works according to how Leo is consuming it.  This helps SAM release independently by letting it block deployments if it would break Leo's expectations.  In turn, this lets Leo unit test it's own code without having to call SAM for real since it can now trust that SAM would never deploy something that breaks Leo's expectations without resolving that with the team first.

This pattern is then expected to be repeated for consumers of Leo service/s, giving Leo the same benefit of gated deployment based on api consumer expectations brokered to Pact.io and verifiable by Leo.  Collectively this is what enables independent release, because it frees all parties from having to run heavy integration tests with real running service dependencies.  The Pact contract broker establishes trust that each service can focus primarily on testing itself in isolation.

araina2018:
Thanks Miguel. Perhaps as a separate exercise, what will answer Qi's question is demonstrating an example of Leo contract test where Leo is the provider?

msilva-broad:
> Thanks Miguel. Perhaps as a separate exercise, what will answer Qi's question is demonstrating an example of Leo contract test where Leo is the provider?

absolutely... continuing the "chain" would immediately demonstrate the need for the Leo api consumer expectations to be populated in order for Leo to run it's own canideploy command.  This demonstrates the shared investment and payoff... though at this stage of the pilot I think we are being careful to keep overall scope in check.

pact4s/src/test/scala/org/broadinstitute/dede/workbench/leonardo/consumer/SamClientSpec.scala
ichengchang: I'll discuss this in today's walkthrough.  Here I'm using the following code snippet in [HttpSamDAOSpec](https://github.com/DataBiosphere/leonardo/blob/develop/http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpSamDAOSpec.scala#L47-L67) as a source of truth for establishing the pact. There are also other must less restrictive mocks such as [this](https://github.com/DataBiosphere/leonardo/blob/develop/http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpSamDAOSpec.scala#L90-L98) that could be used. So I guess this is something the Leo team will have to decide when writing consumer tests. My hunch is to start with the least restrictive one and gradually increase the level of restrictions to meet the contract goals.

pact4s/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+Otherwise, you can run the command inside a docker container with OpenJDK 17 installed. 
+```
+docker run --rm -v $PWD:/working \
+                -v jar-cache:/root/.ivy \
+                -v jar-cache:/root/.ivy2 \
+                -w /working \
+                sbtscala/scala-sbt:openjdk-17.0.2_1.7.2_2.13.10 \
+                sbt "project pact4s" clean test
+```
```
Qi77Qi, line 46: I think we can assume most ppl have java 17 installed...so probably just say `sbt "project pact4s" clean test`?

pact4s/src/main/scala/org/broadinstitute/dede/workbench/leonardo/consumer/FakeClient.scala
```diff
@@ -0,0 +1,68 @@
...
+  def fetchResource(id: String)(implicit decoder: Decoder[Resource]): F[Option[Resource]]
+
+  def createResource(resource: Resource): F[Unit]
+}
+
+/*
+ This class represents the consumer view of a fake provider that implements the following endpoints:
```
Qi77Qi: I kinda prefer just remove this fake client

project/Dependencies.scala
```diff
@@ -126,10 +128,17 @@ object Dependencies {
   val googleCloudNio: ModuleID =  "com.google.cloud"    % "google-cloud-nio"      % "0.124.21" % Test // brought in for FakeStorageInterpreter
 
   val circeYaml =         "io.circe"          %% "circe-yaml"           % "0.14.2"
+  val circeCore =         "io.circe"          %% "circe-core"           % "0.14.3"
```
Qi77Qi: this is brought in transitively. we don't need to specify here

project/Dependencies.scala
```diff
@@ -126,10 +128,17 @@ object Dependencies {
...
   val http4sBlazeServer = "org.http4s"        %% "http4s-blaze-server"  % http4sVersion
   val http4sPrometheus = "org.http4s" %% "http4s-prometheus-metrics" % http4sVersion
   val http4sDsl =         "org.http4s"        %% "http4s-dsl"           % http4sVersion
+  val http4sEmberClient = "org.http4s"        %% "http4s-ember-client"  % http4sVersion
+  val http4sEmberServer = "org.http4s"        %% "http4s-ember-server"  % http4sVersion
+  val http4sCirce       = "org.http4s"        %% "http4s-circe"  % http4sVersion
   val guava: ModuleID =   "com.google.guava"  % "guava"                 % guavaV
+  val typelevelCat =      "org.typelevel"     %% "munit-cats-effect-3"  % munitCatsEffectV % Test
```
Qi77Qi: can we just use scalatest instead of bringing yet another testing library?

project/Dependencies.scala
```diff
@@ -126,10 +128,17 @@ object Dependencies {
   val googleCloudNio: ModuleID =  "com.google.cloud"    % "google-cloud-nio"      % "0.124.21" % Test // brought in for FakeStorageInterpreter
 
   val circeYaml =         "io.circe"          %% "circe-yaml"           % "0.14.2"
+  val circeCore =         "io.circe"          %% "circe-core"           % "0.14.3"
   val http4sBlazeServer = "org.http4s"        %% "http4s-blaze-server"  % http4sVersion
   val http4sPrometheus = "org.http4s" %% "http4s-prometheus-metrics" % http4sVersion
   val http4sDsl =         "org.http4s"        %% "http4s-dsl"           % http4sVersion
+  val http4sEmberClient = "org.http4s"        %% "http4s-ember-client"  % http4sVersion
```
Qi77Qi, line 134: can we use the default http client instead of ember client?

pact4s/README.md
ichengchang: The `docker` will still be necessary if the consumer tests are being kicked off in a CI/CD pipeline such as GitHub Action. The runner is not guaranteed to possess the right version of OpenJDK. I have updated the README to state this fact.

pact4s/src/main/scala/org/broadinstitute/dede/workbench/leonardo/consumer/FakeClient.scala
ichengchang: The fake client has been removed.

project/Dependencies.scala
ichengchang: You're correct, the `circe-core` will be brought in transitively. Removed.

project/Dependencies.scala
ichengchang: Removed munit test library.

project/Dependencies.scala
ichengchang: By default client, I am assuming you mean the client used here [HttpDockerDAOSpec.scala#L68](https://github.com/DataBiosphere/leonardo/blob/develop/http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpDockerDAOSpec.scala#L68). I have updated the code accordingly. If that's not what you meant please let me know. The reason I use Ember instead of Blaze is because the latter has been deprecated.

Qi77Qi (commented):
thumbing assuming this is still WIP and things will be iterated after Kai finishes Sam work.

but I'll leave approving to @mspector since he's more knowledgeable about pact

knapii-developments (approved):

mspector (approved):
Looks good to me. I agree that changes are likely after we get the Sam provider-side tests set up, and also as we learn more as a group.

jack-r-warren:
jenkins retest

Qi77Qi:
jenkins retest

Qi77Qi:
```
[info] org.broadinstitute.dsde.workbench.leonardo.notebooks.NotebookGATKSpec *** ABORTED *** (11 minutes, 7 seconds)
[info]   java.lang.RuntimeException: terra-quality-6c305663/automation-test-ar6lnr6xz errored due to List(RuntimeError(Jupyter, Welder failed to start after 10 minutes.,None,2022-12-15T17:04:29Z,Some(TraceId(aae892c2e6a240edb8aef1ab9224b7d8/3939911508519804487))))
```

jenkins retest

PR #2997, opened 2022-11-10
ID-347 Remove Martha and use DRSHub in tests

andy7i:
What:
Adds configuration options for Leonardo to use drshub instead of martha in unit tests.
https://broadworkbench.atlassian.net/browse/ID-347

Why:
So we can test switching over to using DrShub instead of Martha.

How:
Updating unit tests.

---

In all cases:


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/Config.scala
```diff
@@ -607,6 +607,10 @@ object Config {
   implicit private val galaxyDrsUrlReader: ValueReader[GalaxyDrsUrl] = stringValueReader.map(GalaxyDrsUrl)
 
   implicit private val galaxyAppConfigReader: ValueReader[GalaxyAppConfig] = ValueReader.relative { config =>
+    val drsUrl = config.as[String]("drsResolver") match {
+      case "martha" => config.as[GalaxyDrsUrl]("marthaUrl")
+      case "drshub" => config.as[GalaxyDrsUrl]("drshubUrl")
```
Qi77Qi: any reason we can't switch to drshub yet?

andy7i: good question. @tlangs do you know the answer to this?

andy7i: the ticket mentions "Just configuration options that aren't utilized yet"

tlangs: Yeah. We haven't performance-tested DRSHub yet, and making things talk to it without first doing that seems unwise. The plan is to test it sequentially in all pre-prod envs to make sure nothing breaks, then performance test it  to make sure it can at least match Martha, then push to prod.

http/src/main/resources/reference.conf
```diff
@@ -513,7 +513,9 @@ gke {
     # Templated by firecloud-develop
     postgres.password = "replace-me"
     orchUrl = "https://firecloud-orchestration.dsde-dev.broadinstitute.org/api/"
-    drsUrl = "https://us-central1-broad-dsde-dev.cloudfunctions.net/martha_v3"
+    drsResolver = "martha"
+    marthaUrl = "https://us-central1-broad-dsde-dev.cloudfunctions.net/martha_v3"
+    drshubUrl = "https://drshub.dsde-dev.broadinstitute.org/api/"
```
tlangs: This needs to be `https://drshub.dsde-dev.broadinstitute.org/api/v4/drs/resolve`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/Config.scala
tlangs: Yeah. We want to make sure DRSHub can match Martha's performance before going to prod. The plan is to configure all services to be able to talk to DRSHub, then flip the switches in non-prod envs sequentially and see what happens. We'll want to perf test it before going to prod as well. So for now, things should continue using Martha.

tlangs (changes requested):

tlangs (approved):

Qi77Qi (commented):
is related firecloud-develop PR merged?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/Config.scala
Qi77Qi: I probably would've just used different value for different envs instead of introducing 2 config values.

So basically just adding `drshubUrl`, and the value for it will vary depending on the env, which will be a firecloud-develop PR

http/src/main/resources/reference.conf
```diff
@@ -513,7 +513,9 @@ gke {
     # Templated by firecloud-develop
     postgres.password = "replace-me"
     orchUrl = "https://firecloud-orchestration.dsde-dev.broadinstitute.org/api/"
-    drsUrl = "https://us-central1-broad-dsde-dev.cloudfunctions.net/martha_v3"
+    drsResolver = "martha"
+    marthaUrl = "https://us-central1-broad-dsde-dev.cloudfunctions.net/martha_v3"
+    drshubUrl = "https://drshub.dsde-dev.broadinstitute.org/api/v4/drs/resolve"
```
rtitle: Noting that this URL actually gets sent to Galaxy, and Galaxy makes DRS calls from python. I think this is the code: https://github.com/anvilproject/fs.anvilfs

Does the DRSHub API have the same shape as martha? E.g. same auth, params, etc. If so then it should be safe, but ideally we'd test with Galaxy before cutting over.

rtitle (commented):
This looks fine to me but I defer to @Qi77Qi. :) I agree fc-develop needs to merge first because this introduces a new environmental config value. I left a comment on the fc-develop PR about BEE-ificiation.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/Config.scala
tlangs: So, I'm hearing you'd rather have the rendered environment determine the `drsUrl`, rather than doing it in code?

Qi77Qi: yea...no big deal...they do the same thing...just feels the current approach is slightly more complicated than necessary

When later we do the full switch, it'd be easier to just change the config value in firecloud-develop too instead of having to change both leo and firecloud-develop

http/src/main/resources/reference.conf
Qi77Qi: @tlangs @andy7i can u comment on this..I thikn this is the most important thing to confirm besides making sure `fireloud-develop` is merged before this one

andy7i:
As noted in the PR description, I'll wait to merge this PR until after the fc-dev PR has merged.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/Config.scala
andy7i: Going with this approach, I'm realizing that the unit tests can either test Martha or DRSHub but not both.  I'm wondering if we should leave the Martha code or if we want to switch over tests to DRSHub now. I think could keep the Martha code until we're ready to release the DRSHub code. No strong preference.

http/src/main/resources/reference.conf
andy7i: based on the documentation, it looks very similar. 
https://drshub.dsde-dev.broadinstitute.org/#/drsHub/resolveDrs
https://github.com/broadinstitute/martha#martha-v3

@tlangs do you know off the top of your head whether the two APIs are functionally equivalent (for example, parameters, response, etc.)?

andy7i:
jenkins retest - startup failed

```
10:51:38 3:51PM ERR 1 execution errors:
10:51:38 jobmanager: Command "ARGOCD_SERVER=ap-argocd.dsp-devops.broadinstitute.org ARGOCD_AUTH_TOKEN=****** argocd --header Proxy-Authorization: Bearer ****** --grpc-web app get jobmanager-jenkins-swat-leo-5446 -o yaml" exited with status 20:
10:51:38 time="2022-11-17T15:48:08Z" level=fatal msg="rpc error: code = Unknown desc = Post \"[https://ap-argocd.dsp-devops.broadinstitute.org:443/application.ApplicationService/Get\](https://ap-argocd.dsp-devops.broadinstitute.org/application.ApplicationService/Get/)": dial tcp: lookup ap-argocd.dsp-devops.broadinstitute.org on 169.254.169.254:53: read udp 172.17.0.2:44840->169.254.169.254:53: i/o timeout"
```
https://fc-jenkins.dsp-techops.broadinstitute.org/job/fiab-start/87003/console

README.md
```diff
@@ -37,6 +37,10 @@ Ensure docker is running. Spin up MySQL locally:
 ```
 $ ./docker/run-mysql.sh start leonardo  
 ```
+If you have an M1 laptop, pull the x86_64 mysql image:
+```
+docker pull --platform linux/x86_64 mysql:5.6
```
tlangs: if you're running M1, wouldn't you want the `linux/arm64` mysql image?

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/config/ConfigSpec.scala
```diff
@@ -115,7 +115,7 @@ final class ConfigSpec extends AnyFlatSpec with Matchers {
       true,
       DbPassword("replace-me"),
       GalaxyOrchUrl("https://firecloud-orchestration.dsde-dev.broadinstitute.org/api/"),
-      GalaxyDrsUrl("https://us-central1-broad-dsde-dev.cloudfunctions.net/martha_v3"),
+      GalaxyDrsUrl("https://drshub.dsde-dev.broadinstitute.org/api/v4/drs/resolve"),
```
tlangs, line 118: If we do this now, Leo tests will talk to DRSHub, but deployed Leo will talk to Martha. Do we want that?

andy7i: yeah, I also had that question: https://github.com/DataBiosphere/leonardo/pull/2997#discussion_r1024532347

README.md
andy7i: running m1, we need to force pull the linux/x86_64 image because the linux/arm64 one does not exist for 5.6
Screen Shot 2022-11-17 at 12 23 27 PM

I'll make that more explicit!

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/config/ConfigSpec.scala
andy7i: also, do you know the answer to this question? https://github.com/DataBiosphere/leonardo/pull/2997#discussion_r1024537436

README.md
andy7i: I guess the other way to solve this is to add the docker pull command to https://github.com/DataBiosphere/leonardo/blob/develop/docker/run-mysql.sh

Maybe I'll pull this out into a separate PR.

andy7i: pulling this into a new PR https://github.com/DataBiosphere/leonardo/pull/3023

andy7i:
jenkins retest
```
11:21:25 4:21PM ERR 500 Internal Server Error from https://firecloudorch.jenkins-swat-leo-5476.bee.envs-terra.bio/register/profile ({"causes":[],"message":"The http server closed the connection unexpectedly before delivering responses for 1 outstanding requests","source":"FireCloud","stackTrace":[]})
```
https://fc-jenkins.dsp-techops.broadinstitute.org/job/fiab-start/87054/console

andy7i:
jenkins retest - different failure each time. unstable tests?

gpolumbo-broad:
When we worked with Justin on Thursday to do the independisam release, he was commenting that the Leo tests were having all sorts of problems.  Not sure what they were or if they're ongoing or if they affect only the main branch or what.  But I know Leo tests were struggling.

Qi77Qi:
@gpolumbo-broad if you're referring to Justin Yang, I think he's probably talking about when running in github workflow, which has its own problems...Regarding tests, there isn't known consistent failures...what I've seen myself is lots of fiab-start errors

project/Settings.scala
```diff
@@ -41,7 +41,7 @@ object Settings {
     "-language:implicitConversions",     // Allow definition of implicit functions called views
     "-unchecked",                        // Enable additional warnings where generated code depends on assumptions.
     "-Xcheckinit",                       // Wrap field accessors to throw an exception on uninitialized access.
-    "-Xfatal-warnings",                  // Fail the compilation if there are any warnings.
+//    "-Xfatal-warnings",                  // Fail the compilation if there are any warnings.
```
Qi77Qi: this needs to be reverted

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/Config.scala
Qi77Qi: for just tests, just use whatever that's in reference.conf

http/src/main/resources/reference.conf
tlangs: They should be equivalent, as DRSHub is a direct port of Martha.

Qi77Qi: cool..once this is deployed..can we also make sure things are working as expected in Galaxy on dev?

cc @afgane

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/Config.scala
andy7i: just to clarify, you're saying we should keep the old values (referencing Martha), correct?

If so, I'll leave this PR open until the cutover to DRSHub.

http/src/main/resources/reference.conf
andy7i: yeah, how would we do that? Would these be manual tests / are they written down somewhere?

Qi77Qi: I don't actually know very well how it's being used either...was hoping @afgane could provide some more info

andy7i: ok this PR is more of a cleanup PR so I'm going to put this under [ID-347](https://broadworkbench.atlassian.net/browse/ID-347) instead. 

@afgane please let us know how to test galaxy either here, slack (#dsp-identiteam) or on https://broadworkbench.atlassian.net/browse/ID-313

rtitle:
Just checking back on this PR, seems it's not doing much anymore -- just changing the default `drsUrl` in reference.conf which is overridden anyway by firecloud-develop. 

So I'm happy we landed on a simpler solution to just configure the DRS URL environmentally -- am OK with merging or closing this PR.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/Config.scala
Qi77Qi: what I'm saying is for tests, you can just update it to whatever value you change to in `reference.conf`...like here in this PR, you updated to drs url, so just use that in test too...

The real value is controlled by config value in firecloud-develop, so whether or not we want to merge this PR is up-to-you...doesn't really matter whether we merge now or later as long as `firecloud-develop` has correct value for each env

andy7i: oh gotcha, thanks for clarifying! I'll go ahead and merge this then to reduce code hanging around

andy7i:
@Qi77Qi do you mind reviewing this PR in the next week or two?

Qi77Qi:
I thumbed your merging comment https://github.com/DataBiosphere/leonardo/pull/2997#discussion_r1030552008  ...guess I  could explicitly approve if u want

Qi77Qi (approved):

andy7i:
> I thumbed your merging comment [#2997 (comment)](https://github.com/DataBiosphere/leonardo/pull/2997#discussion_r1030552008) ...guess I could explicitly approve if u want

Thank you!

PR #3013, opened 2022-11-16
[TOAZ-233] Update VM use lz resources

jordansafer:
Draft changes for adding landing zone resources for VMs

---

In all cases:


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpWsmDao.scala
```diff
@@ -153,6 +156,78 @@ class HttpWsmDao[F[_]](httpClient: Client[F], config: HttpWsmDaoConfig)(implicit
       )(onError)
     } yield res
 
+  override def getLandingZoneResources(billingProfileId: String, userToken: Authorization)(implicit
```
jordansafer, line 166: I don't like having this in the WsmDao, because it belongs a layer above WSM. Right now it's here so AppInterp and RuntimeV2Interp can share it. This also means the other LZ APIs (getLandingZone, listLandingZoneResourcesByType) are no longer called outside of WSM, and I could make them private if we go with this.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala
```diff
@@ -82,13 +82,7 @@ class RuntimeV2ServiceInterp[F[_]: Parallel](config: RuntimeServiceConfig,
...
 
       leoAuth <- samDAO.getLeoAuthToken
-      relayNamespaceOpt <- wsmDao.getRelayNamespace(workspaceId, req.region, leoAuth)
-      relayNamespace <- F.fromOption(
-        relayNamespaceOpt,
-        BadRequestException(s"Workspace ${workspaceId} doesn't have relay namespace provisioned appropriately",
-                            Some(ctx.traceId)
-        )
-      )
+      // TODO/PR question: do we still need this RelayNamespace for GCP?
```
jordansafer, line 84: I see we have targeted GCP support in a few places in the design. These changes shouldn't prevent future GCP support, but they bias in the direction of GCP also providing resources via landingZone. TODO check on how we create GCP VMs today-

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala
```diff
@@ -157,8 +160,8 @@ class RuntimeV2ServiceInterp[F[_]: Parallel](config: RuntimeServiceConfig,
             _ <- publisherQueue.offer(
               CreateAzureRuntimeMessage(savedRuntime.id,
```
jordansafer, line 162: Since we are explicitly specifying this as a Create**Azure**RuntimeMessage, I'm in favor of throwing an error in the GCP case, and making landingZoneResources required instead of optional

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/MonitorAtBoot.scala
```diff
@@ -331,22 +332,24 @@ class MonitorAtBoot[F[_]](publisherQueue: Queue[F, LeoPubsubMessage],
...
           )
-          rns <- F.fromOption(relayNamespaceOpt,
-                              MonitorAtBootException(s"no relay namespace found for ${wid}", traceId)
-          )
+          workspaceDescOpt <- wsmDao.getWorkspace(wid, leoAuth)
+          workspaceDesc <- F.fromOption(workspaceDescOpt, WorkspaceNotFoundException(wid, traceId))
+
+
+          // Get the Landing Zone Resources for the app for Azure
+          landingZoneResourcesOpt <- wsmDao.getLandingZoneResources(workspaceDesc.spendProfile, leoAuth)
```
jordansafer: I'm not familiar with this code path, and am not sure the leoAuth token has permission to request landingZone resources (the above wsm calls are both for wsm managed resources). TODO to check on this-

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpWsmDao.scala
rtitle: Good point that this code needs to be shared. Another pattern we've done is put things in companion objects, example: https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeServiceInterp.scala#L851

This seems _okay_ to put in `WsmDAO` though, since it's just calling a couple WSM APIs and parsing the responses.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala
rtitle: Yeah, the fact that it uses Relay makes it Azure-specific (Relay is an Azure product). Ideally the V2 endpoints would support GCP and Azure, but require a WSM-powered workspace. So there is maybe some tech debt here, but for now I think we can always use the Relay namespace from the Landing Zone.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala
rtitle: That seems reasonable to me to not increase scope too much, though I still think eventually this method should support GCP.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/MonitorAtBoot.scala
rtitle: It probably doesn't! :) You could maybe use https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/SamDAO.scala#L105 to call the LZ API

rtitle: This reminds me that I don't think we touched `MonitorAtBoot` for the app stuff -- this code tries to replay messages if Back Leo restarts while it is monitoring something. I can look at this..

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpWsmDao.scala
Qi77Qi: I'm ok with having it in WsmDao for now...Making the other 2 methods private sound good to me!

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/WsmDao.scala
```diff
@@ -86,6 +86,10 @@ trait WsmDao[F[_]] {
     ev: Ask[F, AppContext]
   ): F[Option[WorkspaceDescription]]
 
+  def getLandingZoneResources(billingProfileId: String, userToken: Authorization)(implicit
```
rtitle, line 88: It'd be nice to remove methods in WsmDAO that we aren't using anymore, like `createNetwork`.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
```diff
@@ -101,8 +105,9 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
         case x: CloudContext.Azure => F.pure(x.value)
       }
       hcName = RelayHybridConnectionName(params.runtime.runtimeName.asString)
-      primaryKey <- azureRelay.createRelayHybridConnection(params.relayeNamespace, hcName, cloudContext)
+      primaryKey <- azureRelay.createRelayHybridConnection(params.landingZoneResources.relayNamespace, hcName, cloudContext)
       createDiskAction = createDisk(params, auth)
+      // TODO PR question: can we use one of the LZ subnets for this? Possibly the compute subnet?
```
rtitle: Yes! We shouldn't explicitly create a network anymore, we should use the compute subnet I think.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpWsmDao.scala
```diff
@@ -153,6 +156,78 @@ class HttpWsmDao[F[_]](httpClient: Client[F], config: HttpWsmDaoConfig)(implicit
...
+      .flatMap(_.deployedResources)
+      .filter(_.resourceType.equalsIgnoreCase(resourceType))
+      .headOption
+      .flatMap { r =>
+        if (useParent) r.resourceParentId.flatMap(_.split('/').lastOption)
+        else r.resourceName.orElse(r.resourceId.flatMap(_.split('/').lastOption))
+      }
+      .fold(
+        F.raiseError[String](
+          AppCreationException(s"${resourceType} resource with purpose ${purpose} not found in landing zone")
```
Qi77Qi, line 256: the error probably should be updated?

What's a LZ resource name btw? can u give an example if u have it handy

rtitle (commented):
Answered questions, I think you're on the right track

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala
```diff
@@ -157,8 +160,8 @@ class RuntimeV2ServiceInterp[F[_]: Parallel](config: RuntimeServiceConfig,
...
                                         workspaceId,
-                                        relayNamespace,
                                         storageContainer.resourceId,
+                                        landingZoneResourcesOpt,
```
Qi77Qi: shall we just pass relayNamespace from the LZ directly here? it seems the change will be a lot smaller that way...given back leo doesn't really need other info from the LZ either

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala
jordansafer: Oh! Makes sense, thank you :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/MonitorAtBoot.scala
jordansafer: For the Sam API, I'm still skeptical since we had the previous issue with apps where the pet access token wasn't available. I'm a little confused about how leoAuth works in general, I'll reach out to chat about it-

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpWsmDao.scala
jordansafer: Yes, the names are autogenerated by the landing zone service based on the azure resource requirements (different resources get names of different lengths). Here's an example from logs of some names:
```
{
  "clusterName":"lz87b5acf237a7394344358a9",
  "batchAccountName":"lze5ac32cc7822076fc571a8",
  "relayNamespace":"lz1aad27142d64b2999b8ceae00ada4b3c9d2bf58a098ca336",
  "storageAccountName":"lz9b7e56e0d2d94c6031edc2",
   "vnetName":"lzf43fb0d47b61bd524490895ca52933b304f3ceec72736d76220753ef17bc8d",
  "batchNodesSubnetName":"lzf43fb0d47b61bd524490895ca52933b304f3ceec72736d76220753ef17bc8d",
  "aksSubnetName":"lzf43fb0d47b61bd524490895ca52933b304f3ceec72736d76220753ef17bc8d"
}
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala
jordansafer: i was thinking this too, it's a lot to pass everything... but now I saw from rob we do need the vnet and compute subnet too, so I'm leaning towards sending the whole object-

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
```diff
@@ -101,8 +105,9 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
         case x: CloudContext.Azure => F.pure(x.value)
       }
       hcName = RelayHybridConnectionName(params.runtime.runtimeName.asString)
-      primaryKey <- azureRelay.createRelayHybridConnection(params.relayeNamespace, hcName, cloudContext)
+      primaryKey <- azureRelay.createRelayHybridConnection(params.landingZoneResources.relayNamespace, hcName, cloudContext)
```
Qi77Qi: I thought hybrid connection is also created somewhere now? or maybe I'm misremembering things?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala
Qi77Qi: sounds good...

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
rtitle: We talked about making it a WSM resource but it never actually happened.. so for now Leo is still creating it directly (which I think is fine).

Qi77Qi: yea that's what i was thinking....fwiw, I think it's totally fine to be managed by Leo

Qi77Qi:
btw, I think we also need to update https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala#L92 to get the workspace storage container from LZ now

Qi77Qi:
wanna correct my previous comment. I think it's https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpWsmDao.scala#L291 needs to be updated

jordansafer:
- still testing in BEEs

jordansafer:
Testing succeeded on BEE!

jordansafer:
Jenkins retest

jordansafer:
`[org.broadinstitute.dsde.workbench.leonardo.runtimes.RuntimeDataprocSpec.should stop/start a Dataproc cluster with workers and preemptible workers](https://fc-jenkins.dsp-techops.broadinstitute.org/job/leonardo-fiab-test-runner/53562/testReport/junit/org.broadinstitute.dsde.workbench.leonardo.runtimes/RuntimeDataprocSpec/should_stop_start_a_Dataproc_cluster_with_workers_and_preemptible_workers/)`
Doesn't appear to be related
-  Jenkins retest

jordansafer:
Jenkins retest

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala
rtitle: Can probably remove this comment now

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
```diff
@@ -101,8 +101,13 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
...
       }
       hcName = RelayHybridConnectionName(params.runtime.runtimeName.asString)
-      primaryKey <- azureRelay.createRelayHybridConnection(params.relayeNamespace, hcName, cloudContext)
+      primaryKey <- azureRelay.createRelayHybridConnection(params.landingZoneResources.relayNamespace,
+                                                           hcName,
+                                                           cloudContext
+      )
       createDiskAction = createDisk(params, auth)
+      // TODO make a ticket. We're still calling WSM createVM API, which depends on a WSM owned network. We're migrating
+      // to the shared LZ subnet. We could create the new VM directly from Leo or make an updated API in WSM
```
rtitle, line 109: Hm - I had thought the `networkId` field in the WSM `createVm` payload was optional, and it would use the LZ network if not present. Is this not working? I can check too..

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/WsmDao.scala
rtitle: Would still be good to do this, don't think we should be using the ip or relayNamespace methods anymore?

rtitle (approved):
Went over this again, LGTM if tested and working.

I am curious why using the LZ network/subnet for VM creation isn't working, might try this myself in a BEE..

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
jordansafer: The problem is that the Hybrid Connection is currently created by Leo using the old Azure Relay/network. So what needs to happen there is:
1. I need to close out the changes for the createHybridConnection API in WSM!
1. Leo can migrate to calling the API and using the LZ relay namespace.

jordansafer:
Jenkins retest (BPM disk mounting error)

jordansafer:
Jenkins retest

jordansafer:
Jenkins retest

rtitle:
I deployed this branch to a BEE and manually made a VM and AKS app :) I'm going to admin-merge since fiab-start is broken currently.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpWsmDao.scala
```diff
@@ -153,7 +163,101 @@ class HttpWsmDao[F[_]](httpClient: Client[F], config: HttpWsmDaoConfig)(implicit
...
+                                                       "Microsoft.Storage/storageAccounts",
+                                                       SHARED_RESOURCE,
+                                                       false
+      )
+      postgresName <- getLandingZoneResourceName(groupedLzResources,
+                                                 "microsoft.dbforpostgresql/servers",
+                                                 SHARED_RESOURCE,
+                                                 false
+      )
+      logAnalyticsWorkspaceName <- getLandingZoneResourceName(groupedLzResources,
```
Qi77Qi, line 209: @jordansafer is there a way to get all the resources in one call...

PR #3014, opened 2022-11-16
[TOAZ-279] Fix app unique constraint; DDL update

rtitle:
https://broadworkbench.atlassian.net/browse/TOAZ-279

Fixing a couple things I noticed on dev.

1. `APP` table unique constraint was on `(appName, nodepoolId)`. This meant we couldn't delete & recreate Azure apps with the same name. Normally we include `destroyedDate` in the index to allow this. (Note APP already has a non-null `destroyedDate` column so I think this should be fine.)

2. I had renamed the values in [this enum](https://github.com/DataBiosphere/leonardo/blob/develop/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala#L253-L277) but there are existing DB records with the old values -- so it was throwing a deserialization exception when I tried to list my existing apps in the DB. Added a DDL update to fix existing records.

---

In all cases:


Qi77Qi (approved):
thanks! :shipit:

PR #2983, opened 2022-11-07
Add platform to mysql docker run commands

mikenev:
Add the --platform params to the docker run command, otherwise M1 macs error out starting up the image.

---
Have you read [CONTRIBUTING.md](https://github.com/DataBiosphere/leonardo/blob/develop/CONTRIBUTING.md) lately? If not, do that first.

I, the developer opening this PR, do solemnly pinky swear that:

- [x] I've documented my API changes in Swagger

In all cases:

- [ ] Get a thumbsworth of review and PO signoff if necessary
- [ ] Verify all tests go green (you can re-run automation tests with a comment saying `jenkins retest`
- [ ] Run the automation tests multiple times in parallel to weed out instability if applicable via a comment saying `multi-test`
- [ ] Squash and merge; Delete your branch after this
- [ ] Test this change deployed correctly and works on dev environment after deployment

PR #3023, opened 2022-11-17
ID-322 fix mysql issue for testing locally on m1 laptops

andy7i:

running m1, we need to force pull the linux/x86_64 image because the linux/arm64 one does not exist for 5.6 and results in an error.

Screen Shot 2022-11-17 at 12 23 27 PM

pulled this out from https://github.com/DataBiosphere/leonardo/pull/2997/files#r1025481349

---

In all cases:


tlangs (approved):

docker/run-mysql.sh
```diff
@@ -7,6 +7,10 @@ start() {
     echo "attempting to remove old $CONTAINER container..."
     docker rm -f $CONTAINER || echo "docker rm failed. $CONTAINER not found."
 
+    # this is a necessary step for m1 laptops since a linux/arm64 image does not exist for mysql:5.6
+    echo "pulling mysql docker image"
+    docker pull --platform linux/x86_64 mysql:5.6
```
Qi77Qi, line 12: would this break for non m1 laptop cuz it'll always download this image and use it for non m1 laptop too?

andy7i: this command pulls the non-m1 image, so it will not affect non-m1 laptops. (There is no m1 / arm64 image which was causing issues on m1 laptops. So that's why this command pulls the x86_64 image.)

Also, if the image already exists then the image is not downloaded again:
```
pulling mysql docker image
5.6: Pulling from library/mysql
Digest: sha256:20575ecebe6216036d25dab5903808211f1e9ba63dc7825ac20cb975e34cfcae
Status: Image is up to date for mysql:5.6
docker.io/library/mysql:5.6
starting up mysql container...
```

andy7i:
jenkins retest - different failure each time. unstable tests?

andy7i:
@Qi77Qi do you mind reviewing this PR in the next week or so?

docker/run-mysql.sh
Qi77Qi: so this image is for non-m1 laptop? how does this fix things for m1 laptop then?

also I tagged u in https://github.com/DataBiosphere/leonardo/pull/2983#issuecomment-1319119036, which I was hoping for some clarification :)

andy7i: I think m1 laptops can run x86_64 images using an emulator (Rosetta) but not the other way around.

andy7i: so for intel machines it doesn't make a difference but for m1 laptops, it provides an image that 1) exists, and 2) can be run on m1 machines.

I thumbed #2983 because that seems to be a more elegant solution than this one

PR #3024, opened 2022-11-17
TOAZ-280: move workspaceId from KUBERNETES_CLUSTER to APP table

rtitle:
https://broadworkbench.atlassian.net/browse/TOAZ-280

* Schema change: moved nullable `workspaceId` column from `KUBERNETES_CLUSTER` to `APP` table. This is more correct because Apps are per-workspace entities; which clusters may be shared across workspaces (on Azure).
   * This was causing bugs where apps were being created/listed in the wrong workspaces, if there were multiple workspaces in a landing zone. 
* Cleaned up some code/tests around `KubernetesServiceDbQueries`, see inline comments

I tested manually on a BEE by creating 2 apps in 2 workspaces -- I confirmed the get/list app endpoints work as expected now.

---

In all cases:


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/AppComponent.scala
```diff
@@ -151,7 +155,7 @@ object appQuery extends TableQuery(new AppTable(_)) {
           )(DBIO.successful(_))
         )
 
-      // here, we enforce uniqueness on (AppName, GoogleProject) for active apps
+      // here, we enforce uniqueness on (AppName, CloudContext) for active apps
```
rtitle: This logic means that Azure app names must be unique within an entire MRG (not just unique with a workspace). I'm not actually sure if this a requirement in our AKS deployment, but keeping the logic in place anyway.

1 reason I don't like our app data model: these kinds of constraints aren't expressible in the DB and thus enforced in code.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesClusterComponent.scala
```diff
@@ -163,11 +157,11 @@ object kubernetesClusterQuery extends TableQuery(KubernetesClusterTable(_)) {
     ).map(_.headOption)
 
   // this retrieves the nodepool and namespaces associated with a cluster
-  def getMinimalActiveClusterByName(
+  def getMinimalActiveClusterByCloudContext(
```
rtitle, line 160: Renamed method based on what it does

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesServiceDbQueries.scala
```diff
@@ -1,26 +1,33 @@
...
 import org.broadinstitute.dsde.workbench.model.TraceId
 
+import java.time.Instant
 import scala.concurrent.ExecutionContext
 
 object KubernetesServiceDbQueries {
-  // the 'full' here means that all joins possible are done, meaning all fields in the cluster, nodepool, and app will be present
-  // if you just need the cluster, nodepools, and cluster-wide namespaces, see the minimal join in the KubernetesClusterComponent
+
+  /**
```
rtitle, line 23: I tried to comment the public methods here, and make clear which should be used for v1 apps and v2 apps.

In general v2 apps should query based on `workspace`, and v1 should query based on `cloudContext`.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesServiceDbQueries.scala
```diff
@@ -31,20 +38,26 @@ object KubernetesServiceDbQueries {
...
+      kubernetesClusterQuery,
       nodepoolQuery,
-      if (includeDeleted) appQuery else nonDeletedAppQuery,
+      filterByWorkspaceId(if (includeDeleted) appQuery else nonDeletedAppQuery, workspaceId),
       labelFilter
     )
 
-  // Called by MonitorAtBoot to determine which apps need monitoring
+  /**
+   * List all apps that need monitoring. Called by MonitorAtBoot.
```
rtitle, line 59: TODO for myself: look at `MonitorAtBoot` code, make sure it works for v2 apps

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesServiceDbQueries.scala
```diff
@@ -177,21 +204,6 @@ object KubernetesServiceDbQueries {
       }
     } yield validatedApp
 
-  // if there are no apps associated with this cluster, None will be returned even if the cluster exists.
```
rtitle, line 180: These methods weren't used anywhere so I deleted them

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
```diff
@@ -758,10 +760,7 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
   private[service] def getSavableCluster(
     userEmail: WorkbenchEmail,
     cloudContext: CloudContext,
-    now: Instant,
-    numNodepools: Option[NumNodepools],
```
rtitle, line 762: `numNodepools` and `clusterName` were never passed to this method, so removed them.

(I think this was a holdover to when we have a batch-nodepool-creation endpoint, which was removed.)

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/db/AppServiceDbQueriesSpec.scala
```diff
@@ -1,442 +0,0 @@
-package org.broadinstitute.dsde.workbench.leonardo
```
rtitle, line 1: I moved all tests in this file to `KubernetesServiceDbQueriesSpec`, since that is the class being tested (there is no class named `AppServiceDbQueries`)

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/db/AppServiceDbQueriesSpec.scala
```diff
@@ -1,442 +0,0 @@
...
-  }
-
-  it should "error if cluster exists and is creating on save" in isolatedDbTest {
-    val makeCluster1 = makeKubeCluster(1).copy(status = KubernetesClusterStatus.Provisioning)
-    val makeCluster2 = makeKubeCluster(2).copy(status = KubernetesClusterStatus.Precreating)
-
-    makeCluster1.save()
-    makeCluster2.save()
-
-    val saveCluster1 = Some(makeCluster1)
```
rtitle, line 254: The pattern `Some(a).map(f).get` pattern was repeated a few places which made me chuckle/ 🤦

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/AppComponent.scala
Qi77Qi: this was probably fine for v1 route...maybe we should use sth different for v2? so that the uniqueness is at workspace level?

rtitle: I thought about that -- the k8s deployments do name things like `{{ appName }}-something` so I wasn't sure if there would be name conflicts if we allow duplicate appNames within a cluster. Technically everything is scoped to the namespace in k8s so maybe it's ok. But was thinking this is safer (and the UI generates unique appNames anyway)..

Qi77Qi: it just gets harder to reason why different level of uniqueness is being enforced...especially AoU reuses names (altho I talked to Yonghao and they'll potentially start using unique names too)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesServiceDbQueries.scala
rtitle: Looking at `MonitorAtBoot` code, it looks like it explicitly doesn't handle Azure, which is good (better than trying to publish the wrong message). Thinking I make a separate ticket/PR for `MonitorAtBoot` logic.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/AppComponent.scala
rtitle: Yeah it is a gotcha.. I can try changing v2 apps to be unique by (name, workspace) and see what happens

rtitle: [done](https://github.com/DataBiosphere/leonardo/pull/3024/commits/3c83fe867044e4d52c280c3060ded19a21800529)

rtitle: I successfully created 2 apps with the same name in different workspaces 🎉

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesServiceDbQueries.scala
rtitle: MonitorAtBoot: https://broadworkbench.atlassian.net/browse/TOAZ-281

rtitle:

rtitle:
Automation tests have failed a few times in this PR, with no test failures but red anyway:

https://fc-jenkins.dsp-techops.broadinstitute.org/job/leonardo-fiab-test-runner/53224/

Smells a little fishy, will investigate..

rtitle:
jenkins retest

Qi77Qi (approved):

PR #3038, opened 2022-11-22
[IA-3710][IA-3874] Renamed Azure to JupyterLab from runtimeModels.

n8landolt:
CLOSED: Had another branch that was cleaner

https://github.com/DataBiosphere/leonardo/pull/3046


Qi77Qi:
@n8landolt  which api and which field we're trying to refactor? I'd start from there instead of trying to refactor any model classes (things will get really messy)

PR #3047, opened 2022-11-29
[IA-3716] Fix v2 list

jdcanas:
So the issue with runtimes showing up due via creator filtering is partially an issue with the V2 list endpoint. The list function in v2 doesn't perform the fallback properly to handle runtimes without

Proof in curl below (the auth token is expired, note the fact its the same request, but the v1 endpoint returns more runtimes than the v2). This makes a UI running against local leo work properly for v1 runtimes. 

**Still need to test the creator filtering for runtimes created with v2 endpoint.** Want to socialize this fix first though.

image

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala
@@ -339,14 +339,19 @@ class RuntimeV2ServiceInterp[F[_]: Parallel](config: RuntimeServiceConfig,
 
       (runtimesUserIsCreator, runtimesUserIsNotCreator) = runtimes.partition(_.auditInfo.creator == userInfo.userEmail)
 
+      // Here, we optimize the SAM lookups based on whether we will need to use the google project fallback
+      // IF (workspaceId) EXISTS we need WSM resource ID sam lookup + workspaceId Fallback
+      // ELSE we need googleProject fallback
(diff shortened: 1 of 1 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala
```diff
@@ -352,10 +354,21 @@ class RuntimeV2ServiceInterp[F[_]: Parallel](config: RuntimeServiceConfig,
         .map(_.getOrElse(List.empty))
 
       // We must also check the RuntimeSamResourceId in sam to support already existing and newly created google runtimes
-      samVisibleRuntimeSamResourceIds <- NonEmptyList
-        .fromList(runtimesUserIsNotCreator.map(_.samResource))
-        .traverse(ids => authProvider.filterUserVisible(ids, userInfo))
-        .map(_.getOrElse(List.empty))
+      runtimesAndProjects = runtimesUserIsNotCreator.map(r => (GoogleProject(r.cloudContext.asString), r.samResource))
```
Qi77Qi: `(GoogleProject(r.cloudContext.asString)` doesn't seem right cuz you the cloud context could be either GCP or Azure

jdcanas: This is the fallback for runtimes created with the v1 endpoint. the `GoogleProject` cast is for compatibility with the `authProvider`

Qi77Qi: this is not fallback for just runtimes created via v1 api tho..runtimes created via v2 will show up here too

Qi77Qi:
closing this PR since it's superceded by https://github.com/DataBiosphere/leonardo/pull/3061

PR #3059, opened 2022-12-06
Minor change to avoid confusion in API Docs for CreateAppV2 endpoint

aaronkanzer:
Bumps example in `CreateAppV2` endpoint to be more appropriate for current use cases.

See https://broadinstitute.slack.com/archives/C02GJ782BAA/p1670361183668239?thread_ts=1670350497.658939&cid=C02GJ782BAA as an example

---

In all cases:


jdcanas (approved):
Looks good, likely was copy pasted. Thanks! It might fail automation tests for unrelated reasons, in which case I'll force-push it for you...

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -2550,9 +2550,7 @@ components:
           schema:
             $ref: "#/components/schemas/CreateAppRequest"
           example:
-            diskConfig:
```
Qi77Qi, line 2553: I think we're reusing the same example for both v1 and v2 api...you'll need to create a separate model for v2 api if you'd like to update the example

Qi77Qi (changes requested):
I think we're reusing the same example for both v1 and v2 api...you'll need to create a separate model for v2 api if you'd like to update the example

jdcanas:
> I think we're reusing the same example for both v1 and v2 api...you'll need to create a separate model for v2 api if you'd like to update the example

Ah yeah, you'll need to create a copy of this https://github.com/DataBiosphere/leonardo/blob/167242c2d1f7467ee358fc65b0d1629c0666abcd/http/src/main/resources/swagger/api-docs.yaml#L3554

aaronkanzer:
@jdcanas @Qi77Qi Updated here! Let me know when you have the chance to re-review -- thanks!

Qi77Qi (approved):

Qi77Qi:
thanks @aaronkanzer ....sry abt the delay

PR #3062, opened 2022-12-07
TOAZ-281 Implement Azure support in MonitorAtBoot.

Ghost-in-a-Jar:
This pr adds the implementation for MonitorAtBoot for azure apps. This included adding a codepath for createappv2message and deleteappv2message in the MonitorAtBoot.scala class. I also added some tests to the MonitorAtBootSpec file. 

Will test out the functionality in a bee before merging.

---

In all cases:


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/MonitorAtBoot.scala
```diff
@@ -351,6 +458,101 @@ class MonitorAtBoot[F[_]](publisherQueue: Queue[F, LeoPubsubMessage],
...
+      .flatMap(_.headOption)
+      .flatMap { r =>
+        if (useParent) r.resourceParentId.flatMap(_.split('/').lastOption)
+        else r.resourceName.orElse(r.resourceId.flatMap(_.split('/').lastOption))
+      }
+      .fold(
+        F.raiseError[String](
+          AppCreationException(s"${resourceType} resource with purpose ${purpose} not found in landing zone")
+        )
+      )(F.pure)
```
Ghost-in-a-Jar: These two fns are taken from app service, maybe they should be in a trait?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/MonitorAtBoot.scala
```diff
@@ -157,61 +177,148 @@ class MonitorAtBoot[F[_]](publisherQueue: Queue[F, LeoPubsubMessage],
...
+                for {
+                  tokenOpt <- samDAO.getCachedArbitraryPetAccessToken(app.auditInfo.creator)
+                  userToken <- F.fromOption(
+                    tokenOpt,
+                    MonitorAtBootException(
+                      s"Unable to get pet token for ${app.auditInfo.creator} for app ${app.id} in Provisioning status",
+                      appContext.traceId
+                    )
+                  )
+                  workspaceId: WorkspaceId = app.workspaceId.getOrElse(
```
rtitle: Can we use `F.fromOption` here (and in places where we need to throw exceptions)?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/MonitorAtBoot.scala
```diff
@@ -157,61 +177,148 @@ class MonitorAtBoot[F[_]](publisherQueue: Queue[F, LeoPubsubMessage],
...
+                    org.http4s.Credentials.Token(AuthScheme.Bearer, userToken)
+                  )
+                  workspaceDescOpt <- wsmDao.getWorkspace(workspaceId, bearerAuth)
+                  workspaceDesc <- F.fromOption(workspaceDescOpt,
+                                                WorkspaceNotFoundException(workspaceId, appContext.traceId)
+                  )
+
+                  landingZoneResources <- getLandingZoneResources(workspaceDesc.spendProfile, bearerAuth)
+                  storageContainer <- wsmDao.getWorkspaceStorageContainer(workspaceId, bearerAuth)
+                  workspaceId = app.workspaceId.getOrElse(
```
rtitle: Another `F.fromOption`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/MonitorAtBoot.scala
```diff
@@ -351,6 +458,101 @@ class MonitorAtBoot[F[_]](publisherQueue: Queue[F, LeoPubsubMessage],
         )
       case x => F.raiseError(MonitorAtBootException(s"Unexpected status for runtime ${runtime.id}: ${x}", traceId))
     }
+
+  private def getLandingZoneResources(billingProfileId: String, userToken: Authorization)(implicit
```
rtitle: I don't really like duplicating this code -- can we put this method in a shared location?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/MonitorAtBoot.scala
rtitle: Yes - let's not duplicate this code

rtitle:
Logic LGTM, main comment is around sharing code between `MonitorAtBoot` and `AppServiceInterp`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/MonitorAtBoot.scala
Ghost-in-a-Jar: sounds good yeah

jordansafer (approved):
Looks good! I think we're starting to repeat the same pattern a lot for dealing with SAM/pet credentials for Azure, so a minor thought in the back of my head is that if we keep having this we should refactor a little down the road.

jordansafer:
Looks like scalafmt strikes again

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/MonitorAtBoot.scala
```diff
@@ -57,7 +62,9 @@ class MonitorAtBoot[F[_]](publisherQueue: Queue[F, LeoPubsubMessage],
       .parEvalMapUnordered(10) { case (a, n, c) =>
         for {
           now <- F.realTimeInstant
-          implicit0(traceId: Ask[F, TraceId]) = Ask.const[F, TraceId](TraceId(s"BootMonitor${now}"))
+          implicit0(traceId: Ask[F, AppContext]) = Ask.const[F, AppContext](
```
Qi77Qi, line 65: any reason why we need `AppContext` here?

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/MonitorAtBootSpec.scala
```diff
@@ -242,6 +335,22 @@ class MonitorAtBootSpec extends AnyFlatSpec with TestComponent with LeonardoTest
     res.unsafeRunSync()(cats.effect.unsafe.IORuntime.global)
   }
 
+  it should "ignore non-monitored apps in Azure" in isolatedDbTest {
```
Qi77Qi, line 338: what's non-monitored apps?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/MonitorAtBoot.scala
Ghost-in-a-Jar: Yeah to string it through to handleApp, there is a downstream fn that requires an implicit AppContext

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/MonitorAtBootSpec.scala
Ghost-in-a-Jar: any app that doesnt have the status Prvisioning or Deleting when leo starts up

rtitle (approved):
I re-reviewed this and it looks good. Tests failed due to known Google groups issues. Going to merge this.

PR #3064, opened 2022-12-08
IA-3885/ adding optional timeout to CreateRuntimeRequest

LizBaldo:
Added an optional `timeoutInMinutes` argument to the `CreateRuntimeRequest` endpoint. The API is expecting the custom time to be an integer representing the number of minutes the user is willing to wait for a Runtime to be created before calling it quit and erroring out.

This change ended up impacting quite a lot of files and I will leave some breadcrumb comments to point towards where the logic actually is. 

https://broadworkbench.atlassian.net/browse/IA-3885

When testing locally, I am able to error out after only a couple of minutes:

request body:
```
{
  "runtimeConfig": {
    "cloudService": "gce",
    "machineType": "n1-standard-4",
    "diskSize": 100
  },
  "toolDockerImage": "us.gcr.io/broad-dsp-gcr-public/terra-jupyter-gatk:2.0.2",
  "timeoutInMinutes": 2
}
```

status after a couple of minutes:
```
"status": "Error",
...
"errors": [
    {
      "errorMessage": "Jupyter, Welder failed to start after 2 minutes.",
      "errorCode": null,
      "timestamp": "2022-12-13T21:18:25Z",
      "traceId": "817ef6e0-559d-412b-8b1d-08f5a5af038a"
    }
  ],
```

---

In all cases:


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/RuntimeRoutes.scala
```diff
@@ -436,6 +436,7 @@ object RuntimeRoutes {
       wr <- c.downField("welderRegistry").as[Option[ContainerRegistry]]
       s <- c.downField("scopes").as[Option[Set[String]]]
       cv <- c.downField("customEnvironmentVariables").as[Option[LabelMap]]
+      tm <- c.downField("timeoutInMinutes").as[Option[Int]]
```
LizBaldo, line 439: We are getting it as an Integer, which simplifies the message encoding / decoding immensely

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/RuntimeRoutes.scala
```diff
@@ -448,7 +449,8 @@ object RuntimeRoutes {
       tdi,
       wr,
       s.getOrElse(Set.empty),
-      cv.getOrElse(Map.empty)
+      cv.getOrElse(Map.empty),
+      tm.map(x => x minutes)
```
LizBaldo, line 453: The integer then gets transformed into an optional FiniteDuration type

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BaseCloudServiceRuntimeMonitor.scala
```diff
@@ -493,12 +504,17 @@ abstract class BaseCloudServiceRuntimeMonitor[F[_]] {
             ).map(b => (imageType, b))
           )
       }
-      // wait for 10 minutes for tools to start up before time out.
+      // wait for tools to start up before time out. Use the timeout value if available
```
LizBaldo: This is really where the meat of this work is. If the timeout is not None then use that, else use the default

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/leoPubsubMessageSubscriberModels.scala
```diff
@@ -217,7 +219,8 @@ object LeoPubsubMessage {
         runtime.welderEnabled,
         runtime.customEnvironmentVariables,
         runtimeConfig,
-        traceId
+        traceId,
+        timeoutInMinutes.map(x => x.toMinutes.toInt)
```
LizBaldo: The pubsub messages only handle integers, not FiniteDuration

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/MonitorAtBoot.scala
```diff
@@ -37,7 +37,7 @@ class MonitorAtBoot[F[_]](publisherQueue: Queue[F, LeoPubsubMessage],
         for {
           now <- F.realTimeInstant
           implicit0(traceId: Ask[F, TraceId]) = Ask.const[F, TraceId](TraceId(s"BootMonitor${now}"))
-          _ <- handleRuntime(r)
+          _ <- handleRuntime(r, None) // Should we pass in the custom timeout here, and if so how?
```
LizBaldo, line 44: @Qi77Qi FYI this is a weird blocker that I am unable to answer myself. How should I pass the timeout argument in this function?

Qi77Qi: let's leave it out for now...this class is for recovering cases when runtimes are left in transit status after Leo is restarted.

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -3036,6 +3036,10 @@ components:
         defaultClientId:
           type: string
           description: The default Google Client ID.
+        timeoutInMinutes:
+          type: integer
+          description: Custom image deployment time out in minutes.
+            If nothing id provided, a default of 10 minutes will be used.
```
Qi77Qi: one caveat of this is there’s this defines a timeout for the entire create runtime flow…so if user’s input is more than 30 min, it won’t work…I don’t think we need to do anything intelligently update this overall timeout yet..but maybe mention that currently, this timeout can't be more than 30 min? if in the future, user requests to increase this overall timeout, we can do that

Qi77Qi (approved):
one minor comment..otherwise lgtm 🎉 

thumbing once tests pass

http/src/main/resources/swagger/api-docs.yaml
LizBaldo: Ohhh, I was not aware of that 30 minutes timeout! I will definitely update the documentation to reflect that, thanks! @n8landolt it should likely be reflected in the UI (e.g. a validation that makes sure you are getting an integer between 1 and 30)

LizBaldo: @Qi77Qi I also remember that you would have preferred to name this variable differently, would you like me to change it as well?

Qi77Qi: yea...I'm struggling with names too...maybe `checkToolsInterruptAfter`? altho, I think in UI, we should just say "timeout" so that it's meaningful to them

LizBaldo: Maybe let's have a mini huddle with @n8landolt today and decide on the naming for both the ui and leo?

Qi77Qi: just corrected my last comment (think I had an extra n't)...

LizBaldo: Ahah indeed I get it now! `timeoutInMinutes` on the UI side, but `checkToolsInterruptAfter` in Leo past the endpoint :) FYI @n8landolt

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BaseCloudServiceRuntimeMonitor.scala
```diff
@@ -528,12 +539,18 @@ abstract class BaseCloudServiceRuntimeMonitor[F[_]] {
...
+        case Some(duration) => duration
+        case None =>
+          monitorConfig.checkTools.interruptAfter
+      }
       availableTools <- streamFUntilDone(
         checkTools,
         monitorConfig.checkTools.maxAttempts,
         monitorConfig.checkTools.interval
-      ).interruptAfter(monitorConfig.checkTools.interruptAfter).compile.lastOrError
+      ).interruptAfter(timeout).compile.lastOrError
```
n8landolt: Would the default config interval and max attempts terminate at the default 10 minute timeout even if we set the interruptAfter?

Qi77Qi: ah yea, that too...so I think we need to adjust the max attempts and interval as well if user defines timeout

LizBaldo: Gotcha, thanks for the catch @n8landolt 
I will implement what you had previously
https://github.com/DataBiosphere/leonardo/blob/5150e7dc0d7dda5f525a2e09fe3b12afd95da822/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BaseCloudServiceRuntimeMonitor.scala#L474

I will also modify the corresponding unit test

LizBaldo:
jenkins retest

n8landolt (approved):
Approval pending fix on the max attempts/interval.

LizBaldo:
I ended up modifying the interval using the custom time out instead of modifying the max number of attempts. it felt cleaner to avoid rounding to the nearest integer.
I also added a small unit test to make sure that the interval is modified the way we want. Everything seems to be working as expected. The only issue preventing me from merging are the jenkins automation tests 🤷‍♀️

LizBaldo:
multi-test

LizBaldo:
jenkins retest

LizBaldo:
jenkins retest

LizBaldo:
jenkins retest

LizBaldo:
jenkins retest

LizBaldo:
jenkins retest

LizBaldo:
jenkins retest

LizBaldo:
jenkins retest

LizBaldo:
fiab-start failure

jenkins retest

LizBaldo:
fiab-start failure

jenkins retest

LizBaldo:
jenkins retest

LizBaldo:
jenkins retest

LizBaldo:
jenkins retest

Qi77Qi:
since tests are quite flaky these days and we've seen a successful run for the PR, I'm going to merge this as is

PR #3061, opened 2022-12-07
IA-3888 Optimize listRuntimeV2 performance

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-3888

* Support `creator=role` query string for listRuntimeV2 api
* Update DB query to filter based on an optional `creatorOnly` field for better DB query performance
* Update underlying query to [plain SQL](https://scala-slick.org/doc/3.0.0/sql.html#splicing-literal-values) for better performance (more than 1 min to a few seconds).
* Removed workspace permission check fallback. After talking to @rtitle , it turns out Sam model created by WSM is hierarchical, so by asking sam if a user can read a runtime is sufficient (Sam will take care of checking if the caller has permission at workspace level as well)

Test
1. Run Leo locally against dev DB
2. Run the following cmd as `hermione`
```
curl 'https://local.dsde-dev.broadinstitute.org:30443/api/v2/runtimes?saturnAutoCreated=true&includeLabels=saturnWorkspaceNamespace%2CsaturnWorkspaceName' \
  -H 'Accept: */*' \
  -H 'Accept-Language: en-US,en;q=0.9' \
  -H "Authorization: Bearer `gcloud auth print-access-token`" \
  -H 'Cache-Control: no-cache' \
  -H 'Connection: keep-alive' \
  -H 'Origin: http://localhost:3000' \
  -H 'Pragma: no-cache' \
  -H 'Referer: http://localhost:3000/' \
  -H 'Sec-Fetch-Dest: empty' \
  -H 'Sec-Fetch-Mode: cors' \
  -H 'Sec-Fetch-Site: cross-site' \
  -H 'User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/108.0.0.0 Safari/537.36' \
  -H 'X-App-ID: Saturn' \
  -H 'sec-ch-ua: "Not?A_Brand";v="8", "Chromium";v="108", "Google Chrome";v="108"' \
  -H 'sec-ch-ua-mobile: ?0' \
  -H 'sec-ch-ua-platform: "macOS"' \
  --compressed
```
3. Before the optimization, the DB query will take more than 1 min to finish, and it'll cause curl to timeout. After the optimization, the DB query only takes about 2 seconds to finish. (I put log message before and after the DB query for debugging)

---
[...]

PR #3074, opened 2022-12-16
IA-3903 PROD-761 A few bug fixes and remove duplicate Sam ID for Azure runtime

Qi77Qi:
1. No longer save Azure Runtime's WSM resourceId in `RUNTIME_CONTROLLED_RESOURCE` because this resourceId is the same as runtime's Sam resource id, which already exists in `CLUSTER` table as`internalId`. This also simplifies things a bit in a few places where we can just use `internalId` as the runtime's Sam ID for checking permissions instead of having to read the value from DB separately.
2. Fix a bug where we're joining CLUSTER and RUNTIME_CONFIG table on the wrong key..Introduced in [this PR](https://github.com/DataBiosphere/leonardo/pull/3061)...It should've been caught by at least one unit test, but for some reason that test didn't fail in CI...I'll spend a bit more time looking into how to run `RuntimeServiceDbQueriesSpec` more reliably in CI
3. Fix a bug where we're checking `creator` role for Azure runtime when calling Sam API. But it should be `writer`.
4. Fix a bug where workspace owner can't see runtimes created other users in the worksapce

---

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/RuntimeServiceDbQueries.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/RuntimeServiceDbQueries.scala
@@ -262,11 +262,11 @@ object RuntimeServiceDbQueries {
     listRuntimesHelper(labelMap, includeDeleted, creatorOnly, workspaceId, cp)
   }
 
-  def listRuntimesHelper(labelMap: LabelMap,
-                         includeDeleted: Boolean,
-                         creatorOnly: Option[WorkbenchEmail],
-                         workspaceId: Option[WorkspaceId],
-                         cloudContextOrCloudProvider: Option[Either[CloudContext, CloudProvider]]
+  private def listRuntimesHelper(labelMap: LabelMap,
+                                 includeDeleted: Boolean,
+                                 creatorOnly: Option[WorkbenchEmail],
+                                 workspaceId: Option[WorkspaceId],
+                                 cloudContextOrCloudProvider: Option[Either[CloudContext, CloudProvider]]
   ): DBIO[Vector[ListRuntimeResponse2]] = {
     val runtimeQueryFilteredByCreator = creatorOnly match {
       case Some(creator) => List(s"C.`creator` = '${creator.value}'")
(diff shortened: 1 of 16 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/RuntimeServiceDbQueries.scala
```diff
@@ -403,7 +403,7 @@ object RuntimeServiceDbQueries {
               left join `LABEL` L on (L.`resourceId` = FILTERED_CLUSTER.id) and (L.`resourceType` = 'runtime') 
               #${labelMapFilters}
           ) AS LABEL_FILTERED 
-          inner join `RUNTIME_CONFIG` RG on LABEL_FILTERED.id = RG.`id` 
+          inner join `RUNTIME_CONFIG` RG on LABEL_FILTERED.runtimeConfigId = RG.`id`
```
Qi77Qi, line 406: this is the most important bug fix for this PR

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala
```diff
@@ -141,7 +139,7 @@ class RuntimeV2ServiceInterp[F[_]: Parallel](config: RuntimeServiceConfig,
               cloudContext,
               userInfo,
               req,
-              RuntimeSamResourceId(samResource.resourceId),
+              RuntimeSamResourceId(vmSamResourceId.toString),
```
Qi77Qi, line 142: we were previously saving the workspaceId as the runtime's sam id, which was not correct

jdcanas (approved):

Qi77Qi:
```
[info] org.broadinstitute.dsde.workbench.leonardo.notebooks.NotebookGCEDataSyncingSpec *** ABORTED *** (11 minutes, 30 seconds)
[info]   java.lang.RuntimeException: terra-quality-6fb9bcfa/automation-test-a3utqmv6z errored due to List(RuntimeError(Jupyter, Welder failed to start after 10 minutes.,None,2022-12-19T19:45:10Z,Some(TraceId(90507a1f466b4c7299cf79c9e187c8b9/3939911508519804487))))
```

jenkins retest

PR #3076, opened 2022-12-19
IA-3823 List disks allows filtering by creator

mlilynolting:
Unit tests pass for DiskServiceInterp (there is no spec for DiskServiceDbQueries; it would be reasonable to create one for this work but I have left this undone), and I've verified the endpoints work via local Swagger.

---

In all cases:

### DIFF
--- a/http/src/main/resources/swagger/api-docs.yaml
+++ b/http/src/main/resources/swagger/api-docs.yaml
@@ -861,14 +861,6 @@ paths:
           schema:
             type: string
         - in: query
-          name: creator
-          description: >
-            Optional filter that includes only persistent disks created by the given user. Accepts an email.
-            Note: currently only the requesting user's email is supported (making this a duplicate of "role=creator").
-          required: false
-          schema:
-            type: string
-        - in: query
           name: includeDeleted
           description: Optional filter that includes any persistent disks with a Deleted status.
           required: false
@@ -935,14 +927,6 @@ paths:
             Note: this string format is a workaround because Swagger doesn't support free-form
             query string parameters. The recommended way to use this endpoint is to specify the
             labels as top-level query string parameters. For instance: GET /api/google/v1/disks?key1=val1&key2=val2.
-          required: false
-          schema:
-            type: string
-        - in: query
-          name: creator
-          description: >
-            Optional filter that includes only persistent disks created by the given user. Accepts an email.
-            Note: currently only the requesting user's email is supported (making this a duplicate of "role=creator").
           required: false
           schema:
             type: string
--- a/http/src/main/resources/swagger/api-docs.yaml
+++ b/http/src/main/resources/swagger/api-docs.yaml
@@ -861,6 +861,14 @@ paths:
           schema:
             type: string
         - in: query
+          name: creator
+          description: >
+            Optional filter that includes only persistent disks created by the given user. Accepts an email.
+            Note: currently only the requesting user's email is supported (making this a duplicate of "role=creator").
+          required: false
+          schema:
+            type: string
+        - in: query
           name: includeDeleted
           description: Optional filter that includes any persistent disks with a Deleted status.
           required: false
@@ -873,6 +881,12 @@ paths:
             Optional label keys of the labels returned in response. Example:
             Querying by key1,key2,key3
             returns all labels key1/val1 key2/val2 and key3 and val3 for each persistent disk in response
+          required: false
+          schema:
+            type: string
+        - in: query
+          name: role
+          description: Optional filter that excludes persistent disks you did not create. Accepts "creator" or nothing.
           required: false
           schema:
             type: string
@@ -925,6 +939,14 @@ paths:
           schema:
             type: string
         - in: query
+          name: creator
+          description: >
+            Optional filter that includes only persistent disks created by the given user. Accepts an email.
+            Note: currently only the requesting user's email is supported (making this a duplicate of "role=creator").
+          required: false
+          schema:
+            type: string
+        - in: query
           name: includeDeleted
           description: Optional filter that includes any persistent disks with a Deleted status.
           required: false
@@ -937,6 +959,12 @@ paths:
             Optional label keys of the labels returned in response. Example:
             Querying by key1,key2,key3
             returns all labels key1/val1 key2/val2 and key3 and val3 for each persistent disk in response
+          required: false
+          schema:
+            type: string
+        - in: query
+          name: role
+          description: Optional filter that excludes persistent disks you did not create. Accepts "creator" or nothing.
           required: false
           schema:
             type: string
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskServiceInterp.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskServiceInterp.scala
@@ -192,8 +192,13 @@ class DiskServiceInterp[F[_]: Parallel](config: PersistentDiskConfig,
   ): F[Vector[ListPersistentDiskResponse]] =
     for {
       ctx <- as.ask
+      // TODO processListParameters may parse the key-value param "includeRole=creator" as a label
       paramMap <- F.fromEither(processListParameters(params))
-      disks <- DiskServiceDbQueries.listDisks(paramMap._1, paramMap._2, cloudContext).transaction
+      // TODO includeRole in processListParameters
+      // TODO where do we enforce the acceptable values for an API param? includeRole may only be 'creator' or nothing
+      creatorOnly = params.get(includeRoleKey).contains(SamRole.Creator.asString)
+      createdBy = if (creatorOnly) Some(userInfo.userEmail) else None
+      disks <- DiskServiceDbQueries.listDisks(paramMap._1, paramMap._2, createdBy, cloudContext).transaction
       _ <- ctx.span.traverse(s => F.delay(s.addAnnotation("Done DB call")))
       diskAndProjects = disks.map(d =>
         (GoogleProject(d.cloudContext.asString), d.samResource)
@@ -358,6 +363,7 @@ object DiskServiceInterp {
 
     for {
       // check the labels do not contain forbidden keys
+      // TODO also blacklist `includeRoleKey`
       labels <-
         if (allLabels.contains(includeDeletedKey))
           Left(IllegalLabelKeyException(includeDeletedKey))
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskServiceInterp.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskServiceInterp.scala
@@ -193,14 +193,7 @@ class DiskServiceInterp[F[_]: Parallel](config: PersistentDiskConfig,
     for {
       ctx <- as.ask
       paramMap <- F.fromEither(processListParameters(params))
-      creatorOnly =
-        // Support filtering by creator either by role=creator query string, or creator=<user email> label
-        if (
-          params
-            .get(creatorOnlyKey)
-            .contains(creatorOnlyValue) || params.get(creatorOnlyValue).exists(_ == userInfo.userEmail.toString)
-        ) Some(userInfo.userEmail)
-        else None
+      creatorOnly <- F.fromEither(processCreatorOnlyParameter(userInfo, params))
       disks <- DiskServiceDbQueries.listDisks(paramMap._1, paramMap._2, creatorOnly, cloudContext).transaction
       _ <- ctx.span.traverse(s => F.delay(s.addAnnotation("Done DB call")))
       diskAndProjects = disks.map(d =>
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskServiceInterp.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskServiceInterp.scala
@@ -192,13 +192,16 @@ class DiskServiceInterp[F[_]: Parallel](config: PersistentDiskConfig,
   ): F[Vector[ListPersistentDiskResponse]] =
     for {
       ctx <- as.ask
-      // TODO processListParameters may parse the key-value param "includeRole=creator" as a label
       paramMap <- F.fromEither(processListParameters(params))
-      // TODO includeRole in processListParameters
-      // TODO where do we enforce the acceptable values for an API param? includeRole may only be 'creator' or nothing
-      creatorOnly = params.get(includeRoleKey).contains(SamRole.Creator.asString)
-      createdBy = if (creatorOnly) Some(userInfo.userEmail) else None
-      disks <- DiskServiceDbQueries.listDisks(paramMap._1, paramMap._2, createdBy, cloudContext).transaction
+      creatorOnly =
+        // Support filtering by creator either by role=creator query string, or creator=<user email> label
+        if (
+          params
+            .get(creatorOnlyKey)
+            .contains(creatorOnlyValue) || params.get(creatorOnlyValue).exists(_ == userInfo.userEmail.toString)
+        ) Some(userInfo.userEmail)
+        else None
+      disks <- DiskServiceDbQueries.listDisks(paramMap._1, paramMap._2, creatorOnly, cloudContext).transaction
       _ <- ctx.span.traverse(s => F.delay(s.addAnnotation("Done DB call")))
       diskAndProjects = disks.map(d =>
         (GoogleProject(d.cloudContext.asString), d.samResource)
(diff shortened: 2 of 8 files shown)


http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -860,6 +860,14 @@ paths:
           required: false
           schema:
             type: string
+        - in: query
+          name: creator
```
Qi77Qi: key should `role`?

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -924,6 +938,14 @@ paths:
...
```
Qi77Qi: same

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskServiceInterpSpec.scala
```diff
@@ -449,6 +449,50 @@ class DiskServiceInterpSpec extends AnyFlatSpec with LeonardoTestSuite with Test
...
+                            0
+    ) // this email is white listed
+
+    val res = for {
+      disk1 <- makePersistentDisk(Some(DiskName("d1"))).save()
+      // Make second disk belonging to different user than the calling user
+      disk2 <- LeoLenses.diskToCreator
+        .set(WorkbenchEmail("a_different_user@example.com"))(makePersistentDisk(Some(DiskName("d2"))))
+        .save()
+      listResponse <- diskService.listDisks(userInfo, None, Map("role" -> "creator"))
```
Qi77Qi, line 488: can u also add a test that without either param, both disks should be returned

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskServiceInterp.scala
```diff
@@ -193,7 +193,15 @@ class DiskServiceInterp[F[_]: Parallel](config: PersistentDiskConfig,
     for {
       ctx <- as.ask
       paramMap <- F.fromEither(processListParameters(params))
-      disks <- DiskServiceDbQueries.listDisks(paramMap._1, paramMap._2, cloudContext).transaction
+      creatorOnly =
+        // Support filtering by creator either by role=creator query string, or creator=<user email> label
+        if (
+          params
```
Qi77Qi: can probably make this a utility function in `package.scala` so that it can be used in various places

sth like
```
def extractCreator(callerEmail: WorkbenchEmail, queryStrings: Map[String, String]): Option[WorkbenchEmail] =???
```

Qi77Qi (commented):
looks great in general 👍 a few minor comments

http/src/main/resources/swagger/api-docs.yaml
mlilynolting: This is to mirror your changes to the listRuntimes endpoints - role=creator and creator=me@my.email are both supported

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskServiceInterp.scala
mlilynolting: agreed! Should I extract your logic from the listRuntimes endpoints to the helper as well? Would it make sense to fold this into the logic of `processListParameters` in that file, since that currently handles parsing out `includeDeleted` and `includeLabels`?

http/src/main/resources/swagger/api-docs.yaml
Qi77Qi: oh I see..let's not mention this in doc? kinda wanna just nudge ppl to use `role=creator` anyway...we're only supporting it mostly becuz that's how terra UI is using the API today

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskServiceInterpSpec.scala
mlilynolting: Yes! Thanks for noting that negative case

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskServiceInterp.scala
Qi77Qi: >agreed! Should I extract your logic from the listRuntimes endpoints to the helper as well? 

sounds good...

>Would it make sense to fold this into the logic of processListParameters in that file, since that currently handles parsing out includeDeleted and includeLabels?

nah, I don't really like that function as is..let's not make it even more complicated

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskServiceInterpSpec.scala
n8landolt: Isn't that the test above?
"list disks belonging to other users"

n8landolt (approved):

http/src/main/resources/swagger/api-docs.yaml
mlilynolting: ahhh ok! will do

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskServiceInterpSpec.scala
mlilynolting: I added a new case showing a mixed set of disks (i created + others created) all get returned w/o the param

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/package.scala
```diff
@@ -82,4 +85,33 @@ package object service {
         )
       )
 
+  private[service] def processCreatorOnlyParameter(
+    userInfo: UserInfo,
+    params: LabelMap
+  ): Either[ParseCreatorOnlyException, Option[WorkbenchEmail]] =
+    // Support filtering by creator either by role=creator query string, or creator=<user email> label
```
LizBaldo: super small nit, but could you also add in the comment the preferred method and why to disseminate the context shared by you and Qi in the PR comments?

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskServiceInterpSpec.scala
```diff
@@ -449,6 +449,72 @@ class DiskServiceInterpSpec extends AnyFlatSpec with LeonardoTestSuite with Test
     res.unsafeRunSync()(cats.effect.unsafe.IORuntime.global)
   }
 
+  it should "list disks belonging to self and others, if not filtered by role=creator" in isolatedDbTest {
+    val (diskService, _) = makeDiskService()
+    val userInfo = UserInfo(OAuth2BearerToken(""),
+                            WorkbenchUserId("userId"),
+                            WorkbenchEmail("user1@example.com"),
+                            0
+    ) // this email is white listed
```
LizBaldo: nit: `white list ` -> `allowed list`

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskServiceInterpSpec.scala
```diff
@@ -449,6 +449,72 @@ class DiskServiceInterpSpec extends AnyFlatSpec with LeonardoTestSuite with Test
...
+    res.unsafeRunSync()(cats.effect.unsafe.IORuntime.global)
+  }
+
+  it should "list disks belonging to self only, if filtered by role=creator" in isolatedDbTest {
+    val (diskService, _) = makeDiskService()
+    val userInfo = UserInfo(OAuth2BearerToken(""),
+                            WorkbenchUserId("userId"),
+                            WorkbenchEmail("user1@example.com"),
+                            0
+    ) // this email is white listed
```
LizBaldo: Same comment as above :)

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskServiceInterpSpec.scala
```diff
@@ -449,6 +449,72 @@ class DiskServiceInterpSpec extends AnyFlatSpec with LeonardoTestSuite with Test
...
+    res.unsafeRunSync()(cats.effect.unsafe.IORuntime.global)
+  }
+
+  it should "list disks belonging to one creator only, if filtered by creator=user@email" in isolatedDbTest {
+    val (diskService, _) = makeDiskService()
+    val userInfo = UserInfo(OAuth2BearerToken(""),
+                            WorkbenchUserId("userId"),
+                            WorkbenchEmail("user1@example.com"),
+                            0
+    ) // this email is white listed
```
LizBaldo: final one :)

LizBaldo (approved):
LGTM, and I learned a lot of scala while reviewing, thanks for sharing! I left very minor documentation-related comments, feel free to ignore if you disagree :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/package.scala
mlilynolting: yes!

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskServiceInterpSpec.scala
mlilynolting: will do! for other readers: I went looking for industry conversation around these kinds of changes and the best writeup I found was [this Wired article](https://www.wired.com/story/tech-confronts-use-labels-master-slave/). `deny`/`allow` seems clear enough to me!

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/package.scala
```diff
@@ -82,4 +85,39 @@ package object service {
...
+          if (map.get(creatorOnlyValue).exists(_ == userInfo.userEmail.toString))
+            Either.right(Some(userInfo.userEmail))
+          else
+            Either.left(
+              ParseCreatorOnlyException(
+                s"Failed to process invalid value for ${creatorOnlyValue}. The only currently supported value is your own user email."
+              )
+            )
+        case _ =>
+          Either.right(None)
```
Qi77Qi: this seems a lot more complicated than https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala#L319-L326

can u add some unit tests for this helper...also we can replace https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala#L319-L326 with the function as well

Not sure if we need to have `ParseCreatorOnlyException` either, can u just reuse `BadRequestException`..Also populate the `traceId` properly

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskServiceInterpSpec.scala
```diff
@@ -442,20 +443,130 @@ class DiskServiceInterpSpec extends AnyFlatSpec with LeonardoTestSuite with Test
...
+      disk1 <- makePersistentDisk(Some(DiskName("d1"))).save()
+      // Make second disk belonging to different user than the calling user
+      disk2 <- LeoLenses.diskToCreator
+        .set(WorkbenchEmail("a_different_user@example.com"))(makePersistentDisk(Some(DiskName("d2"))))
+        .save()
+      listResponse <- diskService.listDisks(userInfo, None, Map("creator" -> "a_different_user@example.com"))
+    } yield ()
+
+    a[ParseCreatorOnlyException] should be thrownBy {
+      res.unsafeRunSync()(cats.effect.unsafe.IORuntime.global)
```
Qi77Qi: u can do `res.attempt.unsafeRunSync()` and assert the reuslt should be a left value instead

sth like 
```
res.attempt.unsafeRunSync()(cats.effect.unsafe.IORuntime.global) shouldBe Left(ParseCreatorExceotion(???)))
```

I'd also have this test be just testing the helper function, cuz this is the behavior of the helper

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/package.scala
mlilynolting: It's definitely more complex to read! I wanted to make sure that the method handled errors similarly to the others in `package.scala`, and also that the errors were relevant to the specific param passed (`role=` vs `creator=`). These can definitely change to `BadRequestException` (I was using the examples of `IllegalLabelKeyException` etc) - will do!

I plan to refactor the listRuntimes methods in a separate PR, per a convo with Liz elsewhere on this page.

Will also add unit tests and the traceId~

mlilynolting: If there is a simpler way to handle the Either() return structure please let me know; I'd much prefer a cleaner implementation if that's possible

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskServiceInterpSpec.scala
mlilynolting: I can move these cases to the helper's spec, and explore this syntax for error checking

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/package.scala
Qi77Qi: >I plan to refactor the listRuntimes methods in a separate PR, per a convo with Liz elsewhere on this page.

I'd prefer we do that in this PR since it's not that big anyway...Otherwise I'd like to see a ticket number for the follow up PR or an actual PR already :) just so I don't have to remember this

>I'd much prefer a cleaner implementation if that's possible

I think what u have is fine...but 
I'd probably do sth like 
```
  private[service] def processCreatorOnlyParameter(
    callerEmail: WorkbenchEmail,
    params: Map[String, String],
    traceId: TraceId
  ): Either[ParseCreatorOnlyException, Option[WorkbenchEmail]] = {
    val creatorOnlyFromRoleKey = params.get(creatorOnlyKey) match {
        case Some(value) => Either.cond(value == creatorOnlyValue.value, Some(callerEmail), BadRequestException("???", Some(traceId))) 
        case None => none.asRight[BadRequestException]
      } 
    
    // For backwards compatibility. xxxx
    val creatorOnlyFromCreatorKey = params.get(creatorOnlyValue) match {
        case Some(value) => Either.cond(value == callerEmail.value, Some(callerEmail), BadRequestException("???", Some(traceId))) 
        case None => none.asRight[BadRequestException]
      }
    
    creatorOnlyFromRoleKey orElse creatorOnlyFromCreatorKey
}
```

Note I'm not using `LabelMap` as type for `params` because it is not. LabelMap indicates the input is all labels, but that's not true.
I also updated `UserInfo` to `WorkbenchEmail` cuz that's all u need for this helper, u don't need any other info from `UserInfo`

mlilynolting: Magnificent thank you!!

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/PackageSpec.scala
```diff
@@ -21,4 +25,36 @@ class PackageSpec extends AnyFlatSpec with Matchers {
...
+    val email = WorkbenchEmail("user1@example.com")
+    val params: Map[String, String] = Map.empty
+    processCreatorOnlyParameter(email, params, TraceId(UUID.randomUUID())) shouldBe Right(None)
+  }
+
+  it should "reject creatorOnly parameter when role=not_creator" in {
+    val email = WorkbenchEmail("user1@example.com")
+    val params = Map("role" -> "owner")
+    val res = processCreatorOnlyParameter(email, params, TraceId(UUID.randomUUID()))
+    res.isLeft
```
mlilynolting: Do we need to check if this is a BadRequestException? since it's typed that way could it ever be anything else? I tried a number of ways to assert that the value contained in the Left is a BadRequestException and couldn't get anything to successfully compile, so either this is unnecessary or I'm bad at Scala (likely)

Qi77Qi: you can do sth like `res shouldBe Left(BadRequestException(???))`

Qi77Qi (approved):

PR #3081, opened 2022-12-20
[TOAZ-233] remove extra network creation

jordansafer:
Follow-on PR for VM landing zone integration. This PR removes network creation step and switches to using the landing zone network (which is now the default for the wsm createVM API)

---

In all cases:


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala
```diff
@@ -82,13 +82,7 @@ class RuntimeV2ServiceInterp[F[_]: Parallel](config: RuntimeServiceConfig,
...
 
       leoAuth <- samDAO.getLeoAuthToken
-      relayNamespaceOpt <- wsmDao.getRelayNamespace(workspaceId, req.region, leoAuth)
-      relayNamespace <- F.fromOption(
-        relayNamespaceOpt,
-        BadRequestException(s"Workspace ${workspaceId} doesn't have relay namespace provisioned appropriately",
-                            Some(ctx.traceId)
-        )
-      )
+      // TODO/PR question: do we still need this RelayNamespace for GCP?
```
Qi77Qi: we don't have RelayNamespace for GCP...r u wondering about Azure?

Qi77Qi (commented):
lgtm once this is tested in BEE! thanks @jordansafer !

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpWsmDao.scala
```diff
@@ -153,7 +142,101 @@ class HttpWsmDao[F[_]](httpClient: Client[F], config: HttpWsmDaoConfig)(implicit
...
+                                                         false
+      )
+      aksSubnetName <- getLandingZoneResourceName(groupedLzResources, "DeployedSubnet", AKS_NODE_POOL_SUBNET, false)
+      computeSubnetName <- getLandingZoneResourceName(groupedLzResources,
+                                                      "DeployedSubnet",
+                                                      WORKSPACE_COMPUTE_SUBNET,
+                                                      false
+      )
+      postgresSubnetName <- getLandingZoneResourceName(groupedLzResources, "DeployedSubnet", POSTGRESQL_SUBNET, false)
+    } yield LandingZoneResources(
```
Qi77Qi: if we have to call LZ api separately for each of the resources. Maybe we can have 2 models, `LandingZoneResourcesForApp` and `LandingZoneResourcesForVM`? Reason being for creating VM, many of the resources aren't needed

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala
jordansafer: You're right, I got confused- removing this!

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpWsmDao.scala
jordansafer: Edit: It is doing what we want :). The method`getLandingZoneResourceName(groupedLzResources, "DeployedSubnet", POSTGRESQL_SUBNET, false)` doesn't make any API calls, it's just a helper fn to find the resource id we are looking for in the map of lz resources. The map of all resources is returned from the api call on line 158.

Hm. We should only be calling one time, because the LandingZone API returns a payload with all of the resources listed. Let me double-check that we are doing this correctly. For this reason we should only need 1 model, since all the resources are returned from the API anyway, so all the information is available to us-

Qi77Qi: ah ok cool...just wanna make sure we dont' fire off 10 LZ calls for every runtime or app creation

jordansafer:
Successfully tested VM creation with the latest changes, and it's working as expected :). Was able to create a VM and run some python in my notebook

Qi77Qi (approved):
looks good to me! thanks @jordansafer !

Qi77Qi:
jenkins retest

PR #3082, opened 2022-12-21
IA-3926 add wait in updateDataprocImageGroupMembership

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-3926

* This is the error when the issue happens.
```
com.google.api.gax.rpc.PermissionDeniedException: io.grpc.StatusRuntimeException: PERMISSION_DENIED: Required 'compute.images.get' permission for 'projects/broad-dsp-gcr-public/global/images/aou-leo-image-dataproc-2-0-51-debian10-215b822'
	at com.google.api.gax.rpc.ApiExceptionFactory.createException(ApiExceptionFactory.java:98)
	at com.google.api.gax.rpc.ApiExceptionFactory.createException(ApiExceptionFactory.java:41)
```
From the 4 sample clusters I've looked at, the 2 that failed creation had the following log, indicating there was group addition happened during cluster creation, while the 2 successful creation didn't go through that becuz members were previously added already.
Screen Shot 2022-12-21 at 5 51 41 PM

Think what happened for some automation tests and users in prod is adding member to the dataproc image user group is taking extra long lately…
This PR adds some polling in leo to make sure the adding member has finished before we proceed dataproc cluster creation

* Remove a spec that's already covered

---

In all cases:

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/runtimes/RuntimeDataprocSpec.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/runtimes/RuntimeDataprocSpec.scala
@@ -96,57 +96,6 @@ class RuntimeDataprocSpec
     res.unsafeRunSync()(cats.effect.unsafe.IORuntime.global)
   }
 
-  "should create a Dataproc cluster with workers and preemptible workers" taggedAs Retryable in { project =>
-    val runtimeName = randomClusterName
-
-    // 2 workers and 5 preemptible workers
-    val createRuntimeRequest = defaultCreateRuntime2Request.copy(
-      runtimeConfig = Some(
-        RuntimeConfigRequest.DataprocConfig(
-          Some(2),
-          Some(MachineTypeName("n1-standard-4")),
-          Some(DiskSize(130)),
-          Some(MachineTypeName("n1-standard-4")),
-          Some(DiskSize(150)),
-          None,
-          Some(5),
-          Map.empty,
-          None,
-          true,
-          false
-        )
-      ),
-      toolDockerImage = Some(ContainerImage(LeonardoConfig.Leonardo.hailImageUrl, ContainerRegistry.GCR))
-    )
-
-    val res = dependencies.use { dep =>
-      implicit val client = dep.httpClient
-      for {
-        // create runtime
-        getRuntimeResponse <- LeonardoApiClient.createRuntimeWithWait(project, runtimeName, createRuntimeRequest)
-        runtime = ClusterCopy.fromGetRuntimeResponseCopy(getRuntimeResponse)
-
-        // check cluster status in Dataproc
-        _ <- verifyDataproc(project, runtime.clusterName, dep.dataproc, 2, 5, RegionName("us-central1"))
-
-        // check output of yarn node -list command
-        implicit0(authToken: AuthToken) <- Ron.authToken()
-        _ <- IO(
-          withWebDriver { implicit driver =>
-            withNewNotebook(runtime, Python3) { notebookPage =>
-              val output = notebookPage.executeCell("""!yarn node -list""")
-              output.get should include("Total Nodes:")
-            }
-          }
-        )
-
-        _ <- LeonardoApiClient.deleteRuntime(project, runtimeName)
-      } yield ()
-    }
-
-    res.unsafeRunSync()(cats.effect.unsafe.IORuntime.global)
-  }
-
   "should stop/start a Dataproc cluster with workers and preemptible workers" taggedAs Retryable in { project =>
     val runtimeName = randomClusterName
 
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreter.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreter.scala
@@ -13,7 +13,6 @@ import com.google.api.services.directory.model.Group
 import com.google.cloud.compute.v1.{Operation, Tags}
 import com.google.cloud.dataproc.v1.{RuntimeConfig => _, _}
 import com.typesafe.scalalogging.LazyLogging
-import org.broadinstitute.dsde.workbench.{google2, DoneCheckable, DoneCheckableInstances}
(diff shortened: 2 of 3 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreter.scala
```diff
@@ -625,44 +628,64 @@ class DataprocInterpreter[F[_]: Parallel](
...
-              )
-              _ <- updateGroupMembership(config.groupsConfig.dataprocImageProjectGroupEmail,
-                                         apiServiceAccountEmail,
-                                         createCluster
-              )
-            } yield ()
-          }
-      } yield ()
-    }
+    for {
```
Qi77Qi, line 631: diff looks bigger than reality becuz I removed the `parseImageProject` call cuz I don't think it's necessary

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreter.scala
```diff
@@ -105,6 +105,9 @@ class DataprocInterpreter[F[_]: Parallel](
     with LazyLogging {
 
   import dbRef._
+  val isMemberofGroupDonecheckable = new DoneCheckable[Boolean] {
```
LizBaldo: super nit, but `isMemberofGroupDoneCheckable`? I am still getting used to scala naming conventions so feel free to ignore

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreter.scala
LizBaldo: Could you elaborate on why you think it is not necessary anymore?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreter.scala
```diff
@@ -625,44 +628,66 @@ class DataprocInterpreter[F[_]: Parallel](
...
+            // for more information
+            dataprocServiceAccountEmail = WorkbenchEmail(
+              s"service-${projectNumber}@dataproc-accounts.iam.gserviceaccount.com"
+            )
+            _ <- updateGroupMembership(config.groupsConfig.dataprocImageProjectGroupEmail,
+                                       dataprocServiceAccountEmail,
+                                       createCluster
+            )
+            _ <-
+              if (createCluster)
```
LizBaldo, line 662: Could you add a comment explaining why a wait is added when a cluster is created?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreter.scala
```diff
@@ -625,44 +628,66 @@ class DataprocInterpreter[F[_]: Parallel](
...
+                waitUntilMemberAdded(apiServiceAccountEmail)
+              else F.unit
+          } yield ()
+        }
+    } yield ()
+
+  private def waitUntilMemberAdded(memberEmail: WorkbenchEmail): F[Boolean] = {
```
LizBaldo, line 680: Could you add a docstring to thins function explaining what it does and why? For instance what do the 60 and 5 seconds stand for, and was there any rational behind using these specific values? 
Should they live as constants elsewhere for instance?

LizBaldo (commented):
LGTM, I mostly have requests around documentation to make it easier for fellow devs to understand what is happening.
Should the new function be tested as well?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreter.scala
Qi77Qi: all it's doing is parsing the project out of a custom dataproc link and do stuff if the parsing is successful, which most likely originally somewhere we're using the parsed project, but that's not the case in the current code

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreter.scala
Qi77Qi: ?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreter.scala
Qi77Qi: is the function name clear what it's supposed to do?

> what do the 60 and 5 seconds stand for,

that's documented in `streamUntilDoneOrTimeout`

> was there any rational behind using these specific values

They're my estimates for now. Will need to monitor what happens in tests and ppl's usage to see if adjustment is needed

>Should they live as constants elsewhere for instance?

I don't think that adds too much value

Qi77Qi:
jenkins retest

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/runtimes/RuntimeDataprocSpec.scala
```diff
@@ -96,57 +96,6 @@ class RuntimeDataprocSpec
     res.unsafeRunSync()(cats.effect.unsafe.IORuntime.global)
   }
 
-  "should create a Dataproc cluster with workers and preemptible workers" taggedAs Retryable in { project =>
```
Qi77Qi, line 99: this is already covered in `should stop/start a Dataproc cluster with workers and preemptible workers` test below

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreter.scala
LizBaldo: `c` to `C` in `checkable`, but again I am not very familiar with naming conventions so feel free to ignore

LizBaldo (approved):
Thanks for addressing! Let's try it :) 🤞

PR #3088, opened 2023-01-03
[IA-3106] Github action to Create Vm Image

jdcanas:
Github action to Create GCE and Dataproc VM images via a daisy workflow

successful build here: https://github.com/DataBiosphere/leonardo/actions/runs/3943212134/jobs/6747789691 

(not triggered on most recent commits due to the goal for it be triggered manually)

### DIFF
--- a/.github/workflows/custom_image_generation.yml
+++ b/.github/workflows/custom_image_generation.yml
@@ -80,6 +80,10 @@ jobs:
       env:
         GCP_DAISY_GCR_KEY: ${{secrets.GCP_DAISY_GCR_KEY}}
 
+    - name: check
+      run: |
+        ls -al $GITHUB_WORKSPACE/jenkins/gce-custom-images
+
     - name: run script
       run: |
         gsutil ls $GCE_IMAGE_BUCKET || gsutil mb -p $GOOGLE_PROJECT -l $REGION $GCE_IMAGE_BUCKET
@@ -134,6 +138,9 @@ jobs:
           echo "$TEST_NAME"
           echo 'lsing'
           ls -al $GITHUB_WORKSPACE/jenkins/dataproc-custom-images
+        shell: bash
+        env:
+          GCP_DAISY_GCR_KEY: ${{secrets.GCP_DAISY_GCR_KEY}}
 
       - name: Init submodule
         run: |
--- a/.github/workflows/custom_image_generation.yml
+++ b/.github/workflows/custom_image_generation.yml
@@ -68,7 +68,7 @@ jobs:
     - name: Run image generation script
       run: |
          echo "echoing workspace: $GITHUB_WORKSPACE"
-         bash OUTPUT_FILE_PATH="$GITHUB_WORKSPACE/$OUTPUT_FILE_RELATIVE_PATH" $GITHUB_WORKSPACE/jenkins/gce-custom-images/create_gce_image.sh
+         OUTPUT_FILE_PATH="$GITHUB_WORKSPACE/$OUTPUT_FILE_RELATIVE_PATH" $GITHUB_WORKSPACE/jenkins/gce-custom-images/create_gce_image.sh
 
     - name: Output image URI
       run: |
@@ -111,7 +111,7 @@ jobs:
 
       - name: Run image generation script
         run: |
-          bash OUTPUT_FILE_PATH="$GITHUB_WORKSPACE/$OUTPUT_FILE_RELATIVE_PATH" $GITHUB_WORKSPACE/jenkins/dataproc-custom-images/create_dataproc_image.sh
+          OUTPUT_FILE_PATH="$GITHUB_WORKSPACE/$OUTPUT_FILE_RELATIVE_PATH" $GITHUB_WORKSPACE/jenkins/dataproc-custom-images/create_dataproc_image.sh
 
       - name: Output image URI
         run: |
--- a/.github/workflows/custom_image_generation.yml
+++ b/.github/workflows/custom_image_generation.yml
@@ -73,6 +73,7 @@ jobs:
     - name: Output image URI
       run: |
         cat $GITHUB_WORKSPACE/$OUTPUT_FILE_RELATIVE_PATH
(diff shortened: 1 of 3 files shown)


.github/workflows/custom_image_generation.yml
```diff
@@ -0,0 +1,159 @@
...
+#             - 'jenkins/gce-custom-images/**' 
+
+#     - name: display progress
+#       run: | 
+#         echo "finished inspecting the repo to determine which image generation scripts to run@@@@@@@@@@@"
+
+jobs:
+  run-image-generation-script-gce:
+    runs-on: ubuntu-latest
+    # needs: inspect-changes
```
mikeflinn:
```suggestion
    permissions:
      contents: 'read'
      id-token: 'write'
    # needs: inspect-changes
```

You need to add this block in order for workload id federation to work

.github/workflows/custom_image_generation.yml
```diff
@@ -0,0 +1,159 @@
...
+        
+        # Make the image public
+        gcloud beta compute images add-iam-policy-binding \
+        projects/${GOOGLE_PROJECT}/global/images/${OUTPUT_BASE_NAME}-gce-cos-image-${GITHUB_SHA:0:6} \
+        --member='allAuthenticatedUsers' \
+        --role='roles/compute.imageUser'
+
+  run-image-generate-script-dataproc:
+    runs-on: ubuntu-latest
+    # needs: inspect-changes
```
mikeflinn:
```suggestion
    permissions:
      contents: 'read'
      id-token: 'write'
    # needs: inspect-changes
```
Same thing here

.github/workflows/custom_image_generation.yml
```diff
@@ -0,0 +1,159 @@
...
+      - name: test step 2
+        run: echo "test 2"
+
+      - name: Create svc acct file
+        run: |
+          echo "$GCP_DAISY_GCR_KEY" > "$GITHUB_WORKSPACE/jenkins/dataproc-custom-images/$GOOGLE_CREDENTIAL_FILE_NAME"
+          echo 'echoing test name'
+          echo "$TEST_NAME"
+          echo 'catting'
+          cat $GITHUB_WORKSPACE/jenkins/dataproc-custom-images/$GOOGLE_CREDENTIAL_FILE_NAME
```
mikeflinn: Please never cat the contents of a service account file. Github actions logs are public, so any one who can see the leo repo could see this key. This should be done the same way as on line 74.

.github/workflows/custom_image_generation.yml
```diff
@@ -0,0 +1,159 @@
...
+
+      - name: Init submodule
+        run: |
+          git submodule sync
+          git submodule update --init --recursive
+
+      - name: Run image generation script
+        run: |
+          docker run --rm -v $GITHUB_WORKSPACE/jenkins/dataproc-custom-images:/work:ro gcr.io/google.com/cloudsdktool/cloud-sdk:266.0.0 bash -c "cd /work/dataproc-custom-images \
+          && gcloud auth activate-service-account --key-file=/work/$GOOGLE_CREDENTIAL_FILE_NAME \
```
mikeflinn: I think this needs to be the same file path you setup in line 126. I don't think there is actually anything at this `/work/$GOOGLE_CREDENTIAL_FILE_NAME` path. Maybe a copy paste from Jenkins?

.github/workflows/custom_image_generation.yml
jdcanas: ahh... this was just a sanity check, its not working currently, I just wanted to check if the file was indeed empty

.github/workflows/custom_image_generation.yml
jdcanas: I'm fairly sure line 139 is mounting the directory with the key into the docker image's `work` dir. I'll remove the `:ro` though, not sure what that is.

mikeflinn: Ah yeah I misread it, my bad. `:ro` means read-only volume

.github/workflows/custom_image_generation.yml
```diff
@@ -0,0 +1,169 @@
...
+      DATAPROC_IMAGE_VERSION_FORMATTED: 2-0-51-debian10
+      DATAPROC_IMAGE_BUCKET: gs://leo-dataproc-image-creation-logs
+      TEST_NAME: ${OUTPUT_BASE_NAME}-gce-cos-image-${GITHUB_SHA:0:6}
+      GCP_DAISY_GCR_KEY: ${{secrets.GCP_DAISY_GCR_KEY}}
+
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Create svc acct file
+        run: 'echo "$GCP_DAISY_GCR_KEY" > "$GITHUB_WORKSPACE/jenkins/dataproc-custom-images/$GOOGLE_CREDENTIAL_FILE_NAME"'
```
mikeflinn:
```suggestion
        run: 'echo "$GCP_DAISY_GCR_KEY" | base64 -d > "$GITHUB_WORKSPACE/jenkins/dataproc-custom-images/$GOOGLE_CREDENTIAL_FILE_NAME"'
```

Qi77Qi:
where can we find the generated image name?

also can we delete [test.tmp](https://github.com/DataBiosphere/leonardo/pull/3088/files#diff-1886643ed4b95f41725a642a0e1bc314b4b4c1658b74dfabc1b0227b05117c76)

jdcanas:
> where can we find the generated image name?
> 
> also can we delete [test.tmp](https://github.com/DataBiosphere/leonardo/pull/3088/files#diff-1886643ed4b95f41725a642a0e1bc314b4b4c1658b74dfabc1b0227b05117c76)

Deleted, and I added a step OUTPUT_IMAGE_URI that prints it. Unfortunately, there's not a good way to display it on the github actions UI that I found with my cursory googling..

n8landolt (approved):

jdcanas:
jenkins retest

Qi77Qi (approved):
looks great.. :shipit:

jdcanas:
jenkins retest

PR #3112, opened 2023-01-19
[IA-3934] Add updateDateAccessed

jdcanas:


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala
```diff
@@ -254,6 +255,34 @@ class RuntimeV2ServiceInterp[F[_]: Parallel](config: RuntimeServiceConfig,
...
+        }
+
+      _ <- ctx.span.traverse(s => F.delay(s.addAnnotation("Done auth call for update date accessed runtime permission")))
+      _ <- F
+        .raiseError[Unit](
+          RuntimeNotFoundException(runtime.cloudContext, runtimeName, "permission denied", Some(ctx.traceId))
+        )
+        .whenA(!hasPermission)
+
+      _ <- dateAccessUpdaterQueue.offer(UpdateDateAccessMessage(runtimeName, runtime.cloudContext, ctx.now))
```
LizBaldo: The original design for this endpoint is for Azure runtimes only, would it make sense to constrain it to that specific cloud context? 
It does not seem that there is anything preventing a user from calling this patch on a gcp runtime, but that could mess up with how autopause works there. What do you think?

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeServiceV2InterpSpec.scala
```diff
@@ -44,15 +44,19 @@ class RuntimeServiceV2InterpSpec extends AnyFlatSpec with LeonardoTestSuite with
   val wsmDao = new MockWsmDAO
 
   // used when we care about queue state
-  def makeInterp(queue: Queue[IO, LeoPubsubMessage], wsmDao: WsmDao[IO] = wsmDao) =
-    new RuntimeV2ServiceInterp[IO](serviceConfig, whitelistAuthProvider, wsmDao, mockSamDAO, queue)
+  def makeInterp(queue: Queue[IO, LeoPubsubMessage],
+                 wsmDao: WsmDao[IO] = wsmDao,
+                 dateAccessedQueue: Queue[IO, UpdateDateAccessMessage] = QueueFactory.makeDateAccessedQueue()
+  ) =
+    new RuntimeV2ServiceInterp[IO](serviceConfig, whitelistAuthProvider, wsmDao, mockSamDAO, queue, dateAccessedQueue)
```
LizBaldo: NIT, I know this existed before your change, but would it be difficult to change `whitelistAuthProvider` to `allowlistAuthProvider`? We should try to clean up some of leo's language whenever we see it

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeServiceV2InterpSpec.scala
```diff
@@ -867,4 +871,75 @@ class RuntimeServiceV2InterpSpec extends AnyFlatSpec with LeonardoTestSuite with
       .isInstanceOf[ParseLabelsException] shouldBe true
   }
 
+  it should "update date accessed when user has permission" in isolatedDbTest {
+    val userInfo = UserInfo(OAuth2BearerToken(""),
+                            WorkbenchUserId("userId"),
+                            WorkbenchEmail("user1@example.com"),
+                            0
+    ) // this email is white listed
```
LizBaldo: `allowed`

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeServiceV2InterpSpec.scala
```diff
@@ -867,4 +871,75 @@ class RuntimeServiceV2InterpSpec extends AnyFlatSpec with LeonardoTestSuite with
...
+    } yield msg shouldBe Some(UpdateDateAccessMessage(runtimeName, cluster.cloudContext, ctx.now))
+    res.unsafeRunSync()(cats.effect.unsafe.IORuntime.global)
+  }
+
+  it should "not update date accessed when user doesn't have permission" in isolatedDbTest {
+    val userInfo = UserInfo(OAuth2BearerToken(""),
+                            WorkbenchUserId("bad1"),
+                            WorkbenchEmail("bad1@example.com"),
+                            0
+    ) // this email is not white listed
```
LizBaldo: `not on the allow list`

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/util/QueueFactory.scala
```diff
@@ -17,4 +17,6 @@ object QueueFactory {
     Queue.bounded[IO, Event[LeoPubsubMessage]](queueSize).unsafeRunSync()
 
   def asyncTaskQueue() = Queue.bounded[IO, Task[IO]](queueSize).unsafeRunSync()
+
+  def makeDateAccessedQueue() = Queue.bounded[IO, UpdateDateAccessMessage](10).unsafeRunSync()
```
LizBaldo, line 21: What does the `10` stand for here, no need to change anything, just curious :)

LizBaldo (commented):
LGTM! I only have a couple of renaming nits, and a more open question about whether we should restrict the use of this endpoint to azure runtimes only, but I will let other chime in :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala
Qi77Qi: would be nice if we can refactor the msg to be an ADT, sth like
```
sealed abstract class UpdateDateAccessMessage extends Product with Serializable
object UpdateDateAccessMessage {
 final case class UpdateDateAccessMessageRuntimeV1(runtimeName: RuntimeName, cloudContext: CloudContext, dateAccessd: Instant) extends UpdateDateAccessMessage {
  override def toString: String =
    s"Message: ${cloudContext.asStringWithProvider}/${runtimeName.asString}, ${dateAccessd.toEpochMilli}"
}
 final case class UpdateDateAccessMessageRuntimeV2(runtimeName: RuntimeName, workspaceId: WorkspaceId, dateAccessd: Instant) extends UpdateDateAccessMessage {
  override def toString: String =
    s"Message: ${workspaceId}/${runtimeName.asString}, ${dateAccessd.toEpochMilli}"
}
}
```
Just so we can slowly move away from relying on `cloudContext`

Qi77Qi (commented):
lgtm overall...would be nice to do some e2e testing as well

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala
jdcanas: This may not be applicable to google runtimes, but I also think the direction we should try to head with v2 runtimes is make them as general as possible. The less cloud specific these endpoints the better is my opinion.

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/util/QueueFactory.scala
jdcanas: The size allocated for the queue in memory

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala
jdcanas: Lets chat in our 1-on-1 tomorrow on this!

n8landolt (approved):

LizBaldo (approved):
Thanks for addressing my comments 👍

jdcanas:
jenkins retest

jdcanas:
jenkins retest

jdcanas:
jenkins retest

jdcanas:
jenkins retest

PR #3113, opened 2023-01-20
[IA-3973] Leo - add instrumentation-enabled to the azure config

n8landolt:
Firecloud develop PR:
https://github.com/broadinstitute/firecloud-develop/pull/3218

---

In all cases:


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/InstrumentationEnabledConfig.scala
```diff
@@ -0,0 +1,5 @@
+package org.broadinstitute.dsde.workbench.leonardo.config
+
+case class InstrumentationEnabledConfig(
```
Qi77Qi: I don't think we need this to be a config on it's own....it should just be part of `AKSInterpreterConfig`

Qi77Qi (changes requested):
also fc-develop PR is needed before this can be merged

PR #3121, opened 2023-01-23
Add missing AppStatus enum values to api-docs.yaml

jmthibault79:
These are present in kubernetesModel.scala

---

In all cases:


http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -2625,11 +2625,16 @@ components:
     AppStatus:
       type: string
       enum:
-        - DELETING
         - DELETED
+        - DELETING
         - ERROR
+        - PRECREATTING
```
LizBaldo: Nit, `PRECREATING`

LizBaldo (approved):
Thanks for adding these!

http/src/main/resources/swagger/api-docs.yaml
Qi77Qi: oh, we really shouldn't be exposing these statues..I'll have a separate PR to not exposing these details to users

jmthibault79: thanks.  Should I fix this spelling or remove all of the PRE ?

jmthibault79: I removed the PRE.  Now it's just adding STARTING and alphabetizing.

Qi77Qi (approved):
thanks!

PR #3127, opened 2023-01-26
IA-3950 Update createRuntimeV2 to support PD

lucymcnatt:
Adds `useExistingDisk` boolean flag in the CreateRuntimeV2 endpoint to indicate if an existing disk should be attached to the new runtime.

**Behavior**

When true:
- Finds existing disk(s) for the given workspace created by the user
- Error if there are multiple / no disks found
- Error if the disk isn't in 'ready' state

**Added**
- useExistingDisk: Boolean to `CreateAzureRuntime` as a query param
- resourceId: `Option[WSMResourceId]` to persistent disk table (necessary in back leo for createVMRequest)

---

In all cases:


http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -2591,7 +2600,6 @@ components:
             $ref: "#/components/schemas/CreateAzureRuntimeRequest"
           example:
             labels: { saturnRuntimeName: "saturn-ab9134" }
-            region: "eastus"
```
lucymcnatt, line 2594: Qi took region out of this request a bit ago

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
```diff
@@ -301,31 +306,64 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
...
-          params.runtimeConfig.region
-        )
-      )
-      diskResp <- wsmDao.createDisk(request, leoAuth)
-      _ <- controlledResourceQuery
-        .save(params.runtime.id, diskResp.resourceId, WsmResourceType.AzureDisk)
-        .transaction
-      _ <- persistentDiskQuery.updateStatus(disk.id, DiskStatus.Ready, ctx.now).transaction
-    } yield diskResp
+    params.useExistingDisk match {
```
lucymcnatt, line 308: main back leo logic

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala
```diff
@@ -126,17 +127,46 @@ class RuntimeV2ServiceInterp[F[_]: Parallel](config: RuntimeServiceConfig,
...
-                userInfo,
-                cloudContext,
-                DiskName(req.azureDiskConfig.name.value),
-                config.azureConfig.diskConfig,
-                req,
-                landingZoneResources.region,
-                ctx.now
-              )
-            )
+            diskId <- useExistingDisk match {
```
lucymcnatt, line 153: main front leo logic

core/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/RuntimeRoutesTestJsonCodec.scala
```diff
@@ -82,9 +82,12 @@ object RuntimeRoutesTestJsonCodec {
   implicit val azureDiskConfigEncoder: Encoder[CreateAzureDiskRequest] =
     Encoder.forProduct4("labels", "name", "size", "diskType")(x => CreateAzureDiskRequest.unapply(x).get)
   implicit val createAzureRuntimeRequestEncoder: Encoder[CreateAzureRuntimeRequest] =
-    Encoder.forProduct5("labels", "machineSize", "customEnvironmentVariables", "disk", "autopauseThreshold")(x =>
-      CreateAzureRuntimeRequest.unapply(x).get
-    )
+    Encoder.forProduct5("labels",
```
LizBaldo: I am not crazy, this is just a reformatting change correct? 😅

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20230201_add_wsmResourceId_to_persistent_disk.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
LizBaldo, line 1: I might just be a bit confused, but didn't this column already exist in this table given that we do already support PDs in google environment? Or is this table only related to the V2 endpoints?

http/src/main/resources/swagger/api-docs.yaml
LizBaldo, line 2708: Just a cleanup from a past PR not using the region anymore, correct?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
```diff
@@ -464,6 +464,9 @@ object clusterQuery extends TableQuery(new ClusterTable(_)) {
   def getClusterById(id: Long)(implicit ec: ExecutionContext): DBIO[Option[Runtime]] =
     fullClusterQueryById(id).result map { recs => unmarshalFullCluster(recs).headOption }
 
+  def getClusterByPersistentDiskId(diskId: DiskId)(implicit ec: ExecutionContext): DBIO[Runtime] =
+    clusterQuery.filter(_.persistentDiskId === diskId).headOption
```
LizBaldo: So we are expecting this query to return a list of N identical clusters and return the first element only?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/RuntimeControlledResourceComponent.scala
```diff
@@ -38,11 +38,28 @@ object controlledResourceQuery extends TableQuery(new RuntimeControlledResourceT
...
       .headOption
 
+  def getWsmRecordFromResourceId(resourceId: WsmControlledResourceId,
+                                 resourceType: WsmResourceType
+  ): DBIO[Option[RuntimeControlledResourceRecord]] =
+    controlledResourceQuery
+      .filter(_.resourceId === resourceId)
+      .filter(_.resourceType === resourceType)
+      .result
+      .headOption
```
LizBaldo: I have the same question as above about why we feel comfortable just using the first element of the list

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala
```diff
@@ -126,17 +127,46 @@ class RuntimeV2ServiceInterp[F[_]: Parallel](config: RuntimeServiceConfig,
...
+              // if using existing disk, find disk in users workspace
+              case true =>
+                for {
+                  disks <- DiskServiceDbQueries
+                    .listDisks(Map.empty, includeDeleted = false, Some(userInfo.userEmail), Some(cloudContext))
+                    .transaction
+                  // check if 0 or mult disks
```
LizBaldo: Nit: `multiple`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala
```diff
@@ -126,17 +127,46 @@ class RuntimeV2ServiceInterp[F[_]: Parallel](config: RuntimeServiceConfig,
...
+                  disk <- disks.length match {
+                    case 1 => F.pure(disks.head)
+                    case 0 => F.raiseError(NoPersistentDiskException(workspaceId))
+                    case _ => F.raiseError(MultiplePersistentDisksException(workspaceId, disks.length))
```
LizBaldo: Nit, but do you want to list the disk names in the error message as well?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala
```diff
@@ -126,17 +127,46 @@ class RuntimeV2ServiceInterp[F[_]: Parallel](config: RuntimeServiceConfig,
...
+                      cloudContext,
+                      DiskName(req.azureDiskConfig.name.value),
+                      config.azureConfig.diskConfig,
+                      req,
+                      landingZoneResources.region,
+                      ctx.now
+                    )
+                  )
+                  d <- persistentDiskQuery.save(pd).transaction
+                } yield d.id
```
LizBaldo: Why yield the id now?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/MonitorAtBoot.scala
```diff
@@ -366,6 +366,7 @@ class MonitorAtBoot[F[_]](publisherQueue: Queue[F, LeoPubsubMessage],
           wid,
           storageContainer.resourceId,
           landingZoneResources,
+          false, // TODO (me) not sure if this should always be false?
```
LizBaldo: I would say so since it is the default behavior anyway but might be nice to check indeed

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
```diff
@@ -663,23 +703,27 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
...
-        .transaction
-      _ <- diskResourceOpt.traverse { disk =>
-        // TODO: once we start supporting persistent disk, we should not delete disk anymore
-        wsmDao.deleteDisk(
-          DeleteWsmResourceRequest(
-            e.workspaceId,
-            disk.resourceId,
-            DeleteControlledAzureResourceRequest(
-              WsmJobControl(WsmJobId(s"delete-disk-${ctx.traceId.asString.take(10)}"))
+      // if no existing disk being used, delete new disk created
```
LizBaldo: I am not sure I understand, there should be nothing to delete? It feels contradictory to the TODO comment 
// TODO: once we start supporting persistent disk, we should not delete disk anymore

LizBaldo (commented):
This looks really good to me, I just have a bunch of question to make sure I really understand the underlying logic and the acceptance criteria before I can give a 👍

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
lucymcnatt: so this TODO was here before my PR, since this is in `handleAzureRuntimeCreationError` basically the current behavior is if it errors during creation, delete the disk it created while creating the runtime if its exists, but with persistent disks we don't want to delete the persistent disk. It grabs from the CONTROLLED_RESOURCE table to decide which disk to delete, which is updated with the new runtimeId when using an existing disk, so the traverse may execute when using an existing disk as well

core/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/RuntimeRoutesTestJsonCodec.scala
lucymcnatt: oh yeah when I saved this file it reformatted automatically, sorry for the confusion!

http/src/main/resources/swagger/api-docs.yaml
lucymcnatt: yes! I noticed this was still hanging around from Qi's PR from a while ago

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
lucymcnatt: so the goal of this is to get the most recent runtime the PD was attached to, this has made me realize I should verify how the results of this query are ordered though

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20230201_add_wsmResourceId_to_persistent_disk.xml
lucymcnatt: so I just checked to make sure, but yeah the RUNTIME_CONTROLLED_RESOURCE table is only referenced in Azure pubsubhandler, so the resourceId seems like an azure-specific artifact. As far as I can tell, leo communicates more directly with google, or at least doesn't do so through the workspace manager so it has no need for the WSMResourceId

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/RuntimeControlledResourceComponent.scala
lucymcnatt: RUNTIME_CONTROLLED_RESOURCE table is unique by runtimeId and resourceType and rather than adding a new entry when a PD is transferred to a new disk, my logic updates the entry with the new runtimeId. I.E. it should be unique by resourceId as well. I could add that constraint in the changeset as well?

Or the alternative would be to add an entry when the PD is transferred to a new runtime, but this introduces more complications and I'd probably want to add some sort of timestamp to the table.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala
lucymcnatt: oh yeah that's a good idea

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
LizBaldo: Gotcha, yes, ordering by dateCreated would do the trick. Could you also add a comment to this function, explaining that?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/RuntimeControlledResourceComponent.scala
LizBaldo: That makes sense, thanks! Maybe add a small comment to explain your choice here?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala
LizBaldo: Probably disk id as well actually, might as well :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
LizBaldo: Ahhh, that makes a lot of sense, thanks for explaining!

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala
lucymcnatt: in both branches of this case, the id is yielded and the result is used on line 185, we don't need any of the other fields to send the createRuntime message to back leo.

LizBaldo (approved):
LGTM! 👍 if you tested it wither in a BEE or local swagger :)

mlilynolting:
I haven't yet run the code but this looks great. What are the different paths to validate - seems like given a runtime with a persistent disk, delete the runtime but retain the disk, then POST to runtime endpoint with useExistingDisk true | false? And do this for a GCP runtime running Jupyter/Rstudio, a GCP runtime running Galaxy (? is this what a cluster is for? I think I'm turned around on this point), and the runtime v1 endpoint, as well as all the above with the runtime v2 endpoint, and then do it again for an Azure runtime via the v2 endpoint?

lucymcnatt:
For your question on validating Lily, this PR is specifically targeting azure runtimes with the v2 endpoint, the main path would be:
- create an azure runtime
- delete the runtime deleteDisk = false
- create a new runtime with useExistingDisk = true
- delete the runtime with deleteDisk = true

You can validate using a GCP runtime as well if you'd like (and that's a good reminder for me to)

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -882,7 +892,7 @@ paths:
                 $ref: "#/components/schemas/ErrorReport"
 
   "/api/v2/runtimes/{workspaceId}/deleteAll":
-    post:
+    delete:
```
jdcanas: We debated on this for a bit, but it is intentionally POST. The thought is `delete` is an HTTP verb on a specific resource. Since `deleteAll` does not specify a resource (and could even result in nothing being deleted with a successful status code if no resources exist), `POST` best describes the action this endpoint performs (generic side-effectful update of state)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala
```diff
@@ -129,17 +130,55 @@ class RuntimeV2ServiceInterp[F[_]: Parallel](config: RuntimeServiceConfig,
...
+                  // check disk is ready
+                  _ <- F
+                    .raiseError(PersistentDiskNotReadyException(disk.id, disk.status))
+                    .whenA(disk.status != DiskStatus.Ready)
+
+                  // check previously attached runtime is deleted
+                  runtimeOpt <- clusterQuery.getClusterFromDiskId(disk.id).transaction
+                  _ <- F
+                    .raiseError(
+                      new Exception(
```
jdcanas: Lets use a `LeoException` with the proper status code (500) instead of a generic exception. This ensures the caller of the API gets the proper HTTP code in the case this somehow fails

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala
```diff
@@ -129,17 +130,55 @@ class RuntimeV2ServiceInterp[F[_]: Parallel](config: RuntimeServiceConfig,
...
+                        s"This should never happen. No previous runtime associated with disk ${disk.id}"
+                      )
+                    )
+                    .whenA(runtimeOpt.isEmpty)
+                  _ <- F
+                    .raiseError(NoUnattachedPersistentDiskException(disk.id, runtimeOpt.get.runtimeName))
+                    .whenA(runtimeOpt.get.status != RuntimeStatus.Deleted)
```
jdcanas: We should try to avoid `.get` whenever possible in non-test code. Its an unsafe operation that can easily bite you during refactoring (we are reliant on a previous line to check empty here). Here, this is likely better expressed as something like the following (traverse might not be the exact method you want, but you get the point)
```
runtimeOpt.traverse { runtime => runtime.status match {
        case RuntimeStatus.Deleted => F.raiseError(NoUnattachedPersistentDiskException(disk.id, runtimeOpt.get.runtimeName))
        case _ => F.unit
    }
}
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala
```diff
@@ -129,17 +130,55 @@ class RuntimeV2ServiceInterp[F[_]: Parallel](config: RuntimeServiceConfig,
...
+                  id = disk.id
```
jdcanas: why not just `yield disk.id`?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala
```diff
@@ -129,17 +130,55 @@ class RuntimeV2ServiceInterp[F[_]: Parallel](config: RuntimeServiceConfig,
...
+                } yield id
+              case _ =>
```
jdcanas: would be nice to use `case false` here, and add a comment to explain what this branch is doing. Always helps when nesting complex `for` comprehensions

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
```diff
@@ -301,31 +306,70 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
...
+              params.useExistingDisk
+            )
+          )
+          _ <- persistentDiskQuery.updateStatus(disk.id, DiskStatus.Ready, ctx.now).transaction
+          _ <- controlledResourceQuery
+            .updateRuntime(wsmDisk.resourceId, WsmResourceType.AzureDisk, params.runtime.id)
+            .transaction // set runtime to new runtimeId
+          diskResp = CreateDiskResponse(wsmDisk.resourceId)
+        } yield diskResp
+      case _ =>
```
jdcanas: `case false` here or use an if so its clear :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
```diff
@@ -301,31 +306,70 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
...
+          wsmDisk <- F.fromOption(
+            diskResourceOpt,
+            AzureRuntimeCreationError(
+              params.runtime.id,
+              params.workspaceId,
+              s"WSMResource:${resourceId} not found for disk id:${params.runtimeConfig.persistentDiskId.value}",
+              params.useExistingDisk
+            )
+          )
+          _ <- persistentDiskQuery.updateStatus(disk.id, DiskStatus.Ready, ctx.now).transaction
```
jdcanas: Hmm... why are we updating it to `Ready` here? Don't we check in front leo if its `Ready`, and if not, throw an error? I would expect that behavior to occur here as well...

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeServiceV2InterpSpec.scala
```diff
@@ -194,6 +197,65 @@ class RuntimeServiceV2InterpSpec extends AnyFlatSpec with LeonardoTestSuite with
...
+      _ <- persistentDiskQuery.updateStatus(disk.id, DiskStatus.Ready, now).transaction
+
+      err <- runtimeV2Service
+        .createRuntime(userInfo, name1, workspaceId, true, defaultCreateAzureRuntimeReq)
+        .attempt
+
+    } yield err shouldBe Left(NoUnattachedPersistentDiskException(disk.id, name0))
+    res.unsafeRunSync()(cats.effect.unsafe.IORuntime.global)
+  }
+
```
jdcanas, line 340: I'd like to see a test that ensures `createRuntime` behaves properly if called twice in a row with `useExistingDisk = true`.

I.e., the first time it should succeed and attach the second disk. The second time it should fail, as that existing disk is now attached. It's a bit redundant, but I think its a common enough use case to smash 'send' twice in a row with the same request to ensure we are doing the right thing.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
```diff
@@ -301,31 +306,70 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
...
+          _ <- controlledResourceQuery
+            .updateRuntime(wsmDisk.resourceId, WsmResourceType.AzureDisk, params.runtime.id)
+            .transaction // set runtime to new runtimeId
+          diskResp = CreateDiskResponse(wsmDisk.resourceId)
+        } yield diskResp
+      case _ =>
+        for {
+          ctx <- ev.ask
+          diskOpt <- persistentDiskQuery.getById(params.runtimeConfig.persistentDiskId).transaction
+          disk <- F.fromOption(diskOpt, new RuntimeException("no disk found"))
```
jdcanas: We typically avoid generic exceptions, and use something that inherits from `PubsubHandleMessageError`. For example, `AzureRuntimeCreationError`. This allows us to use consistent error handling. Error handling lives in `messageHandler` in `LeoPubsubMessageSubscriber`, I'd just double check to  ensure the error is being passed to a handler/case match you're happy with

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
```diff
@@ -301,31 +306,70 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
...
+              AzureDiskName(disk.name.value),
+              disk.size,
+              params.runtimeConfig.region
+            )
+          )
+          diskResp <- wsmDao.createDisk(request, leoAuth)
+          _ <- controlledResourceQuery
+            .save(params.runtime.id, diskResp.resourceId, WsmResourceType.AzureDisk)
+            .transaction
+          _ <- persistentDiskQuery.updateStatus(disk.id, DiskStatus.Ready, ctx.now).transaction
```
jdcanas, line 379: Hmm... same comment, why do we need to update it to `Ready` here, didn't we already error if it wasn't `Ready`?

jdcanas: Ah, just kidding, this is the branch where the disk does not exist. Ignore me :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
```diff
@@ -663,23 +709,28 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
...
-        .transaction
-      _ <- diskResourceOpt.traverse { disk =>
-        // TODO: once we start supporting persistent disk, we should not delete disk anymore
-        wsmDao.deleteDisk(
-          DeleteWsmResourceRequest(
-            e.workspaceId,
-            disk.resourceId,
-            DeleteControlledAzureResourceRequest(
-              WsmJobControl(WsmJobId(s"delete-disk-${ctx.traceId.asString.take(10)}"))
+      deleteDiskAction =
```
jdcanas: Hmm... semantically, I find it reads a bit weird to assign this to a variable and check `e.useExistingDisk` after. I would prefer to lead with `e.useExistingDisk match` here, so its clear we only delete newly-created disks associated with this VM creation. Maybe a comment is in order too, any time we are deleting disks should be very well commented 😬

jdcanas (commented):
Looks very good overall! Few things here and there

http/src/main/resources/swagger/api-docs.yaml
lucymcnatt: ohhh ok, I honestly thought it was a mistake, but I should've checked with you guys, that makes sense!

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
lucymcnatt: good catch, this was leftover from when I wasn't checking the status in front leo, I removed it

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
lucymcnatt: totally makes sense, I switched it to be a case and added a comment!

PR #3128, opened 2023-01-27
[AJ-824 AJ-822 AJ-827] allow individual Cromwhelm containers to be enabled/disabled

rtitle:
Needs testing with https://github.com/broadinstitute/cromwhelm/pull/178.

Adds ability in Leo to control which services in the `cromwhelm` chart are enabled. Initially we would like to enable WDS-only, then enable the others over time.

---

In all cases:


http/src/main/resources/reference.conf
```diff
@@ -215,23 +215,23 @@ azure {
      # See https://github.com/broadinstitute/cromwhelm/blob/main/terra-batch-libchart/templates/_reverse-proxy.tpl#L81-L114
      # for list of services.
      services = [
```
rtitle: Figured I can use this array of services to control which are enabled/disabled in cromwhelm. This array also populates the `proxyUrls` in the Leo `getApp` JSON response.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/KubernetesAppConfig.scala
```diff
@@ -74,4 +75,19 @@ final case class CoaAppConfig(chartName: ChartName,
...
+      case ServiceName("cromwell")                      => Cromwell
+    }
+    .toSet
+}
+
+sealed trait CoaService
+object CoaService {
+  final case object Cbas extends CoaService
+  final case object Wds extends CoaService
+  final case object Cromwell extends CoaService
```
jgainerdewar, line 94: Wondering if we should include TES here as well. I assume there will be a time when we are deploying TES in dev, but not ready to deploy to prod yet. We could also just lump that in with Cromwell.

rtitle: Yeah it might make sense to track TES here, I could imagine wanting to conditionally enable/disable it. I can add an enum value, it won't be used yet though.

aaronkanzer: Just as an FYI, I commented out TES in https://github.com/broadinstitute/cromwhelm/pull/168

Let me know if it would be helpful to uncomment and surround the TES Deployment template with similar `enabled` flags :D

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -411,6 +413,11 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
         // Sam configs
         raw"sam.url=${config.samConfig.server}",
 
+        // Enabled services configs
```
aaronkanzer, line 417: Do we need to also include `cbasUI` here? (Perhaps it is tied 1<>1 with Cbas being present)?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/KubernetesAppConfig.scala
rtitle: Yeah -- I think fine to leave it commented, as it didn't work anyway. We can re-add TES to cromwhelm when we're ready to support it.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
rtitle: Oh -- good catch! Yeah will fix that

rtitle:
Tested on a BEE along with https://github.com/broadinstitute/cromwhelm/pull/178, it works well.

rtitle:
jenkins retest

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -327,16 +328,17 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
...
-        cbasDao
-          .getStatus(relayBaseUri, authHeader)
-          .handleError(_ => false),
-        wdsDao
-          .getStatus(relayBaseUri, authHeader)
-          .handleError(_ => false)
-        // TODO (TOAZ-241): add cromwell to the status checks once it starts up
-        // cromwellDao.getStatus(relayBaseUri, authHeader).handleError(_ => false)
-      ).sequence
+      op = config.coaAppConfig.coaServices
```
aaronkanzer, line 331: Does `CbasUI` need to be included in these services as well?

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreterSpec.scala
```diff
@@ -141,6 +141,10 @@ class AKSInterpreterSpec extends AnyFlatSpecLike with TestComponent with Leonard
       "identity.resourceId=identity-id," +
       "identity.clientId=identity-client-id," +
       "sam.url=https://sam.dsde-dev.broadinstitute.org/," +
+      "cbas.coaEnabled=true," +
```
aaronkanzer, line 144: Just double-checking here (e.g. if we need to set some of these false) -- I assume we are OK with all being `enabled` from the Leo side, then overriding those values in the downstream Helm chart?

Dockerfile
```diff
@@ -29,8 +29,8 @@ ENV TERRA_APP_VERSION 0.5.0
 ENV GALAXY_VERSION 1.6.1
 ENV NGINX_VERSION 4.3.0
 # If you update this here, make sure to also update reference.conf:
-ENV CROMWELL_CHART_VERSION 0.2.187
-ENV CROWELL_ON_AZURE_CHART_VERSION 0.2.187
+ENV CROMWELL_CHART_VERSION 0.2.190
+ENV CROWELL_ON_AZURE_CHART_VERSION 0.2.190
```
davidangb: now that cromwhelm is updated, this should be `0.2.191` throughout

Edit: unless you'd rather leave it to #3133

davidangb (approved):
looks good, did not test myself

Dockerfile
rtitle: I bumped it to 191 -- would rather just have 1 Leo PR

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreterSpec.scala
rtitle: Yeah I would prefer to have Leo control the enablement instead of the helm chart, so we can set it per-env. See this PR which sets them all enabled for dev/BEE but WDS-only for alpha/staging/prod: https://github.com/broadinstitute/firecloud-develop/pull/3228

aaronkanzer: sounds good!

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
aaronkanzer: Ah disregard -- there isn't a DAO there for CBAS UI

aaronkanzer (approved):

jordansafer (approved):

rtitle:

rtitle:
Tested both cases (all services, wds-only) on a BEE with a recent Azure landing zone. WDS spun up and I was able to upload/download TSVs. AKS looks happy -- 8 deployments, 2 nodes, no Error'd pods.

Swatomation has been flaky because of Google groups issues which is unrelated -- given manual testing was done I am admin-merging this.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -411,6 +414,12 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
         // Sam configs
         raw"sam.url=${config.samConfig.server}",
 
+        // Enabled services configs
+        raw"cbas.coaEnabled=${config.coaAppConfig.coaServices.contains(Cbas)}",
```
aaronkanzer, line 418: Flag for `cromwhelm` updates

PR #3134, opened 2023-01-31
[IA-4010] Delete All Apps associated with a workspaceId

jdcanas:
This is tested in live Leo. It works in the base case present in the public preview, where a the workspace owner has an app created by default and no additional users in the workspace will have an app.

However, it **does not** work for workspaces with apps not owned by the workspace owner, due to a sam permission model issue. More details can be found [here](https://broadinstitute.slack.com/archives/C02GJ782BAA/p16751965562040890) with the ticket for this fix found [here](https://broadworkbench.atlassian.net/browse/IA-4012)

### DIFF
--- a/http/src/main/resources/swagger/api-docs.yaml
+++ b/http/src/main/resources/swagger/api-docs.yaml
@@ -1732,7 +1732,7 @@ paths:
     post:
       summary: Deletes ALL apps associated with the given workspace Id
       description: deletes all Apps in a workspace
-      operationId: deleteAllAppV2
+      operationId: deleteAllAppsV2
       tags:
         - appsV2
       parameters:
@@ -1751,11 +1751,11 @@ paths:
             default: false
       responses:
         "202":
-          description: App deletion request accepted
+          description: App(s) deletion request accepted
         "403":
           description: User does not have permission to perform action on App(s)
         "404":
-          description: App not found
+          description: App(s) not found
           content:
             application/json:
               schema:
--- a/http/src/main/resources/swagger/api-docs.yaml
+++ b/http/src/main/resources/swagger/api-docs.yaml
@@ -1728,6 +1728,45 @@ paths:
                 items:
                   $ref: "#/components/schemas/ListAppResponse"
 
+  "/api/apps/v2/{workspaceId}/deleteAll":
+    post:
+      summary: Deletes ALL apps associated with the given workspace Id
+      description: deletes all Apps in a workspace
+      operationId: deleteAllAppV2
+      tags:
+        - appsV2
+      parameters:
+        - in: path
+          name: workspaceId
+          description: workspaceId
+          required: true
+          schema:
+            type: string
+        - in: query
+          name: deleteDisk
+          description: Whether or not the disk associated with the apps should be deleted. Default to false if not specified.
+          required: false
+          schema:
+            type: boolean
+            default: false
+      responses:
+        "202":
+          description: App deletion request accepted
+        "403":
+          description: User does not have permission to perform action on App(s)
+        "404":
+          description: App not found
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/ErrorReport"
+        "500":
+          description: Internal Error
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/ErrorReport"
+
   ## Deprecated Notebook API ##
 
   "/notebooks/{googleProject}/{clusterName}":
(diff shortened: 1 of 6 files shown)


http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -1728,6 +1728,45 @@ paths:
                 items:
                   $ref: "#/components/schemas/ListAppResponse"
 
+  "/api/apps/v2/{workspaceId}/deleteAll":
+    post:
```
LizBaldo, line 1732: So post is correct, and not delete here?

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -1728,6 +1728,45 @@ paths:
...
+      summary: Deletes ALL apps associated with the given workspace Id
+      description: deletes all Apps in a workspace
+      operationId: deleteAllAppV2
```
LizBaldo: Nit: `deleteAllAppsV2`?

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -1728,6 +1728,45 @@ paths:
...
+        - in: query
+          name: deleteDisk
+          description: Whether or not the disk associated with the apps should be deleted. Default to false if not specified.
+          required: false
+          schema:
+            type: boolean
+            default: false
+      responses:
+        "202":
+          description: App deletion request accepted
```
LizBaldo: Nit: App(s) deletion request accepted

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -1728,6 +1728,45 @@ paths:
...
+        "403":
+          description: User does not have permission to perform action on App(s)
+        "404":
+          description: App not found
```
LizBaldo: Nit: App(s) not found

jdcanas:
jenkins retest

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterpSpec.scala
```diff
@@ -1907,6 +1907,75 @@ final class AppServiceInterpSpec extends AnyFlatSpec with LeonardoTestSuite with
     res.swap.toOption.get.getMessage shouldBe s"Persistent disk Gcp/${workspaceId}/disk is already formatted by CROMWELL"
   }
 
+  it should "V2 Azure - deleteAllApp, update all status appropriately, and queue multiple messages" in isolatedDbTest {
```
LizBaldo, line 1910: Would it make sense to have a similar test with apps in Google as well since the V2 route is cloud context agnostic?

LizBaldo (approved):
LGTM 👍 Thanks for providing me with a template for doing a similar thing with runtimes :)

I left a few nit comments, as well as a question around wether we should be testing this with google as well.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppService.scala
```diff
@@ -64,6 +65,10 @@ trait AppService[F[_]] {
     appName: AppName,
     deleteDisk: Boolean
   )(implicit as: Ask[F, AppContext]): F[Unit]
+
+  def deleteAllAppV2(userInfo: UserInfo, workspaceId: WorkspaceId, deleteDisk: Boolean)(implicit
```
n8landolt: For consistency, it seems like we want the plural, `deleteAllAppsV2`

http/src/main/resources/swagger/api-docs.yaml
jdcanas: Yeah, this is what Rob recommended. DELETE is for a single entity, and PATCH isn't quite right either since no resources are modified. POST is a generic catch-all for 'modify resources with side effects'

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -1728,6 +1728,45 @@ paths:
...
+        - in: query
+          name: deleteDisk
+          description: Whether or not the disk associated with the apps should be deleted. Default to false if not specified.
+          required: false
+          schema:
+            type: boolean
+            default: false
+      responses:
+        "202":
+          description: App(s) deletion request accepted
```
cahrens, line 1754: So this call is asynchronous, correct? That means that we will not have a way of knowing when/if the deletion ultimately succeeds. What is the suggestion for us to determine that? Should we poll WSM to see if resources still exist, or poll a getApp method? It would be great if there was a way we could poll the deletion itself (like a job control ID being returned) so that if it fails, we can present the user with the cause of failure.

jdcanas: Right now leo does not support 'jobs' like WSM does. The resources themselves are only ever soft-deleted, so you can query the list endpoint with the `workspaceId` and `includeDelete=true`. Each app deletion will be processed separately and in parallel, so any apps that fail deletion will be marked with `Error`, and any error(s) included in the list response.

PR #3136, opened 2023-01-31
[SUP-802] Added logic to prevent a disk from being deleted when it's in a valid state

n8landolt:
https://broadworkbench.atlassian.net/browse/SUP-802

This ticket fixes a bug where a runtime errors out and deletes the user's PD without request.

To recreate - make a jupyter runtime, delete the runtime, create a new runtime with a failing user script, and then the PD is set to deleted, however it still exists on the Google side.

For reviewers - is this added logic the outcome we want? Or do we want to just leave the PD alone? "Acceptance critera" isn't defined, so I made some assumptions about the outcome.

---

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BaseCloudServiceRuntimeMonitor.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BaseCloudServiceRuntimeMonitor.scala
@@ -167,7 +167,7 @@ abstract class BaseCloudServiceRuntimeMonitor[F[_]] {
                 persistentDiskOpt <- rc.persistentDiskId.flatTraverse(did =>
                   persistentDiskQuery.getPersistentDiskRecord(did).transaction
                 )
-                _ = persistentDiskOpt match {
+                _ <- persistentDiskOpt match {
                   case Some(value) =>
                     if (value.status == DiskStatus.Creating || value.status == DiskStatus.Failed) {
                       persistentDiskOpt.traverse_(d =>
@@ -175,7 +175,7 @@ abstract class BaseCloudServiceRuntimeMonitor[F[_]] {
                           .updateStatus(d.id, DiskStatus.Deleted, ctx.now)
                           .transaction
                       )
-                    }
+                    } else F.unit
                   case None => F.unit
                 }
               } yield ()
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BaseCloudServiceRuntimeMonitor.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BaseCloudServiceRuntimeMonitor.scala
@@ -167,11 +167,17 @@ abstract class BaseCloudServiceRuntimeMonitor[F[_]] {
                 persistentDiskOpt <- rc.persistentDiskId.flatTraverse(did =>
                   persistentDiskQuery.getPersistentDiskRecord(did).transaction
                 )
-                _ <- persistentDiskOpt.traverse_(d =>
-                  googleDisk.deleteDisk(googleProject, rc.zone, d.name) >> persistentDiskQuery
-                    .updateStatus(d.id, DiskStatus.Deleted, ctx.now)
-                    .transaction
-                )
+                _ = persistentDiskOpt match {
+                  case Some(value) =>
+                    if (value.status == DiskStatus.Creating || value.status == DiskStatus.Failed) {
+                      persistentDiskOpt.traverse_(d =>
+                        googleDisk.deleteDisk(googleProject, rc.zone, d.name) >> persistentDiskQuery
+                          .updateStatus(d.id, DiskStatus.Deleted, ctx.now)
+                          .transaction
+                      )
+                    }
+                  case None => F.unit
+                }
               } yield ()
             }
           } yield ()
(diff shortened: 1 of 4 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BaseCloudServiceRuntimeMonitor.scala
```diff
@@ -167,11 +167,17 @@ abstract class BaseCloudServiceRuntimeMonitor[F[_]] {
...
                   persistentDiskQuery.getPersistentDiskRecord(did).transaction
                 )
-                _ <- persistentDiskOpt.traverse_(d =>
-                  googleDisk.deleteDisk(googleProject, rc.zone, d.name) >> persistentDiskQuery
-                    .updateStatus(d.id, DiskStatus.Deleted, ctx.now)
-                    .transaction
-                )
+                _ = persistentDiskOpt match {
+                  case Some(value) =>
+                    if (value.status == DiskStatus.Creating || value.status == DiskStatus.Failed) {
```
LizBaldo, line 172: Am I understanding correctly that we want to delete any disk in a non ready status? In that case, it might be easier to have the following condition:
`if (value.status != DiskStatus.Ready)`

LizBaldo: Regardless, adding a comment explaining the expected behavior would be nice :)

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BaseCloudServiceRuntimeMonitorSpec.scala
```diff
@@ -255,7 +255,7 @@ class BaseCloudServiceRuntimeMonitorSpec extends AnyFlatSpec with Matchers with
       // handleCheckTools should have timed out after 10 seconds and the runtime should remain in Deleted status
       elapsed should be >= 10000L
       status shouldBe Some(RuntimeStatus.Deleted)
-      diskStatus shouldBe (Some(DiskStatus.Deleted))
+      diskStatus shouldBe (Some(DiskStatus.Ready))
```
LizBaldo, line 258: For this file and the next, it would be nice to keep the existing test that triggers the disk deletion)

LizBaldo (commented):
Am I understanding correctly that

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BaseCloudServiceRuntimeMonitor.scala
n8landolt: Personally, I don't think we should affect the PD at all. But if we do want to, do we know what the 'Restoring' status is used for? I would think we don't want to delete a 'Restoring' disk as well.

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BaseCloudServiceRuntimeMonitorSpec.scala
n8landolt: Agreed! I'll get a test in there before I request your review again.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BaseCloudServiceRuntimeMonitor.scala
jdcanas: I like the idea of being explicit in the statuses we want to delete instead of the inverse. Because users often save disks between runtimes, I'd hate to eagerly delete.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -1365,11 +1364,17 @@ class LeoPubsubMessageSubscriber[F[_]](
...
                         persistentDiskQuery.getPersistentDiskRecord(did).transaction
                       )
-                      _ <- persistentDiskOpt.traverse_(d =>
-                        googleDiskService.deleteDisk(googleProject, rc.zone, d.name) >> persistentDiskQuery
-                          .updateStatus(d.id, DiskStatus.Deleted, now)
-                          .transaction
-                      )
+                      _ = persistentDiskOpt match {
+                        case Some(value) =>
+                          if (value.status == DiskStatus.Creating || value.status == DiskStatus.Failed) {
```
jdcanas, line 1369: Huh... I'm surprised this compiles?

This is probably the reason your test is failing. You never actually run the IO from `persistentDiskOpt.traverse_(d =>
                              googleDiskService.deleteDisk(googleProject, rc.zone, d.name) >> persistentDiskQuery
                                .updateStatus(d.id, DiskStatus.Deleted, now)
                                .transaction
                            )`

In other words, Line 1367 should read 
`_ <- persistentDiskOpt match {` not `_ = persistentDiskOpt match {`

Furthermore, that will cause a syntax error, as the type from the `if() { ... }` branch is F[_] and the type of the else branch is None (aka not included, and should likely be `else F.unit`)

mlilynolting:
How would I test the two distinct error handlers, as a reviewer? The LeoPubSubMessageSubscriber and the BaseCloudServiceRuntimeMonitor presumably are triggering these sections of code in different situations, and I'm not sure when one would fire over the other.

LizBaldo (approved):
Thanks for putting back the tests!

You have my 👍, assuming that you confirmed the logic and expected user experience with Alessandro?

jdcanas (approved):

n8landolt:
jenkins retest

n8landolt:
jenkins retest

n8landolt:
jenkins retest

PR #3138, opened 2023-01-31
[IA-4009] Delete All Runtimes associated with a workspaceId

LizBaldo:
Jira ticket: https://broadworkbench.atlassian.net/browse/IA-4009

Tested locally on leonardo 👍


http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/CommonTestData.scala
```diff
@@ -544,6 +544,19 @@ object CommonTestData {
     Some(0)
   )
 
+  val default2CreateAzureRuntimeReq = CreateAzureRuntimeRequest(
```
jdcanas: Nit: if you'd like, you can `.copy` on the original  `defaultCreateAzureRuntimeReq` and just change the disk name.

I.E. (just replace `diskReq` with actual param name)
defaultCreateAzureRuntimeReq.copy(diskReq = defaultCreateAzureRuntimeReq.diskReq.copy(diskName = AzureDiskName("diskName2")))
`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala
```diff
@@ -265,6 +265,22 @@ class RuntimeV2ServiceInterp[F[_]: Parallel](config: RuntimeServiceConfig,
...
+    for {
+      ctx <- as.ask
+      runtimes <- RuntimeServiceDbQueries
+        .listRuntimesForWorkspace(Map.empty, true, None, Some(workspaceId), None)
+        .map(_.toList)
+        .transaction
+      // In ListRuntimeResponse2, the runtimeName is weirdly named clusterName instead
+      runtime_names = runtimes.map(r => r.clusterName)
+
+      _ <- runtime_names.traverse(runtime_name => deleteRuntime(userInfo, runtime_name, workspaceId, deleteDisk))
```
jdcanas: I don't think this is exactly what we want. This gets the job done, but it causes a lot of extra DB queries, which can be expensive. See what I did for apps, I split the default `delete` function into two parts, the db lookup, then a function that takes the app for the rest of the flow. Should be possible here to make a third function that takes a `runtime`

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/CommonTestData.scala
LizBaldo: Ohh, this is a lot nicer indeed, I will do that, thanks for the tip :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala
LizBaldo: Good point, thanks for the catch, I will look closer at your PR and update :)

LizBaldo: Actually, the db queries are different between the get active runtime, and the list runtime. One returns a `ClusterRecord` that contains a lot more information than the `ListRuntimeResponse2`, so separating the two seems a bit delicate

LizBaldo: I scratched my head for 10 more minutes and I do not see an obvious way of not querying for the active runtime record every time 🤔

LizBaldo:
jenkins retest

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -885,6 +885,45 @@ paths:
...
+        - in: query
+          name: deleteDisk
+          description: Whether or not the disk associated with the apps should be deleted. Default to false if not specified.
+          required: false
+          schema:
+            type: boolean
+            default: false
+      responses:
+        "202":
+          description: Runtime(s) deletion request accepted
```
cahrens, line 907: So this call is asynchronous, correct? That means that we will not have a way of knowing when/if the deletion ultimately succeeds. What is the suggestion for us to determine that? Should we poll WSM to see if resources still exist, or poll a getRuntime method? It would be great if there was a way we could poll the deletion itself (like an job control ID being returned) so that if it fails, we can present the user with the cause of failure.

LizBaldo: You should be able to call the `ListRuntimes` method for that workspace to see the statuses @cahrens

cahrens: OK, we also have a few other questions, so I scheduled a meeting for Workspaces and IA to touch base.

LizBaldo:
jenkins retest

jdcanas (approved):

PR #3139, opened 2023-02-01
IA-3995: enable retries in http4s client

rtitle:
https://broadworkbench.atlassian.net/browse/IA-3995

I think we inadvertently weren't enabling retries in the http4s client which we commonly use for WSM, Sam, and others. Theory is this made us un-resilient to socket exceptions, etc.

---

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/Boot.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/Boot.scala
@@ -725,9 +725,9 @@ object Boot extends IOApp {
       httpClientWithLogging = Http4sLogger[F](logHeaders = true, logBody = false, logAction = Some(s => logAction(s)))(
         httpClient
       )
-      client = if (withRetry) Retry(retryPolicy)(httpClientWithLogging) else httpClientWithLogging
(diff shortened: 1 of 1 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/Boot.scala
```diff
@@ -736,7 +736,7 @@ object Boot extends IOApp {
             meteredClient = Metrics[F](
               metricsOps,
               classifierFunc
-            )(httpClientWithLogging)
+            )(clientWithRetry)
```
rtitle, line 739: This code path previously wasn't using the retry-enabled client

LizBaldo, line 739: ohhh good catch 👀

LizBaldo (approved):

rtitle:
BEE testing looks good, I think -- at least, I was able to create/delete a VM a few times. I also validated that all cloud resources (VM, 2 disks, nic) were all deleted after Leo runtime deletion.

aherbst-broad (approved):

rtitle:
wow swatomation passed

PR #3144, opened 2023-02-06
[IA-4010] add deleteAll exception and test

jdcanas:
Adds an exception if you call deleteAll and there is a non-deletable app

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
@@ -1236,7 +1236,7 @@ case class AppCannotBeDeletedByWorkspaceIdException(workspaceId: WorkspaceId,
 case class DeleteAllAppsCannotBePerformed(workspaceId: WorkspaceId, apps: List[App], traceId: TraceId)
     extends LeoException(
       s"App(s) in workspace ${workspaceId.value.toString} with (name(s), status(es)) ${apps
-          .map(app => s"(${app.appName.value},${app.status})")} cannot be deleted their status(es).",
+          .map(app => s"(${app.appName.value},${app.status})")} cannot be deleted due to their status(es).",
       StatusCodes.Conflict,
       traceId = Some(traceId)
     )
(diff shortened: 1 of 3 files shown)


jdcanas:
jenkins retest

n8landolt (approved):

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
```diff
@@ -1227,12 +1233,10 @@ case class AppCannotBeDeletedByWorkspaceIdException(workspaceId: WorkspaceId,
       traceId = Some(traceId)
     )
 
-case class AppCannotBeCreatedException(googleProject: GoogleProject,
```
lucymcnatt, line 1230: why is this exception class being removed?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
```diff
@@ -1227,12 +1233,10 @@ case class AppCannotBeDeletedByWorkspaceIdException(workspaceId: WorkspaceId,
...
-                                       appName: AppName,
-                                       status: AppStatus,
-                                       traceId: TraceId
-) extends LeoException(
-      s"App ${googleProject.value}/${appName.value} cannot be created in ${status} status.",
+case class DeleteAllAppsCannotBePerformed(workspaceId: WorkspaceId, apps: List[App], traceId: TraceId)
+    extends LeoException(
+      s"App(s) in workspace ${workspaceId.value.toString} with (name(s), status(es)) ${apps
+          .map(app => s"(${app.appName.value},${app.status})")} cannot be deleted their status(es).",
```
lucymcnatt: nit: should this be "cannot be deleted due to their status(es)"?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
jdcanas: Its not used anywhere, just cleaning up

lucymcnatt (approved):

PR #3153, opened 2023-02-08
[IA-3957] Add environment variables to azure_vm_init_script

n8landolt:
Users would like access to certain environment variables. This script update will add these environment variables to the available kernel.json files.

Note - for some reason the `/anaconda/envs/py38_default/share/jupyter/kernels/python3/kernel.json` doesn't "take" the env variables.

You can view the available kernel directories by running the following in the jupyter notebook:

`!jupyter kernelspec list`

To read it, run `!cat <dir>/kernel.json`

---

In all cases:


http/src/main/resources/init-resources/azure_vm_init_script.sh
```diff
@@ -47,6 +47,13 @@ WELDER_OWNER_EMAIL="${13:-dummy}"
 WELDER_STAGING_BUCKET="${14:-dummy}"
 WELDER_STAGING_STORAGE_CONTAINER_RESOURCE_ID="${15:-dummy}"
 
+# Envs for Azure
+WORKSPACE_ID="${10:-dummy}"
```
rtitle: I'm not sure this will actually set the EVs for the user in their notebook kernel -- I think this would just set the variable for this script. Even `export FOO=BAR` may not do it -- in the past we had to hook into kernel startup to set EVs for users to see.

Happy to chat more, or be proven wrong..

n8landolt: What's the process for that?
I don't have much familiarity with this - I'm just attempting to follow the ticket's notes. Would appreciate some guidance!

http/src/main/resources/init-resources/azure_vm_init_script.sh
```diff
@@ -47,6 +47,12 @@ WELDER_OWNER_EMAIL="${13:-dummy}"
...
```
rtitle: I think these args need to start at `${16-dummy}` and go to `${19-dummy}`

http/src/main/resources/init-resources/azure_vm_init_script.sh
```diff
@@ -99,3 +105,10 @@ docker run -d --restart always --network host --name welder \
 --env STAGING_BUCKET=$WELDER_STAGING_BUCKET \
 --env SHOULD_BACKGROUND_SYNC="false" \
 $WELDER_WELDER_DOCKER_IMAGE
+
+echo "{ 'env': { 'WORKSPACE_ID': '${WORKSPACE_ID}', 'WORKSPACE_STORAGE_CONTAINER_ID': '${WORKSPACE_STORAGE_CONTAINER_ID}', 'WORKSPACE_NAME': '${WORKSPACE_NAME}', 'WORKSPACE_STORAGE_CONTAINER_URL': '${WORKSPACE_STORAGE_CONTAINER_URL}' }}" | jq . >> wsenv.json
```
rtitle: I think the quotes are a little messed up here. Json fields need double quotes instead of single. Maybe use `jq` to generate the json too? This seems to work:
```
jq --null-input \
--arg workspace_id $WORKSPACE_ID \
--arg workspace_storage_container_id $WORKSPACE_STORAGE_CONTAINER_ID \
--arg workspace_name $WORKSPACE_NAME \
--arg workspace_storage_container_url $WORKSPACE_STORAGE_CONTAINER_URL \
'{ "env": { "WORKSPACE_ID": $workspace_id, "WORKSPACE_STORAGE_CONTAINER_ID": $workspace_storage_container_id, "WORKSPACE_NAME": $workspace_name, "WORKSPACE_STORAGE_CONTAINER_URL": $workspace_storage_container_url }}' \
> wsenv.json
```

http/src/main/resources/init-resources/azure_vm_init_script.sh
```diff
@@ -99,3 +105,10 @@ docker run -d --restart always --network host --name welder \
...
+jq -s . /usr/local/share/jupyter/kernels/julia-1.6/kernel.json wsenv.json > /usr/local/share/jupyter/kernels/julia-1.6/kernel.json
```
rtitle: Was testing and this makes a top-level array with the existing json and `env` structure, which isn't correct.
```
jupyter@vm0186808a1:~$  jq -s . /usr/local/share/jupyter/kernels/julia-1.6/kernel.json wsenv.json 
[
  {
    "display_name": "Julia 1.6.0",
    "argv": [
      "/opt/julia/julia-1.6.0/bin/julia",
      "-i",
      "--color=yes",
      "--project=@.",
      "/opt/julia/latest/packages/packages/IJulia/AQu2H/src/kernel.jl",
      "{connection_file}"
    ],
    "language": "julia",
    "env": {},
    "interrupt_mode": "signal"
  },
  {
    "env": {
      "WORKSPACE_ID": "foo",
      "WORKSPACE_STORAGE_CONTAINER_ID": "bar",
      "WORKSPACE_NAME": "name",
      "WORKSPACE_STORAGE_CONTAINER_URL": "url"
    }
  }
]
```

Using `-s add` seems to work. Fixed:
```
jupyter@vm0186808a1:~$ jq -s add /usr/local/share/jupyter/kernels/julia-1.6/kernel.json wsenv.json
{
  "display_name": "Julia 1.6.0",
  "argv": [
    "/opt/julia/julia-1.6.0/bin/julia",
    "-i",
    "--color=yes",
    "--project=@.",
    "/opt/julia/latest/packages/packages/IJulia/AQu2H/src/kernel.jl",
    "{connection_file}"
  ],
  "language": "julia",
  "env": {
    "WORKSPACE_ID": "foo",
    "WORKSPACE_STORAGE_CONTAINER_ID": "bar",
    "WORKSPACE_NAME": "name",
    "WORKSPACE_STORAGE_CONTAINER_URL": "url"
  },
  "interrupt_mode": "signal"
}
```

http/src/main/resources/reference.conf
```diff
@@ -194,7 +194,7 @@ azure {
         type = "CustomScript",
         version = "2.1",
         minor-version-auto-upgrade = true,
-        file-uris = ["https://raw.githubusercontent.com/DataBiosphere/leonardo/d6f5eea1a9299f8ef95cf1dc3eaf40594af26782/http/src/main/resources/init-resources/azure_vm_init_script.sh"]
+        file-uris = ["https://raw.githubusercontent.com/DataBiosphere/leonardo/24d19970c5785b5e8f94d7660dcc8a0b4c497ff9/http/src/main/resources/init-resources/azure_vm_init_script.sh"]
```
rtitle: So Leo will always use the `azure_vm_init_script.sh` from this git hash. It looks like https://raw.githubusercontent.com/DataBiosphere/leonardo/24d19970c5785b5e8f94d7660dcc8a0b4c497ff9/http/src/main/resources/init-resources/azure_vm_init_script.sh is an older version of script which doesn't have the changes in this branch. That's probably why you weren't seeing the changes.

I think you want https://raw.githubusercontent.com/DataBiosphere/leonardo/ad6056757f684e2bd82fc1d90240dab7db333862/http/src/main/resources/init-resources/azure_vm_init_script.sh to point at the current head version in this branch. 

Or even better use https://raw.githubusercontent.com/DataBiosphere/leonardo/IA-3957-include-additional-create-parameters/http/src/main/resources/init-resources/azure_vm_init_script.sh to have it track the branch for convenience (just change back to hash before merge).

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
```diff
@@ -63,14 +66,35 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
         case x: RuntimeConfig.AzureConfig => F.pure(x)
         case x => F.raiseError(new RuntimeException(s"this runtime doesn't have proper azure config ${x}"))
       }
+      leoAuth <- samDAO.getLeoAuthToken
+      workspaceDescOpt <- wsmDao.getWorkspace(msg.workspaceId, leoAuth)
+      workspaceDesc <- F.fromOption(workspaceDescOpt, WorkspaceNotFoundException(msg.workspaceId, ctx.traceId))
+      storageContainerOpt <- wsmDao.getWorkspaceStorageContainer(msg.workspaceId, leoAuth)
```
rtitle: Maybe minor point -- but I think we're already passing the storage account + container in the `CreateAzureRuntimeMessage`? We could probably just construct the `workspaceStorageContainerUrl` with information we already have instead of making 3 new calls to WSM. Something like:

```
s"https://${params.landingZoneResources.storageAccountName.value}.blob.core.windows.net/${params.storageContainer.name.vale}"
```

See also: https://learn.microsoft.com/en-us/rest/api/storageservices/naming-and-referencing-containers--blobs--and-metadata#resource-uri-syntax

http/src/main/resources/init-resources/azure_vm_init_script.sh
```diff
@@ -99,3 +105,10 @@ docker run -d --restart always --network host --name welder \
...
+
```
rtitle: Also in addition to Julia probably want to update these guys:
```
jupyter@vm0186808a1:~$ jupyter kernelspec list
Available kernels:
  python3      /anaconda/envs/py38_default/share/jupyter/kernels/python3
  ir           /usr/local/share/jupyter/kernels/ir
  julia-1.6    /usr/local/share/jupyter/kernels/julia-1.6
```

rtitle:
Made some suggestions above, including how to (hopefully) get the changes to `azure_vm_init_script.sh` actually applied! Let me know of any questions.

Also I was poking and I _think_ the init script + output get placed in `/var/log/azure/custom-script/` directory on the DSVM. Unfortunately though that dir is owned by root so I can't actually look at it from JupyterLab as `jupyter` user. It is possible to ssh directly & sudo into these VMs. Also kind of wondering if we could `chmod -R r+jupyter /var/log/azure/custom-script` (or something similar) in the init script just to help with debugging this stuff. But just an idea, haven't tried.

http/src/main/resources/reference.conf
rtitle: Also this setup is a little odd, feels a bit error-prone and brittle to use direct references to git for the init script. There may be another way to do this (but separate from this PR).

http/src/main/resources/init-resources/azure_vm_init_script.sh
n8landolt: WORKSPACE_ID and WORKSPACE_STORAGE_ID are both already in the args for AzurePubsubHandler, I only added WORKSPACE_NAME and WORKSPACE_STORAGE_CONTAINER_URL, which dould be 16 and 17 in the args list.

http/src/main/resources/reference.conf
n8landolt: Huh. When I tried to do this before, I couldn't see my changes on the browser. It seemed like it was taking it's time to update the 'raw' file. 🤷 Seems fine now.

http/src/main/resources/init-resources/azure_vm_init_script.sh
rtitle: Ah, I see. Maybe could just rename the above variable to `WORKSPACE_ID` and reuse it? Same for `WORKSPACE_STORAGE_CONTAINER_ID`

n8landolt: Sounds good!

http/src/main/resources/reference.conf
```diff
@@ -194,7 +194,7 @@ azure {
         type = "CustomScript",
         version = "2.1",
         minor-version-auto-upgrade = true,
-        file-uris = ["https://raw.githubusercontent.com/DataBiosphere/leonardo/8dc8a0171fa4267197305eddddafb9dda8ce97fa/http/src/main/resources/init-resources/azure_vm_init_script.sh"]
+        file-uris = ["https://raw.githubusercontent.com/DataBiosphere/leonardo/IA-3957-include-additional-create-parameters/http/src/main/resources/init-resources/azure_vm_init_script.sh"]
```
n8landolt, line 197: What's best practice for keeping this line correct?
I know it shouldn't stay as-is. 
Do I find a permalink for the file that matches this PR? (Seems like historically this is what we do.)
Do we want to keep this reference.conf pointing to the most up-to-date version from dev?

n8landolt: Note - ConfigReaderSpec and WsmCodecSpec won't succeed as-is, they'll need updating to whatever we decide with this.

http/src/main/resources/init-resources/azure_vm_init_script.sh
```diff
@@ -42,13 +42,17 @@ CONTENTSECURITYPOLICY_FILE=$8
 
 # Envs for welder
 WELDER_WSM_URL=${9:-localhost}
-WELDER_WORKSPACE_ID="${10:-dummy}"
-WELDER_STORAGE_CONTAINER_RESOURCE_ID="${11:-dummy}"
+WORKSPACE_ID="${10:-dummy}"
```
LizBaldo: Why the name change here? If these variables are shared across services, not just Welder maybe we should rename everything? Or leave a comment explaining what this corresponds to?

http/src/main/resources/init-resources/azure_vm_init_script.sh
```diff
@@ -112,12 +116,22 @@ docker run -d --restart always --network host --name welder \
...
+--env STORAGE_CONTAINER_RESOURCE_ID=$WORKSPACE_STORAGE_CONTAINER_ID \
 --env STAGING_STORAGE_CONTAINER_RESOURCE_ID=$WELDER_STAGING_STORAGE_CONTAINER_RESOURCE_ID \
 --env OWNER_EMAIL=$WELDER_OWNER_EMAIL \
 --env CLOUD_PROVIDER="azure" \
 --env LOCKING_ENABLED=false \
 --env STAGING_BUCKET=$WELDER_STAGING_BUCKET \
 --env SHOULD_BACKGROUND_SYNC="false" \
 $WELDER_WELDER_DOCKER_IMAGE
+
+jq --null-input \
```
LizBaldo, line 130: Could you leave a comment explaining that this exposes the desired environment variables to the jupyter kernel?

http/src/main/resources/reference.conf
LizBaldo: My vote is to do the permalink from this feature branch only. It is easier to do here, and will be updated next time a dev makes a change to that script

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
```diff
@@ -111,8 +114,9 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
         params,
         auth
       )
-
       samResourceId = WsmControlledResourceId(UUID.fromString(params.runtime.samResource.resourceId))
+      wsStorageContainerUrl =
```
LizBaldo, line 124: Could you add a comment explaining what this change does?

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandlerSpec.scala
```diff
@@ -448,7 +469,14 @@ class AzurePubsubHandlerSpec
...
         }
 
-        msg = CreateAzureRuntimeMessage(runtime.id, workspaceId, storageContainerResourceId, landingZoneResources, None)
+        msg = CreateAzureRuntimeMessage(runtime.id,
+                                        workspaceId,
+                                        storageContainerResourceId,
+                                        landingZoneResources,
+                                        None,
+                                        "",
+                                        ContainerName("dummy")
```
LizBaldo, line 570: Why the empty Workspace name is this testing a specific edge case?

LizBaldo (commented):
LGTM! I am guessing that this is not really unit testable, correct?
Most of my comments are requests to add documentation for the next dev who might tackle these types of changes :)

http/src/main/resources/init-resources/azure_vm_init_script.sh
n8landolt: Rob suggested a name change over me declaring another variable with the same value.
I'm not sure if the other variables will be used for anything else, or are applicable. I could guess the email, bucket, and storage container resource id may be generic? 🤷

n8landolt: I'll add a comment, but I'm not sure how descriptive I can really be

LizBaldo (approved):
LGTM! Thanks for addressing my comments :)

n8landolt:
jenkens retest

n8landolt:
jenkins retest

n8landolt:
jenkins retest

PR #3169, opened 2023-02-16
[IA-3998] Multi-user V2 app creation

LizBaldo:
Jira ticket: https://broadworkbench.atlassian.net/browse/IA-3998

This PR is blocked by https://github.com/broadinstitute/sam/pull/989

---

In all cases:

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,10 @@ sbt "testOnly *LeoPubsubMessageSubscriberSpec -- -z "handle Azure StopRuntimeMes
 ```
 where `map` is a substring within the test name.
 
+If you made a change to the leonardo Db by adding a changeset xml file, and then adding that file path to the changelog
+file, you have to set `initWithLiquibase = true` in the leonardo.conf file for these changes to be reflected in the unit
+tests.
+
 Once you're done, tear down MySQL.
 ```
 ./docker/run-mysql.sh stop leonardo
--- a/README.md
+++ b/README.md
@@ -65,7 +65,8 @@ where `map` is a substring within the test name.
 
 If you made a change to the leonardo Db by adding a changeset xml file, and then adding that file path to the changelog
 file, you have to set `initWithLiquibase = true` in the leonardo.conf file for these changes to be reflected in the unit
-tests.
+tests. Once youare done testing your changes, make sure to switch it back to `initWithLiquibase = false`, as this can do
+some damage if you are running local Leo against Dev!
 
 Once you're done, tear down MySQL.
 ```
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodec.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodec.scala
@@ -96,8 +96,7 @@ object JsonCodec {
   implicit val formattedByEncoder: Encoder[FormattedBy] = Encoder.encodeString.contramap(_.asString)
   implicit val toolEncoder: Encoder[Tool] = Encoder.encodeString.contramap(_.asString)
   implicit val runtimeConfigTypeEncoder: Encoder[RuntimeConfigType] = Encoder.encodeString.contramap(_.asString)
-  implicit val accessScopeEncoder: Encoder[AppAccessScope] = Encoder.encodeString.contramap(_.toString)
-
+  implicit val appAccessScopeEncoder: Encoder[AppAccessScope] = Encoder.encodeString.contramap(_.toString)
 
   implicit val cloudContextEncoder: Encoder[CloudContext] = Encoder.forProduct2(
     "cloudProvider",
@@ -653,8 +652,8 @@ object JsonCodec {
   implicit val kubeClusterStatusDecoder: Decoder[KubernetesClusterStatus] = Decoder.decodeString.emap(s =>
     KubernetesClusterStatus.stringToObject.get(s).toRight(s"Invalid cluster status ${s}")
   )
-  implicit val accessScopeDecoder: Decoder[AppAccessScope] =
-    Decoder.decodeString.emap(s => AppAccessScope.stringToObject.get(s).toRight(s"Invalid accessScope ${s}"))
+  implicit val appAccessScopeDecoder: Decoder[AppAccessScope] =
+    Decoder.decodeString.emap(s => AppAccessScope.stringToObject.get(s).toRight(s"Invalid app accessScope ${s}"))
   implicit val appSamIdDecoder: Decoder[AppSamResourceId] =
     Decoder.forProduct2("resourceId", "accessScope")(AppSamResourceId.apply)
 
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodec.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodec.scala
@@ -655,7 +655,7 @@ object JsonCodec {
   implicit val appAccessScopeDecoder: Decoder[AppAccessScope] =
     Decoder.decodeString.emap(s => AppAccessScope.stringToObject.get(s).toRight(s"Invalid app accessScope ${s}"))
   implicit val appSamIdDecoder: Decoder[AppSamResourceId] =
-    Decoder.forProduct2("resourceId", "accessScope")(AppSamResourceId)
+    Decoder.forProduct2("resourceId", "accessScope")(AppSamResourceId.apply)
 
   implicit val namespaceNameDecoder: Decoder[NamespaceName] =
     Decoder.decodeString.emap(s => KubernetesName.withValidation(s, NamespaceName).leftMap(_.getMessage))
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodec.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodec.scala
@@ -96,6 +96,8 @@ object JsonCodec {
   implicit val formattedByEncoder: Encoder[FormattedBy] = Encoder.encodeString.contramap(_.asString)
   implicit val toolEncoder: Encoder[Tool] = Encoder.encodeString.contramap(_.asString)
   implicit val runtimeConfigTypeEncoder: Encoder[RuntimeConfigType] = Encoder.encodeString.contramap(_.asString)
+  implicit val accessScopeEncoder: Encoder[AppAccessScope] = Encoder.encodeString.contramap(_.toString)
+
 
   implicit val cloudContextEncoder: Encoder[CloudContext] = Encoder.forProduct2(
     "cloudProvider",
@@ -277,7 +279,8 @@ object JsonCodec {
       case _                                   => status.toString
     }
   )
-  implicit val kubeSamIdEncoder: Encoder[AppSamResourceId] = Encoder.encodeString.contramap(_.resourceId)
+  implicit val kubeSamIdEncoder: Encoder[AppSamResourceId] =
+    Encoder.forProduct2("resourceId", "accessScope")(x => AppSamResourceId.unapply(x).get)
   implicit val namespaceEncoder: Encoder[NamespaceName] = Encoder.encodeString.contramap(_.value)
   implicit val appNameEncoder: Encoder[AppName] = Encoder.encodeString.contramap(_.value)
   implicit val appStatusEncoder: Encoder[AppStatus] = Encoder.encodeString.contramap { x =>
@@ -650,6 +653,8 @@ object JsonCodec {
   implicit val kubeClusterStatusDecoder: Decoder[KubernetesClusterStatus] = Decoder.decodeString.emap(s =>
     KubernetesClusterStatus.stringToObject.get(s).toRight(s"Invalid cluster status ${s}")
   )
+  implicit val accessScopeDecoder: Decoder[AppAccessScope] =
+    Decoder.decodeString.emap(s => AppAccessScope.stringToObject.get(s).toRight(s"Invalid accessScope ${s}"))
   implicit val appSamIdDecoder: Decoder[AppSamResourceId] =
     Decoder.forProduct2("resourceId", "accessScope")(AppSamResourceId.apply)
 
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodec.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodec.scala
@@ -650,7 +650,9 @@ object JsonCodec {
   implicit val kubeClusterStatusDecoder: Decoder[KubernetesClusterStatus] = Decoder.decodeString.emap(s =>
     KubernetesClusterStatus.stringToObject.get(s).toRight(s"Invalid cluster status ${s}")
   )
-  implicit val appSamIdDecoder: Decoder[AppSamResourceId] = Decoder.decodeString.map(AppSamResourceId)
+  implicit val appSamIdDecoder: Decoder[AppSamResourceId] =
+    Decoder.forProduct2("resourceId", "accessScope")(AppSamResourceId.apply)
+
   implicit val namespaceNameDecoder: Decoder[NamespaceName] =
     Decoder.decodeString.emap(s => KubernetesName.withValidation(s, NamespaceName).leftMap(_.getMessage))
   implicit val namespaceIdDecoder: Decoder[NamespaceId] = Decoder.decodeLong.map(NamespaceId)
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodec.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodec.scala
@@ -655,7 +655,7 @@ object JsonCodec {
   implicit val appAccessScopeDecoder: Decoder[AppAccessScope] =
     Decoder.decodeString.emap(s => AppAccessScope.stringToObject.get(s).toRight(s"Invalid app accessScope ${s}"))
   implicit val appSamIdDecoder: Decoder[AppSamResourceId] =
-    Decoder.forProduct2("resourceId", "accessScope")(AppSamResourceId.apply)
+    Decoder.forProduct2("resourceId", "accessScope")(AppSamResourceId)
 
   implicit val namespaceNameDecoder: Decoder[NamespaceName] =
     Decoder.decodeString.emap(s => KubernetesName.withValidation(s, NamespaceName).leftMap(_.getMessage))
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodec.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodec.scala
@@ -278,8 +278,7 @@ object JsonCodec {
       case _                                   => status.toString
     }
   )
-  implicit val kubeSamIdEncoder: Encoder[AppSamResourceId] =
-    Encoder.forProduct2("resourceId", "accessScope")(x => AppSamResourceId.unapply(x).get)
+  implicit val appSamIdEncoder: Encoder[AppSamResourceId] = Encoder.encodeString.contramap(_.resourceId)
   implicit val namespaceEncoder: Encoder[NamespaceName] = Encoder.encodeString.contramap(_.value)
   implicit val appNameEncoder: Encoder[AppName] = Encoder.encodeString.contramap(_.value)
   implicit val appStatusEncoder: Encoder[AppStatus] = Encoder.encodeString.contramap { x =>
@@ -654,8 +653,7 @@ object JsonCodec {
   )
   implicit val appAccessScopeDecoder: Decoder[AppAccessScope] =
     Decoder.decodeString.emap(s => AppAccessScope.stringToObject.get(s).toRight(s"Invalid app accessScope ${s}"))
-  implicit val appSamIdDecoder: Decoder[AppSamResourceId] =
-    Decoder.forProduct2("resourceId", "accessScope")(AppSamResourceId.apply)
+  implicit val appSamIdDecoder: Decoder[AppSamResourceId] = Decoder.decodeString.map(s => AppSamResourceId(s, None))
 
   implicit val namespaceNameDecoder: Decoder[NamespaceName] =
     Decoder.decodeString.emap(s => KubernetesName.withValidation(s, NamespaceName).leftMap(_.getMessage))
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/appRoutesModels.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/appRoutesModels.scala
@@ -2,7 +2,22 @@ package org.broadinstitute.dsde.workbench.leonardo.http
 
 import org.broadinstitute.dsde.workbench.google2.DiskName
 import org.broadinstitute.dsde.workbench.google2.KubernetesSerializableName.ServiceName
-import org.broadinstitute.dsde.workbench.leonardo.{App, AppAccessScope, AppError, AppName, AppStatus, AppType, AuditInfo, CloudContext, CloudProvider, KubernetesCluster, KubernetesRuntimeConfig, LabelMap, Nodepool, WorkspaceId}
+import org.broadinstitute.dsde.workbench.leonardo.{
+  App,
+  AppAccessScope,
+  AppError,
+  AppName,
+  AppStatus,
+  AppType,
+  AuditInfo,
+  CloudContext,
+  CloudProvider,
+  KubernetesCluster,
+  KubernetesRuntimeConfig,
+  LabelMap,
+  Nodepool,
+  WorkspaceId
+}
 import org.http4s.Uri
 
 import java.net.URL
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/appRoutesModels.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/appRoutesModels.scala
@@ -42,6 +42,7 @@ final case class GetAppResponse(appName: AppName,
                                 customEnvironmentVariables: Map[String, String],
                                 auditInfo: AuditInfo,
                                 appType: AppType,
+                                accessScope: Option[AppAccessScope],
                                 labels: LabelMap
 )
 
@@ -56,6 +57,7 @@ final case class ListAppResponse(cloudProvider: CloudProvider,
                                  appType: AppType,
                                  diskName: Option[DiskName],
                                  auditInfo: AuditInfo,
+                                 accessScope: Option[AppAccessScope],
                                  labels: LabelMap
 )
 
@@ -81,6 +83,7 @@ object ListAppResponse {
           a.appType,
           a.appResources.disk.map(_.name),
           a.auditInfo,
+          a.appAccessScope,
           a.labels.filter(l => labelsToReturn.contains(l._1))
         )
       }
@@ -104,6 +107,7 @@ object GetAppResponse {
       appResult.app.customEnvironmentVariables,
       appResult.app.auditInfo,
       appResult.app.appType,
+      appResult.app.appAccessScope,
       appResult.app.labels
     )
 }
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/appRoutesModels.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/appRoutesModels.scala
@@ -2,27 +2,14 @@ package org.broadinstitute.dsde.workbench.leonardo.http
 
 import org.broadinstitute.dsde.workbench.google2.DiskName
 import org.broadinstitute.dsde.workbench.google2.KubernetesSerializableName.ServiceName
-import org.broadinstitute.dsde.workbench.leonardo.{
-  App,
-  AppError,
-  AppName,
-  AppStatus,
-  AppType,
-  AuditInfo,
-  CloudContext,
-  CloudProvider,
-  KubernetesCluster,
-  KubernetesRuntimeConfig,
-  LabelMap,
-  Nodepool,
-  WorkspaceId
-}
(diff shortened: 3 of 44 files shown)


core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodec.scala
```diff
@@ -633,7 +633,9 @@ object JsonCodec {
   implicit val kubeClusterStatusDecoder: Decoder[KubernetesClusterStatus] = Decoder.decodeString.emap(s =>
     KubernetesClusterStatus.stringToObject.get(s).toRight(s"Invalid cluster status ${s}")
   )
-  implicit val appSamIdDecoder: Decoder[AppSamResourceId] = Decoder.decodeString.map(AppSamResourceId)
+  implicit val appSamIdDecoder: Decoder[AppSamResourceId] =
+    Decoder.forProduct2("resourceId", "accessScope")(AppSamResourceId.apply)
```
LizBaldo: sbt is yelling at me because of:

```
could not find implicit value for parameter decodeA1: io.circe.Decoder[Option[org.broadinstitute.dsde.workbench.leonardo.AppAccessScope]]
Decoder.forProduct2("resourceId", "accessScope")(AppSamResourceId.apply)
```

Fixing this depends on whether my new implementation of the AppSamResourceId makes sense

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/appRoutesModels.scala
```diff
@@ -2,27 +2,14 @@ package org.broadinstitute.dsde.workbench.leonardo.http
...
-  WorkspaceId
-}
+import org.broadinstitute.dsde.workbench.leonardo.{App, AppAccessScope, AppError, AppName, AppStatus, AppType, AuditInfo, CloudContext, CloudProvider, KubernetesCluster, KubernetesRuntimeConfig, LabelMap, Nodepool, WorkspaceId}
 import org.http4s.Uri
 
 import java.net.URL
 
 final case class CreateAppRequest(kubernetesRuntimeConfig: Option[KubernetesRuntimeConfig],
                                   appType: AppType,
+                                  accessScope: Option[AppAccessScope],
```
LizBaldo, line 27: Since `CreateAppRequest` is shared between the V1 and V2 endpoints, I chose to keep `accessScope` optional until the Sam V2 Resource creation

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/samModels.scala
```diff
@@ -24,8 +24,12 @@ object SamResourceId {
     override def resourceType: SamResourceType = SamResourceType.Project
   }
 
-  final case class AppSamResourceId(resourceId: String) extends SamResourceId {
-    override def resourceType: SamResourceType = SamResourceType.App
+  final case class AppSamResourceId(resourceId: String, accessScope: Option[AppAccessScope]) extends SamResourceId {
+    override def resourceType: SamResourceType = accessScope match {
+      case Some(AppAccessScope.UserPrivate)     => SamResourceType.App
+      case Some(AppAccessScope.WorkspaceShared) => SamResourceType.SharedApp
+      case None                                 => SamResourceType.App
```
LizBaldo, line 31: This is where I chose to enforce the default value for the accessScope. My main reason for having this logic in here is because I did not want to carry along two different classes for the app and shared app. Doing this makes the rest a lot nicer.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpSamDAO.scala
```diff
@@ -256,6 +256,46 @@ class HttpSamDAO[F[_]](httpClient: Client[F],
...
+                                           workspaceId: WorkspaceId
+  )(implicit
+    sr: SamResource[R],
+    encoder: Encoder[R],
+    ev: Ask[F, TraceId]
+  ): F[Unit] =
+    for {
+      traceId <- ev.ask
+      // TODO: Check if this is the best way to get the token, there are no petSA for Workspace resources, only google-projects
+      authHeader = Authorization(Credentials.Token(AuthScheme.Bearer, userInfo.accessToken.token))
```
LizBaldo, line 275: I went down a rabbit hole in the SAM API documentation, but could not find a better way to get the token, am I missing something?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/AppV2Routes.scala
```diff
@@ -169,13 +169,15 @@ object AppV2Routes {
...
         a <- x.downField("appType").as[Option[AppType]]
+        s <- x.downField("accessScope").as[Option[AppAccessScope]]
         d <- x.downField("diskConfig").as[Option[PersistentDiskRequest]]
         l <- x.downField("labels").as[Option[LabelMap]]
         cv <- x.downField("customEnvironmentVariables").as[Option[LabelMap]]
         dp <- x.downField("descriptorPath").as[Option[Uri]]
         ea <- x.downField("extraArgs").as[Option[List[String]]]
       } yield CreateAppRequest(c,
                                a.getOrElse(AppType.Galaxy),
+                               s,
```
LizBaldo, line 180: Not enforcing the default here as I need to do it anyways in `AppSamResourceId` for the V1 apps

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
```diff
@@ -103,7 +103,8 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
         F.raiseError[Unit](AppAlreadyExistsException(cloudContext, appName, c.app.status, ctx.traceId))
       )
 
-      samResourceId <- F.delay(AppSamResourceId(UUID.randomUUID().toString))
+      // Shared app SAM resource is not enabled for V1 endpoint
```
LizBaldo, line 106: I could leave the accessScope as a None here.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeoAuthProvider.scala
```diff
@@ -103,6 +110,15 @@ object SamResourceAction {
     def actionAsString(a: AppAction): String = a.asString
   }
 
+  implicit def SharedAppSamResourceAction = new SharedAppSamResource
```
LizBaldo: I have to admit that I am not sure what this does... It does not seem to be used anywhere else (but I keep getting burned by these implicits)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -369,7 +370,7 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
         raw"relaylistener.targetHost=http://coa-${release.asString}-reverse-proxy-service:8000/",
         raw"relaylistener.samUrl=${config.samConfig.server}",
         raw"relaylistener.samResourceId=${samResourceId.resourceId}",
-        raw"relaylistener.samResourceType=kubernetes-app",
+        raw"relaylistener.samResourceType=${samResourceId.resourceType}",
```
LizBaldo: either `kubernetes-app` or `kubernetes-app-shared`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpSamDAO.scala
LizBaldo: Fine as long as called synchronously, but maybe use this instead `getCachedArbitraryPetAccessToken`

LizBaldo: Do this to stay consistent with the Azure resources

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodec.scala
LizBaldo: Need implicit decoder for that type

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/appRoutesModels.scala
rtitle: Makes sense - we should default to the existing behavior (of private apps) if `accessScope` is not passed.

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -1599,6 +1599,13 @@ paths:
           required: true
           schema:
             type: string
+        - in: query
```
rtitle: We probably want to have `accessScope` in the `CreateAppRequest` object in the request body, instead of in the query string.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpSamDAO.scala
rtitle: Thinking more, I feel like using `UserInfo` is fine (e.g. how it is now). The other branch in `SamAuthProvider.notifyResourceCreatedV2` calls `createResourceWithUserInfo` so we're already using `UserInfo` for runtime creations. Doing similar for apps would avoid an extra Sam hop. If we ever need to do this asynchronously we can revisit.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpSamDAO.scala
```diff
@@ -256,6 +256,46 @@ class HttpSamDAO[F[_]](httpClient: Client[F],
...
+      loggingCtx = Map("traceId" -> traceId.asString)
+      _ <- logger.info(loggingCtx)(
+        s"Creating ${sr.resourceType.asString} resource in sam v2 for ${workspaceId}/${sr.resourceIdAsString(resource)}"
+      )
+      _ <- metrics.incrementCounter(s"sam/createResource/${sr.resourceType.asString}")
+      policies = Map[SamPolicyName, SamPolicyData](
+        SamPolicyName.Creator -> SamPolicyData(List(creatorEmail), List(SamRole.Creator))
+      )
+      parent = SerializableSamResource(SamResourceType.Workspace, WorkspaceResourceSamResourceId(workspaceId))
```
rtitle, line 284: Cool - so this method + logic works for both `kubernetes-app` and `kubernetes-app-shared` resource types, correct?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/SamDAO.scala
```diff
@@ -70,7 +70,19 @@ trait SamDAO[F[_]] {
 
   /** Creates a Sam resource R with the provided google project as the parent resource using a GCP pet credential
      for the given email/project. */
-  def createResourceWithParent[R](resource: R, creatorEmail: WorkbenchEmail, googleProject: GoogleProject)(implicit
+  def createResourceWithGcpParent[R](resource: R, creatorEmail: WorkbenchEmail, googleProject: GoogleProject)(implicit
```
rtitle: Nit - wonder if `createResourceWithGoogleProjectParent` is slightly more descriptive.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
rtitle: Wonder if we should throw a 400 Bad Request if accessScope is passed to the v1 endpoint?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
```diff
@@ -103,7 +103,8 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
...
+      samResourceId <- F.delay(AppSamResourceId(UUID.randomUUID().toString, AppAccessScope.UserPrivate))
```
rtitle: Nit - can use `F.pure` for pure values

rtitle: Er nevermind - it's probably not pure due to `randomUUID()`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpSamDAO.scala
LizBaldo: Sounds good, I will leave a comment in the code to explain

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
LizBaldo: good idea 👍

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpSamDAO.scala
LizBaldo: Yep!

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeoAuthProvider.scala
rtitle: These implicits maybe go overboard a bit. I think basically it's trying to tie together Sam resource types + actions at compile time. So it's impossible to call `hasResourcePermission(resource, action)` for an action that doesn't match the resource type (e.g. it doesn't compile).

… Which is kinda cool, but at the same time I struggle to remember how this code works (and I wrote it!), so it feels like the complexity isn't really worth it.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
rtitle: 💯

rtitle (commented):
Just a few nits -- but overall the approach looks good! It seems like this will do the right thing.

http/src/main/resources/swagger/api-docs.yaml
LizBaldo: What do you mean by that? The swagger doc points towards the `query` option https://swagger.io/docs/specification/describing-parameters/

rtitle: With `query`, the accessScope param would be in the query string:
```
POST /api/apps/v2/{workspaceId}/{appName}?accessScope={accessScope}
{
  "appType": "CROMWELL",
  …
}
```

But our normal pattern for POSTs is to put fields in the JSON body:
```
POST /api/apps/v2/{workspaceId}/{appName}
{
  "appType": "CROMWELL",
  "accessScope": "shared",
  …
}
```

I think this would mean updating the `CreateAppRequest` object schema [here](https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/resources/swagger/api-docs.yaml#L3700) instead.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeoAuthProvider.scala
LizBaldo: I ended up modifying the SamResource object to have ResourceType and SamPolicies as functions instead of values. This makes things more flexible, and avoid having to pass around the `accessScope` parameter

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/ProxyService.scala
```diff
@@ -194,16 +227,21 @@ class ProxyService(
...
     for {
       ctx <- ev.ask[AppContext]
-      cacheResult <- samResourceCache.cachingF(key)(None)(
+      cacheSamResourceId <- samResourceCache.cachingF(key)(None)(
         getSamResourceFromDb(key)
       )
-      resourceId = cacheResult.map(AppSamResourceId)
+      // TODO modify the cache to store more than just the App Resource ID string
+      // TODO REENABLE CACHE IMMEDIATELY AFTER FIGURING OUT THE ABOVE
+      cacheAppAccessScopeString <- getSamAppAccessScopeFromDb(key)
```
LizBaldo: This is an issue I ran into because of how the cache is being mocked for unit tests. The cache can only contain one string the resource id, but we also need to retrieve the access scope as well now. Feels like this opens the door to unit tests doing some IO stuff, which we should avoid. i would love some help on refactoring the cache :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeoAuthProvider.scala
```diff
@@ -19,39 +19,44 @@ import org.broadinstitute.dsde.workbench.model.{TraceId, UserInfo, WorkbenchEmai
 
 // Typeclass representing a Sam resource and associated policies
 sealed trait SamResource[R] {
-  def resourceType: SamResourceType
+  def resourceType(r: R): SamResourceType
```
LizBaldo, line 22: By moving from val to def, we avoid having to pass around the accessScope parameter. It is way cleaner this way and make the implicits happy!

README.md
```diff
@@ -63,6 +63,10 @@ sbt "testOnly *LeoPubsubMessageSubscriberSpec -- -z "handle Azure StopRuntimeMes
 ```
 where `map` is a substring within the test name.
 
+If you made a change to the leonardo Db by adding a changeset xml file, and then adding that file path to the changelog
+file, you have to set `initWithLiquibase = true` in the leonardo.conf file for these changes to be reflected in the unit
```
rtitle, line 67: Just be careful enabling this if running local Leo against the dev DB 😅

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/SamAuthProvider.scala
```diff
@@ -124,10 +136,11 @@ class SamAuthProvider[F[_]: OpenTelemetryMetrics](
     for {
       ctx <- ev.ask
 
-      resourcePolicies <- samDao
-        .getResourcePolicies[R](authHeader)
-      res = resourcePolicies.filter { case (_, pn) =>
-        sr.policyNames.contains(pn)
+      resourcePolicies <- resources.toList.flatTraverse(resource =>
```
rtitle: It looks like this is now calling Sam `getResourcePolicies` once per resource which doesn't seem correct -- seems we want to call it once per resource _type_ ? Then aggregate the results & filter.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/SamAuthProvider.scala
```diff
@@ -142,13 +155,17 @@ class SamAuthProvider[F[_]: OpenTelemetryMetrics](
...
     val authHeader = Authorization(Credentials.Token(AuthScheme.Bearer, userInfo.accessToken.token))
     for {
-      projectPolicies <- samDao.getResourcePolicies[ProjectSamResourceId](authHeader)
-      owningProjects = projectPolicies.collect { case (r, SamPolicyName.Owner) =>
-        r.googleProject
-      }
-      resourcePolicies <- samDao
-        .getResourcePolicies[R](authHeader)
-      res = resourcePolicies.filter { case (_, pn) => sr.policyNames.contains(pn) }
+      projectPolicies <- resources.toList.flatTraverse(resource =>
```
rtitle: Yeah same as above -- unless I'm missing something it feels like this will result in much more Sam queries since we're calling `/api/resources/v2/{resourceType}` for each resource passed to this method, instead of only once. Maybe the code needs some restructuring to deal with the fact that there may be multiple resource types handled by this method..

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/SamAuthProvider.scala
LizBaldo: Oh shoot, you are completely right! I was so focused on getting the compiler to work that I missed this, sorry!

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/LeoProfile.scala
```diff
@@ -224,8 +224,13 @@ private[leonardo] object LeoProfile extends MySQLProfile {
...
     implicit val appNameColumnType: BaseColumnType[AppName] =
       MappedColumnType.base[AppName, String](_.value, AppName.apply)
+    implicit val appAccessScopeColumnType: BaseColumnType[AppAccessScope] =
+      MappedColumnType.base[AppAccessScope, String](
+        _.toString,
+        s => AppAccessScope.stringToObject.getOrElse(s, throw ColumnDecodingException(s"invalid app access scope ${s}"))
+      )
     implicit val appSamResourceIdColumnType: BaseColumnType[AppSamResourceId] =
-      MappedColumnType.base[AppSamResourceId, String](_.resourceId, AppSamResourceId.apply)
+      MappedColumnType.base[AppSamResourceId, String](_.resourceId, s => AppSamResourceId.apply(s, None))
```
rtitle, line 233: Hm, is it an issue that we're deserializing `AppAccessScope` to `None` here? I guess maybe this field isn't used in the DB layer?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/AppComponent.scala
```diff
@@ -103,11 +106,12 @@ object appQuery extends TableQuery(new AppTable(_)) {
       app.nodepoolId,
       app.appType,
       app.appName,
+      app.appAccessScope,
       app.workspaceId,
       app.status,
       app.chart,
       app.release,
-      app.samResourceId,
+      AppSamResourceId(app.samResourceId.resourceId, app.appAccessScope),
```
rtitle, line 114: Ah re my comment in `LeoProfile`, I guess it works because the `AppSamResourceId` is copied here. 👍

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/ProxyService.scala
rtitle: As discussed earlier, we can probably update the cache value to contain multiple fields as `(String, String)` or something.

rtitle (commented):
Some comments -- overall approach looks good, some comments in the Sam layer.

Also was checking Jenkins automation tests and see errors:
```
message: Failed to create app terra-quality-ca740cf8/automation-test-app-ad4cyerfz, status: 400 Bad Request body: {"source":"leonardo","message":"AuthProvider error: The request content was malformed:\ncould not deserialize value object: {\"accessScope\":null,\"resourceId\":\"438f5694-530b-4252-8142-2b3f0ee8a5bd\"}","statusCode":400,"exceptionClass":"class org.broadinstitute.dsde.workbench.leonardo.dao.AuthProviderException","traceId":"93a372c6b71a4b518851437a89b3ba41/3939911508519804487"}
```

There may be some codec that still needs updating..

README.md
LizBaldo: Added a little comment to switch it back to false after the test and a cautionary tale re running against Dev 😬

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/LeoProfile.scala
LizBaldo: I could not figure out a better way to grab the ID string out of the AppSamResourceId object, this seems to work, but lmk if you know a more elegant handling

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/AppComponent.scala
LizBaldo: Yep! I can definitely add a little unit test to double check!

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/ProxyService.scala
LizBaldo: Done, the cache now contains a tuple with both the string id and the appAccessScope :)

LizBaldo:
@rtitle, thanks for taking over from me while I am on PTO! I think this PR is very, very close, but I can't get the createAppRequest to decode properly (error that you linked in your comment above). I suspect it that [this decoder](https://github.com/DataBiosphere/leonardo/blob/fef2d02559ff6dc12b028fa4094de53ad41da75c/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodec.scala#L657) needs to be updated, but I was not able to figure it out in time, I am so sorry!

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/SamAuthProvider.scala
```diff
@@ -141,14 +154,22 @@ class SamAuthProvider[F[_]: OpenTelemetryMetrics](
...
+    val resourceTypes = resources.map(r => sr.resourceType(r._2)).toList.toSet
     for {
-      projectPolicies <- samDao.getResourcePolicies[ProjectSamResourceId](authHeader)
-      owningProjects = projectPolicies.collect { case (r, SamPolicyName.Owner) =>
-        r.googleProject
-      }
-      resourcePolicies <- samDao
-        .getResourcePolicies[R](authHeader)
-      res = resourcePolicies.filter { case (_, pn) => sr.policyNames.contains(pn) }
+      projectPolicies <- resourceGoogleProjects.toList.flatTraverse(resourceGoogleProject =>
```
rtitle: I think this is still making too many Sam queries -- it's calling Sam for every googleProject resource in the input list, but really it should only be making at most 3 Sam calls:

* 1 for google-project resource type
* 1 for kubernetes-app resource type (if any exist in the input list)
* 1 for kubernetes-app-shared resource type (if any exist in the input list)

(I think for apps the code is correct, just missing the logic for google-project.)

This may seem minor but could have a real impact if a large `listApps` query ends up pounding Sam.

@LizBaldo really no need to address this tonight -- I can try to push a commit later/tomorrow & see this though.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/SamAuthProvider.scala
```diff
@@ -162,7 +183,9 @@ class SamAuthProvider[F[_]: OpenTelemetryMetrics](
   ): F[Set[WorkspaceResourceSamResourceId]] = {
     val authHeader = Authorization(Credentials.Token(AuthScheme.Bearer, userInfo.accessToken.token))
     for {
-      workspacePolicies <- samDao.getResourcePolicies[WorkspaceResourceSamResourceId](authHeader)
+      workspacePolicies <- resources.toList.flatTraverse(resource =>
```
rtitle: Same here, I think -- should just make 1 Sam query for workspace resource type

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/SamAuthProvider.scala
rtitle: Er, this method is not apps-specific but my overall comment stands  -- it should be calling Sam once for `google-project` resource type and once for each distinct resource type in the set of input resources (of type `R`).

rtitle: Actually wasn't too bad, pushed a commit

rtitle:
OK will check out the decoders

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/SamAuthProvider.scala
LizBaldo: Ohh that makes a lot of sense, I was trying to generalize and be fancy, but might as well be explicit

LizBaldo:
Thanks @rtitle 🙏 , taking one last glance, and I might ping you once I am back about serializing AppSamResourceId, I don't think I was understanding how it was supposed to behave properly 😬

rtitle:
Only `RuntimeAutopauseSpec` failed after the latest change, which seems unrelated.

However I was thinking about this and probably want to make a couple additional changes before merge:

1. For shared apps, don't link the user's pet identity via pod-managed-identity. The original [doc](https://docs.google.com/document/d/1Qqljdgnzojca-8xVdpYZaEWBZs0htf59JA5Yfbeg8po/edit) goes into this a bit. I think we probably still need to design which identity we _do_ want to link; but for now we can probably just not link anything and have it default to a system identity (with no permissions).

2. Allow-list app types which can launch shared. I'm not certain we want to enable this across the board yet. E.g. Cromwell probably shouldn't be launchable as a shared app. Thinking we can have an allow-list, and add the WDS-only chart when it's available.

I can work on these changes.

rtitle:
I threw this on a BEE and created a cromwell-app (in private mode). Unfortunately it failed to spin up because all proxy endpoints were responding 403. 😭 

Need to investigate more..

EDIT: hopefully [fixed the bug](https://github.com/DataBiosphere/leonardo/pull/3169/commits/cf2ec1ae2fbc72b68c60713398b42fec2751ac9c), retesting..

rtitle:
OK I managed to verify this in a BEE!

1. Created user-private app; verified only creator can use it (same behavior as today)
2. Created workspace-shared app. App creator and a workspace reader can view/connect to the app. Only creator can delete.
3. Verified shared app has no pod-identity linking.
4. Verified app type not on the allow list cannot be launched in shared mode.

Approving this; will verify tests and maybe do 1 more review pass, then merge.

rtitle (approved):

cahrens:
@LizBaldo I believe you have a typo in the test package name-- "dede" instead of "dsde".

LizBaldo:
@cahrens I am a bit confused sorry, this is a pretty old PR and does not use a test package, I will dm you

PR #3174, opened 2023-02-21
update VM status to deleted only when WSM job succeeds

lucymcnatt:
When a delete azure runtime calls that error on first try, on the second try the request appears to go through fine, but the VM is still running in the Azure portal, change moves the status update to 'deleted' only when the WSM call executes successfully so that the status stays in the errored state for the second call.

relevant [slack thread]( https://broadinstitute.slack.com/archives/GB1VAM14J/p1676671086876319 )

---

In all cases:


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
```diff
@@ -576,6 +573,8 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
           case Some(resp) =>
             resp.jobReport.status match {
               case WsmJobStatus.Succeeded =>
+                dbRef.inTransaction(clusterQuery.updateClusterStatus(runtime.id, RuntimeStatus.Deleted, ctx.now))
```
rtitle: So these are all `IO`s, which need to be chained together to form a single `IO` which is returned by this `match` statement. Otherwise they won't run.

So need something like:
```
dbRef.inTransaction(..) >> logger.info(..) >> deleteDiskAction
```

or

```
for {
  _ <- dbRef.inTransaction(..)
  _ <- logger.info(..)
  _ <- deleteDiskAction
yield ()
```

… which are both syntax for the same thing.

lucymcnatt: ohhh yeah gotcha, I'll fix that

rtitle (commented):
The change looks good (once Scala issue is fixed) -- nice find.

So if I'm understanding correctly, now when a deletion request fails, a user will see the error message and the runtime will be moved to Error status. The user can then retry the deletion again.

Maybe in parallel we can look at making the Leo->WSM calls more reliable to reduce frequency of these errors -- but most important to not leak resources. 👍

lucymcnatt:
Yes yeah, before if a deletion request failed, the cluster status would still be updated to `deleted`, but it will now only be `deleted` if the WSM job succeeds. Then on a second request if the first errors, the cluster status should be `error` and the deleteVm WSM call should execute properly.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
LizBaldo: for comprehensions for the win!

LizBaldo (commented):
That makes sense to me! Happy to. 👍 once the unit tests are passing

lucymcnatt:
yeah I've got the change switched to a for comp, just have to fix a couple failing tests

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
```diff
@@ -576,7 +573,11 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
           case Some(resp) =>
             resp.jobReport.status match {
               case WsmJobStatus.Succeeded =>
-                deleteDiskAction
+                for {
+                  _ <- dbRef.inTransaction(clusterQuery.updateClusterStatus(runtime.id, RuntimeStatus.Deleted, ctx.now))
+                  _ <- logger.info(ctx.loggingCtx)("runtime is deleted successfully")
```
rtitle: While we're touching this maybe improve the log msg? :) 
```
s"Runtime ${msg.runtimeId} is deleted successfully"
```

or something with more breadcrumbs..

rtitle (approved):
:shipit:

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
lucymcnatt: good idea

lucymcnatt:
jenkins retest

lucymcnatt:
jenkins retest

PR #3183, opened 2023-02-27
[IA-4089] Disable custom AppLifeCycle automation test

LizBaldo:
https://broadworkbench.atlassian.net/browse/IA-4089

---

In all cases:

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/apps/AppLifecycleSpec.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/apps/AppLifecycleSpec.scala
@@ -62,23 +62,24 @@ class AppLifecycleSpec
     )
   )
 
-  // Test custom app first so that there will be a GKE cluster created already for the next two tests
-  "create CUSTOM app, start/stop, delete it" in { googleProject =>
-    test(
-      googleProject,
-      createAppRequest(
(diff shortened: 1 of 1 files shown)


rtitle (approved):

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/apps/AppLifecycleSpec.scala
```diff
@@ -62,23 +62,24 @@ class AppLifecycleSpec
     )
   )
 
-  // Test custom app first so that there will be a GKE cluster created already for the next two tests
```
rtitle, line 65: Hm, were we relying on a side effect of this test to create a GKE cluster before the subsequent (parallel) tests?

LizBaldo: Yeah, I was just thinking about that, I might make these two tests sequential instead of in parallel, but i was curious to see if this woks as is

LizBaldo: turns out it worked! Merging as is since I will have a follow up PR shortly reenabling the test with the new image anyways

PR #3188, opened 2023-02-28
IA-4069: try again to fix connection closed errors, and fix WSM error handling

rtitle:
https://broadworkbench.atlassian.net/browse/IA-4069 redux

See inline comments. Planning to do some more BEE-testing, but wanted to get PR out.

---

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/Boot.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/Boot.scala
@@ -71,7 +71,7 @@ import org.typelevel.log4cats.StructuredLogger
 import org.typelevel.log4cats.slf4j.Slf4jLogger
 import scalacache.caffeine._
 
-import java.net.InetSocketAddress
+import java.net.{InetSocketAddress, SocketException}
 import java.nio.file.Paths
 import java.time.Instant
 import java.util.concurrent.TimeUnit
@@ -741,7 +741,16 @@ object Boot extends IOApp {
     metricsPrefix: Option[String],
     withRetry: Boolean
   ): Resource[F, org.http4s.client.Client[F]] = {
-    val retryPolicy = RetryPolicy[F](RetryPolicy.exponentialBackoff(30 seconds, 5))
+    // Retry all SocketExceptions to deal with pooled HTTP connections getting closed on the server-side.
+    // See https://broadworkbench.atlassian.net/browse/IA-4069.
+    val retryPolicy = RetryPolicy[F](
+      RetryPolicy.exponentialBackoff(30 seconds, 5),
+      (req, result) =>
+        result match {
+          case Left(e) if e.isInstanceOf[SocketException] => true
+          case _                                          => RetryPolicy.defaultRetriable(req, result)
+        }
+    )
 
     for {
       httpClient <- client
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/Boot.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/Boot.scala
@@ -754,6 +754,7 @@ object Boot extends IOApp {
         .withConnectTimeout(30 seconds)
         .withRequestTimeout(60 seconds)
         .withMaxTotalConnections(100)
+        .withMaxIdleDuration(30 seconds)
         .resource
       httpClientWithLogging = Http4sLogger[F](logHeaders = true, logBody = false, logAction = Some(s => logAction(s)))(
         httpClient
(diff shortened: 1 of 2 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/Boot.scala
```diff
@@ -754,6 +754,7 @@ object Boot extends IOApp {
         .withConnectTimeout(30 seconds)
         .withRequestTimeout(60 seconds)
         .withMaxTotalConnections(100)
+        .withMaxIdleDuration(30 seconds)
```
rtitle, line 766: Doc for this field:
```
  * @param maxIdleDuration maximum time a connection can be idle and still
  * be borrowed.  Helps deal with connections that are closed while
  * idling in the pool for an extended period.
```

The previous (and default) setting was `Infinite`.

Our theory is that http4s blaze-client maintains a connection pool with connections that never expire. However spring-based services (like WSM) have an idle timeout of 1 minute. So Leo was getting connections from the pool that had already been terminated by the server, leading to `HTTP connection closed`. 

Try setting the idle duration to a smaller value to see if it has a positive impact. Traffic should be low enough that reestablishing connections shouldn't be that much overhead. Note this affects _all_ Leo clients, not just WSM. Planning to do some BEE-load-testing of this.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
```diff
@@ -653,7 +653,14 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
                   )
                 )
             }
-          case None => deleteDiskAction
+          case None =>
+            // Error'd runtimes might not have a WSM resourceId and therefore no WsmJobStatus.
```
rtitle, line 665: Trying to fix this scenario which I think is currently happening to a few users:

1. User creates runtime
2. Back Leo processes `createRuntime` request
3. WSM call fails; Leo moves runtime to `Error` status
4. User tries to delete Error'd runtime
5. Front Leo moves runtime to `Deleting` status
6. Back Leo treats it as a no-op and doesn't fail; but never actually moves it to `Deleted` status
7. User is stuck with a runtime they can never delete, and get errors like this when they try to create a new one: 
image

LizBaldo (approved):

aherbst-broad:
Might be for another PR/ticket: it would be good to get 👀 on the # of connection closed errors we are encountering via a grafana dashboard (and really any other failures to create VMs or other resources) so we can know if this is fixed or not more easily than manually scraping logs or relying on user reports.

rtitle:
@aherbst-broad we do have WSM client metrics here: https://grafana.dsp-devops.broadinstitute.org/d/470LS8A7k/leonardo?orgId=1&viewPanel=310

It comes from http4s -- I _think_ `Abnormal Termination` includes connection closed errors.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/Boot.scala
rtitle: Testing in a BEE, I managed to get `HTTP connection closed` (from Sam, not WSM) even with this change. The Sam call was a GET though, so it retried and succeeded.

Thinking of making it retry all "connection closed" errors, regardless of method (GET/POST).

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/Boot.scala
```diff
@@ -741,7 +741,16 @@ object Boot extends IOApp {
...
     withRetry: Boolean
   ): Resource[F, org.http4s.client.Client[F]] = {
-    val retryPolicy = RetryPolicy[F](RetryPolicy.exponentialBackoff(30 seconds, 5))
+    // Retry all SocketExceptions to deal with pooled HTTP connections getting closed on the server-side.
+    // See https://broadworkbench.atlassian.net/browse/IA-4069.
+    val retryPolicy = RetryPolicy[F](
+      RetryPolicy.exponentialBackoff(30 seconds, 5),
+      (req, result) =>
+        result match {
+          case Left(e) if e.isInstanceOf[SocketException] => true
```
rtitle, line 750: Was debating whether to check for `"HTTP connection closed"` string as well but seems okay to just retry all `ScoketExceptions`..

lucymcnatt (approved):

rtitle:
Having a bit of trouble reproducing this on my BEE -- maybe traffic is lower, plus there is only 1 Leo/WSM/Sam instance (no load balancer) so the setup is a little different. But in general things seem to be working well, so think I'll try to get this merged and observe in dev.

rtitle:
jenkins retest

gdlex4015:
Rob, can not find IA-4909 issue in Jira. Is that the right number?

rtitle:
Sorry - should be https://broadworkbench.atlassian.net/browse/IA-4069

PR #3212, opened 2023-03-08
[IA-4067] Update GKM to 2.5.0 and fix job resource requests

afgane:
This updates GKM that uses the latest version of Galaxy (v23.0).


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -1507,8 +1507,8 @@ class GKEInterpreter[F[_]](
     // Machine type info
     val maxLimitMemory = machineType.memorySizeInGb
     val maxLimitCpu = machineType.numOfCpus
-    val maxRequestMemory = maxLimitMemory - 5
-    val maxRequestCpu = maxLimitCpu - 3
+    val maxRequestMemory = maxLimitMemory - 15
```
rtitle, line 1510: Just curious about this - why increase the difference between requests & limits memory/cpu? (15G just seems like a lot.)

rtitle (approved):
I didn't test this but the change LGTM ✅ -- happy to merge if you all are happy with it.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
afgane: We added 2 new services to the chart (rabbitmq & tusd) and these were the values I saw on a cluster that had just started Galaxy.

afgane:
I'd be glad to see this merged so we test on dev. (If the past month is any indication, there will be more iterations here but this is hopefully one big step closer to getting Galaxy running again.)

lucymcnatt:
jenkins retest

rtitle:
jenkins retest

jdcanas (approved):

rtitle:
Not sure why `fiab-start` is failing -- jenkins retest

PR #3215, opened 2023-03-08
[IA-4123] Dont use trace id as wsm job id

jdcanas:
I didn't test this manually since the repro steps are a bit much, but its a pretty straightforward change (and we honestly shouldn't be using the trace id anyways). Will test with a local leo soon.

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
@@ -522,7 +522,7 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
               msg.workspaceId,
               wsmResourceId,
               DeleteControlledAzureResourceRequest(
-                WsmJobControl(WsmJobId(s"delete-vm-${wsmResourceId.toString.take(10)}"))
+                WsmJobControl(WsmJobId(s"delete-vm-${wsmResourceId.value.toString.take(10)}"))
               )
             ),
             auth
@@ -552,7 +552,7 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
               msg.workspaceId,
               stagingBucketResourceId.resourceId,
               DeleteControlledAzureResourceRequest(
-                WsmJobControl(WsmJobId(s"del-staging-${stagingBucketResourceId.resourceId.toString.take(10)}"))
+                WsmJobControl(WsmJobId(s"del-staging-${stagingBucketResourceId.resourceId.value.toString.take(10)}"))
               )
             ),
             auth
@@ -591,7 +591,7 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
 
       getDeleteJobResultOpt = msg.wsmResourceId.flatTraverse(wsmResourceId =>
         wsmDao.getDeleteVmJobResult(
-          GetJobResultRequest(msg.workspaceId, WsmJobId(s"delete-vm-${wsmResourceId.toString.take(10)}")),
+          GetJobResultRequest(msg.workspaceId, WsmJobId(s"delete-vm-${wsmResourceId.value.toString.take(10)}")),
           auth
         )
       )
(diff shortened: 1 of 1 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
```diff
@@ -771,7 +772,7 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
             e.workspaceId,
             network.resourceId,
             DeleteControlledAzureResourceRequest(
-              WsmJobControl(WsmJobId(s"delete-networks-${ctx.traceId.asString.take(10)}"))
+              WsmJobControl(WsmJobId(s"delete-networks-${network.resourceId.value.toString.take(10)}"))
```
mlilynolting: opt: Could all of these maybe rely on passing an id to a little helper like `toJobId(jobName, resourceId`, since we're doing a little bit of special logic (first 10 digits)?

mlilynolting:
I see how this resolves the issue where multiple resource actions triggered by the same thread would silently fail because of duplicate job ids (since the traceId was the same).

Is it possible that we would have a similar duplicate job id issue caused by different threads attempting the same action on the same resource, because of the resource id being the same? For create/delete requests I don't think this would be an issue (as these probably can be idempotent), but if there are, say, multiple update actions on obj1 it seems like we'd lose jobs that duplicated the `update-obj1` job id. Is this worth fixing by generating job ids with resourceId + traceId?

mlilynolting (commented):
LGTM with nits - if my comments don't feel actionable, let me know and I'll approve.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
lucymcnatt: ^^

lucymcnatt (approved):

aherbst-broad:
This looks reasonable; I remain curious about what causes the disappearance of runtime records.

jdcanas:
> I see how this resolves the issue where multiple resource actions triggered by the same thread would silently fail because of duplicate job ids (since the traceId was the same).
> 
> Is it possible that we would have a similar duplicate job id issue caused by different threads attempting the same action on the same resource, because of the resource id being the same? For create/delete requests I don't think this would be an issue (as these probably can be idempotent), but if there are, say, multiple update actions on obj1 it seems like we'd lose jobs that duplicated the `update-obj1` job id. Is this worth fixing by generating job ids with resourceId + traceId?

This isn't an issue currently, as we use different IDs associated with the  leo record in any actions on the resource (currently only create/delete actions operate over a resouce). I could see us wanting to generate a unique ID entirely, but this way you can figure out the WSM job ID for any resource from the leo record, which I like.

jdcanas:
> This looks reasonable; I remain curious about what causes the disappearance of runtime records.

cc @aherbst-broad 

This was tricky to track down. Im 90% sure the following logic is to blame. Curious to hear your thoughts, but my conclusion is not a bug. Essentially, we dont eagerly exit with an error if WSM errors out on `deleteVm` due to some other cleanup we need to do, and proceed to poll the WSM jobId to determine what action to take. If there's no WSM job id, then we get `WsmJobStatus.None`, and proceed to mark the resource as deleted due to the comments in-line below. However, if, due to this leo bug, there's no WSM job ID, we also execute the same logic. My conclusion is not a bug it gets marked deleted, because it is the fault of the initial bug in leo that causes no WSM job ID to exist.

```
            resp.jobReport.status match {
              case WsmJobStatus.Succeeded =>
               ...
              case WsmJobStatus.Failed =>
            ...
              case WsmJobStatus.Running =>
               ...
          case None =>
            // Error'd runtimes might not have a WSM resourceId and therefore no WsmJobStatus.
            // We still want deletion to succeed in this case.
            for {
              _ <- dbRef.inTransaction(clusterQuery.updateClusterStatus(runtime.id, RuntimeStatus.Deleted, ctx.now))
              _ <- logger.info(ctx.loggingCtx)(s"runtime ${msg.runtimeId} with name ${runtime.runtimeName} is deleted successfully")
              _ <- deleteDiskAction
            } yield ()
```

mlilynolting (approved):

aherbst-broad:
@jdcanas hrm, interesting. It seems like if WSM returns anything other than a 2xx when leo issues the delete request, Leo should not delete the runtime record. 

Agreed that it's less likely now that the trace bug is fixed but I am 100% there will be other scenarios where WSM returns a non-200 (5xx seems like a possiblity for instance). Might be worth further investigation/ticketing/spiking to ensure leo doesn't leak these resources if it's unsure of the WSM state.

rtitle (approved):
LGTM 👍

aherbst-broad (approved):
lgtm, still want to think a bit more about the premature deletion of runtime records.

jdcanas:
jenkins retest

PR #3218, opened 2023-03-09
[WIP] [IA-4057] require user enabled directive on each direct Leo API route

mlilynolting:
Add an explicit check to Sam for user isEnabled, to all direct Leo resource routes. Fix ProxyService's non-check of enabled.

Consider this more of a sketch of an intention than an attempt at working code - my StandardEnabledUserDirectives class is the loadbearing element and I'm not sure it's the correct approach (can it actually get access to SamDAO's instance, `samDAO`?).

I'd deeply appreciate any eyes with notions of how to write an Akka Directive or properly surface errors during execution.

---

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/AppV2Routes.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/AppV2Routes.scala
@@ -19,51 +19,53 @@ import org.broadinstitute.dsde.workbench.model.UserInfo
 import org.broadinstitute.dsde.workbench.openTelemetry.OpenTelemetryMetrics
 import org.http4s.Uri
 
-class AppV2Routes(kubernetesService: AppService[IO], userInfoDirectives: UserInfoDirectives)(implicit
+class AppV2Routes(kubernetesService: AppService[IO], userInfoDirectives: UserInfoDirectives, enabledUserDirectives: EnabledUserDirectives)(implicit
   metrics: OpenTelemetryMetrics[IO]
 ) {
 
   val routes: server.Route = traceRequestForService(serviceData) { span =>
     extractAppContext(Some(span)) { implicit ctx =>
       userInfoDirectives.requireUserInfo { userInfo =>
-        CookieSupport.setTokenCookie(userInfo) {
-          pathPrefix("apps" / "v2" / workspaceIdSegment) { workspaceId =>
-            pathEndOrSingleSlash {
-              parameterMap { params =>
-                get {
-                  complete(
-                    listAppV2Handler(userInfo, workspaceId, params)
-                  )
-                }
-              }
-            } ~ pathPrefix(Segment) { appNameString =>
-              RouteValidation.validateNameDirective(appNameString, AppName.apply) { appName =>
-                post {
-                  entity(as[CreateAppRequest]) { req =>
-                    complete(
-                      createAppV2Handler(userInfo, workspaceId, appName, req)
-                    )
-                  }
-                } ~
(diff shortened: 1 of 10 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/EnabledUserDirectives.scala
```diff
@@ -0,0 +1,9 @@
+package org.broadinstitute.dsde.workbench.leonardo.http
+package api
+
+import akka.http.scaladsl.server.Directive
+import org.broadinstitute.dsde.workbench.model.UserInfo
+
+trait EnabledUserDirectives {
```
mlilynolting, line 7: This and some of the other plumbing for these new Directives are cargo-cult copies of UserInfoDirectives. I don't immediately see why this or UserInfoDirectives should be a `trait` since it's passed as a param into HttpRoutes by Boot.scala. Orch uses `with *Directives`; maybe that's better use of the architecture.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/AppV2Routes.scala
```diff
@@ -19,51 +19,53 @@ import org.broadinstitute.dsde.workbench.model.UserInfo
 import org.broadinstitute.dsde.workbench.openTelemetry.OpenTelemetryMetrics
 import org.http4s.Uri
 
-class AppV2Routes(kubernetesService: AppService[IO], userInfoDirectives: UserInfoDirectives)(implicit
+class AppV2Routes(kubernetesService: AppService[IO], userInfoDirectives: UserInfoDirectives, enabledUserDirectives: EnabledUserDirectives)(implicit
```
mlilynolting, line 22: Tons of whitespace changes in these files; reload with https://github.com/DataBiosphere/leonardo/pull/3218/files?w=1 to ignore

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/StandardEnabledUserDirectives.scala
```diff
@@ -0,0 +1,41 @@
+package org.broadinstitute.dsde.workbench.leonardo.http
+package api
+
+import akka.http.scaladsl.server.Directive
+// TODO akka require isn't in this package - where is it? see https://doc.akka.io/docs/akka-http/current/routing-dsl/directives/custom-directives.html#require-trequire
```
mlilynolting, line 5: And also, how does it work - I clearly don't understand the syntax needed to define and use Directives

rtitle:
Before diving into the code (and the akka `Directive` API ;)), question on the approach. I am a little nervous about a blanket extra Sam check to _all_ Leo requests. My understanding is we are only missing the enabled-user check for specific scenarios, right? E.g. during `list` calls when the result set is empty. Should we be more targeted and only do the explicit enabled-check in those cases?

I think all the `list` APIs (runtimes, disks, apps; v1 and v2) go through one of the `filter` methods in [LeoAuthProvider](https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeoAuthProvider.scala#L165-L185). So maybe that could be a good place to add the enabled-user check. We could have logic to do the enabled-user check only if the input list is empty. What do you think?

mlilynolting:
Yes, I could add the check to only the empty list cases in each list endpoint. I'll put something together!

mlilynolting:
Closing, replaced by https://github.com/DataBiosphere/leonardo/pull/3220 !

PR #3220, opened 2023-03-10
[IA-4057] Check that user is enabled when showing empty lists

mlilynolting:
…akes no resource permission checks (because resources returned are an empty list)

When an endpoint would return an empty list, and the user is not enabled in Sam, fail with 401 Forbidden, don't serve an empty list.

Validated with trivial unit test for SamAuthProvider + testing in my BEE with a user which has not accepted ToS.

Steps to validate:
- Create resources (disks, apps, runtimes) in workspaces (Google and Azure) using an authenticated user.
- With an unauthenticated user (you've signed up, but have not accepted terms of service):
	- Authorize through Leonardo Swagger (e.g. https://leonardo.mnolting.bee.envs-terra.bio)
	- Make a request to an endpoint (the issue was seen on endpoints without a resource context (e.g. https://leonardo.mnolting.bee.envs-terra.bio/#/apps/listApp) which returned no resources (e.g. when filtered by _labels nonexistent=bar))
	- See that requests matching any resources are rejected with a 403 (existing behavior)
	- See that requests matching no resources are rejected with a 401 (bugfix as of this PR)

Note also the ProxyService issue I outlined in comments & https://broadworkbench.atlassian.net/browse/IA-4131.

---

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/SamAuthProvider.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/SamAuthProvider.scala
@@ -285,7 +285,6 @@ class SamAuthProvider[F[_]: OpenTelemetryMetrics](
           )
     } yield ()
 
-
   override def isCustomAppAllowed(userEmail: WorkbenchEmail)(implicit ev: Ask[F, TraceId]): F[Boolean] =
     samDao.isGroupMembersOrAdmin(config.customAppCreationAllowedGroup, userEmail)
 
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/SamAuthProvider.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/SamAuthProvider.scala
@@ -257,6 +257,35 @@ class SamAuthProvider[F[_]: OpenTelemetryMetrics](
           )
     } yield samUserInfo.userEmail
 
+  /**
+   * Confirm the Sam user is enabled (accepted terms of service etc).
+   * Raises an AuthProviderException if user does not exist or is disabled.
+   */
+  override def isUserEnabled(petOrUserInfo: UserInfo)(implicit ev: Ask[F, TraceId]): F[Boolean] =
+    for {
+      traceId <- ev.ask
+      samUserInfoOpt <- samDao.getSamUserInfo(petOrUserInfo.accessToken.token)
+      samUserInfo <- F.fromOption(
+        samUserInfoOpt,
+        AuthProviderException(
+          traceId,
+          s"[SamAuthProvider.isUserEnabled] Subject info not found for ${petOrUserInfo.userEmail.value}",
+          StatusCodes.Unauthorized
+        )
+      )
+      _ <-
+        if (samUserInfo.enabled) F.unit
+        else
+          F.raiseError(
+            AuthProviderException(
+              traceId,
+              s"[SamAuthProvider.isUserEnabled] User ${samUserInfo.userEmail.value} is disabled.",
+              StatusCodes.Forbidden
+            )
+          )
+    } yield samUserInfo.enabled
+
+
   override def isCustomAppAllowed(userEmail: WorkbenchEmail)(implicit ev: Ask[F, TraceId]): F[Boolean] =
     samDao.isGroupMembersOrAdmin(config.customAppCreationAllowedGroup, userEmail)
 
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/SamAuthProvider.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/SamAuthProvider.scala
@@ -251,7 +251,7 @@ class SamAuthProvider[F[_]: OpenTelemetryMetrics](
           F.raiseError(
             AuthProviderException(
               traceId,
-              s"[SamAuthProvider.lookupOriginatingUserEmail] User ${samUserInfo.userEmail.value} is disabled.",
+              s"[SamAuthProvider.lookupOriginatingUserEmail] User ${samUserInfo.userEmail.value} is disabled",
               StatusCodes.Forbidden
             )
           )
@@ -261,7 +261,7 @@ class SamAuthProvider[F[_]: OpenTelemetryMetrics](
    * Confirm the Sam user is enabled (accepted terms of service etc).
    * Raises an AuthProviderException if user does not exist or is disabled.
    */
-  override def isUserEnabled(petOrUserInfo: UserInfo)(implicit ev: Ask[F, TraceId]): F[Boolean] =
+  override def checkUserEnabled(petOrUserInfo: UserInfo)(implicit ev: Ask[F, TraceId]): F[Unit] =
     for {
       traceId <- ev.ask
       samUserInfoOpt <- samDao.getSamUserInfo(petOrUserInfo.accessToken.token)
@@ -269,7 +269,7 @@ class SamAuthProvider[F[_]: OpenTelemetryMetrics](
         samUserInfoOpt,
         AuthProviderException(
           traceId,
-          s"[SamAuthProvider.isUserEnabled] Subject info not found for ${petOrUserInfo.userEmail.value}",
+          s"[SamAuthProvider.checkUserEnabled] Subject info not found for ${petOrUserInfo.userEmail.value}",
           StatusCodes.Unauthorized
         )
       )
@@ -279,11 +279,11 @@ class SamAuthProvider[F[_]: OpenTelemetryMetrics](
           F.raiseError(
             AuthProviderException(
               traceId,
-              s"[SamAuthProvider.isUserEnabled] User ${samUserInfo.userEmail.value} is disabled.",
+              s"[SamAuthProvider.checkUserEnabled] User ${samUserInfo.userEmail.value} is disabled",
               StatusCodes.Forbidden
             )
           )
-    } yield samUserInfo.enabled
+    } yield ()
 
 
   override def isCustomAppAllowed(userEmail: WorkbenchEmail)(implicit ev: Ask[F, TraceId]): F[Boolean] =
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
@@ -1141,9 +1141,10 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
     samVisibleAppsOpt <- NonEmptyList.fromList(samResources).traverse { apps =>
       authProvider.filterUserVisible(apps, userInfo)
     }
-    _ = if (samVisibleAppsOpt.isEmpty) authProvider.isUserEnabled(userInfo)
     res = samVisibleAppsOpt match {
-      case None => Vector.empty
+      case None =>
+        authProvider.checkUserEnabled(userInfo)
+        Vector.empty
       case Some(samVisibleApps) =>
         val samVisibleAppsSet = samVisibleApps.toSet
         // we construct this list of clusters by first filtering apps the user doesn't have permissions to see
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
@@ -1143,8 +1143,7 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
     }
 
     res = samVisibleAppsOpt match {
-      case None =>
-        Vector.empty
+      case None => Vector.empty
       case Some(samVisibleApps) =>
         val samVisibleAppsSet = samVisibleApps.toSet
         // we construct this list of clusters by first filtering apps the user doesn't have permissions to see
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
@@ -1167,7 +1167,8 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
           .flatMap(c => ListAppResponse.fromCluster(c, Config.proxyConfig.proxyUrlBase, labels))
           .toVector
     }
-    // If no visible clusters, check if user should be able to use this endpoint at all
+    // We authenticate actions on resources. If there are no visible clusters,
+    // we need to check if user should be able to see the empty list.
     _ <- if (res.isEmpty) authProvider.checkUserEnabled(userInfo) else F.unit
   } yield res
 
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
@@ -1141,9 +1141,9 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
     samVisibleAppsOpt <- NonEmptyList.fromList(samResources).traverse { apps =>
       authProvider.filterUserVisible(apps, userInfo)
     }
+
     res = samVisibleAppsOpt match {
       case None =>
-        authProvider.checkUserEnabled(userInfo)
         Vector.empty
       case Some(samVisibleApps) =>
         val samVisibleAppsSet = samVisibleApps.toSet
@@ -1168,6 +1168,8 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
           .flatMap(c => ListAppResponse.fromCluster(c, Config.proxyConfig.proxyUrlBase, labels))
           .toVector
     }
+    // If no visible clusters, check if user should be able to use this endpoint at all
+    _ <- if (res.isEmpty) authProvider.checkUserEnabled(userInfo) else F.unit
   } yield res
 
 }
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
@@ -1141,7 +1141,7 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
     samVisibleAppsOpt <- NonEmptyList.fromList(samResources).traverse { apps =>
       authProvider.filterUserVisible(apps, userInfo)
     }
-
+    _ = if (samVisibleAppsOpt.isEmpty) authProvider.isUserEnabled(userInfo)
     res = samVisibleAppsOpt match {
       case None => Vector.empty
       case Some(samVisibleApps) =>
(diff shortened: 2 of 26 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/SamAuthProvider.scala
```diff
@@ -257,6 +257,35 @@ class SamAuthProvider[F[_]: OpenTelemetryMetrics](
           )
     } yield samUserInfo.userEmail
 
+  /**
+   * Confirm the Sam user is enabled (accepted terms of service etc).
+   * Raises an AuthProviderException if user does not exist or is disabled.
+   */
+  override def isUserEnabled(petOrUserInfo: UserInfo)(implicit ev: Ask[F, TraceId]): F[Boolean] =
```
rtitle: As implemented this never returns `false`.
I think this should either return `F[Unit]` and raise an error (and also name it `checkUserEnabled` or something); or return `F[Boolean]` and not raise an error for the `false` case.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
```diff
@@ -1141,7 +1141,7 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
     samVisibleAppsOpt <- NonEmptyList.fromList(samResources).traverse { apps =>
       authProvider.filterUserVisible(apps, userInfo)
     }
-
+    _ = if (samVisibleAppsOpt.isEmpty) authProvider.isUserEnabled(userInfo)
```
rtitle: `samVisibleAppsOpt` is checked in the below match expression -- maybe put this there?

Also wondering what the difference is between the `None` case and `Some(empty list)`, and whether we need to check enablement for both. 🤔 
```
res <- samVisibleAppsOpt match {
  case None =>  // check enablement
  case Some(apps) if apps.isEmpty => // check enablement
  case Some(apps) => // logic
}
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeServiceInterp.scala
```diff
@@ -255,6 +255,7 @@ class RuntimeServiceInterp[F[_]: Parallel](config: RuntimeServiceConfig,
             userInfo
           )
       }
+      _ = if (samVisibleRuntimesOpt.isEmpty) authProvider.isUserEnabled(userInfo)
```
rtitle: Similar comment to above to reuse existing `match` expression.

Also noting by best practice shouldn't have an `if` without an `else`. In Scala (and FP languages in general) `if..else` is an expression which returns a value.

Also needs to be `_ <- authProvider.checkUserEnabled(userInfo)` since the method returns an `F[..]`. The `<-` makes it `flatMap` the `F`. `=` is for pure expressions.

rtitle (commented):
I know it's draft but just a few suggestions above! Thanks

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
mlilynolting: My read was that Some(empty list) wasn't possible because the type is `NonEmptyList`.

mlilynolting: (and my initial attempt to insert it into the None case was unsuccessful because my Scala instincts are trash - I can move there once I confirm it's working here..)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/SamAuthProvider.scala
mlilynolting: That makes sense, and speaks to a concern I had but couldn't articulate properly.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
mlilynolting: Moved!

mlilynolting:
Feedback addressed - building to my BEE to test now.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
```diff
@@ -1141,9 +1141,10 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
     samVisibleAppsOpt <- NonEmptyList.fromList(samResources).traverse { apps =>
       authProvider.filterUserVisible(apps, userInfo)
     }
-
     res = samVisibleAppsOpt match {
-      case None => Vector.empty
+      case None =>
+        authProvider.checkUserEnabled(userInfo)
```
rtitle: So this is a common Scala pitfall. The `authProvider.checkUserEnabled` won't actually execute here because the `F` it returns is not chained to the surrounding for-comp (flatMap). 

In a for-comp, `<-` should be used when the RHS expression is an `F[..]` and `=` should be used when the RHS is a pure expression.

So this will execute `f1`, `f2`, `f3` in sequence:
```
for {
  _ <- f1
  _ <- f2
  x = 42
  _ <- f3(x)
} yield ()
```

This is wrong:
```
for {
  _ <- f1
  _ = f2 // will not execute!!
  _ <- f3
} yield ()
```

So I think this code needs to look something like:
```
res <- samVisibleAppsOpt match {
  case None => 
    // returns F[Vector[something]]
    authProvider.checkUserEnabled(userInfo) >> Vector.empty
  case Some(samVisibleApps) =>
    // same logic but now needs to return an F[Vector[something]] instead of Vector[something]
}
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/ProxyService.scala
```diff
@@ -115,6 +115,11 @@ class ProxyService(
           userInfo <- googleOauth2Service.getUserInfoFromToken(token)
           _ <-
             if (checkUserEnabled) for {
+              // TODO [] this code seems to assume a non-enabled user will return an error or an empty response. However,
```
rtitle: I'm trying to wrap my head around this (pre-existing) code. When do we call this with `checkUserEnabled == false`? And it looks like this code is just calling sam `samDAO.getSamUserInfo` and checking for a response but not actually checking the enabled flag. 🤯 

Maybe this code should use the new `samAuthProvider.checkUserEnabled` method. And investigate how we're using the `checkUserEnabled` flag.

mlilynolting: I agree with your read here: I think that the /proxy endpoints never actually check for the user being enabled. I've hit them under what I think are the appropriate conditions to cause this check, and the error I get is a resource access authorization failure - which would only be thrown when the code gets past this check and actually tries to grab the resource in the URI. There's nothing actually exposed because all the current /proxy endpoints are in the context of specific resources (there are no list endpoints, etc).

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
rtitle: Btw I know it was my idea -- but if changing the `match` expression is too unwieldy in practice could also go back to separate check:
```
_ <- if (samVisibleAppsOpt.isEmpty)  authProvider.checkUserEnabled(userInfo) else F.unit 
res = samVisibleAppsOpt match { … }
```
(I don't actually care that much, so whatever is more readable :))

mlilynolting: Thank you!! Question, would I be able to accomplish the same thing with a `getOrElse`?
```
    samVisibleAppsOpt <- NonEmptyList.fromList(samResources).traverse { apps =>
      authProvider.filterUserVisible(apps, userInfo)
    }.getOrElse(authProvider.checkUserEnabled(userInfo))
```

mlilynolting: Oh, no, that wouldn't work, because then it wouldn't be a None...

mlilynolting: What if I snuck it into its own line? Then I wouldn't have to pollute the cases with a new type wrapper:
```
    _ <- samVisibleAppsOpt.getOrElse(authProvider.checkUserEnabled(userInfo))
```

mlilynolting: Maybe I should stop liveblogging my ignorance ^-^

rtitle:
```
  _ <- samVisibleAppsOpt.getOrElse(authProvider.checkUserEnabled(userInfo))
```
The types don't work in ^^. The RHS needs to return a `F[Unit]`. In the above code the `else` case returns an `F[Unit]` but not the `get` case. I didn't try but it probably doesn't compile..

mlilynolting: Yeah, I had realized :) I'll get there

rtitle: This would type check:
```
 _ <- samVisibleAppsOpt.map(_ => F.unit).getOrElse(authProvider.checkUserEnabled(userInfo))
```

mlilynolting: Type checking is suggesting that the Option here is wrapping a List, not a NonEmptyList anymore (maybe it lost that type in the traverse step?)

So I might need to cover the case of Some([]) | None after all.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/ProxyService.scala
mlilynolting: I created https://broadworkbench.atlassian.net/browse/IA-4131

mlilynolting: To answer your other Qs, we get user info "without" the enabled check (which isn't happening in any case but w/e) in ProxyRoutes:
```
          pathPrefix("proxy" | "notebooks") {

            corsSupport.corsHandler {

              refererHandler {
                // "apps" proxy routes
                pathPrefix("google" / "v1" / "apps") {
                  pathPrefix(googleProjectSegment / appNameSegment / serviceNameSegment) {
                    (googleProject, appName, serviceName) =>
                      extractUserInfoWithoutUserEnabledCheck(implicitly) { userInfo =>
```
and a couple other routes as well.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
mlilynolting: I moved the checks to the end of the methods, since many of them were filtering inside the cases and I wanted to keep this access check a little bit separate for clarity.

mlilynolting:
I'd like to rename WhitelistAuthProvider to AllowListAuthProvider but I am a little worried about side effects/dependencies.

rtitle:
Oh yeah we should. It should be safe to refactor. I don’t think anything is using it today — once upon a time we had pluggable AuthProviders in Leo, but nowadays everything is standardized on Sam (and not likely to change).

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/SamAuthProvider.scala
```diff
@@ -251,12 +251,40 @@ class SamAuthProvider[F[_]: OpenTelemetryMetrics](
...
               StatusCodes.Forbidden
             )
           )
     } yield samUserInfo.userEmail
 
+  /**
+   * Confirm the Sam user is enabled (accepted terms of service etc).
+   * Raises an AuthProviderException if user does not exist or is disabled.
+   */
+  override def checkUserEnabled(petOrUserInfo: UserInfo)(implicit ev: Ask[F, TraceId]): F[Unit] =
```
n8landolt, line 264: We should probably have more tests the SamAuthProviderSpec that tests the negative cases

mlilynolting: @n8landolt done!

leonardo-example.conf
```diff
@@ -134,10 +134,10 @@ akka {
 # Authorization implementation config
 auth {
   # To use your own provider implementation, subclass org.broadinstitute.dsde.workbench.leonardo.model.LeoAuthProvider and provide the class name here
-  providerClass = "org.broadinstitute.dsde.workbench.leonardo.auth.WhitelistAuthProvider"  # A default auth provider
+  providerClass = "org.broadinstitute.dsde.workbench.leonardo.auth.AllowlistAuthProvider"  # A default auth provider
   providerConfig { # Do not remove providerConfig, even if empty
     # Any provider specific config should be specified here
-    # For the whitelist provider, specify a comma separated list of user emails that should be allowed access
+    # For the allowlist provider, specify a comma separated list of user emails that should be allowed access
     whitelist = [EMAIL1,EMAIL2]
```
mlilynolting, line 141: Leaving external configs as is to avoid breaking anyone's setup

mlilynolting:
Reverified everything's working - all looks good after latest changes!

mlilynolting:
jenkins retest

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskServiceInterp.scala
```diff
@@ -230,6 +230,8 @@ class DiskServiceInterp[F[_]: Parallel](config: PersistentDiskConfig,
             )
             .toVector
       }
+      // If no visible disks, check if user should be able to use this endpoint at all
+      _ <- if (res.isEmpty) authProvider.checkUserEnabled(userInfo) else F.unit
```
lucymcnatt, line 235: question, why are we checking if the user should be able to use the endpoint after the rest of the logic is executed? Would it not make more sense to do this check initially and give the error response if the user isn't enabled before trying to fetch info from the DB?

mlilynolting: This was my feeling initially, but the idea was to only do this check if no resources were visible. Could move this to the top but if so I'd want to just do it without a conditional - apply the second call to all requests.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
```diff
@@ -1167,6 +1167,8 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
           .flatMap(c => ListAppResponse.fromCluster(c, Config.proxyConfig.proxyUrlBase, labels))
           .toVector
     }
+    // If no visible clusters, check if user should be able to use this endpoint at all
+    _ <- if (res.isEmpty) authProvider.checkUserEnabled(userInfo) else F.unit
```
jdcanas, line 1172: Hmm... I know that in all cases except for when the list is empty sam handles TOS logic, so this is technically correct. However, I think its a bit more readable to check this up front and fail early rather than query the db first. (since failure is related to user info, which we have before DB calls)

mlilynolting: I'm totally down with that - are we okay with the overhead of always making this additional request? This would add one additional round trip to Sam for every list request from an enabled user. @jdcanas @lucymcnatt

jdcanas: Yeah... I guess I'm not too sure how often this is happening. If we think it will frequently rejected, we should same time on DB call. If we want to be cautious and think its an edge case, we can leave it like this and save some us the lookup time. Could leave it like this, and add a metric call to count if this happens often so we can reverse it? Performance-wise it could be worth doing it lazily at the end like this...

mlilynolting: I don't imagine this would be happening very often because it only applies to users who have registered but have not yet accepted ToS.

lucymcnatt: would it make sense to add a follow-up to make this the only Sam check, ie remove it from wherever its currently hitting Sam because that check will have already been taken care of with this call?

mlilynolting: I'd prefer to leave it as is, since the current design does not actually expose any info - the security "loophole" only allows unauthenticated users to deduce that there are no results for their query; if they make a request which would actually return resources Sam will block them as intended. Adding overhead to authenticated requests seems like a poor trade for closing that hole.

jdcanas: We still need the other Sam checks (which ask if the user has permission to perform the relevant action on the resource). The check we're adding here just asks if they have permission to even use terra basically.

jdcanas:
Hmm.. a little confused as to what next steps you're recommending in https://broadworkbench.atlassian.net/browse/IA-4131.

Are you advocating we don't add this Sam check to `setCookie` and leave the proxy routes as is? If so, do we need a separate ticket to track it as opposed to a sub-task to make the case to the security team?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
lucymcnatt: if we decide to leave as is, it would be helpful if the comment was a little clearer on why this check is happening where it is, something like `If no visible clusters, Sam wasn't called to verify users permission to perform action, need to check if user should be able to use this endpoint at all`

mlilynolting: > would it make sense to add a follow-up to make this the only Sam check, ie remove it from wherever its currently hitting Sam because that check will have already been taken care of with this call?

I think what you're asking is whether we should add this additional check to Sam to all calls. Because Leo's existing requests to Sam happen in the context of a resource + an action - so "for disk D, user U, action A, can U perform A on D?"

The check I'm adding has only the user and the action, and without a resource there isn't enough info to satisfy the above permission model. So we'll still need to call Sam for all the resources involved in the request, and the end result works out to calling Sam an extra time for good requests (your recommendation) vs calling Sam a bunch more times for bad requests (my code). 

Definitely upsides and downsides to both! For example, a malicious requester could perhaps flood these endpoints with requests forcing lots of pointless lookups, since we're actually doing expensive resource checks for unauthenticated users. But I imagine there are plenty of ways to DDoS Terra and it might be out of scope to try to close those vulnerabilities.

mlilynolting:
@jdcanas IA-4131 isn't a subtask, it's just a separate bug ticket. I'm not recommending a particular resolution at this stage, just identifying that the proxy routes do not actually check the user is enabled, despite the code apparently being written with the assumption that a check is happening.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
lucymcnatt: thanks for the context! I think with that info I'm fine going down the route of only checking if empty to minimize the extra resource check for that's unnecessary in most cases.

jdcanas (approved):

lucymcnatt (approved):

PR #3221, opened 2023-03-13
[AJ-777] support for wds as a new app type

jordansafer:
Changes:
- new apptype added
- new configs pointing to the wds chart in terra-helmfile
- new cases for wds apptype referring to the wds chart and version
- update buildCromwellChartOverrideValues/chart override values for wds chart
- update status checks for wds app type
- update tests, swagger, compilation fixes
- end to end testing on BEE

---

In all cases:


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -179,12 +189,6 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
           } yield Some(petMi)
       }
 
-      // Get the batch account key
```
jordansafer, line 182: Since this call is specific to cromwell, only make it for a cromwell appType

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
```diff
@@ -329,6 +334,7 @@ object AppType {
       case Galaxy   => FormattedBy.Galaxy
       case Cromwell => FormattedBy.Cromwell
       case Custom   => FormattedBy.Custom
+      case Wds      => FormattedBy.Cromwell
```
jordansafer, line 339: Creating a new "FormattedBy" enum adds a lot of boiler plate for no functional change. For this reason I'm in favor of keeping this as is, but am open to going back through an renaming the current enum

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -216,18 +226,38 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
...
+              .run(authContext)
+          } yield ()
+        case AppType.Wds =>
+          for {
+            _ <- helmClient
+              .installChart(
+                app.release,
+                app.chart.name,
+                app.chart.version,
+                buildWdsChartOverrideValues(
```
jordansafer, line 246: We could combine this with the buildCromwellChartOverrideValues function and take appType as a param. I think this is cleaner since wds overrides requires different parameters & depends on different resource creation, so I expect these to diverge to some extent over time

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -354,33 +384,46 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
       token <- F.fromOption(tokenOpt, AppCreationException(s"Pet not found for user ${userEmail}", Some(ctx.traceId)))
       authHeader = Authorization(Credentials.Token(AuthScheme.Bearer, token))
 
-      op = config.coaAppConfig.coaServices
+      services = appType match {
+        case AppType.Wds => config.wdsAppConfig.coaServices
```
jordansafer, line 391: possibly I should rename this to `wdsService`? So this would read, `set(config.wdsAppConfig.wdsService)`. The logic pattern could also be entirely changed, if we went with a separate case for WDS appType. I like that this shares the wds getStatus call for both cromwell and wds app types, since it is the same status call.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -449,15 +494,52 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
         // Enabled services configs
         raw"cbas.enabled=${config.coaAppConfig.coaServices.contains(Cbas)}",
         raw"cbasUI.enabled=${config.coaAppConfig.coaServices.contains(CbasUI)}",
-        raw"wds.enabled=${config.coaAppConfig.coaServices.contains(Wds)}",
+        raw"wds.enabled=${config.coaAppConfig.coaServices.contains(Wds)}", // TODO remove after WDS chart migration
```
jordansafer, line 500: As far as migration:
- first we'll deploy these changes adding support for a WDS only app
- next we'll deploy a change shifting autocreation to autocreate the WDS only app. This be in rawls or TerraUI, depending on what mechanism we're using for autodeploy
- *How does CBAS find the WDS URL? We'll need to make sure CBAS is in sync for the migration
- finally, we can remove the logic for WDS in the cromwell App in leo, and remove the helmchart from cromwhelm

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
aaronkanzer: Is this supposed to be `FormattedBy.Cromwell` as well? Or something else?

aaronkanzer: Could we leave a docstring here perhaps explaining why `FormattedBy.Cromwell` is still compatible?

http/src/main/resources/reference.conf
```diff
@@ -238,6 +238,21 @@ azure {
...
 
+   wds-app-config {
+     instrumentation-enabled = false
+     chart-name = "/leonardo/wds"
+     chart-version = "0.3.0"
+     release-name-suffix = "wds-rls"
+     namespace-name-suffix = "wds-ns"
+     ksa-name = "wds-ksa"
+     services = [
+       {
```
aaronkanzer, line 249: I assume we are waiting to remove the following clause from `coa-app-config` above until the de-coupling is official?

```
 {
   name = "wds"
   kind = "ClusterIP"
 },
```

jordansafer: yup! exactly- see the migration comment below-

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/KubernetesAppConfig.scala
```diff
@@ -87,6 +87,19 @@ final case class CoaAppConfig(chartName: ChartName,
...
+                              ksaName: KsaName,
+                              services: List[ServiceConfig],
+                              instrumentationEnabled: Boolean
+) extends KubernetesAppConfig {
+  override lazy val kubernetesServices: List[KubernetesService] = services.map(s => KubernetesService(ServiceId(-1), s))
+  override val serviceAccountName = ServiceAccountName(ksaName.value)
+  def coaServices: Set[CoaService] = Set(Wds)
+}
+
 sealed trait CoaService
```
aaronkanzer, line 103: Do we eventually need something similar here for `Wds` as a service?

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreterSpec.scala
```diff
@@ -166,6 +168,33 @@ class AKSInterpreterSpec extends AnyFlatSpecLike with TestComponent with Leonard
...
+      "config.applicationInsightsConnectionString=applicationInsightsConnectionString," +
+      "config.subscriptionId=sub," +
+      s"config.region=${azureRegion}," +
+      "general.leoAppInstanceName=app," +
+      s"general.workspaceManager.workspaceId=${workspaceId.value}," +
+      "identity.name=identity-name," +
+      "identity.resourceId=identity-id," +
+      "identity.clientId=identity-client-id," +
+      "sam.url=https://sam.dsde-dev.broadinstitute.org/," +
+      "fullnameOverride=coa-rel-1," +
```
aaronkanzer: Just for consistency can we use `wds-` here instead of `coa-

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreterSpec.scala
```diff
@@ -275,6 +304,62 @@ class AKSInterpreterSpec extends AnyFlatSpecLike with TestComponent with Leonard
     deletion.unsafeRunSync()(cats.effect.unsafe.IORuntime.global)
   }
 
+  it should "create and poll a wds app, then successfully delete it" in isolatedDbTest {
```
aaronkanzer: Can you parameterize this test with the `it should "create and poll a coa app, then successfully delete it" ` test?

Or do the Scala spirits not allow that to be easily done here?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
aaronkanzer: This seems fine by me :D

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
davidangb: I agree this is an ugly change - it'll even require a liquibase migration - but I think we should make the change. I feel like this will confuse people down the line and it'll be a prerequisite if we ever want to have a different PD configuration.

I'm fine making this change in a targeted follow-on PR - get this PR merged, then update the `FormattedBy` separately.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
davidangb: this is a little confusing but it's also a minefield. Should we also rename `CoaService` to something more generic? I don't have a strong opinion, fine with this as-is.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
davidangb: you should claim AJ-909 :)

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpWdsDAOSpec.scala
```diff
@@ -66,7 +66,7 @@ class HttpWdsDAOSpec extends AnyFlatSpec with Matchers with LeonardoTestSuite wi
     )
 
     val wdsDAO = new HttpWdsDAO(okWds)
-    val res = wdsDAO.getStatus(Uri.unsafeFromString("https://test.com/wds/status"), authHeader)
+    val res = wdsDAO.getStatus(Uri.unsafeFromString("https://test.com/wds/status"), authHeader, AppType.Cromwell)
```
davidangb: how about adding tests for /status if WDS is AppType.WDS?

davidangb (commented):
hopefully minor comments inline. The code in the PR looks good - I don't have a good sense if anything is missing, though.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
jordansafer: leaving as-is

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
jordansafer: claimed!

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/KubernetesAppConfig.scala
jordansafer: I don't think so. These are used as an enum to handle different cases for different Coa services, so Wds won't need this unless we are devising a system of conditionally deploying apps on WDS. Which we aren't now :)

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreterSpec.scala
jordansafer: this will always be coa, regardless of appType. This comes from the "appChartPrefix" which is currently hardcoded to "coa". I'll update it to not be hardcoded-

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreterSpec.scala
jordansafer: yup!

jordansafer:
unrelated timeout failure on galaxy app deletion. jenkins retest

jordansafer:
jenkins retest

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpWdsDAOSpec.scala
```diff
@@ -46,28 +46,56 @@ class HttpWdsDAOSpec extends AnyFlatSpec with Matchers with LeonardoTestSuite wi
...
 
-    val response =
+    val upResponse =
+      """
+        |{
+        |  "status": "DOWN"
+        |}
+      """.stripMargin
+
+    val downResponse =
```
davidangb, line 64: nit: looks like the `upResponse` and `downResponse` are transposed - I don't think this has effect on the tests but could be confusing variable naming.

davidangb (approved):
one nit on variable naming, otherwise :+1:!

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpWdsDAOSpec.scala
jordansafer: ah! the test was broken. we were returning "DOWN" when "UP" was expected, and the flipped naming here made it incorrectly pass! I added positive and negative cases to the test so it would fail in the future if names were flipped. Fixed now :)

aaronkanzer (approved):

jordansafer:
jenkins retest

PR #3239, opened 2023-03-20
[IA-4094] deleteDisk and getDisk V2 routes

lucymcnatt:
Added V2 routes for deleteDisk and getDisk
 - queries by disk id and workspace id
 - only compatible with azure currently (need workspaceId to be propagated for google disks)
 - adds workspaceId as a field to the PersistentDisk table
 - deletes and polls disk before updating database

Part of a larger effort with [IA-3955](https://github.com/DataBiosphere/terra-ui/pull/3823) to enable Azure PDs

---

In all cases:

[IA-3955]: https://broadworkbench.atlassian.net/browse/IA-3955?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -1230,6 +1230,86 @@ paths:
               schema:
                 $ref: "#/components/schemas/ErrorReport"
 
+  "/api/v2/disks/{workspaceId}/azure/{name}":
```
n8landolt: I'm concerned about treating name as a query param when we could just use the persistent disk ID.

Name should be used more in a "search" or "filter" function, not a "get"

I believe we use this incorrect pattern elsewhere, but I would like to see us move away from it.

A quick search: https://stackoverflow.com/questions/31506341/should-a-rest-api-select-on-a-id-or-a-name-field

n8landolt: Chat GPT's take on this:
"When designing REST APIs, it is generally recommended to query resources by their unique identifier, which is typically an ID. The primary reason for this is that using IDs to identify resources provides a clear and unambiguous way to reference a specific resource, which can help prevent errors and confusion.

Querying by name can be more ambiguous, as multiple resources may have the same name, and names can change over time. Additionally, names may contain characters that are not URL-safe and may need to be encoded, which can make URLs longer and less readable."

n8landolt: The only thing I can think of - is this name the "primary key" in Azure? In that case, I can see a case for that being used here - however, having the argument be "name" doesn't instill confidence.

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoApiClient.scala
```diff
@@ -364,13 +363,13 @@ object LeonardoApiClient {
             method = Method.POST,
             headers = Headers(authHeader, defaultMediaType, traceIdHeader),
             uri = rootUri
-              .withPath(Uri.Path.unsafeFromString(s"/api/google/v1/disks/${googleProject.value}/${diskName.value}")),
+              .withPath(Uri.Path.unsafeFromString(s"/api/google/v1/disks/${googleProject.value}/${diskName}")),
```
jdcanas: I don't think any of the changes from `${diskName.value}` to `${diskName}` are valid in this file. `AzureDiskName` and `DiskName` both seem to have the same `final case class A(value: String) extends AnyVal` definition, meaning usage in string interpolation like done here will result in a string `"...A(value)..."`.

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/runtimes/RuntimeGceSpec.scala
```diff
@@ -45,7 +44,7 @@ class RuntimeGceSpec
 
   "should create a GCE instance in a non-default zone" in { project =>
     val runtimeName = randomClusterName
-    val diskName = genDiskName.sample.get
+    val diskName = randomDiskName()
```
jdcanas: hmm.... this has to be a bug, its used as a function here and a string in other places such as `RuntimeCreationDiskSpec`?

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodec.scala
```diff
@@ -82,7 +81,7 @@ object JsonCodec {
   implicit val urlEncoder: Encoder[URL] = Encoder.encodeString.contramap(_.toString)
   implicit val zoneNameEncoder: Encoder[ZoneName] = Encoder.encodeString.contramap(_.value)
   implicit val regionNameEncoder: Encoder[RegionName] = Encoder.encodeString.contramap(_.value)
-  implicit val diskNameEncoder: Encoder[DiskName] = Encoder.encodeString.contramap(_.value)
+  implicit val diskNameEncoder: Encoder[DiskName] = Encoder.encodeString.contramap(_.asString)
```
jdcanas: why does this need changing?

edit: ah, i see now that we define DiskName ourselves. Probably can just keep the definition the same as what it was in workbench libs though?

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodec.scala
```diff
@@ -484,7 +482,7 @@ object JsonCodec {
 
   implicit val zoneDecoder: Decoder[ZoneName] = Decoder.decodeString.map(ZoneName(_))
   implicit val regionDecoder: Decoder[RegionName] = Decoder.decodeString.map(RegionName(_))
-  implicit val diskNameDecoder: Decoder[DiskName] = Decoder.decodeString.emap(s => validateName(s).map(DiskName))
```
jdcanas: Why are we removing diskname validation? This is pretty important as it enforced a valid disk name in POST requests to leo.

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/diskModels.scala
```diff
@@ -21,11 +21,14 @@ final case class PersistentDisk(id: DiskId,
                                 appRestore: Option[AppRestore],
                                 labels: LabelMap,
                                 sourceDisk: Option[DiskLink],
-                                wsmResourceId: Option[WsmControlledResourceId]
+                                wsmResourceId: Option[WsmControlledResourceId],
+                                workspaceId: Option[WorkspaceId]
 ) {
-  def projectNameString: String = s"${cloudContext.asStringWithProvider}/${name.value}"
+  def projectNameString: String = s"${cloudContext.asStringWithProvider}/${name}"
```
jdcanas: I don't think this is valid, its still a value class

http/src/main/resources/swagger/api-docs.yaml
jdcanas: Name is unique in conjunction with a workspace ID, and we should likely be enforcing this with a DB constraint in this PR (as workspaceId is now a new field). I believe we do this for runtimes...

jdcanas: Also, do we need azure in this route? Can the cloud platform not be looked up in WSM via workspace id? I wish we didn't have it in there for v2 runtime  routes...

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/DiskServiceDbQueries.scala
```diff
@@ -85,10 +76,28 @@ object DiskServiceDbQueries {
...
+                                   cloudContextOpt: Option[CloudContext] = None,
+                                   workspaceIdOpt: Option[WorkspaceId] = None
+  )(implicit
     executionContext: ExecutionContext
   ): DBIO[GetPersistentDiskResponse] = {
-    val diskQuery = persistentDiskQuery.findActiveByNameQuery(cloudContext, diskName)
+    val diskQuery = (cloudContextOpt, workspaceIdOpt) match {
+      case (Some(cloudContext), None) => persistentDiskQuery.findActiveByNameQuery(cloudContext, diskName)
+      case (None, Some(workspaceId))  => persistentDiskQuery.findActiveByNameWorkspaceQuery(workspaceId, diskName)
+      // case _ => DBIO.failed()
```
jdcanas: Hmm... this doesn't seem like an exhaustive match?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/PersistentDiskComponent.scala
```diff
@@ -174,12 +180,11 @@ object persistentDiskQuery {
       .filter(_.name === name)
       .filter(_.destroyedDate === dummyDate)
 
-  private[db] def findByNameQuery(cloudContext: CloudContext, name: DiskName) =
+  private[db] def findActiveByNameWorkspaceQuery(workspaceId: WorkspaceId, name: DiskName) =
     tableQuery
-      .filter(
-        _.cloudContext === cloudContext.asCloudContextDb
-      )
+      .filterOpt(Some(workspaceId))(_.workspaceId === _)
```
jdcanas: Here, we assume uniqueness based on Name + workspaceId + destroyedDate. We should likely have a db constraint to enforce this so it does not allow insertion of resources that violate this assumption.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -1120,7 +1124,7 @@ class LeoPubsubMessageSubscriber[F[_]](
               .getDisk(
                 msg.project,
                 zone,
-                d.name
+                GoogleDiskName(d.name.asString)
```
jdcanas: Why did we make `AzureDiskName` be `DiskName` and then add `GoogleDiskName`. Can we not use the same for both? I don't understand how the same field in a database can be decoded to 2 different value classes..

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -1574,11 +1574,11 @@ class GKEInterpreter[F[_]](
       raw"""rbac.serviceAccount=${ksa.value}""",
       // Persistence configs
       raw"""persistence.nfs.name=${namespaceName.value}-${config.galaxyDiskConfig.nfsPersistenceName}""",
-      raw"""persistence.nfs.persistentVolume.extraSpec.gcePersistentDisk.pdName=${nfsDisk.name.value}""",
+      raw"""persistence.nfs.persistentVolume.extraSpec.gcePersistentDisk.pdName=${nfsDisk.name}""",
       raw"""persistence.nfs.size=${nfsDisk.size.gb.toString}Gi""",
       raw"""persistence.postgres.name=${namespaceName.value}-${config.galaxyDiskConfig.postgresPersistenceName}""",
       raw"""galaxy.postgresql.galaxyDatabasePassword=${config.galaxyAppConfig.postgresPassword.value}""",
-      raw"""persistence.postgres.persistentVolume.extraSpec.gcePersistentDisk.pdName=${postgresDiskName.value}""",
+      raw"""persistence.postgres.persistentVolume.extraSpec.gcePersistentDisk.pdName=${postgresDiskName}""",
```
jdcanas: I don't believe this is correct, its still a value class?

http/src/main/resources/swagger/api-docs.yaml
n8landolt: Yeah we probably should - and also I think in past experiences "name" can be user defined - in this instance it is not, so I think my concerns are invalid.

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -3037,33 +3117,33 @@ components:
...
         - labels
+        - formattedBy
       properties:
         id:
           type: integer
-          description: Internal Leonardo ID of the persistent disk
-        googleProject:
-          type: string
-          description: The google project used to create the persistent disk
+          description: The user-supplied ID of the persistent disk
```
n8landolt: nit: I believe the IDs being used here for both Azure and GCP are "Cloud-platform provided"
"user-supplied" while correct is a bit vague, and could mean the user decides what this is.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskServiceInterp.scala
```diff
@@ -384,7 +384,8 @@ object DiskServiceInterp {
       None,
       labels,
       sourceDisk.map(_.diskLink),
-      None
+      None,
+      None // TODO: workspace must be present for V2 routes
```
lucymcnatt, line 388: Open question:
Because workspaceId is a new field and deleteDisk requires it, only new PDs will be able to be deleted:
 - should the option be grayed out in the UI for old disks?
 - should we update the DB and propagate those values?

jdcanas: Hmm... probably the first option. A tooltip along the lines if 'Azure disks without a workspace id must be deleted with an associated runtime'. Worth mentioning that we won't actually be able to call this endpoint in the UI without  a workspaceId. 

I think its ok to essentially 'orphan' disks in between this migration, since azure is very much a WIP

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/PersistentDiskComponent.scala
jdcanas: the syntax can be found in multiple places in liquibase code ie.

`<addUniqueConstraint columnNames="googleProject, clusterName, destroyedDate" constraintName="IDX_CLUSTER_UNIQUE" tableName="CLUSTER"/>`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskServiceInterp.scala
lucymcnatt: yeah option 1 seemed like the most probable, as long as we have a mechanism for cleaning up those orphaned disks (do we have a current cronjob that does this?) then I'm fine with that.

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/diskModels.scala
```diff
@@ -56,6 +57,7 @@ object DiskStatus extends Enum[DiskStatus] {
   final case object Ready extends DiskStatus
   final case object Deleting extends DiskStatus
   final case object Deleted extends DiskStatus
+  final case object Error extends DiskStatus
```
jdcanas, line 60: Hmm... interesting... seems like we will need to add this to the `DiskStatus` enum in terra-ui

lucymcnatt: I though this was a useful status if the wsm process errored during deletion, so instead of `deleting` the user would see `error` and know to try deletion again rather than waiting.

jdcanas: Yeah, for sure useful, just commenting that we need to make sure the UI knows this is a valid status (and getting typescript to tell us if we are  missing it in case statements)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/PersistentDiskComponent.scala
```diff
@@ -174,12 +180,11 @@ object persistentDiskQuery {
...
 
-  private[db] def findByNameQuery(cloudContext: CloudContext, name: DiskName) =
+  private[db] def findActiveByIdWorkspaceQuery(workspaceId: WorkspaceId, id: DiskId) =
     tableQuery
-      .filter(
-        _.cloudContext === cloudContext.asCloudContextDb
-      )
-      .filter(_.name === name)
+      .filterOpt(Some(workspaceId))(_.workspaceId === _)
+      .filter(_.id === id)
```
jdcanas: hmm... this query is a bit weird. If id itself is a primary key enforced by table unique constraints, and it's required (aka non-Optional) for this function, does filtering by these other things do anything? Seems a trivial optimization to just filter by ID.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskServiceInterp.scala
jdcanas: The cloud environment page should allow users to clean up once they are allowed to delete disks in the UI

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskV2ServiceInterp.scala
```diff
@@ -0,0 +1,132 @@
...
+
+  override def deleteDisk(userInfo: UserInfo, workspaceId: WorkspaceId, diskId: DiskId)(implicit
+    as: Ask[F, AppContext]
+  ): F[Unit] =
+    for {
+      ctx <- as.ask
+      diskOpt <- persistentDiskQuery.getActiveByIdWorkspace(workspaceId, diskId).transaction
+
+      disk <- diskOpt.fold(
+        F.raiseError[PersistentDisk](DiskNotFoundByIdWorkspaceException(diskId, workspaceId, ctx.traceId))
```
jdcanas: probably can just query by diskId and use the workspaceId for just the sam lookup

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskV2ServiceInterp.scala
```diff
@@ -0,0 +1,132 @@
...
+      )(
+        F.pure
+      )
+
+      _ <- F
+        .raiseUnless(disk.status.isDeletable)(
```
jdcanas, line 85: seems like we are checking if its deletable twice. We should probably eliminate this check, and not give the caller any info about the disk before the authProvider check https://github.com/DataBiosphere/leonardo/pull/3239/files#diff-f824ca8a3e1de98e9abaa9465267c92afa550a384bda7338bf7d32b3b0c4bcc5R94

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskV2ServiceInterp.scala
```diff
@@ -0,0 +1,132 @@
...
+          DiskCannotBeDeletedException(diskId, disk.status, disk.cloudContext, ctx.traceId)
+        )
+      hasPermission <-
+        if (disk.auditInfo.creator == userInfo.userEmail)
+          F.pure(true)
+        else
+          authProvider
+            .isUserWorkspaceOwner(WorkspaceResourceSamResourceId(workspaceId), userInfo)
```
jdcanas: hmm... is this the correct auth check? I don't think a user needs to be a workspaceOwner to see the disk (i.e. receive the message that the disk isn't in a deletable status at the very least).  Shouldn't we first check if they can see the disk, and 404 otherwise:

```
      listOfPermissions <- authProvider.getActionsWithProjectFallback(disk.samResource, googleProject, userInfo)
      hasReadPermission = listOfPermissions._1.toSet
        .contains(PersistentDiskAction.ReadPersistentDisk) || listOfPermissions._2.toSet
        .contains(ProjectAction.ReadPersistentDisk)
```

and then check if the disk is deletable (and 409 if it is not in a deletable status)

```
      hasDeletePermission = listOfPermissions._1.toSet
        .contains(PersistentDiskAction.DeletePersistentDisk) || listOfPermissions._2.toSet
        .contains(ProjectAction.DeletePersistentDisk)
      _ <- if (hasDeletePermission) F.unit else F.raiseError[Unit](ForbiddenError(userInfo.userEmail))
      // throw 409 if the disk is not deletable
      _ <-
        if (disk.status.isDeletable) F.unit
        else F.raiseError[Unit](DiskCannotBeDeletedException(googleProject, disk.name, disk.status, ctx.traceId)) 
```
cc @LizBaldo @rtitle since y'all have done more Sam work than me recently

Took this code directly from v1 disk routes, not sure if anything has changed between azure disks and google disks in sam...

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
```diff
@@ -823,6 +786,142 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
...
+                                            WsmControlledResourceId
+                                 ], // can be either runtimeId or diskResourceId in order to get WSMcontrolled resource
+                                 workspaceId: WorkspaceId,
+                                 auth: Authorization
+  )(implicit
+    ev: Ask[F, AppContext]
+  ): F[Unit] =
+    for {
+      ctx <- ev.ask
+      diskResourceOpt <- id match {
```
jdcanas, line 794: can this code be simplified by using the diskId?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskV2ServiceInterp.scala
LizBaldo: Hummm, it makes sense to me that only the disk creator or workspace owner should have the permission to delete disks no?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
```diff
@@ -823,6 +786,142 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
...
+      taskToRun = for {
+        // We need to wait until WSM deletion job to be done to update the database
+        resp <- streamFUntilDone(
+          getDeleteJobResult,
+          config.deleteDiskPollConfig.maxAttempts,
+          config.deleteDiskPollConfig.interval
+        ).compile.lastOrError
+
+        _ <- resp.jobReport.status match {
+          case WsmJobStatus.Succeeded =>
```
jdcanas, line 872: feels like this pattern of 'perform x,y,z based on exhasutive match of WsmJobStatus` should be encapsulated elsewhere, but a bit out of scope here.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskV2ServiceInterp.scala
jdcanas: I agree they should have permission, my point is more so that workspace owners are a subset of the users that have permission. Its possible we need to ask `isTheUserWorkspaceOwner || doesUserHaveReadPermission` first, then `isDiskDeletable && (doesUserHaveDeletePermission || isTheUserWorkspaceOwner)`. Whether or not we need `isTheUserWorkspaceOwner || xxx` is dependent on whether the owner inherits `xxx (i.e. read/delete)` based on workspace ownership.

lucymcnatt: My thinking was following the same rules for disk deletion as runtime deletion because that's the process that currently controls azure disk deletion.

LizBaldo: Damn, I see both points, so the use case would be a non-workspace owner, but maybe somebody who has write permission on the runtime and even though they canot delete the runtime they should still be able to delete the underlying disk? Sounds a bit scary to me

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/PersistentDiskComponent.scala
lucymcnatt: so I wrote it this way to enforce that the disk has to be from the specified workspace in the call, otherwise a user could put a workspace they have ownership over and the diskId of a disk outside that scope, however, I guess I could just check that the workspaceIds match after grabbing the disk in the getDisk function.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskV2ServiceInterp.scala
jdcanas: No, this shouldn't have anything to do with runtimes. Disks are distinct entities in Sam with permissions separate from the workspace. 

Image the following scenario:
A workspace with 2 users, User 1 is owner and User 2 is a regular User

User 2 creates a runtime with a disk.

User 1 and User 2 should be able to delete that disk then. This code does not allow User 2 to delete the disk.

LizBaldo: Fair, we should trust the `doesUserHaveDeletePermission` resource in Sam then 👍

lucymcnatt: ok I'll update the call

jdcanas: Hmm... so digging a bit I don't think this is possible for azure disks with how they are created. When we create v1 disks, we register it with sam in v1 interp. We don't do this for v2 disks created as part of the azure runtime. This seems like a bit of an oversight, we should probably register azure disks in `createRuntime` now that we are allowing the life cycle to be managed independently.
```
            samResource <- F.delay(PersistentDiskSamResourceId(UUID.randomUUID().toString))
            disk <- F.fromEither(
              convertToDisk(userInfo.userEmail,
                            petSA,
                            cloudContext,
                            diskName,
                            samResource,
                            config,
                            req,
                            ctx.now,
                            sourceDiskOpt
              )
            )
            _ <- authProvider
              .notifyResourceCreated(samResource, userInfo.userEmail, googleProject)
              .handleErrorWith { t =>
                log.error(t)(
                  s"[${ctx.traceId}] Failed to notify the AuthProvider for creation of persistent disk ${disk.projectNameString}"
                ) >> F.raiseError(t)
              }
```

lucymcnatt: I actually already added that notifyresourcecreated call [here](https://github.com/DataBiosphere/leonardo/pull/3239/files#diff-b6c8ddc1824e19ffe2fc9a0ab8b0e372d1cfc4944dc67c23b6f1a5411da904d0R196-R197) (it was previously being handled by WSM I'm pretty sure), so it should be trivial to check the delete permission

jdcanas: Ah ok, yeah, took a while to hunt it down on my end. So is notifying sam not being handled by WSM for disks like it is for runtimes? (Just wondering what would happen if we notify then WSM notifies)

Sounds like we are on the same page in terms of the permission lookup for delete though.

lucymcnatt: oh yeah I went down a rabbit hole at one point to figure out where this was happening and WSM has logic that skips re-notifying if the resource already exists in Sam.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/PersistentDiskComponent.scala
jdcanas: Hmm... I'll leave it up to you. I would expect that if you ask sam 'does user A have permission to do B to resource C in workspace D', it would trivially error if the resource is not in the workspace. Not sure whether we want to trust that or add redundancy.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
lucymcnatt: the issue is that in the RUNTIME_CONTROLLED_RESOURCE table only has runtimeId, wsmResourceId, and the resourceType, so to make this code simpler it would involve more complicated querying, both to get the diskId from the runtimeId (multiple joins) before passing it into this func, and then to get the WsmControlledResource from the diskId (also multiple joins).

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
```diff
@@ -507,43 +501,15 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
...
+    implicit val wsmDeleteVmDoneCheckable: DoneCheckable[GetDeleteJobResult] = (v: GetDeleteJobResult) =>
+      v.jobReport.status.equals(WsmJobStatus.Succeeded) || v.jobReport.status == WsmJobStatus.Failed
     for {
       ctx <- ev.ask
 
       runtimeOpt <- dbRef.inTransaction(clusterQuery.getClusterById(msg.runtimeId))
       runtime <- F.fromOption(runtimeOpt, PubsubHandleMessageError.ClusterNotFound(msg.runtimeId, msg))
       auth <- samDAO.getLeoAuthToken
 
-      // Delete the VM in WSM
```
lucymcnatt, line 519: here I basically moved the WSM job polling into the wsmResourceId fold because I added an adaptError to the deleteDisk call which, in the case that the wsmResourceId was None and the deleteDisk call errored, the runtime would be updated to 'Deleted' and the destroyedDate updated, but then the error would be caught by the handleAzureRuntimeDeletionError, which then set the status to 'Error'.

LizBaldo (approved):
LGTM after pair reviewing together!

I am okay with either adding a unit test to check the state of deleting a runtime that has no wsm resource id associated with it, or leave a TODO with a follow up JIRA ticket to address :)

PR #3241, opened 2023-03-21
[IA-4125] terra-azure-relay-listeners version bump; should set dateAc…

mlilynolting:
…cessed on Azure runtimes

Not yet tested properly; need to spin up a new Azure VM and verify dateAccessed is changing.


rtitle:
Just noticed this.. is this to pick up https://github.com/DataBiosphere/terra-azure-relay-listeners/pull/27 which calls the Leo `setDateAccessed` endpoint?

If so I wonder if Leo also needs to provide new config parameters to the Listener to set these fields: https://github.com/DataBiosphere/terra-azure-relay-listeners/pull/27/files#diff-4dab07009b19ea187f607a407b00997345505227a852e7115e4dcd086f27a388R24-R28

rtitle:
I also just dropped a [note](https://github.com/DataBiosphere/terra-azure-relay-listeners/pull/27/files#r1143984309) in the listener PR -- just thinking we probably only want to enable this in runtime-listener deployments, not app-listener deployments.

mlilynolting:
These are most applicable comments; will address.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
```diff
@@ -147,7 +147,9 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
           stagingContainerName.value,
           stagingContainerResourceId.value.toString,
           params.workspaceName,
-          wsStorageContainerUrl
+          wsStorageContainerUrl,
+          "https://leonardo.dsde-dev.broadinstitute.org", // ServiceHost field needs the leonardo URL to hit when calling dateAccessed - how do we get this value?
```
mlilynolting: @rtitle @giventocode  what is the real source of this URL? What should go here?

mlilynolting:
> I also just dropped a [note](https://github.com/DataBiosphere/terra-azure-relay-listeners/pull/27/files#r1143984309) in the listener PR -- just thinking we probably only want to enable this in runtime-listener deployments, not app-listener deployments.

@rtitle I think that omitting changes from AKSInterpreter should leave us clear of the app-listener deployments, but I could be misunderstanding.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
rtitle: The Leo config template has a `leonardoHostname` function which templates in the correct URL per environment: 

https://github.com/broadinstitute/firecloud-develop/blob/dev/base-configs/leonardo/leonardo.conf.ctmpl#L19

Unfortunately I don't see an existing config field for this (I guess we never needed it before) -- maybe you could add something like:

```
application {
  …
  leoUrl = "{{executeTemplate "leonardoHostname"}}
}
```

[here](https://github.com/broadinstitute/firecloud-develop/blob/dev/base-configs/leonardo/leonardo.conf.ctmpl#L215)?

http/src/main/resources/init-resources/azure_vm_init_script.sh
```diff
@@ -104,6 +110,10 @@ docker run -d --restart always --network host --name listener \
 --env LISTENER_RELAYCONNECTIONNAME=$RELAY_CONNECTION_NAME \
 --env LISTENER_SAMINSPECTORPROPERTIES_SAMRESOURCEID=$SAMRESOURCEID \
 --env LISTENER_SAMINSPECTORPROPERTIES_SAMURL=$SAMURL \
+--env LISTENER_SETDATEACCESSEDINSPECTORPROPERTIES_SERVICEHOST=$LEONARDOURL \
```
rtitle: I think you also need to enable the `setDateAccessed` inspector as well -- by default only the [samChecker](https://github.com/DataBiosphere/terra-azure-relay-listeners/blob/main/service/src/main/resources/application.yml#L20) is enabled.

My syntax could be off but something like:
```
--env LISTENER_REQUESTINSPECTORS_0=samChecker \
--env LISTENER_REQUESTINSPECTORS_1=setDateAccessed
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
mlilynolting: Done, hopefully, here: https://github.com/broadinstitute/firecloud-develop/pull/3269

rtitle:
> I think that omitting changes from AKSInterpreter should leave us clear of the app-listener deployments, but I could be misunderstanding.

Yeah I think that's right -- should be no Leo changes needed to the app stuff.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
jdcanas: You will need to add this field in `leo.conf` and an env var in `local.env` to be compatible with the independent release changes

http/src/main/resources/init-resources/azure_vm_init_script.sh
```diff
@@ -59,9 +59,15 @@ SERVER_APP_ALLOW_ORIGIN="*"
...
 SERVER_APP_WEBSOCKET_URL="wss://${RELAY_NAME}.servicebus.windows.net/${HCVAR}/${RELAY_CONNECTION_NAME}"
 
-#Relay listener configuration
+# Relay listener configuration
 RELAY_CONNECTIONSTRING="Endpoint=sb://${RELAY_NAME}.servicebus.windows.net/;SharedAccessKeyName=listener;SharedAccessKey=${RELAY_CONNECTION_POLICY_KEY};EntityPath=${RELAY_CONNECTION_NAME}"
 
+# Relay listener configuration - setDateAccessed listener
+LEONARDOURL="${18:-dummy}"
+RUNTIMENAME="${19:-dummy}"
+DATEACCESSEDSLEEPSECONDS=60 # supercedes default defined in terra-azure-relay-listeners/service/src/main/resources/application.yml
```
jdcanas: we typically use snake_case and caps for bash ENV vars

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/ApplicationConfig.scala
```diff
@@ -10,5 +11,16 @@ case class ApplicationConfig(applicationName: String,
                              leoGoogleProject: GoogleProject,
                              leoServiceAccountJsonFile: Path,
                              leoServiceAccountEmail: WorkbenchEmail,
+                             leoUrlBase: String,
```
jdcanas: Seems like this should be type `URL` instead of string so we get validation at config reading time?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
LizBaldo: I merged the firecloud-develop PR above, without it we won't be able to test these changes

LizBaldo (approved):
Tested this on a BEE pointing to both the backend and front end changes and was able to:
- Create a VM with a 5 minutes autopause threshold
- Stare at it for 5 minutes
- See the VM pausing automagically after 5 minutes

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/ApplicationConfig.scala
LizBaldo: Just seeing that sorry, will address before merging!

PR #3242, opened 2023-03-22
[IA-4160] Add app/runtime version to metrics

rtitle:
https://broadworkbench.atlassian.net/browse/IA-4160

Small metrics-only change -- add tag to specify _version_ of runtime/apps. For apps this is the chart version; for runtimes this is the imageUrl. This will be useful for seeing what versions are running in the wild; and monitoring that app upgrades are working (in the future).

See also design doc about app upgrades: https://docs.google.com/document/d/1T-QzAgFGEmWJEhoUMQ0wzFAFrdC-gSWqLrEgcqEKAuI/edit#

Demo of this working (from my BEE).

Apps:
image

Runtimes:
image

---

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoMetricsMonitor.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoMetricsMonitor.scala
@@ -95,7 +95,8 @@ class LeoMetricsMonitor[F[_]](config: LeoMetricsMonitorConfig,
                       a.appType,
                       a.status,
                       getRuntimeUI(a.labels),
-                      getAzureCloudContext(c.cloudContext)
+                      getAzureCloudContext(c.cloudContext),
+                      a.chart
       ) -> 1
     )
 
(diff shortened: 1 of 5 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoMetricsMonitor.scala
```diff
@@ -108,18 +109,18 @@ class LeoMetricsMonitor[F[_]](config: LeoMetricsMonitorConfig,
   private[monitor] def countRuntimesByDbStatus(allRuntimes: List[RuntimeMetrics]): Map[RuntimeStatusMetric, Int] = {
     val allContainers = for {
       r <- allRuntimes
-      // Exclude welder to not double-count runtimes. Assume every runtime has a jupyter or rstudio
-      // container, but not both.
-      c <- r.containers if c != RuntimeContainerServiceType.WelderService
+      // Only care about Jupyter, RStudio, or Azure image types.
+      // Assume every runtime has exactly 1 of these.
+      imageTypes = Set(RuntimeImageType.Jupyter, RuntimeImageType.RStudio, RuntimeImageType.Azure)
```
LizBaldo, line 114: Nit, but i would maybe rename these to be more specific:
- JupyterGce
- RStudioGce
- JupyterAzure

LizBaldo (approved):
LGTM! Thanks for adding these :) I just have a small comment about renaming for better clarity

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoMetricsMonitor.scala
rtitle: I agree -- these enums predated this change though, I can see how large the refactor would be.

rtitle: I think I am going to leave the enums as-is. Looks like `RuntimeImageType` also gets serialized to the database, so changing it would likely also involve a DB migration.
```
MySQL [leonardo]> select imageType, count(1) from CLUSTER_IMAGE group by 1 order by 2 desc;
+----------------+----------+
| imageType      | count(1) |
+----------------+----------+
| Jupyter        |    58485 |
| Welder         |    58464 |
| Proxy          |    56461 |
| BootSource     |    56283 |
| CryptoDetector |    25639 |
| RStudio        |      988 |
| Azure          |      688 |
| Listener       |      570 |
+----------------+----------+
8 rows in set (0.876 sec)
```

LizBaldo: Oh yeah, no worries then! Thanks for looking into it :)

PR #3247, opened 2023-03-28
IA-4174: Remove defunct WSM APIs from Leo

rtitle:
Context in: https://broadworkbench.atlassian.net/browse/IA-4174

I will test on a BEE, first to replicate the issue on dev, then verify this fix.

---

In all cases:

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/WsmDao.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/WsmDao.scala
@@ -170,7 +170,7 @@ final case class CreateDiskRequest(workspaceId: WorkspaceId,
                                    diskData: CreateDiskRequestData
 )
 
-final case class CreateDiskRequestData(name: AzureDiskName, size: DiskSize, region: com.azure.core.management.Region)
+final case class CreateDiskRequestData(name: AzureDiskName, size: DiskSize)
 
 final case class CreateDiskResponse(resourceId: WsmControlledResourceId)
 
@@ -429,7 +429,7 @@ object WsmEncoders {
     )
 
   implicit val diskRequestDataEncoder: Encoder[CreateDiskRequestData] =
-    Encoder.forProduct3("name", "size", "region")(x => (x.name.value, x.size.gb, x.region.toString))
+    Encoder.forProduct2("name", "size")(x => (x.name.value, x.size.gb))
   implicit val createDiskRequestEncoder: Encoder[CreateDiskRequest] =
     Encoder.forProduct2("common", "azureDisk")(x => (x.common, x.diskData))
 
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/WsmDao.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/WsmDao.scala
@@ -10,7 +10,6 @@ import org.broadinstitute.dsde.workbench.azure._
 import org.broadinstitute.dsde.workbench.leonardo.JsonCodec.{
   azureImageEncoder,
   azureMachineTypeEncoder,
-  azureRegionEncoder,
   googleProjectDecoder,
   runtimeNameEncoder,
   storageContainerNameDecoder,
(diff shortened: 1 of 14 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/WsmDao.scala
```diff
@@ -33,10 +31,6 @@ import java.time.ZonedDateTime
 import java.util.UUID
 
 trait WsmDao[F[_]] {
-  def createIp(request: CreateIpRequest, authorization: Authorization)(implicit
```
rtitle, line 36: I deleted the methods/models from `WsmDao` around IP, network, and relay namespace which are not used anymore.

Side note: I would _really_ like if Leo migrated to using published Java client libraries for WSM (and Sam) in the future. That would reduce boilerplate code and surface for potential contract-related bugs.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala
```diff
@@ -308,6 +308,16 @@ class RuntimeV2ServiceInterp[F[_]: Parallel](config: RuntimeServiceConfig,
         .raiseError[Unit](RuntimeNotFoundException(runtime.cloudContext, runtimeName, "permission denied"))
         .whenA(!hasPermission)
 
+      // Query WSM for Landing Zone resources
```
rtitle, line 311: `DeleteAzureRuntimeMessage` now needs `LandingZoneResources`, since it needs the relay namespace in order to delete the hybrid connection. This is a similar pattern to `CreateAzureRuntimeMessage`.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoMetricsMonitor.scala
```diff
@@ -115,7 +115,8 @@ class LeoMetricsMonitor[F[_]](config: LeoMetricsMonitorConfig,
       c <- r.images.filter(i => imageTypes.contains(i.imageType)).headOption
     } yield Map(
       RuntimeStatusMetric(r.cloudContext.cloudProvider,
-                          c,
+                          c.imageType,
```
rtitle, line 118: This is an unrelated metrics fix, which I also happened to notice since the release on 3/28. Storing `RuntimeImage` in the `RuntimeStatusMetric` and `RuntimeHealthMetric` case classes was breaking aggregation because it holds a timestamp field. Changed that type to just contain the fields we care about for metrics (imageType and imageUrl).

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/WsmDao.scala
LizBaldo: That is a very good point, I will create a tech debt ticket for us to address that

LizBaldo: https://broadworkbench.atlassian.net/browse/IA-4175

LizBaldo (approved):

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
```diff
@@ -585,12 +577,12 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
...
         case x: CloudContext.Azure => F.pure(x.value)
       }
-      _ <- relayNamespaceOpt.traverse(ns =>
-        azureRelay.deleteRelayHybridConnection(
-          ns,
-          RelayHybridConnectionName(runtime.runtimeName.asString),
-          cloudContext
-        )
+
+      // Delete hybrid connection for this VM
```
aherbst-broad, line 579: This is a good catch, explains why the connections were leaking.

rtitle: Yup, trying to verify in a BEE that they are being deleted now

rtitle:
Testing notes:

First I reproduced the error in my BEE with dev code: 
image

It works with the Leo fix:
image


Also verified the relay hybrid connection now gets deleted upon VM deletion. I think this partially resolves https://broadworkbench.atlassian.net/browse/IA-4097, will comment in that ticket.

rtitle:
Metrics look good too -- will plan to merge after green tests.

rtitle:
RStudioSpec failed which is a known flake. Going to merge this.

PR #3253, opened 2023-03-30
[IA-4149][IA-4097] Allow runtime creation when there is a deleting runtime and clean up hybrid connections

jdcanas:
Tested in a bee


mlilynolting (approved):

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeServiceInterp.scala
```diff
@@ -241,9 +241,10 @@ class RuntimeServiceInterp[F[_]: Parallel](config: RuntimeServiceConfig,
     for {
       ctx <- as.ask
       (labelMap, includeDeleted, _) <- F.fromEither(processListParameters(params))
+      excludeStatuses = if (includeDeleted) List.empty else List(RuntimeStatus.Deleted)
```
LizBaldo, line 244: So we are not propagating this change to GCP runtimes? I am fine with it, but I would leave a comment in the code saying so and linking the ticket, otherwise this split in behavior is going to be confusing for the next dev working in this file next.

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeServiceV2InterpSpec.scala
```diff
@@ -332,6 +331,154 @@ class RuntimeServiceV2InterpSpec extends AnyFlatSpec with LeonardoTestSuite with
...
+        Some(context.traceId),
+        workspaceDesc.get.displayName,
+        ContainerName("dummy")
+      )
+      message shouldBe expectedMessage
+    }
+    res.unsafeRunSync()(cats.effect.unsafe.IORuntime.global)
+  }
+
+  it should "create a runtime if one exists but in deleting status" in isolatedDbTest {
```
LizBaldo, line 410: This would be the only test that I would expect to be added as part of this PR, it is scary to see that the two above did not exist before 😱 Thanks for adding that coverage!

LizBaldo (approved):
LGTM with only a couple of minor comments. 
For clarity though, this PR only addresses allowing the runtime creation, not cleaning up the hybrid connection ticket. Could you please clean up the PR name to only reflect the one ticket and link it in the PR description? Thanks!

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeServiceInterp.scala
jdcanas: GCP is already different insofar that we do not enforce one runtime per workspace. This is new to v2 because WSM enforces it. Leo did not enforce this in the past, and I don't think we should start

jdcanas:
jenkins retest

env/local.env
```diff
@@ -4,7 +4,7 @@
 # Terra Application Config
 GOOGLE_PROJECT=broad-dsde-dev
 LEO_SVC_ACCT_EMAIL=leonardo-dev@broad-dsde-dev.iam.gserviceaccount.com
-LEO_URL_BASE=leonardo.dsde-dev.broadinstitute.org
+LEO_URL_BASE=https://leonardo.dsde-dev.broadinstitute.org
```
LizBaldo, line 7: sorry!

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -320,6 +320,13 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
 
       clusterName = landingZoneResources.clusterName // NOT the same as dbCluster.clusterName
 
+      // Delete hybrid connection for this app
```
LizBaldo, line 323: Nice so when tested on a BEE you can confirm that the hybrid connection is indeed deleted from the azure portal as well?

LizBaldo (approved):

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
jdcanas: Tested in a BEE, hybrid connection is deleted properly!

lucymcnatt (approved):

PR #3254, opened 2023-03-31
[IA-4148] Bumped version number for Azure DSVM

n8landolt:
https://broadworkbench.atlassian.net/browse/IA-4148

---

In all cases:

Erdal has tested the image and found it works for his use case!

Screenshot 2023-04-20 at 10 18 00 AM
Screenshot 2023-04-20 at 10 18 20 AM


http/src/main/resources/init-resources/azure_vm_init_script.sh
```diff
@@ -132,6 +132,32 @@ LEONARDO_URL="${18:-dummy}"
 RUNTIME_NAME="${19:-dummy}"
 DATEACCESSED_SLEEP_SECONDS=60 # supercedes default defined in terra-azure-relay-listeners/service/src/main/resources/application.yml
 
+# Save output to file
```
LizBaldo: This makes sense for me to keep if:
- Maybe slightly modify to `Save output to file for debugging purposes`
- The users would not access this file via terra-ui or the terminal

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
```diff
@@ -149,9 +149,11 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
         )
 
         val cmdToExecute =
-          s"echo \"${contentSecurityPolicyConfig.asString}\" > csp.txt && bash azure_vm_init_script.sh ${arguments
+          s"echo \"${contentSecurityPolicyConfig.asString}\" > csp.txt && bash azure_vm_init_script.sh ${
+            arguments
               .map(s => s"'$s'")
-              .mkString(" ")}"
+              .mkString(" ")
+          } > /var/log/azure_vm_init_script.log"
```
LizBaldo: Why do you need the env_vars.text? If you echo all variables they should already show up in this log file no?

mlilynolting (approved):
scalafmt issues but otherwise looks peachy

http/src/main/resources/init-resources/azure_vm_init_script.sh
n8landolt: Agree on the comment!
Is there an issue with them having access to this data?
To my knowledge it should be at `/home/leoAdmin/` so users won't have access. I could explicitly specify this as well.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
n8landolt: This was added by Justin - I guess this is a good point!

Do we have a preference? I think Justin's solution may make more sense.

n8landolt: Although - this doesn't seem to label the arguments. 
Maybe I change my init script output to also be in /var/log, and name the file better? Such as `azure_vm_init_script_args.txt`

n8landolt: I think I'll just echo the output and it will automatically be recorded in the azure_vm_init_script.log instead of me creating that text file at /home/leoAdmin

LizBaldo: That would be my preferred option as well, as long as this log file is indeed hidden from the user

n8landolt: It appears as it stands `/var/log/` is available to the user to read.

What is the concern with them being able to see this info?

LizBaldo: I think we would need to ask AppSec at this point just to be sure, you could also comment out the log dump, it would be easy enough to trigger when debugging again.

n8landolt: I think I'd prefer to keep the logs, but I'll just restrict access to the log dump to read only for root. Since we have SSH capabilities now, this would make the debugging process a LOT easier. It's not easy to update this script, especially in prod.

LizBaldo: Sure, as long as the access is restricted and user cannot see it, feel free to keep it in

n8landolt: @LizBaldo See screenshots - users can't see the log output. File is owned by root, and only root can read! So when we SSH, we have access still!

LizBaldo: Amazing, thanks so much! 👍

LizBaldo (approved):

PR #3255, opened 2023-03-31
[IA-4065] Support Hail batch as a new app type on azure

rtitle:
JIRA: https://broadworkbench.atlassian.net/browse/IA-4065

See design doc: https://docs.google.com/document/d/1L6ScP6Xjp9TYC5VCQHdsRRBKsGVjwjaWb2sDusCNid0/edit#

This PR adds a new Leo AppType `HAIL_BATCH` to deploy the Hail Batch app on Azure. It is integrated similar to Cromwell and WDS as a top-level AppType. We expect to go through production readiness and then launch this behind a feature flag.

* Related terra-ui PR: todo
* BEE workspace for testing: https://terra.rtitle-azure.bee.envs-terra.bio/#workspaces/rt-apr-19/Hail%20Batch%20Testing%20Redux

---

In all cases:


Dockerfile
```diff
@@ -32,6 +32,7 @@ ENV NGINX_VERSION 4.3.0
 ENV CROMWELL_CHART_VERSION 0.2.220
 ENV CROWELL_ON_AZURE_CHART_VERSION 0.2.220
 ENV WDS_CHART_VERSION 0.3.0
+ENV HAIL_BATCH_CHART_VERSION 0.1.4
```
rtitle: We will likely iterate on the helm chart a bit more and bump the version -- Hail team has said there are fixes forthcoming.

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/samModels.scala
```diff
@@ -24,7 +24,7 @@ object SamResourceId {
     override def resourceType: SamResourceType = SamResourceType.Project
   }
 
-  final case class AppSamResourceId(resourceId: String, accessScope: Option[AppAccessScope]) extends SamResourceId {
+  final case class AppSamResourceId(resourceId: String)(accessScope: Option[AppAccessScope]) extends SamResourceId {
```
rtitle: This is not strictly relevant to hail batch, but is to fix a bug I noticed when testing workspace-shared apps in Leo. Can give more context…

http/src/main/resources/reference.conf
```diff
@@ -258,6 +259,22 @@ azure {
...
+     release-name-suffix = "hail-rls"
+     namespace-name-suffix = "hail-ns"
+     ksa-name = "hail-ksa"
+     services = [
+       {
+         name = "batch"
+         kind = "ClusterIP"
+       }
+     ]
+     enabled = true
```
rtitle, line 277: I introduced an `enabled` flag for all app types (defaulting to true). Was just thinking we could potentially merge this PR with hail-batch disabled in prod until compliance review happens.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -425,16 +446,23 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
...
-      case AppType.Cromwell => s"http://$appChartPrefix-${release.asString}-reverse-proxy-service:8000/"
-      case AppType.Wds      => s"http://$appChartPrefix-${release.asString}-wds-svc:8080"
+      case AppType.Cromwell  => s"http://coa-${release.asString}-reverse-proxy-service:8000/"
+      case AppType.Wds       => s"http://wds-${release.asString}-wds-svc:8080"
+      case AppType.HailBatch => "http://batch:8080"
       case AppType.Galaxy | AppType.Custom =>
         F.raiseError(AppCreationException(s"App type $appType not supported on Azure"))
     }
 
+    // Hail batch serves requests on /{appName}/batch and uses relative redirects,
```
rtitle, line 469: This was a pain to figure out -- when we generalize apps we need to keep differences in proxy-redirects in mind.

rtitle:
Hey @LizBaldo I'm still going through some paperwork and testing but I think this is probably ready for initial review. See [design doc](https://docs.google.com/document/d/1L6ScP6Xjp9TYC5VCQHdsRRBKsGVjwjaWb2sDusCNid0/edit#) for more context (also linked in the PR description). Thanks!

Dockerfile
```diff
@@ -40,7 +41,7 @@ COPY --from=helm-go-lib-builder /helm-go-lib-build/helm-scala-sdk/helm-go-lib /l
 # Install the Helm3 CLI client using a provided script because installing it via the RHEL package managing didn't work
 RUN curl -fsSL -o get_helm.sh https://raw.githubusercontent.com/helm/helm/master/scripts/get-helm-3 && \
     chmod 700 get_helm.sh && \
-    ./get_helm.sh --version v3.2.4 && \
+    ./get_helm.sh --version v3.11.2 && \
```
rtitle, line 44: Bumped helm version in Leo -- was needed for `oci` chart, and probably good to do anyway. (App deploys seem to work still, including Galaxy)

http/src/main/resources/reference.conf
```diff
@@ -258,6 +259,22 @@ azure {
          path = "/"
        }
      ]
+     enabled = true
+   }
+
+   hail-batch-app-config {
+     chart-name = "/leonardo/hail-batch-terra-azure"
```
dmohs, line 266: @rtitle Where is this? I'd love to get a look at it.

rtitle: That's the path cached in Leo -- in the Dockerfile this is the pull command:
```
helm pull oci://us-docker.pkg.dev/hail-vdc/terra-dev-public/hail-batch-terra-azure --version 0.1.4 --untar
```

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/samModels.scala
LizBaldo: Oh yeah, I definitely introduced this, I would love some more context if you can share 🙏

Dockerfile
```diff
@@ -59,14 +60,17 @@ RUN helm repo add ingress-nginx https://kubernetes.github.io/ingress-nginx && \
...
     helm pull terra-app-setup-charts/terra-app-setup --version $TERRA_APP_SETUP_VERSION --untar && \
     helm pull galaxy/galaxykubeman --version $GALAXY_VERSION --untar && \
     helm pull terra/terra-app --version $TERRA_APP_VERSION --untar  && \
     helm pull ingress-nginx/ingress-nginx --version $NGINX_VERSION --untar && \
     helm pull cromwell-helm/cromwell --version $CROMWELL_CHART_VERSION --untar && \
     helm pull cromwell-helm/cromwell-on-azure --version $CROWELL_ON_AZURE_CHART_VERSION --untar && \
     helm pull terra-helm/wds --version $WDS_CHART_VERSION --untar && \
-    helm repo update && \
+    helm pull oci://us-docker.pkg.dev/hail-vdc/terra-dev-public/hail-batch-terra-azure --version $HAIL_BATCH_CHART_VERSION --untar && \
+    # TODO kind of a hack, remove when hail-batch removes relaylistener from their chart
```
LizBaldo: Why do we need to remove the relay listener (either here or in the chart)?

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
```diff
@@ -322,21 +322,24 @@ object AppType {
...
     override def toString: String = "CUSTOM"
   }
 
   def values: Set[AppType] = sealerate.values[AppType]
   def stringToObject: Map[String, AppType] = values.map(v => v.toString -> v).toMap
 
-  /** Host formatting for an App. Currently, for Azure there is only 1 configuration,
-   * and FormattedBy.Cromwell is expected. */
+  /** Disk formatting for an App. Currently, for Azure there is only 1 configuration,
+   * and FormattedBy.Cromwell is expected for WDS and HailBatch. */
```
LizBaldo: So this means that a disk used for a Cromwell app can be reused for a WDS or Hail app as well, I am just curious as why this is a desired behavior?
I guess what I am asking is why we cannot share between Galaxy and cromwell, but it is okay to share between cromwell and WDS/Hail

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpHailBatchDAO.scala
```diff
@@ -0,0 +1,38 @@
...
+import org.http4s.client.dsl.Http4sClientDsl
+import org.http4s.headers.Authorization
+
+class HttpHailBatchDAO[F[_]](httpClient: Client[F])(implicit
+  F: Async[F],
+  metrics: OpenTelemetryMetrics[F]
+) extends HailBatchDAO[F]
+    with Http4sClientDsl[F] {
+
+  override def getStatus(baseUri: Uri, authHeader: Authorization)(implicit
```
LizBaldo, line 26: Could you leave a comment explainng the difference between the regular status, and the driver status? I went through the design doc, but it still is a bit unclear to me

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeoAuthProvider.scala
```diff
@@ -46,10 +46,9 @@ object SamResource {
   class AppSamResource extends SamResource[AppSamResourceId] {
     def resourceIdAsString(r: AppSamResourceId): String = r.resourceId
     def resourceType(r: AppSamResourceId): SamResourceType = r.resourceType
-    def policyNames(r: AppSamResourceId): Set[SamPolicyName] = r.resourceType match {
-      case SamResourceType.SharedApp => Set(SamPolicyName.Owner, SamPolicyName.User)
-      case _                         => Set(SamPolicyName.Creator, SamPolicyName.Manager)
-    }
+    // Union of shared and private kubernetes-app roles
```
LizBaldo: Why this change? Private app should not have the same set of policy names as shared apps no?

Dockerfile
rtitle: Ah, because the hail chart originally had a relay listener deployment for testing prior to Leo integration. In Leo though the relay listener is deployed by the setup chart (so no need for the app chart to be aware of it). I'll ask the Hail team to remove this for the next revision..

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/samModels.scala
rtitle: Yup, just need to page it back into memory. 😵‍💫 

This came up because I was trying to spin up a shared hail-batch app for the Hail team to use for testing, and noticed some strangeness with Terra UI calling `listAppsV2`. But it's a _little_ out of scope because we're not really planning on making Hail Batch a shared app, for the MVP anyway.

It relates to the algorithm we're using for `listApps`:
```
get apps from Leo DB
get user-visible apps from Sam, filtered by policy
take intersection of above 2 sets
```

In step 3 we're relying on equality of the `AppSamResourceId` type. In Leo we have the `accessScope` field. But in the Sam deserialization it's set to `None`, because the Sam model doesn't really have a concept of access scope. So the equality check was failing, and `listAppsV2` was returning `[]` even though I had a shared app.

The above fix is a Scala trick to make `accessScope` a separate parameter list, which means it isn't included in the case class `equals()` method. I suppose I could have also changed the Sam deserialization logic to set `accessScope` if the resource type is `kubernetes-app-shared`. 

In hindsight I think it would have been better to make a separate type `case class SharedAppSamResourceId(resourceId: String)` instead of adding the `accessScope` field to `AppSamResourceId`. It would have avoided weird errors like this, really hard to predict that in advance though. And _really_, I think we should work toward just rewriting the Leo-Sam integration using the generated Sam Java client, which is probably less error prone and easier to understand.

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
rtitle: That is a good point. I honestly didn't think too hard about this, just kind of copied the approach WDS was using.

It's a little moot because we're not doing any disk management for Azure apps anyway -- e.g. we don't support plugging in an existing disk to an Azure app. So maybe it makes sense not to impose a lot of restrictions just yet. It's a little unfortunate the code is structured in a way to make us define the disk behavior.

rtitle: I guess the only app types which support disk management (on GCP) is `Galaxy` and `Custom`. So maybe the 3 disk categories should be `Galaxy, Custom, Everything Else` (which this code is kind of doing).

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpHailBatchDAO.scala
rtitle: Will do!

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/LeoAuthProvider.scala
rtitle: Yeah this kind of relates to my above Sam comment -- basically when we get the list of `AppSamResourceId`s from Sam, we lose the `accessScope` field. So just testing all the policy names when filtering down the Sam result (this is only used for listing resources).

… As I'm thinking about this more though, I wonder if it would be better to actually fix the Sam deserialization to _not_ lose `accessScope`. Let me take another pass at doing this cleaner…

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/samModels.scala
LizBaldo: Ohh, that makes sense, i did struggle a lot with this when working on the multi-user PR back in Q1 😢 We have a ticket to rewrite the sam-leo integration using the Sam java client for next sprint btw, so I would not do much refactoring here.

rtitle: Oh cool! That sounds good then, can leave it this way in this PR since it is working :)

Dockerfile
rtitle: This is removed now, resolving

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/samModels.scala
rtitle: I actually found (I think) a more natural way of doing this, so resolving this comment for now.

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
rtitle: Updated comment a bit to make this clear

rtitle:
Hey @LizBaldo , finally getting back to this. I applied PR feedback (I think) and cleaned up some of the Sam stuff. Hail team has also iterated several times on the chart version. Wondering if I could get another review pass? Again, I'm thinking of merging this when ready but disabling `HAIL_BATCH` app type in prod until compliance review happens. Thanks!

http/src/main/resources/reference.conf
```diff
@@ -202,7 +202,7 @@ azure {
         minor-version-auto-upgrade = true,
         file-uris = ["https://raw.githubusercontent.com/DataBiosphere/leonardo/1b7690cde14132d6478d5c014341a01b557c6132/http/src/main/resources/init-resources/azure_vm_init_script.sh"]
       }
-      listener-image = "terradevacrpublic.azurecr.io/terra-azure-relay-listeners:8fe77a0"
+      listener-image = "terradevacrpublic.azurecr.io/terra-azure-relay-listeners:decc8b0"
```
rtitle, line 205: This new version includes this bugfix: https://github.com/DataBiosphere/terra-azure-relay-listeners/pull/28

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
```diff
@@ -1145,10 +1145,22 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
   ): F[Vector[ListAppResponse]] = for {
     _ <- as.ask
     samResources = allClusters.flatMap(_.nodepools.flatMap(_.apps.map(_.samResourceId)))
-    samVisibleAppsOpt <- NonEmptyList.fromList(samResources).traverse { apps =>
-      authProvider.filterUserVisible(apps, userInfo)
+    // Partition apps by shared vs. private access scope and make a separate Sam call for each, then re-combine them.
+    // The reason we do this shared vs. private app access scope is represented by different Sam resource types/actions,
+    // and therefore needs different decoders to process the result list..
```
LizBaldo, line 1158: Nice, I wish i had thought of that in my original PR 🙈

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
```diff
@@ -1145,10 +1145,22 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
...
+    partition = samResources.partition(_.accessScope.exists(_ == AppAccessScope.WorkspaceShared))
+    samVisibleSharedAppsOpt <- NonEmptyList.fromList(partition._1).traverse { apps =>
+      authProvider.filterUserVisible(apps, userInfo)(implicitly, sharedAppSamIdDecoder, implicitly)
```
LizBaldo, line 1161: What does the `(implicitly, sharedAppSamIdDecoder, implicitly)` call do? Just curious

LizBaldo (approved):
Thanks for addressing my comments! It is pretty clear what the difference  between the batch and batch-driver is now 👍 
Separating AppResourceId and SharedAppResourceId makes a lot of sense too :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
rtitle: So `filterUserVisible` takes 3 implicit arguments:
```
(implicit
    sr: SamResource[R],
    decoder: Decoder[R],
    ev: Ask[F, TraceId]
  )
```
I want to _explicitly_ pass the `decoder` parameter. Scala doesn't let you pass implicits by name though -- so I can't do this:
```
 authProvider.filterUserVisible(apps, userInfo)(decoder = sharedAppSamIdDecoder)
```
`implicitly` is just a keyword to look up the value from the implicit context (so I avoid having to figure out what the other 2 implicits actually are). I could have also written it this way I guess:
```
    samVisibleSharedAppsOpt <- NonEmptyList.fromList(partition._1).traverse { apps =>
      implicit val d = sharedAppSamIdDecoder
      authProvider.filterUserVisible(apps, userInfo)
    }
```
^ that way I have an implicit `Decoder` in scope so the compiler is happy.

rtitle:
Helmfile change is in to disable this in prod: https://github.com/broadinstitute/terra-helmfile/pull/4105

This PR is rebased -- while tests run I'm also going to do some manual sanity checking in my BEE to make sure it all still works. Then hopefully plan to merge it today.

PR #3258, opened 2023-04-04
[IA-4053] mount azure pd to new working directory

LizBaldo:
In order to enable file uploads persistence when deattaching / reattaching PDs we actually need to format and mount the disk in the first place...

Tested the following successfully on a BEE:
- Created a VM with a new PD
- Uploaded a picture to the `persistent_disk` folder
- Paused and resumed the VM and confirmed the picture was still there
- Deleted the VM without deleting the disk
- Created a new VM with the old disk and the picture was still there


http/src/main/resources/init-resources/azure_vm_init_script.sh
```diff
@@ -152,3 +152,29 @@ jq --null-input \
...
+AllsdDisks=($(/usr/bin/lsblk --nodeps --noheadings --output NAME --paths | grep -i "sd"))
+FreesdDisks=()
+for Disk in "${AllsdDisks[@]}"; do
+    Mounts="$(/usr/bin/lsblk --noheadings --output MOUNTPOINT "${Disk}" | grep -vE "^$")"
+    if [ "${Mounts}" == "" ]; then
+        FreesdDisks+=("${Disk}")
+    fi
+done
+
+DISK_DEVICE_ID= basename ${FreesdDisks[0]}
```
rtitle: missing backtick?
```
DISK_DEVICE_ID=`basename ${FreesdDisks[0]}`
```
Not sure..

http/src/main/resources/init-resources/azure_vm_init_script.sh
```diff
@@ -152,3 +152,29 @@ jq --null-input \
...
+
+## Only format disk is it hasn't already been formatted
+## Maybe check if the working directory exists already?
+if [ ! -d ${WORK_DIRECTORY} ] ; then
```
rtitle: Thinking more, not sure this will work as intended… `/home/jupyter/persistent_disk` will _never_ exist on the VM before you mount the disk. I think this is why on GCE Leo kept track of whether a disk was previously formatted or not, and passed a flag to the script..

rtitle:
Afk a bit, will look more later.

Before I forget: should also test disk remains attached after VM stop/start (deallocate/reallocate). I think the GCE stuff involved modifying `/etc/fstab` which was recommended by docs..

http/src/main/resources/init-resources/azure_vm_init_script.sh
LizBaldo: Yeah, we were thinking about that too, I am thinking about mounting the disk first then trying to check if the directory exist

LizBaldo:
> Afk a bit, will look more later.
> 
> Before I forget: should also test disk remains attached after VM stop/start (deallocate/reallocate). I think the GCE stuff involved modifying `/etc/fstab` which was recommended by docs..

That's a very good point, will test this as well as soon as I can mount it 😅

http/src/main/resources/init-resources/azure_vm_init_script.sh
```diff
@@ -152,3 +152,38 @@ jq --null-input \
 /anaconda/envs/py38_default/bin/jupyter kernelspec list | awk 'NR>1 {print $2}' | while read line; do jq -s add $line"/kernel.json" wsenv.json > tmpkernel.json && mv tmpkernel.json $line"/kernel.json"; done
 /anaconda/envs/azureml_py38/bin/jupyter kernelspec list | awk 'NR>1 {print $2}' | while read line; do jq -s add $line"/kernel.json" wsenv.json > tmpkernel.json && mv tmpkernel.json $line"/kernel.json"; done
 /anaconda/envs/azureml_py38_PT_and_TF/bin/jupyter kernelspec list | awk 'NR>1 {print $2}' | while read line; do jq -s add $line"/kernel.json" wsenv.json > tmpkernel.json && mv tmpkernel.json $line"/kernel.json"; done
+
+#echo "Formatting and mounting persistent disk..."
```
LizBaldo: I am a bit puzzled as to why I am now running into the following error when trying to create a VM with this script on my BEE:

```
ListKernelSpecs] WARNING | Config option `kernel_spec_manager_class` not recognized by `ListKernelSpecs`. 
Unable to load extension: pydevd_plugins.extensions.types.pydevd_plugin_pandas_types
Unable to load extension: pydevd_plugins.extensions.types.pydevd_plugin_pandas_types
```

http/src/main/resources/init-resources/azure_vm_init_script.sh
LizBaldo: I do not think so? I will know as soon as i can get this part of the script to actually run and not fail on the pre-existing steps 😡

http/src/main/resources/init-resources/azure_vm_init_script.sh
LizBaldo: This was a red herring from the vm extension script, the error is indeed somewhere in the added changes, see the wonderful suite of investigative commits below...

LizBaldo:
I think this is a must for reviewers of this PR... https://www.destroyallsoftware.com/talks/wat
Bash is obviously the best language

http/src/main/resources/init-resources/azure_vm_init_script.sh
```diff
@@ -3,6 +3,53 @@ set -e
...
+## The PD should be the only `sd` disk that is not mounted yet
+AllsdDisks=($(lsblk --nodeps --noheadings --output NAME --paths | grep -i "sd"))
+echo ${AllsdDisks[@]}
+FreesdDisks=()
+echo ${FreesdDisks[@]}
+for Disk in "${AllsdDisks[@]}"; do
+    echo $Disk
+    Mounts="$(lsblk --noheadings --output MOUNTPOINT "${Disk}" | grep -vE "^$")"
+    echo $Mounts
+    echo "${Mounts[0]}"
```
LizBaldo: I might need the help of a bash or azure vm custom script wizzard. This line (and the rest up to mounting that I cannot perform due to permission issues) works perfectly fine from a VM jupyter terminal but somehow fails in this script

n8landolt: It may be helpful for debugging to make a new machine with port 22 open and run these commands, not sure if you've directly connected to a machine yet.
If you need some help with getting this set up LMK

LizBaldo: I might, thanks! So far I am testing these commands from the jupyter terminal of a working VM in dev

http/src/main/resources/init-resources/azure_vm_init_script.sh
LizBaldo: I changed the logic to trying to mount the PD directly, if the PD has a formatted partition then it has data on it and should be mountable directly. If not then it must be a new disk and we have to set it up from scratch

LizBaldo:
I am reaching the end of what I can do here and would appreciate some help 😢 

On a brand new DSVM launched from the azure portal, I can ssh into it, copy the azure_vm_init.sh script and run it without issues. It does what I want (find the PD, attempts to mount it and if this is not possible then partition and format first).

However when creating via the terra UI on my BEE the script fails on line 19 (the if statement checking if the mount is empty). If I remove `set -e` from the script then provisioning the extension succeeds and the VM is successfully running in the Azure portal. BUT it stays stuck in `CREATING` on terra UI and ends up failing because of a timeout issue.

rtitle:
Hey @LizBaldo i can help look more tomorrow, I have very few meetings so should be able to focus

LizBaldo:
FYI @rtitle the VM provision succeeded with my latest commit. The trick now is to actually test file upload and persistence on a BEE :)

rtitle:
Oh so mounting to `/home/jupyter` didn't work? Actually that makes sense, I think the mount dir may need to be empty (e.g. not have pre-existing files)?

LizBaldo:
For Monday's eyes, but this works great on my BEE and should be ready for your 👍

LizBaldo:
Now everything really works for real. The PD UUID is properly exposed, and the data disk is remounted when we reboot.

LizBaldo:
Once you give me a 👍 I will make the links to the script permanent with the git sha, leaving it to the base branch right now in case you have changes to request (it is much faster to test that way)

http/src/main/resources/reference.conf
```diff
@@ -200,7 +200,7 @@ azure {
         type = "CustomScript",
         version = "2.1",
         minor-version-auto-upgrade = true,
-        file-uris = ["https://raw.githubusercontent.com/DataBiosphere/leonardo/b34c5b0807d5b6ff50d9f2c25328fae53c6b4b96/http/src/main/resources/init-resources/azure_vm_init_script.sh"]
+        file-uris = ["https://raw.githubusercontent.com/DataBiosphere/leonardo/IA-4053-mount-existing-pd-to-azure-vm/http/src/main/resources/init-resources/azure_vm_init_script.sh"]
```
n8landolt: I think this path isn't permanent - I think I remember trying to do this previously and it eventually broke.
If you clicked "permalink" to get this link then I am mistaken.

LizBaldo: Yes, this is temporary while I debug, I will switch to the permanent pass before merging

lucymcnatt (approved):

PR #3261, opened 2023-04-04
IA-3955: removing workspaceId from disk v2 route

lucymcnatt:
workspaceId isn't required in the initial logic for the deleteDisk v2 route, so this removes it from the path

---

In all cases:


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskV2ServiceInterp.scala
```diff
@@ -120,3 +123,10 @@ case class DiskCannotBeDeletedAttachedException(id: DiskId, workspaceId: Workspa
       StatusCodes.Conflict,
       traceId = Some(traceId)
     )
+
+case class DiskWithoutWorkspaceException(id: DiskId, traceId: TraceId)
```
n8landolt, line 127: I assume this scenario should never happen?

Is the reason for this so a user can't delete a disk they're not allowed to?

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskV2ServiceInterpSpec.scala
```diff
@@ -192,9 +192,36 @@ class DiskV2ServiceInterpSpec extends AnyFlatSpec with LeonardoTestSuite with Te
...
         )
       )
-      err <- diskV2Service.deleteDisk(userInfo, workspaceId, disk.id).attempt
+      err <- diskV2Service.deleteDisk(userInfo, disk.id).attempt
     } yield err shouldBe Left(DiskCannotBeDeletedAttachedException(disk.id, workspaceId, ctx.traceId))
 
     res.unsafeRunSync()(cats.effect.unsafe.IORuntime.global)
   }
+
+  it should "fail to delete a disk if it has no workspaceId" in isolatedDbTest {
```
n8landolt, line 201: Is it possible to create a disk without a workspaceId?

n8landolt: Adding in this logic concerns me, it seems to be just another layer of obscure functionality that restricts user control - if this scenario happens, manual intervention is required.

lucymcnatt: with an Azure disk no, but with a GCP runtime yes. The workspaceId column in the PERSISTENT_DISK table is an option, so in order to get the workspaceId I used that fromOption logic. I don't think this exception should happen very often/at all, but thought it might be useful when we switch over to using the v2 routes for GCP as well.

n8landolt: Ahh it seems like it's used by a dependency - WSM.
In Leo it seems like workspaceId is strictly for logging, but once we make the WSM delete request it needs the workspaceId

env/local.env
```diff
@@ -101,3 +101,6 @@ LEO_EXECUTION_MODE=backLeo
 
 # Akka logging
 AKKA_LOG_LEVEL=INFO
+
+# Java
+JAVA_HOME=java.home
```
n8landolt, line 106: Is this change intentional?

http/src/main/resources/leo.conf
```diff
@@ -123,7 +123,7 @@ akka {
         },
         {
           # Default trust store
-          path: ${JAVA_HOME}/lib/security/cacerts
+          path: ${?JAVA_HOME}/lib/security/cacerts
```
n8landolt, line 126: Same with this - is this intentional?

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskV2ServiceInterpSpec.scala
lucymcnatt: yup yeah that's why its needed in the backend call, every call to WSM for azure disks, runtimes etc requires a workspaceId

http/src/main/resources/leo.conf
lucymcnatt: ohh yeah, I can get rid of them cuz they're out of the scope of this change, but leo tests don't run locally without JAVA_HOME being defined

n8landolt: I'm okay with the java changes going in - it doesn't affect anything other than the local tests.

n8landolt (approved):

lucymcnatt:
jenkins retest

PR #3281, opened 2023-04-13
AJ-939, IA-4193 source workspace id parameter

calypsomatic:
Note: changes to `Dockerfile` and `reference.conf` are temporary for testing and will not be a part of the final PR.
Changes for [IA-4193](https://broadworkbench.atlassian.net/browse/IA-4193) in support of [AJ-939](https://broadworkbench.atlassian.net/browse/AJ-939): WDS needs to receive a `sourceWorkspaceId`. This adds the `sourceWorkspaceId` as a parameter to the `createApp` API, and then adds it to the helm chart overrides (if present).

[IA-4193]: https://broadworkbench.atlassian.net/browse/IA-4193?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[AJ-939]: https://broadworkbench.atlassian.net/browse/AJ-939?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -3858,6 +3858,9 @@ components:
           items:
             type: string
           description: Extra arguments to pass to the application. Only used if appType is CUSTOM.
+        sourceWorkspaceId:
+          type: string
+          description: Id of source workspace if app is a clone.
```
LizBaldo: Maybe say that this is optional

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -116,7 +116,7 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
       )
 
       _ <- logger.info(ctx.loggingCtx)(
-        s"Begin app creation for app ${params.appName.value} in cloud context ${params.cloudContext.asString}"
+        s"Begin app creation for app ${params.appName.value} in cloud context ${params.cloudContext.asString} with variables ${app.customEnvironmentVariables}"
```
LizBaldo: Looks like a leftover from the previous PR, but it really can't hurt to add this to the logs 👍

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -510,8 +512,14 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
         // general configs
         raw"fullnameOverride=$appChartPrefix-${release.asString}",
         raw"instrumentationEnabled=${config.coaAppConfig.instrumentationEnabled}"
-      ).mkString(",")
-    )
+      )
+
+    val updatedLs = sourceWorkspaceId match { // TODO move to buildWdsChartOverrideValues after WDS chart migration
+      case Some(value) => ls ::: List(raw"config.sourceWorkspaceId=${value.value}")
```
LizBaldo: That makes sense to me, mad at scala for having list concatenation defined as ::: though 😡

LizBaldo (commented):
LGTM! Left very minor comments, ping me when you want a final 👍

LizBaldo:
@davidangb do you need me to review something specific again? I did not give a 👍 yet given that the PR is a draft.

davidangb:
@LizBaldo whoops sorry! That was a misclick, please ignore that.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -466,9 +468,10 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
                                                      storageContainer: StorageContainerResponse,
                                                      batchAccountKey: BatchAccountKey,
                                                      applicationInsightsConnectionString: String,
-                                                     appChartPrefix: String
-  ): Values =
-    Values(
+                                                     appChartPrefix: String,
+                                                     sourceWorkspaceId: Option[WorkspaceId]
+  ): Values = {
+    val ls =
```
aaronkanzer: Could we use a more descriptive variable name than `ls`?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -510,8 +513,14 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
         // general configs
         raw"fullnameOverride=$appChartPrefix-${release.asString}",
         raw"instrumentationEnabled=${config.coaAppConfig.instrumentationEnabled}"
-      ).mkString(",")
-    )
+      )
+
+    val updatedLs = sourceWorkspaceId match { // TODO remove after WDS chart migration
+      case Some(value) => ls ::: List(raw"provenance.sourceWorkspaceId=${value.value}")
```
aaronkanzer: Trying to track this down, but where does `provenance` live in our Helm chart values exactly?

aaronkanzer: ah jk -- found it in one of your other PRs... 🤦

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
calypsomatic: Sure why not

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/AppComponent.scala
```diff
@@ -204,7 +207,8 @@ object appQuery extends TableQuery(new AppTable(_)) {
         diskOpt.map(_.id),
         if (saveApp.app.customEnvironmentVariables.isEmpty) None else Some(saveApp.app.customEnvironmentVariables),
         saveApp.app.descriptorPath,
-        if (saveApp.app.extraArgs.isEmpty) None else Some(saveApp.app.extraArgs)
+        if (saveApp.app.extraArgs.isEmpty) None else Some(saveApp.app.extraArgs),
+        if (saveApp.app.sourceWorkspaceId.isEmpty) None else saveApp.app.sourceWorkspaceId
```
davidangb: I think this is redundant. If `saveApp.app.sourceWorkspaceId.isEmpty` is true, then `saveApp.app.sourceWorkspaceId` will be a `None`. So you can just use:
```suggestion
        saveApp.app.sourceWorkspaceId
```
similar to how `saveApp.app.descriptorPath` is specified at line 209.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -514,8 +517,14 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
...
         raw"fullnameOverride=$appChartPrefix-${release.asString}",
         raw"instrumentationEnabled=${config.coaAppConfig.instrumentationEnabled}"
-      ).mkString(",")
-    )
+      )
+
+    val updatedLs = sourceWorkspaceId match { // TODO remove after WDS chart migration
+      case Some(value) => valuesList ::: List(raw"provenance.sourceWorkspaceId=${value.value}")
+      case None        => valuesList
+    }
```
davidangb, line 525: This is perfectly fine as-is, no need to change anything. For the sake of discussing Scala goodness, some thoughts:

You could rewrite this using `.map`, which is a bit more functional-programming-esque than an explicit `match` statement here:
```scala
val updatedLs = valuesList ++ sourceWorkspaceId.map(v => raw"provenance.sourceWorkspaceId=${v.value}").toList
```
This also relies on the behavior of `Option.toList` to create an empty list for None and a single-value list for Some.

~Also, beware the difference between `:::` and `++` when concatenating lists. The former prepends the right-hand value to the list, while the latter appends. In this particular case I don't think it matters at all, but in other cases it certainly could, and might lead to confusion. At lists of this size the performance impact should be nonexistent.~

EDIT: this last paragraph, now stricken, is incorrect … please ignore!

davidangb (commented):
minor comments inline, looking good!

I don't have a great sense in Leo if there's anything you missed - the code changes here look good.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
rtitle: +1 David's suggestion is a bit more idiomatic, but also ok as is.

Does the helm chart handle no value for `provenance.sourceWorkspaceId` if the app is not a clone? E.g. it doesn't need some default value?

rtitle (approved):
LGTM! Thanks for making the DB migration too

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
calypsomatic: In my testing, it was injecting the placeholder value - i.e. "RUNTIME_PARAMETER", so I used a blank string "".  I'm happy to hear a better solution.

LizBaldo (approved):
LGTM! :shipit:

davidangb (approved):
lgtm, nice job!

PR #3289, opened 2023-04-18
IA-4192: passing userAccessToken to helm values

lucymcnatt:
Paired with: https://github.com/broadinstitute/cromwhelm/pull/234
Add the userAccessToken to WDS and Cromwell helm charts, required to clone apps.

---

In all cases:


lucymcnatt:
Right now the userAccessToken is the same for all apps created by a user (across workspaces) and expires 1hr after the user authorizes, are people ok with that behavior?

http/src/main/resources/leo.conf
```diff
@@ -99,6 +99,7 @@ akka {
     client.idle-timeout = 1 hour
     parsing.max-to-strict-bytes = 128m
     server.parsing.max-content-length = 128m
+    server.parsing.max-uri-length = 16k
```
lucymcnatt: this was just to make authorization work on my BEE, we have already have a PR going in to fix

Dockerfile
```diff
@@ -58,13 +58,14 @@ RUN helm repo add ingress-nginx https://kubernetes.github.io/ingress-nginx && \
 # pulling `terra-app-setup` locally and add cert files to the chart.
 # Leonardo will install the chart from local version.
 # We are also caching charts so they are not downloaded with every helm-install
+COPY ./cromwell-on-azure-0.2.219.tgz /leonardo
+RUN tar -xzf /leonardo/cromwell-on-azure-0.2.219.tgz -C /leonardo
```
lucymcnatt: for testing purposes

http/src/main/resources/leo.conf
LizBaldo: I think Lily just merged it so you should be able to merge develop in

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -231,6 +232,7 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
                   storageContainer,
                   BatchAccountKey(batchAccountKey),
                   applicationInsightsComponent.connectionString(),
+                  params.userAccessToken,
```
LizBaldo: Since this is a temporary solution, I think it would be nice to put a TODO before this line (as well as in the WDS case below to warn future devs.

TODO: Remove once permanent solution utilizing the multi-user sam app identity has been implemented

LizBaldo (approved):
LGTM! I only have one small request to add a TODO to remove this once the permanent solution is implemented. There might be a small typo in the helm values passed in the unit test, but that should be a relatively easy fix, lmk if you need help.

If AJ folks are good with using this user info token then :shipit:

http/src/main/resources/leo.conf
rtitle: Ah so the length of the swagger auth url was hitting this? 🤦 Makes sense

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/leoPubsubMessageSubscriberModels.scala
```diff
@@ -569,10 +570,11 @@ object LeoPubsubCodec {
     Decoder.forProduct2("name", "resourceId")(StorageContainerResponse.apply)
 
   implicit val createAppV2Decoder: Decoder[CreateAppV2Message] =
-    Decoder.forProduct7("appId",
+    Decoder.forProduct8("appId",
                         "appName",
                         "workspaceId",
                         "cloudContext",
+                        "userAccessToken",
```
rtitle: Just mentioning: another way to do this besides passing the original user-token in the PubSub message would be to mint a user-pet token in `AKSInterpreter` using `samDao.getCachedArbitraryPetAccessToken`. We are already using that method in `AKSInterpreter` for other things:

https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala#L367

I guess benefits of that would be (a) no need to pass a sensitive token through PubSub; and (b) the minted token would always have a lifetime of 60 minutes; while the user-token might have a lifetime of <60 minutes, since the refresh is handled by the UI.

rtitle (approved):
I suggested another approach inline -- but I don't feel too strongly as this should work too, so approving either way.

lucymcnatt:
jenkins retest

PR #3291, opened 2023-04-18
[DDO-2761] Have Leonardo's Dockerfile CMD use bash

jack-r-warren:
As discussed in [this thread](https://broadinstitute.slack.com/archives/CA3NP1733/p1681835775202119). Leonardo's CMD is silently dropping environment variables with periods because it's using `sh` under the hood (because it's using the "shell" form of CMD). You can test the drop behavior by running the following two commands from the root of Leo's repo:

```bash
docker run --env-file ./env/local.env -it ubuntu /bin/bash -c 'printenv | grep VALID'
```

```bash
docker run --env-file ./env/local.env -it ubuntu /bin/sh -c 'printenv | grep VALID'
```

We do still need to handle $JAVA_OPTS and Leo's variable jar name, so we'll use the [(Docker-recommended) "exec" form of CMD](https://docs.docker.com/engine/reference/builder/#cmd) with `bash` instead.

I deployed this to my BEE and it made Leo start respecting the FRAME_ANCESTORS.* and VALID_HOSTS.* environment variables that were being set, and I think that's about as good of a test as we'll get. Going forward, if this breaks when Leo is running on foundation, we'll see a bunch of errors very early on in lower environments when Leonardo starts rejecting every single referer and iframe.

---

> There are none!

In all cases:

- [x] Verify all tests go green (you can re-run automation tests with a comment saying `jenkins retest`
- [x] Run the automation tests multiple times in parallel to weed out instability if applicable via a comment saying `multi-test`
- [x] Squash and merge; Delete your branch after this


Dockerfile
```diff
@@ -78,4 +78,9 @@ RUN curl -fsSL -o /terra-docker-versions-candidate.json \
     https://storage.googleapis.com/terra-docker-image-documentation/terra-docker-versions-candidate.json
 
 # Add Leonardo as a service (it will start when the container starts)
-CMD java $JAVA_OPTS -jar $(find /leonardo -name 'leonardo*.jar')
+# 1. "Exec" form of CMD necessary to avoid `sh` stripping environment variables with periods in them,
+#    used for Lightbend config
+# 2. $JAVA_OPTS and filesystem like /leonardo/leonardo*.jar both necessary as long as Leonardo runs on 
+#    Kubernetes without foundation (firecloud-develop requires former, old chart requires latter)
+# We use the "exec" form but call `bash` to accomplish both 1 and 2
```
LizBaldo, line 85: 🤯

LizBaldo (approved):
Whoa, thanks for fixing that 🤯 
:shipit:

em-may (approved):

PR #3292, opened 2023-04-18
[IA-4252] remove disk_config from default JSON for creating an Azure app

mlilynolting:
Remove the default disk_config from createAppV2. Verified locally and on my BEE.


http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -1676,7 +1676,7 @@ paths:
           schema:
             type: string
       requestBody:
-        $ref: "#/components/requestBodies/CreateAppRequest"
+        $ref: "#/components/requestBodies/CreateAzureAppRequest"
```
jdcanas: why are we further diverging V2 request models from being non Azure/google specific

mlilynolting: Is that an ideal toward which I should be working? Would renaming the new terms with "V2" instead of "Azure" be preferable?

mlilynolting: Went ahead and renamed

jdcanas (approved):

http/src/main/resources/swagger/api-docs.yaml
jdcanas: Yeah, was a little unclear as to whether the model itself was azure-specific, so didn't want to be prescriptive. In general, I'd say we want to make v2 routes as non-azure-specific as possible as the eventual goal is to remove the google-project-dependent v1 routes and use workspaceId for all resources (which is why we can't make them completely agnostic today)

LizBaldo (approved):

PR #3293, opened 2023-04-18
[RW-9951][IA-4269] update custom env var type to allow larger char limit

yonghaoy:

---

In all cases:

### DIFF
--- a/http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20230418_update_custom_env_length.xml
+++ b/http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20230418_update_custom_env_length.xml
@@ -1,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="leonardo" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+    <changeSet logicalFilePath="leonardo" author="yonghaoy" id="update_custom_env_length">
+        <modifyDataType
+                columnName="customClusterEnvironmentVariables"
+                newDataType="text"
+                tableName="CLUSTER"/>
+        <modifyDataType
+                columnName="customClusterEnvironmentVariables"
+                newDataType="text"
+                tableName="APP"/>
+    </changeSet>
+</databaseChangeLog>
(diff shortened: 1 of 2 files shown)


http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20230418_update_custom_env_length.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="leonardo" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+    <changeSet logicalFilePath="leonardo" author="yonghaoy" id="update_custom_env_length">
+        <modifyDataType
+                columnName="customClusterEnvironmentVariables"
+                newDataType="text"
```
LizBaldo: Instead of changing to `text`, could you keep varchar, but increase to the desired size?
e.g. `newDataType="varchar(4064)"`

yonghaoy: I throught using text will give us a lot more space and should be good enough moving forward.
If keep using varchar but increase to 4064, I am concerned we will hit this error again as we will kept adding env vars into runtime to help user manage their dataset asset. 
Not DB expert, what is the drawbacks of changing from `varchar` to `text`?

LizBaldo: varchar is stored inline and usually faster to access than text

LizBaldo: You can increase varchar to the size you need, I was just giving you an example

yonghaoy: I see. Doing some research. Is [this post](https://stackoverflow.com/questions/1476987/how-to-store-more-than-255-char-in-mysql-database) correct? 
Leo is using mysql 5.0.3 or higher, and I can just set to `varchar(65535)` ?

LizBaldo: We are using 5.6 so this should work yes, the only other reason why I am picky is that other stakeholders are using this column as well, I do not see why changing it to text would make this noticeably slower for others but I would like to make the smallest possible change here

yonghaoy: SG! Thanks

yonghaoy:
Unit test fails at 
```
iquibase.exception.LiquibaseException: liquibase.exception.MigrationFailedException: Migration failed for changeset leonardo::update_custom_env_length::yonghaoy:
     Reason: liquibase.exception.DatabaseException: Row size too large. The maximum row size for the used table type, not counting BLOBs, is 65535. This includes storage overhead, check the manual. You have to change some columns to TEXT or BLOBs [Failed SQL: (1118) ALTER TABLE leotestdb.CLUSTER MODIFY customClusterEnvironmentVariables VARCHAR(65535)]
```
trying smaller values

LizBaldo (approved):
LGTM! If you keep being forced to lower the value to something you think will be way too small, then I am fine reverting to text, but would like to avoid it if possible

yonghaoy:
> LGTM! If you keep being forced to lower the value to something you think will be way too small, then I am fine reverting to text, but would like to avoid it if possible

Still finalizing the character we would need. Currently we just about to above 2048 character limit so we are thinking just to use 8128 for now. It's less likely we will hit that limit, or we may just hit restful API size limit first... 
Does that SG?

yonghaoy:
jenkins retest

PR #3280, opened 2023-04-13
IA-3961: hybrid connection name includes workspaceId

lucymcnatt:
Adding workspaceId to the relayHybridConnectionName for apps
Adding the name in the customEnvironmentVariables on app creation so that new apps with this custom var will delete the correct relayHybridConnection

---
Have you read [CONTRIBUTING.md](https://github.com/DataBiosphere/leonardo/blob/develop/CONTRIBUTING.md) lately? If not, do that first.

I, the developer opening this PR, do solemnly pinky swear that:

- [ ] I've documented my API changes in Swagger

In all cases:

- [ ] Get a thumbsworth of review and PO signoff if necessary
- [ ] Verify all tests go green (you can re-run automation tests with a comment saying `jenkins retest`
- [ ] Run the automation tests multiple times in parallel to weed out instability if applicable via a comment saying `multi-test`
- [ ] Squash and merge; Delete your branch after this
- [ ] Test this change deployed correctly and works on dev environment after deployment

PR #3307, opened 2023-04-24
IA-3961: hybrid connection name includes workspaceId (2nd try)

lucymcnatt:
My [original PR](https://github.com/DataBiosphere/leonardo/pull/3280) caused errors because the incorrect proxyURL was being generated, [thread here](https://broadinstitute.slack.com/archives/C02GJ782BAA/p1682356982603629)

This reverts the reverse commit c8f1292698db194169e69c0465670b6fde1355c1 and adds changes to the getProxyURLs function to make sure those are correct with the naming changes.

__Jira ticket__: https://broadworkbench.atlassian.net/browse/[IA-3961]

## Dependencies 
[original PR](https://github.com/DataBiosphere/leonardo/pull/3280) 

## Summary of changes

This reverts the reverse commit c8f1292698db194169e69c0465670b6fde1355c1 and adds changes to the getProxyURLs function to make sure those are correct with the naming changes.

### What

Changes the relayHybridConnection name to have the workspaceId as well as the appId.
### Why

Because the appId is not unique across workspaces, users can get access to other apps data

### What to test

- create a new workspace, create an Azure VM and open a jupyter notebook

### Who tested and where

- [ ] I validated this change 

[IA-3961]: https://broadworkbench.atlassian.net/browse/IA-3961?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
@@ -394,12 +394,17 @@ final case class App(id: AppId,
       // A service can optionally define a path; otherwise, use the name.
       val leafPath = service.config.path.map(_.value).getOrElse(s"/${service.config.name.value}")
       // GCP uses a Leo proxy endpoint: e.g. https://notebooks.firecloud.org/google/v1/apps/{project}/{app}/{service}
-      // Azure uses Azure relay: e.g. https://{namespace}.servicebus.windows.net/{app}/{service}
+      // Azure uses Azure relay: e.g. https://{namespace}.servicebus.windows.net/{app}-{workspaceId}/{service}
       val proxyPathOpt = cluster.cloudContext match {
         case CloudContext.Gcp(project) =>
           Some(s"${proxyUrlBase}google/v1/apps/${project.value}/${appName.value}${leafPath}")
         case CloudContext.Azure(_) =>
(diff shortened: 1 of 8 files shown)


rtitle (approved):
Thanks - it seems like this will fix the earlier issue. Should test the Leo code changes against pre-existing and new apps (to make sure both naming schemes still work).

I was wondering if there is any thought about migrating old apps to use the new naming scheme after this PR is in production? I guess migration would be a little tricky because we'd not only have to rename the hybrid connection; but also probably redeploy the listener with new values..

lucymcnatt:
> Thanks - it seems like this will fix the earlier issue. Should test the Leo code changes against pre-existing and new apps (to make sure both naming schemes still work).
> 
> I was wondering if there is any thought about migrating old apps to use the new naming scheme after this PR is in production? I guess migration would be a little tricky because we'd not only have to rename the hybrid connection; but also probably redeploy the listener with new values..

Yeah I'll retest against both app types, to your second point, the migration is a bit tricky, I'm not sure how we'd manage it effectively without user disruption, but open to suggestions!

lucymcnatt:
tested with old and new naming conventions + creating new workspace and VM in the UI, everything seems to be functioning normally

lucymcnatt:
jenkins retest

PR #3308, opened 2023-04-25
[IA-4272] modify disk attachment logic for azure

jdcanas:
Main problem is we didn't correctly handle the concept of 'attached' for azure.

This is not heavily tested in a bee since it was a bit hard to reproduce consistently, but this field wasn't being used properly as is. Honestly, not super sure what is happening. The RUNTIME_CONFIG table uses `persistentDiskId` as a FOREIGN KEY referencing the PERSISTENT_DISK tables primary key `id`, but the RUNTIME_CONFIG table does not enforce this constraint. Indeed, this was not an issue for GCP as we enforce it in the app. However, the logic for azure disks violated this constraint and now there are multiple instances of the same `persistentDiskId` for multiple runtimes (because we viewed those attached to deleted runtimes as valid). This isn't that big of a problem moving forward, the logic handles it, but its impossible to enforce this constraint now that it is violated in prod without messy manual database manipulation risking user disks.  

Still need to do BEE testing, but unit tests exercise all logic introduced, so ready for some eyes.

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
@@ -528,7 +528,7 @@ object clusterQuery extends TableQuery(new ClusterTable(_)) {
       .result
       .headOption
 
-  def getLastClusterWithDiskId(diskId: DiskId)(implicit ec: ExecutionContext): DBIO[Option[ClusterRecord]] =
+  def getClusterWithDiskId(diskId: DiskId)(implicit ec: ExecutionContext): DBIO[Option[ClusterRecord]] =
     // get most recently created runtime with the specified persistent disk
     for {
       runtimeConfigId <- runtimeConfigs
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskV2ServiceInterp.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskV2ServiceInterp.scala
@@ -98,11 +98,10 @@ class DiskV2ServiceInterp[F[_]: Parallel](config: PersistentDiskConfig,
       workspaceId <- F.fromOption(disk.workspaceId, DiskWithoutWorkspaceException(diskId, ctx.traceId))
 
       // check that disk isn't attached to a runtime
-      previousRuntime <- clusterQuery.getLastClusterWithDiskId(diskId).transaction
-      _ <- previousRuntime.traverse(runtime =>
-        F.raiseError[Unit](DiskCannotBeDeletedAttachedException(diskId, workspaceId, ctx.traceId))
-          .whenA(runtime.status != RuntimeStatus.Deleted)
-      )
+      isAttached <- persistentDiskQuery.isDiskAttached(diskId).transaction
+      _ <- F
+        .raiseError[Unit](DiskCannotBeDeletedAttachedException(diskId, workspaceId, ctx.traceId))
+        .whenA(isAttached)
 
       _ <- persistentDiskQuery.markPendingDeletion(disk.id, ctx.now).transaction
 
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeServiceInterp.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeServiceInterp.scala
@@ -1204,7 +1204,7 @@ final case class DiskNotSupportedException(traceId: TraceId)
 
 final case class DiskAlreadyAttachedException(cloudContext: CloudContext, name: DiskName, traceId: TraceId)
     extends LeoException(
-      s"Persistent disk ${cloudContext.asStringWithProvider}/${name.value} is already attached to another runtime",
+      s"Your persistent disk ${cloudContext.asStringWithProvider}/${name.value} is already attached to another runtime. You might need to wait a few minutes if you just deleted a runtime.",
       StatusCodes.Conflict,
       traceId = Some(traceId)
     )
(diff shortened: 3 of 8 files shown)


LizBaldo:
Should we chat with @lucymcnatt? I understand the issue you found here, but I am not sure I am grasping how this would fix the bug experienced by the user (aka we let them create a new VM too early after deleting the previous one and the PD is not yet available)

jdcanas:
> Should we chat with @lucymcnatt? I understand the issue you found here, but I am not sure I am grasping how this would fix the bug experienced by the user (aka we let them create a new VM too early after deleting the previous one and the PD is not yet available)

The solution itself is fairly simple, and the same way we do it for GKE. Here, we make sure we label when a disk is no longer attached at the proper in point in backend monitoring https://github.com/DataBiosphere/leonardo/pull/3308/files#diff-e140eb1deb8547995938a18b2bea7c0a97756d55ca5f62d3fa7592a99faabaa4R621, and [here](https://github.com/DataBiosphere/leonardo/pull/3308/files#diff-e140eb1deb8547995938a18b2bea7c0a97756d55ca5f62d3fa7592a99faabaa4R621) we actually check that field in the frontend, which we weren't not doing before (disks are already labeled properly as attached via the field in the RUNTIME_CONFIG table, which `completeDeletion` handles in the first link).

In other words, we eagerly kick back in the UI from front leo if the user tries to use an unattached disk (with the same messanging we do for google), instead of submitting it to backleo to send it to WSM an WSM eventually kicking back (the backend polling makes sure we dont mark a disk as unattached until WSM is finished with any deletion jobs)

lucymcnatt:
To Liz's point, while this may fix an issue in how we're handing attachment in Leo, the error we're seeing in the ticket is from the Azure side. The issue being that the VM deletion in Azure is sometimes slow enough that it causes a conflict on re-attachment.

jdcanas:
> To Liz's point, while this may fix an issue in how we're handing attachment in Leo, the error we're seeing in the ticket is from the Azure side. The issue being that the VM deletion in Azure is sometimes slow enough that it causes a conflict on re-attachment.

Yes, before though, for azure disks, it was considered unattached as soon as the previous runtime was marked 'deleting' from the deletion request in front leo. Now we do not mark the disk as unattached until the processing in the back leo has completed (Aka runtime marked 'deleted')

LizBaldo (commented):
Okay, I think I get it, instead of getting a WSM failure after a few minutes, this would fail a lot faster by getting an error from leo directly saying that this disk is still attached.

Does this mean that this should get a sister PR in the UI to make the message more user friendly? (e.g. Persistent disk is still attached to the previous VM, please wait a few minutes before trying again)

LizBaldo:
Or better, poll the persistent disk status (with a reasonable timeout) and only send the createRuntime request when it is ready, abstracting all of that disk attachment logic away from the user

jdcanas:
> Okay, I think I get it, instead of getting a WSM failure after a few minutes, this would fail a lot faster by getting an error from leo directly saying that this disk is still attached.
> 
> Does this mean that this should get a sister PR in the UI to make the message more user friendly? (e.g. Persistent disk is still attached to the previous VM, please wait a few minutes before trying again)

I  can modify the message leo returns that the UI displays to mention this error could possibly require them to wait and try again (but note not all instances of this error will necessarily mean that)

jdcanas:
> Or better, poll the persistent disk status (with a reasonable timeout) and only send the createRuntime request when it is ready, abstracting all of that disk attachment logic away from the user

Hmm. I don't think I agree with this approach.

This is a terra internal service complaining it is still attached in the error message. We now wait for any WSM jobs associated with the vm it is complaining is still attached in the initial error message, which is the only pollable thing we have with regard to disk attachment. Therefore, this fix is likely to solve any instances of this in the wild

LizBaldo:
Sounds good, happy to help test on a BEE if you want to share the testing steps

jdcanas:
It took a while to test in a bee, but I finally re-created the issue in dev, then deleted/recreated several VMs in my bee with these chages and did not encounter the issue.

LizBaldo:
Cool, let me give it a try on my BEE :)

LizBaldo:
So I just tested on my BEE, it looks better now, in the sense that the error pops up immediately and makes more sense:
```
{
  "source": "leonardo",
  "message": "Persistent disk Azure/0cb7a640-45a2-4ed6-be9f-63519f86e04b/3efc5bdf-be0e-44e7-b1d7-c08931e3c16c/mrg-terra-dev-previ-20230214105302/saturn-pd-452bc30e-1d06-46c9-8e02-ea170f884d62 is already attached to another runtime",
  "statusCode": 409,
  "exceptionClass": "class org.broadinstitute.dsde.workbench.leonardo.http.service.DiskAlreadyAttachedException",
  "traceId": "501484e7-c7cc-45a0-847a-20138ee1d6c4"
}
```

It still takes a few minutes / 4 or 5 retries for this error to disappear and the VM to actually create, which is not a great user experience, so I think we should pair this with a `please wait a few minutes before retrying` message

jdcanas:
> So I just tested on my BEE, it looks better now, in the sense that the error pops up immediately and makes more sense:
> 
> ```
> {
>   "source": "leonardo",
>   "message": "Persistent disk Azure/0cb7a640-45a2-4ed6-be9f-63519f86e04b/3efc5bdf-be0e-44e7-b1d7-c08931e3c16c/mrg-terra-dev-previ-20230214105302/saturn-pd-452bc30e-1d06-46c9-8e02-ea170f884d62 is already attached to another runtime",
>   "statusCode": 409,
>   "exceptionClass": "class org.broadinstitute.dsde.workbench.leonardo.http.service.DiskAlreadyAttachedException",
>   "traceId": "501484e7-c7cc-45a0-847a-20138ee1d6c4"
> }
> ```
> 
> It still takes a few minutes / 4 or 5 retries for this error to disappear and the VM to actually create, which is not a great user experience, so I think we should pair this with a `please wait a few minutes before retrying` message

the better message is included in a later commit, forgot to push 😅

lucymcnatt (approved):

LizBaldo (approved):
Thanks for fixing this! We should figure out a way of elegantly handling the error for the user

jdcanas:
jenkins retest

PR #3309, opened 2023-04-25
[IA-4169] Fix improperly configured dataproc machines.

n8landolt:
__Jira ticket__: https://broadworkbench.atlassian.net/browse/IA-4169

## Summary of changes

Hail/Dataproc users are finding their machine's memory isn't being configured correctly.

### What

- Add spark.driver.memory to the Software Config builder
- Add a DataProcMachineSpecification type for mapping MachineTypeName to / from CPU and Memory.

### Why

- dataproc users are unable to properly run their data analyses.

## How to test
Create a spark node OR cluster. **With the HAIL image!** 
Take note of the "non-worker" memory allocation.

Screenshot 2023-05-04 at 3 16 26 PM
Once it's running, view output from `!cat /etc/spark/conf/spark-defaults.conf`
Verify spark.driver.memory is set to a number approx 80% of the total available memory.
Screenshot 2023-05-04 at 3 27 03 PM

Additionally, verify Jupyter docker image is allowed up to 20% of total available memory.

image

Note - you may need to find the resource in the GCC. dataproc machines are tagged with `-m` after the name seen in terra-ui.
gcloud compute ssh --zone "us-central1-a" "<runtimeName>" --project "<projectName>"

on the machine, run `docker stats`
Verify the jupyter-server image has a max limit set to ~20% available GB.
Screenshot 2023-05-04 at 3 25 10 PM

### Who tested and where

- [x] I validated this change 

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/runtimeModels.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/runtimeModels.scala
@@ -529,9 +529,13 @@ final case class MemorySize(bytes: Long) extends AnyVal {
   override def toString: String = bytes.toString + "b"
 }
 object MemorySize {
-  def fromKb(kb: Double): MemorySize = MemorySize((kb * 1024).toLong)
-  def fromMb(mb: Double): MemorySize = MemorySize((mb * 1048576).toLong)
-  def fromGb(gb: Double): MemorySize = MemorySize((gb * 1073741824).toLong)
+  val kbInBytes = 1024
+  val mbInBytes = 1048576
+  val gbInBytes = 1073741824
+
+  def fromKb(kb: Double): MemorySize = MemorySize((kb * kbInBytes).toLong)
+  def fromMb(mb: Double): MemorySize = MemorySize((mb * mbInBytes).toLong)
+  def fromGb(gb: Double): MemorySize = MemorySize((gb * gbInBytes).toLong)
 }
 
 /**
@@ -539,7 +543,7 @@ object MemorySize {
  * See https://docs.docker.com/compose/compose-file/compose-file-v2/#cpu-and-other-resources
  * for other types of resources we may want to add here.
  */
-final case class RuntimeResourceConstraints(memoryLimit: MemorySize)
+final case class RuntimeResourceConstraints(memoryLimit: MemorySize, totalMachineMemory: MemorySize)
 
 final case class RuntimeMetrics(cloudContext: CloudContext,
                                 runtimeName: RuntimeName,
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -71,7 +71,13 @@ dataproc {
 
   customDataprocImage = "projects/broad-dsp-gcr-public/global/images/leo-dataproc-image-2-0-51-debian10-2023-04-12-19-18-34"
 
-  dataprocReservedMemory = 6g
+  # The ratio of memory allocated to spark. 0.8 = 80%.
+  # Hail/Spark users generally allocate 80% of the ram to the JVM.
+  sparkMemoryConfigRatio = 0.8
+  # This allows at minimmum 4gb to be reserved away from Hail/Spark for Jupyter/system processes
+  # Jupyter has a suggested a minimum of 1.5gb, welder has a minimum of 0.5g, and we want some space for system processes.
+  # This limit is used to define the max memory available to Jupyter.
+  minimumRuntimeMemoryInGb = 4
 
   monitor {
     initialDelay = 30 seconds
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreter.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreter.scala
@@ -154,9 +154,9 @@ class DataprocInterpreter[F[_]: Parallel](
           .compile
           .drain
 
-        resourceConstraints <- getClusterResourceContraints(params.runtimeProjectAndName,
-                                                            machineConfig.masterMachineType,
-                                                            machineConfig.region
+        jupyterResourceConstraints <- getDataprocRuntimeResourceContraints(params.runtimeProjectAndName,
+                                                                           machineConfig.masterMachineType,
+                                                                           machineConfig.region
         )
 
         templateParams = RuntimeTemplateValuesConfig.fromCreateRuntimeParams(
@@ -169,7 +169,7 @@ class DataprocInterpreter[F[_]: Parallel](
           config.proxyConfig,
           config.clusterFilesConfig,
           config.clusterResourcesConfig,
-          Some(resourceConstraints),
+          Some(jupyterResourceConstraints),
           false
         )
         templateValues = RuntimeTemplateValues(templateParams, Some(ctx.now), false)
@@ -267,7 +267,11 @@ class DataprocInterpreter[F[_]: Parallel](
               .getOrElse((None, None))
           } else (None, None)
 
-        softwareConfig = getSoftwareConfig(googleProject, params.runtimeProjectAndName.runtimeName, machineConfig)
+        softwareConfig = getSoftwareConfig(googleProject,
+                                           params.runtimeProjectAndName.runtimeName,
+                                           machineConfig,
+                                           jupyterResourceConstraints
+        )
 
         // Enables Dataproc Component Gateway. Used for enabling cluster web UIs.
         // See https://cloud.google.com/dataproc/docs/concepts/accessing/dataproc-gateways
@@ -462,7 +466,7 @@ class DataprocInterpreter[F[_]: Parallel](
         LeoLenses.cloudContextToGoogleProject.get(params.runtimeAndRuntimeConfig.runtime.cloudContext),
         new RuntimeException("this should never happen. Dataproc runtime's cloud context should be a google project")
       )
-      resourceConstraints <- getClusterResourceContraints(
+      resourceConstraints <- getDataprocRuntimeResourceContraints(
         RuntimeProjectAndName(params.runtimeAndRuntimeConfig.runtime.cloudContext,
                               params.runtimeAndRuntimeConfig.runtime.runtimeName
         ),
@@ -741,9 +745,9 @@ class DataprocInterpreter[F[_]: Parallel](
       List(deleteBucket, deleteCluster, removeIamRoles).parSequence_
     }
 
-  private[leonardo] def getClusterResourceContraints(runtimeProjectAndName: RuntimeProjectAndName,
-                                                     machineType: MachineTypeName,
-                                                     region: RegionName
+  private[leonardo] def getDataprocRuntimeResourceContraints(runtimeProjectAndName: RuntimeProjectAndName,
+                                                             machineType: MachineTypeName,
+                                                             region: RegionName
   )(implicit
     ev: Ask[F, AppContext]
   ): F[RuntimeResourceConstraints] =
@@ -769,11 +773,19 @@ class DataprocInterpreter[F[_]: Parallel](
       _ <- logger.debug(ctx.loggingCtx)(s"Resolved machine type: ${resolvedMachineType.toString}")
       total = MemorySize.fromMb(resolvedMachineType.getMemoryMb.toDouble)
     } yield {
-      // total - dataproc allocated - welder allocated
-      val dataprocAllocated = config.dataprocConfig.dataprocReservedMemory.map(_.bytes).getOrElse(0L)
-      val welderAllocated = config.welderConfig.welderReservedMemory.map(_.bytes).getOrElse(0L)
-      val result = MemorySize(total.bytes - dataprocAllocated - welderAllocated)
-      RuntimeResourceConstraints(result)
+      // For dataproc, we don't need much memory for Jupyter.
+
+      // Hail reserves about 80% of memory available to the machine to the JVM.
+      val sparkMemoryConfigRatio = config.dataprocConfig.sparkMemoryConfigRatio.getOrElse(0.8)
+      // We still want a minimum to run Jupyter and other system processes.
+      val minRuntimeMemoryGb = config.dataprocConfig.minimumRuntimeMemoryInGb.getOrElse(4.0)
+      val runtimeAllocatedMemory =
+        Math.max(
+          (total.bytes * (1 - sparkMemoryConfigRatio)).toLong,
+          MemorySize.fromGb(minRuntimeMemoryGb).bytes
+        )
+
+      RuntimeResourceConstraints(MemorySize(runtimeAllocatedMemory), MemorySize(total.bytes))
     }
 
   /**
@@ -856,14 +868,21 @@ class DataprocInterpreter[F[_]: Parallel](
     }
   }
 
-  private def getSoftwareConfig(googleProject: GoogleProject,
-                                runtimeName: RuntimeName,
-                                machineConfig: RuntimeConfig.DataprocConfig
+  // This file is worth keeping track of. It's the source of truth for hail machine provisioning
+  // https://github.com/hail-is/hail/blob/75f351d43a6c6b2e87e7d37531be63aaf11af9df/hail/python/hailtop/hailctl/dataproc/start.py#L20
+  def getSoftwareConfig(googleProject: GoogleProject,
+                        runtimeName: RuntimeName,
+                        machineConfig: RuntimeConfig.DataprocConfig,
+                        jupyterResourceConstraints: RuntimeResourceConstraints
   ): SoftwareConfig = {
     val dataprocProps = if (machineConfig.numberOfWorkers == 0) {
       // Set a SoftwareConfig property that makes the cluster have only one node
       Map("dataproc:dataproc.allow.zero.workers" -> "true")
     } else Map.empty[String, String]
+
+    val memoryLimitInMb =
+      (jupyterResourceConstraints.totalMachineMemory.bytes - jupyterResourceConstraints.memoryLimit.bytes) / MemorySize.mbInBytes
+    val driverMemoryProp = Map("spark:spark.driver.memory" -> s"${memoryLimitInMb}m")
 
     val yarnProps = Map(
       // Helps with debugging
@@ -893,7 +912,7 @@ class DataprocInterpreter[F[_]: Parallel](
     SoftwareConfig
       .newBuilder()
       .putAllProperties(
-        (dataprocProps ++ yarnProps ++ stackdriverProps ++ requesterPaysProps ++ knoxProps ++ machineConfig.properties).asJava
+        (driverMemoryProp ++ dataprocProps ++ yarnProps ++ stackdriverProps ++ requesterPaysProps ++ knoxProps ++ machineConfig.properties).asJava
       )
       .addOptionalComponents(Component.DOCKER)
       .build()
--- a/http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreterSpec.scala
+++ b/http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreterSpec.scala
@@ -9,7 +9,7 @@ import com.google.api.client.googleapis.json.GoogleJsonResponseException
 import com.google.api.client.googleapis.testing.json.GoogleJsonResponseExceptionFactoryTesting
 import com.google.api.client.testing.json.MockJsonFactory
 import com.google.api.gax.longrunning.OperationFuture
-import com.google.cloud.compute.v1.{Instance, Operation}
+import com.google.cloud.compute.v1.{Instance, MachineType, Operation}
 import com.google.cloud.dataproc.v1.ClusterOperationMetadata
 import com.google.protobuf.Empty
 import kotlin.NotImplementedError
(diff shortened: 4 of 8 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreter.scala
```diff
@@ -863,7 +871,12 @@ class DataprocInterpreter[F[_]: Parallel](
     val dataprocProps = if (machineConfig.numberOfWorkers == 0) {
       // Set a SoftwareConfig property that makes the cluster have only one node
       Map("dataproc:dataproc.allow.zero.workers" -> "true")
-    } else Map.empty[String, String]
+    } else {
+      getDataProcMachineSpecificationFromMachineTypeName(machineConfig.workerMachineType) match {
```
n8landolt: For some reason I've been assuming this configuration is for the worker machines, but now after rereading the comments in the ticket, I think this is actually for the master machine?

So I may need to move this logic from this conditional and also pull from `machineConfig.masterMachineType`

n8landolt: Ahhhhh driver == "master"

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/runtimeModels.scala
```diff
@@ -585,3 +585,34 @@ object RuntimeOperation {
     val asString = "stopping"
   }
 }
+
+final case class DataProcMachineSpecification(machineTypeName: MachineTypeName, cpu: Int, memory: Double)
+
+object MachineTypes {
```
rtitle: Just want to point to some [existing code](https://github.com/DataBiosphere/leonardo/blob/c8f1292698db194169e69c0465670b6fde1355c1/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GceInterpreter.scala#L515) which dynamically resolves the machine type in GCE to get the CPU/memory… could be a way to avoid enumerating them all.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreter.scala
```diff
@@ -864,6 +869,16 @@ class DataprocInterpreter[F[_]: Parallel](
       // Set a SoftwareConfig property that makes the cluster have only one node
       Map("dataproc:dataproc.allow.zero.workers" -> "true")
     } else Map.empty[String, String]
+    // Reserve 3.5g to 4g (math.floor) memory for Welder, Jupyter, and Server processes. This value is arbitrary.
```
rtitle: Interesting.. so before was the driver allocating all the memory on the master VM, causing contention with Jupyter/welder containers?

Makes sense to reserve some memory for Juypter/welder/other processes. Wondering how you came up with 3.5G? (It seems reasonable, just curious)

rtitle:
Just dropped some feedback but looks good.. you could probably verify this using the Spark web UIs (which are linked from Terra UI).

jdcanas:
Hmm.. how is this covered by automation tests if users were encountering this in the wild? It thought it was affecting all users not just a few sporadically. Were you able to reproduce the original issue and see if this resolves it?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreter.scala
n8landolt: It seemed like this property wasn't being set at all? So I'm not entirely sure what the original value was.
I think the main issue was the driver wasn't allocating ENOUGH memory, it seemed like there's a `6g` constant in reference.conf.

n8landolt: I calculated it by finding elsewhere in the code that Welder needed 0.5g, and saw online Jupyter also needed 0.5g MINIMUM. Then I figured 2 ish for system processes, so I rounded up slightly so in most cases it'd really be 4g left free (there's a case at 7.5 which will resolve to 4g for the JVM, which seemed like a good minimum.)

rtitle: OK - I seem to remember if `spark:spark.driver.memory` was not specified it would default to all the memory on the JVM. But not 100% sure. You should be able to check the `-Xmx` of the driver Java process and compare before/after.

n8landolt: It seems like the spark.driver.memory will set the JVM memory to the same amount!

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/runtimeModels.scala
n8landolt: Seems like this is a reasonable approach! Seems to allocate more memory for the system / welder than I was planning to. 
I also switched to megs over gigs for better precision.

n8landolt: Hey @rtitle , I'm a bit confused about the output of the getClusterResourceConstraints. 
I'm not sure how the total - dataprocallocated - welder allocated is being used.

Essentially what I want to do is set `spark:spark.driver.memory` to the most amount of memory possible for the machine. And from what I understand, this is what "dataproc" is.

So with getClusterResourceConstraints, it seems like this 6gb is being subtracted from available memory to reserve for dataproc... but when I go to use this function to get data for dataproc, I'm actually reserving memory for dataproc (The spark JVM).
For lower memory machines, this 6gb subtraction is killer. For higher memory machines, it's less significant, and results from my previous implementation will be closer to this implementation.

What am I missing?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreter.scala
danking: The default value for `spark.driver.memory`, in Spark at least, [is `1g`](https://spark.apache.org/docs/latest/configuration.html). I do not know if Dataproc sets this intelligently. When users start clusters with `hailctl` we explicitly set the driver and executor memory (as well as memory overhead and a Hail-specific environment variable).

Regarding memory allocated to non-Spark driver-node processes: we default to 80% of the machine's RAM allocated to Spark. The rest is left open for Jupyter, etc.

danking:
To reproduce this issue, you need to create a large memory allocation on the driver node. I don't know exactly what users were executing, so it's a bit hard to reproduce exactly.

If you just want to verify that you can make a large allocation in driver RAM, this should suffice:

```
import hail as hl
# You will need workers with at least 4 * 1024 * 4 * 1024 / n_partitions bytes of RAM
ht = hl.utils.range_table(4 * 1024, n_partitions=128)
ht = ht.annotate(x = hl.range(1024))
large_array_in_spark_driver = ht.aggregate(hl.agg.collect(x), _localize=False)
hl.sum(large_array_in_spark_driver.map(hl.sum)).collect()
```
That should print `[536346624]`. It uses some special Hail functionality (`_localize=False`) which explicitly places things in memory on the driver rather than in Jupyter. It should place ~4 bytes * 1024 * 4 * 1024 = 16GiB in RAM.

http/src/main/resources/reference.conf
```diff
@@ -71,7 +71,8 @@ dataproc {
 
   customDataprocImage = "projects/broad-dsp-gcr-public/global/images/leo-dataproc-image-2-0-51-debian10-2023-04-12-19-18-34"
 
-  dataprocReservedMemory = 6g
+  jupyterReservedMemory = 1.5g
```
LizBaldo: What is the reasoning behind this memory allocation? Leaving a comment might be nice here

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreter.scala
```diff
@@ -154,11 +154,9 @@ class DataprocInterpreter[F[_]: Parallel](
           .compile
           .drain
 
-        resourceConstraints <- getClusterResourceContraints(params.runtimeProjectAndName,
-                                                            machineConfig.masterMachineType,
-                                                            machineConfig.region
+        jupyterResourceConstraints = RuntimeResourceConstraints(
+          MemorySize(config.dataprocConfig.jupyterReservedMemory.map(_.bytes).getOrElse(MemorySize.fromGb(1.5).bytes))
```
LizBaldo: Shouldn't this 1.5 be read from Config instead of being hardcoded?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreter.scala
```diff
@@ -769,10 +776,13 @@ class DataprocInterpreter[F[_]: Parallel](
       _ <- logger.debug(ctx.loggingCtx)(s"Resolved machine type: ${resolvedMachineType.toString}")
       total = MemorySize.fromMb(resolvedMachineType.getMemoryMb.toDouble)
     } yield {
-      // total - dataproc allocated - welder allocated
-      val dataprocAllocated = config.dataprocConfig.dataprocReservedMemory.map(_.bytes).getOrElse(0L)
-      val welderAllocated = config.welderConfig.welderReservedMemory.map(_.bytes).getOrElse(0L)
-      val result = MemorySize(total.bytes - dataprocAllocated - welderAllocated)
+      // For dataproc, we don't need much memory for Jupyter.
+      val jupyterAllocated =
+        config.dataprocConfig.jupyterReservedMemory.map(_.bytes).getOrElse(MemorySize.fromGb(1.5).bytes)
```
LizBaldo: Same comment here and below about not hardcoding the memory allocated

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreter.scala
```diff
@@ -769,10 +776,13 @@ class DataprocInterpreter[F[_]: Parallel](
...
+      val welderAllocated = config.welderConfig.welderReservedMemory.map(_.bytes).getOrElse(MemorySize.fromGb(.5).bytes)
+      val otherReservedMemory =
+        config.dataprocConfig.systemReservedMemory.map(_.bytes).getOrElse(MemorySize.fromGb(2).bytes)
```
LizBaldo: Should this be in the reference.conf config as well?

LizBaldo (commented):
LGTM! Worth testing now :) I only left a few comments about not hardcoding the memories allocated

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreter.scala
n8landolt: It's trying to read from config, Scala reads it in as optional, but if it can't find it in config it will default to 1.5.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreter.scala
n8landolt: Same reply, their correct values will be in reference. The "OrElse" part should never happen.

http/src/main/resources/reference.conf
n8landolt: Minimum suggested for Jupyter is 1.5g. I'll add a comment!

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreter.scala
LizBaldo: Oh boy, I need to get back to drinking coffee after lunch so sorry 🤦‍♀️

http/src/main/resources/reference.conf
```diff
@@ -71,7 +71,9 @@ dataproc {
 
   customDataprocImage = "projects/broad-dsp-gcr-public/global/images/leo-dataproc-image-2-0-51-debian10-2023-04-12-19-18-34"
 
-  dataprocReservedMemory = 6g
+  # We may be able to increase this and have overlap with memory allocated to dataproc.
```
n8landolt: I didn't want to put TOO much in the comment (I can if we want)

Basically we might be able to have the Jupyter Docker image allotted a maximum greater than 2g, and just have dataproc memory be 80% of machine memory. If there's some overlap, that's fine. For example:
40GB machine. Dataproc Memory: 32g. Jupyter Docker Max memory: 10g. (32+10 has an overlap of 2gs)

The reason I bring this up is I saw a comment about sometimes wanting to allocate more memory to python. So we MAY be restricting them here.
It also seems like we don't restrict the resources of the crypto mining detector or welder. If I remember correctly, they had had a cap of around 11g on a 15 gb machine. So, the memory space isn't RESERVED for the running docker images, it's just a max limit it can pull from what I can tell from SSHing and using docker ps.

LizBaldo: I am all for adding that context to the comments, the more details the better 👍

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreter.scala
```diff
@@ -769,10 +776,13 @@ class DataprocInterpreter[F[_]: Parallel](
...
-      val dataprocAllocated = config.dataprocConfig.dataprocReservedMemory.map(_.bytes).getOrElse(0L)
-      val welderAllocated = config.welderConfig.welderReservedMemory.map(_.bytes).getOrElse(0L)
-      val result = MemorySize(total.bytes - dataprocAllocated - welderAllocated)
+      // For dataproc, we don't need much memory for Jupyter.
+      val jupyterAllocated =
+        config.dataprocConfig.jupyterReservedMemory.map(_.bytes).getOrElse(MemorySize.fromGb(2).bytes)
+      val welderAllocated = config.welderConfig.welderReservedMemory.map(_.bytes).getOrElse(MemorySize.fromGb(.5).bytes)
+      val result = MemorySize(
+        ((total.bytes - jupyterAllocated - welderAllocated) * 0.8).toLong
+      ) // Hail reserves about 80% of memory available to the machine to the JVM.
```
LizBaldo: Is there a link to a documentation about hail using 80%?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreter.scala
```diff
@@ -856,15 +866,22 @@ class DataprocInterpreter[F[_]: Parallel](
     }
   }
 
-  private def getSoftwareConfig(googleProject: GoogleProject,
-                                runtimeName: RuntimeName,
-                                machineConfig: RuntimeConfig.DataprocConfig
+  // This file is worth keeping track of. It's the source of truth for hail machine provisioning
```
LizBaldo, line 871: 👍

http/src/main/resources/reference.conf
n8landolt: Update,
The 80-20% split is for Spark - Jupyter and system processes.
Going to refactor.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreter.scala
```diff
@@ -266,8 +266,17 @@ class DataprocInterpreter[F[_]: Parallel](
               }
               .getOrElse((None, None))
           } else (None, None)
-
-        softwareConfig = getSoftwareConfig(googleProject, params.runtimeProjectAndName.runtimeName, machineConfig)
+//
+//        dataprocResourceConstraints <- getDataprocResourceConstraints(params.runtimeProjectAndName,
```
LizBaldo: Should this be removed instead of commented out?

http/src/main/resources/reference.conf
```diff
@@ -71,7 +71,9 @@ dataproc {
 
   customDataprocImage = "projects/broad-dsp-gcr-public/global/images/leo-dataproc-image-2-0-51-debian10-2023-04-12-19-18-34"
 
-  dataprocReservedMemory = 6g
+  # The ratio of memory allocated to spark. 0.8 = 80%.
```
LizBaldo, line 74: Is this documented anywhere? Would be helpful for the next dev working on it

http/src/main/resources/reference.conf
```diff
@@ -71,7 +71,9 @@ dataproc {
...
+  sparkMemoryConfigRatio = 0.8
+  minimumRuntimeMemoryInGb = 4
```
LizBaldo, line 80: So the original 1.5G was too low?

LizBaldo: Would be great to leave a comment explaining the reasoning behind this value

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreter.scala
```diff
@@ -769,11 +778,16 @@ class DataprocInterpreter[F[_]: Parallel](
...
-      val dataprocAllocated = config.dataprocConfig.dataprocReservedMemory.map(_.bytes).getOrElse(0L)
-      val welderAllocated = config.welderConfig.welderReservedMemory.map(_.bytes).getOrElse(0L)
-      val result = MemorySize(total.bytes - dataprocAllocated - welderAllocated)
-      RuntimeResourceConstraints(result)
+      // For dataproc, we don't need much memory for Jupyter.
+
+      // Hail reserves about 80% of memory available to the machine to the JVM.
+      val sparkMemoryConfigRatio = config.dataprocConfig.sparkMemoryConfigRatio.getOrElse(0.8)
+      // We still want a minimum to run Jupyter and other system processes.
+      val minRuntimeMemoryGb = 4
```
LizBaldo: This should also be read from the configs no?

LizBaldo (commented):
Testing it now, there are a couple of little things that might still need to be addressed in the meantime

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreter.scala
n8landolt: Screenshot 2023-05-04 at 4 01 09 PM

Additionally chatted briefly with hail users in their version of slack.

http/src/main/resources/reference.conf
n8landolt: What specifically? The reasoning?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreter.scala
n8landolt: Oh yes! I forgot to wrap that part up.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreter.scala
LizBaldo: Gotcha, I was hoping that this would be documented in their doc, but no worries if not :)

http/src/main/resources/reference.conf
LizBaldo: All good, if it is not an official hail default documented somewhere this will do

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreterSpec.scala
```diff
@@ -188,7 +188,31 @@ class DataprocInterpreterSpec
     )
   }
 
-  it should "calculate cluster resource constraints" in isolatedDbTest {
+  it should "calculate cluster resource constraints fpr" in isolatedDbTest {
```
jdcanas: `fpr` probably a leftover test string?

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreterSpec.scala
```diff
@@ -201,15 +225,60 @@ class DataprocInterpreterSpec
...
+    val dataProcSoftwareConfig = dataprocInterp().getSoftwareConfig(
+      GoogleProject("MyGoogleProject"),
+      RuntimeName("MyRuntimeName"),
+      runtimeConfig,
+      resourceConstraints
+    )
+
+    val propertyMap = dataProcSoftwareConfig.getPropertiesMap()
+    resourceConstraints.totalMachineMemory shouldBe MemorySize.fromGb(104)
+    resourceConstraints.memoryLimit shouldBe MemorySize((MemorySize.fromGb(104).bytes * (1 - 0.8)).toLong)
```
jdcanas: can the .8 here be sourced from the same constraints the interp/`getSoftwareConfig` is given?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreter.scala
```diff
@@ -856,15 +865,22 @@ class DataprocInterpreter[F[_]: Parallel](
...
+                        machineConfig: RuntimeConfig.DataprocConfig,
+                        jupyterResourceConstraints: RuntimeResourceConstraints
   ): SoftwareConfig = {
     val dataprocProps = if (machineConfig.numberOfWorkers == 0) {
       // Set a SoftwareConfig property that makes the cluster have only one node
       Map("dataproc:dataproc.allow.zero.workers" -> "true")
     } else Map.empty[String, String]
 
+    val memoryLimitInMb =
+      (jupyterResourceConstraints.totalMachineMemory.bytes - jupyterResourceConstraints.memoryLimit.bytes) / 1048576
```
jdcanas: what is 1048576? Maybe a comment? Should it be a field in `RuntimeResourceConstraints` sourced from a config instead of a floating magic number?

jdcanas (approved):
few small nits, looks good otherewise

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreter.scala
```diff
@@ -769,11 +773,16 @@ class DataprocInterpreter[F[_]: Parallel](
...
-      val result = MemorySize(total.bytes - dataprocAllocated - welderAllocated)
-      RuntimeResourceConstraints(result)
+      // For dataproc, we don't need much memory for Jupyter.
+
+      // Hail reserves about 80% of memory available to the machine to the JVM.
+      val sparkMemoryConfigRatio = config.dataprocConfig.sparkMemoryConfigRatio.getOrElse(0.8)
+      // We still want a minimum to run Jupyter and other system processes.
+      val minRuntimeMemoryGb = config.dataprocConfig.minimumRuntimeMemoryInGb.getOrElse(4)
+      val runtimeAllocatedMemory =
+        Math.max((total.bytes * (1 - sparkMemoryConfigRatio)).toLong, MemorySize.fromGb(minRuntimeMemoryGb).bytes)
```
jdcanas: looks like there's a type mismatch here

n8landolt: I'm not seeing that issue here

LizBaldo (approved):
I was able to see the 80% spark and 20% jupyter memory allocation on my BEE as well, thanks for sharing the detailed testing steps :)

👍 on my end, feel free to merge after addressing Justin's comments!

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreterSpec.scala
n8landolt: Yep! Fixed and added more description

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreter.scala
n8landolt: Added into MemorySize. LMK if you have a better suggestion for the name of the variables

n8landolt:
jenkins retest

PR #3313, opened 2023-05-01
[IA-4189] Make sure resource creator still has workspace access

LizBaldo:
__Jira ticket__: https://broadworkbench.atlassian.net/browse/IA-4189

## Summary of changes

This only impacts V2 endpoints as this is related to workspace access. The following endpoints have been changed with an additional workspace permission check:
- Runtime: list, get, start, stop, and delete
- App: list, get, and delete
- Disk: get and delete

### What

- Add a check to ensure that resource creator also has access to the parent workspace

### Why

- This is a security issue, if a resource creator has been removed from the parent workspace, then this user should not have access to the workspace child resources regardless of the creator role

### What to test
You can do this on your BEE, with user Ab being your broad email address, and user B being your throwaway test gmail address for instance

- User A creates a workspace and a (v2 runtime / v2 app / v2 disk)
- User A shares the workspace with user B
- User B creates a (v2 runtime / v2 app / v2 disk) in that workspace
- At this points, user B has access to both user A and user B (v2 runtime / v2 app / v2 disk)
- User A removes user B from the workspace
- At this point, user B no longer has access to any (v2 runtime / v2 app / v2 disk) via any endpoint

### Who tested and where

- [ ] I validated this change 


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/SamAuthProvider.scala
```diff
@@ -194,6 +194,32 @@ class SamAuthProvider[F[_]: OpenTelemetryMetrics](
     } yield roles.contains(SamRole.Owner)
   }
 
+  def filterWorkspaceReader(
```
LizBaldo, line 197: My thinking here is that as long as there is a policy set for that user, they must be at least a reader

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/SamAuthProvider.scala
```diff
@@ -194,6 +194,32 @@ class SamAuthProvider[F[_]: OpenTelemetryMetrics](
...
+    } yield readingWorkspaces.toSet
+  }
+
+  def isUserWorkspaceReader(
+    workspaceResource: WorkspaceResourceSamResourceId,
+    userInfo: UserInfo
+  )(implicit ev: Ask[F, TraceId]): F[Boolean] = {
+    val authHeader = Authorization(Credentials.Token(AuthScheme.Bearer, userInfo.accessToken.token))
+    for {
+      roles <- samDao.getResourceRoles(authHeader, workspaceResource)
```
LizBaldo, line 217: Same thinking here, if there is a role associated with the user, it is at least reader

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala
```diff
@@ -554,12 +594,13 @@ class RuntimeV2ServiceInterp[F[_]: Parallel](config: RuntimeServiceConfig,
               case None => List.empty
             }
           } yield
-          // runtimesUserIsCreator: runtimes user is creator
+          // runtimesUserIsCreatorWithoutWorkspaceId: runtime user is creator, legacy V1 without workspace
```
LizBaldo: The logic above feels a bit dicey to me, would appreciate some extra scrutiny :)

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeServiceV2InterpSpec.scala
```diff
@@ -530,8 +541,48 @@ class RuntimeServiceV2InterpSpec extends AnyFlatSpec with LeonardoTestSuite with
...
+          defaultCreateAzureRuntimeReq
+        )
+      azureCloudContext <- wsmDao.getWorkspace(workspaceId, dummyAuth).map(_.get.azureContext)
+      clusterOpt <- clusterQuery
+        .getActiveClusterByNameMinimal(CloudContext.Azure(azureCloudContext.get), runtimeName)(
+          scala.concurrent.ExecutionContext.global
+        )
+        .transaction
+      cluster = clusterOpt.get
+      getResponse <- azureService2.getRuntime(userInfo, runtimeName, workspaceId).attempt
```
LizBaldo: I am trying to get the runtime using a different authprovider that does not contain the creator's email address anymore, but it does not seem to be picked up, I might need so pairing again 🤔

LizBaldo:
@lucymcnatt and @jdcanas this is ready for review :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/SamAuthProvider.scala
lucymcnatt: have you verified this? ie that its empty when a user has no access to a resource?

LizBaldo: That is a very good point, I was surprised to see that the unit tests are basically swapping in a completely different logic for that function (aka user being contained in the allowlist). 
This is based on my manual testing (ongoing, setting things up on my BEE now), but if you have an idea to properly test this as a unit test lmk!

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
```diff
@@ -436,7 +436,14 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
       .transaction
 
     res <- filterAppsBySamPermission(allClusters, userInfo, paramMap._3)
-  } yield res
+    // Make sure that the user still has access to the resource parent workspace
+    resWithWorkspace <- res.traverseFilter { resource =>
```
lucymcnatt: would it make sense to check the users workspace permissions at the beginning (possibly even before querying) because the result of this `authProvider.isUserWorkspaceReader` would be the same for all apps? then just return nothing if the user doesn't have workspace permissions

LizBaldo: Oh yes, thanks for the catch! In my mind the list endpoint could eventually work across workspaces which is why I put it after, but no need to do that eight now while the endpoint still depends on having a workspace id (the fact that runtimes don't was confusing to me at first)

lucymcnatt: oh gotcha, if you'd rather go with the more future-proof option I'm fine with that! with maybe a comment on why

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
```diff
@@ -677,18 +689,22 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
     ctx <- as.ask
     listOfPermissions <- authProvider.getActions(app.samResourceId, userInfo)
 
+    hasWorkspacePermission <- authProvider.isUserWorkspaceReader(WorkspaceResourceSamResourceId(workspaceId), userInfo)
```
lucymcnatt: kind of a general comment, but would it make sense to fail faster when the user doesn't have workspace permission, have that be the first check, error if user doesn't so that we aren't doing the sam checks/other logic unless we have to?

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/auth/AllowlistAuthProvider.scala
```diff
@@ -18,7 +18,7 @@ import org.broadinstitute.dsde.workbench.model.{TraceId, UserInfo, WorkbenchEmai
 
 class AllowlistAuthProvider(config: Config, saProvider: ServiceAccountProvider[IO]) extends LeoAuthProvider[IO] {
 
-  val allowlist = config.as[Set[String]]("whitelist").map(_.toLowerCase)
+  val allowlist = config.as[Set[String]]("allowlist").map(_.toLowerCase)
```
lucymcnatt, line 21: :thumbsup:

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
LizBaldo: That is a very good point, let me do another pass at this and switch things around a bit (hopefully that won't change the unit tests, which were the most annoying things to set up here)

README.md
```diff
@@ -23,9 +23,10 @@ We use JIRA instead of the issues page on Github. If you would like to see what
 ## Java client library
 for sbt:
 
-```libraryDependencies += "org.broadinstitute.dsde.workbench" %% "leonardo-client" % "0.1-<git hash>"```
+```libraryDependencies += "org.broadinstitute.dsde.workbench" %% "leonardo-client" % "1.3.6-<git hash>"```
```
jdcanas, line 26: what exactly is the logic behind this versioning?

LizBaldo: Not related to this PR per say, just some general clean up of our java client library doc. Not sure what the actual logic behind versioning is really, which is why this is an example and the user should look at what is available on Jfrog (link provided below)

jdcanas (commented):
Hmm.. the general approach seems sound but I can't help think this is something Sam should be doing. Yes, its something Leo handles improperly, but I would think that if I ask sam 'does user B have permissions to do X to Y in Workspace Z', if user B is not in Workspace Z it should say no... Right now, we are kind of doing what the UI does with `listRuntimes`, and figuring out if the user has a `currentRuntime` through some side effects. Unless there is an explicit Sam endpoint that asks `is user B in Workspace Z`, it seems a bit risky to try figure out the answer to that question through side effects like whether the user has Read permissions on some resources (which, I don't think this even works, it sounds like the Read/Write permissions aren't cleaned up when the user is removed since we ask Sam on every request to begin with)

LizBaldo:
Very fair point, let me assess what Sam does and the level of effort to modify anything there, I might be duplicating things here

LizBaldo:
This is where using the SAM library would help a lot :)

jdcanas:
I'd almost say this should be part of the existing endpoints to ask if a User has permission to perform an Action on a Resource... Sam should say no if the user has been removed from the workspace (which it clearly doesn't). Might be worth asking the Identiteam...

rtitle:
I started going through this - will leave more detailed comments in a bit. The general approach seems 👍 though.

Re @jdcanas' comment -- yeah it would be nice if our Sam resources encapsulated all access control. Our existing resources (`runtime`, `persistent-disk`, `kubernetes-app`) are kind of independent from workspaces -- which was actually by design a long time ago, when these resources could actually span workspaces. We should think about adapting things though. Potentially we could model more things as WSM [controlled|referenced] resources and use that permission model (which has the right workspace semantics). Drew and I have talked about this a bit.

Food for thought.. but short term I think this PR is the right way to go (and I'll look in more detail shortly!)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/SamAuthProvider.scala
```diff
@@ -194,6 +194,30 @@ class SamAuthProvider[F[_]: OpenTelemetryMetrics](
     } yield roles.contains(SamRole.Owner)
   }
 
+  def filterWorkspaceReader(
+    resources: NonEmptyList[WorkspaceResourceSamResourceId],
```
rtitle, line 198: It seems like `resources` is not used in this method.. it doesn't seem to be filtering anything, just returning all the workspace resource IDs from Sam. I'll look further down to try and understand better how this method is used..

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/SamAuthProvider.scala
rtitle: This makes sense I think. Reader is the minimum permission for workspace -- so if you're at least a reader, you can see all your Leo resources in the workspace.

rtitle: You could also explicitly check the read permission by `samDao.hasResourcePermission(workspaceResource, SamAction.Read, authHeader)`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
```diff
@@ -429,14 +429,22 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
...
-      .listFullAppsByWorkspaceId(Some(workspaceId), paramMap._1, paramMap._2)
-      .transaction
+  ): F[Vector[ListAppResponse]] =
+    for {
+      // Make sure that the user still has access to the resource parent workspace
+      hasWorkspacePermission <- authProvider.isUserWorkspaceReader(
+        WorkspaceResourceSamResourceId(workspaceId),
+        userInfo
+      )
+      _ <- F.raiseUnless(hasWorkspacePermission)(ForbiddenError(userInfo.userEmail))
```
rtitle, line 439: I was considering whether 403 or 404 makes more sense for no-workspace-permission (e.g. we may want to pretend the workspace doesn't exist). But I am not sure, 403 seems fine I think.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskV2ServiceInterp.scala
```diff
@@ -40,23 +43,28 @@ class DiskV2ServiceInterp[F[_]: Parallel](config: PersistentDiskConfig,
         .getGetPersistentDiskResponseV2(diskId, ctx.traceId)
         .transaction
 
-      // If user is creator of the disk, they should definitely be able to see the disk.
-      hasPermission <-
-        if (diskResp.auditInfo.creator == userInfo.userEmail) F.pure(true)
```
rtitle, line 45: Seems we are losing the `if user is creator return true` logic. Which is fine with me.. not sure why it was needed. Just want to make sure there is no negative impact to this?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala
```diff
@@ -479,53 +504,51 @@ class RuntimeV2ServiceInterp[F[_]: Parallel](config: RuntimeServiceConfig,
         if (creatorOnly.isDefined) {
```
rtitle, line 504: Oh dear what is the logic in this method… 🤯

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskV2ServiceInterp.scala
LizBaldo: This is the source of AppSec worries, they do not want any user not having access to the parent workspace to have access to a child resource regardless of a creator status

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/SamAuthProvider.scala
LizBaldo: Oh yeah, I need to clean this up, probably too hastily copied it from the `filterWorkspaceOwner` function 😬

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/SamAuthProvider.scala
LizBaldo: That sounds like a cleaner way that does not rely on a side effect, I will implement that, thanks!

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala
```diff
@@ -479,53 +504,51 @@ class RuntimeV2ServiceInterp[F[_]: Parallel](config: RuntimeServiceConfig,
...
               )
             )
           for {
-            // If user is not creator for the runtime, then we check if they're workspace owner
-            runtimesUserIsNotCreatorWithWorkspaceIdAndSamResourceId <- runtimesUserIsNotCreatorWithWorkspaceSamIds
+            // We check if the user has access to the workspace
+            runtimesUserWithWorkspaceIdAndSamResourceId <- runtimesUserWithWorkspaceSamIds
               .traverse(workspaces =>
-                authProvider.filterWorkspaceOwner(
+                authProvider.filterWorkspaceReader(
```
rtitle, line 532: OK I think I see -- `filterWorkspaceReader` is just returning the resources from Sam (it's not doing any filtering). The filtering is occurring in the next call to `mapFilter`.

Seems like it was this way before this PR, but this code feels quite confusing to me..

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala
LizBaldo: From a security perspective, 404 would make more sense I think, but we are going to have users lose access to their resources, and getting a 403 will let them know exactly why

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala
LizBaldo: Yeah, my brain hurt a bit when I saw that 😱 My change should simplify things a lot though since we do not have to keep track of the creator status anymore

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala
```diff
@@ -553,13 +576,11 @@ class RuntimeV2ServiceInterp[F[_]: Parallel](config: RuntimeServiceConfig,
...
-          //                                                      Note here we don't need to check workspace level permission because for WSM created resources, permissions
-          //                                                      built hierarchically. So by asking Sam if user can read a runtime will implicitly check user's permission at workspace level
-          // samVisibleRuntimesWithoutWorkspaceId: runtimes user is not creator, but user can view the runtime according to Sam
-          runtimesUserIsCreator ++ samUserVisibleRuntimesUserIsNotCreatorWithWorkspace ++ samVisibleRuntimesWithoutWorkspaceId
+          // samVisibleRuntimesUserWithWorkspaceId: User has access to both the runtime and the parent workspace (need to check in case the user lost workspace access after runtime creation for instance)
+          // samVisibleRuntimesWithoutWorkspaceId: Legacy runtimes created on GCP that have o associated workspace ID and we fallback on google project-level permissions
+          samVisibleRuntimesUserWithWorkspaceId ++ samVisibleRuntimesWithoutWorkspaceId
```
rtitle, line 583: Again losing the `runtimesUserIsCreator` check.. which is fine, just want to confirm no negative effects.

Maybe one reason this existed was to avoid delays in Sam resource creation. One thing to test: if you create a runtime in the UI, does it still immediately display in the UI (is it immediately list-able after create)?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/SamAuthProvider.scala
rtitle: Yeah I see `filterWorkspaceOwner` behaves the same way.. and I think the filtering _is_ occurring, just outside of this method (see below comment).

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskV2ServiceInterp.scala
rtitle: Ah right, that makes sense.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala
LizBaldo: I believe that the UI has separate logic to list runtimes, but that will be easy to check on my BEE 👍

rtitle (approved):
👍 , seems like this hit all the Leo endpoints and should do what we want. 

There is some tech debt/refactoring to be done of the Leo/Sam integration, but outside the scope of this PR I think.

One thing -- this covers the Leo `/api` endpoints I think, but there is also the Leo proxy/listener Sam check of the `connect` action to a runtime/app. Should that check workspace permission too? Otherwise a user would still be able to connect to their app after losing workspace permission (though it would be harder because they can no longer list it).

LizBaldo:
Thanks! Let me address your comments and look into the proxy logic :)

LizBaldo:
Bleh, it actually looks like the workspace is never passed in the proxy app or runtime requests since they are all part of the v1 endpoint, I can set it up so that these functions take a workspace id as an option and check the permission in there, but this is getting into the V1 to V2 migration territory and might be outside of the scope of this PR

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala
LizBaldo: Agreed, this function (and the list runtime one) is trying to do too much and could benefit from a larger refactoring

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/SamAuthProvider.scala
LizBaldo: I might actually stick to checking the user role, there is no clear mapping to  a read sam action that I can see in their doc (unless read_policies maybe?)

lucymcnatt (approved):

LizBaldo:
Just tested on my BEE and things are working as expected 👍 
I do not think that there is any harm in merging to dev today since the next monolith release won't happen until next Wednesday

mlilynolting (approved):

PR #3338, opened 2023-05-10
[IA-4268] RStudio on GKE as a new App Type

LizBaldo:

This PR implements the installation of the new RStudio helm chart developed by AOU: https://github.com/DataBiosphere/terra-app/pull/34

Sister PR in terra helmfile repo to disable this App in prod until we successfully go through the production readiness checklist https://github.com/broadinstitute/terra-helmfile/pull/4195

### What

- Creates a new RStudio app type that supports welder file syncing and detachable PDs
- Add this app type to the app metrics we monitor
- When creating the app in the GKE interpreter, the specificity of RStudio is the need to create a staging bucket for Welder to use (to dump logs into)

### Why

- AOU needs RStudio to run as an app in Q2 2023, which means that we unfortunately do not have the time to enable new app types via a custom helm chart URL option, but have to create a new app type from scratch (similar to the work done for Hail and WDS recently)

### What to test

- Update your BEE with the latest image generated by this PR
- Make sure that you have a google project ready to use
- From your BEE leo swagger page, create an RStudio app using the apps/createApp v1 endpoint
```
{
  "diskConfig": {
    "name": "disk1"
  },
  "appType": "RSTUDIO",
    "WORKSPACE_NAME": "<YOUR_WORKSPACE_NAME>"
  }
}
```
- Monitor your leonardo BEE log, as well as the k8s workload to ensure that the app is being installed properly and both the rstudio and welder pods are running
- To set up Welder you need to:
- - Create a storage bucket in your google project (e.g. `test-welder`) and make sure to give it the correct permissions (easier to copy the permissions  from the staging bucket that was created with the app in that google project)
- - Send a welder post request to create a storage link to that test bucket, e.g.
```
curl 'https://leonardo.<YOUR-BEE-NAME>.bee.envs-terra.bio/proxy/google/v1/apps/<YOUR-GOOGLE-PROJECT>/<YOUR-APP-NAME>/welder-service/storageLinks' \
  -H 'authority: leonardo.<YOUR-BEE-NAME>.bee.envs-terra.bio' \
  -H 'accept: */*' \
  -H 'accept-language: en-US,en;q=0.9' \
  -H 'authorization: Bearer <YOUR-ACCESS-TOKEN>' \
  -H 'content-type: application/json' \
  -H 'origin: https://terra.<YOUR-BEE-NAME>.bee.envs-terra.bio' \
  -H 'referer: https://leonardo.dsde-dev.broadinstitute.org' \
  --data-raw '{"localBaseDirectory":"","localSafeModeBaseDirectory":"","cloudStorageDirectory":"gs://<YOUR-BUCKET-NAME>/notebooks","pattern":".+(\\.R|\\.Rmd)$"}' \
  --compressed
```
- Generate a cookie in your browser from the leo BEE swagger page (proxy/setCookie endpoint)
- You might need to use something like the ModHeader pluggin to add `https://leonardo.dsde-dev.broadinstitute.org` as the Referer
- Open the RStudio app (the url should be `https://leonardo.e<YOUR-BEE-NAME>.bee.envs-terra.bio/proxy/google/v1/apps/<YOUR-GOOGLE-PROJECT>/<YOUR-APP-NAME>/rstudio-service/`). Note that the first time it might redirect you to get authenticated, I have not quite figured this one yet. So what I ended up doing is to first go to the `/rstudio-service/auth-sign-in` endpoint that redirects me to the swagger page and then I can access the url above.
- You should be able to create R and Rmd files, and you can then check that they are being synced by welder to your test storage bucket
- Also load a non R file, like a picture of a bunny 🐰 so we can test if the persistent disk reattachment works in the steps below
- Delete the app but keep the disk
- Create another RStudio app using the same disk
- Open the app again and make sure the data from the old app is still in there

### Who tested and where

- [ ] I validated this change 


http/src/main/resources/reference.conf
```diff
@@ -596,6 +596,22 @@ gke {
     dbPassword = "replace-me"
     enabled = true
   }
+  rstudioApp{
```
jdcanas: space

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -504,6 +518,7 @@ class GKEInterpreter[F[_]](
                 }
             } yield ()
         case AppType.Cromwell => persistentDiskQuery.updateLastUsedBy(diskId, app.id).transaction
+        case AppType.RStudio  => F.unit // TODO grab the pv and pvc
```
jdcanas: Is there a follow-up ticket for this?

LizBaldo: Good catch, I do not believe that we need to support PD for RStudio here, but if we do it will be part of this PR so no need for a follow up ticket

Dockerfile
```diff
@@ -44,21 +45,23 @@ RUN curl -fsSL -o get_helm.sh https://raw.githubusercontent.com/helm/helm/master
...
     helm repo add terra-helm https://terra-helm.storage.googleapis.com && \
+    helm repo add aou-rstudio-chart https://storage.googleapis.com/terra-app-helm/aou-rstudio-chart && \
     helm repo update
 
 
 # .Files helm helper can't access files outside a chart. Hence in order to populate cert file properly, we're
 # pulling `terra-app-setup` locally and add cert files to the chart.
 # Leonardo will install the chart from local version.
 # We are also caching charts so they are not downloaded with every helm-install
+# TODO pull the RSTudio chart version and untar
```
rtitle: maybe remove TODO :)

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
```diff
@@ -340,9 +344,9 @@ object AppType {
    */
   def appTypeToFormattedByType(appType: AppType): FormattedBy =
     appType match {
-      case Galaxy                     => FormattedBy.Galaxy
-      case Custom                     => FormattedBy.Custom
-      case Cromwell | Wds | HailBatch => FormattedBy.Cromwell
+      case Galaxy                               => FormattedBy.Galaxy
+      case Custom                               => FormattedBy.Custom
+      case Cromwell | Wds | HailBatch | RStudio => FormattedBy.Cromwell
```
rtitle: Haven't made it through the rest of the PR yet.. but wondering if we do actually want to ensure that RStudio disks are only used with RStudio apps? Since I think RStudio does rely on disk state (what with welder, etc).

http/src/main/resources/leo.conf
```diff
@@ -27,6 +27,9 @@ gke {
   hail-batch-app-config {
```
rtitle: This isn't your change, but is `hail-batch-app-config` in the wrong place here? I think it should be in `azure {  }` not `gke { .. }`. Probably my mistake 😞

LizBaldo: You might be right, happy to move it!

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpRStudioAppDAO.scala
```diff
@@ -0,0 +1,36 @@
...
+import org.broadinstitute.dsde.workbench.openTelemetry.OpenTelemetryMetrics
+import org.http4s._
+import org.http4s.client.Client
+import org.http4s.client.dsl.Http4sClientDsl
+import org.http4s.headers.Authorization
+
+/**
+ * HTTP client for RStudio app.
+ */
+class HttpRStudioAppDAO[F[_]](httpClient: Client[F])(implicit
```
rtitle: Just thinking about this -- for GCP we abstracted all app status checking behind the common [HttpAppDAO](https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpAppDAO.scala). It's only for Azure that we started making DAOs per service -- partly because of the "combined" nature of many of the Azure apps (e.g. a single app instance exposes many services we need to status check).

I guess I'm wondering is `HttpRStudioAppDAO` actually needed or could we leverage `HttpAppDAO` for status checking?

rtitle: … I guess we'll use this when/if we support rstudio-app on Azure so it's probably good to have. 👍

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -1194,6 +1209,94 @@ class GKEInterpreter[F[_]](
...
+        s"Installing helm chart for RStudio app ${appName.value} in cluster ${cluster.getClusterId.toString}"
+      )
+
+      googleProject <- F.fromOption(
+        LeoLenses.cloudContextToGoogleProject.get(cluster.cloudContext),
+        new RuntimeException("trying to create an azure runtime in GKEInterpreter. This should never happen")
+      )
+
+      // Create the staging bucket to be used by Welder
+      stagingBucketName = generateUniqueBucketName("leostaging-" + appName.value)
```
rtitle, line 1242: Does welder actually use the staging bucket for anything? I know it talks to the workspace bucket but I don't recall a dependency on the extra staging bucket.

Just wondering if this migrated from the GCE installation -- we could avoid a bucket creation if we can be sure welder doesn't use it.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -1194,6 +1209,94 @@ class GKEInterpreter[F[_]](
...
+
+      // Currently we always retry.
+      // The main failure mode here is helm install, which does not have easily interpretable error codes
+      retryConfig = RetryPredicates.retryAllConfig
+      _ <- tracedRetryF(retryConfig)(
+        helmInstall,
+        s"helm install for RSTUDIO app ${appName.value} in project ${cluster.cloudContext.asString}"
+      ).compile.lastOrError
+
+      googleProject <- F.fromOption(
```
rtitle: Nit: you already have the `googleProject` variable, don't need to grab it again

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -1194,6 +1209,94 @@ class GKEInterpreter[F[_]](
...
+        LeoLenses.cloudContextToGoogleProject.get(cluster.cloudContext),
+        new RuntimeException("trying to create an azure runtime in GKEInterpreter. This should never happen")
+      )
+      // Poll the app until it starts up
+      last <- streamFUntilDone(
+        config.RStudioAppConfig.services
+          .map(_.name)
+          .traverse(s => appDao.isProxyAvailable(googleProject, appName, s)),
```
rtitle, line 1283: Yeah related to above comment -- I think here it's just using `HttpAppDAO` for polling, not `HttpRStudioAppDAO`.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpRStudioAppDAO.scala
LizBaldo: Yeah, I am thinking about our future app enablement work, it might be nice to have this be explicitly for Rstudio app

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
LizBaldo: Good catch, thanks!

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
LizBaldo: I might have misunderstood what the Rstudio chart wanted, it is asking for a staging bucket, but I guess that I should feed it the workspace bucket instead of creating a new one. I got confuse with how things are set up in GCE

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -1596,6 +1699,59 @@ class GKEInterpreter[F[_]](
...
+      raw"""welder.extraEnv.name=OWNER_EMAIL""",
+      raw"""welder.extraEnv.value=${userEmail.value}"""
+    )
+
+    List(
+      // Node selector
+      raw"""nodeSelector.cloud\.google\.com/gke-nodepool=${nodepoolName.value}""",
+      // Persistence
+      raw"""persistence.size=${disk.size.gb.toString}G""",
+      raw"""persistence.gcePersistentDisk=${disk.name.value}""",
```
rtitle, line 1746: OK so we are attaching the persistent disk to the app here. We may want to introduce a `FormattedBy.RStudio` enum to the disk object. I'm also curious what happens at app deletion time -- can it successfully detach the disk and re-attach to a new app instance? I know Galaxy handles that today but not honestly sure how much of the disk logic is portable to other apps.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
rtitle: We should check the actual welder config for what it expects, what I am thinking is Yonghao just copy-pasted the staging bucket param when he made the helm chart :) 

(copy-pasting all the way down)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
LizBaldo: Okay, let me dig into that, both cromwell and galaxy have this restore logic that I should likely try to reproduce

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
rtitle: To be clear I think this code is doing the right thing, just was commenting re usage of HttpAppDAO vs HttpRstudioAppDAO

rtitle (commented):
Overall looks good. Made some comments, honestly went a little fast but can revisit. Main thing I'm curious about is whether disk detach/reattachment works out of the box. 

Also curious how much you were able to test -- any issues when running this stuff on a BEE?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
LizBaldo: But maybe I should switch to using the HttpRstudioAppDAO get status instead?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
rtitle: Yeah I know for Galaxy we explicitly tried to support detachable-PDs. I don't think Cromwell actually uses the PD for anything (it doesn't store any state on the disk). For RStudio though we probably actually want to support detachable PDs if possible.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
LizBaldo: Yeah, I do not see any mention of any of the welder extraEnv that Yonghao added in the chart in leonardo. Looks like Welder configs are:
```
welder {
  # Set to deploy welder to clusters with the given label
  deployWelderLabel = "saturnVersion"

  # Set to upgrade welder on clusters with the given label
  updateWelderLabel = "saturnVersion"

  # Leo will only deploy welder to clusters created after this date.
  # Clusters created prior to this date will not have welder deployed and will have delocalization disabled.
  deployWelderCutoffDate = "2019-08-01"

  welderReservedMemory = 768m
}
```


But that would make any of the Welder-specific values in the chart (google project, staging bucket, cluster name and owner email) obsolete?

rtitle: This is the welder conf, you can see the supported env vars:

https://github.com/DataBiosphere/welder/blob/master/server/src/main/resources/reference.conf

Well sure enough, I see
```
staging-bucket-name = ${STAGING_BUCKET}
```

I wonder what it's used for… but anyway this probably answers my question about whether it's actually needed. :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
rtitle: No strong preference, I think `HttpAppDAO` is probably fine. I think they are both doing the same thing, just hit the root URL and check for an ok http status.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
LizBaldo: Just talking with Justin, it is used to dump the welder logs!

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -1194,6 +1213,95 @@ class GKEInterpreter[F[_]](
...
+                                                             nodepoolName,
+                                                             namespaceName,
+                                                             disk,
+                                                             ksaName,
+                                                             userEmail,
+                                                             stagingBucketName
+      )
+      _ <- logger.info(ctx.loggingCtx)(s"Chart override values are: $chartValues")
+
+      _ <- F.delay("rm -rf /leonardo/aou-rstudio-chart" !!)
```
LizBaldo: NOTE TO SELF: REMOVE BEFORE MERGING!!!

LizBaldo:
@yonghaoy and @rtitle I still want to do some more testing tomorrow morning, especially around the PD reattachment logic, but I fell like this is a good stop for a another round of review :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -1596,6 +1704,61 @@ class GKEInterpreter[F[_]](
...
+    val ingress = List(
+      raw"""ingress.enabled=true""",
+      raw"""ingress.annotations.nginx\.ingress\.kubernetes\.io/auth-tls-secret=${namespaceName.value}/ca-secret""",
+      raw"""ingress.annotations.nginx\.ingress\.kubernetes\.io/proxy-redirect-from=https://${k8sProxyHost}""",
+      raw"""ingress.annotations.nginx\.ingress\.kubernetes\.io/proxy-redirect-to=${leoProxyhost}""",
+      raw"""ingress.annotations.nginx\.ingress\.kubernetes\.io/rewrite-target=/${rewriteTarget}""",
+      raw"""ingress.hosts[0].host=${k8sProxyHost}""",
+      raw"""ingress.hosts[0].paths[0].path=${rstudioIngressPath}${"(/|$)(.*)"}""",
+      raw"""ingress.hosts[0].paths[0].backend.service.name=aou-rstudio""",
+      raw"""ingress.hosts[0].paths[0].backend.service.port.number=8787""",
```
yonghaoy: should `service.name, service.port.number` provided by Helm so that it can match the port/service name provided in `deployment.yam`, and `svc.yaml`?

yonghaoy (approved):
Thanks!

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
LizBaldo: i could not find a way of gracefully do that though given that we are getting the path from leo within this loop

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
LizBaldo: Soooo, I was hoping to avoid having to pass custom environment variables for this app but it looks like the PD reattachement logic depends on knowing the `WORKSPACE_NAME_KEY` so I need to make a small follow up commit tomorrow to add that as part of the app install.

yonghaoy: `WORKSPACE_NAME_KEY ` or `WORKSPACE_NAME`? Cromwell requires `WORKSPACE_NAME` and I just merge a PR in AoU to set that for both RStudio and Cromwell. If RStudio just needs WORKSPACE_NAME then it would work for now.

LizBaldo: They should be the same: `val WORKSPACE_NAME_KEY = "WORKSPACE_NAME"`,  Looking into the db now to understand how this information is stored and maybe it is just a matter of sending a create app request with the `WORKSPACE_NAME` in it as a custom environment variable, maybe there is no code change needed

LizBaldo: Yep, everything works great and the disk is reattaching without any lost data! Just need to send the createApp request with a `WORKSPACE_NAME`, I will update the testing steps above :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpAppDAO.scala
```diff
@@ -14,18 +14,32 @@ class HttpAppDAO[F[_]: Async](val kubernetesDnsCache: KubernetesDnsCache[F], cli
...
-      case HostReady(targetHost, _, _) => // Update once we support Relay for apps
-        client
-          .successful(
-            Request[F](
-              method = Method.GET,
-              uri = Uri.unsafeFromString(
-                s"https://${targetHost.address}/proxy/google/v1/apps/${googleProject.value}/${appName.value}/${serviceName.value}/"
+      case HostReady(targetHost, _, _) =>
+        serviceName match {
+          case ServiceName("welder-service") =>
```
rtitle, line 19: Maybe this could be DRY'd a little bit since only the URL is different?
```
case HostReady(targetHost, _, _) =>
  val url = serviceName match {
    case ServiceName("welder") => "…/status"
    case _ => …
  }
  client.successful(…)
}
```

rtitle: Also in the back of my mind is restructuring code so adding new apps/services don't entail adding a million new `case` statements… but that's out of scope of this PR I think.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
rtitle: It might be nice if the `ingress.yaml` in the helm chart were a little less templated, so Leo didn't have to pass these hardcoded values..

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -1596,6 +1698,61 @@ class GKEInterpreter[F[_]](
...
+      raw"""ingress.tls[0].hosts[0]=${k8sProxyHost}"""
+    )
+
+    val welder = List(
+      raw"""welder.extraEnv[0].name=GOOGLE_PROJECT""",
+      raw"""welder.extraEnv[0].value=${cluster.cloudContext.asString}""",
+      raw"""welder.extraEnv[1].name=STAGING_BUCKET""",
+      raw"""welder.extraEnv[1].value=${stagingBucket.value}""",
+      raw"""welder.extraEnv[2].name=CLUSTER_NAME""",
+      raw"""welder.extraEnv[2].value=${cluster.clusterName.value}""",
```
rtitle: Perhaps this should be the appName?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
rtitle: 🎉

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -1733,6 +1890,7 @@ final case class GKEInterpreterConfig(vpcNetworkTag: NetworkTag,
                                       galaxyAppConfig: GalaxyAppConfig,
                                       cromwellAppConfig: CromwellAppConfig,
                                       customAppConfig: CustomAppConfig,
+                                      RStudioAppConfig: RStudioAppConfig,
```
rtitle: Nit: lower-case variable name `rstudioAppConfig: RStudioAppConfig`

rtitle (approved):
Some minor nits but overall looks good! ✅ 

Great job putting this together and testing everything.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
LizBaldo: Okay, let me have a second pass at the`ingress.yaml` file

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
LizBaldo: Ohhh, I thought this was the k8s name! But from Welder's perspective this was intended as the cluster as in dataproc cluster name?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpAppDAO.scala
LizBaldo: That makes a lot of sense, I need to step up my scala syntax a bit, thanks!

yonghaoy (approved):
This LG. 
one more question, when do you expect a new client lib to get published? @jmthibault79 is working on integrating with java Leo client lib now.

LizBaldo:
a snapshot version is published for each PR commit, and a bumped up version will be publish once I merge this PR to dev

jmthibault79:
@yonghaoy note that we are currently blocked from switching to the client by https://broadworkbench.atlassian.net/browse/IA-4329

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
LizBaldo: I think I greatly simplified the ingress yaml file, see https://github.com/DataBiosphere/terra-app/pull/34/commits/389f058075ecbd516ef722d86dfeb787194a371b

Everything is still working great on my BEE, but I am hesitant merging this PR around 5pm ET on a Friday before a long weekend in case nobody is around to help me revert it so I might wait until Monday evening / Tuesday morning 😬

LizBaldo:
jenkins retest

LizBaldo:
jenkins retest

LizBaldo:
jenkins retest

LizBaldo:
FYI @Qi77Qi this is the main leo PR enabling RSTUDIO on GKE

PR #3349, opened 2023-05-12
[IA-4081] CORS hardening: require Origin header on proxy endpoints

mlilynolting:
A simpler attempt at requiring the Origin header.

__Jira ticket__: https://broadworkbench.atlassian.net/browse/IA-4081

## Summary of changes

When Leonardo receives a request to a `/proxy` or `/notebook` endpoint, it authenticates the request using the provided Cookie and serves the response as a proxied call to an external resource. If the request's Origin header misses the referer allowlist (in app configs/terra-helmfile), deny it. If it hits the allowlist, respond with an Access-Control-Allow-Origin header (= the request origin). 

If the Origin header is not supplied, respond with a wildcard, to support existing proxy behavior.

RefererConfig is already used as an allowlist for the Referer header. To allow testing strict origin rules in environments which require permissive referer rules, add the `refererConfig.originStrict` config.

### Testing on a BEE
BEEs are configured in `terra-helmfile` to allow all ("*") referers, since each BEE has a unique hostname. 
You can push a branch to `terra-helmfile` with changes to `/values/app/leonardo/bee.yaml.gotmpl`:
* Configure `refererConfig.originStrict` to `true`
* Add your own BEE's hostname to VALID_HOSTS for `leonardo` and `terra` - e.g.
```
        # Referer Config
        VALID_HOSTS.0: localhost:3000
        VALID_HOSTS.1: bvdp-saturn-dev.appspot.com
        VALID_HOSTS.2: leonardo.mnolting.bee.envs-terra.bio
        VALID_HOSTS.3: terra.mnolting.bee.envs-terra.bio
        VALID_HOSTS.4: "*"
        IS_REFERER_CONFIG_ENABLED: true
        IS_ORIGIN_STRICT: true
```
Then deploy Leonardo on your BEE against your custom helmchart.

#### Naive
With _either_ a configuration that contains no wildcard in `refererConfig.validHosts` or has `refererConfig.originStrict true`:
- Make a request against any Leo /proxy or /notebook endpoint..
 - [ ] With an Origin header which is not in the refererConfig => Failure
	- [ ] Without an Origin header => Success (Access-Control-Allow-Origin: *)
	- [ ] With an Origin header https://foo.gov in the refererConfig => Success (Access-Control-Allow-Origin: https://foo.gov)

#### Thorough
- Create a runtime and an Analysis. Open it - the UI makes tens of requests against `/proxy` and `/notebook` endpoints to grab external resources. This should all Just Work.
- Duplicate any of these requests, add a valid Origin, and resend over cURL - should succeed
- Duplicate and add an _invalid_ Origin header. Resend over cURL - should deny access.

### Who tested and where

- [x] I validated this change 

### DIFF
--- a/http/src/main/resources/leo.conf
+++ b/http/src/main/resources/leo.conf
@@ -72,6 +72,7 @@ refererConfig {
   # Hence we include '*' for fiab and alpha environments.
   validHosts = ${?VALID_HOSTS}
   enabled = ${?IS_REFERER_CONFIG_ENABLED}
+  originStrict = ${?IS_ORIGIN_STRICT}
 }
 
 oidc {
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -806,6 +806,7 @@ refererConfig {
   # list of valid hosts and enabled flag is environment specific so it's overriden in firecloud develop
   validHosts = []
   enabled = false // false for dev and QA environments, true everywhere else
+  originStrict = false // true only if "*" in validHosts is to be ignored by origin validation
 }
 
 swagger {
--- a/http/src/main/resources/swagger/api-docs.yaml
+++ b/http/src/main/resources/swagger/api-docs.yaml
@@ -2113,8 +2113,6 @@ paths:
   ## Proxy API ##
 
   "/proxy/{googleProject}/{runtimeName}/jupyter":
-    parameters:
-      - $ref: '#/components/parameters/originHeader'
     get:
       summary: Access Jupyter (if installed) on a Leonardo runtime
       description: >
@@ -2172,8 +2170,6 @@ paths:
               schema:
                 $ref: "#/components/schemas/ErrorReport"
   "/proxy/{googleProject}/{runtimeName}/jupyter/lab":
-    parameters:
-      - $ref: '#/components/parameters/originHeader'
     get:
       summary: Access Jupyter Lab (if installed) on a Leonardo runtime
       description: >
@@ -2231,8 +2227,6 @@ paths:
               schema:
                 $ref: "#/components/schemas/ErrorReport"
   "/proxy/{googleProject}/{runtimeName}/rstudio":
-    parameters:
-      - $ref: '#/components/parameters/originHeader'
     get:
       summary: Access RStudio (if installed) on a Leonardo runtime
       description: >
@@ -2290,8 +2284,6 @@ paths:
               schema:
                 $ref: "#/components/schemas/ErrorReport"
   "/proxy/google/v1/apps/{googleProject}/{appName}/{serviceName}":
-    parameters:
-      - $ref: '#/components/parameters/originHeader'
     get:
       summary: Access an app if the app and service name exists
       description: >
@@ -2355,8 +2347,6 @@ paths:
               schema:
                 $ref: "#/components/schemas/ErrorReport"
   "/proxy/setCookie":
-    parameters:
-      - $ref: '#/components/parameters/originHeader'
     get:
       summary: Sets a browser cookie needed to authorize connections to a Leonardo runtime
       description: >
@@ -2386,8 +2376,6 @@ paths:
               schema:
                 $ref: "#/components/schemas/ErrorReport"
   "/proxy/invalidateToken":
-    parameters:
-      - $ref: '#/components/parameters/originHeader'
     get:
       summary: Invalidates a token
       description: >
@@ -2409,8 +2397,6 @@ paths:
               schema:
                 $ref: "#/components/schemas/ErrorReport"
   "/proxy/{googleProject}/{runtimeName}/setCookie":
-    parameters:
-      - $ref: '#/components/parameters/originHeader'
     get:
       deprecated: true
       summary: (Deprecated) Sets a browser cookie needed to authorize connections to a Leonardo runtime
@@ -2459,8 +2445,6 @@ paths:
               schema:
                 $ref: "#/components/schemas/ErrorReport"
   "/proxy/{googleProject}/{runtimeName}/yarn":
-    parameters:
-      - $ref: '#/components/parameters/originHeader'
     get:
       summary: YARN Resource Manager
       description: >
@@ -2516,8 +2500,6 @@ paths:
               schema:
                 $ref: "#/components/schemas/ErrorReport"
   "/proxy/{googleProject}/{runtimeName}/jobhistory":
-    parameters:
-      - $ref: '#/components/parameters/originHeader'
     get:
       summary: MapReduce Job History
       description: >
@@ -2573,8 +2555,6 @@ paths:
               schema:
                 $ref: "#/components/schemas/ErrorReport"
   "/proxy/{googleProject}/{runtimeName}/apphistory":
-    parameters:
-      - $ref: '#/components/parameters/originHeader'
     get:
       summary: YARN Application Timeline
       description: >
@@ -2630,8 +2610,6 @@ paths:
               schema:
                 $ref: "#/components/schemas/ErrorReport"
   "/proxy/{googleProject}/{runtimeName}/sparkhistory":
-    parameters:
-      - $ref: '#/components/parameters/originHeader'
     get:
       summary: Spark History Server
       description: >
@@ -2687,8 +2665,6 @@ paths:
               schema:
                 $ref: "#/components/schemas/ErrorReport"
   "/proxy/{googleProject}/{runtimeName}/hdfs":
-    parameters:
-      - $ref: '#/components/parameters/originHeader'
     get:
       summary: HDFS NameNode
       description: >
@@ -2749,16 +2725,6 @@ paths:
 ##########################################################################################
 
 components:
-  parameters:
-    originHeader:
-      - in: header
-        name: Origin
-        schema:
-          type: string
-          enum:
-            - "https://example.com" # how do I make this render the current host?
-        required: true
-        description: The Origin header
   requestBodies:
     CreateRuntimeRequest:
       content:
@@ -2824,6 +2790,7 @@ components:
             labels: { saturnRuntimeName: "saturn-ab9134" }
             machineSize: "Standard_DS1_v2"
             disk: { labels: {}, name: "disk1", size: 50 }
+
   securitySchemes:
     oidc:
       type: oauth2
--- a/http/src/main/resources/swagger/api-docs.yaml
+++ b/http/src/main/resources/swagger/api-docs.yaml
@@ -2113,6 +2113,8 @@ paths:
   ## Proxy API ##
 
   "/proxy/{googleProject}/{runtimeName}/jupyter":
+    parameters:
+      - $ref: '#/components/parameters/originHeader'
     get:
       summary: Access Jupyter (if installed) on a Leonardo runtime
       description: >
@@ -2170,6 +2172,8 @@ paths:
               schema:
                 $ref: "#/components/schemas/ErrorReport"
   "/proxy/{googleProject}/{runtimeName}/jupyter/lab":
+    parameters:
+      - $ref: '#/components/parameters/originHeader'
     get:
       summary: Access Jupyter Lab (if installed) on a Leonardo runtime
       description: >
@@ -2227,6 +2231,8 @@ paths:
               schema:
                 $ref: "#/components/schemas/ErrorReport"
   "/proxy/{googleProject}/{runtimeName}/rstudio":
+    parameters:
+      - $ref: '#/components/parameters/originHeader'
     get:
       summary: Access RStudio (if installed) on a Leonardo runtime
       description: >
@@ -2284,6 +2290,8 @@ paths:
               schema:
                 $ref: "#/components/schemas/ErrorReport"
   "/proxy/google/v1/apps/{googleProject}/{appName}/{serviceName}":
+    parameters:
+      - $ref: '#/components/parameters/originHeader'
     get:
       summary: Access an app if the app and service name exists
       description: >
@@ -2347,6 +2355,8 @@ paths:
               schema:
                 $ref: "#/components/schemas/ErrorReport"
   "/proxy/setCookie":
+    parameters:
+      - $ref: '#/components/parameters/originHeader'
     get:
       summary: Sets a browser cookie needed to authorize connections to a Leonardo runtime
       description: >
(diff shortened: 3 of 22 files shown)


http/src/main/resources/swagger/api-docs.yaml
mlilynolting, line 1: This doesn't actually populate the Origin header - I tried to follow https://swagger.io/docs/specification/describing-parameters/ and combine Common Parameters for Various Paths + Constant Parameters, but the result is garbage.

project/Settings.scala
mlilynolting, line 1: revert this

rtitle:
Hey @mlilynolting,

I was just looking at this PR and doing some testing on dev. The Leo code to validate `Origin` seems correct, but from my dev testing I'm not actually seeing `Origin` come through from Terra UI iframe requests. 

Here is a request (copied as curl) from the jupyter iframe:
```
curl 'https://leonardo.dsde-dev.broadinstitute.org/proxy/terra-dev-9daa5eba/saturn-d3c5419c-98f6-43d7-a497-6f3824221f31/jupyter/notebooks/RT%20-%20UI%20test/edit/Test.ipynb' \
  -H 'authority: leonardo.dsde-dev.broadinstitute.org' \
  -H 'accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3;q=0.7' \
  -H 'accept-language: en-US,en;q=0.9' \
  -H 'cache-control: no-cache' \
  -H 'cookie: GCP_IAP_UID=100182905304480635229; LeoToken=<token>' \
  -H 'pragma: no-cache' \
  -H 'referer: https://bvdp-saturn-dev.appspot.com/' \
  -H 'sec-ch-ua: "Google Chrome";v="113", "Chromium";v="113", "Not-A.Brand";v="24"' \
  -H 'sec-ch-ua-mobile: ?0' \
  -H 'sec-ch-ua-platform: "macOS"' \
  -H 'sec-fetch-dest: iframe' \
  -H 'sec-fetch-mode: navigate' \
  -H 'sec-fetch-site: cross-site' \
  -H 'sec-fetch-user: ?1' \
  -H 'upgrade-insecure-requests: 1' \
  -H 'user-agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/113.0.0.0 Safari/537.36' \
  --compressed
```

Here is a (non-iframe) AJAX request from the UI:
```
curl 'https://leonardo.dsde-dev.broadinstitute.org/api/google/v1/disks?role=creator&saturnWorkspaceName=RT%20-%20UI%20test&saturnWorkspaceNamespace=general-dev-billing-account&includeLabels=saturnApplication%2CsaturnWorkspaceName%2CsaturnWorkspaceNamespace' \
  -H 'authority: leonardo.dsde-dev.broadinstitute.org' \
  -H 'accept: */*' \
  -H 'accept-language: en-US,en;q=0.9' \
  -H 'authorization: Bearer <token>' \
  -H 'cache-control: no-cache' \
  -H 'origin: https://bvdp-saturn-dev.appspot.com' \
  -H 'pragma: no-cache' \
  -H 'referer: https://bvdp-saturn-dev.appspot.com/' \
  -H 'sec-ch-ua: "Google Chrome";v="113", "Chromium";v="113", "Not-A.Brand";v="24"' \
  -H 'sec-ch-ua-mobile: ?0' \
  -H 'sec-ch-ua-platform: "macOS"' \
  -H 'sec-fetch-dest: empty' \
  -H 'sec-fetch-mode: cors' \
  -H 'sec-fetch-site: cross-site' \
  -H 'user-agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/113.0.0.0 Safari/537.36' \
  -H 'x-app-id: Saturn' \
  --compressed
```

I was reading about [Origin](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Origin#description) and it does seem like it can be `null` in some circumstances. Maybe this makes sense that it's set for actual cross-origin AJAX requests, but not for same-origin requests like the Jupyter iframe. 

Not sure if there's a way we can force the iframe to always send `Origin`. Or maybe we need to slightly change our Leo validation approach. Not sure yet (will sleep on it) but wanted to raise..

rtitle:
I wonder if a reasonable solution would be for Leo to validate Origin only if it’s set; and not do any validation if Origin is unset. That may still prevent these cross-origin attack scenarios, but still allow direct page loads to work. Should validate this with AppSec, but that’s 1 potential idea.

http/src/main/resources/leo.conf
```diff
@@ -72,6 +72,7 @@ refererConfig {
   # Hence we include '*' for fiab and alpha environments.
   validHosts = ${?VALID_HOSTS}
   enabled = ${?IS_REFERER_CONFIG_ENABLED}
+  originStrict = ${?IS_ORIGIN_STRICT}
```
rtitle, line 75: Would we ever set this to true? Seems like this flag is used to "override" the presence of `*` in validHosts. Looks like we only include `*` in BEE and alpha environments today (I think that's because the hostnames are dynamic). Just thinking if this config will always be false, then why have it :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/CorsSupport.scala
```diff
@@ -23,20 +27,49 @@ class CorsSupport(contentSecurityPolicy: ContentSecurityPolicyConfig) {
...
+   * Enable strict mode by either removing the wildcard (*) from
+   * refererConfig.validHosts or setting refererConfig.originStrict to true.  */
+  private val isOriginStrict: Boolean = !refererConfig.validHosts.contains("*") || refererConfig.originStrict
+
+  private val handleOrigin: Directive0 =
+    if (!refererConfig.enabled || !isOriginStrict) pass
+    else
+      optionalHeaderValueByType(Origin) flatMap {
+        case Some(origin) =>
+          logger.info(s"CorsSupport | Request origin detected: ${origin}. Checking validity")
```
rtitle: Wondering if these should be debug-level logs, since every proxy request will pass through this..

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/CorsSupport.scala
```diff
@@ -23,20 +27,49 @@ class CorsSupport(contentSecurityPolicy: ContentSecurityPolicyConfig) {
...
+          logger.info(s"CorsSupport | Request origin not provided. Permitting")
+          pass
+      }
+
+  private def getValidOriginRange: HttpOriginRange.Default = {
+    def validOrigins: Set[HttpOrigin] = refererConfig.validHosts
+      .filter(_ != "*")
+      .map(uri => if (uri.last == '/') uri.slice(0, uri.length - 1) else uri)
+      .flatMap { uriString =>
+        Set(HttpOrigin(s"http://${uriString}"), HttpOrigin(s"https://${uriString}"))
```
rtitle, line 59: Curious if adding both schemes is necessary, or if you could just do `HttpOrigin(uriString)` and akka would match it.

http/src/main/resources/reference.conf
```diff
@@ -807,6 +807,7 @@ refererConfig {
   # list of valid hosts and enabled flag is environment specific so it's overriden in firecloud develop
   validHosts = []
   enabled = false // false for dev and QA environments, true everywhere else
+  originStrict = false // true only if "*" in validHosts is to be ignored by origin validation
```
LizBaldo, line 809: Can you link the sister PR in terra helmfile that does this? it is related to Rob's comment above

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/RuntimeV2Routes.scala
```diff
@@ -19,15 +19,15 @@ import JsonCodec._
...
 
-class RuntimeV2Routes(saturnIframeExtentionHostConfig: RefererConfig,
+class RuntimeV2Routes(saturnIframeExtensionHostConfig: RefererConfig,
                       runtimeV2Service: RuntimeV2Service[IO],
                       userInfoDirectives: UserInfoDirectives
 )(implicit
   metrics: OpenTelemetryMetrics[IO]
 ) {
   // See https://github.com/DataBiosphere/terra-ui/blob/ef88f396a61383ee08beb65a37af7cae9476cc20/src/libs/ajax.js#L1358
   private val allValidSaturnIframeExtensions =
```
rtitle, line 29: Forgot we were using `RefererConfig` here as well. This seems duplicative of `RuntimeRoutes`, could we factor out to a common method?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/CorsSupport.scala
```diff
@@ -23,20 +27,49 @@ class CorsSupport(contentSecurityPolicy: ContentSecurityPolicyConfig) {
     )
   }
 
+  /** Whether to allow any Origin header value or to restrict to the allowlist.
+   * Enable strict mode by either removing the wildcard (*) from
```
LizBaldo, line 40: This feels redundant, aka the flag is set to True only in environments where there is a wildcard (BEE and alpha), so why have it?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/CorsSupport.scala
```diff
@@ -23,20 +27,49 @@ class CorsSupport(contentSecurityPolicy: ContentSecurityPolicyConfig) {
...
+
+  private val handleOrigin: Directive0 =
+    if (!refererConfig.enabled || !isOriginStrict) pass
+    else
+      optionalHeaderValueByType(Origin) flatMap {
+        case Some(origin) =>
+          logger.info(s"CorsSupport | Request origin detected: ${origin}. Checking validity")
+          checkSameOrigin(getValidOriginRange)
+        case None =>
+          logger.info(s"CorsSupport | Request origin not provided. Permitting")
```
LizBaldo: Have you checked with AppSec that this was okay?

rtitle (approved):
A few nits but this looks good! I also did a quick test on my BEE and I think I see the desired behavior -- proxy requests succeed with a valid Origin or no Origin; and fail with 403 with an invalid origin. 🚀

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/CorsSupport.scala
LizBaldo: I second that, this might make our logs a bit too noisy

mlilynolting: These were mostly for debugging, and I think are safe to remove entirely

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/CorsSupport.scala
mlilynolting: Needed! the `uriString` here omits the scheme, `HttpOrigin` requires it, and I wasn't sure which of http/https we'd be expecting (and doing one doesn't automatically allow the other)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/RuntimeV2Routes.scala
mlilynolting: It's duplicative but this variable is unused - I created https://broadworkbench.atlassian.net/browse/IA-4317 to track this

http/src/main/resources/reference.conf
mlilynolting: There's no sister PR because this config is really only needed in debugging scenarios, in environments (BEEs) where the "*" in validHosts is required for functionality. So this defaults to false and can be optionally set to true to tweak behavior (when you want referers permissive but origins strict). Once we're happy with this PR, we could optionally remove this setting entirely.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/CorsSupport.scala
mlilynolting: Setting this to true when there is a wildcard in the list lets Origins be checked against a strict allowlist while any Referer is permitted.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/CorsSupport.scala
mlilynolting: Will confirm today.

http/src/main/resources/leo.conf
mlilynolting: Explained below!

mlilynolting:
jenkins retest

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/RuntimeRoutes.scala
```diff
@@ -36,7 +36,15 @@ class RuntimeRoutes(saturnIframeExtentionHostConfig: RefererConfig,
...
-    saturnIframeExtentionHostConfig.validHosts.map(s => s"https://${s}/jupyter-iframe-extension.js")
+    saturnIframeExtensionHostConfig.validHosts
+      .filter(_ != "*")
+      .map(uri => if (uri.last == '/') uri.slice(0, uri.length - 1) else uri)
+      .map(s => s"https://${s}/jupyter-iframe-extension.js")
+
+  private def isValidSaturnIframeExtension(uri: String): Boolean =
+    !saturnIframeExtensionHostConfig.enabled ||
+      saturnIframeExtensionHostConfig.validHosts.contains("*") ||
+      allValidSaturnIframeExtensions(uri)
```
mlilynolting, line 47: These changes I think constitute a bugfix - the check for valid iframe extensions previously ignored the refererConfig's `enabled` flag.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/RuntimeV2Routes.scala
mlilynolting: Realized that the field this is checked against for v1 runtime create doesn't apply to v2 runtimes (I think?) Closed the ticket.

mlilynolting:
jenkins retest

PR #3361, opened 2023-05-22
[IA-4343] Set dateAccessed to current time on runtime create

mlilynolting:
Subtask of
__Jira ticket__: https://broadworkbench.atlassian.net/browse/IA-4283

### What

Use Instant.now instead of ctx.now for dateAccessed when setting runtime status to Running.

### Why

`AzurePubsubHandler#monitorCreateRuntime` has an `implicit ev: Ask[F, AppContext]` param, and sets `ctx <- ev.ask` as its first step. `ctx` thus is an instance of `AppContext`, whose `generate` method initializes `now` as the current `java.time.Instant` as of the initial HTTP create request. So `ctx.now` is always equal to the request time.

`monitorCreateRuntime` polls WorkspaceManager until the create action succeeds, at which point it sets the runtime’s `status`, `hostIp`, and `dateAccessed`. `dateAccessed` is set to `ctx.now` - which is the request time, not the time that the runtime status changes to "Running".

### What to test

- Create an Azure runtime through the UI with an `autopauseThreshold` = X minutes; note the time (t0)
- When runtime starts (moves from Creating -> Running status), note the time (t1)
- X minutes after creation request (t0 + X) runtime is running, not paused
- X minutes after start (t1 + X) runtime is paused

### Who tested and where

- [ ] I validated this change 

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
@@ -414,7 +414,6 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
       v.jobReport.status.equals(WsmJobStatus.Succeeded) || v.jobReport.status == WsmJobStatus.Failed
     for {
       ctx <- ev.ask
-
       auth <- samDAO.getLeoAuthToken
       getWsmJobResult = wsmDao.getCreateVmJobResult(GetJobResultRequest(params.workspaceId, params.jobId), auth)
 
@@ -463,7 +462,7 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
             val hostIp = s"${params.relayNamespace.value}.servicebus.windows.net"
 
             for {
-              _ <- clusterQuery.updateClusterHostIp(params.runtime.id, Some(IP(hostIp)), ctx.now).transaction
+              _ <- clusterQuery.updateClusterHostIp(params.runtime.id, Some(IP(hostIp)), Instant.now).transaction
               // then poll the azure VM for Running status, retrieving the final azure representation
               _ <- streamUntilDoneOrTimeout(
                 isJupyterUp,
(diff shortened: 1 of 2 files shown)


jdcanas (approved):
ah nice catch, that would do it

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
```diff
@@ -463,7 +462,7 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
             val hostIp = s"${params.relayNamespace.value}.servicebus.windows.net"
 
             for {
-              _ <- clusterQuery.updateClusterHostIp(params.runtime.id, Some(IP(hostIp)), ctx.now).transaction
+              _ <- clusterQuery.updateClusterHostIp(params.runtime.id, Some(IP(hostIp)), Instant.now).transaction
```
rtitle: Nit: since `Instant.now` is technically a side effect, we normally wrap it in an `F` context. You can do something like:
```
for {
  now <- nowInstant
  _ <- clusterQuery.updateClusterHostIp(params.runtime.id, Some(IP(hostIp)), now).transaction
  …
} yield ()
```
…which should still have the effect of getting the current instant (rather than the instant the AppContext was created).

rtitle (approved):
Makes sense to update dateAccessed to the timestamp the runtime transitioned to `Running`. 👍 

Still curious why it's not being continually updated by the listener…

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
mlilynolting: This is definitely not second nature yet-- I appreciate the tip :)

mlilynolting:
jenkins retest

mlilynolting:
jenkins retest

mlilynolting:
jenkins retest

mlilynolting:
jenkins retest

PR #3364, opened 2023-05-22
[IA-4183][IA-4303] Merge dependabot changes and security patch

jdcanas:


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/DbReference.scala
```diff
@@ -31,7 +31,8 @@ object DbReference extends LazyLogging {
       val liquibase = new Liquibase(liquibaseConfig.changelog, resourceAccessor, liquibaseConnection)
 
       changelogParameters.foreach { case (key, value) => liquibase.setChangeLogParameter(key, value) }
-      liquibase.update(new Contexts())
+
+      liquibase.update(new Contexts(), new LabelExpression())
```
LizBaldo, line 35: What does this change do?

LizBaldo (approved):
LGTM, thanks! I just have a minor question.

Looks like the integration test failures are unrelated, but it would be nice to see them pass before merging, happy to monitor and merge if it does not happen by the time you leave for PTO :)

jdcanas:
jenkins retest

LizBaldo (approved):

PR #3367, opened 2023-05-23
[IA-4305] bump wb-libs versions for security vulnerability

jdcanas:


LizBaldo (approved):

project/Dependencies.scala
```diff
@@ -17,16 +17,16 @@ object Dependencies {
...
+  val serviceTestV = s"3.1-$workbenchLibsHash"
+  val workbenchModelV = s"0.18-$workbenchLibsHash"
 
   // TODO update to 0.26 - DataprocInterpreter relies on deprecated class MemberType
-  val workbenchGoogleV = s"0.23-4b46aac"
-  val workbenchGoogle2V = s"0.26-$workbenchLibsHash"
-  val workbenchOpenTelemetryV = s"0.4-$workbenchLibsHash"
-  val workbenchOauth2V = s"0.3-$workbenchLibsHash"
-  val workbenchAzureV = s"0.3-$workbenchLibsHash"
+  val workbenchGoogleV = s"0.27-4b46aac"
```
mlilynolting: Has the todo above been addressed? Does this not complain anymore?

project/Dependencies.scala
```diff
@@ -17,16 +17,16 @@ object Dependencies {
...
+  val serviceTestV = s"3.1-$workbenchLibsHash"
+  val workbenchModelV = s"0.18-$workbenchLibsHash"
 
   // TODO update to 0.26 - DataprocInterpreter relies on deprecated class MemberType
-  val workbenchGoogleV = s"0.23-4b46aac"
-  val workbenchGoogle2V = s"0.26-$workbenchLibsHash"
-  val workbenchOpenTelemetryV = s"0.4-$workbenchLibsHash"
-  val workbenchOauth2V = s"0.3-$workbenchLibsHash"
-  val workbenchAzureV = s"0.3-$workbenchLibsHash"
+  val workbenchGoogleV = s"0.26-$workbenchLibsHash"
```
mlilynolting: Same question as before - can we remove the TODO above? Does this work without complaining about MemberType etc?

mlilynolting (changes requested):

project/Dependencies.scala
jdcanas: ah good catch, ill revert

PR #3393, opened 2023-06-02
[IA-4268] redirect proxy to Rstudio root url

LizBaldo:
Follow up to https://github.com/DataBiosphere/leonardo/pull/3338

We need to redirect the RStudio service to its root URL to avoid sending users to the auth-sign-in page, see https://broadinstitute.slack.com/archives/CA3NP1733/p1685652363954569


yonghaoy (approved):

LizBaldo:
@Qi77Qi small follow up PR re RSTUDIO on GKE

PR #3401, opened 2023-06-06
[IA-4291] use Wsm DB to decide if runtime should be deleted

lucymcnatt:
__Jira ticket__: https://broadworkbench.atlassian.net/browse/IA4291

### What

When we decide whether to send the wsmResouceId to back leo (which determines if a delete call to WSM is made), this changes the logic to use the WSM java client to check the WSM DB as the source of truth.

### Why

Before we were relying on Leo's database and checking if the hostIp was set, but this was an unreliable metric to decide if the wsmResource should be deleted.

### What to test

- Create an azure VM
- Delete your vm
- check that it was deleted in the Azure portal and/or examine to logs to ensure there was no "no-op for wsm.deleteVm" message

### Who tested and where

- [ ] I validated this change 


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala
```diff
@@ -312,12 +314,29 @@ class RuntimeV2ServiceInterp[F[_]: Parallel](config: RuntimeServiceConfig,
           )
       }
 
-      // We only update IP to properly after the VM is created in WSM. Hence, if IP is not defined, we don't need to pass the VM resourceId
-      // to back leo, where it'll trigger a WSM call to delete the VM
-      wsmVMResourceSamId = runtime.hostIp.map(_ => WsmControlledResourceId(UUID.fromString(runtime.internalId)))
-      // If there's no controlled resource record, that means we created the DB record in front leo but failed somewhere in back leo (possibly in polling WSM)
-      // This is non-fatal, as we still want to allow users to clean up the db record if they have permission.
-      // We must check if they have permission other ways if we did not get an ID back from WSM though
+      // get wsm api
```
lucymcnatt, line 315: logic to use wsmClient to get vm status here

local/proxies.sh
```diff
@@ -62,7 +62,7 @@ dbconnect() {
         exit 1
     fi
 
-    mysql -u ${DB_USER} -p${DB_PASSWORD} --host 127.0.0.1 --port 3306
+    mysql -u ${DB_USER} -p${DB_PASSWORD} --host 127.0.0.1 --port 3306 leonardo
```
lucymcnatt, line 65: auto-connect to leo DB when running dbConnect locally, just a small tweak

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/runtimeModels.scala
```diff
@@ -315,7 +315,7 @@ object RuntimeConfig {
 
   // Azure machineType maps to `com.azure.resourcemanager.compute.models.VirtualMachineSizeTypes`
   final case class AzureConfig(machineType: MachineTypeName,
-                               persistentDiskId: DiskId,
+                               persistentDiskId: Option[DiskId],
```
LizBaldo, line 318: This is a bonus change so we won't get errored at for no reason at times?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/RuntimeConfigTable.scala
```diff
@@ -259,9 +259,7 @@ object RuntimeConfigTable {
       case CloudService.AzureVm =>
         RuntimeConfig.AzureConfig(
           machineType,
-          persistentDiskId.getOrElse(
-            throw new SQLDataException("persistentDiskId field should not be null for Azure.")
```
LizBaldo, line 263: Could you remind me in what situation is it acceptable to have a null persistent disk id?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala
```diff
@@ -312,12 +314,29 @@ class RuntimeV2ServiceInterp[F[_]: Parallel](config: RuntimeServiceConfig,
...
+      // if the vm is found in WSM and has a deletable state,
+      // then the resourceId is passed to back leo to make the delete call to WSM
+      // (state can be BROKEN, CREATING, DELETING, READY, UPDATING or NULL)
+      wsmVMResourceSamId = Try(wsmAzureResourceApi.getAzureVm(workspaceId.value, wsmResourceId.value)) match {
+        case Success(result) =>
+          val vmState = result.getMetadata.getState.getValue
+          log.info(
+            s"Runtime ${runtimeName.asString} with resourceId ${wsmResourceId.value} has a state of $vmState in WSM"
+          )
+          if (List("BROKEN", "READY").contains(vmState))
```
LizBaldo: Nit: I would create a val deletableStates with this list in to make things more explicit

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/LivenessRoutesSpec.scala
```diff
@@ -8,14 +8,16 @@ import org.broadinstitute.dsde.workbench.leonardo.db.TestComponent
...
 
 class LivenessRoutesSpec
     extends AnyFlatSpec
     with Matchers
     with ScalatestRouteTest
     with LeonardoTestSuite
     with TestComponent
-    with TestLeoRoutes {
+    with TestLeoRoutes
+    with MockitoSugar {
```
LizBaldo, line 20: I have seen this so many times by now, i really want a mojito!

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeServiceV2InterpSpec.scala
```diff
@@ -804,6 +840,115 @@ class RuntimeServiceV2InterpSpec extends AnyFlatSpec with LeonardoTestSuite with
...
+          landingZoneResources,
+          Some(context.traceId)
+        )
+      message shouldBe expectedMessage
+    }
+
+    res.unsafeRunSync()(cats.effect.unsafe.IORuntime.global)
+  }
+
+  it should "not delete a runtime in a non-deletable status" in isolatedDbTest {
```
LizBaldo: Nit: Could you precise in a leonardo non-deletable status?

local/proxies.sh
LizBaldo: Thank you!!!

project/Dependencies.scala
```diff
@@ -235,4 +243,28 @@ object Dependencies {
     http4sCirce,
     scalaTest
   )
+
+  val wsmDependencies: List[ModuleID] = List(
+    "bio.terra" % "workspace-manager-client" % "0.254.452-SNAPSHOT"
+      exclude("com.sun.activation", "jakarta.activation"),
+    /*
+    1. WSM is looking for the rs-api under javax.*.
```
LizBaldo: Should this comment be deleted or an actual TODO?

LizBaldo (approved):
Looks great to me!

The sad path is hard to reproduce since the error was not happening consistently. But since you were able to test the happy path, there should b no harm merging to dev and then see if it fixes the sad VMs in there :)

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/runtimeModels.scala
lucymcnatt: yes! I was being blocked by it in testing this change, so I thought I'd just include a fix to the `persistentDiskId field should not be null for Azure` error as well

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/RuntimeConfigTable.scala
lucymcnatt: So we started seeing this come up once Justin introduced the disk attachment/detachment logic for azure VMs. So now when you `completeDeletion` on an Azure VM (get success signal from WSM) it sets the status to 'deleted' and diskId to None for that Vm's runtimeConfig

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala
lucymcnatt: that makes sense, will do

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/LivenessRoutesSpec.scala
lucymcnatt: ooh same....maybe at state park next time

project/Dependencies.scala
lucymcnatt: oh yup that was taken directly from rawls, but that may have been before I did a fetch update, so I'll check on that and (probably) delete this

lucymcnatt:
jenkins retest

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala
```diff
@@ -312,12 +314,31 @@ class RuntimeV2ServiceInterp[F[_]: Parallel](config: RuntimeServiceConfig,
...
+      // get wsm api
+      wsmAzureResourceApi = wsmClientProvider.getControlledAzureResourceApi(userInfo.accessToken.token)
+      wsmResourceId = WsmControlledResourceId(UUID.fromString(runtime.internalId))
+
+      // if the vm is found in WSM and has a deletable state,
+      // then the resourceId is passed to back leo to make the delete call to WSM
+      // (state can be BROKEN, CREATING, DELETING, READY, UPDATING or NULL)
+      deletableStatus = List("BROKEN", "READY")
+
+      wsmVMResourceSamId = Try(wsmAzureResourceApi.getAzureVm(workspaceId.value, wsmResourceId.value)) match {
```
rtitle: Friendly Scala tip.. the blocking WSM call should probably be "lifted" into the F type and chained in the for-comp since it's a side effect. Also the `log` statements below are also `F`s but not chained to anything, so they probably won't execute as is. One way to write this could be:
```
      attempt <- F.delay(wsmAzureResourceApi.getAzureVm(workspaceId.value, wsmResourceId.value)).attempt
      wsmVMResourceSamId <- attempt match {
        case Right(result) =>
          val vmState = result.getMetadata.getState.getValue
          val res = if (deletableStatus.contains(vmState)) Some(wsmResourceId) else None
          log
            .info(ctx.loggingCtx)(
              s"Runtime ${runtimeName.asString} with resourceId ${wsmResourceId.value} has a state of $vmState in WSM"
            )
            .as(res)
        case Left(e) =>
          log
            .info(ctx.loggingCtx)(
              s"No wsm record found for runtime ${runtimeName.asString} No-op for wsmDao.deleteVm, ${e.getMessage}"
            )
            .as(None)
```

rtitle (commented):
I left a Scala comment.. but mainly wanted to say Thank you for integrating the WSM client!! 🎉

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala
lucymcnatt: ohh thank you! Yeah I realized that those logs weren't executing, that makes sense

rtitle (approved):
Thanks!

lucymcnatt:
jenkins retest

lucymcnatt:
jenkins retest

lucymcnatt:
jenkins retest

PR #3443, opened 2023-06-16
[IA-4302] upgrade app pubsub message

LizBaldo:
__Jira ticket__: https://broadworkbench.atlassian.net/browse/IA-4302

UI-related ticket implementing the new `UPDATING` app status there: https://github.com/DataBiosphere/terra-ui/pull/3983

### What

- Implements a new type of pubsub message to handle app updates (for both GKE and AKS)
- This does not yet implement the replica rescaling as app teams were hesitant about it and would rather go for some minor downtime for the MVP
- Implements the admin api endpoint to trigger the upgrade app message

### Why

- This is step 2  and 3 of the [app update designs](https://broadworkbench.atlassian.net/wiki/spaces/IA/pages/edit-v2/2746056705), and is making use of the new chartupgrade helm function provided by the helm-scala sdk repo 

## Testing these changes

TODO - Test on a BEE for each apptype/cloud context combination

### Who tested and where

- [ ] I validated this change 


http/src/main/resources/reference.conf
```diff
@@ -929,6 +929,11 @@ pubsub {
       interval = 3 seconds
       interruptAfter = 10 minutes
     }
+    updateApp {
```
LizBaldo, line 1010: I figured we could use the same as for starting an app

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/LeoPublisher.scala
```diff
@@ -118,6 +118,9 @@ final class LeoPublisher[F[_]](
           appQuery.updateStatus(m.appId, AppStatus.Stopping).transaction
         case m: LeoPubsubMessage.StartAppMessage =>
           appQuery.updateStatus(m.appId, AppStatus.Starting).transaction
+        case _: LeoPubsubMessage.UpdateAppMessage =>
+          F.unit
+          //TODO - See if we need to transition to an `UPDATING` status
```
LizBaldo: I would rather not do that as part of the MVP if we can avoid it

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/Boot.scala
```diff
@@ -644,7 +644,8 @@ object Boot extends IOApp {
         googleDiskService,
         appDescriptorDAO,
         nodepoolLock,
-        googleDependencies.googleResourceService
+        googleDependencies.googleResourceService,
+        googleDependencies.googleComputeService
```
LizBaldo, line 668: We need this so I can query the memory and gpu based on the machine type for the Galaxy app :'(

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -408,44 +581,63 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
     } yield ()
   }
 
+  private[util] def pollApp(userEmail: WorkbenchEmail, relayBaseUri: Uri, appType: AppType)(implicit
```
LizBaldo, line 678: Since polling is used for both creating and updating now, might as well make this a bit DRY

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/LeoPublisher.scala
jdcanas: Whats the logic behind this decision?

LizBaldo: Mostly consistency with the pre-existing update messages for disk and runtime

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/Boot.scala
jdcanas: Hmm.. is this needed because the machineType is in the DB but not the gpu information?

LizBaldo: Indeed :(

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/LeoPublisher.scala
jdcanas: Ah, so for google runtimes `Updating` is reserved for mirroring the actual google status of the runtime, which is only used for updating certain fields (we delete/re-create for a lot of 'updates' a user may request to their vm). In this case however, I think it is important to capture the concept that this leo resource is updating. Especially in a world where WSM is mirroring the cloud resource status already, it seems particularly relevant for leo to know "hey there is actually something going on here", especially since in this case we are doing the helm commands directly.

LizBaldo: Gotcha, I can do that, especially if we end up having some downtime, the status will be important 👍

jdcanas: Yeah, also thinking in the context of `deleteAll`. Not sure if an app being updated is safe to be deleted

LizBaldo: Good point!

jdcanas: Should be an easy UI change, but will be required if we change the app models:
- the UI model for app statuses here: https://github.com/DataBiosphere/terra-ui/blob/dev/src/libs/ajax/leonardo/models/app-models.ts#L17
- And the corresponding display status mapping and type https://github.com/DataBiosphere/terra-ui/blob/dev/src/libs/ajax/leonardo/models/app-models.ts#L45

LizBaldo: Ahh, now I remember why we did not want to add this to the MVP, but it does feel safer to include it, even if it requires a small UI change

LizBaldo: Done :) https://github.com/DataBiosphere/terra-ui/pull/3983

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/leoPubsubMessageSubscriberModels.scala
```diff
@@ -1141,6 +1166,19 @@ object PubsubHandleMessageError {
...
+
+  final case class AppNotFound(appId: Long, message: LeoPubsubMessage) extends PubsubHandleMessageError {
+    override def getMessage: String =
+      s"Unable to process transition finished message ${message} for app ${appId} because it was not found in the database"
+
+    val isRetryable: Boolean = false
+  }
+  final case class AppInvalidState(appId: Long, message: LeoPubsubMessage) extends PubsubHandleMessageError {
+    override def getMessage: String =
+      s"Unable to process transition finished message ${appId}for app ${appId} because its status was not RUNNING"
```
rtitle: Typo -- `${message}` instead of `${appId}`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -342,6 +343,215 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
...
     } yield ()
 
+  override def updateAndPollApp(params: UpdateAKSAppParams)(implicit ev: Ask[F, AppContext]): F[Unit] = {
+    for {
+      ctx <- ev.ask
+
+      workspaceId <- F.fromOption(
+        params.workspaceId,
+        AppUpdateException(
+          s"${params.appName} must have a Workspace in the Azure cloud context | trace id: ${ctx.traceId}"
```
rtitle: Should this pass `Some(ctx.traceId)` instead of embedding it in the message?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -342,6 +343,215 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
...
+        .transaction
+      dbApp <- F.fromOption(
+        dbAppOpt,
+        AppNotFoundException(CloudContext.Azure(params.cloudContext),
+                             params.appName,
+                             ctx.traceId,
+                             "No active app found in DB"
+        )
+      )
+      _ <- logger.info(ctx.loggingCtx)(s"Updating app $params.appName in workspace $params.workspaceId")
```
rtitle: Does the string interpolation need `${..}`?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -342,6 +343,215 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
...
+
+      app = dbApp.app
+      namespaceName = app.appResources.namespace.name
+      kubernetesNamespace = KubernetesNamespace(namespaceName)
+
+      // Grab the LZ and storage container information associated with the workspace
```
rtitle, line 392: The other app messages (create/delete) get the `landingZoneResources` and workspace-storage container from the pubsub message, instead of querying it here. If we're triggering UpdateApp messages via Leo API then maybe we should do it the same way. Not sure I have a strong preference though…

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -342,6 +343,215 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
...
+        case SamResourceType.SharedApp => F.pure(None)
+        case _ =>
+          for {
+            msi <- buildMsiManager(params.cloudContext)
+            petMi <- F.delay(
+              msi.identities().getById(app.googleServiceAccount.value)
+            )
+
+            // Assign the pet managed identity to the VM scale set backing the cluster node pool
+            _ <- assignVmScaleSet(landingZoneResources.clusterName, params.cloudContext, petMi)
```
rtitle: I don't think we should need to `assignVmScaleSet` for upgrades, since it should already be assigned..

rtitle (approved):
I didn't test this. But honestly it all looks correct. Also I see the logic better now in wanting to implement the front-Leo API instead of having the cron jobs emit pubsub messages directly -- front Leo can apply logic around which app types/versions are allowed to be updated; do necessary pre-setup, etc.

Nice work!

LizBaldo:
Thanks @rtitle! i will address your comment and open another PR pointing to this one that implements the new API endpoint, it will be easier to test from there :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
LizBaldo: Good catch, I must have copied this pattern from another place without thinking twice about it 🙈

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
LizBaldo: Agreed, let me take a stab at the front leo implementation and see if I can harmonize a bit more with what the create/delete endpoints are already doing

LizBaldo:
@rtitle, @jgainerdewar, @jdcanas and @lucymcnatt 

I tested the functionality with the dry-run on and off and was able to successfully upgrade WDS and Cromwell on AKS, and Cromwell and RStudio on GKE. 
I was not able to upgrade or install older versions of Hail Batch and Custom apps and so could not really test upgrading
I was able to confirm that the Galaxy helm chart is not compatible with an automatic helm upgrade, the team has been informed, and it is okay to exclude existing Galaxy versions from the upgrade for now.

This is ready for a final review and hopefully getting a 👍 so we can merge :)

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/adminRoutesModels.scala
```diff
@@ -0,0 +1,47 @@
...
+import org.broadinstitute.dsp.ChartVersion
+
+final case class UpdateAppsRequest(appType: AppType,
+                                   cloudProvider: CloudProvider,
+                                   appVersionsInclude: List[ChartVersion],
+                                   appVersionsExclude: List[ChartVersion],
+                                   googleProject: Option[GoogleProject],
+                                   workspaceId: Option[WorkspaceId],
+                                   appNames: List[AppName],
+                                   dryRun: Boolean
```
jdcanas, line 14: maybe a comment on the usage of this field?

http/src/main/resources/reference.conf
```diff
@@ -250,6 +250,38 @@ azure {
      ]
      enabled = true
      database-enabled = false
+     chart-versions-to-exclude-from-updates = [
```
jdcanas, line 254: where are these sources from?

http/src/main/resources/reference.conf
```diff
@@ -268,6 +300,20 @@ azure {
...
```
jdcanas, line 305: same thing as above, maybe some comment, maybe some comment inline about where this comes from?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -358,6 +359,243 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
...
+        )
+      )
+
+      // Grab records from the database
+      dbAppOpt <- KubernetesServiceDbQueries
+        .getActiveFullAppByName(CloudContext.Azure(params.cloudContext), params.appName)
+        .transaction
+      dbApp <- F.fromOption(
+        dbAppOpt,
+        AppNotFoundException(CloudContext.Azure(params.cloudContext),
```
jdcanas, line 380: Hmm... I'm not sure we want a fatal error in this case. Maybe just a log? I'm imagining a scenario where we submit the request and the app is deleted in the interim

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -358,6 +359,243 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
 
     } yield ()
 
+  override def updateAndPollApp(params: UpdateAKSAppParams)(implicit ev: Ask[F, AppContext]): F[Unit] = {
```
jdcanas, line 362: this method seems like the definition of unscalable... maybe  there's a better encapsulation scheme? I'm imagining something like a mini-interp/object for each specific app with a method to generate the `chartOverrideValues` for that specific app. That way, instead of a massive case match here, the code would read something like `buildChartOverridesGenerator(app.appType).getOverrides()`. This general pattern would apply to a good bit of different methods and clean the case matching out of the base interp logic. I'll leave it up to you whether you think it should be done in this PR..

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/auth/AllowlistAuthProvider.scala
```diff
@@ -159,4 +159,6 @@ class AllowlistAuthProvider(config: Config, saProvider: ServiceAccountProvider[I
       }
     }
   } yield filteredResources.toSet
+
+  override def isAdminUser(userInfo: UserInfo)(implicit ev: Ask[IO, TraceId]): IO[Boolean] = ???
```
jdcanas, line 163: Can this be `F.pure(true)`? Probably shouldn't commit placeholder code..

jdcanas (approved):
LGTM overall, few comments

http/src/main/resources/reference.conf
LizBaldo: It is the responsibility of each app developer to inform of non backward compatible versions in there. We are not super happy with this design and will likely refactor soon, but i will add a comment :)

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -3872,6 +3926,42 @@ components:
...
+          type: string
+          description: If provided, update apps in this workspace
+        appNames:
+          type: array
+          description: If provided, apps with these names will be updated
+          items:
+            type: string
+        dryRun:
+          type: boolean
+          description: If true, report on what apps would be updated but take no action. Defaults to false.
```
LizBaldo, line 3963: @jdcanas the dryRun option is explained here, is that enough or you would rather have it explained somewhere else in the code as well?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
LizBaldo: Once we are in the audit freezing window, I am totally down to refactor this

LizBaldo: We should also refactor the crateAndPollApp method the same way

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
LizBaldo: Oh I see, we would want to avoid a `DELETED` status to go back to `ERROR`

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/auth/AllowlistAuthProvider.scala
LizBaldo: @jdcanas I see a lot of the `???` pattern all over the leonardo Mock functions, is that expected?

jdcanas: Hmm... I guess it's fine in Mock functions, but that seems to indicate that this code is never actually tested since `isAdminUser` is never hit on the mock 🤔

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/adminRoutesModels.scala
LizBaldo: Done at the swagger schema doc level already :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
LizBaldo: I am following the exact same pattern as for createAndPollApp though, should we modify the behavior there as well?

http/src/main/resources/reference.conf
```diff
@@ -577,6 +627,25 @@ gke {
     minMemoryGb = 5
     minNumOfCpus = 3
     enabled = true
+    # App developers - Please keep the list of non-backward compatible versions in the list below
+    chartVersionsToExcludeFromUpdates = [
```
rtitle, line 631: Curious if there's a way to disable updates for a given app type? E.g. say we want to prevent people updating Galaxy for now..

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/KubernetesAppConfig.scala
```diff
@@ -34,6 +23,15 @@ sealed trait KubernetesAppConfig {
...
 
   def enabled: Boolean
+
+  // Each app can configure its own list of chart versions to NEVER update. To be used
+  // when changes aren't backward-compatible and we know an update would be destructive.
+  def chartVersionsToExcludeFromUpdates: List[ChartVersion]
+
+  // These are defined by each implementing class. Each config type
+  // corresponds to a specific app type and cloud provider.
+  def cloudProvider: CloudProvider
```
rtitle, line 33: I was wondering if app configs should be cloud specific… but I guess they are today, I guess due to differences in the charts/services.

http/src/main/resources/reference.conf
LizBaldo: There is no specific flag to disable, for now I simply added all versions to the exclude list 😬

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/KubernetesAppConfig.scala
LizBaldo: They are all cloud specific for now yes

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesServiceDbQueries.scala
```diff
@@ -60,8 +61,39 @@ object KubernetesServiceDbQueries {
...
+        .filter(_.cloudProvider === cloudProvider)
+        .filterOpt(googleProject) { case (clusterTable, gp) =>
+          clusterTable.cloudContextDb === CloudContextDb(gp.value)
+        },
+      nodepoolQuery,
+      appQuery
+        .filter(_.status === (AppStatus.Running: AppStatus))
+        .filter(_.appType === appType)
+        .filter(_.chart =!= targetVersion)
+        .filterIf(chartVersionsToInclude.nonEmpty)(_.chart inSetBind chartVersionsToInclude)
```
rtitle, line 88: Fancy slick stuff :) Looks good to me, may be a good candidate to unit test (given different input, does the query return the correct things..)

LizBaldo: Oh indeed, we tested all possible combination on the BEE using the dryRun mode, but I did not think of transposing this to unit tests. The tricky part is having the correct older versions of the apps installed, but that might be mockable

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -358,6 +359,243 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
...
+      }
+
+      // Get relay hybrid connection pool information
+      hcName = RelayHybridConnectionName(s"${params.appName.value}-${workspaceId.value}")
+      relayDomain = s"${landingZoneResources.relayNamespace.value}.servicebus.windows.net"
+      relayEndpoint = s"https://${relayDomain}/"
+      relayPath = Uri.unsafeFromString(relayEndpoint) / hcName.value
+
+      // Generate the app values to pass to helm at the upgrade chart step
+      chartOverrideValues <- app.appType match {
```
rtitle, line 422: For future ticket -- we should also update the relaylistener deployment, included in the setup chart.

rtitle (approved):
Looks good! Nice testing and demos. 🚢  :shipit: 🥳 🎉 🚀

http/src/main/resources/reference.conf
rtitle: Ok, just means we need to keep that version list up to date..

LizBaldo:
jenkins retest

LizBaldo:
Integration tests have been notoriously flaky and this error is completely unrelated to the changes:

`org.broadinstitute.dsde.workbench.leonardo.runtimes.RuntimeCreationDiskSpec.create runtime and attach an existing persistent disk or clone`

Force merging now

PR #3469, opened 2023-06-29
IA-4402 enableIntranodeVisibility for AoU apps

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-4402

With `enableIntraNodeVisibility`, we'll be able to see more inter node traffic logs, which will make it possible for us to trace back to a user if what they do triggers egress alert.

This PR turns on `enableIntraNodeVisibility` for GKE apps
1. For envs with `enable-custom-app-check` being false. This is mostly for dev envs.
2. For envs with `enable-custom-app-check` being true, it's only on for in AoU projects and AoU supported apps (RStudio, custom apps, cromwell).

We could potentially turn this flag on for all apps, but figured it's less risky to have it on only for AoU projects for now.

Tested things are working in BEE.
Screenshot 2023-06-29 at 1 25 53 PM

- [ ] I validated this change 

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/package.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/package.scala
@@ -17,6 +17,8 @@ package object leonardo {
   val SECURITY_GROUP = "security-group"
   val SECURITY_GROUP_HIGH = "high"
 
+  val AOU_UI_LABEL = "all-of-us"
+
   implicit def http4sBody[F[_], A](body: A)(implicit encoder: EntityEncoder[F, A]): Entity[F] =
     encoder.toEntity(body)
 
@@ -53,8 +55,6 @@ package object leonardo {
       retryOnPermissionDenied
     )
   }
-
-  def isAoUProject(labels: Map[String, String]): Boolean = labels.get(SECURITY_GROUP).isDefined
 
   val allSupportedRegions = List(
     RegionName("us-central1"),
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/package.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/package.scala
@@ -14,6 +14,9 @@ package object leonardo {
   val autoPauseOffValue = 0
   val traceIdHeaderString = ci"X-Cloud-Trace-Context"
 
+  val SECURITY_GROUP = "security-group"
+  val SECURITY_GROUP_HIGH = "high"
+
   implicit def http4sBody[F[_], A](body: A)(implicit encoder: EntityEncoder[F, A]): Entity[F] =
     encoder.toEntity(body)
 
@@ -50,6 +53,8 @@ package object leonardo {
       retryOnPermissionDenied
     )
   }
+
+  def isAoUProject(labels: Map[String, String]): Boolean = labels.get(SECURITY_GROUP).isDefined
 
   val allSupportedRegions = List(
     RegionName("us-central1"),
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
@@ -813,13 +813,11 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
   /**
    * Short-cuit if app creation isn't allowed, and return whether or not the user's project is AoU project
    *
-   * If custom app check isn't enabled, we'll just turn on intranode visibility; otherwise, we only turn it on for AoU projects
-   *
    * @param userEmail
    * @param googleProject
    * @param descriptorPath
    * @param ev
-   * @return true if intranode visibility should be enabled; false otherwise
+   * @return 
    */
   private[service] def checkIfAppCreationIsAllowed(userEmail: WorkbenchEmail,
                                                    googleProject: GoogleProject,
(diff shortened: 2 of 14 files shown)


core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/package.scala
```diff
@@ -51,6 +54,8 @@ package object leonardo {
     )
   }
 
+  def isAoUProject(labels: Map[String, String]): Boolean = labels.get(SECURITY_GROUP).isDefined
```
LizBaldo: Cool, I did not know that, so only AOU projects have a "security-group" label?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
```diff
@@ -83,11 +80,18 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
       )
       _ <- F.raiseWhen(!hasPermission)(ForbiddenError(userInfo.userEmail))
 
-      _ <- req.appType match {
+      enableIntraNodeVisibility <- req.appType match {
+        case AppType.Galaxy | AppType.HailBatch | AppType.Wds => F.pure(false)
```
LizBaldo: Up to you, but I would not necessarily worry about WDS or HailBatch since they are only implemented for Azure

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
```diff
@@ -823,39 +841,41 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
             projectLabels <- googleResourceService.getLabels(googleProject)
             isAppAllowed <- projectLabels match {
               case Some(labels) =>
+                // as long as the project has SECURITY_GROUP label, it is an AoU project
```
LizBaldo: 👍

LizBaldo: Not within the scope of this PR, but it would be nice to have a more explicit label to check if a project is an AOU one (similar to the aou label in the leo db for runtimes)

LizBaldo (approved):
Have not tested, but that makes sense to me 👍 

I just left a few very minor comments :) Thanks for jumping on this!

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
Qi77Qi: maybe..we can definitely add that if it becomes a pattern. Btw, where do you see the aou label for runtimes?

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/package.scala
Qi77Qi: yep...

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
LizBaldo: I am fairly sure that the label table in leo DB differenciates between the terra (`terraLabel`) and aou (`allOfUsLabel`) resources but that might be for the UI definition of a runtime, not the leo one 🤔

Qi77Qi: ah right...UI sets some labels to differentiate terra vs AoU as well...I guess that could be an alternative solution too...I'll look into that

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -121,6 +122,7 @@ class GKEInterpreter[F[_]](
       kubeNetwork = KubernetesNetwork(googleProject, network)
       kubeSubNetwork = KubernetesSubNetwork(googleProject, dbCluster.region, subnetwork)
 
+      networkConfig = new model.NetworkConfig().setEnableIntraNodeVisibility(params.enableIntraNodeVisibility)
```
yonghaoy, line 125: Change LG. But sometimes Google libraries can not handle "default" value well. Worth to check what other default fields for `NetworkConfig` if we did not set those. 
(I remembered it caused an issue before)

yonghaoy (approved):

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
Qi77Qi: @yonghaoy I compared cluster before and after the change, and the `Networking` section looks the same to me in console, so I think we can give it a try

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
Qi77Qi: @LizBaldo updated things to use label to differentiate..lmk if u have more comments

LizBaldo: LGTM, maybe you want to update the comment above to reflect the use of the UI label?

Qi77Qi: I removed the comment for now...I think in reality, the statement is true today, but the comment is no longer super related to the code anymore

Qi77Qi:
```
terra-quality-8772546f/automation-test-awpboqkgz errored due to List(RuntimeError(Jupyter, Welder failed to start after 10 minutes.,None,2023-06-29T20:59:22Z,Some(TraceId(4dd9c16abc244a5a925808de70dd6e35/3939911508519804487))))
```

jenkins retest 

if the next run fails again on GCE VM related test, think I'm going to just merge it since the change has nothing to do with GCE VM

PR #3484, opened 2023-07-07
[IA-4418] Allow creator of a runtime to view it, if they can view the workspace

mlilynolting:
__Jira ticket__: https://broadworkbench.atlassian.net/browse/IA-4415

## Summary

- Permit runtime creator to always perform reads on the runtime.
- Sam permission checks will fail if the Azure runtime errors during create. Terra-workspace-manager will clean up Sam's policy records following the Azure VM failing and deleting itself. A Leo record, thus orphaned, should still be readable by the creator, who may want to check its errors for an explanation.
- This fix allows users with Workspace Reader to access runtimes they created. This behavior is consistent with listRuntimesV2, which shows a user with Workspace Reader all runtimes in a workspace (creatorOnly off) or all runtimes they created in a workspace (creatorOnly on). The list endpoint does not authorize user access to any individual runtime via Sam, only the workspace, so it avoids the issue fixed here.

### What to test

- Create a runtime under conditions which will cause an error in Azure (you could create it in a billing project with quota limits or availability limits for your region, or you could manually interrupt the creation from the Azure Portal side so that terra-workspace-manager decides create has failed)
- Open the Cloud Environment modal for your failed runtime
- You should be able to view a meaningful error message whose ultimate source was Azure.

### Who tested and where

- [x] I validated this change 


LizBaldo (commented):
This looks like a good change, but it might lead to inconsistencies and confuse the user if it does not extend to the following endpoint as well:
- List Runtime V2 
- List Runtime V1
- Get Runtime V1
- List App (V1 and V2)
- Get App (V1 and V2)

mlilynolting:
> This looks like a good change, but it might lead to inconsistencies and confuse the user if it does not extend to the following endpoint as well:
> 
>     * List Runtime V2
> 
>     * List Runtime V1
> 
>     * Get Runtime V1
> 
>     * List App (V1 and V2)
> 
>     * Get App (V1 and V2)

@LizBaldo True! I'm going to put a new card into the sprint to add those accesses without blocking this bugfix: https://broadworkbench.atlassian.net/browse/IA-4431

I don't know if v1 endpoints should be covered here.

mlilynolting:
jenkins retest

mlilynolting:
jenkins retest

mlilynolting:
jenkins retest

mlilynolting:
Another swat run is ongoing but not picked up by the actions panel above. https://fc-jenkins.dsp-techops.broadinstitute.org/job/leonardo-PR-fiab-test/12652/

LizBaldo (approved):
👍 from me as long as the List Runtime V2 endpoint is consistent (which might already be the case, but if you could confirm that would be great)

mlilynolting:
Passed swat https://fc-jenkins.dsp-techops.broadinstitute.org/job/leonardo-fiab-test-runner/56162/

PR #3501, opened 2023-07-18
WX-1194 Upgrade apps endpoint

jgainerdewar:
Still working on testing, putting up PR for early code review.

__Jira ticket__: https://broadworkbench.atlassian.net/browse/WX-1194

### What

-

### Why

-

### What to test

- 

### Who tested and where

- [ ] I validated this change 


jgainerdewar:
This doesn't seem to be running tests, do I need different permissions on the repo?

LizBaldo:
@jgainerdewar doing a first pass at the review now, since the branch is pointing to the other upgrade feature branch and not develop, tests are not automatically running, I am happy to merge my leo PR so you can point to develop instead if that is easier, lmk!

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/adminRoutesModels.scala
```diff
@@ -0,0 +1,47 @@
...
+                                   cloudProvider: CloudProvider,
+                                   appVersionsInclude: List[ChartVersion],
+                                   appVersionsExclude: List[ChartVersion],
+                                   googleProject: Option[GoogleProject],
+                                   workspaceId: Option[WorkspaceId],
+                                   appNames: List[AppName],
+                                   dryRun: Boolean
+)
+
+final case class ListUpdateableAppsResponse(workspaceId: Option[WorkspaceId],
```
LizBaldo: Nit, but this class is only describing one app correct? If so I would rename to `UpdatableAppResonse`

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/adminRoutesModels.scala
```diff
@@ -0,0 +1,47 @@
...
+                                            appName: AppName,
+                                            appType: AppType,
+                                            auditInfo: AuditInfo,
+                                            chart: Chart,
+                                            accessScope: Option[AppAccessScope],
+                                            labels: LabelMap
+)
+
+object ListUpdateableAppsResponse {
+  def fromClusters(clusters: List[KubernetesCluster]): List[ListUpdateableAppsResponse] = for {
```
LizBaldo: I am still ramping up with scala so I might be completely off, but my understanding is that we want to traverse through the list of clusters and generate the corresponding list of UpdatableAppResponse for each app within each nodepool of a cluster?

I am not sure the code below would do that? If it does, then I might have made my life a lot more complicated using scala 😅

http/src/main/resources/reference.conf
```diff
@@ -576,6 +579,7 @@ gke {
     minMemoryGb = 5
     minNumOfCpus = 3
     enabled = true
+    chartVersionsToExcludeFromUpdates = []
```
LizBaldo, line 583: Arg, I completely understand why you used this different format, but I never noticed the difference before and it is driving me a bit crazy 😬

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -1878,6 +1880,37 @@ paths:
               schema:
                 $ref: "#/components/schemas/ErrorReport"
 
+  ## Admin API ##
+
+  "/api/admin/v2/apps/update":
+    post:
+      summary: Updates apps
```
LizBaldo: Nit: Updates k8s apps

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -1878,6 +1880,37 @@ paths:
...
+      description: Updates a designated collection of apps (both v1 and v2)
+      operationId: updateApps
+      tags:
+        - admin
+      requestBody:
+        $ref: "#/components/requestBodies/UpdateAppsRequest"
+      responses:
```
LizBaldo, line 1894: Would we need a 404 or a permission denied response as well?

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -3872,6 +3914,42 @@ components:
...
+        - cloudProvider
+      type: object
+      properties:
+        appType:
+          $ref: '#/components/schemas/AppType'
+        cloudProvider:
+          $ref: '#/components/schemas/CloudProvider'
+        appVersionsInclude:
+          type: array
+          description: Apps currently running these chart versions will be updated. Applied after a config-level default exclusion list.
```
LizBaldo: Nit, but add `If provided` at the beginning for this one and the one below to match the rest of the optional properties?

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -3958,6 +4036,37 @@ components:
           type: object
           description: The labels of each app in the response whose key is in includeLabels in the request. Of type Map[String,String]
 
+    ListUpdateableAppsResponse:
+      description: out-of-date apps
```
LizBaldo: Nit: List of out-of-date apps properties?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/SamAuthProvider.scala
```diff
@@ -312,6 +312,9 @@ class SamAuthProvider[F[_]: OpenTelemetryMetrics](
   override def isCustomAppAllowed(userEmail: WorkbenchEmail)(implicit ev: Ask[F, TraceId]): F[Boolean] =
     samDao.isGroupMembersOrAdmin(config.customAppCreationAllowedGroup, userEmail)
 
+  override def isAdminUser(userInfo: UserInfo)(implicit ev: Ask[F, TraceId]): F[Boolean] =
```
LizBaldo, line 315: Thank you ❤️

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/KubernetesAppConfig.scala
```diff
@@ -34,6 +23,15 @@ sealed trait KubernetesAppConfig {
   def kubernetesServices: List[KubernetesService]
 
   def enabled: Boolean
+
+  // Each app can configure its own list of chart versions to NEVER update. To be used
```
LizBaldo, line 27: I am thinking that we should probably advertise this to the Galaxy, AOU, and WDS folks as well. It should likely go into their `contributing` guidelines. I can take an action item to gather this requirement from them 👍

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpSamDAO.scala
```diff
@@ -506,6 +506,33 @@ class HttpSamDAO[F[_]](httpClient: Client[F],
           } yield admins.contains(workbenchEmail)
     } yield res
 
+  override def isAdminUser(userInfo: UserInfo)(implicit
```
LizBaldo, line 519: FYI, we have it in our backlog to tear this whole file apart and use the sam client directly, sorry you had to do this 😬

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesServiceDbQueries.scala
```diff
@@ -59,6 +60,36 @@ object KubernetesServiceDbQueries {
       labelFilter
     )
 
+  /**
+    * List all RUNNING apps that are not on the given chart version. Used to decide which apps to update.
```
LizBaldo: Nit: given -> current?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/ConfigReader.scala
```diff
@@ -22,6 +16,18 @@ object ConfigReader {
     ConfigSource
       .fromConfig(org.broadinstitute.dsde.workbench.leonardo.config.Config.config)
       .loadOrThrow[AppConfig]
+
+  // Encapsulates config for all apps Leo administers
```
LizBaldo, line 20: Nice! Gosh at some point we really need to clean up the GKE vs AKS apps divide

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AdminServiceInterp.scala
```diff
@@ -0,0 +1,89 @@
...
+        case Some(conf) => F.pure(conf)
+        case None => F.raiseError(NoMatchingAppError(req.appType, req.cloudProvider, Option(ctx.traceId)))
+      }
+
+      // Query the database for the collection of updateable apps matching filters
+      excludedVersions = (appConfig.chartVersionsToExcludeFromUpdates ++ req.appVersionsExclude).distinct
+      matchingApps <- KubernetesServiceDbQueries.listAppsForUpdate(appConfig.chart,
+                                                                   req.appType,
+                                                                   req.cloudProvider,
+                                                                   req.appVersionsInclude.map(Chart(appConfig.chartName, _)),
```
LizBaldo, line 55: We want to filter by chart version, not by chart name correct?

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AdminServiceInterpSpec.scala
```diff
@@ -0,0 +1,138 @@
...
+      config
+    )
+
+    val res = interp.updateApps(userInfo, updateAppsRequest.copy(dryRun = false))
+      .unsafeRunSync()(cats.effect.unsafe.IORuntime.global)
+
+    res.length shouldEqual 1
+    res.map(_.appName.value).contains("app2") shouldBe true
+
+    // Verify that no update message was send
```
LizBaldo: Nit: Verify that the update message was send?

LizBaldo (commented):
Looks really clear and thorough to me, I can't thank you enough for doing this ❤️ 

I left a few minor comments and I have a question around how `ListUpdateableAppsResponse` is constructed, but other than that it looks good to me.

I think I am going to go ahead and maybe merge the [IA-4302-upgradeApp-pubsub-message](https://github.com/DataBiosphere/leonardo/tree/IA-4302-upgradeApp-pubsub-message) PR so you can rebase to develop and properly run the tests

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/adminRoutesModels.scala
jgainerdewar: Believe it or not this code does do that! Each line of the `for` block will expand the lines below so they run once for each item in its collection, like:
```
val i = List(List(1, 2, 3), List(4, 5, 6))
for {
  a <- i
  b <- a
} yield b

>>> List(1, 2, 3, 4, 5, 6)
```

LizBaldo: Ohh, this is making me hate scala a lot less, TIL, thank you ❤️

http/src/main/resources/reference.conf
jgainerdewar: Yeah same 😬 I opted to stay locally consistent and not overhaul the key format of the entire file but this makes me itch.

http/src/main/resources/swagger/api-docs.yaml
jgainerdewar: That's a good point, I'm not actually sure we're returning the right error HTTP codes under the relevant conditions, will address.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesServiceDbQueries.scala
jgainerdewar: I actually think I want to rename `currentChart` to `updateTargetChart` or similar, I don't think this function cares about new vs old or current - it just cares about "the chart to update to."

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/ConfigReader.scala
jgainerdewar: 💯 💯

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AdminServiceInterp.scala
jgainerdewar: Yeah, but it looks like the database value is the full `${chartName}-${chartVersion}` string, and the chart name will be the same for all instances of a given (app type, cloud) so it seemed better to do the database query based on `Chart` rather than `ChartVersion`. I might be missing an edge case where this doesn't make sense, though.

http/src/main/resources/reference.conf
jgainerdewar: Suggestions for alternate names for this config setting welcome, this one is unwieldy but I wanted the intention to be really clear.

LizBaldo:
@jgainerdewar FYI, I resolved some merge conflicts in https://github.com/DataBiosphere/leonardo/pull/3443, I am uneasy merging it to dev, but if you wanted to merge your PR in right now and continue developing so your branch points to develop lmk

http/src/main/resources/swagger/api-docs.yaml
jgainerdewar: (Reminded myself that we ARE returning the right codes, updating Swagger doc)

jgainerdewar:
I think that probably makes sense... we want to test the changes together anyway, and I would like to get tests running on this.

LizBaldo (approved):

PR #3504, opened 2023-07-20
TOAZ-306 delete errored app

dvoet:
__Jira ticket__: https://broadworkbench.atlassian.net/browse/TOAZ-306

## Summary of changes

delete any resources that maybe have created for an errored app

### What

- remove special case for errored apps
- make the deleteApp tolerant of missing resources
- stop uninstalling the helm chart because deleting the namespace does all we need

### Why

- errored apps that are not cleaned up appropriately cost money and can prevent workspace deletion

### What to test

- 

### Who tested and where

- [ ] I validated this change 


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -386,33 +393,15 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
           RelayHybridConnectionName(name),
           cloudContext
         )
-
-      // Authenticate helm client
```
LizBaldo, line 390: I am understanding correctly that we don't need this because we have already deleted the underlying K8s resources anyways?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -386,33 +393,15 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
...
-        .run(authContext)
+        .handleErrorWith {
+          case e: ManagementException if e.getResponse.getStatusCode == StatusCodes.NotFound.intValue =>
+            logger.info(s"${name} does not exist to delete in ${cloudContext}")
+          case e => F.raiseError[Unit](e)
+        }
 
       client <- kubeAlg.createAzureClient(cloudContext, landingZoneResources.clusterName)
 
-      // Poll until all pods in the app namespace are deleted
```
LizBaldo, line 404: Would it make sense to keep polling, or is this done by WSM under the hood?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -371,11 +374,15 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
       _ <- logger.info(ctx.loggingCtx)(s"Deleting app $appName in workspace $workspaceId")
 
       app = dbApp.app
+
+      // Delete WSM resources associated with the app
```
LizBaldo, line 378: How can we ensure that the corresponding Sam resource is also deleted? Is this handled by WSM?

LizBaldo (approved):
Thanks for looking into this!

I think this makes sense to my, I get why the helm uninstall is redundant. My only question is around making sure that the Sam Resource gets deleted as well. 

But if things are looking good on your BEE then am okay :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
dvoet: correct

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
dvoet: there is polling below for the namespace being gone

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
dvoet: WSM handles that

dvoet: Or maybe another way to think about it: these are WSM resources, the fact that there are Sam resources behind them is a WSM implementation detail, Leo should not care.

dvoet:
jenkins retest

dvoet:
jenkins retest

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
rtitle: I'm a little uneasy with skipping the `helm uninstall` -- will namespace deletion work reliably when there are active helm deployments? Will it leak helm state somehow (e.g. will the deployment still show up in `helm list`)?

dvoet:
between https://fc-jenkins.dsp-techops.broadinstitute.org/job/leonardo-fiab-test-runner/56340/ and https://fc-jenkins.dsp-techops.broadinstitute.org/job/leonardo-fiab-test-runner/56345/ there is full coverage in automation tests

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
dvoet: I have confirmed that `helm list` does not indicate any leakage

PR #3509, opened 2023-07-21
RW-9838 Refactor existing RStudio app support to be more generic

Qi77Qi:
https://precisionmedicineinitiative.atlassian.net/browse/RW-9838

https://github.com/all-of-us/workbench/pull/7851 needs to be merged shortly after this

[This doc](https://docs.google.com/document/d/1RaQRMqAx7ymoygP6f7QVdBbZC-iD9oY_XLNMe_oz_cs/edit) has more context for this PR

1. Update to createApp API should be backwards compatible createApp API for the most part. One thing that's different is that the proxy path for RStudio will be `/app` instead of `/rstudio`
2. Change to createApp api includes adding `ALLOWED` app type and `allowedChartName` as described in [the doc](https://docs.google.com/document/d/1RaQRMqAx7ymoygP6f7QVdBbZC-iD9oY_XLNMe_oz_cs/edit)

TODOs:
- [x] Update swagger yaml
- [x] Test start/stop
- [x] Check metrics categorized by `AppType`
- [x] Test App upgrade

**Tests**
1. Tested in BEE with this req:
```
POST https://{{leoUrl}}/api/google/v1/apps/{{project}}/{{appName}}
accept: */*
Authorization: Bearer {{token}}
Content-Type: application/json

{
  "diskConfig": {
    "name": "{{appName}}"
  },
  "appType": "RSTUDIO",
  "labels": {"all-of-us": "true"}
}
```

Screenshot 2023-07-25 at 1 47 43 PM

2. Tested in BEE with new contract:
```
{
  "diskConfig": {
    "name": "qi-rstudio-1-new"
  },
  "appType": "ALLOWED",
  "allowedChartName": "aou-rstudio-chart",
    "WORKSPACE_NAME": "qi-ws"
  }
}
```

3. Tested start/stop/delete
4. Tested reusing the same disk with a second app
5. Tested creating multiple apps as same user on the same cluster, and found a bug as mentioned in [here](https://broadinstitute.slack.com/archives/C05HS98C0V8/p1690570413087969)
6. Tested ALLOWED app upgrade path
```
2023/08/07 18:12:57 preparing upgrade for xyrvde-rstudio-rls
2023/08/07 18:12:57 resetting values to the chart's original version
2023/08/07 18:13:02 performing update for xyrvde-rstudio-rls
2023/08/07 18:13:02 creating upgraded release for xyrvde-rstudio-rls
2023/08/07 18:13:02 checking 6 resources for changes
2023/08/07 18:13:02 Looks like there are no changes for PersistentVolume "xyrvde-rstudio-rls-aou-rstudio-chart-pv"
2023/08/07 18:13:02 Looks like there are no changes for PersistentVolumeClaim "xyrvde-rstudio-rls-aou-rstudio-chart-pvc"
2023/08/07 18:13:02 Looks like there are no changes for Service "xyrvde-rstudio-rls-aou-rstudio-chart"
2023/08/07 18:13:02 Looks like there are no changes for Service "xyrvde-rstudio-rls-aou-rstudio-chart-welder"
2023/08/07 18:13:02 Patch Deployment "xyrvde-rstudio-rls-aou-rstudio-chart" in namespace xyrvde-rstudio-ns
2023/08/07 18:13:02 Looks like there are no changes for Ingress "xyrvde-rstudio-rls-aou-rstudio-chart"
2023/08/07 18:13:02 updating status for upgraded release for xyrvde-rstudio-rls
2023/08/07 18:13:02 Finished upgrading to release xyrvde-rstudio-rls
```

- [ ] I validated this change 

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/appRoutesModels.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/appRoutesModels.scala
@@ -3,6 +3,7 @@ package org.broadinstitute.dsde.workbench.leonardo.http
 import org.broadinstitute.dsde.workbench.google2.DiskName
 import org.broadinstitute.dsde.workbench.google2.KubernetesSerializableName.ServiceName
 import org.broadinstitute.dsde.workbench.leonardo.{
+  AllowedChartName,
   App,
   AppAccessScope,
   AppError,
@@ -17,12 +18,14 @@ import org.broadinstitute.dsde.workbench.leonardo.{
   Nodepool,
   WorkspaceId
 }
+import org.broadinstitute.dsp.ChartName
 import org.http4s.Uri
 
 import java.net.URL
 
 final case class CreateAppRequest(kubernetesRuntimeConfig: Option[KubernetesRuntimeConfig],
                                   appType: AppType,
+                                  allowedChartName: Option[AllowedChartName],
                                   accessScope: Option[AppAccessScope],
                                   diskConfig: Option[PersistentDiskRequest],
                                   labels: LabelMap = Map.empty,
@@ -42,6 +45,7 @@ final case class GetAppResponse(appName: AppName,
                                 customEnvironmentVariables: Map[String, String],
                                 auditInfo: AuditInfo,
                                 appType: AppType,
+                                chartName: ChartName,
                                 accessScope: Option[AppAccessScope],
                                 labels: LabelMap
 )
@@ -54,6 +58,7 @@ final case class ListAppResponse(workspaceId: Option[WorkspaceId],
                                  proxyUrls: Map[ServiceName, URL],
                                  appName: AppName,
                                  appType: AppType,
+                                 chartName: ChartName,
                                  diskName: Option[DiskName],
                                  auditInfo: AuditInfo,
                                  accessScope: Option[AppAccessScope],
@@ -79,6 +84,7 @@ object ListAppResponse {
           a.getProxyUrls(c, proxyUrlBase),
           a.appName,
           a.appType,
+          a.chart.name,
           a.appResources.disk.map(_.name),
           a.auditInfo,
           a.appAccessScope,
@@ -105,6 +111,7 @@ object GetAppResponse {
       appResult.app.customEnvironmentVariables,
       appResult.app.auditInfo,
       appResult.app.appType,
+      appResult.app.chart.name,
       appResult.app.appAccessScope,
       appResult.app.labels
     )
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
@@ -329,8 +329,7 @@ object AllowedChartName {
   def fromChartName(chartName: ChartName): Option[AllowedChartName] = {
     val splittedString = chartName.asString.split("/")
     if (splittedString.size == 3) {
-      val splitChartNameAndVersion = splittedString(2).split("-")
-      stringToObject.get(splitChartNameAndVersion.dropRight(1).mkString("-"))
+      stringToObject.get(splittedString(2))
     } else None
   }
 }
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
@@ -324,6 +324,15 @@ object AllowedChartName {
     def asString: String = "aou-sas-chart"
   }
   def stringToObject: Map[String, AllowedChartName] = sealerate.values[AllowedChartName].map(v => v.asString -> v).toMap
+
+  // Chartname from DB has the following format: /leonardo/cromwell-0.2.291
+  def fromChartName(chartName: ChartName): Option[AllowedChartName] = {
+    val splittedString = chartName.asString.split("/")
+    if (splittedString.size == 3) {
+      val splitChartNameAndVersion = splittedString(2).split("-")
+      stringToObject.get(splitChartNameAndVersion.dropRight(1).mkString("-"))
+    } else None
+  }
 }
 
 sealed abstract class AppType
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
@@ -313,6 +313,19 @@ object ErrorSource {
   def stringToObject: Map[String, ErrorSource] = values.map(v => v.toString -> v).toMap
 }
 
+sealed abstract class AllowedChartName extends Product with Serializable {
+  def asString: String
+}
+object AllowedChartName {
+  final case object RStudio extends AllowedChartName {
+    def asString: String = "aou-rstudio-chart"
+  }
+  final case object Sas extends AllowedChartName {
+    def asString: String = "aou-sas-chart"
+  }
+  def stringToObject: Map[String, AllowedChartName] = sealerate.values[AllowedChartName].map(v => v.asString -> v).toMap
+}
+
 sealed abstract class AppType
 object AppType {
   case object Galaxy extends AppType {
@@ -330,8 +343,9 @@ object AppType {
     override def toString: String = "HAIL_BATCH"
   }
 
-  case object RStudio extends AppType {
-    override def toString: String = "RSTUDIO"
+  // See more context in https://docs.google.com/document/d/1RaQRMqAx7ymoygP6f7QVdBbZC-iD9oY_XLNMe_oz_cs/edit
+  case object Allowed extends AppType {
+    override def toString: String = "ALLOWED"
   }
 
   case object Custom extends AppType {
@@ -350,7 +364,7 @@ object AppType {
     appType match {
       case Galaxy                     => FormattedBy.Galaxy
       case Custom                     => FormattedBy.Custom
-      case RStudio                    => FormattedBy.RStudio
+      case Allowed                    => FormattedBy.Allowed
       case Cromwell | Wds | HailBatch => FormattedBy.Cromwell
     }
 }
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -678,13 +678,17 @@ gke {
     # App developers - Please keep the list of non-backward compatible versions in the list below
     chartVersionsToExcludeFromUpdates = []
   }
-  rstudioApp {
+  allowedApp {
     # If you update this here, be sure to update in the dockerfile
-    chartName = "/leonardo/aou-rstudio-chart"
-    chartVersion = "0.2.0"
+    # This is really just a prefix of the chart name. Full chart name for AOU app will be
+    # this chartName in config file appended with allowedChartName from createAppRequest
+    chartName = "/leonardo/"
+    rstudioChartVersion = "0.2.0"
+    sasChartVersion = "0.1.0"
+
     services = [
           {
-            name = "rstudio-service"
+            name = "app"
             kind = "ClusterIP"
           },
           {
@@ -699,6 +703,7 @@ gke {
     # App developers - Please keep the list of non-backward compatible versions in the list below
     chartVersionsToExcludeFromUpdates = []
   }
+
   customApp {
     chartName = "/leonardo/terra-app"
     # If you update this here, be sure to update in the dockerfile
--- a/http/src/main/resources/swagger/api-docs.yaml
+++ b/http/src/main/resources/swagger/api-docs.yaml
@@ -15,7 +15,7 @@ tags:
   - name: disks
     description: Persistent Disks API.
   - name: apps
-    description: Experimental App API, Support for Google Kubernetes Engine and Azure Kubernetes Service. Under active Development.
+    description: Apps API. Support Google Kubernetes Engine and Azure Kubernetes Service.
   - name: proxy
     description: Proxy API
   - name: admin
--- a/http/src/main/resources/swagger/api-docs.yaml
+++ b/http/src/main/resources/swagger/api-docs.yaml
@@ -15,9 +15,7 @@ tags:
   - name: disks
     description: Persistent Disks API.
   - name: apps
-    description: Experimental App API
-  - name: appsV2
-    description: Apps API v2. Support for Google Kubernetes Engine and Azure Kubernetes Service. Under active Development.
+    description: Experimental App API, Support for Google Kubernetes Engine and Azure Kubernetes Service. Under active Development.
   - name: proxy
     description: Proxy API
   - name: admin
@@ -1663,7 +1661,7 @@ paths:
         Default labels appName, googleProject, serviceAccount, and creator cannot be overridden.
       operationId: createAppV2
       tags:
-        - appsV2
+        - apps
       parameters:
         - in: path
           name: workspaceId
@@ -1709,7 +1707,7 @@ paths:
         Poll this to find out when your app has finished starting up.
       operationId: getAppV2
       tags:
-        - appsV2
+        - apps
       parameters:
         - in: path
           name: workspaceId
@@ -1750,7 +1748,7 @@ paths:
       description: deletes an App
       operationId: deleteAppV2
       tags:
-        - appsV2
+        - apps
       parameters:
         - in: path
           name: workspaceId
@@ -1795,7 +1793,7 @@ paths:
       description: List all apps for a given workspaceId.
       operationId: listAppsV2
       tags:
-        - appsV2
+        - apps
       parameters:
         - in: path
           name: workspaceId
@@ -1847,7 +1845,7 @@ paths:
       description: deletes all Apps in a workspace
       operationId: deleteAllAppsV2
       tags:
-        - appsV2
+        - apps
       parameters:
         - in: path
           name: workspaceId
@@ -2866,6 +2864,10 @@ components:
         - WDS
         - HAIL_BATCH
         - RSTUDIO
+    AllowedChartName:
+      type: string
+      enum:
+        - aou-rstudio-chart
     AppStatus:
       type: string
       enum:
@@ -3938,6 +3940,8 @@ components:
           $ref: '#/components/schemas/PersistentDiskRequest'
         appType:
           $ref: '#/components/schemas/AppType'
+        allowedChartName:
+          $ref: '#/components/schemas/AllowedChartName'
         accessScope:
           $ref: '#/components/schemas/AppAccessScope'
         kubernetesRuntimeConfig:
--- a/http/src/main/resources/swagger/api-docs.yaml
+++ b/http/src/main/resources/swagger/api-docs.yaml
@@ -2863,7 +2863,7 @@ components:
         - GALAXY
         - WDS
         - HAIL_BATCH
-        - RSTUDIO
+        - ALLOWED
     AllowedChartName:
       type: string
       enum:
@@ -3227,7 +3227,7 @@ components:
             - GCE
             - CROMWELL
             - CUSTOM
-            - RSTUDIO
+            - ALLOWED
           description: App that formatted disk, missing if unformatted
 
     GetPersistentDiskV2Response:
@@ -3288,7 +3288,7 @@ components:
             - GCE
             - CROMWELL
             - CUSTOM
-            - RSTUDIO
+            - ALLOWED
           description: App that formatted disk, missing if unformatted
 
     InstanceKey:
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/Config.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/Config.scala
@@ -690,8 +690,8 @@ object Config {
     )
   }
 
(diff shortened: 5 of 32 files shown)


http/src/main/resources/reference.conf
```diff
@@ -603,13 +603,17 @@ gke {
...
+    # This is really just a prefix of the chart name. Full chart name for AOU app will be
+    # this chartName in config file appended with allowedChartName from createAppRequest
+    chartName = "/leonardo/"
+    rstudioChartVersion = "0.2.0"
+    sasChartVersion = "0.1.0"
+
     services = [
           {
-            name = "rstudio-service"
+            name = "app"
```
LizBaldo, line 691: `aou-app`?

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
```diff
@@ -309,6 +309,19 @@ object ErrorSource {
   def stringToObject: Map[String, ErrorSource] = values.map(v => v.toString -> v).toMap
 }
 
+sealed abstract class AllowedChartName extends Product with Serializable {
```
LizBaldo, line 316: I understand the bigger picture, but I think it would be more prudent to isolate this to AOU apps for now, and rename to `AOUAllowed` for the chart names and AppType.
Also the code below is always referencing `aou*` so it will be easier to be consistent throughout

That will also help us make the AOU app upgrade logic explicit (we will need to modify our code to loop through each chart within the app type for AOU moving forward)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/AppV2Routes.scala
```diff
@@ -168,16 +169,33 @@ object AppV2Routes {
...
         d <- x.downField("diskConfig").as[Option[PersistentDiskRequest]]
         l <- x.downField("labels").as[Option[LabelMap]]
         cv <- x.downField("customEnvironmentVariables").as[Option[LabelMap]]
         dp <- x.downField("descriptorPath").as[Option[Uri]]
         ea <- x.downField("extraArgs").as[Option[List[String]]]
         swi <- x.downField("sourceWorkspaceId").as[Option[WorkspaceId]]
+
+        optStr <- x.downField("appType").as[Option[String]]
+        cn <- x.downField("allowedChartName").as[Option[AllowedChartName]]
+        // TODO: once AOU has migrated to use the new app type, we can use much simpler version instead of this workaround for backwards compatibility
```
LizBaldo, line 182: 👍

LizBaldo (commented):
I really like being able to bundle up all AOU apps into one type to streamline the code. My main comment is around the naming convention, and to make it explicit for Devs that these are AOU apps only for now, so renaming from `Allowed` to `AouAllowed` would help a lot in that regard.

@jgainerdewar FYI this will break our AppType to app mapping assumption in the upgrade logic though so we will have to reconcile and will likely need to have an optional chartName argument in the upgrade app message so we know which of the aou app to update 🤔 

@Qi77Qi I did not check in details, but would this refactoring also impact the stop/start/delete endpoints as well?

http/src/main/resources/reference.conf
Qi77Qi: no.. I want this to be generic...it doesn't really need to be "app" specific

rtitle:
So if I understand correctly, `ALLOWED` is kind of a meta-appType which bundles together different potential charts. This should make it easier to integrate new charts, because we won't need to introduce a new AppType and add boilerplate code each time. `ALLOWED` is similar to `CUSTOM` in this way -- a difference is that `CUSTOM` templates things into a common helm chart; whereas `ALLOWED` supports many different helm charts.

I think this makes sense at a high level. Things I'm thinking about:

1. We're still going to need to map helm values to individual charts. I think that's a source of a lot of pain today. I would also love to introduce a standard set of Leo values that consistent across charts. I guess that's kind of orthogonal to this PR.
2. Should we think a bit more about categories? Is allowed-for-AoU, allowed-for-Terra, allowed-for-ToA the same thing or should they be tracked separately?
3. Wondering if there's anything today that depends on the current `AppType` enumeration which would break. Upgrades, metrics, ?

Those are my high level thoughts, will check out code too…

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/appRoutesModels.scala
```diff
@@ -23,6 +24,7 @@ import java.net.URL
 
 final case class CreateAppRequest(kubernetesRuntimeConfig: Option[KubernetesRuntimeConfig],
                                   appType: AppType,
+                                  allowedChartName: Option[AllowedChartName],
```
rtitle, line 28: So a caller would pass this to create an app: 
```
POST /api/apps/...
{
   …,
   "appType": ALLOWED",
   "allowedChartName": "sas"
}
```
right?

Is the `allowedChartName` stored in Leo and return in the `getApp`/`listApp` responses? I imagine the UI would need to know the chart name in order to display the right icon, etc.

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
rtitle: I guess a question is whether we envision the set of AoU-allowed apps being different than Terra-allowed apps (or even Terra-on-Azure-allowed-apps)?

These app types are really categories/bundles of potential apps. Code that deals with app types today (UI, metrics, upgrades, etc) will need to have knowledge about these special meta-appTypes.

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
```diff
@@ -346,7 +360,7 @@ object AppType {
     appType match {
       case Galaxy                     => FormattedBy.Galaxy
       case Custom                     => FormattedBy.Custom
-      case RStudio                    => FormattedBy.RStudio
+      case Allowed                    => FormattedBy.RStudio
```
rtitle: Wonder what to do about disk support. Do we want a disk formatted by Allowed-rstudio to be able to be used by Allowed-sas?

http/src/main/resources/reference.conf
```diff
@@ -603,13 +603,17 @@ gke {
     dbPassword = "replace-me"
     enabled = true
   }
-  rstudioApp {
+  aouApp {
```
rtitle: Would be good to keep the terminology consistent. The app type is `ALLOWED`, so maybe this should be `allowedApp { … }`?

http/src/main/resources/reference.conf
```diff
@@ -603,13 +603,17 @@ gke {
...
     # If you update this here, be sure to update in the dockerfile
-    chartName = "/leonardo/aou-rstudio-chart"
-    chartVersion = "0.1.0"
+    # This is really just a prefix of the chart name. Full chart name for AOU app will be
+    # this chartName in config file appended with allowedChartName from createAppRequest
```
rtitle, line 684: The caller of Leo then needs to know the exact chart name? 

Think about injection attacks, if we're taking user input and appending to file system paths.

(Thinking what if I passed `"allowedChartName": "aou-sas; rm -rf /"` heh)

http/src/main/resources/reference.conf
```diff
@@ -603,13 +603,17 @@ gke {
...
+    rstudioChartVersion = "0.2.0"
+    sasChartVersion = "0.1.0"
+
     services = [
           {
-            name = "rstudio-service"
+            name = "app"
             kind = "ClusterIP"
           },
           {
```
rtitle, line 694: This is kind of interesting, we're assuming welder comes with all "allowed" charts. Which is true for rstudio/sas, but not everything needs welder (Cromwell for example).

Are we thinking of `ALLOWED` apps as welder-enabled apps?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
```diff
@@ -82,7 +84,19 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
...
       enableIntraNodeVisibility = req.labels.get(AOU_UI_LABEL).isDefined
       _ <- req.appType match {
-        case AppType.Galaxy | AppType.HailBatch | AppType.Wds | AppType.RStudio | AppType.Cromwell => F.unit
+        case AppType.Galaxy | AppType.HailBatch | AppType.Wds | AppType.Cromwell => F.unit
+        case AppType.Allowed =>
+          req.allowedChartName match {
+            case Some(cn) =>
+              cn match {
+                case AllowedChartName.RStudio => F.unit
+                case AllowedChartName.Sas     => F.unit // TODO: check group permission check here
```
rtitle, line 93: Does one need to be in a special group in order to launch SAS?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
```diff
@@ -1133,18 +1149,32 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
         )
       )(app => app.namespaceName.asRight[Throwable])
 
+      potentialNewChart =
```
rtitle, line 1146: I see, we're building the actual chart+version here and persisting to the database. So the APP table has
* appType: ALLOWED
* chart: the-actual-chart:version

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/BuildHelmChartValues.scala
rtitle, line 1: I didn't realize till now we moved helm values creation to a dedicated file. I like it :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/BuildHelmChartValues.scala
```diff
@@ -245,7 +245,7 @@ private[leonardo] object BuildHelmChartValues {
                                                stagingBucket: GcsBucketName,
                                                customEnvironmentVariables: Map[String, String]
   ): List[String] = {
-    val rstudioIngressPath = s"/proxy/google/v1/apps/${cluster.cloudContext.asString}/${appName.value}/rstudio-service"
+    val rstudioIngressPath = s"/proxy/google/v1/apps/${cluster.cloudContext.asString}/${appName.value}/app"
```
rtitle, line 248: Yeah so I imagine we'll need to add a `buildSasAppChartOverrideValuesString` method, right?

It would be so nice if we had a common `buildAppChartOverrideValuesString` with a standard set of Leo values, and app charts built to that spec…

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -1299,9 +1301,10 @@ class GKEInterpreter[F[_]](
         s"helm install for RSTUDIO app ${appName.value} in project ${cluster.cloudContext.asString}"
       ).compile.lastOrError
 
+      _ <- F.sleep(1 minutes) // RStudio won't be up right away, hence wait a bit before we start ping the services
```
rtitle: 1 minute could be a bit long, it could start before that..

http/src/main/resources/reference.conf
Qi77Qi: so the decoder will reject this. Btw, the chart name will be the fullname is what I'm currently thinking like (aou-sas-chart)...

```
  implicit val allowedChartNameDecoder: Decoder[AllowedChartName] =
    Decoder.decodeString.emap(x => AllowedChartName.stringToObject.get(x).toRight("chart name not allowed"))
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
Qi77Qi: yea...it's not implemneted in this PR since SAS won't get enabled yet...but it will be becuz for development purpose, only ppl who have SAS license can access it

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/BuildHelmChartValues.scala
Qi77Qi: yea for sure...definitely future optimization...
It'll be more clear how we can optimize once we attempt adding SAS.

http/src/main/resources/reference.conf
Qi77Qi: Conceptually, `ALLOWED` and whether welder-enabled should be orthogonal...Right now, `ALLOWED` only supports RSTUDIO (will be SAS soon), so I'm just adding welder here, but we can evolve things in the future to make this conditional (like a `welderEnabled` flag)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
Qi77Qi: yea exactly...what's persisted to DB will be the truth...what's in config is not entirely the final value

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/appRoutesModels.scala
Qi77Qi: good point...I'll add `ChartName` to response (I don't think it needs to be `AllowedChartName` cuz value from DB will already be validated anyway, and not all apps have `AllowedChartName` value)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
Qi77Qi: removed this for now

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
Qi77Qi: had offline discussions about naming...

ALLOWED is meant to be generic in a sense that it's a category of apps doesn't need to be specific to AoU

Will check metrics usage

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -15,9 +15,7 @@ tags:
   - name: disks
     description: Persistent Disks API.
   - name: apps
-    description: Experimental App API
-  - name: appsV2
-    description: Apps API v2. Support for Google Kubernetes Engine and Azure Kubernetes Service. Under active Development.
+    description: Experimental App API, Support for Google Kubernetes Engine and Azure Kubernetes Service. Under active Development.
```
rtitle: Should we remove `Experimental`? How about:
```
Apps API. Support Google Kubernetes Engine and Azure Kubernetes Service.
```

rtitle (approved):

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -1006,29 +1001,51 @@ class GKEInterpreter[F[_]](
...
+                  for {
+                    _ <- appErrorQuery
+                      .save(
+                        dbApp.app.id,
+                        AppError(e.getMessage, ctx.now, ErrorAction.StopApp, ErrorSource.App, None, Some(ctx.traceId))
+                      )
+                      .transaction
+                    _ <- appQuery.markAsDeleted(dbApp.app.id, ctx.now).transaction
+                    _ <-
+                      nodepoolQuery.markAsDeleted(dbApp.nodepool.id, ctx.now).transaction
```
Qi77Qi, line 1022: I added this logic here to handle NOT_FOUND a bit better...

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -2866,6 +2864,10 @@ components:
         - WDS
         - HAIL_BATCH
         - RSTUDIO
+    AllowedChartName:
```
LizBaldo, line 2867: Could you replace the RSTUDIO appType with ALLOWED in the api doc as well?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/Config.scala
```diff
@@ -689,15 +689,15 @@ object Config {
...
+    AllowedAppConfig(
       config.as[ChartName]("chartName"),
-      config.as[ChartVersion]("chartVersion"),
+      config.as[ChartVersion]("rstudioChartVersion"),
+      config.as[ChartVersion]("sasChartVersion"),
       config.as[NamespaceNameSuffix]("namespaceNameSuffix"),
       config.as[ReleaseNameSuffix]("releaseNameSuffix"),
       config.as[List[ServiceConfig]]("services"),
       config.as[ServiceAccountName]("serviceAccountName"),
-      config.as[Boolean]("enabled"),
```
LizBaldo, line 701: This make me think that there is likely a terra helm file PR that should be done with this, see https://github.com/broadinstitute/terra-helmfile/blob/34f228ec6e4f39ee4195ddb784015ee98fcdebeb/values/app/leonardo/live/prod.yaml#L167

LizBaldo (commented):
@Qi77Qi this looks good to me, I left a couple of comments, but I mainly have one question around app upgrade:

Given the app type ALLOWED, how would the upgrade be able to only upgrade RSTUDIO apps and not the SAS ones?

LizBaldo (approved):
Discussed with @Qi77Qi offline and I am good with upgrading all ALLOWED apps together for now 👍

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/Config.scala
Qi77Qi: for sure! I'll remove that once the PR is in!

Qi77Qi:
we haven't talked about upgrade story for `ALLOWED` apps...so punting that for now. but will potentially need it to be chart specific as we add more `ALLOWED` app

Qi77Qi:
```
 java.lang.Exception: Failed to create new billing project and workspace: <html>
<head><title>503 Service Temporarily Unavailable</title></head>
<body>
<center><h1>503 Service Temporarily Unavailable</h1></center>
<hr><center>nginx</center>
</body>
</html>
```

jenkins retest

Qi77Qi:
fiab-start failure

```
18:40:06 #1: Post "https://firecloudorch.jenkins-swat-leo-10039.bee.envs-terra.bio/register/profile": read tcp 172.17.0.1:49612->35.223.159.191:443: read: connection reset by peer
18:40:06 #2: 409 Conflict from https://firecloudorch.jenkins-swat-leo-10039.bee.envs-terra.bio/register/profile ({"causes":[{"causes":[],"message":"user harry.potter@quality.firecloud.org is already registered","source":"sam","stackTrace":[],"statusCode":409}],"message":"user harry.potter@quality.firecloud.org is already registered","source":"Sam","stackTrace":[],"statusCode":409})
```

jenkins retest

Qi77Qi:
unrelated errors
```
terra-quality-65d4fcf3/automation-test-asjlsh9hz errored due to List(RuntimeError(Failed to create cluster 24 due to 403 Forbidden POST https://storage.googleapis.com/upload/storage/v1/b/leoinit-automation-test-as-cd9b9515-10e8-4f61-a23f-d71868699db3/o?name=welder-docker-compose-gce.yaml&uploadType=resumable {   "error": {     "code": 403,     "message": "leonardo-qa@broad-dsde-qa.iam.gserviceaccount.com does not have storage.objects.create access to the Google Cloud Storage object. Permission 'storage.objects.create' denied on resource (or it may not exist).",     "errors": [       {         "message": "leonardo-qa@broad-dsde-qa.iam.gserviceaccount.com does not have storage.objects.create access to the Google Cloud Storage object. Permission 'storage.objects.create' denied on resource (or it may not exist).",         "domain": "global",         "reason": "forbidden"       }     ]   } } ,None,2023-08-09T00:31:28Z,Some(TraceId(668f35ac475548408edd3e46c02bf2c2/3939911508519804487))))
```

jenkins retest

Qi77Qi:
Leonardo build failed
```
21:20:32 Error: failed to fetch https://raw.githubusercontent.com/cloudve/helm-charts/anvil/charts/galaxykubeman-2.5.2.tgz : 502 Bad Gateway
21:20:32 helm.go:84: [debug] failed to fetch https://raw.githubusercontent.com/cloudve/helm-charts/anvil/charts/galaxykubeman-2.5.2.tgz : 502 Bad Gateway
21:20:32 helm.sh/helm/v3/pkg/getter.(*HTTPGetter).get
```

jenkins retest

Qi77Qi:
```
{"causes":[],"message":"Error creating workspace tmp-billing-project-bee2c6c329/0c5329b9-1786-4826-949f-2ebb637ba8ba-azure-test-workspace. Please try again.","source":"rawls","stackTrace":[],"statusCode":500}
```

it's all random faliures at this point...think I'm going to merge

PR #3516, opened 2023-07-25
[IA-4452] Infer cloudProvider from cloudContext and include it in DB …

mlilynolting:
…cluster queries to leverage the unique key index

__Jira ticket__: https://broadworkbench.atlassian.net/browse/[ticket_number]

### What

-

### Why

-

### What to test

- 

### Who tested and where

- [ ] This change is covered by automated tests 
  - _NB: Rerun automation tests on this PR by commenting `jenkins retest` or `jenkins multi-test`._
- [ ] I validated this change 
- [ ] Primary reviewer validated this change 
- [ ] I validated this change in the dev environment

PR #3514, opened 2023-07-24
[IA-4450] Remove explicit timeout on Proxy DNS host lookups

mlilynolting:
Remove timeouts wrapping DNS cache lookups used by the ProxyDAO.
Bump runtime DNS cache expiry time to 10 seconds. 
Bump runtime and aks DNS cache max size to 1000. 

__Jira ticket__: https://broadworkbench.atlassian.net/browse/IA-4450

## Dependencies
PROD-848, PROD-852 - "due to 5 seconds" timeout bug preventing launch of apps/runtimes

## Summary of changes

Increase cache expiry time and max size for the DNS host caches which the ProxyDAO uses to get the IP address associated with a given runtime/app. These methods are called everywhere Leo attempts to communicate with its runtimes or AKS apps, and the various settings - timeout 5 seconds, expiry time 5 seconds, max size 100 - have been in place for years. An increasing number of users report Jupyter startup failures which we traced to these timeouts.

Removing the timeout allows the DB lookup to take as long as it takes. This seems consistent with other DB lookups in Leonardo. (risky?)

Bumping cache max sizes decreases the frequency of cache misses. Other caches in Leo are set to 1000 or greater sizes; no other cache is as small as 100. (safe)

Bumping runtime cache expiry from 5 to 10 seconds also decreases miss frequency. Runtime IP address changes on stop/start, and a stale reference will fail, but I don't think that 10 seconds of a stale reference will break anything, since we retry attempts to send/receive to proxied servers. (risky?)

### What to test

- Stop and start GCP and Azure runtimes
- Create, update, and delete GCP and Azure runtimes
- Stop and start GCP and Azure apps
- Create, update, and delete GCP and Azure apps

### Who tested and where

- [ ] This change is covered by automated tests 
  - _NB: Rerun automation tests on this PR by commenting `jenkins retest` or `jenkins multi-test`._
- [ ] I validated this change 
- [ ] Primary reviewer validated this change 
- [ ] I validated this change in the dev environment

PR #3519, opened 2023-07-26
[IA-4449] Improve proxy DNS cache performance

mlilynolting:
* Use index in Cluster queries
* Remove explicit timeouts on DNS host lookups
* Allocate more resources to DNS caches

__Jira ticket__: https://broadworkbench.atlassian.net/browse/IA-4449

This PR combines https://github.com/DataBiosphere/leonardo/pull/3516 and https://github.com/DataBiosphere/leonardo/pull/3514

## Dependencies
Follows PROD-848, PROD-852 - "due to 5 seconds" timeout bug preventing launch of apps/runtimes

## Summary of changes

1. Add `cloudProvider` to queries against the CLUSTER table. This table is configured with only one index: `IDX_KUBERNETES_CLUSTER_UNIQUE_V2  | 1. cloudProvider  | 2. cloudContext  |  3. destroyedDate`. This is a multiple-column index:

> If the table has a multiple-column index, _any leftmost prefix of the index_ can be used by the optimizer to look up rows. For example, if you have a three-column index on (col1, col2, col3), you have indexed search capabilities on (col1), (col1, col2), and (col1, col2, col3). (see https://dev.mysql.com/doc/refman/8.0/en/multiple-column-indexes.html)

Existing functions depend on methods like `getRuntimeQueryByUniqueKey` which use only `cloudContext` and `destroyedDate` - presumably an artifact of the GCP-only days.

```
  def getRuntimeQueryByUniqueKey(cloudContext: CloudContext,
                                 clusterName: RuntimeName,
                                 destroyedDateOpt: Option[Instant]
  ) = {
    val destroyedDate = destroyedDateOpt.getOrElse(dummyDate)
    val baseQuery = clusterQuery
	  // without a filter on cloudProvider here, the index cannot be leveraged
      .filter(_.cloudContextDb === cloudContext.asCloudContextDb)
      .filter(_.runtimeName === clusterName)
      .filter(_.destroyedDate === destroyedDate)
```

2. Increase cache expiry time and max size for the DNS host caches which the ProxyDAO uses to get the IP address associated with a given runtime/app. These methods are called everywhere Leo attempts to communicate with its runtimes or AKS apps, and the various settings - timeout 5 seconds, expiry time 5 seconds, max size 100 - have been in place for years. An increasing number of users report Jupyter startup failures which we traced to these timeouts.

3. Removing the timeout allows the DB lookup to take as long as it takes. This seems consistent with other DB lookups in Leonardo.

4. Bumping cache max sizes decreases the frequency of cache misses. Other caches in Leo are set to 1000 or greater sizes; no other cache is as small as 100. (safe)

## Next Steps/Not Doing

Timeouts are also observed on app start and create. However the DB queries which create app responses are highly complex and more planning is needed before these can be optimized.

### What to test

- Stop and start GCP and Azure runtimes
- Create, update, and delete GCP and Azure runtimes
- Stop and start GCP and Azure apps
- Create, update, and delete GCP and Azure apps

### Who tested and where

- [x] I validated this change 

### DIFF
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -719,14 +719,14 @@ clusterFiles {
 # so we need to ensure we don't use stale cache entries.
 runtimeDnsCache {
   cacheExpiryTime = 5 seconds
-  cacheMaxSize = 10000
+  cacheMaxSize = 1000
 }
 
 # Kubernetes expiration can be higher because the IP is at the cluster level, which is
 # consistent across app pause/resume. Clusters are garbage collected ~1 hour after app deletion.
 kubernetesDnsCache {
   cacheExpiryTime = 10 minutes
-  cacheMaxSize = 10000
+  cacheMaxSize = 100
 }
 
 mysql {
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -719,14 +719,14 @@ clusterFiles {
 # so we need to ensure we don't use stale cache entries.
 runtimeDnsCache {
   cacheExpiryTime = 10 seconds
-  cacheMaxSize = 1000
+  cacheMaxSize = 10000
 }
 
 # Kubernetes expiration can be higher because the IP is at the cluster level, which is
 # consistent across app pause/resume. Clusters are garbage collected ~1 hour after app deletion.
 kubernetesDnsCache {
   cacheExpiryTime = 10 minutes
-  cacheMaxSize = 1000
+  cacheMaxSize = 10000
 }
 
 mysql {
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -718,15 +718,15 @@ clusterFiles {
 # The expiration is low because the IP changes when a runtime is pause/resumed,
 # so we need to ensure we don't use stale cache entries.
 runtimeDnsCache {
-  cacheExpiryTime = 5 seconds
-  cacheMaxSize = 100
+  cacheExpiryTime = 10 seconds
+  cacheMaxSize = 1000
 }
 
 # Kubernetes expiration can be higher because the IP is at the cluster level, which is
 # consistent across app pause/resume. Clusters are garbage collected ~1 hour after app deletion.
 kubernetesDnsCache {
   cacheExpiryTime = 10 minutes
-  cacheMaxSize = 100
+  cacheMaxSize = 1000
 }
 
 mysql {
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -718,7 +718,7 @@ clusterFiles {
 # The expiration is low because the IP changes when a runtime is pause/resumed,
 # so we need to ensure we don't use stale cache entries.
 runtimeDnsCache {
-  cacheExpiryTime = 10 seconds
+  cacheExpiryTime = 5 seconds
   cacheMaxSize = 10000
 }
 
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/ProxyDAO.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/ProxyDAO.scala
@@ -7,8 +7,6 @@ import cats.effect.implicits._
 import org.broadinstitute.dsde.workbench.leonardo.dns._
 import org.http4s.Uri
 
-import scala.concurrent.duration._
-
 sealed trait HostStatus extends Product with Serializable
 object HostStatus {
   final case object HostNotFound extends HostStatus
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/ProxyDAO.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/ProxyDAO.scala
@@ -43,11 +43,10 @@ object Proxy {
   ): F[HostStatus] =
     runtimeDnsCache
       .getHostStatus(RuntimeDnsCacheKey(cloudContext, runtimeName))
-      .timeout(10 seconds)
 
   def getAppTargetHost[F[_]: Async](kubernetesDnsCache: KubernetesDnsCache[F],
                                     cloudContext: CloudContext,
                                     appName: AppName
   ): F[HostStatus] =
-    kubernetesDnsCache.getHostStatus(KubernetesDnsCacheKey(cloudContext, appName)).timeout(10 seconds)
+    kubernetesDnsCache.getHostStatus(KubernetesDnsCacheKey(cloudContext, appName))
 }
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/ProxyDAO.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/ProxyDAO.scala
@@ -3,7 +3,6 @@ package dao
 
 import akka.http.scaladsl.model.Uri.Host
 import cats.effect.Async
-import cats.effect.implicits._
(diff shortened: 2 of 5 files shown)


rtitle (approved):
Good find! Yeah `cloudProvider` is in the unique index now:
```
MySQL [leonardo]> SHOW INDEXES FROM CLUSTER;
+---------+------------+-------------------------------------------+--------------+-----------------+-----------+-------------+----------+--------+------+------------+---------+---------------+
| Table   | Non_unique | Key_name                                  | Seq_in_index | Column_name     | Collation | Cardinality | Sub_part | Packed | Null | Index_type | Comment | Index_comment |
+---------+------------+-------------------------------------------+--------------+-----------------+-----------+-------------+----------+--------+------+------------+---------+---------------+
| CLUSTER |          0 | PRIMARY                                   |            1 | id              | A         |       96111 |     NULL | NULL   |      | BTREE      |         |               |
| CLUSTER |          0 | IDX_CLUSTER_UNIQUE_V2                     |            1 | cloudProvider   | A         |           2 |     NULL | NULL   |      | BTREE      |         |               |
| CLUSTER |          0 | IDX_CLUSTER_UNIQUE_V2                     |            2 | cloudContext    | A         |       96111 |     NULL | NULL   |      | BTREE      |         |               |
| CLUSTER |          0 | IDX_CLUSTER_UNIQUE_V2                     |            3 | runtimeName     | A         |       96111 |     NULL | NULL   |      | BTREE      |         |               |
| CLUSTER |          0 | IDX_CLUSTER_UNIQUE_V2                     |            4 | destroyedDate   | A         |       96111 |     NULL | NULL   |      | BTREE      |         |               |
| CLUSTER |          0 | IDX_CLUSTER_RUNTIME_NAME_WORKSPACE_UNIQUE |            1 | workspaceId     | A         |         620 |     NULL | NULL   | YES  | BTREE      |         |               |
| CLUSTER |          0 | IDX_CLUSTER_RUNTIME_NAME_WORKSPACE_UNIQUE |            2 | runtimeName     | A         |       96111 |     NULL | NULL   |      | BTREE      |         |               |
| CLUSTER |          0 | IDX_CLUSTER_RUNTIME_NAME_WORKSPACE_UNIQUE |            3 | destroyedDate   | A         |       96111 |     NULL | NULL   |      | BTREE      |         |               |
| CLUSTER |          1 | IDX_CLUSTER_STATUS                        |            1 | status          | A         |          16 |     NULL | NULL   |      | BTREE      |         |               |
| CLUSTER |          1 | IDX_CLUSTER_CLUSTER_SERVICE_ACCOUNT       |            1 | serviceAccount  | A         |       96111 |     NULL | NULL   | YES  | BTREE      |         |               |
| CLUSTER |          1 | FK_CLUSTER_RUNTIME_CONFIG_ID              |            1 | runtimeConfigId | A         |       96111 |     NULL | NULL   |      | BTREE      |         |               |
+---------+------------+-------------------------------------------+--------------+-----------------+-----------+-------------+----------+--------+------+------------+---------+---------------+
11 rows in set (0.036 sec)
```

mlilynolting:
jenkins retest

http/src/main/resources/reference.conf
```diff
@@ -719,14 +719,14 @@ clusterFiles {
 # so we need to ensure we don't use stale cache entries.
 runtimeDnsCache {
   cacheExpiryTime = 5 seconds
-  cacheMaxSize = 100
+  cacheMaxSize = 10000
```
Qi77Qi: I don't think this is necessary... this is prod metrics for this cache size
Screenshot 2023-07-26 at 3 53 28 PM

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/ProxyDAO.scala
```diff
@@ -43,11 +40,10 @@ object Proxy {
...
     runtimeDnsCache
       .getHostStatus(RuntimeDnsCacheKey(cloudContext, runtimeName))
-      .timeout(10 seconds)
 
   def getAppTargetHost[F[_]: Async](kubernetesDnsCache: KubernetesDnsCache[F],
                                     cloudContext: CloudContext,
                                     appName: AppName
   ): F[HostStatus] =
-    kubernetesDnsCache.getHostStatus(KubernetesDnsCacheKey(cloudContext, appName)).timeout(10 seconds)
+    kubernetesDnsCache.getHostStatus(KubernetesDnsCacheKey(cloudContext, appName))
```
Qi77Qi, line 48: I think this is still useful guard, but maybe we can respond to the timeout a bit better in leo

http/src/main/resources/reference.conf
```diff
@@ -719,14 +719,14 @@ clusterFiles {
...
 }
 
 # Kubernetes expiration can be higher because the IP is at the cluster level, which is
 # consistent across app pause/resume. Clusters are garbage collected ~1 hour after app deletion.
 kubernetesDnsCache {
   cacheExpiryTime = 10 minutes
-  cacheMaxSize = 100
+  cacheMaxSize = 10000
```
Qi77Qi: same comment...

mlilynolting: Agreed, will leave as is

http/src/main/resources/reference.conf
mlilynolting: Thank you! I'm going to bump this one to 1000 only

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/ProxyDAO.scala
mlilynolting: My concern is if timing this out leaves us with a permanent cache miss/aborted query loop. I'd advocate for removing the timeout and accepting that a slow query execution will at least populate the cache.

mlilynolting: If we're concerned about lots of duplicate queries stacking up as multiple threads try to getTargetHost during the long execution time (a cache miss storm), we could:
1. Warm the cache on boot (prepopulate with ip values)
2. Set a lock on the cache lookup via a keyed mutex, so only one DB query for that key is allowed at a time

(https://stackoverflow.com/questions/23303463/dealing-with-concurrency-issues-when-caching-for-high-traffic-sites)

mlilynolting: @rtitle and @Qi77Qi , is there a way we could see how many cache misses are going to the database at once? Or would we need to add totally new metrics to determine this?

Qi77Qi (approved):

mlilynolting:
Jigsaw swat pass:

https://fc-jenkins.dsp-techops.broadinstitute.org/job/leonardo-fiab-test-runner/56452/consoleFull
  should stop/start a Dataproc cluster with workers and preemptible workers - failed with compute.images.get perms proc
  NotebookPyKernelSpec - aborted with Jupyter, Welder failed to start (JWF)

https://fc-jenkins.dsp-techops.broadinstitute.org/job/leonardo-fiab-test-runner/56454/consoleFull
  should create a Dataproc cluster in a non-default region - failed with compute.images.get propagation
  should stop/start a Dataproc cluster with workers and preemptible workers - cancelled with flicker (TestFailedDueToTimeoutException: ... Stopping was not equal to Stopped.)
  create runtime and attach an existing persistent disk or clone - cancelled with flicker (JWF)

https://fc-jenkins.dsp-techops.broadinstitute.org/job/leonardo-fiab-test-runner/56464/consoleFull
  environment variables should be available in RStudio - failed with Selenium timeout
  NotebookRKernelSpec - aborted with JWF

PR #3542, opened 2023-08-02
[IA-4468] Adjust front Leo DNS cache settings

rtitle:
__Jira ticket__: https://broadworkbench.atlassian.net/browse/IA-4468

### What

Two changes here:

1. Increase RuntimeDnsCache expiry from 5 seconds to 1 minute. The expiry is low because runtime IPs change during pause/resume. But 1 minute should be safe to allow for a pause/resume cycle. With 5 seconds we were getting a lot of evictions + cache loads, resulting in a lot of DB queries.

2. In addition to the caffeine cache we have an unbouded `Map[Host, IP]` which plugs into ProxyResolver. I was looking at akka-http Host class and I'm not sure it's a good HashMap key -- it's not a case class and doesn't have equals or hashCode implemented. So it might have been relying on reference equality and thus the Map may have been growing quickly with all the cache loads we were doing. We just unwrap the String anyway in proxy resolver so I change the map key to a String.

### Why

Attempt to address the GC pressure + restarts we're seeing in prod starting on 8/1.

The changes here are a little speculative (but shouldn't hurt). Analyzing a heap dump might give more insight.

### What to test

- 

### Who tested and where

- [ ] I validated this change 

### DIFF
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -718,7 +718,7 @@ clusterFiles {
 # The expiration is low because the IP changes when a runtime is pause/resumed,
 # so we need to ensure we don't use stale cache entries.
 runtimeDnsCache {
-  cacheExpiryTime = 5 seconds
+  cacheExpiryTime = 1 minute
   cacheMaxSize = 1000
 }
 
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dns/KubernetesDnsCache.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dns/KubernetesDnsCache.scala
@@ -1,6 +1,5 @@
 package org.broadinstitute.dsde.workbench.leonardo.dns
 
-import akka.http.scaladsl.model.Uri.Host
 import cats.effect.{Async, Ref}
 import cats.syntax.all._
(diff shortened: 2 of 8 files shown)


LizBaldo (approved):

lucymcnatt (approved):

mlilynolting (approved):

rtitle:
Tested on a BEE:
1. Created GCP runtime, ran a notebook
2. Saved notebook a bunch of times
3. Pause/resumed the runtime (def took more than 1 minute)
4. Ran the notebook again, still worked

rtitle:
Jenkins tests failing:
```
++ set +x
14:58 ERR error retrieving OAuth client credentials from Vault: Error making API request.

URL: GET https://clotho.broadinstitute.org:8200/v1/secret/dsp/identity-aware-proxy/dsp-tools-k8s/dsp-tools-k8s-iap-oauth_client-credentials.json
Code: 403. Errors:
```

jenkins retest

lucymcnatt:
jenkins retest

rtitle:
Tests ran this time -- `RuntimeDataprocSpec` and `RuntimeCreationDiskSpec` failed which I think are pretty common/known transient failures. So probably going to admin merge this if no objections.

LizBaldo:
Go for it 👍

PR #3545, opened 2023-08-03
WM-2101: New workflows apps config

cjllanwarne:
## Summary of changes

- Adds app definitions for two new app types: WORKFLOWS_APP and CROMWELL_RUNNER_APP.
- No deployment logic yet, that is to follow
- Enough configuration for us to move ahead with adding a CI/CD flow from `terra-helmfile` to auto-update these apps when the charts in `terra-helmfile` change.
- Enough data types and definitions to allow adding app deployment logic in smaller steps that are hopefully easier to reason about and review.

## Testing these changes

- Regression testing to make sure nothing was broken, but no live control flow changes here to test, just data structures and config for us to build on top of.


core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
```diff
@@ -348,10 +354,10 @@ object AppType {
...
   def appTypeToFormattedByType(appType: AppType): FormattedBy =
     appType match {
-      case Galaxy                     => FormattedBy.Galaxy
-      case Custom                     => FormattedBy.Custom
-      case RStudio                    => FormattedBy.RStudio
-      case Cromwell | Wds | HailBatch => FormattedBy.Cromwell
+      case Galaxy                                                        => FormattedBy.Galaxy
+      case Custom                                                        => FormattedBy.Custom
+      case RStudio                                                       => FormattedBy.RStudio
+      case Cromwell | Wds | HailBatch | WorkflowsApp | CromwellRunnerApp => FormattedBy.Cromwell
```
jgainerdewar, line 382: I know you're just following the existing pattern here, but... weird pattern, right?

cjllanwarne: Hmm, yeah.
1. I believe this is also entirely unnecessary for these apps which won't really do anything with their persistent disks
2. But might also cause problems if we ever try to run both apps in the same workspace, will they be fighting over disks they don't need 🤔

cjllanwarne: I guess obviously it's not a problem since Cromwell, WDS and HailBatch are already sharing this, but I don't think there's any reason not to have more cases here

Qi77Qi: this can definitely be simplified...we pretty much only care if the app is galaxy, and don't wanna reuse a disk used by galaxy for other apps...all other cases are probably ok

kpierre13 (approved):

http/src/main/resources/reference.conf
```diff
@@ -285,6 +285,52 @@ azure {
...
+    chart-version = "0.1.0"
+    release-name-suffix = "cromwell-runner-app-rls"
+    namespace-name-suffix = "cromwell-runner-app-ns"
+    ksa-name = "cromwell-runner-app-ksa"
+    dockstore-base-url = "https://staging.dockstore.org/"
+    # See https://github.com/broadinstitute/cromwhelm/blob/main/terra-batch-libchart/templates/_reverse-proxy.tpl#L81-L114
+    # for list of services.
+    services = [
+      {
+        name = "cromwell"
```
rtitle: Will having a `/cromwell` path for both workflows-app and cromwell-runner-app conflict when both are running? (Wonder if they should be distinct.)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/KubernetesAppConfig.scala
```diff
@@ -117,6 +117,51 @@ final case class CoaAppConfig(chartName: ChartName,
...
+                                    databaseEnabled: Boolean,
+                                    chartVersionsToExcludeFromUpdates: List[ChartVersion]
+) extends KubernetesAppConfig {
+  override lazy val kubernetesServices: List[KubernetesService] = services.map(s => KubernetesService(ServiceId(-1), s))
+  override val serviceAccountName = ServiceAccountName(ksaName.value)
+
+  val cloudProvider: CloudProvider = CloudProvider.Azure
+  val appType: AppType = AppType.WorkflowsApp
+
+  def workflowsAppServices: Set[WorkflowsAppService] = services
```
rtitle, line 139: Doesn't need to be for this PR.. but wondering if we can remove the `WorkflowsAppService` enum and related methods. I think this was introduced when Cromwell/CBAS/WDS all shared a single chart, and we wanted to pick and choose which services to deploy. Nowadays we have dedicated charts for each service, so probably don't need this granularity anymore.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
rtitle, line 1: I imagine this file will need case statements in `createAndPollApp` and `updateAndPollApp`, methods to generate helm values, etc…. is that for a follow-on PR?

rtitle (approved):

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
cjllanwarne: Yes exactly, that's to follow. There's so many changes that we're planning to do this one step at a time instead of all at once

http/src/main/resources/reference.conf
salonishah11: @rtitle since Chris is on vacation, I am working on getting this PR merged. I am not sure if names for the services need to be unique or not. @Qi77Qi or @LizBaldo do you know if `/cromwell` path for 2 different apps will clash or not?

salonishah11:
jenkins retest

http/src/main/resources/reference.conf
LizBaldo: I do not know if it will crash, but for better clarity it might be nice to name them differently regardless?

kpierre13:
jenkins retest

dillydally414:
jenkins retest

http/src/main/resources/reference.conf
salonishah11: @rtitle or @LizBaldo do you if the service name is changed here a similar name change would need to be made in helm charts somewhere? 🤔

LizBaldo: don't think so, but it depends on your helmchart, I don't know if it expects to use a specific service name

kpierre13:
jenkins retest

http/src/main/resources/reference.conf
salonishah11: Hmm I am not so sure. Since this PR is meant to get us started on the new app types (nothing is deployed yet), I will rename the `/cromwell` paths and if things need to be updated in helm charts they can be done as part of another PR

kpierre13:
jenkins retest

kpierre13:
jenkins retest

PR #3546, opened 2023-08-03
Use slightly newer Broad cloudsqlproxy image

em-may:
Fixes dbconnect by using a slightly newer version of the old cloudsqlproxy (v1).


mlilynolting (approved):
tysm!!

PR #3551, opened 2023-08-08
[AJ-1083] [IA-4441] Relay Listener liveness, metrics, and upgrades

rtitle:
https://broadworkbench.atlassian.net/browse/IA-4441
https://broadworkbench.atlassian.net/browse/AJ-1083

## Dependencies
* relay-listener: https://github.com/DataBiosphere/terra-azure-relay-listeners/pull/45
* wb-libs: https://github.com/broadinstitute/workbench-libs/pull/1447
* terra-helmfile: https://github.com/broadinstitute/terra-helmfile/pull/4411
* terra-app: https://github.com/DataBiosphere/terra-app/pull/46

## Summary of changes

Several Relay Listener improvements (see below).

### What

- Add DAO for listener status endpoint, and populate Leo App Metrics Dashboard
- Switch to terra-helmfile managed chart
- Update relay-listener deployments as part of App Upgrades flow

### Why

- Compliance, system reliability, maintenance

### What to test

Tested on a BEE.

1. Deployed Leo change w/ all dependencies
2. Made a few workspaces/apps
3. Validated listener liveness/readiness is working in k8s
4. Validated `/listenerstatus` endpoint
5. Validated Leo App Metrics dashboard
6. Update an app; validate listener is also updated

### Who tested and where

- [ ] I validated this change 


Dockerfile
```diff
@@ -63,6 +64,8 @@ RUN helm repo add ingress-nginx https://kubernetes.github.io/ingress-nginx && \
 # pulling `terra-app-setup` locally and add cert files to the chart.
 # Leonardo will install the chart from local version.
 # We are also caching charts so they are not downloaded with every helm-install
+COPY ./relay-listener-0.1.0.tgz /leonardo
```
rtitle: TODO: remove

http/src/main/resources/reference.conf
```diff
@@ -215,7 +215,7 @@ azure {
         minor-version-auto-upgrade = true,
         file-uris = ["https://raw.githubusercontent.com/DataBiosphere/leonardo/270bd6aad916344fadc06d1a51629c432da663a8/http/src/main/resources/init-resources/azure_vm_init_script.sh"]
       }
-      listener-image = "terradevacrpublic.azurecr.io/terra-azure-relay-listeners:9c988ec"
+      listener-image = "terradevacrpublic.azurecr.io/terra-azure-relay-listeners:1d97e9d"
```
rtitle: TODO: update hash once Listener PR is merged

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
```diff
@@ -1138,11 +1138,16 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
...
-        )
+        KubernetesName
+          .withValidation(
+            s"${uid}-${gkeAppConfig.releaseNameSuffix.value}",
+            Release.apply
+          )
       )(app => app.release.asRight[Throwable])
+      services =
+        if (cloudContext.cloudProvider == CloudProvider.Azure) {
+          gkeAppConfig.kubernetesServices.appended(RelayListenerChartConfig.service)
```
rtitle: I thought about making the Relay Listener a full-fledged `KubernetesAppConfig`. But it doesn't share all attributes with the other apps -- it's more a thing implicitly attached to Azure apps, than a standalone app config. So here, if the app is Azure, I'm just appending the `relay-listener` service to persist to the DB.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -733,57 +749,6 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
       ).interruptAfter(config.appMonitorConfig.updateApp.interruptAfter).compile.lastOrError
     } yield appOk.isDone
 
-  private[util] def buildSetupChartOverrideValues(release: Release,
```
rtitle, line 749: Moved to `BuildHelmChartValues` object

Dockerfile
Qi77Qi: didnt' realize relay is installed as it's own chart

Qi77Qi (commented):
do u mean to check in relay-listener-0.1.0.tgz?

Dockerfile
rtitle: It used to be part of the setup chart, but we're breaking it into a dedicated chart in terra-helmfile. See https://github.com/broadinstitute/terra-helmfile/pull/4411

rtitle:
> do u mean to check in relay-listener-0.1.0.tgz?

Temporarily yes so Jenkins can build the Leo image -- need to remove before merge though.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
LizBaldo: That makes sense, at some point, should we also think about moving the rest of the setup chart into its own config outside of tera-app and deploy it in a similar manner?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -176,18 +175,21 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
...
       )
 
       _ <- logger.info(ctx.loggingCtx)(
-        s"Setup chart values for app ${params.appName.value} are ${values.asString}"
+        s"Relay listener values for app ${params.appName.value} are ${values.asString}"
       )
 
       _ <- helmClient
         .installChart(
-          getTerraAppSetupChartReleaseName(app.release),
```
LizBaldo, line 188: I think I am a bit confused now, how are we installing the rest of the setup (pod identity etc.) now?

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/ConfigReaderSpec.scala
```diff
@@ -24,7 +24,7 @@ class ConfigReaderSpec extends AnyFlatSpec with Matchers {
   it should "read config file correctly" in {
     val config = ConfigReader.appConfig
     val expectedConfig = AppConfig(
-      TerraAppSetupChartConfig(ChartName("/leonardo/terra-app-setup"), ChartVersion("0.0.19")),
+      TerraAppSetupChartConfig(ChartName("/leonardo/terra-app-setup"), ChartVersion("0.0.20")),
```
LizBaldo: I thought we bumped to 0.1.0?

LizBaldo (commented):
LGTM, I might just need some clarification on a couple of things:
- Are we still installing the setup chart?
- Could this be deployed for runtimes as well? (crazy question, but I would love to only maintain one way of deploying tarl)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
rtitle: That may make sense, the terra-helmfile repo is where most of our "internal" charts are landing these days.

One note: after this change, we no longer need to deploy the setup chart for Azure apps (the listener was the only thing we were using). So it's more of a GCP-only thing now.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
rtitle: With the listener deployment moved, there are no longer any Azure things needed in the setup chart!

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/ConfigReaderSpec.scala
rtitle: Oops, so that's why the test is failing :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
LizBaldo: Ohhh I missed that completely, in my mind the pod identity was deployed via this I don't know why 😓 In that case, this makes perfect sense!

LizBaldo (approved):
LGTM!

rtitle:
> Could this be deployed for runtimes as well? (crazy question, but I would love to only maintain one way of deploying tarl)

Me too! I think we could certainly start using tarl-in-k8s for runtimes (and not need to deploy it on the DSVM anymore). It's probably better security too. May need a little design/exploration but overall it's a good direction I think.

rtitle:
I rebased and retested this on a BEE.. going to merge.

PR #3552, opened 2023-08-08
[IA-4363] Adding Project Permission Check to V1 endpoints

lucymcnatt:

### What

- Adding project-level Sam checks to V1 endpoints:
- Disks (get, list, delete, update)
- Apps (get, list, delete, start, stop)
- Runtimes (get, list, delete, start, stop, update)

### Why

- Users who are removed from projects shouldn't be able to see or interact with those resources

### What to test

- Add a user to a project
- Have them create an app and a runtime
- Have them get, list, start, stop, update successfully
- Remove the user from that workspace
- Attempt to get, list (both by project and listAll), start, stop, update the disk/runtime/app
- Should return a 403 or an empty list

### Who tested and where

- [ ] I validated this change 


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/SamAuthProvider.scala
```diff
@@ -155,19 +155,35 @@ class SamAuthProvider[F[_]: OpenTelemetryMetrics](
     val authHeader = Authorization(Credentials.Token(AuthScheme.Bearer, userInfo.accessToken.token))
     val resourceTypes = resources.map(r => sr.resourceType(r._2)).toList.toSet
     for {
-      projectPolicies <- samDao.getResourcePolicies[ProjectSamResourceId](authHeader, SamResourceType.Project)
-      owningProjects = projectPolicies.collect { case (r, SamPolicyName.Owner) =>
+      projectPolicies <- samDao
+        .getResourcePolicies[ProjectSamResourceId](authHeader, SamResourceType.Project)
+      readingProjects = projectPolicies.map(_._1.googleProject)
+      _ <- logger.info(s"Reading: $readingProjects")
```
LizBaldo: Were these logging statements for debugging? I agree that these would make things a bit easier to debug, but this call is made very frequently and I am a bit afraid of making the logs too noisy

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/SamAuthProvider.scala
```diff
@@ -155,19 +155,35 @@ class SamAuthProvider[F[_]: OpenTelemetryMetrics](
...
       }
+      _ <- logger.info(s"Owned: $ownedProjects")
       resourcePolicies <- resourceTypes.toList.flatTraverse(resourceType =>
         samDao.getResourcePolicies[R](authHeader, resourceType)
       )
+      _ <- logger.info(s"resourcePolicies: $resourcePolicies")
       res = resourcePolicies.filter { case (r, pn) => sr.policyNames(r).contains(pn) }
     } yield resources.filter { case (project, r) =>
-      owningProjects.contains(project) || res.exists(_._1 == r)
+      // user must be reader on project and resource, or a project owner
```
LizBaldo: Is this check redundant with the `isUserProjectReader below`?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskServiceInterp.scala
```diff
@@ -173,6 +173,14 @@ class DiskServiceInterp[F[_]: Parallel](config: PersistentDiskConfig,
   ): F[GetPersistentDiskResponse] =
     for {
       ctx <- as.ask
+
+      // throw 403 if no project-level permission
```
LizBaldo, line 177: I am always conflicted between returning a 403, or a 404. On one end, returning a 404 is safer, but a 403 is a lot more user friendly. I think I am fine with it be we might want to have a bigger conversation as a team and harmonize the error code we want to return

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
```diff
@@ -1262,6 +1303,65 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
...
+    samResources = allClusters.map(a =>
+      // Need the project to filter projects not visible to user
+      (GoogleProject(a.cloudContext.toString), a.nodepools.flatMap(_.apps.map(_.samResourceId)))
+    )
+    apps = samResources.flatMap { case (gp, srList) =>
+      srList.map(sr => (gp, sr.copy(accessScope = sr.accessScope.orElse(Some(AppAccessScope.UserPrivate)))))
+    }
+    // Partition apps by shared vs. private access scope and make a separate Sam call for each, then re-combine them.
+    // The reason we do this shared vs. private app access scope is represented by different Sam resource types/actions,
+    // and therefore needs different decoders to process the result list.
```
LizBaldo, line 1306: 👍

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
```diff
@@ -1262,6 +1303,65 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
     _ <- if (res.isEmpty) authProvider.checkUserEnabled(userInfo) else F.unit
   } yield res
 
+  private def filterAppsBySamProjectPermission(allClusters: List[KubernetesCluster],
```
LizBaldo: I am not 100% sure, but I believe the logic of partitioning apps by access scope already exists somewhere (Rob added it in), would it make sense to reuse it?

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterpSpec.scala
```diff
@@ -819,6 +820,30 @@ final class AppServiceInterpSpec extends AnyFlatSpec with LeonardoTestSuite with
     }
   }
 
+  it should "error on delete if user is removed from app's project" in isolatedDbTest {
```
LizBaldo: Nit, `user` -> `app creator`?

LizBaldo (approved):
LGTM! If you have tested the scenarios from your PR description, could you let us know in the PR as well? If not, happy to help test this.

I have a couple of questions and some very minor comments, but nothing blocking 👍

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/SamAuthProvider.scala
lucymcnatt: oh oops I definitely meant to delete that, thanks!

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/SamAuthProvider.scala
lucymcnatt: I don't know if I understand the redundancy, this is filtering out resources where the user isn't a reader of the associated project

LizBaldo: Ohhhh got it! No issues then, that makes sense :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
lucymcnatt: yeah I tried to reuse that function (`filterAppsBySamPermission`), but there's enough changes required in multiple parts of the function that it required some funky logic to not break it out (still need the original function for v2 endpoints), I can take another stab at it though

LizBaldo: Nah, no need, just wanted to make sure you saw the original logic, if it is different enough, no need to refactor

lucymcnatt:
jenkins retest

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
```diff
@@ -1240,6 +1281,69 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
...
     }
+  } yield samVisibleAppsOpt
+
+  private def getVisibleAppsByProject(allClusters: List[KubernetesCluster], userInfo: UserInfo)(implicit
+    as: Ask[F, AppContext]
+  ): F[Option[List[AppSamResourceId]]] = for {
+    ctx <- as.ask
+    samResources = allClusters.map(a =>
+      // Need the project to filter projects not visible to user
+      (GoogleProject(a.cloudContext.asString), a.nodepools.flatMap(_.apps.map(_.samResourceId)))
```
LizBaldo, line 1298: Ahh, is this the famous `asString` that fixed it for you?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
```diff
@@ -1240,6 +1281,69 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
...
+                                        useGoogleProject: Boolean
+  )(implicit
+    as: Ask[F, AppContext]
+  ): F[Vector[ListAppResponse]] = for {
+    ctx <- as.ask
+
+    // V1 endpoints use google project to determine user access
+    // Azure apps don't have a google project
+    partition = allClusters.partition(_.cloudContext.isInstanceOf[CloudContext.Gcp])
+    samVisibleAppsOpt <- useGoogleProject match {
```
LizBaldo, line 1336: Do we need to grab both gcp and azure when using the google project because of the List ALL apps endpoint? If so could you leave a little comment to explain? I thought this was a bug for a second when reading the code

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
lucymcnatt: ughhh yes, so innocuous, but so annoying

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
lucymcnatt: yeah no prob, and you're correct in that the V1 listAll endpoint handles both Azure and GCP, so I have to handle the case where we want to filter the GCP apps by their project and not the Azure ones. I thought about putting that logic further down the pipeline, but it seemed like it was going to make things more complicated and I'd rather keep  `getVisibleAppsByProject` true to the name

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskServiceInterp.scala
```diff
@@ -192,6 +200,16 @@ class DiskServiceInterp[F[_]: Parallel](config: PersistentDiskConfig,
   ): F[Vector[ListPersistentDiskResponse]] =
     for {
       ctx <- as.ask
+
+      // throw 403 if user doesn't have project permission
+      hasProjectPermission <- cloudContext.traverse(cc =>
```
mlilynolting, line 205: . Is an Option traversable?
. If cloudContext is None, does this permit listing all disks? (is that what the `getOrElse(true)` handles?)
. If a user *is* listing all disks, the filterResourceProjectVisible step will ensure that I see only disks under projects I have read-or-greater access to. Right?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/SamAuthProvider.scala
```diff
@@ -154,20 +154,34 @@ class SamAuthProvider[F[_]: OpenTelemetryMetrics](
...
         r.googleProject
       }
       resourcePolicies <- resourceTypes.toList.flatTraverse(resourceType =>
         samDao.getResourcePolicies[R](authHeader, resourceType)
       )
       res = resourcePolicies.filter { case (r, pn) => sr.policyNames(r).contains(pn) }
     } yield resources.filter { case (project, r) =>
-      owningProjects.contains(project) || res.exists(_._1 == r)
+      // user must be a project owner or at least a reader on the project and resource
+      ownedProjects.contains(project) || (readingProjects.contains(project) && res.exists(_._1 == r))
```
mlilynolting: What is `_._1` here?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/SamAuthProvider.scala
```diff
@@ -154,20 +154,34 @@ class SamAuthProvider[F[_]: OpenTelemetryMetrics](
...
+      projectPolicies <- samDao
+        .getResourcePolicies[ProjectSamResourceId](authHeader, SamResourceType.Project)
+      readingProjects = projectPolicies.map(_._1.googleProject).toSet
+      ownedProjects = projectPolicies.collect { case (r, SamPolicyName.Owner) =>
         r.googleProject
       }
       resourcePolicies <- resourceTypes.toList.flatTraverse(resourceType =>
         samDao.getResourcePolicies[R](authHeader, resourceType)
       )
       res = resourcePolicies.filter { case (r, pn) => sr.policyNames(r).contains(pn) }
```
mlilynolting: This is, like, resource policies which I have, and which apply to the resource I'm trying to view, right?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/SamAuthProvider.scala
mlilynolting: Is it a `SamPolicyName`?

mlilynolting: Oh god is this 1-indexed? Scala wtf

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskServiceInterp.scala
```diff
@@ -201,17 +219,16 @@ class DiskServiceInterp[F[_]: Parallel](config: PersistentDiskConfig,
...
       _ <- ctx.span.traverse(s => F.delay(s.addAnnotation("Done checking Sam permission")))
       res = samVisibleDisksOpt match {
         case None => Vector.empty
         case Some(samVisibleDisks) =>
           val samVisibleDisksSet = samVisibleDisks.toSet
-          // Making the assumption that users will always be able to access disks that they create
           disks
             .filter(d =>
-              d.auditInfo.creator == userInfo.userEmail || samVisibleDisksSet.contains(
+              samVisibleDisksSet.contains(
```
mlilynolting, line 231: What is the behavior now, if I'm the creator of a disk which exists in Leonardo but not in Sam, but I still have project read access? I think that the revised `filterResourceProjectVisible` might handle this case, but then again might not -- it looks like it's checking that the resource exists, so this change might remove support for the case I'm describing

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskServiceInterp.scala
```diff
@@ -192,6 +200,16 @@ class DiskServiceInterp[F[_]: Parallel](config: PersistentDiskConfig,
...
+      // throw 403 if user doesn't have project permission
+      hasProjectPermission <- cloudContext.traverse(cc =>
+        authProvider.isUserProjectReader(
+          cc,
+          userInfo
+        )
+      )
+      _ <- F.raiseWhen(!hasProjectPermission.getOrElse(true))(ForbiddenError(userInfo.userEmail, Some(ctx.traceId)))
+
       paramMap <- F.fromEither(processListParameters(params))
```
mlilynolting, line 213: This is very much existing, but why is this called a `paramMap`? it's a tuple

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
```diff
@@ -1240,6 +1281,69 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
...
+                                        labels: List[String],
+                                        useGoogleProject: Boolean
+  )(implicit
+    as: Ask[F, AppContext]
+  ): F[Vector[ListAppResponse]] = for {
+    ctx <- as.ask
+
+    // V1 endpoints use google project to determine user access
+    // Azure apps don't have a google project
+    partition = allClusters.partition(_.cloudContext.isInstanceOf[CloudContext.Gcp])
```
mlilynolting: Do we only need this in the `true` case of the match below?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeServiceInterp.scala
```diff
@@ -261,11 +281,9 @@ class RuntimeServiceInterp[F[_]: Parallel](config: RuntimeServiceConfig,
         case None => Vector.empty
         case Some(samVisibleRuntimes) =>
           val samVisibleRuntimesSet = samVisibleRuntimes.toSet
-          // Making the assumption that users will always be able to access runtimes that they create
-          // Fix for https://github.com/DataBiosphere/leonardo/issues/821
           runtimes
             .filter(c =>
-              c.auditInfo.creator == userInfo.userEmail || samVisibleRuntimesSet
```
mlilynolting, line 268: I see you've consistently removed this (that's great, this sucks, lol). I think I need clarity on what the behavior is meant to be, now, in the several different edge cases that might arise (app|runtime creator leaves workspace, workspace member sees another's app|runtime, app|runtime errors)

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/dao/MockSamDAO.scala
```diff
@@ -385,6 +390,16 @@ class MockSamDAO extends SamDAO[IO] {
     ev: Ask[IO, TraceId]
   ): IO[Option[UserSubjectId]] = IO.pure(None)
 
+  def addUserToProject(creatorEmail: WorkbenchEmail, googleProject: GoogleProject)(implicit
+    ev: Ask[IO, TraceId]
+  ): IO[Unit] = {
+    val authHeader = userEmailToAuthorization(creatorEmail)
+    val project = ProjectSamResourceId(googleProject)
+    IO {
+      projectUsers = projectUsers |+| Map(authHeader -> Set((project, SamPolicyName.User)))
```
mlilynolting, line 399: what the heck is the TIE fighter operator?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/SamAuthProvider.scala
mlilynolting: (I asked in #scala and learned the awful truth)

lucymcnatt: yeahhhh in this case it refers to the first element of an item in the list of resource policies, which should be the samResourceId

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/SamAuthProvider.scala
lucymcnatt: yeah basically

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/SamAuthProvider.scala
mlilynolting: would it be possible to make that more explicit? maybe something like `res.exists { case (visibleResource, _) => visibleResource == r }`?

lucymcnatt: hmm I can make r more descriptive if that helps, change the case statements to 
`case (resourceId, policyName)`
and
`case (project, resourceId)`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskServiceInterp.scala
lucymcnatt: so the traverse here basically means that the following only executes if `cloudContext` isn't `None`, and it evaluates to `None` otherwise. So `hasProjectPermission` can be None, true or false

The next bit `!hasProjectPermission.getOrElse(true)` ensures that if hasProjectPermission is:
- true --> no error is raised
- false -->  error is raised
- None --> no error is raised, then yeah the rest of the function decides which disks the user can see based on resource and project permissions

The goal of this bit is when a user is listing by project, they are immediately rejected if they don't have project permission (they shouldn't even see an empty list)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskServiceInterp.scala
lucymcnatt: I'd guess because it contains a labelMap
image

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/SamAuthProvider.scala
mlilynolting: Every variable changed from a mysterious acronym to a descriptive name makes me happier so that would be great tyty😅

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskServiceInterp.scala
mlilynolting: Isn't there a use case in Terra for listing all visible disks independent of cloudContext? That's what the Environments page does, right?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskServiceInterp.scala
lucymcnatt: in what situation would you have a disk in Leo and not in Sam? if it errored on creation or something? I think in that case the user wouldn't be able to see the disk because it wouldn't be in the permissions from the sam check. If the user was the project owner they could see it, but not at the reader level (but I would have to check)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskServiceInterp.scala
mlilynolting: a bit misleading😢

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
lucymcnatt: I ended up moving this logic to the function 1 step above this so that I could use the useGoogleProject bool I added

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskServiceInterp.scala
mlilynolting: I'm not sure you can get into this state with Disks, but you definitely can with Runtimes. That's why the getRuntime method has the ugly check for user-is-creator; Leo keeps the record around after Sam has no more knowledge of it

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskServiceInterp.scala
lucymcnatt: yeah so when `cloudContext` is `None`, then it doesn't execute the traverse (ie it doesn't do the auth check) and `hasProjectPermission` will therefore equal `None`, the exception won't be raised and the rest of the function executes like nothing happened

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskServiceInterp.scala
lucymcnatt: I agree there, lots of misleading names in Leo lol

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskServiceInterp.scala
lucymcnatt: hmmm that's a case I didn't know about, so when we notify a resource as deleted to Sam it removes it? In that case the time I could see this being an issue is a project reader using the `includeDeleted` flag and not seeing previous runtimes they've created and deleted

mlilynolting: Yeah - although this might not apply to v1 resources that Leo manages directly through Google's API. I know that for WSM-mediated resources, WSM cleans up (incl. removing from Sam) and Leo retains the record, which will now scan in Sam as forbidden to the creator.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
lucymcnatt: I ended up moving this logic, so it now looks like
```
listApp(...) {
      ...
      partition = allClusters.partition(_.cloudContext.isInstanceOf[CloudContext.Gcp])
      gcpApps <- filterAppsBySamPermission(partition._1, userInfo, paramMap._3, true)
      azureApps <- filterAppsBySamPermission(partition._2, userInfo, paramMap._3, false)
    } yield gcpApps ++ azureApps
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeServiceInterp.scala
lucymcnatt: - if a creator leaves a workspace, they will no longer be able to access any of that workspace's (project's) resources, which is the intended behavior behind this security patch
- the behavior for workspace/project readers and owners shouldn't change
- if they're a reader they should only be able to access resources they have read+ permission on in Sam
- if they're an owner they should be able to access everything in the workspace/project

I'm not sure what error cases you mean here, an errored app/runtime should have the same permissions as a running one I think? And a resource is only set as deleted in Sam after its transitioned to a 'Deleted' status in Leo

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/dao/MockSamDAO.scala
lucymcnatt: TBH I copied that bit of logic from another place in MockSamDAO, so I'm not entirely sure myself

lucymcnatt: but I think it functions as kind of a += in this case

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeServiceInterp.scala
mlilynolting: Nice, I think your final clarification is what I needed. v1 resources just don't have the sync issues that WSM-managed ones do

mlilynolting (approved):

PR #3582, opened 2023-08-15
Fix dev issue: explicitly set array values

rtitle:
This fixes the Leo issue currently on dev. Testing in a BEE. See related helmfile PR: https://github.com/broadinstitute/terra-helmfile/pull/4436

Not the most elegant fix, but I can't see anything better right now.

Before, terra-helmfile would set EVs like this (with dots in the name):
```
VALID_HOSTS.2=bvdp-saturn-dev.appspot.com
VALID_HOSTS.1=*
VALID_HOSTS.0=localhost:3000
```
which let us define list-valued HOCON configs like this in Leo (this behavior is mostly undocumented):
```
validHosts = ${?VALID_HOSTS}
```

Unfortunately, dots in EV names aren't really supported in many Unix shells like bash, and it broke in a recent Debian base image upgrade. Here's a quick [repro](https://gist.github.com/rtitle/c2022672a6f0007b73517717682a08b9).

So instead, we now just pass each array element as a string, as suggested in https://stackoverflow.com/questions/50133291/how-to-provide-an-array-of-values-as-an-env-variable-to-typesafe-lightbend-confi. So we have EVs like:
```
VALID_HOSTS_2=bvdp-saturn-dev.appspot.com
VALID_HOSTS_1=*
VALID_HOSTS_0=localhost:3000
```
and read them in Leo like:
```
validHosts = [
  ${?VALID_HOSTS_0},
  ${?VALID_HOSTS_1},
  ${?VALID_HOSTS_2},
  …
]
...
```
I made "space" for 20 but the config library will only read the ones defined by EVs.

### DIFF
--- a/http/src/main/resources/leo.conf
+++ b/http/src/main/resources/leo.conf
@@ -65,53 +65,57 @@ azure.coa-app-config.dockstore-base-url = ${?DOCKSTORE_BASE_URL}
 azure.wds-app-config.instrumentation-enabled = ${?WDS_INSTRUMENTATION_ENABLED}
 
 contentSecurityPolicy {
-  frameAncestors.0 = ${?FRAME_ANCESTORS_0}
-  frameAncestors.1 = ${?FRAME_ANCESTORS_1}
-  frameAncestors.2 = ${?FRAME_ANCESTORS_2}
-  frameAncestors.3 = ${?FRAME_ANCESTORS_3}
-  frameAncestors.4 = ${?FRAME_ANCESTORS_4}
-  frameAncestors.5 = ${?FRAME_ANCESTORS_5}
-  frameAncestors.6 = ${?FRAME_ANCESTORS_6}
-  frameAncestors.7 = ${?FRAME_ANCESTORS_7}
-  frameAncestors.8 = ${?FRAME_ANCESTORS_8}
-  frameAncestors.9 = ${?FRAME_ANCESTORS_9}
-  frameAncestors.10 = ${?FRAME_ANCESTORS_10}
-  frameAncestors.11 = ${?FRAME_ANCESTORS_11}
-  frameAncestors.12 = ${?FRAME_ANCESTORS_12}
-  frameAncestors.13 = ${?FRAME_ANCESTORS_13}
-  frameAncestors.14 = ${?FRAME_ANCESTORS_14}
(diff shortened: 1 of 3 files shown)


rtitle:

davidangb (approved):
Is there any issue with the extra commas, if you end up with e.g. 
```
validHosts = ['localhost:3000', '*', 'bvdp-saturn-dev.appspot.com', , , , , , , , ]
```

Giving a :+1:, did not test this myself.

rtitle:
@davidangb I was wondering that, I think lightbend config strips nulls automatically. Adding a println with dev configs, I get `localhost:3000,*,bvdp-saturn-dev.appspot.com`.

LizBaldo (approved):
Whoa, interesting find 🤯 Would it make sense to link a comment in terra helm-file pointing to the leo config, saying that the list needs to be expended there if we reach more than 20 Valid hosts EV?

mlilynolting (approved):

em-may:
Fyi that this will break local dev. Please apply the following diff to fix:

```
diff --git a/local/overrides.env b/local/overrides.env
index 713d93d19..87b3eef44 100644
--- a/local/overrides.env
+++ b/local/overrides.env
@@ -31,7 +31,7 @@ SHOULD_INIT_WITH_LIQUIBASE=false
 JAVA_OPTS=-DXmx4G -DXms4G -DXss2M -Dsun.net.spi.nameservice.provider.1=default -Dsun.net.spi.nameservice.provider.2=dns,Jupyter -Djna.library.path=${HELM_BUILD_DIR} -Duser.timezone=UTC -Dapp-service.enable-custom-app-check=false -Dconfig.resource=leo.conf

 # SBT
-SBT_OPTS=-Xmx4G -Xms4G -Xss2M -XX:+UseG1GC -DVALID_HOSTS.0=local.dsde-dev.broadinstitute.org
+SBT_OPTS=-Xmx4G -Xms4G -Xss2M -XX:+UseG1GC -DVALID_HOSTS_0=local.dsde-dev.broadinstitute.org

 # SSL certs and SA key
 ROOT_CA_PEM_PATH=${RENDER_DIR}/rootCA.pem
```

rtitle:
Keeping this PR, but exploring an alternative way with @em-may that may not require 20 placeholder fields in Leo conf.

rtitle:
Closing this in favor of better helmfile fix

PR #3590, opened 2023-08-15
[IA-4495] Remove redundant automation tests

jdcanas:
This PR attempts to chip away at redundant/unnecessary automation tests. The rule of thumb I used for the majority of these removals is:
1. The removed test is covered elsewhere (mainly terra-docker)
2. OR the removed test is not providing much value / not fixable in leo if there is a failure

For example, `NotebookGatkSpec` is duplicated 100% by this test in terra-docker: https://github.com/DataBiosphere/terra-docker/blob/master/terra-jupyter-gatk/tests/gatk_smoke_test.ipynb

This is by no means an exhaustive cleanup, but I'm hoping we can start seeing some wins right away.


automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookGCECustomizationSpec.scala
```diff
@@ -26,72 +26,6 @@ final class NotebookGCECustomizationSpec
...
-
-                // Exercise the translate extensionfailure_screenshots/NotebookGCECustomizationSpec_18-34-37-017.png
-                notebookPage.translateMarkup("Yes") should include("Oui")
-              }
-            }
-          }
-      }
-    }
-
-    "should give cluster user-specified scopes" in { billingProject =>
```
jdcanas, line 66: I removed this test despite not being covered elsewhere because it seems to be testing a feature of the google cloud, @rtitle  thoughts?

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookPyKernelSpec.scala
```diff
@@ -38,48 +38,6 @@ class NotebookPyKernelSpec extends RuntimeFixtureSpec with NotebookTestUtils {
...
-            val getFuzzyWuzzyLocation = "! pip3 show fuzzywuzzy"
-            notebookPage.executeCell(getFuzzyWuzzyLocation, cellNumberOpt = Some(1)).get should include(
-              "/home/jupyter/.local/lib/python3.10/site-packages"
-            )
-          }
-        }
-      }
-    }
-
-    "should NOT be able to run Spark" in { runtimeFixture =>
```
jdcanas, line 64: This seems pretty unnecessary

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookPyKernelSpec.scala
```diff
@@ -38,48 +38,6 @@ class NotebookPyKernelSpec extends RuntimeFixtureSpec with NotebookTestUtils {
         }
     }
 
-    Seq(Python3).foreach { kernel =>
```
LizBaldo, line 41: I did not realize we tested that 'pip' works! I do hope that pip has some tests internally 😅

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookPyKernelSpec.scala
LizBaldo: Agreed, the only use case I see would be to make sure that hail can be imported on the dataproc runtime, since that cannot be tested in terra-docker

LizBaldo (approved):
Good find, that's a satisfactory cleanup. One thing to consider though might be to move these tests into a test-notebook that we can run manually from time to time to make sure that bit upgrades don't break basic functionalities (e.g. I would have run this to test the 3.10 upgrade)

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookGCECustomizationSpec.scala
rtitle: I support this -- feels like it could be unit tested too (Leo reads the scopes from the API request and passes it to Google).

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookPyKernelSpec.scala
rtitle: Sigh 😅

rtitle (approved):
This all looks good to me -- yeah any automation test that is using webdriver to test the contents of the image probably does not need to exist anymore.

PR #3603, opened 2023-08-17
[WM-2102][WM-2159] Create Workflows App with Workload Identity

dillydally414:
__Jira tickets__: 
- https://broadworkbench.atlassian.net/browse/WM-2102
- https://broadworkbench.atlassian.net/browse/WM-2159

### What

- Deploying Workflows App with shared workload identity
- Deploying Workflows App databases with kubernetes service account

### Why

- Creating new app types in Leo to support multi-user collaboration
- Workload identity preferred over pod identity
- Shared app should used shared service account

## Testing these changes

<details>
  <summary>Successful workflow app creation in bee</summary>

  ```json
  {
    "appName": "workflows-test-app12",
    "cloudContext": {
      "cloudProvider": "AZURE",
      "cloudResource": "fad90753-2022-4456-9b0a-c7e5b934e408/62b22893-6bc1-46d9-8a90-806bb3cce3c9/mrg-terra-dev-previ-20230817150542"
    },
    "kubernetesRuntimeConfig": {
      "numNodes": 1,
      "machineType": "Standard_A2_v2",
      "autoscalingEnabled": false
    },
    "errors": [],
    "status": "RUNNING",
    "proxyUrls": {
      "cbas": "https://lzd0510bd6277a20a5fa4c7348da58aba7d26e295bbe59a960.servicebus.windows.net/workflows-test-app12-8f99bc2b-c665-467f-a504-769bb99722fb/cbas",
      "cromwell-reader": "https://lzd0510bd6277a20a5fa4c7348da58aba7d26e295bbe59a960.servicebus.windows.net/workflows-test-app12-8f99bc2b-c665-467f-a504-769bb99722fb/cromwell",
      "listener": "https://lzd0510bd6277a20a5fa4c7348da58aba7d26e295bbe59a960.servicebus.windows.net/workflows-test-app12-8f99bc2b-c665-467f-a504-769bb99722fb/listener"
    },
    "diskName": null,
      "RELAY_HYBRID_CONNECTION_NAME": "workflows-test-app12-8f99bc2b-c665-467f-a504-769bb99722fb"
    },
    "auditInfo": {
      "creator": "dscott.tester@gmail.com",
      "createdDate": "2023-08-23T21:14:43.972271Z",
      "destroyedDate": null,
      "dateAccessed": "2023-08-23T21:14:43.972271Z"
    },
    "appType": "WORKFLOWS_APP",
    "accessScope": "WORKSPACE_SHARED",
    "labels": {
      "cloudContext": "fad90753-2022-4456-9b0a-c7e5b934e408/62b22893-6bc1-46d9-8a90-806bb3cce3c9/mrg-terra-dev-previ-20230817150542",
      "appName": "workflows-test-app12",
      "clusterServiceAccount": "/subscriptions/62b22893-6bc1-46d9-8a90-806bb3cce3c9/resourcegroups/mrg-terra-dev-previ-20230817150542/providers/Microsoft.ManagedIdentity/userAssignedIdentities/pet-26922994279819c684337",
      "creator": "dscott.tester@gmail.com"
    }
  }
  ```
</details>

### What to test

- Creating a new app in an existing workspace (in [my bee](https://leonardo.scottdil-test.bee.envs-terra.bio/#/apps/createAppV2)) with the following body:
```json
{
  "appType": "WORKFLOWS_APP",
  "accessScope": "WORKSPACE_SHARED"
}
```

### Who tested and where

- [X] This change is covered by automated tests - unit 
- [X] I validated this change in [my bee](https://leonardo.scottdil-test.bee.envs-terra.bio/#/apps/createAppV2)
- [X] Primary reviewer validated this change 
  - @kpierre13 was able to create a workflows app in her own workspace as well as view a workflows app I had created in a shared workspcae


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -1104,19 +1238,19 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
     }
   }
 
-  private[util] def maybeCreateWsmIdentityAndDatabase(app: App,
-                                                      workspaceId: WorkspaceId,
-                                                      landingZoneResources: LandingZoneResources,
-                                                      namespace: KubernetesNamespace
+  private[util] def maybeCreateWsmIdentityAndDatabases(app: App,
```
dillydally414: This function was previously only used to create an identity and database for WDS (in LZ postgres), but since we want to create Workflows-App in the same manner it has been generalized to create an identity and database(s) depending on the app type, and should be flexible to new apps with shared databases.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -1104,19 +1238,19 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
...
   )(implicit
     ev: Ask[F, AppContext]
-  ): F[(Option[ServiceAccountName], Option[String])] = {
-    val databaseConfigEnabled = app.appType match {
-      case AppType.Wds => config.wdsAppConfig.databaseEnabled
-      case _           => false
+  ): F[(Option[ServiceAccountName], Option[Map[String, String]])] = {
+    val shouldCreateLZDatabases = app.appType match {
+      case AppType.Wds          => landingZoneResources.postgresServer.isDefined && config.wdsAppConfig.databaseEnabled
+      case AppType.WorkflowsApp => true
```
dillydally414, line 1230: New app types should always be deployed with workload identity and LZ postgres, as they don't have the ability to create postgres deployments via helm

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -1165,14 +1299,29 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
           )
           .transaction
 
-        dbName <- createDatabaseInWsm(app,
+        dbNamePrefixes = app.appType match {
```
dillydally414, line 1283: Generalized to a list of database prefixes rather than hardcoded wds, led to some tricky stuff as far as mapping the (async) database creation across the list.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -1104,19 +1238,19 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
...
+                                                       workspaceId: WorkspaceId,
+                                                       landingZoneResources: LandingZoneResources,
+                                                       namespace: KubernetesNamespace
   )(implicit
     ev: Ask[F, AppContext]
-  ): F[(Option[ServiceAccountName], Option[String])] = {
-    val databaseConfigEnabled = app.appType match {
-      case AppType.Wds => config.wdsAppConfig.databaseEnabled
-      case _           => false
+  ): F[(Option[ServiceAccountName], Option[Map[String, String]])] = {
```
dillydally414: The function now returns a map from database name prefix to database name, rather than a single database name, since workflows app creates two databases on deployment.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/BuildHelmChartValues.scala
```diff
@@ -314,10 +314,11 @@ private[leonardo] object BuildHelmChartValues {
...
   ): Values = {
     val relayTargetHost = appType match {
-      case AppType.Cromwell  => s"http://coa-${release.asString}-reverse-proxy-service:8000/"
-      case AppType.Wds       => s"http://wds-${release.asString}-wds-svc:8080"
-      case AppType.HailBatch => "http://batch:8080"
-      case _                 => "uknown"
+      case AppType.Cromwell     => s"http://coa-${release.asString}-reverse-proxy-service:8000/"
+      case AppType.Wds          => s"http://wds-${release.asString}-wds-svc:8080"
+      case AppType.HailBatch    => "http://batch:8080"
+      case AppType.WorkflowsApp => s"http://workflows-app-${release.asString}-reverse-proxy-service:8000/"
```
dillydally414: Provide relay listener with the reverse proxy uri

http/src/main/resources/reference.conf
```diff
@@ -306,11 +306,11 @@ azure {
       {
         name = "cromwell-reader"
         kind = "ClusterIP"
+        path = "/cromwell"
```
dillydally414, line 309: Health/status check assumes all cromwell services are located at `/cromwell`, rather than creating a new service type and status checker and unnecessary boilerplate, put this service at `/cromwell`. Proxy URL will look something like:
`"cromwell-reader": "https://LZ.servicebus.windows.net/<appName>-<uuid>/cromwell"`

http/src/main/resources/reference.conf
```diff
@@ -306,11 +306,11 @@ azure {
...
       }
     ]
     # Not enabled to make sure people cannot start the (not yet complete) app:
     enabled = false
```
dillydally414, line 313: Should this be enabled in this PR? If not, do we have a ticket to enable it?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -116,12 +116,13 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
...
 
-      // If configured for the app type, call WSM to create a managed identity and postgres database.
-      // This returns a KSA authorized to access the database.
-      (maybeKsaFromDatabaseCreation, maybeDbName) <- maybeCreateWsmIdentityAndDatabase(app,
-                                                                                       params.workspaceId,
-                                                                                       params.landingZoneResources,
-                                                                                       kubernetesNamespace
+      // If configured for the app type, call WSM to create a managed identity and postgres database(s).
+      // This returns a KSA authorized to access the database(s).
+      (maybeKSAFromSharedDatabaseCreation, maybeSharedDbNames) <- maybeCreateWsmIdentityAndDatabases(
```
dillydally414: This is where the identity and databases are created. Scroll down to see the body of this function

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -322,6 +326,52 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
...
+        case AppType.WorkflowsApp =>
+          for {
+            // Get the batch account key
+            batchAccount <- azureBatchService.getBatchAccount(params.landingZoneResources.batchAccountName,
+                                                              params.cloudContext
+            )
+
+            batchAccountKey = batchAccount.getKeys().primary
+
+            // Storage container is required for Workflows app
```
dillydally414, line 341: I don't actually know if this is true. This part was taken from CoA chart creation.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -962,6 +1012,90 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
...
+        raw"provenance.userAccessToken=${userAccessToken}"
+      )
+
+    val postgresConfig = (maybeDatabaseNames, landingZoneResources.postgresServer, ksaName) match {
+      case (Some(dbNames), Some(PostgresServer(dbServer, pgBouncerEnabled)), Some(ksa)) =>
+        List(
+          raw"postgres.host=$dbServer.postgres.database.azure.com",
+          raw"postgres.pgbouncer.enabled=$pgBouncerEnabled",
+          // convention is that the database user is the same as the service account name
+          raw"postgres.user=${ksa.value}",
```
dillydally414: database user is shared KSA

dillydally414 (commented):
Left some initial explanatory comments on important takeaways/changes

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
dillydally414: Did more digging and it looks like this is used more on the cromwell-runner side of things, so it may be safe to not include here. Would appreciate a sanity check

kpierre13 (approved):

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/KubernetesAppConfig.scala
```diff
@@ -139,8 +138,8 @@ final case class WorkflowsAppConfig(chartName: ChartName,
   def workflowsAppServices: Set[WorkflowsAppService] = services
```
rtitle: Consider if the `WorkflowsAppService` enum is needed anymore? It was added early on when we wanted to pick and choose which services were deployed via the `CROMWELL` app type. Nowadays, a given `WORKFLOWS_APP` will also include both `cbas` and `cromwell-reader` services, right? So maybe we don't need this level of granularity anymore.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoMetricsMonitor.scala
```diff
@@ -173,11 +173,12 @@ class LeoMetricsMonitor[F[_]](config: LeoMetricsMonitorConfig,
...
                   .unsafeFromString(baseUri.asString) / s"${app.appName.value}-${app.workspaceId.map(_.value.toString).getOrElse("")}"
                 isUp <- serviceName match {
-                  case ServiceName("cbas")     => cbasDAO.getStatus(relayPath, authHeader).handleError(_ => false)
-                  case ServiceName("cbas-ui")  => cbasUiDAO.getStatus(relayPath, authHeader).handleError(_ => false)
-                  case ServiceName("cromwell") => cromwellDAO.getStatus(relayPath, authHeader).handleError(_ => false)
-                  case ServiceName("wds")      => wdsDAO.getStatus(relayPath, authHeader).handleError(_ => false)
-                  case ServiceName("batch")    => hailBatchDAO.getStatus(relayPath, authHeader).handleError(_ => false)
+                  case ServiceName("cbas")    => cbasDAO.getStatus(relayPath, authHeader).handleError(_ => false)
+                  case ServiceName("cbas-ui") => cbasUiDAO.getStatus(relayPath, authHeader).handleError(_ => false)
+                  case ServiceName("cromwell") | ServiceName("cromwell-reader") =>
```
rtitle, line 178: It looked like `cromwell-writer` was a possible service name too?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -962,6 +1011,79 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
       ).mkString(",")
     )
 
+  private[util] def buildWorkflowsAppChartOverrideValues(release: Release,
```
rtitle, line 1008: Could consider splitting into the [BuildHelmChartValues](https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/BuildHelmChartValues.scala) file

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
rtitle: Is this method specific to shared databases? Maybe we should name it as such: `maybeCreateSharedDatabases` or something.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
rtitle: Agree about the app requiring LZ-postgres (we are not trying to support old LZs with no postgres server).

Idly wondering what would happen if you deployed `WORKFLOWS_APP` in an old LZ _without_ postgres -- I assume it would blow up somehow.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
rtitle: Another note: this implementation might change (not sure exactly how) once Leo adopts https://github.com/DataBiosphere/terra-workspace-manager/pull/1430, which changes some of the WSM APIs used for creating identities, databases, and namespaces.

http/src/main/resources/reference.conf
rtitle: Not sure -- you could also add an EV to `leo.conf` like
```
enabled = ${?WORKFLOWS_APP_ENABLED}
```
 so we could at least turn it on in lower environments (dev, bees, etc).

rtitle (approved):
I didn't test but this looks great, nice work. Left some comments/feedback but overall 👍

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
dillydally414: Is `maybeCreateWsmIdentityAndSharedDatabases` too much of a mouthful? I think it's important to note the identity is created within this method as well

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
dillydally414: I'm not sure how to easily test this but I think it would error out in WSM during the attempted database creation.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
rtitle: That sounds fine… mainly I was just confused what this method was doing until I realized it handle all our shared database creation. Whereas `maybeCreateCromwellDatabases` creates all the user-private DBs needed by cromwell apps.

http/src/main/resources/reference.conf
rtitle: OTOH you could just enable it, which seems to be happening for Cromwell Runner in https://github.com/DataBiosphere/leonardo/pull/3621 :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoMetricsMonitor.scala
dillydally414: Yes, I believe that will be taken care of in #3621

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/KubernetesAppConfig.scala
dillydally414: What are you suggesting to replace the usages of `workflowsAppServices` with?

http/src/main/resources/reference.conf
salonishah11: I like the idea of setting this to true for Dev and BEEs via terra-helmfile charts. This way users can't unexpectedly create these new apps in Prod but we can easily test it in Dev or BEEs.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/KubernetesAppConfig.scala
rtitle: I didn't find all references.. but if it's just being used for `config.workflowsAppConfig.workflowsAppServices.contains(Cbas)`, etc… and it always contains the same list, we could just remove the method.

(Not a big deal, it was just added as a temporary thing which kept getting carried forward :))

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
dillydally414: Looked into this, may be more trouble than it's worth since the values depend on the AKSInterpreter config values, which would need to be passed in and then mocked out again in a different test file since they're already mocked out in AKSInterpreterSpec

http/src/main/resources/reference.conf
dillydally414: Agreed, thanks both!

dillydally414: https://github.com/broadinstitute/terra-helmfile/pull/4527

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/KubernetesAppConfig.scala
dillydally414: Made some changes in my latest commit, let me know if this looks good now?

rtitle (approved):
Thanks, latest change looks good

…although a unit test seems to have failed.

http/src/main/resources/leo.conf
```diff
@@ -161,7 +161,7 @@ azure {
 
   workflows-app-config {
     instrumentation-enabled = ${?WORKFLOWS_APP_INSTRUMENTATION_ENABLED}
-    database-enabled  = ${?WORKFLOWS_APP_DATABASE_ENABLED}
```
cjllanwarne, line 167: Nice catch

http/src/main/resources/reference.conf
cjllanwarne: Do we need an equivalent path for cbas?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoMetricsMonitor.scala
cjllanwarne: TOL I wonder whether just using "cromwell" for both makes sense, but not enough to request that in this PR

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/BuildHelmChartValues.scala
```diff
@@ -314,10 +314,11 @@ private[leonardo] object BuildHelmChartValues {
...
   ): Values = {
     val relayTargetHost = appType match {
-      case AppType.Cromwell  => s"http://coa-${release.asString}-reverse-proxy-service:8000/"
-      case AppType.Wds       => s"http://wds-${release.asString}-wds-svc:8080"
-      case AppType.HailBatch => "http://batch:8080"
-      case _                 => "uknown"
+      case AppType.Cromwell     => s"http://coa-${release.asString}-reverse-proxy-service:8000/"
+      case AppType.Wds          => s"http://wds-${release.asString}-wds-svc:8080"
+      case AppType.HailBatch    => "http://batch:8080"
+      case AppType.WorkflowsApp => s"http://wfa-${release.asString}-reverse-proxy-service:8000/"
```
cjllanwarne: Can/should we combine the cases with a `|`? Happy to leave it as-is though

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/ConfigReaderSpec.scala
```diff
@@ -140,17 +140,19 @@ class ConfigReaderSpec extends AnyFlatSpec with Matchers {
         WorkflowsAppConfig(
           ChartName("/leonardo/workflows-app"),
           ChartVersion("0.24.0"),
-          ReleaseNameSuffix("workflows-app-rls"),
-          NamespaceNameSuffix("workflows-app-ns"),
-          KsaName("workflows-app-ksa"),
+          ReleaseNameSuffix("wfa-rls"),
```
cjllanwarne, line 143: I assume this is because we don't like the `-` in the name? Can we preserve a human readable name like `workflows` though?

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/ConfigReaderSpec.scala
```diff
@@ -140,17 +140,19 @@ class ConfigReaderSpec extends AnyFlatSpec with Matchers {
...
-            ServiceConfig(ServiceName("cromwell-reader"), KubernetesServiceKindName("ClusterIP"))
+            ServiceConfig(ServiceName("cromwell-reader"),
+                          KubernetesServiceKindName("ClusterIP"),
+                          Some(ServicePath("/cromwell"))
+            )
           ),
           instrumentationEnabled = false,
           enabled = false,
           dockstoreBaseUrl = new URL("https://staging.dockstore.org/"),
-          databaseEnabled = false,
```
cjllanwarne, line 153: Nice catch again

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
cjllanwarne: Yeah I don't think we need this. I'd be tempted to try deploying without it and seeing whether anything goes wrong. I guess we can also sanity check by seeing whether the cromwell deployment in the helm chart makes a persistence claim. 

EDIT: maybe we do use it? (but maybe we shouldn't, these all seem very run-specific)
- https://github.com/broadinstitute/terra-helmfile/blob/master/charts/workflows-app/templates/cromwell.yaml#L66
- https://github.com/broadinstitute/terra-helmfile/blob/master/charts/workflows-app/templates/cbas-api-config.yaml#L26
- https://github.com/broadinstitute/terra-helmfile/blob/master/charts/workflows-app/templates/cromwell-config.yaml#L88

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
cjllanwarne: If you feel inclined you could make this a little safer (and maybe easier to use?) by having an ADT something like 

```scala
sealed trait SharedDatabaseNames
case class WdsDatabaseName(String wdsDbName) extends SharedDatabaseNames
case class WorkflowsDatabaseNames(String cbasDbName, String cromwellDbName) extends SharedDatabaseNames
```

and choosing which type to return as appropriate (might be overkill though 😄 )

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
cjllanwarne: Is there an easy way to find out? Might be nice (for us in the near-future!) to fail this early rather than late in an inappropriate LZ

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -1165,14 +1279,29 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
...
                                       workspaceId,
                                       namespace,
-                                      "wds",
+                                      databaseNamePrefix,
                                       wsmApi,
                                       Option(createIdentityResponse.getResourceId)
-        )
-      } yield (Some(ServiceAccountName(identityName)), Some(dbName))
+                  )
+              } yield databaseNamePrefix -> databaseName
```
cjllanwarne: My personal heuristic is that a`for` inside another `for` is often a sign that you might want to extract a sub-function

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
cjllanwarne: Hmm... it looks like this might [already exist](https://github.com/DataBiosphere/leonardo/pull/3603/files#diff-3d6a73abda025b49d1c9b3169828843be606585cfaeb46b798e2a567bbe1fc47R1533)?

cjllanwarne (approved):
LGTM, more TOL level comments than blockers

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/BuildHelmChartValues.scala
dillydally414: My thought was (at least internally) it's nice to differentiate coa from wfa (and cra which Saloni will introduce) via prefix, even if they both use reverse proxy service.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
dillydally414: If anyone has access to an old LZ in bee/dev they could grant me access too I would be more than happy to test it out (though on dev would have to be a post-merge test)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
dillydally414: Yep, I actually convert it into that type when passing into the chart creation. Main reason I didn't was because WDS only has one database and I wasn't sure how to dynamically do this assignment. I think I have a marginally better understanding of Scala now than I did at the start of last week though, so I may give this another shot.

http/src/main/resources/reference.conf
dillydally414: Nope, it will default to `/<name>` AKA `/cbas`

rtitle: If there’s no path defined in config it defaults to the name of the service.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoMetricsMonitor.scala
dillydally414: Had the same thought while working through this. I think one benefit I see of having them labeled differently is that it's easier to pick out whether it's a reader or writer from a proxy URL object, since they both will end in `/cromwell` but would have different keys. Definitely open to changing this but I think it might change more than just this PR as well since there are related CBAS changes already in progress.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
dillydally414: Hmm, I do see the workflows app using the storage account but this may be tied into the question of "what configs does a reader cromwell instance actually need to run properly". I don't see the workflows app using the batch account key.

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/ConfigReaderSpec.scala
dillydally414: This was actually because this led to super long names like `workflows-app-mkcjzg-workflows-app-rls-cbas-depl-78ccb5fd4d9r7v`. Open to other ideas. I think Saloni is planning on shortening hers to cra.

salonishah11: Yeah I am open to suggestions too for Cromwell Runner app since having `cromwell-runner-app` appear twice in names for making it go over 63 character limit and throwing an error.

dillydally414:
jenkins retest

dillydally414:
jenkins retest

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -1288,20 +1288,22 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
...
+                                  wsmApi,
+                                  Option(createIdentityResponse.getResourceId)
+              )
             )
-            .map(list => list.toMap)
 
-      } yield (Some(ServiceAccountName(identityName)), Some(dbNames))
+        typedDbNames = app.appType match {
+          case AppType.Wds          => Some(WdsDatabaseNames(dbNames.head))
+          case AppType.WorkflowsApp => Some(WorkflowsAppDatabaseNames(dbNames.head, dbNames.apply(1)))
```
cjllanwarne, line 1304: Probably not a problem here... but making assumptions about how long a list will be can sometimes backfire. I wonder whether any kind of validation on the size makes sense? Or is that overly paranoid?

cjllanwarne (approved):

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
dillydally414: Tested in a non-postgres LZ - App fails pretty quickly:
```
Error creating Azure app with id 4161 and cloudContext fad90753-2022-4456-9b0a-c7e5b934e408/f557c728-871d-408c-a28b-eb6b2141a087/devTestingMrg: WSM database creation failed for app workflows-test-app. WSM response:
class CreatedControlledAzureDatabaseResult {
   resourceId: null
   azureDatabase: null
   jobReport: class JobReport {
        id: cbas_qyyomd
       description: Create controlled resource CONTROLLED_AZURE_DATABASE; id d327dbea-e39e-4b8e-a7e2-f742e02669dc; name cbas_qyyomd
       status: FAILED
       statusCode: 500
       submitted: 2023-08-31T13:32:13.466407Z
       completed: 2023-08-31T13:32:14.850230Z
       resultURL: https://workspace.dsde-dev.broadinstitute.org/api/workspaces/v1/a65bc4c6-a077-45e0-9a67-b31aec55243e/resources/controlled/azure/databases/cbas_qyyomd/create-result
   }
   errorReport: class ErrorReport {
       message: No shared database found
       statusCode: 500
       causes: []
    }
}
```

PR #3618, opened 2023-08-22
[IA-4307] adjust timeouts

jdcanas:
This PR addresses one main thing:

-It ensures leo waits longer than the WSM job timeout plus some processing time, to ensure a WSM job failure and not a leo timeout issue is saved with a failed WSM job

This should address the failure mode where WSM job is still processing, and leo then tries to clean a disk that is still 'attached'. We still need to submit the disk delete request, but we need to ensure the VM step has 'failed' and can be unrolled. This assumes that the disk job will have terminated within the same 30minutes. 

WSM timeout configured here: https://github.com/DataBiosphere/terra-workspace-manager/blob/main/service/src/main/resources/application.yml#L77


LizBaldo (approved):
LGTM, thanks for adding the links and documentation to the code, this is super clear!

jdcanas:
jenkins retest

lucymcnatt (approved):

jdcanas:
jenkins retest

jdcanas:
jenkins retest

PR #3621, opened 2023-08-22
[WM-2158][WM-2160] Ability to deploy Cromwell Runner app with databases

salonishah11:
__Jira tickets__: 
- https://broadworkbench.atlassian.net/browse/WM-2158
- https://broadworkbench.atlassian.net/browse/WM-2160

Related PR in terra-helmfile that enables Cromwell Runner app for Dev and BEE envs: https://github.com/broadinstitute/terra-helmfile/pull/4540

### What

- Deploying new Cromwell Runner app and its databases using Workfload identity that is scoped to user

### Why

- To support multi-user workflow collaboration

## Testing these changes

Tested it in a BEE and Cromwell Runner app with Cromwell and TES pods was successfully deployed with its respective databases. TES pod is failing to start because of a known issue of current image not working in shared LZ. The new TES image is going to be released soon which should fix the issue.

Pods:
```
% kubectl get pods -n bqwen0-cra-ns
NAME                                                       READY   STATUS             RESTARTS      AGE
cra-bqwen0-cra-rls-cromwell-depl-8794964cf-c7dgr           1/1     Running            0             6m47s
cra-bqwen0-cra-rls-reverse-proxy-deploy-6c74779dc4-5kspf   1/1     Running            0             6m47s
cra-bqwen0-cra-rls-tes-depl-5cc85fb67b-bvcwn               0/1     CrashLoopBackOff   5 (41s ago)   6m47s
listener-6cb58468db-j4cbl                                  1/1     Running            0             6m51s
```

Databases:
Screenshot 2023-08-30 at 4 55 25 PM

### Who tested and where

- [ ] I validated this change 


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoMetricsMonitor.scala
```diff
@@ -173,11 +173,12 @@ class LeoMetricsMonitor[F[_]](config: LeoMetricsMonitorConfig,
...
                   .unsafeFromString(baseUri.asString) / s"${app.appName.value}-${app.workspaceId.map(_.value.toString).getOrElse("")}"
                 isUp <- serviceName match {
-                  case ServiceName("cbas")     => cbasDAO.getStatus(relayPath, authHeader).handleError(_ => false)
-                  case ServiceName("cbas-ui")  => cbasUiDAO.getStatus(relayPath, authHeader).handleError(_ => false)
-                  case ServiceName("cromwell") => cromwellDAO.getStatus(relayPath, authHeader).handleError(_ => false)
-                  case ServiceName("wds")      => wdsDAO.getStatus(relayPath, authHeader).handleError(_ => false)
-                  case ServiceName("batch")    => hailBatchDAO.getStatus(relayPath, authHeader).handleError(_ => false)
+                  case ServiceName("cbas")    => cbasDAO.getStatus(relayPath, authHeader).handleError(_ => false)
+                  case ServiceName("cbas-ui") => cbasUiDAO.getStatus(relayPath, authHeader).handleError(_ => false)
+                  case ServiceName("cromwell") | ServiceName("cromwell-writer") =>
```
rtitle: Ah, https://github.com/DataBiosphere/leonardo/pull/3603 adds `cromwell-reader`. I suspect we want all 3 (cromwell, cromwell-writer, cromwell-reader). 😄

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -130,12 +130,19 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
...
 
+      maybeCromwellRunnerDatabaseNames <- mayCreateCromwellRunnerDatabases(app, params.workspaceId, kubernetesNamespace)
+
       // Determine which type of identity to link to the app: pod identity, workload identity, or nothing.
-      identityType = (maybeKsaFromDatabaseCreation, app.samResourceId.resourceType, maybeCromwellDatabaseNames) match {
-        case (Some(_), _, _)                      => WorkloadIdentity
-        case (None, SamResourceType.SharedApp, _) => NoIdentity
-        case (None, _, Some(_))                   => WorkloadIdentity
-        case (None, _, _)                         => PodIdentity
+      identityType = (maybeKsaFromDatabaseCreation,
```
rtitle: Hopefully when we adopt https://github.com/DataBiosphere/terra-workspace-manager/pull/1430 we can get rid of this logic, and just use WorkloadIdentity across the board.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -863,6 +909,81 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
     Values((valuesList ++ postgresConfig).mkString(","))
   }
 
+  private[util] def buildCromwellRunnerChartOverrideValues(release: Release,
```
rtitle: Consider putting this method in the [BuildHelmChartValues](https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/BuildHelmChartValues.scala) file

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoMetricsMonitor.scala
salonishah11: Yes. I will wait for #3603 to merge first and then resolve this conflict

cromwell-runner-app-0.16.0.tgz
dillydally414, line 1: We should be able to remove this now

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -125,16 +126,18 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
         kubernetesNamespace
       )
 
-      maybeCromwellDatabaseNames <- maybeCreateCromwellDatabases(app,
-                                                                 params.workspaceId,
-                                                                 params.landingZoneResources,
-                                                                 kubernetesNamespace
+      // If configured for the app type, calls WSM to create a managed identity and postgres databases
+      // for CROMWELL and CROMWELL_RUNNER_APP app types
+      maybeSharedCromwellDatabaseNames <- maybeCreateCromwellDatabases(app,
```
dillydally414: I don't think we should call these "shared" since they are per-user, even if they are located in the shared LZ postgres instance

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -1331,10 +1381,37 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
...
+
+        createdDatabases = app.appType match {
+          case AppType.Cromwell =>
+            if (dbNames.size == 3) {
+              Some(CromwellDatabaseNames(dbNames.get(0).get, dbNames.get(1).get, dbNames.get(2).get))
+            } else None
+          case AppType.CromwellRunnerApp =>
+            if (dbNames.size == 2) {
+              Some(CromwellRunnerDatabaseNames(dbNames.get(0).get, dbNames.get(1).get))
+            } else None
```
dillydally414: @cjllanwarne and I discussed the length checking in #3603 and decided against it to have a louder failure if the length is wrong rather than returning None which would progress through further and fail more quietly

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -1532,7 +1609,9 @@ final case class AKSInterpreterConfig(
   listenerChartConfig: ListenerChartConfig
 )
 
-final case class CromwellDatabaseNames(cromwell: String, cbas: String, tes: String)
+sealed trait SharedCromwellDatabaseNames
```
dillydally414: Same comment with naming

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/BuildHelmChartValues.scala
```diff
@@ -404,4 +413,80 @@ private[leonardo] object BuildHelmChartValues {
...
+        raw"persistence.workspaceManager.url=${config.wsmConfig.uri.renderString}",
+        raw"persistence.workspaceManager.workspaceId=${workspaceId.value}",
+        raw"persistence.workspaceManager.containerResourceId=${storageContainer.resourceId.value.toString}",
+
+        // identity configs
+        raw"identity.enabled=${identityType == PodIdentity}",
+        raw"identity.name=${petManagedIdentity.map(_.name).getOrElse("none")}",
+        raw"identity.resourceId=${petManagedIdentity.map(_.id).getOrElse("none")}",
+        raw"identity.clientId=${petManagedIdentity.map(_.clientId).getOrElse("none")}",
+        raw"workloadIdentity.enabled=${identityType == WorkloadIdentity}",
```
dillydally414: I don't think we need these for cromwell-runner 🤔

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/BuildHelmChartValues.scala
```diff
@@ -404,4 +413,80 @@ private[leonardo] object BuildHelmChartValues {
...
+    val postgresConfig = (maybeDatabaseNames, landingZoneResources.postgresServer, petManagedIdentity) match {
+      case (Some(databaseNames), Some(PostgresServer(dbServerName, pgBouncerEnabled)), Some(pet)) =>
+        List(
+          raw"postgres.podLocalDatabaseEnabled=false",
+          raw"postgres.host=$dbServerName.postgres.database.azure.com",
+          raw"postgres.pgbouncer.enabled=$pgBouncerEnabled",
+          // convention is that the database user is the same as the service account name
+          raw"postgres.user=${pet.name()}",
+          raw"postgres.dbnames.cromwell=${databaseNames.cromwellRunner}",
+          raw"postgres.dbnames.tes=${databaseNames.tes}"
```
dillydally414, line 477: We may be able to extract this from the pattern matching since cromwell runner app should only be deployed with these configurations

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreterSpec.scala
```diff
@@ -621,12 +679,16 @@ class AKSInterpreterSpec extends AnyFlatSpecLike with TestComponent with Leonard
     deletion.unsafeRunSync()(cats.effect.unsafe.IORuntime.global)
   }
 
-  for (appType <- List(AppType.Wds, AppType.Cromwell, AppType.HailBatch, AppType.WorkflowsApp))
+  for (
```
dillydally414: CromwellRunnerApp shouldn't be created with WorkspaceShared access, so I'm not sure if these tests should be altered since both this test and the following one create the app with the shared access.

cromwell-runner-app-0.16.0.tgz
salonishah11: Oops I forgot to remove this. Thanks for pointing it out!

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
salonishah11: Yeah I think `shared` in the name is causing some confusion. What I wanted the variable to represent was that it could have either Cromwell or Cromwell Runner databases in it since for both apps we use `maybeCreateCromwellDatabases` method. I will probably revert the name to `maybeCromwellDatabaseNames` only.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
salonishah11: Ah I see. So I should replace `dbNames.get(1).get` with `dbNames.apply(1)` right?

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreterSpec.scala
salonishah11: Ah I see. The following test seems to be creating apps with WSM resources which is right for Cromwell Runner app too no? Update: the other test was also using WorkspaceShared scope.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/BuildHelmChartValues.scala
salonishah11: Hmm I thought that we do want to create Cromwell Runner app with Workload Identity but scoped to the user. @cjllanwarne Is that right?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
dillydally414: I think we can get rid of the if dbNames.size checks

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/BuildHelmChartValues.scala
dillydally414: Yes, the `identity.xyz` configs are all related to the Pod Identity if I'm not mistaken, whereas we only care about the `workloadIdentity.serviceAccountName` since Cromwell Runner should only be created with workload identity and never with Pod

rtitle (approved):
I didn't test but overall this looks good to me.

I should mention that I picked up https://broadworkbench.atlassian.net/browse/IA-4531 which involves some refactoring of this stuff. Basically what I'm doing:

1. Leo always creates WSM-controlled identity, databases, and namespace (a new resource type) for apps
2. Got rid of pod identity entirely; workload identity is always used going forward
3. Broke some stuff out of `AKSInterpreter` -- made a new `AppInstall` trait + subclasses which contain appType-specific logic. `AKSInterpreter` is more generic, and delegates to `AppInstall` for appType-specific logic. I hope this will make this easier to maintain.

Here is the [code so far](https://github.com/DataBiosphere/leonardo/compare/rt-ia-4531) but it's early (only been working on it a ~day) so will need more time to implement and test.

All that said, this PR looks good -- I think this should merge first, just know that this code will be changing (hopefully for the better!).

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/BuildHelmChartValues.scala
rtitle: Agree about just using workload identity -- and in general only passing helm values from Leo that the chart actually uses.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
salonishah11: Yes I have removed that check. I was curious about `.apply(N)` because `.get(N).get` looks ugly

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/BuildHelmChartValues.scala
salonishah11: Ah I got confused about the comment and took it as if we need to Workload Identity or not instead of whether that specific configs `identity.<xyz>` and `workloadIdentity.enabled` is needed to be set in charts or not. Yeah it isn't mentioned in the charts. I think I missed removing it when writing that function. I will remove it now.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -1317,11 +1364,12 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
     ev: Ask[F, AppContext]
   ): F[Option[CromwellDatabaseNames]] = {
     val databaseConfigEnabled = app.appType match {
-      case AppType.Cromwell => config.coaAppConfig.databaseEnabled
-      case _                => false
+      case AppType.Cromwell => config.coaAppConfig.databaseEnabled && landingZoneResources.postgresServer.isDefined
+      case AppType.CromwellRunnerApp => true
```
cjllanwarne, line 1368: I assume it's sorted, but I'm curious how this plays into the WORKFLOWS app and its Cromwell DB

cjllanwarne (approved):

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
salonishah11: Chatted via Slack. For Workflows app [this method](https://github.com/DataBiosphere/leonardo/blob/80f571487b5f904ad109e5f3e890cf1f96716fd6/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala#L1221) is called to create identity and postgres databases

dillydally414 (approved):

PR #3666, opened 2023-09-05
[IA-4531] [IA-3804] Leo should create WSM-controlled kubernetes namespaces for Azure apps

rtitle:
Jira: 
* https://broadworkbench.atlassian.net/browse/IA-4531
* https://broadworkbench.atlassian.net/browse/IA-3804

## Summary of changes

Azure app changes necessary to support multi-user dataplane services, with the right databases & roles.

Design doc: https://docs.google.com/document/d/1FFJRz5AjJTQlghsVDh1wxgQjBZXhgyL1QlHLKKl9bkM/edit?usp=sharing

NOTE: this is a _breaking_ change w.r.t. app upgrade logic. All app versions until present have been excluded from upgrades.

### What

* AKS apps now provision WSM-controlled managed identity; database(s); and kubernetes namespace
* Migrated all apps to Workload Identity, removed support for aad-pod-identity (IA-3804)
* Refactored app type-specific logic to `AppInstall` trait & subclasses

### Why

- See design doc for context; the original model was problematic for shared-apps using LZ-databases in the correct way.

### What to test

I tried to test pretty extensively.

- Unit tests cover all new code
- Tested in a BEE:
   - Created workspace, WDS, Cromwell, imported data tables, ran `sra_to_bam` workflow
   - Deleted workspace, verified apps & all WSM resources were deleted
   - Tested in a fresh LZ, as well as a LZ from a few weeks ago
   - WDS, Cromwell, and Listener upgrades work going forward
   - WDS, Cromwell, and Listener upgrades _fail_ for apps created prior to this branch (before excluding versions)
   - Created Cromwell Runner and Workflows apps in Swagger, verify they reach RUNNING status (though I didn't test much beyond that)

### Who tested and where

- [x] I validated this change 


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/app/CromwellRunnerAppInstall.scala
```diff
@@ -0,0 +1,132 @@
...
+        raw"persistence.storageAccount=${params.landingZoneResources.storageAccountName.value}",
+        raw"persistence.blobContainer=${storageContainer.name.value}",
+        raw"persistence.leoAppInstanceName=${params.app.appName.value}",
+        raw"persistence.workspaceManager.url=${params.config.wsmConfig.uri.renderString}",
+        raw"persistence.workspaceManager.workspaceId=${params.workspaceId.value}",
+        raw"persistence.workspaceManager.containerResourceId=${storageContainer.resourceId.value.toString}",
+
+        // identity configs
+        raw"workloadIdentity.serviceAccountName=${params.ksaName.value}",
+        raw"identity.name=${params.managedIdentityName.value}",
```
rtitle, line 103: `identity.name` is still required -- see https://github.com/broadinstitute/terra-helmfile/commit/010a274595ed155e945adc3099e55385bb2b0bae

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/app/HailBatchAppInstall.scala
```diff
@@ -0,0 +1,54 @@
...
+          raw"persistence.storageAccount=${params.landingZoneResources.storageAccountName.value}",
+          raw"persistence.blobContainer=${storageContainer.name.value}",
+          raw"persistence.workspaceManager.url=${params.config.wsmConfig.uri.renderString}",
+          raw"persistence.workspaceManager.workspaceId=${params.workspaceId.value}",
+          raw"persistence.workspaceManager.containerResourceId=${storageContainer.resourceId.value.toString}",
+          raw"persistence.workspaceManager.storageContainerUrl=https://${params.landingZoneResources.storageAccountName.value}.blob.core.windows.net/${storageContainer.name.value}",
+          raw"persistence.leoAppName=${params.app.appName.value}",
+
+          // identity configs
+          // TODO: can Hail Batch chart migrate to Workfload Identity?
```
rtitle, line 41: I'll reach out to Hail team -- hopefully it should be pretty simple/quick to update their chart to use Workload Identity.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/KubernetesAppConfig.scala
```diff
@@ -209,11 +199,3 @@ final case class AllowedAppConfig(chartName: ChartName,
     "dummy"
   ) // For AoU apps, chart version will vary, and will be populated from user request
 }
-
-sealed trait WorkflowsAppService
```
rtitle, line 212: This trait isn't useful anymore, since each app has their own chart.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -1482,41 +735,155 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
           AppControlledResourceStatus.Created
         )
         .transaction
-    } yield dbName
+    } yield result
   }
 
-  private[util] def deleteAppWsmResources(app: App, workspaceId: WorkspaceId)(implicit
+  private[util] def createWsmKubernetesNamespaceResource(app: App,
```
rtitle, line 762: A lot of the changes in this file are refactoring; but this is new code to create a WSM-controlled `KubernetesNamespace` for https://broadworkbench.atlassian.net/browse/IA-4531

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreterSpec.scala
```diff
@@ -734,71 +150,30 @@ class AKSInterpreterSpec extends AnyFlatSpecLike with TestComponent with Leonard
...
-        mockCbasUiDAO,
-        mockWdsDAO,
-        mockHailBatchDAO,
-        mockWsmDAO,
-        mockKube,
-        mockWsm
-      ) {
-        override private[util] def buildMsiManager(cloudContext: AzureCloudContext) = IO.pure(setUpMockMsiManager)
+  // update and poll each app type
+  // TODO
```
rtitle: OK this is a TODO -- I want to add some unit tests around upgrade (I didn't see any previously)

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSManualTest.scala
```diff
@@ -1,214 +0,0 @@
-package org.broadinstitute.dsde.workbench.leonardo.util
```
rtitle, line 1: I just removed this file. It was useful for testing apps early on, but I don't think it works anymore. It's in git history if we ever need it again.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/app/AppInstall.scala
```diff
@@ -0,0 +1,65 @@
...
+  WorkspaceId
+}
+import org.broadinstitute.dsp.Values
+import org.http4s.Uri
+import org.http4s.headers.Authorization
+
+/**
+ * Defines how to install a Kubernetes App.
+ */
+trait AppInstall[F[_]] {
```
Qi77Qi, line 23: a typclass for apps is cool 👍

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/app/AppInstall.scala
```diff
@@ -0,0 +1,65 @@
...
+
+  /** List of WSM-controlled databases the app requires. */
+  def databases: List[Database]
+
+  /** Builds helm values to be passed to the app. */
+  def buildHelmOverrideValues(params: BuildHelmOverrideValuesParams)(implicit ev: Ask[F, AppContext]): F[Values]
```
Qi77Qi, line 29: this might be a bit too stringent to be applicable to GCP apps 🤔 but can refactor later tho

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/app/AppInstall.scala
```diff
@@ -0,0 +1,65 @@
...
+
+  /** Checks status of the app. */
+  def checkStatus(baseUri: Uri, authHeader: Authorization)(implicit ev: Ask[F, AppContext]): F[Boolean]
+}
+
+object AppInstall {
+
+  /** Maps AppType to AppInstall. */
+  def appTypeToAppInstall[F[_]](wdsAppInstall: WdsAppInstall[F],
```
Qi77Qi, line 38: feels to me this shouldn't be needed. Comipler should be able to find the correct instance depending on the type 🤔 ... can defer to future improvements tho

http/src/main/resources/reference.conf
```diff
@@ -256,7 +256,24 @@ azure {
      enabled = true
      database-enabled = false
      # App developers - Please keep the list of non-backward compatible versions in the list below
+     # All app versions excluded up until the switch to WSM-controlled KubernetesNamespace
+     # https://github.com/DataBiosphere/leonardo/pull/3666
```
rtitle, line 255: I generated these lists by looking at the prod DB, and selecting all versions up to and including the current version in Leo. 
(So after this PR goes in, we'll need 1 more WDS/Cromwell version bump to make them upgradeable going forward.)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -741,16 +415,58 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
...
+        .transaction
+      _ <- wsmDatabases.traverse { database =>
+        deleteWsmResource(workspaceId, app, database)
+      }
+
+      // Then delete namespace resources
+      wsmNamespaces <- appControlledResourceQuery
+        .getAllForAppByType(app.id.id, WsmResourceType.AzureKubernetesNamespace)
+        .transaction
+      deletedNamespace <- wsmNamespaces
```
Qi77Qi, line 431: does any of the delete ops need to be synchronous? like do we need to wait for namespace to be deleted first before deleting identities etc?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/Boot.scala
```diff
@@ -656,6 +664,50 @@ object Boot extends IOApp {
...
+          ConfigReader.appConfig.azure.workflowsAppConfig,
+          ConfigReader.appConfig.drs,
+          samDao,
+          cromwellDao,
+          cbasDao,
+          azureBatchService,
+          azureApplicationInsightsService
+        )
+
+      implicit val appTypeToAppInstall = AppInstall.appTypeToAppInstall(wdsAppInstall,
```
Qi77Qi, line 699: is this implicit actually used?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/app/WorkflowsAppInstall.scala
```diff
@@ -0,0 +1,146 @@
...
+          raw"persistence.storageResourceGroup=${params.cloudContext.managedResourceGroupName.value}",
+          raw"persistence.storageAccount=${params.landingZoneResources.storageAccountName.value}",
+          raw"persistence.blobContainer=${storageContainer.name.value}",
+          raw"persistence.leoAppInstanceName=${params.app.appName.value}",
+          raw"persistence.workspaceManager.url=${params.config.wsmConfig.uri.renderString}",
+          raw"persistence.workspaceManager.workspaceId=${params.workspaceId.value}",
+          raw"persistence.workspaceManager.containerResourceId=${storageContainer.resourceId.value.toString}",
+
+          // identity configs
+          raw"workloadIdentity.serviceAccountName=${params.ksaName.value}",
```
dillydally414, line 106: Should workflows app also still have an identity name here as per https://github.com/DataBiosphere/leonardo/pull/3666/files#r1319924610? I don't see where the identity name is being used in the charts

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/app/CromwellAppInstall.scala
```diff
@@ -0,0 +1,149 @@
...
+      raw"config.landingZoneId=${params.landingZoneResources.landingZoneId}",
+      raw"config.subscriptionId=${params.cloudContext.subscriptionId.value}",
+      raw"config.region=${params.landingZoneResources.region}",
+      raw"config.applicationInsightsConnectionString=${applicationInsightsComponent.connectionString()}",
+
+      // relay configs
+      raw"relay.path=${params.relayPath.renderString}",
+
+      // persistence configs
+      raw"persistence.storageResourceGroup=${params.cloudContext.managedResourceGroupName.value}",
```
jgainerdewar, line 92: A fair number of these are no longer needed - probably makes more sense to leave that cleanup to a future PR, though, and leave this one as a no-op in terms of app install definition.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/app/AppInstall.scala
rtitle: Yes, you're right -- I remember thinking about this now. Maybe there needs to be AKS and GKE variants of `params`. I think starting with this, and later refactoring makes sense though.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/app/AppInstall.scala
rtitle: I kind of modeled it after [ToolDAO](https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/ToolDAO.scala) in Leo. The actual instances are instantiated in `Boot`, and there is an implicit `appTypeToAppInstall: AppType => AppInstall[F]`. This lets you do something like:

```
AppType.WDS.buildHelmOverrideValues(params)
```

There are probably other ways to do it though..

Qi77Qi: oh I see..you want the extension methods on `AppType`... 👍

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
rtitle: Yes, exactly -- the deletion order needs to be: database, namespace, identity. Each of the `deleteWsm*Resource` methods waits for WSM deletion before returning.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/Boot.scala
rtitle: Yeah, see above -- it's passed to `AKSInterpreter` and used for getting the `AppInstall` instance from an `AppType`.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/app/WorkflowsAppInstall.scala
rtitle: I didn't see where the workflows-app [cromwell config](https://github.com/broadinstitute/terra-helmfile/blob/master/charts/workflows-app/templates/cromwell-config.yaml) needs the managed identity name -- maybe because it's a read-only Cromwell, and not submitting jobs?

That said, no harm in Leo passing the value in case Workflows-app needs it for some reason.

jgainerdewar: Correct, it's set in config to be passed to the TES that is running the jobs, so not needed in Workflows app.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/app/CromwellAppInstall.scala
rtitle: Yeah, agreed -- I compared helm values before/after this PR and tried to make everything consistent (modulo the workload identity changes).

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/app/WorkflowsAppInstall.scala
dillydally414: Ah, I didn't see https://github.com/broadinstitute/terra-helmfile/pull/4583 and related discussion. Probably not needed for workflows app then?

rtitle:
jenkins retest

http/src/main/resources/reference.conf
LizBaldo: Gotcha, I am wondering if slightly modifying the app upgrade API to add the option to use an upper bound would be a valid thing to do now

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/app/AppInstall.scala
```diff
@@ -0,0 +1,73 @@
...
+  WorkspaceId
+}
+import org.broadinstitute.dsp.Values
+import org.http4s.Uri
+import org.http4s.headers.Authorization
+
+/**
+ * Defines how to install a Kubernetes App.
+ */
+trait AppInstall[F[_]] {
```
LizBaldo, line 23: Would it make sense to rename to `AKSAppInstall` for now? We can always refactor further later if the GKE apps can follow the same pattern

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/app/AppInstall.scala
```diff
@@ -0,0 +1,73 @@
...
+
+  /** Maps AppType to AppInstall. */
+  def appTypeToAppInstall[F[_]](wdsAppInstall: WdsAppInstall[F],
+                                cromwellAppInstall: CromwellAppInstall[F],
+                                workflowsAppInstall: WorkflowsAppInstall[F],
+                                hailBatchAppInstall: HailBatchAppInstall[F],
+                                cromwellRunnerAppInstall: CromwellRunnerAppInstall[F]
+  ): AppType => AppInstall[F] = _ match {
+    case AppType.Wds               => wdsAppInstall
+    case AppType.Cromwell          => cromwellAppInstall
```
LizBaldo, line 45: Naive question, but do we still need this app type?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/app/HailBatchAppInstall.scala
LizBaldo: Does this mean that we should still leave the pod identity in for now as to not break Hail?

rtitle:
mass failure, jenkins retest

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -1478,41 +756,181 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
...
+      _ <- logger.info(ctx.loggingCtx)(
+        s"Creating $namespacePrefix namespace for app ${app.appName.value} in cloud workspace ${workspaceId.value}"
+      )
+
+      // Build WSM clients
+      wsmApi <- buildWsmControlledResourceApiClient
+      wsmResourceApi <- buildWsmResourceApiClient
+
+      // Name of the WSM resource. Must be unique per workspace.
+      // For shared apps, name it by the appType so it's semantically meaningful.
```
LizBaldo, line 779: Nit that may not need to be addressed ever, but with the change implemented by AOU, appType is not really unique to a chart now, would this eventually break the assumption made here?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -1478,41 +756,181 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
...
+      // There can only be at most 1 shared app type per workspace anyway.
+      // For private apps, use the namespacePrefix to ensure uniqueness.
+      wsmResourceName = app.samResourceId.resourceType match {
+        case SamResourceType.SharedApp => s"${app.appType.toString.toLowerCase}-ns"
+        case _                         => namespacePrefix
+      }
+
+      // Obtain external database IDs by querying WSM and filtering by name
+      // TODO this will get simpler once WSM createNamespace supports passing databases by name instead of UUID
```
LizBaldo: Is there an existing ticket for this work that could be linked here so we know when it's not a blocker anymore?

project/Dependencies.scala
```diff
@@ -144,7 +144,7 @@ object Dependencies {
   val pact4sCirce =       "io.github.jbwheatley"  %% "pact4s-circe"     % pact4sV
   val okHttp =            "com.squareup.okhttp3"  % "okhttp"            % "4.11.0"
 
-  val workSpaceManagerV = "0.254.824-SNAPSHOT"
+  val workSpaceManagerV = "0.254.903-SNAPSHOT"
```
LizBaldo, line 147: I am just curious about the SNAPSHOT version, should we use a more official one?

LizBaldo (approved):
This is amazing! I only have a few nits / comments that should not block you from merging in any way :shipit:

http/src/main/resources/reference.conf
rtitle: I think that probably makes sense -- however, I hope that adding to the exclusion list should be a rare thing going forward.

Qi77Qi (approved):

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/app/AppInstall.scala
rtitle: Yeah, we're still using it for now -- `WORKFLOWS` and `CROMWELL_RUNNER` are behind a feature flag, we will switch to them later.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/app/AppInstall.scala
rtitle: Maybe.. I was hoping to not make this Azure-specific, but I guess it kind of is at the moment. 🤔

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/app/HailBatchAppInstall.scala
rtitle: That would mean adding a bunch of code back to Leo.. I think it should be a really quick change to their Helm chart, hoping they can just do it (will get in touch with them this week).

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
rtitle: I'll piggy back on https://broadworkbench.atlassian.net/browse/WOR-1250, will update this comment.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/app/AppInstall.scala
LizBaldo: Really up to you :) I tend to prefer representing the current state of the code, especially if the GKE refactoring is not going to be a fast follow

project/Dependencies.scala
rtitle: Hm yeah, will try

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
rtitle: Yeah I guess so. :-/ Kind of academic I guess though, since `ALLOWED` cannot be launched in shared mode.

LizBaldo: Oh very true, we might get shared mode there in the future but for now feel free to ignore my nit picking 😬

project/Dependencies.scala
rtitle: …nope, apparently only SNAPSHOT versions are published in Artifactory

PR #3694, opened 2023-09-12
[IA-4399] WSM state reconciliation

lucymcnatt:

### What

-

### Why

-

### What to test

- 

### Who tested and where

- [ ] I validated this change 


rtitle (commented):
Just noticed this -- FYI I'm experimenting with some changes to `WsmApiClientProvider` in https://github.com/DataBiosphere/leonardo/pull/3699, mostly to pass it a `Span` for cross-service tracing. (I don't think it should conflict much with this though.)

Just curious, in `AKSInterpreter` Leo is using the WSM client directly -- i.e. there is no Scala `WsmApiClientDAO` layer in front of it. I couldn't decide if there was value in having a DAO layer vs just using the client where it's needed (so I just opted for the latter). Curious if you considered that too?

PR #3699, opened 2023-09-12
WOR-1250: Update WSM client to use names instead of ids

rtitle:
https://broadworkbench.atlassian.net/browse/WOR-1250

## Dependencies
Depends on https://github.com/DataBiosphere/terra-workspace-manager/pull/1470

## Summary of changes

Updates to new WSM client library to facilitate passing WSM resources by name instead of UUID. This will let us simplify Leo code, and facilitate cloning.

### What

- Update WSM client as an intermediate step. Does not actually pass-by-name yet
- Generates OpenCensus `Span` in Back Leo pubsub handler, and passes it to WSM for tracing

Demo of Tracing, this is a `createAppV2` message:

image

Update: Trace with WSM included:

image

### Why

- Cleaner WSM integration, and cross-service tracing!

### What to test

- Unit tests pass, covering WSM calls and AKS app creation with mocks
- Deployed to a BEE, ran standard test of create workspace -> load data tables -> run workflow
- Looked at traces in GCP console

### Who tested and where

- [ ] I validated this change 


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/WsmApiClientProvider.scala
```diff
@@ -11,27 +17,32 @@ import org.http4s.Uri
...
-  }
+class HttpWsmClientProvider[F[_]](baseWorkspaceManagerUrl: Uri)(implicit F: Async[F]) extends WsmApiClientProvider[F] {
+  private def getApiClient(token: String)(implicit ev: Ask[F, AppContext]): F[ApiClient] =
+    for {
+      ctx <- ev.ask
+      client = new ApiClient() {
+        override def performAdditionalClientConfiguration(clientConfig: ClientConfig): Unit = {
+          super.performAdditionalClientConfiguration(clientConfig)
+          ctx.span.foreach { span =>
+            clientConfig.register(Tracing.getTracer.withSpan(span))
```
rtitle: Pass a `Span` to the WSM client to enable cross-service tracing. Inspired by [Rawls](https://github.com/broadinstitute/rawls/blob/develop/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/workspacemanager/WorkspaceManagerApiClientProvider.scala#L44).

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -138,7 +139,9 @@ class LeoPubsubMessageSubscriber[F[_]](
   private[monitor] def messageHandler(event: Event[LeoPubsubMessage]): F[Unit] = {
     val traceId = event.traceId.getOrElse(TraceId("None"))
     val now = Instant.ofEpochMilli(com.google.protobuf.util.Timestamps.toMillis(event.publishedTime))
-    implicit val appContext = Ask.const[F, AppContext](AppContext(traceId, now))
+    val span = Tracing.getTracer.spanBuilder("leoPubsubMessage").startSpan()
```
rtitle, line 142: Create a `Span` for the lifecycle of pubsub message handling. For now just creates 1 parent span, no children (will experiment though).

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -186,9 +189,11 @@ class LeoPubsubMessageSubscriber[F[_]](
...
     } yield ()
 
-    res.handleErrorWith(e =>
-      logger.error(e)("Fail to process pubsub message for some reason") >> F.delay(event.consumer.ack())
-    )
+    res
+      .handleErrorWith(e =>
+        logger.error(e)("Fail to process pubsub message for some reason") >> F.delay(event.consumer.ack())
+      )
+      .guarantee(F.delay(span.end()))
```
rtitle, line 196: `guarantee` comes from cats-effect [MonadCancel](https://typelevel.org/cats-effect/docs/typeclasses/monadcancel), it works like `finally` :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -655,7 +656,7 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
                                                appInstall: AppInstall[F],
                                                workspaceId: WorkspaceId,
                                                namespacePrefix: String,
-                                               owner: Option[UUID],
+                                               owner: Option[String],
```
rtitle, line 689: The type changes come from the new WSM client, in preparation for accepting names instead of id's.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/WsmApiClientProvider.scala
rtitle: Here's a span for Leo in a BEE!

Unfortunately I can't see WSM in the trace because
https://github.com/broadinstitute/terra-helmfile/blob/master/values/app/workspacemanager/bee.yaml.gotmpl#L87
```
# WSM doesn't currently have permissions for this in QA, it just spams errors
samplingProbability: 0
```

image

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -138,13 +138,21 @@ class LeoPubsubMessageSubscriber[F[_]](
   private[monitor] def messageHandler(event: Event[LeoPubsubMessage]): F[Unit] = {
     val traceId = event.traceId.getOrElse(TraceId("None"))
     val now = Instant.ofEpochMilli(com.google.protobuf.util.Timestamps.toMillis(event.publishedTime))
-    implicit val appContext = Ask.const[F, AppContext](AppContext(traceId, now))
+    implicit val ev = Ask.const[F, AppContext](AppContext(traceId, now, span = None))
+    childSpan(event.msg.messageType.asString).use { implicit ev =>
```
rtitle, line 142: Generate a root `Span` at the pubsub message handler entry point, and pass it downstream in the `AppContext`.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -100,27 +100,34 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
       // Create WSM managed identity if shared app
       wsmManagedIdentityOpt <- app.samResourceId.resourceType match {
         case SamResourceType.SharedApp =>
-          createWsmIdentityResource(app, namespacePrefix, params.workspaceId).map(_.some)
+          childSpan("createWsmIdentityResource").use { implicit ev =>
```
rtitle, line 103: In `AKSInterpreter` I tried to make child Spans for WSM calls and helm calls. See screenshot in PR description.

rtitle:
```
org.broadinstitute.dsde.workbench.leonardo.runtimes.RuntimeDataprocSpec.should create a Dataproc cluster with worker private access
```

jenkins retest

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/package.scala
```diff
@@ -93,6 +94,34 @@ package object http {
...
+   * }
+   */
+  def childSpan[F[_]: Sync](name: String)(implicit ev: Ask[F, AppContext]): Resource[F, Ask[F, AppContext]] =
+    for {
+      ctx <- Resource.eval(ev.ask)
+      newSpan <- Resource.make[F, Span](
+        Sync[F].delay(
+          Tracing.getTracer
+            .spanBuilderWithExplicitParent(name, ctx.span.orNull)
+            .setSampler(Samplers.alwaysSample())
```
rtitle: Do we think `alwaysSample` is safe? Leo pubsub messages are not terribly high.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/WsmApiClientProvider.scala
dvoet: I am curious that this does not need the `JerseyTracingFilter` like rawls uses

rtitle: Um, maybe it does. Leo doesn't have the `bio.terra.common` package. Will look… 👁️

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/WsmApiClientProvider.scala
```diff
@@ -11,27 +17,32 @@ import org.http4s.Uri
...
-    val client: ApiClient = new ApiClient()
-    client.setBasePath(baseWorkspaceManagerUrl.renderString)
-    client
-  }
+class HttpWsmClientProvider[F[_]](baseWorkspaceManagerUrl: Uri)(implicit F: Async[F]) extends WsmApiClientProvider[F] {
+  private def getApiClient(token: String)(implicit ev: Ask[F, AppContext]): F[ApiClient] =
+    for {
+      ctx <- ev.ask
+      client = new ApiClient() {
+        override def performAdditionalClientConfiguration(clientConfig: ClientConfig): Unit = {
```
dvoet, line 32: I am not certain this is thread safe. `performAdditionalClientConfiguration` is called during `ApiClient` construction. The constructed client is wrapped in a `F` which is then chained to other calls. Those other calls are not guaranteed to be on the same thread. I am pretty sure `Tracing.getTracer` uses `ThreadLocal` so calls might get the wrong span or none at all.

LizBaldo (approved):
LGTM, thanks! One nit is that I would have split the PR into to, one for the `Span` setup from the uuid to string change for WSM resource ids.

Just some small lesson learned from the previous PRs, where folks were super excited about the refactoring but missed the namespace change 😬

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/package.scala
dvoet: can you just omit this?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/WsmApiClientProvider.scala
rtitle: Ah yeah of course that's needed to send the Trace ID in a header. I added TCL dependency to Leo and added `JerseyTracingFilter` and I see WSM spans now (also granted WSM SA `Trace Agent` role in broad-dsde-qa):

image

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/WsmApiClientProvider.scala
rtitle: Fascinating… maybe Leo should do something like this then? https://github.com/broadinstitute/rawls/blob/develop/core/src/main/scala/org/broadinstitute/dsde/rawls/util/WithSpanFilter.scala

rtitle:
> One nit is that I would have split the PR into to, one for the Span setup from the uuid to string change for WSM resource ids.

> Just some small lesson learned from the previous PRs, where folks were super excited about the refactoring but missed the namespace change 😬

Thanks @LizBaldo, that's good feedback. I've opportunistically made some improvements during the last couple tickets, but I agree it might be better as a separate PR. Will keep that in mind for the future :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/WsmApiClientProvider.scala
rtitle: …I gave that a shot

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/package.scala
rtitle: Done - i see the frequency is actually configurable in terra-helmfile

PR #3700, opened 2023-09-12
IA-4520 persist app region

cindy-broadinstitute:

### What

- Persist region to table: KUBERNETES_CLUSTER , field: region (which already exists in mysql) 
- set up the API contracts to send this region info for ListApp and GetApp
- add the new contract to Swagger

### Why

- Apps had been using the default region, not the one they were actually in. This allows us to display the actual region.

### What to test

-  Verify Swagger shows this region field correctly when calling the List Apps

### Who tested and where

- [ ] I validated this change 


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
```diff
@@ -638,8 +640,18 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
         }
 
       // Save or retrieve a KubernetesCluster record for the app
+      // debugging
+      regionOpt = landingZoneResourcesOpt.map(_.region)
+      regionName = regionOpt.map(_.name).getOrElse("")
+      _ <- log.info(ctx.loggingCtx)(s"DEBUGGING: azure region: ${regionName}")
```
LizBaldo: Do you still need this log or can it be removed now?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
```diff
@@ -638,8 +640,18 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
         }
 
       // Save or retrieve a KubernetesCluster record for the app
+      // debugging
+      regionOpt = landingZoneResourcesOpt.map(_.region)
+      regionName = regionOpt.map(_.name).getOrElse("")
```
LizBaldo: In what condition would we get an empty string for the region? Would it make sense to leave a comment, or log a warning if the region does not exist?

LizBaldo (commented):
Looks great! I only have a small question around handling the case where we are not getting a region name, should we warn, error, or just let it go?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
cindy-broadinstitute: I am going to remove it, then take this out of Draft

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
cindy-broadinstitute: This is debugging code but I will answer this for where it is used around line 871.
 What is the best practice for checking for this? I changed it to a get() with a comment that it shouldn't be null so the get() will throw an exception if it is.

LizBaldo: Thanks! I would maybe query the dev or prod database to see if there are any apps without a region. If there is an expected case, then a comment should be fine, but if not, throwing an exception is the right thing to do imo

cindy-broadinstitute: Some of the unit tests have have apps without a region apparently (tests for when the user failed to complete something I think). And I am not suitable so don't think I can query prod. Let me know if the change I made looks right.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
```diff
@@ -860,6 +868,8 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
       autoscalingConfig = None
     )
 
+    // regoinName can be empty in some test configurations
```
LizBaldo: Super small nit: `regionName`

LizBaldo (approved):
Just the smallest nit/typo, thanks for addressing my comment, :shipit:

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/appRoutesModels.scala
```diff
@@ -37,6 +38,7 @@ final case class CreateAppRequest(kubernetesRuntimeConfig: Option[KubernetesRunt
 
 final case class GetAppResponse(appName: AppName,
                                 cloudContext: CloudContext,
+                                region: RegionName,
```
mlilynolting, line 41: Remind me - this endpoint is for GCP apps only, right?

mlilynolting: (these endpoints, I should say; get and list)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
```diff
@@ -860,6 +868,8 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
       autoscalingConfig = None
     )
 
+    // regionName can be empty in some test configurations
+    val regionName = if (azureRegionOpt.isEmpty) "" else azureRegionOpt.map(_.name).getOrElse("")
```
mlilynolting, line 872: is `Option.isEmpty` true if the value of the option is `None`?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
```diff
@@ -868,7 +878,9 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
       cloudContext = cloudContext,
       clusterName = clusterName,
       location = config.leoKubernetesConfig.clusterConfig.location,
-      region = config.leoKubernetesConfig.clusterConfig.region,
+      region =
+        if (cloudContext.cloudProvider == CloudProvider.Azure) RegionName(regionName)
```
mlilynolting, line 882: Isn't the `RegionName` class for Google regions? Should this conditional be a `!=`?

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/appRoutesModels.scala
cindy-broadinstitute: Hmm, I thought it was for Azure too

mlilynolting: I'm not sure if Google's RegionName and Azure's region values are compatible 🤔

mlilynolting: Can Scala do union types?

mlilynolting: Or would we need to have two different valid signatures at the API doc level as well as the routes level?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
cindy-broadinstitute: Yes, I tested it as I wasn't sure either.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
cindy-broadinstitute: region is a RegionName

`case class SaveKubernetesCluster(cloudContext: CloudContext,
                                 clusterName: KubernetesClusterName,
                                 location: Location,
                                 region: RegionName,
                                 ingressChart: Chart,
                                 auditInfo: AuditInfo,
                                 defaultNodepool: DefaultNodepool
)

This should be a == because we are setting region in KubernetesCluster table when it is an azure app.

mlilynolting: Right, but that's a legacy of the `region` field up to now being populated only for GKE apps, right? And now we're adding support for Azure, which would have its own region syntax, I thought

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
mlilynolting: Wouldn't this be true for any apps created on AKS before this code is released? Or am I missing something?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
lucymcnatt: you could define the region variable as for SaveKubernetesCluster as `Either[RegionName, Region]`, with region being the Azure region, then change the file that defines the cluster component to reflect that (org/broadinstitute/dsde/workbench/leonardo/db/KubernetesClusterComponent.scala)

lucymcnatt: then region for GCP apps would be `Left(config.leoKubernetesConfig.clusterConfig.region)` and for Azure apps `Right(regionName)`

lucymcnatt: but that might be tricky on the decoder side...

lucymcnatt: I'd be _ok_ with leaving this as is (there's similar logic in runtimeV2ServiceInterp with ZoneName) as long as we leave a comment explaining

mlilynolting: But the regions aren't the same values, are they? How is it going to work if we're trying to persist Azure regions as Google regions?

mlilynolting: I guess they both have code for persisting arbitrary strings as regions. But doesn't that open us up to problems down the line?

lucymcnatt: I guess we could add a new RegionName(value: String) class to the kubernetesModels that would be cloud agnostic

cindy-broadinstitute: It is just a string in the database. So any decoding could be done based on cloud type of azure or google to find the label or valid string based on the strings in Region vs RegionName.

cindy-broadinstitute: @lucymcnatt @mlilynolting - would you have time for a quick meeting to discuss the best practice for RegionName and Region? I think @lucymcnatt's suggestion of having a base RegionName class in kubernetesModels sounds like the right approach, but I need to look into how or if our DAO handles polymorphism and how the contracts are using RegionName. How about 11am?

mlilynolting: Sure! I have a 1:1 at 11 but I'm free between 12 and 2; I can check in if you and Lucy end up meeting earlier

mlilynolting: the way we both overload AND duck-type our data model gives me the heebie-jeebies👻

lucymcnatt (approved):

PR #3713, opened 2023-09-18
[IA-4538] scala steward runs midnight between Sunday and Monday

mlilynolting:
Cleaner. Moved timing to 12:01 AM for disambiguation


LizBaldo (approved):
Thank you :)

PR #3721, opened 2023-09-19
[IA-4567] Leonardo Apps Dashboard isn't populating

rtitle:
https://broadworkbench.atlassian.net/browse/IA-4567

## Summary of changes

* Fix issue where metrics isn't populating
* Add "upgradeable" label on app metrics, for hopefully better organization/filtering in the dashboard
* Opportunistically DRY up some code

### What

- See inline

### Why

- I happened to notice this while monitoring the BroadE workshop dry run

### What to test

- [In progress] Testing metric collection on a BEE

### Who tested and where

- [ ] I validated this change 


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/KubernetesAppConfig.scala
```diff
@@ -33,6 +35,23 @@ sealed trait KubernetesAppConfig extends Product with Serializable {
   def appType: AppType
 }
 
+object KubernetesAppConfig {
+  def configForTypeAndCloud(appType: AppType, cloudProvider: CloudProvider): Option[KubernetesAppConfig] =
```
rtitle, line 39: This logic was in a couple places already -- instead of adding it in a 3rd, tried to DRY up this code a bit.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoMetricsMonitor.scala
```diff
@@ -45,7 +45,7 @@ class LeoMetricsMonitor[F[_]](config: LeoMetricsMonitorConfig,
   clusterToolToToolDao: RuntimeContainerServiceType => ToolDAO[F, RuntimeContainerServiceType],
   ec: ExecutionContext
 ) {
-  private val parallelism = 25
+  private val parallelism = 40
```
rtitle, line 48: These metrics take some time to collect in prod. Also Back Leo CPU is low, so bumping the parallelism a bit.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoMetricsMonitor.scala
```diff
@@ -239,7 +247,7 @@ class LeoMetricsMonitor[F[_]](config: LeoMetricsMonitorConfig,
       .parTraverseN(parallelism) { case (runtime, image, container) =>
         for {
           ctx <- ev.ask
-          isUp <- container.isProxyAvailable(runtime.cloudContext, runtime.runtimeName)
+          isUp <- container.isProxyAvailable(runtime.cloudContext, runtime.runtimeName).handleError(_ => false)
```
rtitle, line 250: I _think_ this was the metrics collection issue. We were getting some exceptions like
```
javax.net.ssl.SSLHandshakeException: No trust manager was able to validate this certificate chain: # of exceptions = 2
```
from runtimes, which weren't being caught and caused the whole metrics collection process to crash.

…I'm still trying to validate this though.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoMetricsMonitor.scala
```diff
@@ -441,6 +449,11 @@ class LeoMetricsMonitor[F[_]](config: LeoMetricsMonitorConfig,
         )
       })
       .getOrElse(List.empty)
+
+  private def isUpgradeable(appType: AppType, cloudProvider: CloudProvider, chart: Chart): Boolean =
```
rtitle, line 453: This was an idea of mine… the current apps dashboard is fairly cluttered, since there are a bunch of old apps in the system which won't be upgraded (some of which are failing health checks, etc). I'm adding an `upgradeable` label to app metrics, which should let us filter those old apps out of the dashboard.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/KubernetesAppConfig.scala
LizBaldo: Nit/question, why `AppType.Allowed` instead of `Allowed` directly?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AdminServiceInterp.scala
```diff
@@ -42,10 +41,10 @@ final class AdminServiceInterp[F[_]: Parallel](authProvider: LeoAuthProvider[F],
       _ <- F.raiseWhen(!hasPermission)(NotAnAdminError(userInfo.userEmail, Option(ctx.traceId)))
 
       // Find the config relevant to this cloud and app type
-      appConfig: KubernetesAppConfig <- adminAppConfig.configForTypeAndCloud(req.appType, req.cloudProvider) match {
-        case Some(conf) => F.pure(conf)
-        case None       => F.raiseError(NoMatchingAppError(req.appType, req.cloudProvider, Option(ctx.traceId)))
-      }
+      appConfig: KubernetesAppConfig <- F.fromOption(
```
LizBaldo, line 44: Ahhh scala, every time I get comfortable with something (e.g. match) I then realize that there is almost always something a lot better to use instead 😓

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoMetricsMonitor.scala
LizBaldo: Good catch, I would not have expected this to crash the entire metrics collection though, just the runtimes 🤔

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoMetricsMonitor.scala
LizBaldo: Nice!

LizBaldo (approved):

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/KubernetesAppConfig.scala
rtitle: Typo I guess, we import:
```
import org.broadinstitute.dsde.workbench.leonardo.AppType._
import org.broadinstitute.dsde.workbench.leonardo._
```
so either works. Will change it to be consistent.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AdminServiceInterp.scala
rtitle: Heh - either is fine honestly, I happen to like `fromOption` :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoMetricsMonitor.scala
rtitle: That's the weird part, I would also expect to see a log line ["Unexpected error occurred during metric monitoring"](https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoMetricsMonitor.scala#L55), but I'm not seeing it. 🤔 
Wondering if the SSL error is causing other issues, this is what I want to dig in a little more on a BEE, and hopefully replicate.

rtitle:
Tested on a BEE, app creation works and I see metrics collection happening. I also see `javax.net.ssl.SSLHandshakeException: No trust manager was able to validate this certificate chain: # of exceptions = 2` for some old Cromwell-GCP apps in my BEE, and it did _not_ crash metrics collection. So hopefully this helps -- merging.

PR #3728, opened 2023-09-21
RW-10808 Add SAS App usage tracking

Qi77Qi:
https://precisionmedicineinitiative.atlassian.net/browse/RW-10808

See [related doc](https://docs.google.com/document/d/1TmSksMNLop6qLKWZAOyr65btWHlKfHUSnI7V1p67Y4o/edit#heading=h.w7w2865nbqac) for more info.

This PR does the following
- Add AppUsage table to track app usage
- Record `startTime` when SAS app is up&running either during app creation or when an existing app is being started
- Record `stopTime` when we receive `deleteApp` or `stopApp` in front leo. I think this is appropriate since users essentially can't use the app anymore at this point

TODO:
- Test things out in BEE e2e (unit tests should have good coverage. but will test things out e2e as well)

- [ ] I validated this change 


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -354,7 +356,7 @@ class GKEInterpreter[F[_]](
       )
 
       // Create KSA
-      ksaName: ServiceAccountName <- F.fromOption(
+      ksaName <- F.fromOption(
```
Qi77Qi, line 359: `:` inside for comprehension will introduce an additional function call...we should avoid doing that...ascribing type anywhere else is encouraged :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/AppUsageComponent.scala
```diff
@@ -0,0 +1,107 @@
...
+          )
+        else
+          appUsageQuery returning appUsageQuery.map(_.id) += AppUsageRecord(AppUsageId(-1), appId, startTime, dummyDate)
+    } yield id
+
+    for {
+      res <- dbio.transaction.attempt
+      id <- res match {
+        case Left(e) if e.getMessage.contains("usage startTime was recorded previously") =>
+          // We should alert if this happens
```
Qi77Qi, line 56: thinking of creating alerts once this PR is merged, and revisiting proper way of addressing the case then

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/AppUsageComponent.scala
```diff
@@ -0,0 +1,107 @@
...
+    F: Sync[F],
+    logger: Logger[F]
+  ): F[AppUsageId] = {
+
+    val dbio = for {
+      recordExists <- appUsageQuery
+        .filter(_.appId === appId)
+        .filter(_.stopTime === dummyDate)
+        .exists
+        .result // Make sure there's no existing appId that doesn't have endTime recorded already
```
LizBaldo, line 42: I am not sure I understand this comment, wouldn't newly created apps not have a stop time?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/AppUsageComponent.scala
LizBaldo: Could you elaborate a bit on this failure case? I am not sure I understand it based on the error message above

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/AppUsageComponent.scala
```diff
@@ -0,0 +1,107 @@
...
+  def id = column[AppUsageId]("id", O.PrimaryKey, O.AutoInc)
+  def appId = column[AppId]("appId")
+
+  def startTime = column[Instant]("startTime", O.SqlType("TIMESTAMP(6)"))
+  def stopTime = column[Instant]("stopTime", O.SqlType("TIMESTAMP(6)"))
+
+  def * = (id, appId, startTime, stopTime) <> (AppUsageRecord.tupled, AppUsageRecord.unapply)
+}
+
+object appUsageQuery extends TableQuery(new AppUasageTable(_)) {
```
LizBaldo, line 28: Just to make sure I understand, every new start/stop pair will be recorded in a new row in the db correct? Nothing is overwritten?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -1215,8 +1218,11 @@ class GKEInterpreter[F[_]](
           } yield ()
         } else {
           for {
+            startTime <- F.realTimeInstant
```
LizBaldo, line 1221: Could you maybe modify the comment below? It sounds more general than just for Galaxy

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20230911_create_app_usage_table.xml
```diff
@@ -0,0 +1,22 @@
...
+            <column name="id" type="BIGINT" autoIncrement="true">
+                <constraints primaryKey="true"/>
+            </column>
+            <column name="appId" type="BIGINT(20)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="startTime" type="TIMESTAMP(6)" defaultValue="0000-00-00 00:00:00.000000">
+                <constraints nullable="false"/>
+            </column>
+            <column name="stopTime" type="TIMESTAMP(6)" defaultValue="1970-01-01 00:00:01.000000">
```
yonghaoy, line 14: (not having too much scala & Leo knowledge)Would that be easier if we make this nullable, then insert null when app is started.
I am wondering if that would make the potential metrics/dashboard less confusing for running apps... 
(was thinking the BQ can just be where `stopTime is not null` but using the current style won't make too complicated).

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20230911_create_app_usage_table.xml
```diff
@@ -0,0 +1,22 @@
...
+                <constraints nullable="false"/>
+            </column>
+        </createTable>
+        <addForeignKeyConstraint baseColumnNames="appId" baseTableName="APP_USAGE" constraintName="FK_APP_USAGE_APP_ID" deferrable="false" initiallyDeferred="false" onDelete="NO ACTION" onUpdate="NO ACTION" referencedColumnNames="id" referencedTableName="APP"/>
+        <addUniqueConstraint columnNames="appId, startTime" constraintName="IDX_APP_USAGE_STARTTIME_UNIQUE" tableName="APP_USAGE"/>
+        <addUniqueConstraint columnNames="appId, startTime, stopTime" constraintName="IDX_APP_USAGE_STOPTIME_UNIQUE" tableName="APP_USAGE"/>
```
yonghaoy: Is this needed if we have the `appId, startTime` as unique key above?

yonghaoy (commented):
Change looks good. 
One minor question is for future use cases that I heard from few meetings. If product wants to use this table for Leo related reliable/usage tracking, it is highly possible that they also want to see Runtime usages. 
What we can do for now: 
1: Make the table name more generic?
2: Make code usable as much as we can. 

But those are future cases, we may not need to consider that as of now...

yonghaoy (approved):
One more question, do we covert Cron job cases? Would that be possible 
1: GKE cluster is deleted/in bad state while App state in Leo is RUNNING. We will change state to ERROR, then also update that usage table. 
2: Not sure other cases can be corner cases here, stuck in CREATING, DELETING; something within Janitor job or other jobs that directly modify Leo DB.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/AppUsageComponent.scala
Qi77Qi: this is when we have a record already for this app and that record doesn't have `stopTime` recorded.

so imagine in DB we have the following already, we should error if leo tries to record another startTime

| id | appId | startTime                  | endTime                    |
|----|-------|----------------------------|----------------------------|
| 1  | 100   | 2023-01-24 20:48:39.666000 | 1970-01-01 00:00:01.000000 |

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/AppUsageComponent.scala
Qi77Qi: see https://github.com/DataBiosphere/leonardo/pull/3728#discussion_r1333599960

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/AppUsageComponent.scala
Qi77Qi: yes..

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
Qi77Qi: I didn't add the comment..but sure :)

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20230911_create_app_usage_table.xml
Qi77Qi: if we have `stopTime` is nullable, then mysql will update other `TIMESTAMP` fields automatically anytime the row is updated. This means `startTime` will be updated when we update `stopTime`...the only way I found to prevent that is to provide a default value

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/AppUsageComponent.scala
LizBaldo: Ohhhh got it, makes complete sense, thanks!

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20230911_create_app_usage_table.xml
Qi77Qi: ah yea, this isn't needed...I was experimenting with somethings

Qi77Qi:
@yonghaoy 

> 1: GKE cluster is deleted/in bad state while App state in Leo is RUNNING. We will change state to ERROR, then also update that usage table.

Yea. That could happen. It wouldn't be handled in this PR tho. It'll be handled in cron jobs

> 2: Not sure other cases can be corner cases here, stuck in CREATING, DELETING; something within Janitor job or other jobs that directly modify Leo DB.

We shouldn't need to worry about stuck in `CREATING` much since we only record usage if the app is actually up&running.
Hopefully we shouldn't need to worry about stuck in `DELETING` much since we record `stopTime` in frontleo before we do the actual GCP deletion.

There're metrics created in this PR that'll warn us if this happens.

yonghaoy (approved):

Qi77Qi:
@LizBaldo lmk if u have more questions/concerns

LizBaldo (approved):
LGTM, thanks for addressing my comments!

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20230911_create_app_usage_table.xml
jmthibault79: could you say more about that?  that is surprising behavior.

PR #3756, opened 2023-09-27
[WM-2161] cromwell runner metadata

mspector:

### What
- Passes external databases (`ReferenceDatabase`) into AKS app helm charts.

### Why
- As part of the "multi-user workflows" effort, `CROMWELL_RUNNER_APP` needs access to the `cromwellmetadata` database that is created with the `WORKFLOWS_APP`. 
- To get that access, the full database name (including namespace suffix; e.g. `cromwellmetadata_xqjonq`) must be passed into the `CROMWELL_RUNNER_APP` helm chart upon creation/update.

### What to test
- Manual test in a BEE to ensure that external database is passed into `CROMWELL_RUNNER_APP` upon app creation
- Updated `AKSInterpreterSpec` unit tests

### Who tested and where

- [x] I validated this change (in a BEE)


mspector:
jenkins retest

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/app/CromwellRunnerAppInstall.scala
```diff
@@ -57,8 +57,11 @@ class CromwellRunnerAppInstall[F[_]](config: CromwellRunnerAppConfig,
       )
 
       // Databases required for Cromwell App
-      dbNames <- F.fromOption(toCromwellRunnerAppDatabaseNames(params.databaseNames),
-                              AppCreationException("Database names required for Cromwell Runner app", Some(ctx.traceId))
+      dbNames <- F.fromOption(
+        toCromwellRunnerAppDatabaseNames(params.databaseNames),
+        AppCreationException(s"Database names required for Cromwell Runner app: ${params.databaseNames}",
```
mspector, line 62: I found this update to the log message helpful, and I thought it was worth keeping (unlike many of my other added logs, now removed)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -97,6 +97,21 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
...
+      wsmResourceDatabases <- F.delay(
+        wsmResourceApi
+          .enumerateResources(params.workspaceId.value, 0, 100, ResourceType.AZURE_DATABASE, StewardshipType.CONTROLLED)
+          .getResources
+          .asScala
+          .toList
+      )
+      referenceResourceDatabaseNames = wsmResourceDatabases
+        .filter(r => referenceDatabaseNames.contains(r.getMetadata().getName()))
+        .map(r => r.getResourceAttributes().getAzureDatabase().getDatabaseName())
```
mspector: you'll see this logic repeated here and in `updateAndPollApp`. I intentionally did not pull this out into a function because it feels more readable this way, but I'm open to other opinions

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -1026,6 +1055,17 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
       }
       wsmApi <- wsmClientProvider.getControlledAzureResourceApi(token)
     } yield wsmApi
+
+  // can these two functions be combined somehow? ask a Scala person for the right way to do it
```
mspector: just caught this stray comment - I'll remove it before merging - but the question stands for any reviewer who wants to answer it!

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -97,6 +97,21 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
         s"Begin app creation for app ${params.appName.value} in cloud context ${params.cloudContext.asString}"
       )
 
+      referenceDatabaseNames = app.appType.databases.collect { case ReferenceDatabase(name) => name }.toSet
+
+      // build WSM resource client
+      wsmResourceApi <- buildWsmResourceApiClient
+      wsmResourceDatabases <- F.delay(
```
rtitle: As an optimization, could we only make this API call if the app actually _has_ any reference databases? 
```
referenceDatabaseNames = app.appType.databases.collect { case ReferenceDatabase(name) => name }.toSet
referenceDatabases <- if (referenceDatabaseNames.nonEmpty) {
  retrieveWsmReferenceDatabases(…)
} else F.pure(List.empty)
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -97,6 +97,21 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
         s"Begin app creation for app ${params.appName.value} in cloud context ${params.cloudContext.asString}"
       )
 
+      referenceDatabaseNames = app.appType.databases.collect { case ReferenceDatabase(name) => name }.toSet
```
rtitle: Maybe a nit, but could we move this new logic a bit down, maybe after L134 after it creates the WSM databases?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
rtitle: See above comment -- after looking at the code, I think it would read a bit cleaner as a separate method.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
rtitle: I guess the token-retrieval logic could be DRYed a bit.. as-is seems ok to me though.

rtitle (commented):
A couple comments, overall looks good though and seems like it works :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/app/CromwellRunnerAppInstall.scala
```diff
@@ -119,16 +122,20 @@ class CromwellRunnerAppInstall[F[_]](config: CromwellRunnerAppConfig,
...
     } yield Values(values.mkString(","))
 
   override def checkStatus(baseUri: Uri, authHeader: Authorization)(implicit ev: Ask[F, AppContext]): F[Boolean] =
     cromwellDao.getStatus(baseUri, authHeader).handleError(_ => false)
 
   private def toCromwellRunnerAppDatabaseNames(dbNames: List[String]): Option[CromwellRunnerAppDatabaseNames] =
-    (dbNames.find(_.startsWith("cromwell")), dbNames.find(_.startsWith("tes")))
+    (dbNames.find(_.startsWith("cromwell")),
+     dbNames.find(_.startsWith("tes")),
+     dbNames.find(_.startsWith("cromwellmetadata"))
```
cjllanwarne: I would expect `dbNames.find(_.startsWith("cromwell"))` to also include `dbNames.find(_.startsWith("cromwellmetadata"))`. Is there any chance of this causing problems?

cjllanwarne: Since we're in the CROMWELL_RUNNER here, does the other Cromwell DB have a longer name that we can search for more specifically? Or does it not matter?

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreterSpec.scala
```diff
@@ -665,7 +666,27 @@ class AKSInterpreterSpec extends AnyFlatSpecLike with TestComponent with Leonard
...
+    // enumerate workspace resources
+    when {
+      resourceApi.enumerateResources(any, any, any, any, any)
+    } thenReturn {
+      val resourceList = new ArrayList[ResourceDescription]
+      val resourceDesc = new ResourceDescription()
+      resourceDesc.metadata(new ResourceMetadata().name("cromwellmetadata"))
+      resourceDesc.resourceAttributes(
+        new ResourceAttributesUnion().azureDatabase(
+          new AzureDatabaseAttributes().databaseName("cromwellmetadata_abcxyz")
```
cjllanwarne, line 679: I don't see this `cromwellmetadata_abcxyz` being used anywhere else in testing. Is there any way to check our wiring through, eg that we are plumbing _this_ value from WDS into the appropriate helm chart value for the cromwell runner?

cjllanwarne (approved):
Looks correct, a few minor thoughts/comments. Would love to see a unit test for extracting the value from the WDS response, and (maybe another test) plumbing it into the chart values

cjllanwarne:
@mspector and @rtitle I was a little surprised we have to go back to WSM. Don't we already go there once to fetch the database information to set up the postgres roles? Couldn't we store it and reuse it?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/app/CromwellRunnerAppInstall.scala
mspector: Good questions - Cromwell DB doesn't seem to have a longer name, but maybe we could give it one. 

I don't have a good intuition about whether this could cause problems. I'll investigate...

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreterSpec.scala
mspector: from what I can tell, that kind of check is happening in the test classes for individual app installation implementations - in this case, `CromwellRunnerAppInstallSpec.scala` (see [here](https://github.com/DataBiosphere/leonardo/pull/3756/files#diff-abe5ada8e5f43528f489448e1cd0a263ea966e967a450bc111802ae8e55e406aR21))

I don't know if we have end-to-end coverage for this kind of plumbing, or if the changes to `CromwellRunnerAppInstallSpec` suffice...

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/app/CromwellRunnerAppInstall.scala
rtitle: Oh gosh, that's a good catch. This could end up passing the wrong DB to the chart. 🤦 

Thinking..

rtitle:
> @mspector and @rtitle I was a little surprised we have to go back to WSM. Don't we already go there once to fetch the database information to set up the postgres roles? Couldn't we store it and reuse it?

We go to WSM to create the databases+identities needed for the `CROMWELL_RUNNER` app. But the `cromwellmetadata` database was already created when the `WORKFLOWS` app was created. So Leo needs to retrieve the full name for that pre-existing database somehow. I thought about having Leo try to remember that information -- but it felt cleaner to use WSM as the source of truth here.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/app/CromwellRunnerAppInstall.scala
rtitle: Trying something -- @mspector do you mind if I push a commit to this PR ?

mspector: not at all, thanks @rtitle

rtitle: Chatted offline with @mspector -- I was thinking of ways to be more generic about these DB names that don't require testing prefixes with `startsWith`. But it's tricky -- will continue to think about this outside the scope of this PR, but @mspector will push the simpler/more direct fix for now.

mspector:
jenkins retest

mspector:
jenkins retest

rtitle (approved):
Looks good!

mspector:
jenkins retest

PR #3766, opened 2023-09-29
doc(IA-4597): update README environment setup steps

cpate4:
This also introduces a `Brewfile` and `.sdkmanrc` for the repo.

__Jira ticket__: https://broadworkbench.atlassian.net/browse/IA-4597

### What

- Consolidate, update, and automate (as much as was possible) the environment setup for `leonardo`

### Why

- My goal is to streamline the environment setup so anyone from new-hires to internal-broadies looking to get `leonardo` spun up can have a seamless experience with little to no team involvement.
- My hope is that by introducing a standard developer setup experience that it will make it easier for the team to help debug issues while also providing a consistent experience for anyone interacting with `leonardo`
- Where things don't make sense to install (e.g. `sdkman` because you already have both a `jdk` and `sbt` installed and configured), the documentation can accommodate that and the team will be versed in supporting those nuances.

### What to test

- I extracted these steps the best I could from memory and my terminal history. The plan is to wipe my machine of the installation (as best I can) and go back through these steps to ensure they are "working" and that I have captured them to the best of my ability. 
- Anyone else that would like to do this is also greatly encouraged.

### Who tested and where

- [ ] I validated this change LOCALLY 


Brewfile.lock.json
```diff
@@ -0,0 +1,415 @@
+{
```
cpate4, line 1: This is based on the `brew bundle` execution - if anyone else runs it subsequently, they will get these versions. Happy to peg versions in the `Brewfile` if need be.

README.md
```diff
@@ -47,16 +56,149 @@ class LeonardoClient(leonardoBasePath: String) {
...
+  # this assumes a default install location for `brew`
+  echo 'eval "$(/opt/homebrew/bin/brew shellenv)"' >> .zprofile
+  ```
+  After the `brew` install, validate that you have a working `brew` installation be executing `brew help`.
+  Feel free to [turn off Anonymous Analytics](https://docs.brew.sh/Analytics) with the following command:
+  ```sh
+  brew analytics off
+  ```
+
+- Following the `brew`-install, you should be able to install the necessary tools to setup the remaining dependencies.
```
LizBaldo, line 120: Ohhh this is really nice, I never even thought of checking if brew could build deterministic environments like poetry does 😅 this is super cool, thanks for making this for us ❤️

lucymcnatt (approved):
This is super valuable, thanks for creating it!!

README.md
```diff
@@ -2,34 +2,43 @@
 
 # Leonardo
 
-Leonardo serves as a way to launch compute within the Terra security boundary. It does so via multiple different cloud hardware virtualization mechanisms, currently leveraging only the Google Cloud Platform.
+`leonardo` serves as a way to launch compute within the Terra security boundary. 
+It does so via multiple different cloud hardware virtualization mechanisms, currently leveraging only the Google Cloud Platform.
```
mlilynolting: Not necessarily the scope of your PR, but this is outdated - we offer GCP and Azure cloud services.

cpate4: I updated this to reflect both GCP and Azure

cindy-broadinstitute (approved):
Ran through it a couple times and now it looks good!

Huge improvement, this should make it much faster for the next onboarders.  

There is still room for the next person to pick up a couple of the 'if you get this error do this' and improve them but don't want to distract from how improved this version is.

PR #3772, opened 2023-09-29
IA-4335 Update `dateAccessed` for GCP apps

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-4335

Today, `dateAccessed` is never updated after app creation. This PR fixes the bug. This PR is important for supporting auto-pause for APPs down the line

Tested e2e in BEE as well
```
mysql> select appName, createdDate, dateAccessed from APP;
+--------------+----------------------------+----------------------------+
| appName      | createdDate                | dateAccessed               |
+--------------+----------------------------+----------------------------+
| qi-sas-825-1 | 2023-09-28 21:16:27.160475 | 2023-09-29 13:50:34.412987 |
+--------------+----------------------------+----------------------------+
1 row in set (0.00 sec)
```

- [ ] I validated this change 


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/DateAccessedUpdater.scala
```diff
@@ -77,7 +79,17 @@ object DateAccessedUpdater {
...
+sealed abstract class UpdateTarget extends Product with Serializable
+object UpdateTarget {
+  final case class Runtime(runtimeName: RuntimeName) extends UpdateTarget {
+    override def toString: String = s"runtime/${runtimeName.asString}"
+  }
+  final case class App(appName: AppName) extends UpdateTarget {
+    override def toString: String = s"app/${appName.value}"
+  }
+}
+final case class UpdateDateAccessMessage(updateTarget: UpdateTarget, cloudContext: CloudContext, dateAccessd: Instant) {
```
mlilynolting: can this typo (dateAccessd) be fixed too? I keep noticing these 😅

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/DateAccessedUpdater.scala
```diff
@@ -45,21 +45,23 @@ class DateAccessedUpdater[F[_]](
...
+    case UpdateTarget.App(appName) =>
+      appQuery.updateDateAccessed(appName, msg.dateAccessd).transaction.void
+  }
 }
 
 object DateAccessedUpdater {
   implicit val updateDateAccessMessageOrder: Ordering[UpdateDateAccessMessage] =
     Ordering.fromLessThan[UpdateDateAccessMessage] { (msg1, msg2) =>
-      if (msg1.cloudContext == msg2.cloudContext && msg1.runtimeName == msg2.runtimeName)
+      if (msg1.cloudContext == msg2.cloudContext && msg1.toString == msg2.toString)
```
mlilynolting: since UpdateTarget's members are case classes, they have `equals` and `hashCode` already defined, right? I think it'd be more stable to use `msg1.cloudContext == msg2.cloudContext && msg1.updateTarget == msg2.updateTarget` here if that's true

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/DateAccessdUpdaterSpec.scala
mlilynolting, line 1: lol, how did the missing E get into the filename too?

mlilynolting (approved):
no blockers, just nice-to-haves

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/AppComponent.scala
```diff
@@ -239,6 +239,11 @@ object appQuery extends TableQuery(new AppTable(_)) {
       .map(a => (a.status, a.destroyedDate, a.diskId))
       .update((AppStatus.Deleted, now, None))
 
+  def updateDateAccessed(appName: AppName, now: Instant): DBIO[Int] =
```
rtitle: Just thinking.. is there any way this could use a DB index to query the app? As-is it's probably scanning the whole APP table. I guess it's tricky with how the tables are set up today..

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/DateAccessedUpdater.scala
```diff
@@ -45,21 +45,23 @@ class DateAccessedUpdater[F[_]](
...
+  private def updateDateAccessed(msg: UpdateDateAccessMessage): F[Unit] = msg.updateTarget match {
+    case UpdateTarget.Runtime(runtimeName) =>
+      metrics.incrementCounter("jupyterAccessCount") >>
+        clusterQuery
+          .clearKernelFoundBusyDateByProjectAndName(msg.cloudContext, runtimeName, msg.dateAccessd)
+          .flatMap(_ => clusterQuery.updateDateAccessedByProjectAndName(msg.cloudContext, runtimeName, msg.dateAccessd))
+          .transaction
+          .void
+    case UpdateTarget.App(appName) =>
+      appQuery.updateDateAccessed(appName, msg.dateAccessd).transaction.void
```
rtitle: Maybe add an `appAccessCount` metric?

rtitle (commented):
Looks fine.. I think there are some unit test failures though.

I'm not sure the solution, but it'd be nice if APP table queries by the Leo proxy (which happen frequently) could use DB indexes.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/AppComponent.scala
mlilynolting: looks like starting with the nonDeleted filter is what's missing the index - `destroyedDate` is in the indexes but only 2nd or 3rd down the list, and `status` isn't at all I think. if we started with `appName` we'd be better off

Qi77Qi: yea, I was a bit suspicious when looking at this..but didn't feel like fixing the existing problem....will take another look

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/DateAccessedUpdater.scala
Qi77Qi: I was thinking the jupyter one isn't very useful, so didn't add one for app either...are you thinking it'd be good to just see the overall traffic?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/DateAccessedUpdater.scala
Qi77Qi: good point :)

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/DateAccessdUpdaterSpec.scala
Qi77Qi: ?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/AppComponent.scala
Qi77Qi: so I think these are the only indexes exist today
```
  PRIMARY KEY (`id`),
  UNIQUE KEY `IDX_APP_UNIQUE` (`appName`,`nodepoolId`,`workspaceId`,`destroyedDate`),
```

not sure what would be good index to add..the real meaningful uniqueness would be (`appName`, `workspaceId`) or `appName`, `cloudContext`)...

Anyways, it's probably not an issue today since we only have 3385 records in prod DB...we don't update `dateAccessed` for every proxy request either, we batch update every 30 seconds...

Another good motivation to rethink DB schema.

Qi77Qi:
@rtitle r u ok with the current state for now? regarding index, I think it'll require a bit more rethinking on app table schemas in general

rtitle (approved):

PR #3782, opened 2023-10-02
[TOAZ-335] Use batch subnet resource id rather than subnet name

jsaun:
__Jira ticket__: https://broadworkbench.atlassian.net/browse/TOAZ-335

### What

- Bug fix: The helm chart for TES needs the resource id of the batch subnet, not the name of the subnet to work. 

### What to test

- 

### Who tested and where

- [ ] I validated this change 


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpWsmDao.scala
```diff
@@ -285,6 +285,22 @@ class HttpWsmDao[F[_]](httpClient: Client[F], config: HttpWsmDaoConfig)(implicit
         AppCreationException(s"could not determine name for resource $resource")
       )
 
+  private def getLandingZoneResourceId(
+    landingZoneResourcesByPurpose: Map[(LandingZoneResourcePurpose, String), List[LandingZoneResource]],
+    resourceType: String,
+    purpose: LandingZoneResourcePurpose,
+    useParent: Boolean
+  ): F[String] =
+    landingZoneResourcesByPurpose
```
rtitle: Could this reuse code in `getLandingZoneResource` and be written as:
```
getLandingZoneResource(landingZoneResourcesByPurpose, resourceType, purpose)
  .flatMap(_.resourceId)
```
?

jsaun: This didn't exactly work, but I was able to rewrite it in a way that uses getLandingZoneResource() instead of querying landingZoneResourcesByPurpose directly.

rtitle (approved):
Looks good - thanks

PR #3801, opened 2023-10-04
RW-11067 small table optimization

Qi77Qi:
https://precisionmedicineinitiative.atlassian.net/browse/RW-11067

1. Drop `dateAccessed` in `NODEPOOL` table. I don't think this field is needed. And we currently never update it anyway
2. add `namespaceName` to `APP` table. I don't think namespace needs a separate table, and I'll have a follow-up PR to drop it once this PR gets to prod.

Test:
Verified the migration looks good by running locally against my instance of leo DB

### What

-

### Why

-

### What to test

- 

### Who tested and where

- [ ] I validated this change 

### DIFF
--- a/http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changelog.xml
+++ b/http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changelog.xml
@@ -106,4 +106,5 @@
     <include file="changesets/20230713_app_controlled_resource_constraints.xml" relativeToChangelogFile="true" />
     <include file="changesets/20230728_app_type_rename.xml" relativeToChangelogFile="true" />
     <include file="changesets/20230911_create_app_usage_table.xml" relativeToChangelogFile="true" />
+    <include file="changesets/20231002_drop_columns_in_nodepool_table.xml" relativeToChangelogFile="true" />
 </databaseChangeLog>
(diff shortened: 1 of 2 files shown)


mlilynolting:
Will dropping `dateAccessed` on nodepool cause issues for the CRUD operations which act on that field? eg. `fromNodepool`, `unmarshalNodepool`, its definition in `NodepoolTable`, `NodepoolRecord` etc. If the dateAccessed fields here are all `null` will that just work? Will having one fewer column than its definition in the Component file cause issues when it's tupled?

mlilynolting:
Are either of these data structures used by Azure resources? Does your local db have Azure data?

mlilynolting:
Right now it looks like namespace records are foreign-keyed to the KUBERNETES_CLUSTER table, not APP. Is app:k8sCluster a 1:1 relationship? What's the rationale for moving the field?

Is your coming refactor going to cover all the places where the Namespace table is expected, and namespaceName is used? What is validation for this going to look like?

Namespace table also has a uniqueness constraint on the name; would it make sense to add that here?

mlilynolting:
Lots of comments for such a small PR 😅 I don't have Jira access to RW-11067 so idk if any of that is answered in the ticket.

EDIT lol, I was looking at Broad's Jira; didn't realize RW tickets were elsewhere :)

Qi77Qi:
> Will dropping dateAccessed on nodepool cause issues for the CRUD operations which act on that field?

Yes...I haven't updated the actual code referencing the field yet..should probably do that before PRing...but mostly for now, I wanna get some feedback if dropping it is reasonable to ppl first


> Are either of these data structures used by Azure resources? 

Azure apps populates namespace as well. 


> Does your local db have Azure data?

No. But I don't think there's any difference in the fields affected for Azure vs GCP apps


> Right now it looks like namespace records are foreign-keyed to the KUBERNETES_CLUSTER table, not APP. Is app:k8sCluster a 1:1 relationship? What's the rationale for moving the field?

You mean namespace table references `clusterId`? I don't think dropping it will affect the table that it's referencing. It's the table that's referencing this `namespace` will be affected, which is `APP`...but that'll all in a future PR


>Is your coming refactor going to cover all the places where the Namespace table is expected, and namespaceName is used? What is validation for this going to look like?

we'll see...I don't have the PR yet :)


> Namespace table also has a uniqueness constraint on the name; would it make sense to add that here?

That's a good point. Need to think abt this a bit


> I don't have Jira access to RW-11067 so idk if any of that is answered in the ticket.

There's nothing on the ticket :) ..this is mostly me trying to make some opportunistic DB schema improvements around apps.

Qi77Qi:
I'm going to close this PR... @mlilynolting had a good point that namespaces are unique at cluster level, but today, our implementation is creating new namespace for every app I think, which can be optimized. So instead of moving namespace to app table, I'll think about improving namespace creation...

PR #3802, opened 2023-10-05
[IA-4295] Stub SAM client

mlilynolting:
__Jira ticket__: https://broadworkbench.atlassian.net/browse/IA-4295

### What

- A `sam` client for `leonardo`! At long last. Not yet actually used, just initialized with a couple of methods that seem useful for my current work.

### Why

- Leo rolls its own Sam client, and it's hard to read for Scala newbies. The published API client for Sam is robust, so why not use it? Also, `listResourceTypes` would allow Leonardo to stop modeling Sam permissions itself (in `samModels` and `LeoAuthProvider`).

### What to test

- 

### Who tested and where

- [ ] I validated this change 

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/SamClient.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/SamClient.scala
@@ -1,0 +1,49 @@
+package org.broadinstitute.dsde.workbench.leonardo
+package dao
+
+import org.broadinstitute.dsde.workbench.client.sam.ApiClient
+import org.broadinstitute.dsde.workbench.client.sam.api.ResourcesApi
+import org.broadinstitute.dsde.workbench.client.sam.api.ConfigApi
+import org.broadinstitute.dsde.workbench.client.sam.model.{ResourceType, UserResourcesResponse}
+
+// Implicit for .asScala
+import scala.jdk.CollectionConverters.ListHasAsScala
+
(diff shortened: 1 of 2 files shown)


cpate4:
NOOB question ...
What is the difference between the client you are proposing here and the one that is contained in this package:
[`org.broadinstitute.dede.workbench.leonardo.consumer`](https://github.com/DataBiosphere/leonardo/blob/57b3c3c12fd54f9c87d35494beecf4191f77d342/pact4s/src/main/scala/org/broadinstitute/dede/workbench/leonardo/consumer/SamClient.scala#L16)

mlilynolting:
> NOOB question ... What is the difference between the client you are proposing here and the one that is contained in this package: [`org.broadinstitute.dede.workbench.leonardo.consumer`](https://github.com/DataBiosphere/leonardo/blob/57b3c3c12fd54f9c87d35494beecf4191f77d342/pact4s/src/main/scala/org/broadinstitute/dede/workbench/leonardo/consumer/SamClient.scala#L16)

@cpate4 I'm not an expert on this area but since this is a class under `/pact4s` I had been thinking of it as a test class, exclusively for contract testing. But it could well be that these classes are supposed to be used for application code as well. I will ask in the #dsp-contract-testing channel.

mlilynolting:
I have answers from the contract testing team! See https://broadinstitute.slack.com/archives/C043YJ40719/p1696534434946319?thread_ts=1696520859.026839&cid=C043YJ40719

tl;dr is to make the base class in prod code, and have a thin mock extension of it available to test code.

PR #3814, opened 2023-10-10
IA-4607: Query LZ Service in Back Leo for runtime and app creation

rtitle:
https://broadworkbench.atlassian.net/browse/IA-4607

## Summary of changes

We observed during [Terra on Azure Load Testing](https://docs.google.com/document/d/1s2bzXs8szI38mYW_rsiawcurohj1lTbKmw6U7GxZVJw/edit) that Front Leo was crashing during many (~80) simultaneous app creations. A major contributor seemed to be the LZ query, which makes a bunch of Azure cloud calls. This PR moves that call to Back Leo, and also adds a cache in front of it.

### What

Before, Front Leo was doing all this in the scope of the `create[Runtime|App]` request:
1. Query WSM `getWorkspace` to get the workspace cloud context
2. Query LZ `getLandingZoneResources` to get the LZ information
3. Query WSM `enumerateResources` to get shared workspace storage container

After this PR, only (1) now happens in the scope of the request. (2) and (3) happen in Back Leo. Additionally, (2) now has a cache since the LZ is immutable.

### Why

- Better support workshop scenarios, when many runtime and app creations may happen simultaneously.

### What to test

TODO: 

- ✅ Unit tests pass
- ✅ Tested runtime and app creation/use/deletion on a BEE
- Tested app update on a BEE
- Scale test 

### Who tested and where

- [ ] I validated this change 

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/runtimeModels.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/runtimeModels.scala
@@ -316,7 +316,7 @@ object RuntimeConfig {
   // Azure machineType maps to `com.azure.resourcemanager.compute.models.VirtualMachineSizeTypes`
   final case class AzureConfig(machineType: MachineTypeName,
                                persistentDiskId: Option[DiskId],
-                               region: com.azure.core.management.Region
+                               region: Option[RegionName]
   ) extends RuntimeConfig {
     val cloudService: CloudService = CloudService.AzureVm
     val configType: RuntimeConfigType = RuntimeConfigType.AzureVmConfig
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpWsmDao.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpWsmDao.scala
@@ -33,10 +33,14 @@ import org.http4s.client.dsl.Http4sClientDsl
 import org.http4s.headers.{`Content-Type`, Authorization}
 import org.typelevel.ci.CIString
 import org.typelevel.log4cats.StructuredLogger
+import scalacache.Cache
 
 import java.util.UUID
 
-class HttpWsmDao[F[_]](httpClient: Client[F], config: HttpWsmDaoConfig)(implicit
+class HttpWsmDao[F[_]](httpClient: Client[F],
+                       config: HttpWsmDaoConfig,
+                       lzResourcesCache: Cache[F, BillingProfileId, LandingZoneResources]
+)(implicit
   logger: StructuredLogger[F],
   F: Async[F],
   metrics: OpenTelemetryMetrics[F]
@@ -125,7 +129,12 @@ class HttpWsmDao[F[_]](httpClient: Client[F], config: HttpWsmDaoConfig)(implicit
       )(onError)
     } yield res
 
-  override def getLandingZoneResources(billingProfileId: String, userToken: Authorization)(implicit
+  override def getLandingZoneResources(billingProfileId: BillingProfileId, userToken: Authorization)(implicit
+    ev: Ask[F, AppContext]
+  ): F[LandingZoneResources] =
+    lzResourcesCache.cachingF(billingProfileId)(None)(getLandingZoneResourcesInternal(billingProfileId, userToken))
+
+  private def getLandingZoneResourcesInternal(billingProfileId: BillingProfileId, userToken: Authorization)(implicit
     ev: Ask[F, AppContext]
   ): F[LandingZoneResources] =
     for {
@@ -297,7 +306,7 @@ class HttpWsmDao[F[_]](httpClient: Client[F], config: HttpWsmDaoConfig)(implicit
         )
     } yield id
 
-  private def getLandingZone(billingProfileId: String, authorization: Authorization)(implicit
+  private def getLandingZone(billingProfileId: BillingProfileId, authorization: Authorization)(implicit
     ev: Ask[F, AppContext]
   ): F[Option[LandingZone]] =
     for {
@@ -307,7 +316,7 @@ class HttpWsmDao[F[_]](httpClient: Client[F], config: HttpWsmDaoConfig)(implicit
           method = Method.GET,
           uri = config.uri
             .withPath(Uri.Path.unsafeFromString("/api/landingzones/v1/azure"))
-            .withQueryParam("billingProfileId", billingProfileId),
+            .withQueryParam("billingProfileId", billingProfileId.value),
           headers = headers(authorization, ctx.traceId, withBody = false)
         )
       )(onError)
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpWsmDao.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpWsmDao.scala
@@ -33,14 +33,10 @@ import org.http4s.client.dsl.Http4sClientDsl
 import org.http4s.headers.{`Content-Type`, Authorization}
 import org.typelevel.ci.CIString
 import org.typelevel.log4cats.StructuredLogger
-import scalacache.Cache
 
 import java.util.UUID
 
-class HttpWsmDao[F[_]](httpClient: Client[F],
-                       config: HttpWsmDaoConfig,
-                       lzResourcesCache: Cache[F, BillingProfileId, LandingZoneResources]
-)(implicit
+class HttpWsmDao[F[_]](httpClient: Client[F], config: HttpWsmDaoConfig)(implicit
   logger: StructuredLogger[F],
   F: Async[F],
   metrics: OpenTelemetryMetrics[F]
@@ -130,11 +126,6 @@ class HttpWsmDao[F[_]](httpClient: Client[F],
     } yield res
 
   override def getLandingZoneResources(billingProfileId: BillingProfileId, userToken: Authorization)(implicit
-    ev: Ask[F, AppContext]
-  ): F[LandingZoneResources] =
-    lzResourcesCache.cachingF(billingProfileId)(None)(getLandingZoneResourcesInternal(billingProfileId, userToken))
-
-  private def getLandingZoneResourcesInternal(billingProfileId: BillingProfileId, userToken: Authorization)(implicit
     ev: Ask[F, AppContext]
   ): F[LandingZoneResources] =
     for {
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/RuntimeConfigTable.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/RuntimeConfigTable.scala
@@ -1,13 +1,13 @@
 package org.broadinstitute.dsde.workbench.leonardo
 package db
 
-import java.sql.SQLDataException
-import java.time.Instant
-import com.azure.core.management.Region
 import org.broadinstitute.dsde.workbench.google2.{MachineTypeName, RegionName, ZoneName}
 import org.broadinstitute.dsde.workbench.leonardo.db.LeoProfile.api._
 import org.broadinstitute.dsde.workbench.leonardo.db.LeoProfile.mappedColumnImplicits._
 import org.broadinstitute.dsde.workbench.leonardo.db.RuntimeConfigTable.toRuntimeConfig
+
+import java.sql.SQLDataException
+import java.time.Instant
 
 class RuntimeConfigTable(tag: Tag) extends Table[RuntimeConfigRecord](tag, "RUNTIME_CONFIG") {
   def id = column[RuntimeConfigId]("id", O.PrimaryKey, O.AutoInc)
@@ -175,8 +175,7 @@ class RuntimeConfigTable(tag: Tag) extends Table[RuntimeConfigRecord](tag, "RUNT
              None,
              r.persistentDiskId,
              None,
-             // TODO: should this be generalized to a type above regionName?
-             Some(RegionName(r.region.toString)),
+             r.region,
              (None, None),
              false,
              false
@@ -260,9 +259,7 @@ object RuntimeConfigTable {
         RuntimeConfig.AzureConfig(
           machineType,
           persistentDiskId,
-          Region.fromName(
-            region.getOrElse(throw new SQLDataException("region field should not be null for Azure.")).value
-          )
+          region
         )
     }
 }
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
@@ -250,7 +250,7 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
       app <- appQuery.save(saveApp, Some(ctx.traceId)).transaction
 
       clusterNodepoolAction = saveClusterResult match {
-        case ClusterExists(_) =>
+        case ClusterExists(_, _) =>
           // If we're using a pre-existing nodepool then don't specify CreateNodepool in the pubsub message
           if (userNodepoolOpt.isDefined) None else Some(ClusterNodepoolAction.CreateNodepool(nodepool.id))
         case ClusterDoesNotExist(c, n) => Some(ClusterNodepoolAction.CreateClusterAndNodepool(c.id, n.id, nodepool.id))
@@ -655,27 +655,10 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
           )
         else F.unit
 
-      // Save or retrieve a nodepool record for the app
-      userNodepoolOpt <- nodepoolQuery
-        .getMinimalByUserAndConfig(originatingUserEmail, cloudContext, machineConfig)
-        .transaction
-      clusterId = saveClusterResult.minimalCluster.id
-      nodepool <- userNodepoolOpt match {
-        case Some(n) =>
-          log.info(ctx.loggingCtx)(
-            s"Reusing user's nodepool ${n.id} in ${saveClusterResult.minimalCluster.cloudContext.asStringWithProvider} with ${machineConfig}"
-          ) >> F.pure(n)
-        case None =>
-          for {
-            _ <- log.info(ctx.loggingCtx)(
-              s"No nodepool with ${machineConfig} found for this user in project ${saveClusterResult.minimalCluster.cloudContext.asStringWithProvider}. Will create a new nodepool."
-            )
-            saveNodepool <- F.fromEither(
-              getUserNodepool(clusterId, cloudContext, originatingUserEmail, machineConfig, ctx.now)
-            )
-            savedNodepool <- nodepoolQuery.saveForCluster(saveNodepool).transaction
-          } yield savedNodepool
-      }
+      // Use the default nodepool of the cluster.
+      // The v1 createApp endpoint has logic for managing nodepools per user in Leonardo.
+      // In the v2 endpoint, nodepools are created upstream of Leonardo.
+      nodepool = saveClusterResult.defaultNodepool.toNodepool()
 
       // Retrieve a pet identity from Sam
       runtimeServiceAccountOpt <- serviceAccountProvider.getClusterServiceAccount(userInfo, cloudContext)
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
@@ -9,9 +9,8 @@ import cats.effect.Async
 import cats.effect.std.Queue
 import cats.mtl.Ask
 import cats.syntax.all._
+import monocle.macros.syntax.lens._
 import org.apache.commons.lang3.RandomStringUtils
-import com.azure.core.management.Region
-import org.broadinstitute.dsde.workbench.azure.AKSClusterName
 import org.broadinstitute.dsde.workbench.google2.GKEModels.{KubernetesClusterName, NodepoolName}
 import org.broadinstitute.dsde.workbench.google2.KubernetesSerializableName.NamespaceName
 import org.broadinstitute.dsde.workbench.google2.{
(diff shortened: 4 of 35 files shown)


rtitle:
jenkins retest

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/runtimeModels.scala
```diff
@@ -316,7 +316,7 @@ object RuntimeConfig {
   // Azure machineType maps to `com.azure.resourcemanager.compute.models.VirtualMachineSizeTypes`
   final case class AzureConfig(machineType: MachineTypeName,
                                persistentDiskId: Option[DiskId],
-                               region: com.azure.core.management.Region
+                               region: Option[RegionName]
```
rtitle, line 319: When Front Leo persists the runtime record, it won't know the region (since that is represented in the LZ). I made region optional in this model (which matches the DB schema where it's also nullable).

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpWsmDao.scala
```diff
@@ -125,7 +129,12 @@ class HttpWsmDao[F[_]](httpClient: Client[F], config: HttpWsmDaoConfig)(implicit
       )(onError)
     } yield res
 
-  override def getLandingZoneResources(billingProfileId: String, userToken: Authorization)(implicit
+  override def getLandingZoneResources(billingProfileId: BillingProfileId, userToken: Authorization)(implicit
+    ev: Ask[F, AppContext]
+  ): F[LandingZoneResources] =
+    lzResourcesCache.cachingF(billingProfileId)(None)(getLandingZoneResourcesInternal(billingProfileId, userToken))
```
rtitle: Added a cache in front of the LZ resources query, since it's an expensive call and LZs are immutable.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
```diff
@@ -600,28 +596,14 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
...
       // Validate the machine config from the request
       // For Azure: we don't support setting a machine type in the request; we use the landing zone configuration instead.
       // For GCP: we support setting optionally a machine type in the request; and use a default value otherwise.
       machineConfig <- (cloudContext.cloudProvider, req.kubernetesRuntimeConfig) match {
         case (CloudProvider.Azure, Some(_)) =>
           F.raiseError(AppMachineConfigNotSupportedException(ctx.traceId))
-        // TODO: pull this from the landing zone instead of hardcoding once TOAZ-232 is implemented
         case (CloudProvider.Azure, None) =>
-          F.pure(KubernetesRuntimeConfig(NumNodes(1), MachineTypeName("Standard_A2_v2"), false))
+          F.pure(KubernetesRuntimeConfig(NumNodes(1), MachineTypeName("unset"), false))
```
rtitle, line 619: `Standard_A2_v2` was a lie anyway. :(
The nodepool machine type is defined in the LZ, just marking it as `unset` in the Leo DB.

This is an example where the Leo DB schema is overloaded to support GCP and Azure, which have different models.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/MonitorAtBoot.scala
```diff
@@ -416,60 +399,46 @@ class MonitorAtBoot[F[_]](publisherQueue: Queue[F, LeoPubsubMessage],
                               MonitorAtBootException(s"no workspaceId found for ${runtime.id.toString}", traceId)
           )
           controlledResourceOpt = WsmControlledResourceId(UUID.fromString(runtime.internalId))
-          leoAuth <- samDAO.getLeoAuthToken
-          workspaceDescOpt <- wsmDao.getWorkspace(wid, leoAuth)
```
rtitle, line 420: I don't _think_ this was right before -- it was querying WSM for the workspace using Leo's token when it should be the user's token.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSAlgebra.scala
```diff
@@ -35,25 +33,6 @@ final case class UpdateAKSAppParams(appId: AppId,
...
   appName: AppName,
   workspaceId: WorkspaceId,
-  landingZoneResourcesOpt: LandingZoneResources,
-  cloudContext: AzureCloudContext
+  cloudContext: AzureCloudContext,
+  billingProfileId: BillingProfileId
 )
-
-/** Enumerates the possible identity modes for an AKS app. */
-sealed trait IdentityType
```
rtitle, line 43: I thought I removed this in a prior PR… it is obsolete, we are using WorkloadIdentity across the board now.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -74,24 +75,6 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
   private def getListenerReleaseName(appReleaseName: Release): Release =
     Release(s"${appReleaseName.asString}-listener-rls")
 
-  private def retrieveWsmReferenceDatabases(resourceApi: ResourceApi,
```
rtitle, line 77: Just moved this method down in the file, next to the other private DB-related methods.

rtitle:
BEE seeding failed.. jenkins retest

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/runtimeModels.scala
Qi77Qi: oh, I think the original assumption is region comes from user..but I guess it's not true anymore becuz it actually comes from LZ now

cc @cindy-broadinstitute in case this affects ur change

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
Qi77Qi: any value in updating it after the runtime is created in leo DB?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
```diff
@@ -874,18 +838,15 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
...
-    // regionName can be empty in some test configurations
-    val regionName = if (azureRegionOpt.isEmpty) "" else azureRegionOpt.map(_.name).getOrElse("")
     for {
       nodepool <- defaultNodepool
       defaultClusterName <- KubernetesNameUtils.getUniqueName(KubernetesClusterName.apply)
-      clusterName = aksClusterName.map(c => KubernetesClusterName(c.value)).getOrElse(defaultClusterName)
     } yield SaveKubernetesCluster(
       cloudContext = cloudContext,
-      clusterName = clusterName,
+      clusterName = defaultClusterName,
```
Qi77Qi, line 842: AKS's cluster name is always the same becuz there will only be one cluster per LZ?

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/runtimeModels.scala
rtitle: Yeah, it's unfortunately a difference in GCP vs Azure.

* For GCP, it's specified in the request (with a configurable default), so we can persist it in Front Leo.
* For Azure, it's not specified in the request, and it always follows the LZ region.

We _could_ update the Leo DB after we query the LZ in back Leo -- this PR is not currently doing that though (and just leaving it null).

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
rtitle: So this is actually the AKS nodepool machine type for apps. We could update it in back Leo, but it would just be informational, since we don't rely on this field for AKS apps.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
rtitle: Yeah there's 1 cluster per LZ. And actually `defaultClusterName` will be a random string but not the actual cluster name.  😬 

This is another place where we could update the Leo DB in back Leo after we query the LZ. (But again, Leo does not rely on this field for Azure.)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpWsmDao.scala
LizBaldo: Just a thought, but I am wondering if it would be worth putting this as a comment into the code, it would help a lot I think

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/runtimeModels.scala
rtitle: Hmm.. does Terra UI actually rely on the runtime region in the API response? I see this in my BEE:

image

I'll look at this more, and look into potentially setting the region to the correct value..

Qi77Qi: yea, I think Cindy worked on this recently

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/RuntimeConfigTable.scala
```diff
@@ -260,9 +259,7 @@ object RuntimeConfigTable {
         RuntimeConfig.AzureConfig(
           machineType,
           persistentDiskId,
-          Region.fromName(
-            region.getOrElse(throw new SQLDataException("region field should not be null for Azure.")).value
-          )
+          region
```
LizBaldo, line 262: So we are now guarantied to get a region from the LZ?

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/runtimeModels.scala
cindy-broadinstitute: The terra-ui support for region is in dev, not in production yet.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
```diff
@@ -874,18 +838,15 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
...
       location = config.leoKubernetesConfig.clusterConfig.location,
       region =
-        if (cloudContext.cloudProvider == CloudProvider.Azure) RegionName(regionName)
+        if (cloudContext.cloudProvider == CloudProvider.Azure) RegionName("unset")
```
LizBaldo, line 845: So this feels like this will break the UI, and especially the work that @cindy-broadinstitute did recently. This means that the UI won't be able to display the app region anymore

LizBaldo (commented):
Moving the LZ query to back leo and caching makes a lot of sense, however as you already noted, we need to know the runtime and app region early on for the UI to work (calculating costs + displaying the region in the cloud environment page)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
cindy-broadinstitute: Yes, this will definitely break the UI. The RegionName (Region for Azure) is saved to the config here and used by the UI.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
```diff
@@ -647,12 +629,7 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
 
       // Save or retrieve a KubernetesCluster record for the app
       saveCluster <- F.fromEither(
-        getSavableCluster(userInfo.userEmail,
-                          cloudContext,
-                          ctx.now,
-                          landingZoneResourcesOpt.map(_.clusterName),
-                          landingZoneResourcesOpt.map(_.region)
```
cindy-broadinstitute, line 667: This was needed to persist the Region.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
rtitle: OK makes sense -- will try to fix.

It would be nice if we could pull the region from the WSM workspace (which we do still query in Front Leo). Will see if that's possible -- failing that, we could asynchronously update this column when we query the LZ in back leo.

Qi77Qi: I think UI probably knows abt the region since it calls WSM anyway...for future improvements tho

mlilynolting: I don't think pulling region from workspace at this point is viable, since there was a feature request to persist region at the app level

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpWsmDao.scala
dvoet: Is the cache bounded in any way or do items age out?

I am also concerned we are adding complexity to overfit for the workshop use case of lots of concurrent users in a single landing zone. Alternatively could we just make the call to get LZ info cheaper?

dvoet: Is the cache bounded in any way or do items age out?

I am also concerned we are adding complexity to overfit for the workshop use case of lots of concurrent users in a single landing zone. Alternatively could we just make the call to get LZ info cheaper?

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/runtimeModels.scala
rtitle: After conversation, I am thinking I will pull the region from WSM from the VM resource, and update the DB in Back Leo. (I haven't pushed that yet.)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpWsmDao.scala
rtitle: The cache size/expiry is defined in Boot.scala with numbers I picked out of thin air: https://github.com/DataBiosphere/leonardo/pull/3814/files#diff-8af5da8d5edb93397a2e5c102c53047173dac803b46343c12568196a90cee553R435-R438

```
underlyingLandingZoneCache = buildCache[BillingProfileId, scalacache.Entry[LandingZoneResources]](
        500,
        4 hours
      )
```

I don't think it's a huge cost in complexity, we use the same caching pattern for several things in Leo.

The one downside I was thinking about is if there's a destructive LZ upgrade which removes/renames things, making the cached value incorrect. But I hope we don't do those kinds of LZ updates in general -- they should be additive/backwards compatible.

rtitle: I don't actually care about the cache _that_ much though -- the main thing was moving it out of the path of the user request.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
```diff
@@ -647,12 +629,7 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
...
-        )
+        getSavableCluster(userInfo.userEmail, cloudContext, ctx.now)
       )
       saveClusterResult <- KubernetesServiceDbQueries.saveOrGetClusterForApp(saveCluster).transaction
```
rtitle, line 634: There's a DB optimization I want to make in here too which I haven't pushed yet.. will update this comment to explain.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
rtitle: Haven't pushed yet, but I am planning to pull the region from the WSM VM resource in back Leo.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpWsmDao.scala
dvoet: I have lost count of all my caching related scars. If you don't care that much about the caching, I would remove it and use this as an excuse to make the LZ api better.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/RuntimeConfigTable.scala
rtitle: Yes, we now persist the region when we create the VM in back leo.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
```diff
@@ -678,27 +655,10 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
...
-            _ <- log.info(ctx.loggingCtx)(
-              s"No nodepool with ${machineConfig} found for this user in project ${saveClusterResult.minimalCluster.cloudContext.asStringWithProvider}. Will create a new nodepool."
-            )
-            saveNodepool <- F.fromEither(
-              getUserNodepool(clusterId, cloudContext, originatingUserEmail, machineConfig, ctx.now)
-            )
-            savedNodepool <- nodepoolQuery.saveForCluster(saveNodepool).transaction
-          } yield savedNodepool
-      }
+      // Use the default nodepool of the cluster.
```
rtitle, line 658: Before, this code was using the v1 pattern where we create a new nodepool record for each unique user in a cluster. This was causing performance issues when there are many users in the same cluster, since each createApp request will query _all_ nodepools in the cluster. :( Moreover these nodepool records were unused for Azure apps, and did not get deleted when the app is deleted.

So I am just creating apps in the "default" nodepool of the cluster. Longer term, when we shift to v2 for GCP, we should revisit our DB schema and probably get rid of KUBERNETES_CLUSTER and NODEPOOL tables.

rtitle:
@LizBaldo  (cc @Qi77Qi @lucymcnatt @cindy-broadinstitute @mlilynolting ) I think I applied PR feedback -- mind doing another pass?

Runtime region should be persisted correctly now -- also verified general runtime/app functionality on a BEE.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
Qi77Qi: do we display cost for Azure? wonder if this info will be needed for cost display

Qi77Qi (approved):
lgtm

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
LizBaldo: We do display cost for Azure, were you able to confirm that the cost display did not break in your BEE UI @rtitle?

LizBaldo: But I am pretty sure it is fine becuse when the machine type is unset for Azure, we default to the cluster region

LizBaldo (commented):
Thanks for addressing the comments, and making a cool improvement to the nodepool query!

I can follow how the runtime VM region get's backfilled into the db, but I am still not clear on how the app region gets persisted to the db, is this via the LZ cluster and won't that break the UI?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
rtitle: I confirmed the cost appears in the UI -- I think it just displays runtime cost though, not app costs. Does the UI report on app (e.g. Cromwell, WDS) app costs?

rtitle: Yeah I think prod doesn't display app costs in the UI. E.g. here's a workspace in PROD with 2 apps and 0 runtimes:
image

So we could correctly set the node type + region for apps, but I'm sort of inclined to leave that for a separate PR if they are unused today.

rtitle: For completeness: my BEE displaying costs for runtimes :) 
image

rtitle: @LizBaldo pointed out to me that the app region is displayed in the Cloud Environments page. Here is my BEE:
image

Here is prod: 
image

Incidentally, prod is incorrect -- `us-central1` is a GCP region. I'll change it to fill in the correct region and machine type for apps as well.

rtitle: Update: fixed with my latest commit. In my BEE:

image

rtitle:
Tested runtime, app functionality in my BEE (including Cloud Envs page) on GCP and Azure. I think it's all working well. Merging.

PR #3815, opened 2023-10-11
IA-4607: Add blocking combinator

rtitle:
https://broadworkbench.atlassian.net/browse/IA-4607

## Summary of changes

Have you noticed Leo log messages like:
```
2023-10-09T17:06:22.499Z [WARNING] Your app's responsiveness to a new asynchronous event (such as a new connection, an upstream response, or a timer) was in excess of 100 milliseconds. Your CPU is probably starving. Consider increasing the granularity of your delays or adding more cedes. This may also be a sign that you are unintentionally running blocking I/O operations (such as File or InetAddress) without the blocking combinator.
```
?

This PR attempts to resolve that. There is more explanation in https://typelevel.org/cats-effect/docs/core/starvation-and-tuning, but it essentially means that cats-effect worker threads (the things that pull and execute `F`s off the stack) are all blocked on I/O and thus the program can't make progress. This PR fixes a few cases, which I'll note inline.

Side note: I don't have definitive proof, but I wonder if this is a contributing factor to some stability issues we've seen recently, such as the autopause monitor dying (https://broadworkbench.atlassian.net/browse/PROD-877), and the metrics collection process stopping (https://broadworkbench.atlassian.net/browse/IA-4601).

### What

- See inline comments for explanation

### Why

- Noticed this while load testing, suspecting it's contributing to Leo instability

### What to test

- Tested runtime + app creation/use/deletion on a BEE
- Unit tests pass
- Verified logs still work

### Who tested and where

- [ ] I validated this change 


http/src/main/resources/logback.xml
```diff
@@ -27,6 +27,10 @@
         </encoder>
     </appender>
 
+    <appender name="async-console" class="ch.qos.logback.classic.AsyncAppender">
```
rtitle, line 30: The cats-effect blocked-thread-detector was complaining about log4cats. I discovered we were using the synchronous appender, which does blocking file I/O. I changed it to an `AsyncAppender`, which writes log messages to a queue, and a background threads pulls and writes to disk.

project/Dependencies.scala
```diff
@@ -17,7 +17,7 @@ object Dependencies {
   val munitCatsEffectV = "1.0.7"
   val pact4sV = "0.10.0"
 
-  private val workbenchLibsHash = "7362eef"
+  private val workbenchLibsHash = "557248c4-SNAP"
```
Qi77Qi: should probably have the new lib published first?

Qi77Qi (approved):

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpSamDAO.scala
```diff
@@ -550,7 +550,7 @@ class HttpSamDAO[F[_]](httpClient: Client[F],
     ).use { credential =>
       val scopedCredential = credential.createScoped(saScopes.asJava)
 
-      F.delay(scopedCredential.refresh).map(_ => scopedCredential.getAccessToken)
+      F.blocking(scopedCredential.refresh).map(_ => scopedCredential.getAccessToken)
```
rtitle, line 553: Not sure how big a deal this is, but it is technically a network call to Google.

(`F.blocking` makes it do the task on a `cachedThreadPool`, not the main io-compute worker thread poll)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpSamDAO.scala
```diff
@@ -593,7 +593,7 @@ class HttpSamDAO[F[_]](httpClient: Client[F],
 
   private def getTokenFromKey(key: Json): F[String] = {
     val keyStream = new ByteArrayInputStream(key.toString().getBytes)
-    F.delay(ServiceAccountCredentials.fromStream(keyStream).createScoped(saScopes.asJava))
+    F.blocking(ServiceAccountCredentials.fromStream(keyStream).createScoped(saScopes.asJava))
```
rtitle, line 596: Also a Google network call

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dns/ProxyResolver.scala
```diff
@@ -52,7 +52,7 @@ object ProxyResolver {
         mapping <- hostToIpMapping.get
         // Use the IP if we have a mapping for it; otherwise fall back to default host name resolution
         h = mapping.get(host).map(_.asString).getOrElse(host)
-        res <- F.delay(new InetSocketAddress(h, port))
+        res <- F.blocking(new InetSocketAddress(h, port))
```
rtitle, line 55: I think this was kind of pathologically bad in Back Leo (though it affects the Front Leo proxy too). Here's the thread dump:
```
[WARNING] A Cats Effect worker thread was detected to be in a blocked state (BLOCKED)
at java.base@17.0.8.1/java.net.InetAddress$NameServiceAddresses.get(Unknown Source)
at java.base@17.0.8.1/java.net.InetAddress.getAllByName0(Unknown Source)
at java.base@17.0.8.1/java.net.InetAddress.getAllByName(Unknown Source)
at java.base@17.0.8.1/java.net.InetAddress.getAllByName(Unknown Source)
at java.base@17.0.8.1/java.net.InetAddress.getByName(Unknown Source)
at java.base@17.0.8.1/java.net.InetSocketAddress.<init>(Unknown Source)
at app//org.broadinstitute.dsde.workbench.leonardo.dns.ProxyResolver$ProxyResolverInterp.$anonfun$resolveInternal$4(ProxyResolver.scala:55)
at app//org.broadinstitute.dsde.workbench.leonardo.dns.ProxyResolver$ProxyResolverInterp$$Lambda$3984/0x00007fcb490022f0.apply(Unknown Source)
at app//cats.effect.IOFiber.runLoop(IOFiber.scala:255)
at app//cats.effect.IOFiber.execR(IOFiber.scala:1360)
at app//cats.effect.IOFiber.run(IOFiber.scala:112)
at app//cats.effect.unsafe.WorkerThread.run(WorkerThread.scala:632)
```

`InetSocketAddress` is a network call, and was being triggered a LOT by the metrics collection process (which http-pings every runtime/app). I have a hunch this was contributing to Back Leo instability.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/Boot.scala
```diff
@@ -869,6 +869,7 @@ object Boot extends IOApp {
         .withConnectTimeout(30 seconds)
         .withRequestTimeout(60 seconds)
         .withMaxTotalConnections(100)
+        .withMaxWaitQueueLimit(1024)
```
rtitle, line 872: This controls the size of the queue of connections int he http4s client (which is constrained by `withMaxTotalConnections` for the number that can be open concurrently). Before it was 120 -- I think it makes sense for the queue to be a bit larger than `withMaxTotalConnections`.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoMetricsMonitor.scala
```diff
@@ -364,7 +364,7 @@ class LeoMetricsMonitor[F[_]](config: LeoMetricsMonitorConfig,
                 val namespace = app.appResources.namespace
                 val labelSelector = s"leoAppName=${app.appName.value}"
                 for {
-                  pods <- F.delay(
+                  pods <- F.blocking(
```
rtitle, line 367: Blocking network call to the k8s API

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -78,7 +78,7 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
                                             referenceDatabaseNames: Set[String],
                                             workspaceId: UUID
   ): F[List[String]] = {
-    val wsmResourceDatabases = F.delay(
+    val wsmResourceDatabases = F.blocking(
```
rtitle, line 81: This was my mistake -- all these WSM calls through the Java client are blocking so should all be wrapped in `F.blocking`, I think.

project/Dependencies.scala
rtitle: TODO: update to non-SNAP

For https://github.com/broadinstitute/workbench-libs/pull/1554, which also replaces `F.delay` with `F.blocking` for some blocking Azure calls.

rtitle: +1

rtitle: @Qi77Qi want to review https://github.com/broadinstitute/workbench-libs/pull/1554 ? 😄

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dns/ProxyResolver.scala
Qi77Qi: didn't realize this tries to resolve the actual url 😵‍💫

http/src/main/resources/logback.xml
rtitle: (I have always used async appenders whenever using log4j/slf4j -- was surprised Leo wasn't)

LizBaldo (approved):
TIL, thank you! :shipit:

PR #3822, opened 2023-10-11
RW-10807 Restrict SAS to only AoU

Qi77Qi:
https://precisionmedicineinitiative.atlassian.net/browse/RW-10807

This PR does two things to restrict access
1. Make sure the request comes from AoU UI (label in createApp request)
2. Make sure the target project is an AoU project (by checking project label)

### Who tested and where

- [ ] I validated this change 


http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterpSpec.scala
```diff
@@ -1295,6 +1300,49 @@ final class AppServiceInterpSpec extends AnyFlatSpec with LeonardoTestSuite with
     }
   }
 
+  it should "reject create SAS app request if it's not from AoU UI" in {
```
LizBaldo, line 1303: What do you think about also having a happy path test to make sure that the request is accepted in case of an AOU UI label + high security group on the google project?

LizBaldo (approved):
LGTM, I am guessing that SAS is the first of many apps like this so it makes sense to look at the project security group instead of a SAS-specific tag. I just have a suggestion to add one more unit test.

yonghaoy (approved):

PR #3825, opened 2023-10-11
[IA-4604] Add metric for autopause

jdcanas:


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/AutopauseMonitor.scala
```diff
@@ -82,6 +82,7 @@ class AutopauseMonitor[F[_]](
         val traceId = TraceId(s"fromAutopause_${UUID.randomUUID().toString}")
         for {
           _ <- clusterQuery.updateClusterStatus(cl.id, RuntimeStatus.PreStopping, now).transaction
+          _ <- metrics.incrementCounter("autoPause/pauseRuntimeCounter")
```
cindy-broadinstitute, line 88: This will increment whenever we have something to stop/freeze. If instead we added a heartbeat (a metric at the top of the thread), then we could alert if we missed a few heartbeats, so 10 seconds of the monitor being down. With the 'autoPause/pauseRuntimeCounter', which is counting runtimes that have been paused, how long would it take for us to determine that the monitor had died?

cindy-broadinstitute: Update to my previous comment: I think we need this pauseRuntimeCounter metric as well as a heartbeat metric. Autopausing is a critical element and should be highly tracked. We need a metric of how many runtimes were paused as well as a heartbeat metric. Then we can alert on the monitor itself as quickly as possible. Adding two metrics will also let us catch the case here the monitor is possibly running, but, not finding runtimes to pause.

jdcanas: Yeah, I'll add the log for a heartbeat. The difference though is that we have a way to alert on a metric, and its pretty safe to assume that if there are no autopauses on a week day something fishy is going on (based on 6 months of historical data). I agree that the log message would be completely fail proof, but there is not as easy of a way to generate a metric based on the tooling I am familiar with, so it would require an additional spike (where as devops has can help generate a grafana based alert)

mlilynolting (approved):

mlilynolting:
✅, scalafmt tho

PR #3831, opened 2023-10-12
WM-2304: Support role=creator for azure listApps

cjllanwarne:
Add a `role=creator` option to Leonardo listAppsV2 in analogy to the option for GCP apps.

- Without the new option specified, all apps are returned, as before ✅ 
- With the new option specified, no apps are returned ✅


Qi77Qi (approved):
looking good 👍 ...just need to run scalafmt (u can config IJ so that it auto reformats on save)

lucymcnatt (approved):

PR #3832, opened 2023-10-13
RW-11059 Drop group check for AoU

Qi77Qi:
https://precisionmedicineinitiative.atlassian.net/browse/RW-11059

AoU will be ready to use consumption model after this PR.

This PR does the following
1. Drop sas_app_users group check for AoU users
2. Keep sas_app_users group check for non_AoU users. This is so that in the future, when we integrate SAS to Terra, devs with individual licenses can be added to this group and work on UI integration before consumption model for Terra is finalized.

Once this PR is merged, SAS will be available for all users in AoU Test environment

Related https://github.com/broadinstitute/terra-helmfile/pull/4701

### Who tested and where

- [ ] I validated this change 

### DIFF
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -1203,7 +1203,7 @@ app-service {
   # Defaults to true, but disabled for fiab runs
   enable-custom-app-check = true
   # Defaults to true, but for Production, we'll use Consumption model (details TBD as of 8/18/2023), which doesn't require Sam group check
-  enable-sas-group-app-check = true
+  enable-sas-app = false
 }
 
 drs {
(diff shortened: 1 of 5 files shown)


LizBaldo (approved):

yonghaoy (approved):
Do we have a feature flag to control SAS is allowed in higher env? If we remove this, does that mean AoU users in prod would be able to create SAS?

Qi77Qi:
@yonghaoy ah yes...I was adding the flag before I went for lunch...will push a bit more changes

Qi77Qi:
@yonghaoy @LizBaldo  

Repurposed `enable-sas-group-app-check` to `enable-sas-app` flag (added a unit test for this too). This flag is only true for `dev` with [this PR](https://github.com/broadinstitute/terra-helmfile/pull/4701)

PR #3841, opened 2023-10-16
[IA-4636] add some config to perhaps fix socket timeout

jdcanas:
I think its the leonardo java client timing out on polling operations. The java client should not be timing out, we have other timeouts in place. 

```
[info]   org.broadinstitute.dsde.workbench.client.leonardo.ApiException: Message: java.net.SocketTimeoutException: timeout
[info] HTTP response code: 0
[info] HTTP response body: null
[info] HTTP response headers: null
[info]   at org.broadinstitute.dsde.workbench.client.leonardo.ApiClient.execute(ApiClient.java:1124)
...
```

Full example run here: https://github.com/broadinstitute/terra-github-workflows/actions/runs/6511714509/job/17687928862#step:7:3924

### DIFF
--- a/automation/src/test/resources/reference.conf
+++ b/automation/src/test/resources/reference.conf
@@ -17,7 +17,7 @@ azure {
     workspaceId="e1aaf25b-b298-46eb-891b-e4c326f29b0c"
 }
 leonardoClient {
-    writeTimeout = 300000 //in ms, 5 minutes * 60 sec/min * 1000 ms/sec
-    readTimeout = 300000 //in ms, 5 minutes * 60 sec/min * 1000 ms/sec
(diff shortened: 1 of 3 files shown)


LizBaldo (approved):
Good to try, but is this something that we would eventually need to do in the http project as well?

jdcanas:
> Good to try, but is this something that we would eventually need to do in the http project as well?

I believe the client here is the `ApiClient` object which is a class generated and packaged with the leonardo client as part of the java lib we are using for code gen. It is not the same client we use for in-app API calls. See the stacktrace (which was in ticket and now is in PR description @LizBaldo ). We should let the consumer of this client decide how they want to configure their timeouts as I see it.

PR #3844, opened 2023-10-17
adding stop/start runtime test to azure swat tests

cindy-broadinstitute:
__Jira ticket__: https://broadworkbench.atlassian.net/browse/4471

### What

- Add a Start Runtime and Stop Runtime test to the automation tests

### Why

- eliminate the need to do manual testing

### What to test

- 
- 
Run the AzureRuntimeSpec

### Who tested and where

- [ ] I validated this change 


automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/azure/AzureRuntimeSpec.scala
```diff
@@ -95,6 +95,55 @@ class AzureRuntimeSpec
         )
         _ = monitorCreateResult.getStatus() shouldBe ClusterStatus.RUNNING
 
+        // Now stop the runtime
```
jdcanas, line 97: I think we should do the stop/start in a separate test where we don't delete the runtime to shorten the length a bit, since we can run this in parallel. Ideally we would be able to cut creation out of this time and have pre-created runtime since creation is tested elsewhere, but I think what I'm proposing is a good compromise. Thoughts?

cindy-broadinstitute: One test creates and deletes a runtime, and this test creates, stops, starts, and deletes the runtime. I would prefer that we eliminate the first test and have this more complete runtime test instead. And, in the Azure Test readiness meeting this morning, they mentioned that we should be cleaning up resources (deleting the runtime). But that might have been talking about the E2E tests. 
So, I don't see the advantage of having two tests, where the second one is a complete duplicate of the first with a couple added steps.
This second test takes only a few minutes longer than the original one.

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoSuite.scala
```diff
@@ -10,7 +10,8 @@ import org.broadinstitute.dsde.workbench.auth.AuthTokenScopes.billingScopes
 import org.broadinstitute.dsde.workbench.config.ServiceTestConfig
 import org.broadinstitute.dsde.workbench.leonardo.BillingProjectFixtureSpec._
 import org.broadinstitute.dsde.workbench.leonardo.TestUser.{Hermione, Ron}
-import org.broadinstitute.dsde.workbench.leonardo.azure.{AzureAutopauseSpec, AzureDiskSpec, AzureRuntimeSpec}
```
lucymcnatt, line 14: nit: this import should go in the list above

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/azure/AzureRuntimeSpec.scala
```diff
@@ -95,6 +95,55 @@ class AzureRuntimeSpec
...
+        _ <- loggerIO.info(
+          s"AzureRuntimeSpec: runtime ${workspaceId}/${runtimeName.asString} in stopping status detected"
+        )
+
+        // Verify the runtime eventually becomes Stopped
+        monitorStopResult <- streamUntilDoneOrTimeout(
+          callGetRuntime,
+          120,
+          10 seconds,
+          s"AzureRuntimeSpec: runtime ${workspaceId}/${runtimeName.asString} did not stop after 20 minutes"
```
lucymcnatt: this can probably be a lower number, I think it usually takes <5 to stop, so I'd maybe cap it at 10

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/azure/AzureRuntimeSpec.scala
```diff
@@ -95,6 +95,55 @@ class AzureRuntimeSpec
...
+        _ <- loggerIO.info(
+          s"AzureRuntimeSpec: runtime ${workspaceId}/${runtimeName.asString} in starting status detected"
+        )
+
+        // Verify the runtime eventually becomes Started
+        monitorStartResult <- streamUntilDoneOrTimeout(
+          callGetRuntime,
+          120,
+          10 seconds,
+          s"AzureRuntimeSpec: runtime ${workspaceId}/${runtimeName.asString} did not start after 20 minutes"
```
lucymcnatt: similar note here, 20 minutes I think is too generous

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/azure/AzureRuntimeSpec.scala
jdcanas: If its not adding a significant amount of time, I'm fine with including it. As long as we make sure we don't approach the max of 1 hour, when the token for the tests will expire

jdcanas (approved):

PR #3849, opened 2023-10-17
RW-11073 Add namespace column in APP table

Qi77Qi:
https://precisionmedicineinitiative.atlassian.net/browse/RW-11073

Reviving https://github.com/DataBiosphere/leonardo/pull/3801

For Azure apps, namespace is unique for each app today. We'll likely want to do the same for Google apps for 2 reasons
1. It'll allow us deleting app via deleting namespace. The current way of deleting apps rely on `helm uninstall` in GCP. This is mostly because Galaxy does custom things during helm uninstall, but all other apps don't do this, so we should be able to just delete namespace during app deletion. This will allow us to delete app without the `RUNNING` app constraint.
2. Make GCP/Azure implementation less different.

This PR does the following
1. Drop `dateAccessed` field from `nodepool` table
2. Add `namespace` column in `APP` table. In Leonardo code, drop `namespace` table related DB DAO. The actual table will still exist after this PR, which I'll plan to drop in a follow up PR

One caveat is this means there's no DB constraint that's making sure for every cluster, namespaces are unique. But I think reality, the chance of namespace naming collision for the same cluster is quite low.

TODO:
- [x] test things out in BEE (specifically PD for apps). 1. create a galaxy app 2. upload a file 3. delete galaxy app 4. create a new galaxy app 5. verify that the file still exists in the app
- [ ] Ran migration against a cloned DB locally

### Who tested and where

- [ ] I validated this change 


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesServiceDbQueries.scala
```diff
@@ -313,29 +313,25 @@ object KubernetesServiceDbQueries {
...
-        persistentDiskQuery.tableQuery on (_._1._1._1._1._1._1.diskId === _.id) joinLeft
+        serviceQuery on (_._1._1._1.id === _.appId) joinLeft
+        persistentDiskQuery.tableQuery on (_._1._1._1._1.diskId === _.id) joinLeft
         labelQuery on { case (c, lbl) =>
           lbl.resourceId
-            .mapTo[DiskId] === c._1._1._1._1._1._1._1.diskId && lbl.resourceType === LabelResourceType.persistentDisk
+            .mapTo[DiskId] === c._1._1._1._1._1.diskId && lbl.resourceType === LabelResourceType.persistentDisk
         } joinLeft
-        appErrorQuery on (_._1._1._1._1._1._1._1._1.id === _.appId)
-    } yield (cluster, nodepool, app, label, clusterNamespaces, appNamespace, services, disks, diskLabels, appError)
```
mlilynolting, line 335: This used to distinguish between `clusterNamespaces` and `appNamespace` - what has changed, now that only APP has namespaceName? Are there any consequences?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/NodepoolComponent.scala
```diff
@@ -32,7 +32,6 @@ case class NodepoolRecord(
   creator: WorkbenchEmail,
   createdDate: Instant,
   destroyedDate: Instant,
-  dateAccessed: Instant,
```
mlilynolting, line 35: What do we gain by deleting this? It's one of the core AuditInfo fields. I understand we aren't setting the value now, but does that mean we will never need it?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesServiceDbQueries.scala
Qi77Qi: In the original implementation, `clusterNamespaces` is all the namespaces for a given cluster, and `appNamespace` is the namespace a given app is in. I removed `namespaces` from `KubernetesCluster`, so we don't need to look up and aggregate all namespaces for a given cluster anymore because I don't think we need it

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/NodepoolComponent.scala
Qi77Qi: if we're not using it (even worse, having bad data for it), I'd prefer removing it to keep things simple. `nodepool` isn't a user facing object, so the likehood of us needing this is very low.

Qi77Qi:
jenkins retest

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/NodepoolComponent.scala
mlilynolting: How would the effort/usefulness compare if we added a line of code to initialize dateAccessed to dateCreated? My concern is that while the user can't see it, Leonardo can, and directly manages it; date of change can be important for traceability.

mlilynolting:
What's the validation plan for this change and migration?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/NodepoolComponent.scala
Qi77Qi: If we want to properly use it, then we'd need update this whenever a user accesses an app on a nodepool, and it'll be different for GCP and Azure.

`NODEPOOL` table isn't useful for Azure apps at all, so I think future plan for this table is TBD as well.

Even if we can update it easily, what does that give us? in cases like runtimes, apps, we have it because we use it for autopause, but for nodepool, there's no meaningful use of it

Qi77Qi:
@mlilynolting  I don't see much risk of this change (all the namespace data still exists after this PR) other than making sure namespaces are populated correct after the DB migration (validated), and a tangential risk around PD re-using, which is also validated. Is there anything in particular you think should be validated?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/NodepoolComponent.scala
mlilynolting: That's fair, and sounds like the removal will not have an impact.

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20231002_drop_columns_in_nodepool_table.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="leonardo" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+    <changeSet logicalFilePath="leonardo" author="qi" id="qi_drop_columns_in_nodepool_table">
+        <dropColumn tableName="NODEPOOL" columnName="dateAccessed"/>
+        <addColumn tableName="APP">
+            <column name="namespace" type="VARCHAR(254)">
+                <constraints nullable="true"/>
```
mlilynolting, line 7: Is namespace nullable now? It seems like the old `namespaceId` was not nullable.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesServiceDbQueries.scala
mlilynolting: ok - does the new code respect the weird uniqueness requirements? "the Galaxy installation requires that release names be unique within a cluster", from LeoAppServiceInterp?

mlilynolting: Ok, got it from the description; it can't make that guarantee.

If a collision does occur, is there a log or failure case that would alert us to the issue? What would happen?

mlilynolting:
> @mlilynolting I don't see much risk of this change (all the namespace data still exists after this PR) other than making sure namespaces are populated correct after the DB migration (validated), and a tangential risk around PD re-using, which is also validated. Is there anything in particular you think should be validated?

First off, thanks for the explanation of the need for this - deleting nonrunning apps would be a big win ^.^

For validation - regresson testing? 

Since I don't know the app model as thoroughly as you do, it's hard for me to tell at a glance that this will preserve all existing behavior. Since this is a schema change affecting apps, I would want to know that GCP and Azure apps' lifecycles still worked as a basic regression, and that existing apps handle the change without trouble. It's only one column, but it's exposed to a decent number of places, so a basic check of all the functions of apps feels appropriate.

mlilynolting:
I guess Azure apps don't use any of this data model, so testing them is not required, right?

mlilynolting:
And Galaxy is the only Google side app? You've done a create-and-delete; it seems like the only other thing is verifying a smooth migration by stopping/starting/deleting/recreating an existing Galaxy app that's had its namespace name migrated to the new schema

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesServiceDbQueries.scala
LizBaldo: Do we have to drop `namespaces` from `KubernetesCluster`? We could keep it in just to maintain the uniqueness constraint

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/NodepoolComponent.scala
LizBaldo: Would this be needed to track app stop/start once we fix it?

Qi77Qi: nope...`dateAccessed` in `APP` table is what we need

Qi77Qi:
@mlilynolting we have automation tests validating most app behaviors https://github.com/DataBiosphere/leonardo/blob/develop/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/apps/AppLifecycleSpec.scala

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20231002_drop_columns_in_nodepool_table.xml
rtitle: Maybe could make `APP.namespace` not-null after the SQL `UPDATE` statement?

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20231002_drop_columns_in_nodepool_table.xml
```diff
@@ -0,0 +1,21 @@
...
+            </column>
+        </addColumn>
+
+        <dropForeignKeyConstraint baseTableName="APP" constraintName="FK_APP_NAMESPACE_ID" />
+        <dropNotNullConstraint tableName="APP" columnName="namespaceId" columnDataType="BIGINT" />
+    </changeSet>
+
+    <changeSet logicalFilePath="leonardo" id="qi_update_app_namespace" author="qi">
+        <sql>
+            UPDATE APP LEFT JOIN NAMESPACE ns ON ns.id = APP.namespaceId
```
rtitle, line 17: Could this be a `JOIN` instead of `LEFT JOIN`? I think all apps will have a namespaceId foreign key.

Qi77Qi: Yea, I think so..I used `LEFT JOIN` just to more forgiving...any downside using `LEFT JOIN`?

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20231002_drop_columns_in_nodepool_table.xml
Qi77Qi: it's not nullable..but I had things more relaxed in this PR...will make it non nullable in the next PR

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesServiceDbQueries.scala
rtitle: Do we use `KubernetesCluster.namespaces`? If not I'd vote to remove it (as this PR does). 

In general I think it's better to model uniqueness at the DB level, instead of through code. What if we just added a DB unique constraint on `APP.release` and `APP.namespace`?

LizBaldo: That works for me, my understanding was that the uniqueness needed to be enforced by cluster, not across the whole app table, but this would fit the bill for sure

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/NodepoolComponent.scala
rtitle: Eventually I hope we can drop the `NODEPOOL` table entirely ;-)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesServiceDbQueries.scala
rtitle: Yeah, technically uniqueness just needs to be per cluster. But that's hard to model in our current DB schema. We use UUIDs anyway so it should be fine (and collisions will fail fast if enforced at the DB level).

Thinking ahead, if we drop `KUBERNETES_CLUSTER` and `NODEPOOL` tables in the future, and move `cloudContext` to the `APP` table, we could make the unique constraint on `(namespace, cloudContext)`.

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20231002_drop_columns_in_nodepool_table.xml
rtitle: Yeah LEFT JOIN is more forgiving.. no downside I guess (it could just leave `null` records if there are any problems).

It's OK with me if there will be a followup PR to add a not-null constraint, etc.

rtitle (approved):
I like the simplification to these tables! Looks good to me. 👍

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesServiceDbQueries.scala
Qi77Qi: >In general I think it's better to model uniqueness at the DB level, instead of through code. What if we just added a DB unique constraint on APP.release and APP.namespace?

So I checked DB and there are apps use the same `release` and `namespace`...I think these are galaxy apps. When we reuse disk for galaxy app, we seem to use the same `release` and `namespace`. We can maybe add `namespace`, `destroyedDate` constraint in the next PR.

I think we can leave these constraints out of this PR for now just so the migration is a bit safer. And once this PR is on dev/prod, we can add appropriate constrains in the follow up PR (we still need to drop `namespace`table once things look good anyway).

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20231002_drop_columns_in_nodepool_table.xml
mlilynolting: oh that makes sense as an interim step while migrating the field

mlilynolting:
> @mlilynolting we have automation tests validating most app behaviors [`develop`/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/apps/AppLifecycleSpec.scala](https://github.com/DataBiosphere/leonardo/blob/develop/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/apps/AppLifecycleSpec.scala?rgh-link-date=2023-10-23T13%3A16%3A13Z)

Passing automation tests are great! Can automation tests validate a migration, though? I'd feel more comfortable knowing that you had made a quick check that a preexisting Galaxy app on dev or elsewhere looked whole and healthy, and handled CRUD operations, after the change.

Qi77Qi:
@mlilynolting , 

>Can automation tests validate a migration, though

no...but the only real change really is just that where we storing "namespace` info in DB level...

mlilynolting:
yeah it's a small change so a small validation! I just am a little uncomfortable with releasing a migration of any scale without at least quickly checking that the resources migrated look healthy.

Qi77Qi:
@mlilynolting so what exactly r u suggesting? Run this branch locally against dev DB? We shouldn't do that becuz running migration off of an unmerged PR is a bad idea

mlilynolting:
Could you build some Galaxy apps on a BEE on develop branch, and then build your branch and check they migrated?

mlilynolting:
I assume that the older legacy cases will only show up in higher environments, but the simple check would be enough for me.

mlilynolting (approved):
checked in about this offline

Qi77Qi:
chatted with @mlilynolting offline...created 2 galaxy apps in dev...will verify they still function once PR gets to dev

Qi77Qi:
Double checked on dev
1. `select * from APP INNER JOIN NAMESPACE ON APP.namespaceId = NAMESPACE.id where APP.namespace != NAMESPACE.namespaceName;` returns nothing as expected
2. Ran 
```select count(*), `release`, namespace, destroyedDate from APP group by `release`, namespace, destroyedDate;```, and there're some more than 1 count, so I won't add unique constraint for this in follow up PR
4. Galaxy app works fine as expected

PR #3851, opened 2023-10-17
AJ-1074 - Have WDS manage certain config variables on its own vs getting them from Leo

yuliadub:
__Jira ticket__: https://broadworkbench.atlassian.net/browse/AJ-1074

### What

- have WDS decide which configuration values to use based on the environment Leo is running in. This requires changes on the terra-helm side as well, PR linked here: https://github.com/broadinstitute/terra-helmfile/pull/4699/files
- Spoke with devops about the best way to propogate env variable from Leo helm to Leo itself. The corresponding leo helm changes live here: https://github.com/broadinstitute/terra-helmfile/pull/4699

### Why

- This greatly simplifies ability to make configuration changes in WDS without having to create multiple PRs to do so the way it happens today. 

### What to test

- Ensure WDS starts up with no issues inside a BEE


http/src/main/resources/leo.conf
```diff
@@ -6,10 +6,6 @@ application {
   leoUrlBase = ${?LEO_URL_BASE}
 }
 
-app-service {
```
yuliadub: hmmm looks like a merge fail possibly, will investigate (i.e. I did not remove this)

calypsomatic (approved):

Dockerfile
```diff
@@ -63,6 +63,8 @@ RUN helm repo add ingress-nginx https://kubernetes.github.io/ingress-nginx && \
     helm repo add aou-sas-chart https://storage.googleapis.com/terra-app-helm/aou-sas-chart && \
     helm repo update
 
+COPY ./wds-4.46.0.tgz /leonardo
```
jladieu: Seems maybe unintentional to have another hardcoded version string in the `Dockerfile`...just checking if this is what you want, or is this just a shortcut leftover from testing something?

http/src/main/resources/reference.conf
```diff
@@ -346,9 +346,11 @@ azure {
   }
 
    wds-app-config {
+     environment = "dev"
+     environment-base = "live"
      instrumentation-enabled = false
      chart-name = "/leonardo/wds"
-     chart-version = "0.46.0"
+     chart-version = "4.46.0"
```
jladieu: Why the big bump from 0 -> 4?

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/app/WdsAppInstallSpec.scala
```diff
@@ -23,12 +23,14 @@ class WdsAppInstallSpec extends BaseAppInstallSpec {
     mockAzureApplicationInsightsService
   )
 
-  it should "build wds override values" in {
+  it should "build wds override values for env" in {
```
jladieu: Is there any reasonable scenario for the `environmentBase="bee"` that would be worth testing similarly?

Dockerfile
davidangb: agree, this would hardcode Leo to a specific wds version

wds-4.46.0.tgz
davidangb, line 1: when removing the other references to wds-4.46.0, remove this .tgz too

Dockerfile
yuliadub: sorry this is just for testing in my bee, I will go ahead and remove this (forgot this step once I completed testing)

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/app/WdsAppInstallSpec.scala
yuliadub: at this point no - since the values are just strings and nothing is different. there used to be a test specific to a bee when config values were going to be formatted differently based on condition but that is now happening inside of terra-helm

jladieu (approved):

davidangb (approved):

jack-r-warren (approved):

yuliadub:
jenkins retest

yuliadub:
jenkins retest

PR #3866, opened 2023-10-24
[IA-4574] add workspace id to get app

jdcanas:
For some reason we have this in listApp but not getApp which makes for some weird UI logic. Figured the more detailed GET should probably have this.


LizBaldo (approved):
Cool, I did not realize that this was already part of the GetAppResponse object, might as well return it indeed 👍 

One thing to triple check on your BEE though would be running a CoA, WDS apps, as we have been bitten by them using an old version of the leo client in the past

PR #3875, opened 2023-10-25
[WM-2238] Attach cloned managed identity and database to its app

salonishah11:
__Jira ticket__: https://broadworkbench.atlassian.net/browse/WM-2238

### What

While creating an app, for Shared apps that support cloning (currently only WORKFLOWS app), Leo will now check if Azure managed identity and "cloned" database for that app already exists in the workspace, and if found use those instead of creating new resources. If they don't exist, as before it will create new managed identity and database resources.

### Why

To support Workflow cloning

### Testing

I tested these changes in a BEE. Steps I executed:
- Create a new workspace
- Using WSM APIs, manually created
   - An Azure managed identity for WORKFLOWS app
   - A "cloned" CBAS database attached to the above created managed identity
- Launched WORKFLOWS app creation through UI

I was able to verify that Leo found the managed identity and cbas database and attached the app to it instead of creating new ones.

### Who tested and where

- [x] I validated this change 


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -115,33 +116,39 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
         }
       }
 
-      // Create WSM managed identity if shared app
+      wsmResourceApi <- buildWsmResourceApiClient
+
+      // Create or fetch WSM managed identity if shared app
       wsmManagedIdentityOpt <- app.samResourceId.resourceType match {
-        case SamResourceType.SharedApp =>
-          childSpan("createWsmIdentityResource").use { implicit ev =>
```
LizBaldo, line 121: Could you make sure to keep the childSpan to track the managed identity creation?

LizBaldo: Nvm, I spoke too fast and it is now built into the `createAzureManagedIdentity` function, all good :)

LizBaldo (approved):
LGTM, the renaming and comments are very clear, thanks so much! 😍 

The unit tests look comprehensive, but out of an abundance of caution could you also do a little regression test on your BEE and make sure that creating non shared apps will create a db and manage identity every time?

salonishah11:
jenkins retest

salonishah11:
@LizBaldo I tested out few additional scenarios in my BEE and they all work as expected:
- I manually created a "cloned" managed identity for WORKFLOWS app and "cloned" cbas database in a workspace. Leo was able to 
   - create managed identity for other shared app WDS as expected
   - create WDS database as expected
   - upon launching Workflows app from UI, Leo was able to attach the "cloned" managed identity to Workflows app and use "cloned" cbas database 
      - Leo was also able to create the other database "cromwellmetadata" as expected 
   - upon launching submission, Leo was also able to successfully create CROMWELL RUNNER app and its databases - cromwell and tes and attach "cromwellmetadata" reference database to runner app

- I created a new workspace and NO manual managed identity or database was created. Leo was able to
   - create WDS managed identity as expected
   - create WDS database as expected
   - upon launching Workflows app from UI,
      - create WORKFLOWS app managed identity 
      - create cbas and cromwellmetadata databases
  - upon launching a submission, Leo was also able to successfully create CROMWELL RUNNER app and its databases - cromwell and tes and attach "cromwellmetadata" reference database to runner app

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -695,6 +704,19 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
     }
   }
 
+  private[util] def createAzureManagedIdentity(app: App, namespacePrefix: String, workspaceId: WorkspaceId)(implicit
```
dillydally414, line 706: TOL, should these managed identities now be created with `COPY_DEFINITION` cloning instructions?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -115,33 +116,39 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
...
+      wsmResourceApi <- buildWsmResourceApiClient
+
+      // Create or fetch WSM managed identity if shared app
       wsmManagedIdentityOpt <- app.samResourceId.resourceType match {
-        case SamResourceType.SharedApp =>
-          childSpan("createWsmIdentityResource").use { implicit ev =>
-            createWsmIdentityResource(app, namespacePrefix, params.workspaceId).map(_.some)
+        // for apps that support cloning, if a managed identity has already been created in the workspace
+        // use that otherwise create a new managed identity
+        case SamResourceType.SharedApp if doesAppTypeSupportCloning(app.appType) =>
```
rtitle: Just curious, do we need to check `doesAppTypeSupportCloning`? Could we just always try to retrieve an existing identity for the app type, and use it if it exists? Just thinking that might be more flexible, if we need to support other cloneable apps in the future.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -746,21 +768,47 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
...
         ctx <- ev.ask
         wsmApi <- buildWsmControlledResourceApiClient
-        res <- appInstall.databases.collect { case d @ CreateDatabase(_, _) => d }.traverse { database =>
-          createWsmDatabaseResource(app, workspaceId, database, namespacePrefix, owner, wsmApi)
-        }
-      } yield res
+        controlledDbsForApp = appInstall.databases.collect { case d @ ControlledDatabase(_, _) => d }
+        // for apps that support cloning, retrieve databases that are might already be created in workspace
+        existingControlledDbsInWorkspace <-
+          if (doesAppTypeSupportCloning(app.appType))
```
rtitle: Same question about whether checking `doesAppTypeSupportCloning` is needed

rtitle (commented):
~Do you want to use this PR to change:~
```
.cloningInstructions(CloningInstructionsEnum.NOTHING)
```
~when creating the WSM database and identity?~

edit: answered above :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
rtitle: +1 (just commented too) -- I think Leo should set clone instructions when creating identities and databases.

salonishah11: Based on my understanding we wanted to do this after changes were made to WSM to handle apps cloning. Is that not right? I had understood that if we create apps and identities with `COPY_DEFINITION` and WSM doesn't have cloning implemented things might break and so the plan was to make this change after WSM work.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
salonishah11: I had thought about this and added `doesAppTypeSupportCloning` check to avoid unnecessary calls to WSM for other apps that don't support cloning. But I am happy to remove that change if you think it is overkill.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
rtitle: Ohh that makes total sense. Thanks 👍

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
rtitle: I think I would value less complexity/maintenance over optimization in this case. These are all async operations, and we are already making a ton of requests to WSM.

rtitle (approved):
Looks good -- just had the 1 comment about `doesAppTypeSupportCloning`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
salonishah11: Sounds good. I will remove that check.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
dillydally414: I guess this depends on what order things merge in (if we merge in an incomplete version of database cloning). But otherwise it shouldn't cause an issue since we only are trying to clone resources specifically enumerated in WSM, and we can just wait to include those resources in WSM until they are fully ready to be cloned (with the last piece of database cloning). The alternative is that we have another small PR later in Leo to send the new clone instructions, and all apps created between now and then would not be cloneable... open to further discussion tomorrow!

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -115,33 +115,37 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
...
-      // Create WSM managed identity if shared app
+      wsmResourceApi <- buildWsmResourceApiClient
+
+      // Create or fetch WSM managed identity if shared app
       wsmManagedIdentityOpt <- app.samResourceId.resourceType match {
         case SamResourceType.SharedApp =>
-          childSpan("createWsmIdentityResource").use { implicit ev =>
-            createWsmIdentityResource(app, namespacePrefix, params.workspaceId).map(_.some)
+          // if a managed identity has already been created in the workspace use that otherwise create a new managed identity
+          retrieveWsmManagedIdentity(wsmResourceApi, app.appType, params.workspaceId.value).flatMap {
```
dillydally414, line 124: TOL, non-blocking: would it be better to pass in the expected identity name to both `retrieve` and `create`, so we are not duplicating the `id<appname>` name schema and introducing a chance they get out of sync?

dillydally414 (approved):

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
salonishah11: Yeah good point. I will extract that into a function.

PR #3879, opened 2023-10-26
RW-11168 Drop namespace table

Qi77Qi:
https://precisionmedicineinitiative.atlassian.net/browse/RW-11168

Validated on prod that [this PR](https://github.com/DataBiosphere/leonardo/pull/3849) had the expected result. So dropping the duplicate column/table for real

I attempted to allow more statues for deleteApp request. But for some reason, delete namespace will get stuck if the app wasn't in `RUNNING` status. My theory is because the way we stopApp currently is by resizing the nodepool to 0, which leaves some pods in bad state, and blocking namespace deletion.

Since we'll be working to switch how to stop/start apps (not resizing nodepools), I think it makes sense to drop this for now. So now this PR mostly only does some minor DB migration to clean up deprecated fields/tables.

### Who tested and where

- [ ] I validated this change 


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -366,12 +366,17 @@ class GKEInterpreter[F[_]](
...
-      googleCluster <- F.fromOption(
-        googleClusterOpt,
-        ClusterCreationException(ctx.traceId,
-                                 s"Cluster not found in Google: ${gkeClusterId} | trace id: ${ctx.traceId}"
-        )
-      )
+      googleCluster <- googleClusterOpt match {
+        case Some(value) => F.pure(value)
+        case None =>
+          kubernetesClusterQuery.markAsDeleted(dbCluster.id, ctx.now).transaction >>
```
Qi77Qi, line 372: this fixes an issue where if a GKE cluster is deleted in google, all Leo does is prompting user that the cluster doesn't exist.

With this change, we're marking the cluster as `DELETED` when leo encounters the issue, and user can try to create a new app again, and leo will just create new app.

This change makes it much easier to work with GKE apps for development, and also benefits real users.

Qi77Qi:
```
13:25:22 17:25 ERR error="2 execution errors:\nconsent: timed out waiting for healthy (Degraded: consent-deployment: 0/80 nodes are available: 26 Insufficient memory, 80 Insufficient cpu. preemption: 0/80 nodes are available: 80 No preemption victims found for incoming pod.)\nontology: timed out waiting for healthy (Degraded: ontology-deployment: pod didn't trigger scale-up: 1 max node group size reached)\n"
```

jenkins retest

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
```diff
@@ -373,13 +376,12 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
...
       _ <- if (hasDeletePermission) F.unit else F.raiseError[Unit](ForbiddenError(userInfo.userEmail))
 
-      canDelete = AppStatus.deletableStatuses.contains(appResult.app.status)
-      _ <-
-        if (canDelete) F.unit
-        else
-          F.raiseError[Unit](
-            AppCannotBeDeletedException(cloudContext, appName, appResult.app.status, ctx.traceId)
-          )
+      // Galaxy is the only app that has custom deletion logic that requires `helm uninstall`. For all other apps, we'll
```
LizBaldo, line 379: Nice 👍

LizBaldo: One suggestion though would be to move this comment into the `checkIfCanBeDeleted` function, because the current status is that we are not yet attempting to delete non galaxy apps by deleting the namespace

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
```diff
@@ -1390,7 +1392,25 @@ object LeoAppServiceInterp {
...
     aSplited(0) == bSplited(0) && aSplited(1) == bSplited(1)
   }
+
+  private[http] def checkIfCanBeDeleted(appType: AppType, appStatus: AppStatus): Either[String, Unit] = {
+    val deletable =
+      appType match {
+        case AppType.Galaxy =>
+          AppStatus.deletableStatuses.contains(appStatus)
+        case _ =>
+          // As of 10/26/2023.
```
LizBaldo, line 1402: This is where I would put the great comment that you have above, I feel like it is some very important context that will help future devs to hopefully relax the constraint 🤞

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -966,7 +978,7 @@ class GKEInterpreter[F[_]](
         .namespaceExists(dbApp.cluster.getClusterId, KubernetesNamespace(dbApp.app.appResources.namespace))
         .map(!_) // mapping to inverse because booleanDoneCheckable defines `Done` when it becomes `true`...In this case, the namespace will exists for a while, and eventually becomes non-existent
 
-      _ <- streamUntilDoneOrTimeout(fa, 30, 5 seconds, "delete namespace timed out")
+      _ <- streamUntilDoneOrTimeout(fa, 60, 5 seconds, "delete namespace timed out")
```
LizBaldo, line 981: Why the increase in timeout here?

LizBaldo (approved):
LGTM, marking the missing GKE clusters as deleted in leo db is a nice touch! I only have a couple of minor comments :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
Qi77Qi: I increased this initially becuz I thought deleting namespace was stuck, and increasing timeout is all we needed. In the end, I realize in some cases, namespaces are really just stuck in deleting, increasing timeout wouldn't help.

But regardless, I think increasing the timeout is nice to do anyway and there's not much downside in this case.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
Qi77Qi: I'll keep this in mind in my future PR...I'll have a few more PRs coming which will likely move code around a bit still :)

PR #3895, opened 2023-10-31
RW-10598  Support updating replica for supporting stop/start (only ALLOWED apps for now)

Qi77Qi:
https://precisionmedicineinitiative.atlassian.net/browse/RW-10598

More changes coming..No need to review yet

See [this](https://docs.google.com/document/d/1T2Sqq9x4YAuhpYZX1budhnU-Hc7J2mKCult69oOsW5o/edit) for more details. [Related PR](https://github.com/broadinstitute/workbench-libs/pull/1592)

This PR changes stop/start to use scaling up/down replicas for `ALLOWED` apps. I think this is a good start as we try out this new approach. For Galaxy to adopt this approach, it'll be tricky because it has a few more deployments instead of just one deployment. I'll try to add CROMWELL in a follow up PR.

Tested on a BEE
1. Stop/start behaves as expected for `ALLOWED` apps. Verified autoscaling is working as expected as well (nodepool gets resized to 0 automatically when there's no PODs running on the nodepool). One rule I'm enforcing here is that if an app is using `replica` for stop/start, then `autoscaling` has to be `true`..In this case, `ALLOWED` app always have `autoscaling` enabled
2. There's no change to all other apps' stop/start behavior.

- [ ] I validated this change 

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/apps/AppLifecycleSpec.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/apps/AppLifecycleSpec.scala
@@ -65,7 +65,7 @@ class AppLifecycleSpec
     googleProject =>
       test(googleProject,
            createAppRequest(AppType.Allowed, "rstudio-test-workspace", None, Some(AllowedChartName.RStudio)),
-           false,
+           true,
            true
       )
   }
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
@@ -1036,7 +1036,7 @@ class GKEInterpreter[F[_]](
             // Update nodepool status to Running and app status to Stopped
             _ <- attemptToStop match {
               case Left(e) =>
-                // This updates the APP back to `RUNNNING` status instead of putting it into `ERROR` status. This
+                // This updates the APP back to `RUNNING` status instead of putting it into `ERROR` status. This
                 // can be confusing to users since they will notice the APP is not stoppable.
                 // Currently, Leo doesn't have a good way to inform users about "an error happened during Stopping",
                 // which I think we should spend some effort design this out.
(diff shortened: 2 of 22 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -1014,101 +1014,47 @@ class GKEInterpreter[F[_]](
...
-                s"setNodepoolSize operation has failed or timed out for nodepool ${dbNodepool.id}"
-              ) >>
-                F.raiseError[Unit](NodepoolStopException(dbNodepool.id))
-        } yield ()
+              .attempt
+
+            // Update nodepool status to Running and app status to Stopped
+            _ <- attemptToStop match {
+              case Left(e) =>
+                // This updates the APP back to `RUNNNING` status instead of putting it into `ERROR` status. This
```
LizBaldo: nit, `RUNNING`

LizBaldo: But yeah, i agree this makes a lot more sense from the user's perspective 👍

LizBaldo (commented):
I really like the refactoring, and you made it super clear / easy for the rest of the app types to follow when we get to that.

One thing though, would it make sense to have an automation test that would spin both Allowed app and successfully stop / start them? I know that AOU wants to move away from E2E tests, but making sure that we don't break this seems pretty important to me

Qi77Qi:
@LizBaldo lmk if u have more comments...an integration for start/stop rstudio app is added (can't really test SAS since it's only available for AoU projects)

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/apps/AppLifecycleSpec.scala
```diff
@@ -65,7 +65,7 @@ class AppLifecycleSpec
     googleProject =>
       test(googleProject,
            createAppRequest(AppType.Allowed, "rstudio-test-workspace", None, Some(AllowedChartName.RStudio)),
-           false,
+           true,
```
LizBaldo, line 68: 👍

LizBaldo (approved):
:shipit: Thanks for switching on the stop/start for Rstudio, it would be create to eventually expand to creating two ALLOWED apps so you can test the stop/start with multiple apps, since that was the original bug :)

Qi77Qi:
@LizBaldo I think the test might already validate that because we're using the same test user, which means all apps will be sharing the same nodepool (before this PR, Galaxy is the only one had the flag enabled)

Btw, I notice galaxy test is disabled

LizBaldo:
Makes sense, yeah I think nathan disabled that test a while back because it was super flaky, we could reenable it

PR #3906, opened 2023-11-02
RW-10711 fix sas proxy bug

Qi77Qi:
This value is passed to the SAS chart to define `sas.commons.web.security.csrf.allowedUris` so that SAS won't reject requests proxied from Leonardo.

In all other envs, Leo's referer is the same as proxy url's domain, but that's not the case for prod leonardo.

So change the variable to use leo's proxy url's domain, which will fix the bug

### Who tested and where

- [ ] I validated this change 

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/BuildHelmChartValues.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/BuildHelmChartValues.scala
@@ -333,7 +333,7 @@ private[leonardo] object BuildHelmChartValues {
           raw"""ingress.path.sas=${ingressPath}${"(/|$)(.*)"}""",
           raw"""ingress.path.welder=${welderIngressPath}${"(/|$)(.*)"}""",
           raw"""ingress.proxyPath=${ingressPath}""",
-          raw"""ingress.referer=${config.leoUrlBase}""",
+          raw"""ingress.referer=${config.proxyConfig.proxyUrlBase.dropRight(6)}""",
           raw"""ingress.annotations.nginx\.ingress\.kubernetes\.io/proxy-redirect-from=http://${k8sProxyHost
               .address()}""",
           raw"""imageCredentials.username=${config.allowedAppConfig.sasContainerRegistryCredentials.username.asString}""",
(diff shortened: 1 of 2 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/BuildHelmChartValues.scala
```diff
@@ -333,7 +333,7 @@ private[leonardo] object BuildHelmChartValues {
           raw"""ingress.path.sas=${ingressPath}${"(/|$)(.*)"}""",
           raw"""ingress.path.welder=${welderIngressPath}${"(/|$)(.*)"}""",
           raw"""ingress.proxyPath=${ingressPath}""",
-          raw"""ingress.referer=${config.leoUrlBase}""",
+          raw"""ingress.referer=${config.proxyConfig.proxyUrlBase.dropRight(6)}""",
```
LizBaldo: Would the order of that config list ever change? If so is there a more robust way to grab the element you need?

LizBaldo (approved):
I have a question about the use of dropright here, but should not be a blocker 👍

yonghaoy (approved):

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/BuildHelmChartValues.scala
Qi77Qi: what `dropRight(6)` does is just dropping the right most 6 digits of the String. So instead of `https://https://notebooks.firecloud.org/proxy/`, it'll become `https://notebooks.firecloud.org/`...

Things will break if we ever change how we do proxy path, which I think isn't very likely. maybe I'll add a comment in the config

LizBaldo: Yeah, a comment is perfectly fine I think :) Thanks!

mlilynolting: It seems like what you need is the URI's Host value, without the `/proxy` part. I'm concerned that the `dropRight(6)` is obscuring the more specific ask of "remove trailing `/proxy` from the base URL."

Would it work to use the existing `ProxyConfig::getProxyServerHostName` method?

Qi77Qi: oh cool..didn't realize the helper exists..

mlilynolting (approved):

PR #3915, opened 2023-11-06
RW-11279  Refactoring autopauseMonitor in preparation of adding more such monitors

Qi77Qi:
There's no functional change to `AutopauseMonitor` in this PR except a minor change to one log msg

- [ ] I validated this change 

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/AutopauseMonitor.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/AutopauseMonitor.scala
@@ -81,7 +81,7 @@ class AutopauseMonitor[F[_]](
     }
   } yield filtered
 
-  override def doAction(a: RuntimeToAutoPause, traceId: TraceId, now: Instant)(implicit F: Async[F]): F[Unit] =
+  override def action(a: RuntimeToAutoPause, traceId: TraceId, now: Instant)(implicit F: Async[F]): F[Unit] =
     for {
       _ <- clusterQuery.updateClusterStatus(a.id, RuntimeStatus.PreStopping, now).transaction
       _ <- publisherQueue.offer(LeoPubsubMessage.StopRuntimeMessage(a.id, Some(traceId)))
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/AutopauseMonitor.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/AutopauseMonitor.scala
@@ -27,11 +27,11 @@ class AutopauseMonitor[F[_]](
   dbRef: DbReference[F],
   ec: ExecutionContext
 ) extends BackgroundProcess[F, RuntimeToAutoPause] {
-  override def monitorType: String =
+  override def name: String =
     "autopause" // autopauseRuntime is more accurate. But keep the current name so that existing metrics won't break
   override def interval: scala.concurrent.duration.FiniteDuration = config.autoFreezeCheckInterval
 
-  override def filterCriteria(cluster: RuntimeToAutoPause, now: Instant)(implicit
+  def filterCriteria(cluster: RuntimeToAutoPause, now: Instant)(implicit
     F: Async[F],
     metrics: OpenTelemetryMetrics[F],
     logger: StructuredLogger[F]
@@ -70,8 +70,16 @@ class AutopauseMonitor[F[_]](
         } else F.pure(isIdle)
     }
 
-  override def dbSource(): F[Seq[RuntimeToAutoPause]] =
-    clusterQuery.getClustersReadyToAutoFreeze.transaction
+  override def getCandidates(now: Instant)(implicit
+    F: Async[F],
+    metrics: OpenTelemetryMetrics[F],
+    logger: StructuredLogger[F]
+  ): F[Seq[RuntimeToAutoPause]] = for {
+    candidates <- clusterQuery.getClustersReadyToAutoFreeze.transaction
+    filtered <- candidates.toList.filterA { a =>
+      filterCriteria(a, now)
+    }
+  } yield filtered
 
   override def action(a: RuntimeToAutoPause, traceId: TraceId, now: Instant)(implicit F: Async[F]): F[Unit] =
     for {
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/AutopauseMonitor.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/AutopauseMonitor.scala
@@ -1,11 +1,10 @@
 package org.broadinstitute.dsde.workbench.leonardo
 package monitor
 
-import java.util.UUID
+import fs2.Stream
 import cats.effect.Async
 import cats.effect.std.Queue
 import cats.syntax.all._
-import fs2.Stream
 import org.typelevel.log4cats.StructuredLogger
 import org.broadinstitute.dsde.workbench.leonardo.config.AutoFreezeConfig
 import org.broadinstitute.dsde.workbench.leonardo.dao.JupyterDAO
@@ -13,7 +12,7 @@ import org.broadinstitute.dsde.workbench.leonardo.db._
 import org.broadinstitute.dsde.workbench.leonardo.http.dbioToIO
 import org.broadinstitute.dsde.workbench.model.TraceId
 import org.broadinstitute.dsde.workbench.openTelemetry.OpenTelemetryMetrics
-
+import cats.Show
 import java.time.Instant
 import scala.concurrent.ExecutionContext
 
@@ -25,84 +24,87 @@ class AutopauseMonitor[F[_]](
   jupyterDAO: JupyterDAO[F],
   publisherQueue: Queue[F, LeoPubsubMessage]
 )(implicit
-  F: Async[F],
-  metrics: OpenTelemetryMetrics[F],
-  logger: StructuredLogger[F],
   dbRef: DbReference[F],
   ec: ExecutionContext
-) {
+) extends BackgroundProcess[F, RuntimeToAutoPause] {
+  override def monitorType: String =
+    "autopause" // autopauseRuntime is more accurate. But keep the current name so that existing metrics won't break
+  override def interval: scala.concurrent.duration.FiniteDuration = config.autoFreezeCheckInterval
 
-  val process: Stream[F, Unit] =
-    (Stream.sleep[F](config.autoFreezeCheckInterval) ++ Stream.eval(
-      autoPauseCheck
-        .handleErrorWith(e => logger.error(e)("Unexpected error occurred during auto-pause monitoring"))
-    )).repeat
+  override def filterCriteria(cluster: RuntimeToAutoPause, now: Instant)(implicit
+    F: Async[F],
+    metrics: OpenTelemetryMetrics[F],
+    logger: StructuredLogger[F]
+  ): F[Boolean] =
+    jupyterDAO.isAllKernelsIdle(cluster.cloudContext, cluster.runtimeName).attempt.flatMap {
+      case Left(t) =>
+        logger.error(s"Fail to get kernel status for ${cluster.projectNameString} due to $t").as(true)
+      case Right(isIdle) =>
+        if (!isIdle) {
+          val maxKernelActiveTimeExceeded = cluster.kernelFoundBusyDate match {
+            case Some(attemptedDate) =>
+              val maxBusyLimitReached =
+                now.toEpochMilli - attemptedDate.toEpochMilli > config.maxKernelBusyLimit.toMillis
+              F.pure(maxBusyLimitReached)
+            case None =>
+              clusterQuery
+                .updateKernelFoundBusyDate(cluster.id, now, now)
+                .transaction
+                .as(false) // max kernel active time has not been exceeded
+          }
 
-  private[monitor] val autoPauseCheck: F[Unit] =
+          maxKernelActiveTimeExceeded.ifM(
+            metrics.incrementCounter("autoPause/maxKernelActiveTimeExceeded") >>
+              logger
+                .info(
+                  s"Auto pausing ${cluster.cloudContext}/${cluster.runtimeName} due to exceeded max kernel active time"
+                )
+                .as(true),
+            metrics.incrementCounter("autoPause/activeKernelClusters") >>
+              logger
+                .info(
+                  s"Not going to auto pause cluster ${cluster.cloudContext}/${cluster.runtimeName} due to active kernels"
+                )
+                .as(false)
+          )
+        } else F.pure(isIdle)
+    }
+
+  override def dbSource(): F[Seq[RuntimeToAutoPause]] =
+    clusterQuery.getClustersReadyToAutoFreeze.transaction
+
+  override def action(a: RuntimeToAutoPause, traceId: TraceId, now: Instant)(implicit F: Async[F]): F[Unit] =
     for {
-      _ <- logger.info(
-        s"[logAlert/autopauseHeartbeat] Executing autopause check at interval of ${config.autoFreezeCheckInterval}"
-      )
-      clusters <- clusterQuery.getClustersReadyToAutoFreeze.transaction
-      now <- F.realTimeInstant
-      pauseableClusters <- clusters.toList.filterA { cluster =>
-        jupyterDAO.isAllKernelsIdle(cluster.cloudContext, cluster.runtimeName).attempt.flatMap {
-          case Left(t) =>
-            logger.error(s"Fail to get kernel status for ${cluster.projectNameString} due to $t").as(true)
-          case Right(isIdle) =>
-            if (!isIdle) {
-              val maxKernelActiveTimeExceeded = cluster.kernelFoundBusyDate match {
-                case Some(attemptedDate) =>
-                  val maxBusyLimitReached =
-                    now.toEpochMilli - attemptedDate.toEpochMilli > config.maxKernelBusyLimit.toMillis
-                  F.pure(maxBusyLimitReached)
-                case None =>
-                  clusterQuery
-                    .updateKernelFoundBusyDate(cluster.id, now, now)
-                    .transaction
-                    .as(false) // max kernel active time has not been exceeded
-              }
-
-              maxKernelActiveTimeExceeded.ifM(
-                metrics.incrementCounter("autoPause/maxKernelActiveTimeExceeded") >>
-                  logger
-                    .info(
-                      s"Auto pausing ${cluster.cloudContext}/${cluster.runtimeName} due to exceeded max kernel active time"
-                    )
-                    .as(true),
-                metrics.incrementCounter("autoPause/activeKernelClusters") >>
-                  logger
-                    .info(
-                      s"Not going to auto pause cluster ${cluster.cloudContext}/${cluster.runtimeName} due to active kernels"
-                    )
-                    .as(false)
-              )
-            } else F.pure(isIdle)
-        }
-      }
-      _ <- metrics.gauge("autoPause/numOfRuntimes", pauseableClusters.length)
-      _ <- pauseableClusters.traverse_ { cl =>
-        val traceId = TraceId(s"fromAutopause_${UUID.randomUUID().toString}")
-        for {
-          _ <- clusterQuery.updateClusterStatus(cl.id, RuntimeStatus.PreStopping, now).transaction
-          _ <- metrics.incrementCounter("autoPause/pauseRuntimeCounter")
-          _ <- logger.info(Map("traceId" -> traceId.asString))(s"Auto freezing runtime ${cl.projectNameString}")
-          _ <- publisherQueue.offer(LeoPubsubMessage.StopRuntimeMessage(cl.id, Some(traceId)))
-        } yield ()
-      }
+      _ <- clusterQuery.updateClusterStatus(a.id, RuntimeStatus.PreStopping, now).transaction
+      _ <- publisherQueue.offer(LeoPubsubMessage.StopRuntimeMessage(a.id, Some(traceId)))
     } yield ()
 }
 
 object AutopauseMonitor {
-  def apply[F[_]](config: AutoFreezeConfig, jupyterDAO: JupyterDAO[F], publisherQueue: Queue[F, LeoPubsubMessage])(
+  def process[F[_]](config: AutoFreezeConfig, jupyterDAO: JupyterDAO[F], publisherQueue: Queue[F, LeoPubsubMessage])(
     implicit
+    dbRef: DbReference[F],
+    ec: ExecutionContext,
     F: Async[F],
-    metrics: OpenTelemetryMetrics[F],
-    logger: StructuredLogger[F],
+    openTelemetry: OpenTelemetryMetrics[F],
+    logger: StructuredLogger[F]
+  ): Stream[F, Unit] = {
+    val autopauseMonitor = apply(config, jupyterDAO, publisherQueue)
+
+    implicit val runtimeToAutoPauseShowInstance: Show[RuntimeToAutoPause] =
+      Show[RuntimeToAutoPause](runtimeToAutoPause => runtimeToAutoPause.projectNameString)
+    autopauseMonitor.process
+  }
+
+  private def apply[F[_]](config: AutoFreezeConfig,
+                          jupyterDAO: JupyterDAO[F],
+                          publisherQueue: Queue[F, LeoPubsubMessage]
+  )(implicit
     dbRef: DbReference[F],
     ec: ExecutionContext
   ): AutopauseMonitor[F] =
     new AutopauseMonitor(config, jupyterDAO, publisherQueue)
+
 }
 
 final case class RuntimeToAutoPause(id: Long,
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/AutopauseMonitor.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/AutopauseMonitor.scala
@@ -31,7 +31,7 @@ class AutopauseMonitor[F[_]](
     "autopause" // autopauseRuntime is more accurate. But keep the current name so that existing metrics won't break
   override def interval: scala.concurrent.duration.FiniteDuration = config.autoFreezeCheckInterval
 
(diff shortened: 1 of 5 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BackgroundProcess.scala
```diff
@@ -0,0 +1,95 @@
...
+                       filteredA.length
+    ) // TODO: update metrics once we add more monitors
+    _ <- filteredA.traverse_ { a =>
+      for {
+        uuid <- F.delay(UUID.randomUUID())
+        traceId = TraceId(s"${monitorType}_${uuid.toString}")
+        _ <- metrics.incrementCounter(
+          s"${monitorType}/pauseRuntimeCounter"
+        ) // TODO: update metrics once we add more monitors
+        _ <- logger.info(Map("traceId" -> traceId.asString))(s"${monitorType} | runtime ${a.show}")
```
Qi77Qi: this log message will look a bit different from before

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BackgroundProcess.scala
```diff
@@ -0,0 +1,95 @@
...
+    showA: Show[A],
+    F: Async[F],
+    openTelemetry: OpenTelemetryMetrics[F],
+    logger: StructuredLogger[F]
+  ): Stream[F, Unit] =
+    (Stream.sleep[F](interval) ++ Stream.eval(
+      check
+        .handleErrorWith {
+          case e: java.sql.SQLTransientConnectionException =>
+            logger.error(e)("DB connection failed transiently") >> F.raiseError[Unit](e)
```
Qi77Qi, line 90: raiseError here instead of recovering from the error

mlilynolting:
I would really like a breakdown of these changes and what the intention here is.

Qi77Qi:
@mlilynolting happy to do a walkthrough..lmk

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BackgroundProcess.scala
```diff
@@ -0,0 +1,92 @@
...
+  ): F[Unit] = for {
+    now <- F.realTimeInstant
+    _ <- logger.info(
+      s"[logAlert/${monitorType}Heartbeat] Executing ${monitorType} check at interval of ${interval}"
+    )
+    candidates <- dbSource()
+    filteredA <- candidates.toList.filterA { a =>
+      filterCriteria(a, now)
+    }
+    _ <- metrics.gauge(s"${monitorType}/numOfRuntimes",
```
LizBaldo: Is a gauge the right metric to collect here? Would a direct count be easier? I know the current gauge metrics on autopause runtimes on Grafana is not easy to interpret because of that

LizBaldo (commented):
I really like this change, and the formalization of what Background Processes should look like 👍 
I would love for @cindy-broadinstitute to have a look as she is the most involved into the Autopause Monitor issue :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BackgroundProcess.scala
Qi77Qi: I just left everything what it was...didn't want to mess with any of these since I know ppl are actively looking at things still :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BackgroundProcess.scala
```diff
@@ -0,0 +1,92 @@
...
+  def filterCriteria(a: A, now: Instant)(implicit
+    F: Async[F],
+    metrics: OpenTelemetryMetrics[F],
+    logger: StructuredLogger[F]
+  ): F[Boolean]
+
+  /**
+   * Get a collection of candidates from the database
+   */
+  def dbSource(): F[Seq[A]]
```
cpate4: Ultimately this could be a list of candidates from a variety of sources ... does it make sense to name this as such?

e.g. `retrieveCandidates()` ... or something to that effect

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BackgroundProcess.scala
```diff
@@ -0,0 +1,92 @@
...
+import scala.concurrent.duration.FiniteDuration
+
+/**
+ * Defines a common interface that certain background processes should implement.
+ * This type of background processes do the following periodically:
+ * 1. Query the database for a list of candidates
+ * 2. Filter the candidates based on some criteria
+ * 3. Perform some action on the filtered candidates
+ */
+trait BackgroundProcess[F[_], A] {
```
mlilynolting, line 21: For name, maybe LeonardoResourceMonitor? There are a number of other *Monitors around the system, which seem to work as long-running continuous jobs.

mlilynolting, line 21: Can A be typed? Isn't it specifically a DB resource or query of some kind?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BackgroundProcess.scala
```diff
@@ -0,0 +1,92 @@
...
+  def monitorType: String
```
mlilynolting: I'd be tempted to call this `name` or `label` since it's primarily a human-readable non-unique label for logs

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BackgroundProcess.scala
```diff
@@ -0,0 +1,92 @@
...
+  def interval: FiniteDuration
+
+  /**
+   * Given a candidate, determine whether it should be filtered out.
+   * true: keep the candidate
+   * false: filter out the candidate
+   */
+  def filterCriteria(a: A, now: Instant)(implicit
```
mlilynolting: Could we combine `filterCriteria` and `dbSource` into a single method called, maybe, `getCandidates`? The caller could handle the DB stuff; I could see there being a wide variety of monitors that might be based on multiple database lookups or querying of other kinds of resources which are not in the db, and it might be simpler to reason about if there were just three stages (get candidates, send logs and metrics, act on candidates).

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/AutopauseMonitor.scala
```diff
@@ -25,84 +24,87 @@ class AutopauseMonitor[F[_]](
...
     F: Async[F],
-    metrics: OpenTelemetryMetrics[F],
-    logger: StructuredLogger[F],
+    openTelemetry: OpenTelemetryMetrics[F],
+    logger: StructuredLogger[F]
+  ): Stream[F, Unit] = {
+    val autopauseMonitor = apply(config, jupyterDAO, publisherQueue)
+
+    implicit val runtimeToAutoPauseShowInstance: Show[RuntimeToAutoPause] =
+      Show[RuntimeToAutoPause](runtimeToAutoPause => runtimeToAutoPause.projectNameString)
```
mlilynolting, line 103: What is a `Show`?

Qi77Qi: https://typelevel.org/cats/typeclasses/show.html

mlilynolting: Thank you! That makes sense :)

Is this something that could move into the `RuntimeToAutopause` definition? Seems like it's basically a better `toString`; it would make sense to have that all encapsulated in one object

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BackgroundProcess.scala
```diff
@@ -0,0 +1,92 @@
...
+    F: Async[F],
+    metrics: OpenTelemetryMetrics[F],
+    logger: StructuredLogger[F]
+  ): F[Unit] = for {
+    now <- F.realTimeInstant
+    _ <- logger.info(
+      s"[logAlert/${monitorType}Heartbeat] Executing ${monitorType} check at interval of ${interval}"
+    )
+    candidates <- dbSource()
+    filteredA <- candidates.toList.filterA { a =>
```
mlilynolting: I know this is existing code, but this is the only place in Leo that uses `filterA`, and I can't really see why. Is this something that would work just as well with `filter`?

mlilynolting: `filteredA` feels confusing as a variable name, given that the `A` here is the generic type `A` and not the `A` in `filterA` which populates the variable

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BackgroundProcess.scala
```diff
@@ -0,0 +1,92 @@
...
+      filterCriteria(a, now)
+    }
+    _ <- metrics.gauge(s"${monitorType}/numOfRuntimes",
+                       filteredA.length
+    ) // TODO: update metrics once we add more monitors
+    _ <- filteredA.traverse_ { a =>
+      for {
+        uuid <- F.delay(UUID.randomUUID())
```
mlilynolting, line 60: This is new, right? Why the delay?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BackgroundProcess.scala
mlilynolting: Why not abstract the metrics and logging side effects to their own method so that `check()` is generic?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BackgroundProcess.scala
```diff
@@ -0,0 +1,92 @@
...
+  def process(implicit
+    showA: Show[A],
+    F: Async[F],
+    openTelemetry: OpenTelemetryMetrics[F],
+    logger: StructuredLogger[F]
+  ): Stream[F, Unit] =
+    (Stream.sleep[F](interval) ++ Stream.eval(
+      check
+        .handleErrorWith { case e =>
+          logger.error(e)("Unexpected error occurred during auto-pause monitoring. Recovering...")
```
mlilynolting: This could use the `monitorType`, right?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BackgroundProcess.scala
Qi77Qi: I want to make sure metrics are sent in one place and in a consistent manner so that individual "monitors" don't have to worry about this...

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BackgroundProcess.scala
Qi77Qi: sounds good 👍

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BackgroundProcess.scala
Qi77Qi: the entire repo is `leonardo`, I don't think there's much point in adding `leonardo` in front of things...`Resource` is quite overloaded too

Qi77Qi: it's a generic type

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BackgroundProcess.scala
Qi77Qi: I can change it to `name` if u prefer...`label` is quite overloaded which means other things in leo code base

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/AutopauseMonitor.scala
Qi77Qi: that wouldn't work becuz `A` is generic....https://github.com/DataBiosphere/leonardo/pull/3915/files#diff-84015af49ce0da0253bc5974a8fbe0b8dac33f7871008ad4e787386a3fdeb722R74 is where `show` is used

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BackgroundProcess.scala
Qi77Qi: no...`filterA` takes `A => F[Boolean]`, filter takes `A => Boolean`

Qi77Qi: it's a standard way to refer to something generic...`filterA` comes from `cats`, not sth defined by us

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BackgroundProcess.scala
Qi77Qi: becuz `UUID.randomUUID()` is not pure, and anything not pure should be wrapped inside `F` is how Leo's code base works

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BackgroundProcess.scala
Qi77Qi: will do

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BackgroundProcess.scala
Qi77Qi: sure...the thinking here is to have the criteria being its own function that can be tested...but I guess that doesn't have to be in `BackgroundProcess`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BackgroundProcess.scala
mlilynolting: Hence `LeonardoResource`... I don't think there's a problem with being specific here, is there? Compare the `WorkspaceManagerResourceMonitor`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/AutopauseMonitor.scala
```diff
@@ -25,84 +24,95 @@ class AutopauseMonitor[F[_]](
...
+  override def name: String =
+    "autopause" // autopauseRuntime is more accurate. But keep the current name so that existing metrics won't break
+  override def interval: scala.concurrent.duration.FiniteDuration = config.autoFreezeCheckInterval
 
-  val process: Stream[F, Unit] =
-    (Stream.sleep[F](config.autoFreezeCheckInterval) ++ Stream.eval(
-      autoPauseCheck
-        .handleErrorWith(e => logger.error(e)("Unexpected error occurred during auto-pause monitoring"))
-    )).repeat
+  def filterCriteria(cluster: RuntimeToAutoPause, now: Instant)(implicit
```
mlilynolting: I think `isCandidate` or `isAutopausable` would make sense for this method name - it acts on one item and returns a boolean

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BackgroundProcess.scala
mlilynolting: What do Runtimes and Apps have in common? They're both cloud environments - would it be better to call this a `CloudEnvironmentMonitor`?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BackgroundProcess.scala
```diff
@@ -0,0 +1,84 @@
...
+  def getCandidates(now: Instant)(implicit
+    F: Async[F],
+    metrics: OpenTelemetryMetrics[F],
+    logger: StructuredLogger[F]
+  ): F[Seq[A]]
+
+  /**
+   * Perform some action on the candidate
+   */
+  def action(a: A, traceId: TraceId, now: Instant)(implicit F: Async[F]): F[Unit]
```
mlilynolting, line 41: This could be called `processCandidate` or `actOnCandidate` or something; it should be a verb.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/AutopauseMonitor.scala
mlilynolting: `A` should be restricted to some degree though, yes? For example currently the only `A` we use is `RuntimeToAutoPause`; it might make sense to have that extend or implement a contract `MonitorCandidate` which could hold the show logic. Then `A <: MonitorCandidate` would both make the code self-documenting and allow the show logic to live where it belongs, with the item it's showing

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BackgroundProcess.scala
mlilynolting: How do you know whether to use `F.delay` over `F.blocking` here? I am not sure whether this small side effect is "thread-blocking"

Qi77Qi: becuz generating a UUID doesn't block any thread

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/AutopauseMonitor.scala
Qi77Qi: no. if we don't have to restrict anything here, we should leave it generic...

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BackgroundProcess.scala
Qi77Qi: We don't typically call apps as `CloudEnvironments` in leo

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/AutopauseMonitor.scala
mlilynolting: The proliferation of generic types makes our legacy code really hard to read, in combination with implicits. I don't understand why this refactor requires relaxing the type-safety of this method from one type (RuntimeToAutopause) to Any

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BackgroundProcess.scala
mlilynolting: The UI does - what are they in our code, though? That's why I suggested "Resources," because I think that's how Leo refers to runtimes, apps, disks, etc - they're Leo resources. I'd love a better word, if there is one

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BackgroundProcess.scala
Qi77Qi: it's not a verb...this function describes the action that will happen

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BackgroundProcess.scala
Qi77Qi: How about we leave it as is until a better name emerges

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/AutopauseMonitor.scala
Qi77Qi: `A` is different from `Any`..everything will still type safe at compile time

Qi77Qi:
jenkins retest

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BackgroundProcess.scala
```diff
@@ -0,0 +1,84 @@
...
+import org.typelevel.log4cats.StructuredLogger
+
+import java.time.Instant
+import java.util.UUID
+import scala.concurrent.duration.FiniteDuration
+
+/**
+ * Defines a common interface that certain background processes should implement.
+ * This type of background processes do the following periodically:
+ * 1. Query the database for a list of candidates
```
yonghaoy, line 18: optional nit: 
<ol>
  <li>Query the database for a list of candidates
  <li>Filter the candidates based on some criteria
  <li>Perform some action on the filtered candidates
</ol>

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/AutopauseMonitor.scala
```diff
@@ -25,84 +24,95 @@ class AutopauseMonitor[F[_]](
...
-    (Stream.sleep[F](config.autoFreezeCheckInterval) ++ Stream.eval(
-      autoPauseCheck
-        .handleErrorWith(e => logger.error(e)("Unexpected error occurred during auto-pause monitoring"))
-    )).repeat
+  def isAutopauseable(cluster: RuntimeToAutoPause, now: Instant)(implicit
+    F: Async[F],
+    metrics: OpenTelemetryMetrics[F],
+    logger: StructuredLogger[F]
+  ): F[Boolean] =
+    jupyterDAO.isAllKernelsIdle(cluster.cloudContext, cluster.runtimeName).attempt.flatMap {
```
yonghaoy, line 39: Is this class just for Jupyter Runtime? If we want to autopause Rstudio on GCE or on GKE, those would be seperate subclass?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/AutopauseMonitor.scala
```diff
@@ -25,84 +24,95 @@ class AutopauseMonitor[F[_]](
...
+      case Left(t) =>
+        logger.error(s"Fail to get kernel status for ${cluster.projectNameString} due to $t").as(true)
```
yonghaoy, line 41: in this case, shall we just log then skip OR we should use the last busyDate? 
I am thinking about the case when Jupyter is stuck in unhealthy state, is that Kernel "busy" or "free".

yonghaoy (commented):
Thanks!

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/AutopauseMonitor.scala
Qi77Qi: this is just for jupyter runtimes. So for supporting apps, we'll create another one that extends `BackgroundProcess`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/AutopauseMonitor.scala
Qi77Qi: I don't wanna change existing logic in this PR. But the thinking here is if we can't get the status of the kernel, we'll just proceed to pause the runtime

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/AutopauseMonitor.scala
mlilynolting: In that case, would you add comments to the scaladoc describing what A is intended to represent?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BackgroundProcess.scala
mlilynolting: function names are usually verbs; why break that pattern? `check()` is a verb, `getCandidates()` is a verb.

cindy-broadinstitute (approved):
lgtm, and I will do more testing on dev as we discussed.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BackgroundProcess.scala
Qi77Qi: I don't think function names need to always be verbs

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/AutopauseMonitor.scala
yonghaoy: emmm. I thought in Terra, I can autopause RStudio on GCE.

yonghaoy (approved):
LG

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/AutopauseMonitor.scala
Qi77Qi: I'm guessing that's an accident..it works becuz `jupyterDao.isAllKernelsIdle` will fail, hence the code will continue to pause rstudio on GCE (theory, haven't verified)

PR #3930, opened 2023-11-13
RW-11352  Clean up SAM resources when nodepool is deleted

Qi77Qi:
When we delete nodepools in leo, we should also delete all the associated SAM resources. Otherwise, these resources will block users from deleting workspaces

### Who tested and where

- [ ] I validated this change 


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/NonLeoMessageSubscriber.scala
```diff
@@ -213,7 +216,16 @@ class NonLeoMessageSubscriber[F[_]](config: NonLeoMessageSubscriberConfig,
   )(implicit ev: Ask[F, AppContext]): F[Unit] =
     for {
       ctx <- ev.ask
-      task = gkeAlg.deleteAndPollNodepool(DeleteNodepoolParams(msg.nodepoolId, msg.googleProject))
+      task = for {
+        _ <- gkeAlg.deleteAndPollNodepool(DeleteNodepoolParams(msg.nodepoolId, msg.googleProject))
+        apps <- appQuery.getAppsByNodepool(msg.nodepoolId).transaction
+        _ <- apps.traverse { app =>
+          authProvider
+            .notifyResourceDeleted(app.samResourceId, app.creator, msg.googleProject)
```
mlilynolting, line 224: It's okay to do this before the GKE deletion is finished, right?

Qi77Qi: `gkeAlg.deleteAndPollNodepool` waits until the nodepool is deleted in GCP

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/NonLeoMessageSubscriberSpec.scala
mlilynolting, line 1: I think the additional call to the authProvider to handle deletion should be checked for in the `handle DeleteNodepoolMessage` test.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/NonLeoMessageSubscriber.scala
mlilynolting: Oh, we don't have to do an `asyncTasks.offer()` call, the way we do below for the whole task?

mlilynolting: Ok, I think I get it - that async offer moves the whole thing to another thread, because it's moving through the steps sequentially. I wasn't sure if the `_ <-` flatMap would block further processing within the for-comp, or if there were some other Scala stuff happening.

Qi77Qi: for comp always runs things 1. sequential 2. short-circuit in case of failure

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/db/AppComponentSpec.scala
```diff
@@ -109,4 +109,22 @@ class AppComponentSpec extends AnyFlatSpecLike with TestComponent {
     }
   }
 
+  it should "get all apps for a given nodepool" in isolatedDbTest {
+    val savedCluster1 = makeKubeCluster(1).save()
+    val savedNodepool1 = makeNodepool(1, savedCluster1.id).save()
+    val savedNodepool2 = makeNodepool(2, savedCluster1.id).save()
+
+    val appName = AppName("test")
```
LizBaldo: I think you should rename it to something else, the unit test is complaining: `org.broadinstitute.dsde.workbench.leonardo.db.AppExistsException: An app with name test already exists for workspace 7ecd8890-2d17-4015-af2c-4d0c0d630499`

LizBaldo (approved):
LGTM :) Just onene minor change needed for the unit test to pass

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/NonLeoMessageSubscriber.scala
```diff
@@ -213,7 +216,16 @@ class NonLeoMessageSubscriber[F[_]](config: NonLeoMessageSubscriberConfig,
   )(implicit ev: Ask[F, AppContext]): F[Unit] =
     for {
       ctx <- ev.ask
-      task = gkeAlg.deleteAndPollNodepool(DeleteNodepoolParams(msg.nodepoolId, msg.googleProject))
+      task = for {
+        _ <- gkeAlg.deleteAndPollNodepool(DeleteNodepoolParams(msg.nodepoolId, msg.googleProject))
+        apps <- appQuery.getAppsByNodepool(msg.nodepoolId).transaction
```
rtitle: Just curious - don't all apps in a nodepool need to be deleted in order for the cronjob to send a `deleteNodepool` message? Just wondering in what circumstance we're deleting a nodepool with non-deleted apps.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/AppComponent.scala
```diff
@@ -298,6 +298,14 @@ object appQuery extends TableQuery(new AppTable(_)) {
   def markPendingDeletion(id: AppId): DBIO[Int] =
     updateStatus(id, AppStatus.Deleting)
 
+  def getAppsByNodepool(
+    nodepoolId: NodepoolLeoId
+  )(implicit ec: ExecutionContext): DBIO[Vector[GetAppsByNodepoolResult]] =
+    appQuery
+      .filter(_.nodepoolId === nodepoolId)
```
rtitle, line 305: Does this want to filter for non-deleted apps or something? Since presumably apps in `DELETED` status are already deleted in Sam.

rtitle (commented):
Just a couple questions about overall context -- I thought expected behavior was to delete `kubernetes-app` Sam resources at app deletion time?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/NonLeoMessageSubscriber.scala
cindy-broadinstitute: Also curious, I was expecting the code to grab the existing apps, then delete the nodepool, then notify the creator that that list of apps had been deleted. (so line 221 comes before line 220)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/AppComponent.scala
Qi77Qi: good point..will do

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/NonLeoMessageSubscriber.scala
Qi77Qi: this happens a lot in Dev projects becuz the underlying clusters get cleaned up periodically..But you're right if an app is deleted in normal flow, Sam resources are cleaned up properly

Qi77Qi: > Also curious, I was expecting the code to grab the existing apps, then delete the nodepool, then notify the creator that that list of apps had been deleted. (so line 221 comes before line 220)

the code 1. deletes the nodepool in GCP 2. grabs non deleted existing apps on the nodepool that's deleted, and tells Sam that they're deleted (that's what `authProvider.notifyResourceDeleted`, 
 there's no notify creator if that's what you're referring to)...

cindy-broadinstitute (approved):
LGTM

Qi77Qi:
seems some dataproc related performance issue happening in GCP...

jenkins retest

PR #3931, opened 2023-11-13
[IA-4525] Remove unneeded tests and reduce reliance on selenium

jdcanas:
This is ready for review. This PR removes usage of selenium as much as possible, and also aggressively trims test. A few notes:
- **The tests will not pass on the PR, see outstanding items**. I have ran them all locally. 
- To replace selenium, I've added a new method to the SSH utility that shells out to gcloud. I initially tried to use the JSSH library in conjunction with private/public keys and the [OSLogin](https://cloud.google.com/compute/docs/oslogin)  google service, but unfortunately this approach does not work since it requires connectivity between the machine running the tests and the VM, which cannot be achieved on github action nodes. I kept the public/private key logic though, since it was working locally (since my local machine on the VPN is on the broad set of IPs we allow list in VM fire wall rules) 
    - Basically I tried to remove the trait `NotebookTestUtils` from as many specs as possible
- A lot of tests used selenium to verify things better done via the command line. These were scattered randomly about different specs. I grouped these similar tests (check java version, check env vars) into a new `RuntimeSystemSpec`.
- There was a common theme of 'test that leo can start a vm with this docker image' that ended up having a lot of selenium related cruft better tested on the docker image themselves. These tests live in terra-docker already for the most part
- Removed Jupyterlab tests
- A lot of tests tried to do too much and duplicated coverage (for example, the patch spec verifying disk attachment)
- A lot of tests were removed, with some [TODOs] left to indicate to reviewers a potential discussion around that test

There's two outstanding items that I need to add pending buffer service changes:
- need to add `tunnel-through-iap` flag to the ssh command 
    - https://cloud.google.com/iap/docs/using-tcp-forwarding
- need to add the fire wall rule required for IAP to leo VMs (discussed with devops, this is essentially the 'blessed' approach from a security standpoint, since it allow lists an ip range in the network within the test project reserved by google for the purpose of proxying to VMs in the project )
    - https://cloud.google.com/iap/docs/using-tcp-forwarding#preparing_your_project_for_tcp_forwarding


jdcanas:
jenkins retest

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/ApiJsonDecoder.scala
```diff
@@ -13,9 +13,10 @@ object ApiJsonDecoder {
     "blockSize"
   )(DiskConfig.apply)
 
-  implicit val getRuntimeResponseCopyDecoder: Decoder[GetRuntimeResponseCopy] = Decoder.forProduct15(
+  implicit val getRuntimeResponseCopyDecoder: Decoder[GetRuntimeResponseCopy] = Decoder.forProduct16(
```
cpate4, line 16: Curious what this change is for? `forProduct15` -> `forProduct16`

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/runtimes/RuntimeStatusTransitionsSpec.scala
```diff
@@ -47,6 +50,11 @@ class RuntimeStatusTransitionsSpec extends BillingProjectFixtureSpec with Parall
       monitorCreateRuntime(billingProject, runtimeName, runtimeRequest)
       Leonardo.cluster.getRuntime(billingProject, runtimeName).status shouldBe ClusterStatus.Running
 
+      // verify we can hit the proxy when the runtime is running
```
jdcanas, line 53: This was another improvement. Before, all tests that created a runtime hit the jupyter server API to confirm 'when leo says the app is running I can hit the jupyter server via the proxy', essentially validating leo's `isProxyAvailable` liveliness check. This should be done only once, and this seemed like the best place to put it.

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/runtimes/RuntimeSystemSpec.scala
```diff
@@ -0,0 +1,104 @@
...
+                                             s"sudo docker exec -it jupyter-server java --version"
+          )
+        } yield output should include("OpenJDK Runtime Environment")
+      }
+      res.unsafeRunSync()(cats.effect.unsafe.IORuntime.global)
+    }
+
+    // TODO: remove selenium from this, needs re-write
+    // see https://broadworkbench.atlassian.net/browse/IA-4723
+    "should set up the users google identity on the vm" ignore { runtimeFixture =>
```
jdcanas, line 68: This was something the hail spec was actually doing. Its an area that could deserve some discussion. Its a good test in my opinion, it essentially tests `when bucket permissions are as expected in the cloud (namely, when there is a file the vm user's pet has permission to access), I can use gsutil commands in the jupyter docker image to interact with it`. Unfortunately, it didn't work out of the box despite me essentially copying it from the hail spec, so I made a ticket to dig into it more and disabled it for now.

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookHailSpec.scala
```diff
@@ -15,170 +15,16 @@ import scala.concurrent.duration._
...
-              .map(_.output.tail.last)
-              .get
-          sparkJobToSucceedcellResult.toInt shouldBe 1000
-        }
-      }
-    }
-
-    // Make sure we can import data from GCS into Hail.
-    // See https://broadworkbench.atlassian.net/browse/IA-1558
-    "should import data from GCS" in { clusterFixture =>
```
jdcanas, line 87: this is an interesting one, see comment in `RuntimeSystemSpec` where I moved this

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/rstudio/RStudioSpec.scala
```diff
@@ -134,7 +132,8 @@ class RStudioSpec extends RuntimeFixtureSpec with NotebookTestUtils with RStudio
     }
 
     // Note this test should be last because the test infrastructure does not close the shiny app
-    "should launch an RShiny app" in { runtimeFixture =>
+    // TODO: can this test be removed? how important is an rshiny app?
```
jdcanas, line 135: Not super sure if we need this from a product perspective, and there's a lot of code that exists to support it. It's the only thing keeping this spec's dependency on selenium around...

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/runtimes/RuntimeGceSpec.scala
```diff
@@ -82,56 +78,57 @@ class RuntimeGceSpec
...
-          }
-        })
-        _ <- LeonardoApiClient.deleteRuntime(project, runtimeName)
-      } yield ()
-    }
-
-    res.unsafeRunSync()(cats.effect.unsafe.IORuntime.global)
-  }
+  // IA-4724
+//  "should be able to create a VM with GPU enabled" ignore { project =>
```
jdcanas, line 81: I commented this out to let me remove the `NotebookTestUtils` dependency without compiler issues. There's a follow-up ticket for a rewrite, which we've prioritized in the next sprint.

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoConfig.scala
```diff
@@ -27,17 +27,21 @@ object LeonardoConfig extends CommonConfig {
...
     val pathToQAJson = gcs.getString("qaJsonFile")
   }
 
+  object Azure {
+    val vmUser = azure.getString("leoVmUser")
+    val vmPassword = azure.getString("leoVmPassword")
+    val bastionName = azure.getString("bastionName")
+    val defaultBastionPort = azure.getInt("defaultBastionPort")
+  }
+
```
cpate4, line 45: None of these _ever_ stand the chance of being used for Google (or another cloud provider)?

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoSuite.scala
```diff
@@ -246,26 +246,27 @@ trait NewBillingProjectAndWorkspaceBeforeAndAfterAll extends BillingProjectUtils
...
+      new RuntimeCreationDiskSpec
+//      new RuntimeAutopauseSpec,
+//      new RuntimePatchSpec,
+//      new RuntimeSystemSpec,
+//      new RuntimeStatusTransitionsSpec,
+//      new NotebookGCECustomizationSpec,
+//      new NotebookGCEDataSyncingSpec,
+//      new RuntimeDataprocSpec,
+//      new RuntimeGceSpec,
+//      new AppLifecycleSpec //TODO: test
```
cpate4, line 257: Do we want to just delete these (vs having them be commented out)?

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoSuite.scala
```diff
@@ -246,26 +246,27 @@ trait NewBillingProjectAndWorkspaceBeforeAndAfterAll extends BillingProjectUtils
...
 final class LeonardoTerraDockerSuite
     extends Suites(
-      new NotebookHailSpec,
-      new NotebookPyKernelSpec,
-      new NotebookRKernelSpec,
-      new RStudioSpec
+//      new NotebookHailSpec,
+//      new NotebookPyKernelSpec,
+//      new NotebookRKernelSpec,
+//      new RStudioSpec
```
cpate4: Same comment re: commented out specs - remove them?

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/RuntimeFixtureSpec.scala
```diff
@@ -160,7 +162,7 @@ object RuntimeFixtureSpec {
         RuntimeConfigRequest.GceConfig(
           machineType = Some(MachineTypeName("n1-standard-4")),
           diskSize = Some(DiskSize(100)),
-          Some(ZoneName("us-east1-c")),
+          Some(runtimeFixtureZone),
```
cpate4, line 173: is there a reason to move this into the constructor? all specs will now have access to this, but I don't see it being used anywhere else

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/SSH.scala
```diff
@@ -33,7 +42,7 @@ object SSH {
       process = Process(
         scriptPath,
         None,
-        "BASTION_NAME" -> LeonardoConfig.Leonardo.bastionName,
+        "BASTION_NAME" -> LeonardoConfig.Azure.bastionName,
```
cpate4, line 44: see comment above re: needing this in other cloud providers.
do we have bastion hosts in google?

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/SSH.scala
```diff
@@ -46,38 +55,99 @@ object SSH {
...
-      _ <- loggerIO.info("Connecting via ssh client")
+
+      _ <- loggerIO.info(s"Connecting via ssh client hostname ${hostName} port $port")
       _ <- IO(client.connect(hostName, port))
-      _ <- loggerIO.info("Authenticating ssh client via password")
-      _ <- IO(client.authPassword(LeonardoConfig.Leonardo.vmUser, LeonardoConfig.Leonardo.vmPassword))
+
+      _ <- loggerIO.info("Authenticating ssh client ")
+      _ <-
+        if (sshConfig.cloudProvider == CloudProvider.Azure)
```
cpate4, line 76: This all looks very Azure specific - question re: re-usability for non-Azure environments

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/SSH.scala
```diff
@@ -46,38 +55,99 @@ object SSH {
...
+      client = OsLoginServiceClient
+        .create()
+
+      settings = client.getSettings()
+      _ <- loggerIO.info(s"settings ${settings}")
+
+      _ <- loggerIO.info("importing ssh public key")
+      _ <- IO(client.importSshPublicKey(request))
+
+    } yield SSHKeyConfig("sa_106387076428805685843", publicKey, privateKey)
```
cpate4: will subsequent runs / calls to this method fail if the key is already installed?
does this need to be unique on the system (and should we check for its existence first)

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/SSH.scala
```diff
@@ -46,38 +55,99 @@ object SSH {
...
+  def executeCommand(hostName: String, port: Int, command: String, sshConfig: SSHRuntimeInfo): IO[CommandResult] =
+    for {
+      output <- SSH.startSSHConnection(hostName, port, sshConfig).use { connection =>
+        executeCommand(connection.session, command)
+      }
+    } yield output
+
+  def executeGoogleCommand(project: GoogleProject, zone: String, runtimeName: RuntimeName, cmd: String): IO[String] = {
+    val sshCommand =
+      s"gcloud compute ssh --zone '${zone}' '${runtimeName.asString}' --project '${project.value}' --command=\"$cmd\" -- -tt"
```
cpate4: `-tt` - `#ftw`

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/apps/AppLifecycleSpec.scala
```diff
@@ -14,7 +14,7 @@ import org.http4s.headers.Authorization
 import org.http4s.{AuthScheme, Credentials}
 import org.scalatest.prop.TableDrivenPropertyChecks
 import org.scalatest.tagobjects.Retryable
-import org.scalatest.Assertion
+import org.scalatest.{Assertion, DoNotDiscover}
```
cpate4: Is this being used?

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoSuite.scala
```diff
@@ -246,26 +246,27 @@ trait NewBillingProjectAndWorkspaceBeforeAndAfterAll extends BillingProjectUtils
...
-      new LabSpec,
-      new RuntimeAutopauseSpec,
-      new RuntimePatchSpec,
-      new RuntimeStatusTransitionsSpec,
-      new NotebookGCECustomizationSpec,
-      new NotebookGCEDataSyncingSpec,
-      new RuntimeDataprocSpec,
-      new RuntimeGceSpec
+      new RuntimeCreationDiskSpec
+//      new RuntimeAutopauseSpec,
```
LizBaldo: You probably have this on your TODO list to clean up? Just making sure :)

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoTestUtils.scala
```diff
@@ -425,16 +438,6 @@ trait LeonardoTestUtils
         val status = Leonardo.cluster.getRuntime(googleProject, runtimeName).status
         status shouldBe ClusterStatus.Stopped
       }
-
-      // Verify notebook error
```
LizBaldo, line 429: Is this a known flaky test? Feels like something not necessarily worth removing, but I am not super opinionated either way :)

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/RuntimeFixtureSpec.scala
```diff
@@ -142,7 +142,9 @@ abstract class RuntimeFixtureSpec
 }
 
 object RuntimeFixtureSpec {
+  // TODO: can we delete this? why are our automation tests simulating env vars used by terra UI? I hope none of our tests depend on this?
```
LizBaldo, line 153: Well it does feel clunky, but we do rely on custom env variables to get workspace info to leo, so making sure that our runtime fixture includes them feels important to me, I don't think we should remove

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/SSH.scala
```diff
@@ -22,7 +31,7 @@ object SSH {
   val loggerIO: StructuredLogger[IO] = Slf4jLogger.getLogger[IO]
 
   // TODO: If multiple tests need to ssh into an azure VM: add a lock of sorts, only one tunnel at a time with same port
-  def startBastionTunnel(runtimeName: RuntimeName, port: Int = LeonardoConfig.Leonardo.defaultBastionPort)(implicit
+  def startBastionTunnel(runtimeName: RuntimeName, port: Int = LeonardoConfig.Azure.defaultBastionPort)(implicit
```
LizBaldo: Is it correct that this is not used at the moment, and is Azure specific? If so could you rename to ` startAzureBastionTunnel` and maybe add a comment that this method works, but is currently not used in favor of the `SSHConnection` one

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/SSH.scala
```diff
@@ -46,38 +55,99 @@ object SSH {
...
       _ <- IO(client.addHostKeyVerifier(new PromiscuousVerifier()))
-      _ <- loggerIO.info("Connecting via ssh client")
+
+      _ <- loggerIO.info(s"Connecting via ssh client hostname ${hostName} port $port")
       _ <- IO(client.connect(hostName, port))
-      _ <- loggerIO.info("Authenticating ssh client via password")
-      _ <- IO(client.authPassword(LeonardoConfig.Leonardo.vmUser, LeonardoConfig.Leonardo.vmPassword))
+
+      _ <- loggerIO.info("Authenticating ssh client ")
+      _ <-
```
LizBaldo, line 75: So if I understand correctly, we ssh via password for Azure VMs, and ssh via key for Google? If so could you add a little comment to the code?

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/SSH.scala
LizBaldo: What does the `"sa_106387076428805685843"` string correspond to? Should this be in a config somewhere?

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/SSH.scala
```diff
@@ -46,38 +55,99 @@ object SSH {
...
+      _ <- loggerIO.info(s"cmd output: $output")
+    } yield output
+  }
+
   def closeTunnel(tunnel: Tunnel): IO[Unit] =
     loggerIO.info(s"Killing tunnel via pid ${tunnel.pid}") >> IO(s"kill ${tunnel.pid}" !!)
 
   // Exec docs/examples: https://www.tabnine.com/code/java/methods/net.schmizz.sshj.connection.channel.direct.Session/exec
-  def executeCommand(session: Session, cmd: String): IO[CommandResult] =
+  private def executeCommand(session: Session, cmd: String): IO[CommandResult] =
```
LizBaldo, line 167: What is the difference between this and the `SSHRuntimeInfo` one? Would it make sense to rename to make clear what is associated with the Bastion method vs the ssh one?

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookHailSpec.scala
```diff
@@ -15,170 +15,16 @@ import scala.concurrent.duration._
...
-            // result.get should include("Coerced sorted dataset")
-
-            // Verify the Hail table
-            val tableResult = notebookPage.executeCellWithCellOutput("samples.count()").map(_.output.last)
-            tableResult shouldBe Some("2504") // rows
-          }
-        }
-      }
-    }
+    // This test is not a no-op, the fixture it mixes in creates a runtime by default with the `toolDockerImage` we override
```
LizBaldo, line 22: 👍

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookHailSpec.scala
```diff
@@ -15,170 +15,16 @@ import scala.concurrent.duration._
...
   implicit def ronToken: AuthToken = ronAuthToken.unsafeRunSync()
 
   // Should match the HAILHASH env var in the Jupyter Dockerfile
-  val expectedHailVersion = "0.2.126"
   val hailTutorialUploadFile = ResourceFile(s"diff-tests/hail-tutorial.ipynb")
   override val toolDockerImage: Option[String] = Some(LeonardoConfig.Leonardo.hailImageUrl)
   override val cloudService: Option[CloudService] = Some(CloudService.Dataproc)
 
   "NotebookHailSpec" - {
-    "should install the right Hail version" in { clusterFixture =>
```
LizBaldo, line 28: Nice! Deleting this makes a lot of sense, thank you :)

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookPyKernelSpec.scala
```diff
@@ -38,76 +41,9 @@ class NotebookPyKernelSpec extends RuntimeFixtureSpec with NotebookTestUtils {
...
-          val query =
-            """! bq query --format=json "SELECT COUNT(*) AS scullion_count FROM [bigquery-public-data.samples.shakespeare] WHERE word='scullion'" """
-          val expectedResult = """[{"scullion_count":"2"}]""".stripMargin
-
-          val result = notebookPage.executeCell(query, timeout = 5.minutes).get
-          result should include(expectedResult)
-        }
-      }
-    }
+    // TODO: [discuss] it does not seem like automation tests should be verifying security policies
```
LizBaldo: 🤔 I understand your point, but we should still have this test somewhere no?

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookPyKernelSpec.scala
```diff
@@ -18,6 +20,7 @@ class NotebookPyKernelSpec extends RuntimeFixtureSpec with NotebookTestUtils {
 
   "NotebookPyKernelSpec" - {
 
+    // TODO: [discuss] Debatable but I like a kernel smoke test... I could see getting rid of this entirely
```
LizBaldo, line 19: I would definitely keep this in for now. The leo dockefile does weird things  with its python path and making sure the kernel starts is important and could pass in terra docker but fail in leo

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookPyKernelSpec.scala
```diff
@@ -38,76 +41,9 @@ class NotebookPyKernelSpec extends RuntimeFixtureSpec with NotebookTestUtils {
...
-          val expectedResult = """[{"scullion_count":"2"}]""".stripMargin
-
-          val result = notebookPage.executeCell(query, timeout = 5.minutes).get
-          result should include(expectedResult)
-        }
-      }
-    }
+    // TODO: [discuss] it does not seem like automation tests should be verifying security policies
 
-    "should update dateAccessed if the notebook is open" in { runtimeFixture =>
```
LizBaldo, line 63: I feel like we should keep this since we rely on dateAccessed to be successfully updated to do our autopause. Maybe we could put the spleep back to 90, or 60 seconds instead of 120?

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookPyKernelSpec.scala
```diff
@@ -38,76 +41,9 @@ class NotebookPyKernelSpec extends RuntimeFixtureSpec with NotebookTestUtils {
...
-          expectedEVs.foreach { case (k, v) =>
-            val res = notebookPage.executeCell(s"os.getenv('$k')")
-            res shouldBe defined
-            res.get shouldBe s"'$v'"
-          }
-        }
-      }
-    }
-
-    // https://github.com/DataBiosphere/leonardo/issues/891
```
LizBaldo, line 103: So I actually caught a bug thanks to this test over the summer. And that is something that we might need to test via leo, in addition to the docker image, because of the dependency on the C bindings in the VM directly

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookRKernelSpec.scala
```diff
@@ -11,149 +12,32 @@ import scala.concurrent.duration._
...
-              |pi <- length(which(unlist(res)))*4.0/samples
-              |cat("Pi is roughly", pi, "\n")""".stripMargin
-
-          notebookPage.executeCell(sparkJob).get should include("Pi is roughly ")
-        }
-      }
-    }
-
-    // See https://github.com/DataBiosphere/leonardo/issues/398
-    "should be able to install mlr" in { runtimeFixture =>
```
LizBaldo, line 70: We have this exact same test in terra-docker already 👍

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/rstudio/RStudioSpec.scala
LizBaldo: Ohhh that is a good question, I did not even know that we supported RShiny apps, but I don't see how this test does that really. It is just another R package to install so it feels like a test that should be on terra-docker anyways https://shiny.posit.co/r/getstarted/shiny-basics/lesson1/index.html

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/runtimes/RuntimeCreationDiskSpec.scala
```diff
@@ -64,25 +64,18 @@ class RuntimeCreationDiskSpec
...
-              val res = notebookPage
-                .executeCell(
-                  "! df -H"
-                )
-                .get
-              res should include("/dev/sdb")
-              res should include("/home/jupyter")
-            }
-          }
+        output <- SSH.executeGoogleCommand(
```
LizBaldo, line 68: 😻

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/runtimes/RuntimeGceSpec.scala
LizBaldo: Thanks!

LizBaldo: I still can't believe this was ignored

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/runtimes/RuntimeSystemSpec.scala
LizBaldo: Is this why the full class is marked as `@DoNotDiscover`?

jdcanas: No, all specs in our suite are marked @DoNotDiscover, because we want to run them through the `LeonardoSuite`. Otherwise, sbt finds them when collecting the tests and runs the spec and the suite, causing any without this annotation to be run twice. 

Funnily enough, `AppLifecycleSpec` is an odd one out in this pattern and chooses to not be in a suite and not include this annotation. I have no idea if this is intentional and I'm not inclined to mess with it given our recent woes with the wrong tests being run in jenkins/gha due to suite nonsense 🤷

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/rstudio/RStudioSpec.scala
jdcanas: I would love to remove this and the other ~5 files that support it 😄 . @rtitle do you have any input here?

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoSuite.scala
jdcanas: For sure. Just keeping this until I get the iap proxy stuff working for a proper green run (currently pending buffer PR merge due to tests + buffer deploy + rawls helmfile update in my bee)

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoTestUtils.scala
jdcanas: Oh this isn't a test. This is our stop runtime api utility method (which allows lets tests decide whether the method monitors for them), which verifies that the jupyter service api is not accessible as part of its completion checks. This particular completion check seems like something that is testing google (really, having a unit test that ensures the google stop function is called when the API is hit is sufficient) and also something quite irrelevant from the concerns of every test that wants to stop a runtime. In fact, not all runtimes even have a jupyter server.

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/RuntimeFixtureSpec.scala
jdcanas: Yeah, its just a bit odd because it indicates we might be testing terra-ui functionality here (specifically the env vars that it will populate). For example, what if terra ui modifies these vars, our runtimes start failing for some reason related to the dependency, but our tests tell us nothing because they are hardcoded here and included in our test runtime creation

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/SSH.scala
jdcanas: Yeah, its only for azure but it is used (and required) for all azure ssh connections. I'll add a comment and rename.

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/SSH.scala
jdcanas: No, we don't use an SSH connection at all for azure, we execute the commands directly via `executeGoogleCommand`. The reasoning is a bit complicated. I'll add a detailed comment

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/SSH.scala
jdcanas: Ah, this in the UNIX name for the leonardo service account. Aka, the user that is set up on all the VMs for ssh access. I'll add it to a config.

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/SSH.scala
jdcanas: This private one interacts with the lib's ssh session to execute the command (internal implementation), and the public one creates the session and executes the command. I'll make sure the private/public modifiers are consistent, and I can rename this one to `startSessionAndExecuteCommand`

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookPyKernelSpec.scala
jdcanas: I think this can be unit tested in the routes?

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookPyKernelSpec.scala
jdcanas: Ah ok :(

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookPyKernelSpec.scala
jdcanas: I think this is fairly unit testable as well. It's essentially testing 'when i hit this api, this database field is updated'

This isn't even using the same API that terra-ui would be hitting to trigger the date accessed update, that would be the proxy traffic. (We only list passively when polling, we never `get` unless you are looking at the compute modal details).

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/RuntimeFixtureSpec.scala
LizBaldo: Very true, should this be moved to a terra ui integration test instead?

jdcanas: Perhaps... I'm not super sure what it does is the issue though. In other words, my question really is "if we remove this set up of custom env vars, will any of our assertions start to fail, which would indicate our tests are not necessarily testing what they should be". Ideally, I'd like to remove this consumer-specific logic and have our tests pass without it. if we find anything useful that we are testing about them, we can decide then if it should live in leo/terra-ui

jdcanas:
jenkins retest

cpate4 (commented):
I don't have much knowledge of the tests being removed ... but for what I did have knowledge of I commented on.

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoSuite.scala
jdcanas: That's just for fast turnaround time while debugging, reverted

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoConfig.scala
jdcanas: No, bastion is an azure specific product

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/ApiJsonDecoder.scala
jdcanas: I'm decoding another field from the `getRuntime` api response. We copy/pasted these models and decoders from the `http` module to the `automation` before the `core` module existed, and the payload of `getRuntime` has more fields now (some of which automation doesn't need, but now I need an additional field). I did try to switch over to the common model in `core`, but its pretty ingrained, and it doesn't make much sense to do this over swapping to the generated API client/models.

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/RuntimeFixtureSpec.scala
jdcanas: Yes, the zone is needed to ssh, so a few specs are consuming `runtimeFixtureZone` now instead of having this magic zone floating around.

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/SSH.scala
jdcanas: [No, bastion is an azure product](https://azure.microsoft.com/en-us/products/azure-bastion)

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/SSH.scala
jdcanas: Yeah, the subsequent else block handles this flow for google

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/SSH.scala
jdcanas: No, you can keep adding as many public keys as you want to the project metadata. I moved the username that was hardcoded here to a config, but it is the unix username of the leonardo sa, so it needs to be the same for all public keys uploaded for the purpose of sshing.

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/RuntimeFixtureSpec.scala
LizBaldo: Ah I see, fair enough, probably safe to remove then

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/SSH.scala
```diff
@@ -46,38 +54,117 @@ object SSH {
...
+  // This method is the main one for any interaction with GCP VMs
+  def executeGoogleCommand(project: GoogleProject, zone: String, runtimeName: RuntimeName, cmd: String): IO[String] = {
+    val dummyCommand =
+      s"gcloud compute ssh --zone '${zone}' '${LeonardoConfig.GCS.leonardoServiceAccountUsername}@${runtimeName.asString}' --project '${project.value}' --tunnel-through-iap -q --command=\"ls\" -- -tt"
+    val sshCommand =
+      s"gcloud compute ssh --zone '${zone}' '${LeonardoConfig.GCS.leonardoServiceAccountUsername}@${runtimeName.asString}' --project '${project.value}' --tunnel-through-iap -q --command=\"$cmd\" -- -tt"
+
+    for {
+      // Without first executing a dummy command, the output will contain a bunch of garble that google spits out because gcloud compute ssh runs ssh-keygen under the covers
+      // Unfortunately, they provide a way to pass args to the subsequent ssh call but not the keygen call, so we need to always execute a command and throw away the output before doing meaningful work
```
LizBaldo, line 153: 😱

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookPyKernelSpec.scala
LizBaldo: That would make sense, at this point this PR is so large though that I would punt that move to a follow up ticket

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookPyKernelSpec.scala
LizBaldo: Same comment as above then, seems like you uncovered a pattern of integration tests that should be unit tests instead

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookPyKernelSpec.scala
LizBaldo: Could you move this test back in?

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/runtimes/RuntimeGceSpec.scala
LizBaldo: Feels like we should push that ticket to be picked up next sprint

LizBaldo (approved):
LGTM! Feel free to merge after moving back in the test checking C bindings within notebooks :)

Dockerfile-tests
```diff
@@ -9,6 +9,21 @@ COPY project /app/project
...
+# Installing the package
+RUN mkdir -p /usr/local/gcloud \
+  && tar -C /usr/local/gcloud -xvf /tmp/google-cloud-sdk.tar.gz \
+  && /usr/local/gcloud/google-cloud-sdk/install.sh
+
+# Adding the package path to local
+ENV PATH $PATH:/usr/local/gcloud/google-cloud-sdk/bin
+
+RUN mkdir -p /root/.ssh
+
```
cpate4, line 28: This looks to be from here:
https://stackoverflow.com/questions/28372328/how-to-install-the-google-cloud-sdk-in-a-docker-image

... and adding the [official documentation](https://cloud.google.com/sdk/docs/install#linux) (for posterity's sake)

automation/project/build.properties
```diff
@@ -0,0 +1 @@
+sbt.version=1.9.2
```
cpate4: two questions ... 
- is this needed?
- and what if this ends up differing from the parent project build.properties?

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/ApiJsonDecoder.scala
cpate4: would it be possible to add a comment to that effect? e.g. what "forProduct16" is supposed to align with? (for our future-selves?)

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/SSH.scala
cpate4: a bastion is also a general concept - 🤪

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/SSH.scala
cpate4: Probably NOT for this code review (aka not a refactor code-review) ...
but this is a code smell - if-else blocks based on things that could/should otherwise be handled by an object of a certain type.

... and yes, I know we have this style of code everywhere

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/notebooks/NotebookPyKernelSpec.scala
cpate4: I am feeling this re: size of PR

http/src/main/resources/reference.conf
```diff
@@ -596,6 +596,45 @@ vpc {
         }
       ]
     }
+    # Allows SSH access from google IAP proxy
+    # IP list obtained from https://cloud.google.com/iap/docs/using-tcp-forwarding#create-firewall-rule
```
cpate4, line 600: Do we need all of these? ALA are we testing in/from a single environment?

cpate4 (approved):
This is massive (*gulp*)
Tests pass. Looking at @LizBaldo for confirmation on deleted tests. Just some minor comments / points-of-clarification

Dockerfile-tests
jdcanas: Yeah, I'll add a comment. Naturally, the official docs didn't work and I had to use the stack overflow docs

automation/project/build.properties
jdcanas: Oh, let me look into this. I think there was something weird about one of the SSH packages I needed where it needed sbt pinned to a lower version than the outer module, but I may have sleepily tried to execute the automation tests from the wrong directory and added this erroneously....

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/ApiJsonDecoder.scala
jdcanas: Hmm... this is a fairly common construct included hundreds of times in our code base (every single model we decode from JSON). The library we use for encoding/decoding (`circe`) needs you to explicitly state how many fields you are attempting to decode and the name of the field. See this file for more examples: https://github.com/DataBiosphere/leonardo/blob/develop/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodec.scala#L107. The number in the function name corresponds to the number of arguments to the function. Not sure why they needed to make a different function instead of using an  equivalent construct to `args...` 🤷

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/SSH.scala
jdcanas: Ah, TIL. Well, either way, `IAP` is the google equivalent product.

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/SSH.scala
jdcanas: Yeah, that's true, we do need to be better about higher levels of abstraction

http/src/main/resources/reference.conf
jdcanas: Oh, good point. I should be able to just use the one region in the automation tests here. Let me test this out...

jdcanas: Huh... ok I'm pretty confused. But we do apparently need all these (source) regions. 
```
terra-quality-eb725221/automation-test-ashspqrlz errored due to List(RuntimeError(ErrorReport(leonardo,Fail to create firewall due to unsupported Region europe-west1,Some(500 Internal Server Error),List(),List(),Some(class org.broadinstitute.dsde.workbench.leonardo.model.LeoException),Some(TraceId(3dfe4ca55a8a4e00864d380e2c15d72f/3939911508519804487))),None,2024-01-05T00:14:07Z,Some(TraceId(3dfe4ca55a8a4e00864d380e2c15d72f/3939911508519804487))))
```

jdcanas:
jenkins retest

jdcanas:
jenkins retest - fiab start

jdcanas:
jenkins retest - unrelated failure

jdcanas:
jenkins retest - jupyter/welder issue

PR #3937, opened 2023-11-14
RW-10598 Stop/start cromwell app on GCP via scaling replicas

Qi77Qi:
https://precisionmedicineinitiative.atlassian.net/browse/RW-10598

1. Change the way we stop/start cromwell App on GCP to use scaling replicas
2. Use a different default nodepool config for apps that rely on scaling replicas for pause/resume. This setting is currently 2 `n1-standard-4` nodes, which we might be able to optimize this over time (I have a GCP support ticket open regarding this), but I think this is ok for this PR since it's not making things any worse (without this, kube system pods won't transfer to the default nodepool and preventing the user nodepool from scaling down). This won't affect Terra since this PR doesn't change default nodepool for non `ALLOWED` and non cromwell on GCP apps. But I'll have a discussion with Moira to confirm the price increase is ok from AoU perspective

Tested on BEE that pause/resume works as expected. Also observed autoscaling kicks in properly to scale down the user nodepool

### Who tested and where

- [ ] I validated this change 

### DIFF
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -640,9 +640,9 @@ gke {
     nodepoolLockCacheMaxSize = 200
   }
   defaultNodepool {
-    machineType = "n1-standard-1"
-    numNodes = 1
-    autoscalingEnabled = false
+    machineType = "n1-standard-2"
+    numNodes = 2
+    autoscalingEnabled = true
     maxNodepoolsPerDefaultNode = 16
   }
   galaxyNodepool {
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -640,7 +640,7 @@ gke {
     nodepoolLockCacheMaxSize = 200
   }
   defaultNodepool {
-    machineType = "n1-standard-2"
+    machineType = "n1-standard-1"
     numNodes = 1
     autoscalingEnabled = false
     maxNodepoolsPerDefaultNode = 16
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -641,6 +641,12 @@ gke {
   }
   defaultNodepool {
     machineType = "n1-standard-2"
+    numNodes = 1
+    autoscalingEnabled = false
+    maxNodepoolsPerDefaultNode = 16
+  }
+  defaultAutoScalingNodepool {
+    machineType = "n1-standard-4"
     numNodes = 2
     autoscalingEnabled = true
     maxNodepoolsPerDefaultNode = 16
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -742,6 +742,7 @@ gke {
     enabled = true
     # App developers - Please keep the list of non-backward compatible versions in the list below
     chartVersionsToExcludeFromUpdates = []
+    numOfReplicas = 1
   }
   allowedApp {
     # If you update this here, be sure to update in the dockerfile
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/KubernetesAppConfig.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/KubernetesAppConfig.scala
@@ -33,6 +33,9 @@ sealed trait KubernetesAppConfig extends Product with Serializable {
   // corresponds to a specific app type and cloud provider.
   def cloudProvider: CloudProvider
   def appType: AppType
+
+  // Only CROMWELL, ALLOWED appTypes have replicas defined properly and used for pause/resume.
+  def numOfReplicas: Option[Int]
 }
 
 object KubernetesAppConfig {
@@ -70,6 +73,8 @@ final case class GalaxyAppConfig(releaseNameSuffix: ReleaseNameSuffix,
 
   val cloudProvider: CloudProvider = CloudProvider.Gcp
   val appType: AppType = AppType.Galaxy
+
+  def numOfReplicas: Option[Int] = None
 }
 
 final case class CromwellAppConfig(chartName: ChartName,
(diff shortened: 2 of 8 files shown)


http/src/main/resources/reference.conf
```diff
@@ -645,6 +645,12 @@ gke {
     autoscalingEnabled = false
     maxNodepoolsPerDefaultNode = 16
   }
+  defaultAutoScalingNodepool {
```
LizBaldo, line 648: Could you leave a comment here around the need o change the machine type and the number of nodes? Like a little summary of what you shared on slack
```regarding pause/resume GKE apps, I was able to get autopause to behave for the first app in the cluster…but I had to do a few tweaks in the default nodepool
move nginx to default nodpeool (making default nodepool to n1-standard-4 suffice nginx’s needs)
make sure metrics-server is moved to default nodepool (for this to work, I had to make the default nodepool having 2 nodes instead of 1)
so in short, this will increase cost for running an app on a GKE cluster…
default nodepool used to be 1 node of  n1-standard-1 ($34.6748175/month)..and now it’ll be 2 nodes of n1-standard-4 ($138.69927*2/month)```

File: http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala [hunk 2]
```diff
@@ -178,18 +182,18 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
...
+      machineConfig = req.kubernetesRuntimeConfig.getOrElse(
         KubernetesRuntimeConfig(
           config.leoKubernetesConfig.nodepoolConfig.galaxyNodepoolConfig.numNodes,
           config.leoKubernetesConfig.nodepoolConfig.galaxyNodepoolConfig.machineType,
-          config.leoKubernetesConfig.nodepoolConfig.galaxyNodepoolConfig.autoscalingEnabled
+          numOfReplicas.fold(
+            config.leoKubernetesConfig.nodepoolConfig.galaxyNodepoolConfig.autoscalingEnabled
+          )(_ =>
+            true
+          ) // if numOfReplicas is defined, then autoscaling has to be enabled because we scale up/down replicas to pause/resume apps.
```
LizBaldo, line 193: Gotcha, it should not impact things as they are until we enable autoscaling for Galaxy (AOU is not exposing this app correct?)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -1082,17 +1091,22 @@ class GKEInterpreter[F[_]](
...
-                dbApp.cluster.getClusterId,
-                KubernetesNamespace(dbApp.app.appResources.namespace),
-                KubernetesDeployment(dbApp.app.appName.value), // appNames are the same as deployments
-                count
-              )
-              .attempt
+            deployments <- kubeService.listDeployments(dbApp.cluster.getClusterId,
+                                                       KubernetesNamespace(dbApp.app.appResources.namespace)
+            )
+            attemptToStopResults <- deployments.parTraverse { deployment =>
```
LizBaldo, line 1097: Should this be renamed to `attemptToStartResults`?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -1082,17 +1091,22 @@ class GKEInterpreter[F[_]](
...
+                  deployment,
+                  count
+                )
+                .attempt
+            }
 
-            // Update nodepool status to Running and app status to Stopped
-            _ <- attemptToStart match {
+            // combineAll will collect the first error
+            _ <- attemptToStopResults.combineAll match {
```
LizBaldo, line 1109: Same comment here :)

LizBaldo (approved):
That makes sense to be :) I have a couple of small comments around documentation and some renaming but nothing major, thanks!

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
Qi77Qi: yep...this doesn't affect Galaxy, and AoU is not using Galaxy

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
Qi77Qi: ah yes...copy paste error :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/KubernetesAppConfig.scala
```diff
@@ -33,6 +33,9 @@ sealed trait KubernetesAppConfig extends Product with Serializable {
   // corresponds to a specific app type and cloud provider.
   def cloudProvider: CloudProvider
   def appType: AppType
+
+  // Only CROMWELL, ALLOWED appTypes have replicas defined properly and used for pause/resume.
+  def numOfReplicas: Option[Int]
```
jgainerdewar, line 38: Based on what you've said about how this works I assume `replicaCount` will always be either 0 or 1 right now. The Cromwell app won't Just Work with replica count > 1, might be good to have some validation somewhere checking that.

Qi77Qi: yea..so right now, I've defaulted cromwell replica to 1....and assume that's the number it needs to scale up to...down the line, I think we should parameterize that in the cromwell chart, and let Leo pass that in so that there won't be mismatch

PR #3943, opened 2023-11-16
[RW-11279] Support auto delete App

yonghaoy:
This is similar to `autoFreeze`, 

https://docs.google.com/document/d/1eWfY8PQoDKVOdIeREuEC9ObLhXlsTtG2ugLwuqov81k/edit#heading=h.kmuxa0wy0za9

- There is a new parameter autoDeleteThreshold in createAppRequest in Duration format
- A new column in App DB for AutoDeleteThresholdInMinutes
- A Monitor similar to autoFreezeMonitor to grab apps ready to delete then delete them. 

### Testing
Create 3 Apps, 

- 1st one with 10min threshold.  `yyu-rstudio-10m`
- 2nd app without threshold,  `yyu-rstudio-empty`
- 3rd app with value `Inf`.    `yyu-rstudio-inf`

Only 10 min app is deleted. And the other two has correct value in DB (i.e. 0)
 
Screenshot 2023-11-29 at 4 48 31 PM

__Jira ticket__: https://precisionmedicineinitiative.atlassian.net/browse/RW-11279

### What

- Delete auto delete app if threshold is specified in request

### Why

- This will help cost saving. 

### What to test

- 

### Who tested and where

- [ ] I validated this change 


http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -4031,6 +4031,10 @@ components:
         sourceWorkspaceId:
           type: string
           description: Optional id of source workspace if app is a clone.
+        autoDeleteThresholdInHours:
```
yonghaoy: where is the logic to insert this into DB table? Automattly in AppCompoment?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/AutoDeleteConfig.scala
```diff
@@ -0,0 +1,9 @@
+package org.broadinstitute.dsde.workbench.leonardo.config
+
+import scala.concurrent.duration.FiniteDuration
+
+case class AutoDeleteConfig(
```
yonghaoy, line 5: The config is in helmfile repo?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/AutoDeleteAppMonitor.scala
```diff
@@ -0,0 +1,112 @@
...
+    F: Async[F],
+    metrics: OpenTelemetryMetrics[F],
+    logger: StructuredLogger[F]
+  ): F[Seq[AppToAutoDelete]] = for {
+    candidates <- appQuery.getAppReadyToAutoDelete.transaction
+  } yield candidates
+
+  override def action(a: AppToAutoDelete, traceId: TraceId, now: Instant)(implicit as: Ask[F, AppContext]): F[Unit] =
+    for {
+      ctx <- as.ask
```
yonghaoy: the `ctx` and `as.ask ` is not regognized and incorrect

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/AutoDeleteAppMonitor.scala
```diff
@@ -0,0 +1,112 @@
...
+            )
+            _ <- appQuery.markAsDeleted(a.id, ctx.now).transaction
+          } yield ()
+        } else {
+          for {
+            _ <- KubernetesServiceDbQueries.markPreDeleting(a.id).transaction
+            deleteMessage = DeleteAppMessage(
+              a.id,
+              a.appName,
+              CloudContext.Gcp,
```
yonghaoy: I just hard coded this as GCP. Is that okay/necessary?

http/src/main/resources/reference.conf
```diff
@@ -1129,6 +1129,12 @@ autoFreeze {
   maxKernelBusyLimit = 24 hours
 }
 
+autoDelete {
+  enableAutoDelete= true
+  autoDeleteAfter = 30 minutes
+  autoDeleteCheckScheduler = 1 minutes
```
Qi77Qi: since realistically in UI, we'll probably give users options to delete after x days...maybe change the interval to be 5 hours or sth? (it does mean to test out auto delete from API, we'll need to wait 5 hrs tho)

leonardo-example.conf
```diff
@@ -182,6 +182,16 @@ autoFreeze {
   autoFreezeCheckScheduler = 1 minutes
 }
 
+#Configs to auto delete after idle period
```
Qi77Qi: we don't need to update this file anymore...don't think the content is kept up-to-date

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20231116_add_auto_delete_threshold_to_app_table.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="leonardo" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+    <changeSet logicalFilePath="leonardo" author="qi" id="yonghao_20231116_add_auto_delete_threshold_to_app_table">
+        <addColumn tableName="APP">
+            <column name="autoDeleteThresholdInMinutes" defaultValue="0" type="SMALLINT">
+                <constraints nullable="true"/>
```
Qi77Qi: think we should make it non-nullable..that way, rows without value will be populated with default value I think

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/package.scala
```diff
@@ -10,8 +10,10 @@ import java.time.Instant
 
 package object leonardo {
   type LabelMap = Map[String, String]
-  // this value is the default for autopause, if none is specified. An autopauseThreshold of 0 indicates no autopause
+  // this value is the default for autopause, if none is specified. An autopause Threshold of 0 indicates no autopause
   val autoPauseOffValue = 0
+  // this value is the default for auto delete, if none is specified. Threshold of 0 indicates no auto delete
```
Qi77Qi, line 15: we should update comment to indicate that this is only how we represent "indefinite" in DB

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -4031,6 +4031,11 @@ components:
         sourceWorkspaceId:
           type: string
           description: Optional id of source workspace if app is a clone.
+        autoDeleteThresholdDuration:
+          type: string
+          description: The duration of idle time to elapse before the app is
```
Qi77Qi: shall we mention "since last time the app is accessed by user". Also mention that `Inf` means indefinite, which means no auto delete

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/package.scala
yonghaoy: added one more line, not sure if that is accurate. PTAL

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/AutoDeleteAppMonitor.scala
```diff
@@ -0,0 +1,134 @@
...
+import org.broadinstitute.dsde.workbench.model.{TraceId, WorkbenchEmail}
+import org.broadinstitute.dsde.workbench.openTelemetry.OpenTelemetryMetrics
+import org.broadinstitute.dsp.ChartName
+import org.typelevel.log4cats.StructuredLogger
+
+import java.time.Instant
+import scala.concurrent.ExecutionContext
+
+/**
+ * This monitor periodically sweeps the Leo database and auto pause clusters that have been running for too long.
```
LizBaldo: auto pause -> auto delete?

LizBaldo (commented):
Hey @yonghaoy and @Qi77Qi  :) Happy to review this, but could you link the design doc in the PR? I do not have access to the JIRA ticket unfortunately. Would you also be okay if I ask other IA folks to have a look? It feels like a big change and I want to make sure that you are getting more eyes on it.

yonghaoy:
> Hey @yonghaoy and @Qi77Qi :) Happy to review this, but could you link the design doc in the PR? I do not have access to the JIRA ticket unfortunately. Would you also be okay if I ask other IA folks to have a look? It feels like a big change and I want to make sure that you are getting more eyes on it.

@LizBaldo We do not have design doc. I will create one since it would be easier to resolve comments over Google Doc. 
For now, I created a placeholder ticket: https://broadworkbench.atlassian.net/browse/IA-4717 and listed my design proposal there. PTAL. 
Also, feel free to add more reviewers. 
Thanks

yonghaoy:
@Qi77Qi @LizBaldo I just tested this change in my Bee and append a screenshot. PTAL

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/AutoDeleteAppMonitor.scala
```diff
@@ -0,0 +1,134 @@
...
+
+final case class AppToAutoDelete(id: AppId,
+                                 appName: AppName,
+                                 appStatus: AppStatus,
+                                 samResourceId: SamResourceId.AppSamResourceId,
+                                 creator: WorkbenchEmail,
+                                 chartName: ChartName,
+                                 cloudContext: CloudContext
+) {
+  def projectNameString: String = s"${CloudContext.Gcp}/${appName}"
```
Qi77Qi: `def projectNameString: String = s"${cloudContext}/${appName}"` instead of hard coding the cloud provider

Qi77Qi (approved):
One minor comment and the submodule change needs to be reverted, but otherwise looks great! :tada:

We should give IA team some time to review still tho

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20231116_add_auto_delete_threshold_to_app_table.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="leonardo" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+    <changeSet logicalFilePath="leonardo" author="yonghao" id="yonghao_20231116_add_auto_delete_threshold_to_app_table">
```
cpate4: The author is captured by the `author`-attribute. Should we include the author's name in the id for the changeset?

http/src/main/resources/reference.conf
```diff
@@ -1130,6 +1130,11 @@ autoFreeze {
   maxKernelBusyLimit = 24 hours
 }
 
+autoDelete {
+  enableAutoDelete= true
+  autoDeleteCheckInterval = 5 hours
```
cpate4: I believe that this is the frequency for which we are going to run the delete process (after going through this PR).

Some questions / comments:
- Would it be possible to name this value consistently with the `autoFreeze {...}` configuration above, and
- I am curious why this is set to `5 hours` and our thresholds are measured in minutes (see database migration)? Do we have any concern with not being expedient enough in deleting the environment per the user's request?

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -4031,6 +4031,11 @@ components:
         sourceWorkspaceId:
           type: string
           description: Optional id of source workspace if app is a clone.
+        autoDeleteThresholdDuration:
+          type: string
+          description: The duration of idle time (i.e. since last time the app is accessed by user) to elapse before the app is
+            auto deleted, e.g. "2 hours, 2h".  
+            A value of Inf means no auto delete.
```
cpate4, line 4038: This is the only place I am seeing this (`CreateAppRequest`). If I do a comparable search for `autopause*`, you will find it in a bunch of other places (most importantly `Get*`, `Update*`, and `Azure`-based runtimes).

That being said, how are we planning on reporting back to the UI the state of the `autoDelete` config and updating its value (outside of the initial `Create`?)

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/appRoutesModels.scala
```diff
@@ -33,7 +34,8 @@ final case class CreateAppRequest(kubernetesRuntimeConfig: Option[KubernetesRunt
                                   customEnvironmentVariables: Map[String, String],
                                   descriptorPath: Option[Uri],
                                   extraArgs: List[String],
-                                  sourceWorkspaceId: Option[WorkspaceId]
+                                  sourceWorkspaceId: Option[WorkspaceId],
+                                  autoDeleteThresholdDuration: Option[Duration]
```
mlilynolting, line 38: Is the choice of a `scala.concurrent` library intentional? It seems like a really nice expressive contract but I wonder about this note at the top of the [doc](https://www.scala-lang.org/api/current/scala/concurrent/duration/Duration.html): 

> This class is not meant as a general purpose representation of time, it is optimized for the needs of scala.concurrent.

Would it make sense to use a `java.time` library instead?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/AutoDeleteAppMonitor.scala
```diff
@@ -0,0 +1,134 @@
...
+    candidates <- appQuery.getAppsReadyToAutoDelete.transaction
+  } yield candidates
+
+  override def action(a: AppToAutoDelete, traceId: TraceId, now: Instant)(implicit F: Async[F]): F[Unit] =
+    for {
+      now <- F.realTimeInstant
+      loggingCtx = Map("traceId" -> traceId.asString)
+      _ <- a.cloudContext match {
+        case CloudContext.Gcp(googleProject) =>
+          if (a.appStatus == AppStatus.Error) {
```
mlilynolting, line 62: Does app status Error imply the corresponding Google cloud resources do not exist?

mlilynolting: And are there any other side effects (audit logs etc) that we're missing out on by not sending a full delete message to back Leo?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/AutoDeleteAppMonitor.scala
```diff
@@ -0,0 +1,134 @@
...
+                Some(traceId)
+              )
+              trackUsage = AllowedChartName.fromChartName(a.chartName).exists(_.trackUsage)
+              _ <- appUsageQuery.recordStop(a.id, now).whenA(trackUsage).recoverWith { case e: FailToRecordStoptime =>
+                logger.error(loggingCtx)(e.getMessage)
+              }
+              _ <- publisherQueue.offer(deleteMessage)
+            } yield ()
+          }
+        case CloudContext.Azure(_) => logger.info(loggingCtx)("Azure is not supported")
```
mlilynolting, line 90: would it make sense to filter Azure candidates out in `getCandidates`?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/AutoDeleteAppMonitor.scala
yonghaoy: I am not 100% sure how Leo deleting app works
I copied this delete app logic from DeleteApp endpoint logic: https://github.com/DataBiosphere/leonardo/blob/bd48425dab3c6cd40f6ed21d1c997f0db35f39cd/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala#L401L425

http/src/test/resources/reference.conf
```diff
@@ -297,6 +297,12 @@ autoFreeze {
   autoFreezeCheckScheduler = 2 second
 }
 
+autoDelete {
+  enableAutoDelete= true
+  autoDeleteAfter = 15 minutes
```
mlilynolting: Is this used?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
```diff
@@ -1248,6 +1249,12 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
         if (req.appType == AppType.Allowed)
           Some(config.leoKubernetesConfig.allowedAppConfig.numOfReplicas)
         else None
+      autoDeleteThresholdInMinutes = req.autoDeleteThresholdDuration match {
+        case None               => autoDeleteOffValue
```
mlilynolting, line 1253: autoPause maps None to its default value (15 minutes); maybe we could follow that pattern and change the configured value for `autoDeleteAfter` to 0 minutes?

http/src/test/resources/reference.conf
```diff
@@ -297,6 +297,12 @@ autoFreeze {
   autoFreezeCheckScheduler = 2 second
 }
 
+autoDelete {
+  enableAutoDelete= true
```
mlilynolting, line 301: It looks like this isn't used (and neither is enableAutoFreeze, lol) - should we remove it?

docker/run-mysql.sh
```diff
@@ -24,7 +24,7 @@ start() {
     docker run --rm \
                --link $CONTAINER:mysql \
                -v $PWD/docker/sql_validate.sh:/working/sql_validate.sh \
-               mysql:$MYSQL_VERSION \
+               mysql/mysql-server:$MYSQL_VERSION \
```
mlilynolting: what does this change do?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/AutoDeleteAppMonitor.scala
mlilynolting: Understood!

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
```diff
@@ -1248,6 +1249,12 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
...
+        case Some(Duration.Inf) => autoDeleteOffValue
```
cpate4, line 1254: I am wondering if the `.Inf` configuration is confusing and if we should just do "on" (because a value is configured) and off (in this case - `0min`). This would keep it consistent with my understanding of how `autopauseRuntime` is working as well.

http/src/test/resources/reference.conf
cpate4: I am not seeing this value be used anywhere - should we remove it?

cpate4 (changes requested):

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/AutoDeleteAppMonitor.scala
yonghaoy: done

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/appRoutesModels.scala
yonghaoy: Actually I am not sure the pros/cons. 
Seems this class is widely used in Leo: https://github.com/search?q=repo%3ADataBiosphere%2Fleonardo%20%22scala.concurrent.duration%22&type=code. so I inherit this pattern ?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
Qi77Qi: so the type for `autoDeleteThresholdDuration` is a `scala.concurrent.duration.Duration`, which can be either `Inf` or a `FiniteDuration`...

`on` isn't a valid value for `scala.concurrent.duration.Duration`...If you're suggesting adding another flag for indicating `on`/`off`, my thinking is that complicates logic becuz u'd be translating a combo of (`Boolean`, `FiniteDuration`) into the DB representation instead of just one value of `Duration`

http/src/main/resources/swagger/api-docs.yaml
Qi77Qi: that's good point...we should probably add in `get`...

`patch` is intentionally left out to keep this PR simple...we do plan to add it later tho

http/src/test/resources/reference.conf
yonghaoy: oh right, it should be deleted. Thanks!

http/src/main/resources/reference.conf
Qi77Qi: https://dspuxteam.invisionapp.com/freehand/Auto-pause-and-auto-delete-GldXLqNSb is some initial mocks..we expect users to enter x days for auto delete, which is why we don't have the monitor run aggressively

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
yonghaoy: I had offline conv with @Qi77Qi , and we feel a similar config like `autoFreezeAfter` is unnecessary in our case. 
The default value when None is present should always be "no auto delete - 0".
We do not know there is a case we want to have different config value per env. So autoDeleteConfig just be:
```
autoDelete {
  enableAutoDelete= true
  autoDeleteCheckInterval = 1 second
}
```

Qi77Qi: for backwards compatibility, we don't want to turn on autodelete for all users when rolling this out...no explicit auto delete should mean no auto delete...we want to be very explicitly opt in regarding delete

http/src/test/resources/reference.conf
yonghaoy: I will have this enabled and config it to run every 1 second, otherwise, my `AutoDeleteAppMonitorSpec` will not pass. 
(Similarly, `enableAutoFreeze ` is needed for `AutopauseMonitorSpec` I think)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/AutoDeleteAppMonitor.scala
Qi77Qi: Azure apps will have the threashold being 0, so they'll be filtered out already. The code as is will be easier to support Azure down the line as well

docker/run-mysql.sh
yonghaoy: ah, sorry, that is a mistake.  Will revert it back. 
(I am using Apple Chip Mac and it does not support Mysql 5. So as a workaround, I change here to mariaDb, forgot to clean it up.)

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/appRoutesModels.scala
Qi77Qi: this is intentional...a threshold is a duration..not a timestamp

> It seems like a really nice expressive contract

what do you mean?

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20231116_add_auto_delete_threshold_to_app_table.xml
yonghaoy: are you suggesting change id 
from `yonghao_20231116_add_auto_delete_threshold_to_app_table`
to `20231116_add_auto_delete_threshold_to_app_table`?
I followed this conversion in the few previous changeset (20231002, and 20231030). Did not notice that some changesets' ID does not have author. will remove it then...

http/src/test/resources/reference.conf
yonghaoy: autoDeleteAfter will be removed. The other two is needed for running Monitor unit tests

http/src/main/resources/reference.conf
yonghaoy: 1: 5 hours is came up from the mock (and probably some user study) from AoU. 
The product decision is to support this via X days (7 day, 14 day, 30 day). So we choose 5 hrs because 1 minute it too aggressive for days duration. Change to 2 hrs for now as start, and we might came back to modify this later once we received more user feedback. 
2: 5 hours here is the frequency of this Monitor to run. It is difference from the threshold in minutes. Those are all Duration format and can be converted by some Scala library. 
If you prefer to be consistent of using minutes, I can try to change it to `300 minutes`?  I am not sure how this string is converted to Scala code, it is not in `JsonCodec`. @Qi77Qi is that correct?

http/src/main/resources/swagger/api-docs.yaml
yonghaoy: I will use follow up few PRs to support get/update/list to make it easier to review and write. 
(there is no UpdateApp api, supporting this would add couple hundreds line of code...)

http/src/main/resources/reference.conf
Qi77Qi: these configs are also `FiniteDuration`...we should use the most convenient unit (hours in this case)...having durations to be `FiniteDuration` allows us to convert them to appropriate unit when needed

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
yonghaoy: I did consider to add extra AutoDeleteEnabled parameter in both Db and API interface but decide not to do it because the extra value is a bit unnecessary and add extra complexity actually. And even for AutoPause, I can pass empty AuthPauseThreshold, 0, to represent no auto pause. This is more complicated and confusing to us because there are too many ways to represent no auto pause. 

One threshold value field is good enough in this case. 
Also  `Inf` is the String represent in Duration: 
https://www.scala-lang.org/api/2.12.4/scala/concurrent/duration/Duration$.html#Inf:scala.concurrent.duration.Duration.Infinite
so I do not think it is confusing. 

happy to chat more. Or in this doc: https://docs.google.com/document/d/1eWfY8PQoDKVOdIeREuEC9ObLhXlsTtG2ugLwuqov81k/edit#heading=h.kmuxa0wy0za9

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/appRoutesModels.scala
mlilynolting: > this class is widely used in Leo

You're right - there's definitely a precedent. 👍 

> what do you mean [by an 'expressive contract']?

I meant that the Duration class looks pretty easy to use, powerful, and intuitive to understand.

The biggest thing I wanted to flag was the note in the javadoc that said it was "not meant as a general purpose representation of time" which made me think `java.time` [might be considered more appropriate for this use case](https://stackoverflow.com/questions/3614380/whats-the-standard-way-to-work-with-dates-and-times-in-scala-should-i-use-java). But a Java lib is less convenient than a Scala one!

All told, this choice is OK with me, not a blocker.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/AutoDeleteAppMonitor.scala
mlilynolting: Won't it generate a lot of logs, one per Azure app, every time the monitor checks?

http/src/test/resources/reference.conf
mlilynolting: Oh! But I don't see that the boolean value is actually used anywhere. Would the actual pause/delete functionality actually be turned off if you set this to `false`?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/AutoDeleteAppMonitor.scala
Qi77Qi: I was wondeirng the same thing..but for now, I think keeping the same logic is safe to do

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/AutoDeleteAppMonitor.scala
yonghaoy: - There is a log about "monitor starts". We will have that even without Azure apps
- When app is about to delete, there will be another log. Since Azure will not returned from getCandidates (because threshold is 0), there will be no log for Azure apps

Qi77Qi: @mlilynolting  no...becuz they won't be returned to begin with from the query (filtered out in https://github.com/DataBiosphere/leonardo/pull/3943/files#diff-370c82ef7303ffa5487e72ee6d57f15a301110695a3a07f298f67a7f105190b1R379 already)...this branch here is not a normal path before we support Azure

http/src/test/resources/reference.conf
yonghaoy: Interesting... That config is not used which means auto pause and auto delete will always be on...
Wondering if we should remove that config if we do not have a case to disable such feature in some env?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/AutoDeleteAppMonitor.scala
mlilynolting: Oh, I see! Thanks 👍

http/src/test/resources/reference.conf
Qi77Qi: I think we can remove the flag...I'm also fine if u want to actually do a branching logic in `boot.scala` based on the flag

mlilynolting: At least I don't think we need to add the new unused config field. (it's probably a good idea to remove the one for autopause too, but that's not your problem lol)

http/src/main/resources/reference.conf
mlilynolting: Do we need to let users know that durations < 5 hours won't necessarily be honored?

Qi77Qi: in UI mock, the unit is days...so I think we're fine in terms of end users...can potentially add this in the API doc mentioning it's checked every x hours only (the 5 hrs is a bit random, the main idea is to not have it run too frequent since it doesn't need to be)

http/src/test/resources/reference.conf
mlilynolting: Adding the branching logic in Boot.scala would also be fine! And that would make it easy to disable the feature if there were some kind of issue

Qi77Qi:
```
16:08:58 21:08 ERR error="1 execution errors:\nmartha: timed out waiting for healthy (Degraded: martha-deployment: Back-off restarting failed container)\n"
```

going to merge as is since things passed earlier before merging in main branch

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
cpate4: I think my preference is to not support `INF` at all.
I believe (if I am reading this correctly) that this is in a public facing API and would prefer to simple not support it as a valid value.

Qi77Qi: we should probably move convo to https://github.com/DataBiosphere/leonardo/pull/3980

but to ur comment, how would you represent the case of not having `autoDelete` enabled?...

fwiw, `Inf` is a valid case for `Duration`

PR #3948, opened 2023-11-20
update http4s version to M40 and update workbench-libs which had the …

cindy-broadinstitute:
…corresponding change

__Jira ticket__: https://broadworkbench.atlassian.net/browse/IA-4705

### What

-

### Why

-

### What to test

- 

### Who tested and where

- [ ] I validated this change 

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -9,7 +9,7 @@ object Dependencies {
   val automationGoogleV = "1.30.5"
   val scalaLoggingV = "3.9.5"
   val scalaTestV = "3.2.17"
-  val http4sVersion = "1.0.0-M38" //Do not upgrade to M40 until workbench-libs does (missing M40 for blaze as of 08/2023
+  val http4sVersion = "1.0.0-M40" //Do not upgrade to M40 until workbench-libs does (missing M40 for blaze as of 08/2023
   val slickV = "3.4.1"
   val guavaV = "32.1.3-jre"
   val monocleV = "3.2.0"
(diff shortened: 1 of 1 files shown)


project/Dependencies.scala
```diff
@@ -9,15 +9,15 @@ object Dependencies {
   val automationGoogleV = "1.30.5"
   val scalaLoggingV = "3.9.5"
   val scalaTestV = "3.2.17"
-  val http4sVersion = "1.0.0-M38" //Do not upgrade to M40 until workbench-libs does (missing M40 for blaze as of 08/2023
+  val http4sVersion = "1.0.0-M40" //Do not upgrade to M40 until workbench-libs does (missing M40 for blaze as of 08/2023
```
cpate4, line 12: out of curiosity, are we using any of the features in 1.0.0-MXX branch?, and if not, does it make sense to go with their officially released version of [v0.23.24](https://github.com/http4s/http4s/releases/tag/v0.23.24)?

Qi77Qi: unfortunately, 1.0.0 is not binary compatible with 0.23 series...

lucymcnatt (approved):

lucymcnatt:
looks like the are some issues downloading http4s packages

project/Dependencies.scala
cindy-broadinstitute: The issue is finding all the packages that have the updated cats-effect (3.5.+)

Blaze, a subproject of http4s, did not support the cats-effect change until recently (even though we had upgrade our cats-effect 2 months ago). 

Blaze can work against 1.0.0-M39, but I couldn't find one for 0.23.23 or 0.23.24.

But with 1.0.0-M40, metrics.prometheus is not supported. 

So now I am back trying 0.23.23 and trying to find compatible Blaze and metrics.prometheus. 

At this point it might be better to back out cats-effect until all our packages that use it are upgraded to support the major change in 3.5.

cindy-broadinstitute:
going to downgrade cats effect instead

PR #3949, opened 2023-11-20
[WM-2380] Use WSM database name to find Azure database name for app install

salonishah11:
__Jira ticket__: https://broadworkbench.atlassian.net/browse/WM-2380

## Summary of changes

This PR changes the way the Azure database names for apps is found and set in helm values. Because of workflow cloning, when a database is cloned by WSM, the Azure database name of the app no longer starts with app prefix. Hence instead of find the Azure database name of an app by checking if the name starts with certain prefix it will now use the WSM database name to find the Azure database name for that database.

### Why

- To support workflow cloning

## Testing these changes

I manually tested these changes in a BEE:
- Tested that launching of apps (WDS, WORKFLOWS, CROMWELL_RUNNER) from UI all work and launch as expected
-  Manually created a cbas database and Workflows app managed identity and then launched WORKFLOWS and CROMWELL_RUNNER app from UI and they attached to the manually created resources as expected
- These changes were also tested along side the WSM changes to created cloned database with dump and restore steps from https://github.com/DataBiosphere/terra-workspace-manager/pull/1524 and Leo was able to successfully attach the WORKFLOWS app to the cloned database

### Who tested and where

- [x] I validated this change 


dillydally414 (approved):

salonishah11:
jenkins retest

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/app/CromwellRunnerAppInstall.scala
```diff
@@ -130,12 +131,15 @@ class CromwellRunnerAppInstall[F[_]](config: CromwellRunnerAppConfig,
...
-    (dbNames.find(name => name.startsWith("cromwell") && !name.startsWith("cromwellmetadata")),
-     dbNames.find(_.startsWith("tes")),
-     dbNames.find(_.startsWith("cromwellmetadata"))
-    )
-      .mapN(CromwellRunnerAppDatabaseNames)
+  def toCromwellRunnerAppDatabaseNames(
+    dbResources: List[WsmControlledDatabaseResource]
+  ): Option[CromwellRunnerAppDatabaseNames] =
+    (dbResources
+       .find(db => db.wsmDatabaseName.startsWith("cromwell") && !db.wsmDatabaseName.startsWith("cromwellmetadata"))
```
lucymcnatt, line 138: since this logic depends on the db name starting with the app type, is it possible to rename the `cromwellmetadata` db to something that won't cause issue since we also have a `cromwell` db type?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/app/CromwellAppInstall.scala
```diff
@@ -139,9 +140,13 @@ class CromwellAppInstall[F[_]](config: CoaAppConfig,
       cbasDao.getStatus(baseUri, authHeader).handleError(_ => false)
     ).sequence.map(_.forall(identity))
 
-  private def toCromwellAppDatabaseNames(dbNames: List[String]): Option[CromwellAppDatabaseNames] =
-    (dbNames.find(_.startsWith("cromwell")), dbNames.find(_.startsWith("cbas")), dbNames.find(_.startsWith("tes")))
-      .mapN(CromwellAppDatabaseNames)
+  private def toCromwellAppDatabaseNames(
+    dbResources: List[WsmControlledDatabaseResource]
+  ): Option[CromwellAppDatabaseNames] =
+    (getAzureDatabaseName(dbResources, "cromwell"),
```
lucymcnatt, line 146: could there be a problem with the `cromwellmetadata` db getting selected here?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/app/CromwellRunnerAppInstall.scala
salonishah11: The database names here are assumed to be prefixed with the prefix mentioned in the databases list in each AppInstall (for example [this](https://github.com/DataBiosphere/leonardo/blob/9354890a356334202adff9d83074811f40ee328a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/app/CromwellRunnerAppInstall.scala#L31-L36)) instead of the app type. I think we can rename the `cromwellmetadata` database to something else but that would be a bigger lift and out of context of this PR.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/app/CromwellAppInstall.scala
salonishah11: I don't think `cromwellmetadata` would appear here because the `dbResources` is fetched from WSM or created based on the databases that apps support and Cromwell app doesn't support `cromwellmetadata` database. It only support `cromwell`, `tes` and `cbas` database (see [here](https://github.com/DataBiosphere/leonardo/blob/9354890a356334202adff9d83074811f40ee328a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/app/CromwellAppInstall.scala#L33-L38)).

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/app/CromwellRunnerAppInstall.scala
lucymcnatt: that works for me, it feels a little brittle at the moment.

lucymcnatt (approved):

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/app/CromwellRunnerAppInstall.scala
salonishah11: Yeah I agree that this check does feel little brittle but it will be hard to change the names of database as it will require helm changes too. In case it helps, the main logic for the condition that is being checked here is same as before but only uses the WSM database name instead of directly Azure name.

salonishah11:
jenkins retest

salonishah11:
jenkins retest

salonishah11:
jenkins retest

salonishah11:
jenkins retest

salonishah11:
jenkins retest

salonishah11:
jenkins retest

salonishah11:
jenkins retest

PR #3962, opened 2023-11-22
IA-4729 Support running Leonardo in Azure

Qi77Qi:
Goal of this PR is to support running `Leonardo` in Azure without having to initialize all GCP specific dependencies, with reasonably minimal code change required. 

See [this](https://docs.google.com/document/d/1Fdy-H9ZPe3_K9i-3Sqw5CNMYhCoRyTpJVuyjG9_E1cE/edit) for more details

~~1. When `leoServiceAccountJsonFile` is defined, Leonardo will be run in GCP mode; Otherwise, it's running in Azure mode~~ (needs larger change where we use potentially different way to identify Leonardo to Sam)
2. In `HttpRoutes`, depending on whether GCP specific services are defined, we're enabling different set of routes. When `GCPModeSpecificServices` is defined, we're running in GCP mode; otherwise in Azure mode
3. Moved GCP specific dependencies out of `LeoAppServiceInterp` class definition, and instead specific methods that rely on GCP will "summon" the dependency implicitly at call site.
4. In `boot`, we're initializing different set up "streams" depending on whether we're running in Azure or GCP mode
5. Added `GCPModeSpecificDependencies` for `LeoPubsubMessageSubscriber` such that a different set of messages are processed depending on whether these dependencies are defined. When `GCPModeSpecificDependencies` is defined, leo is run in `GCP` mode; otherwise in Azure mode.
6. After this PR, GCP mode should function as before; Azure mode isn't fully supported yet. The remaining work to enable running `Leonardo` in Azure is to define proper `CloudTopicPublisher` and `CloudSubscriber` properly (@giventocode is working on this)

### What to test

- 

### Who tested and where

- [ ] I validated this change 


http/src/main/resources/reference.conf
```diff
@@ -6,6 +6,8 @@
 application {
   applicationName = "leonardo"
   #leoGoogleProject = "broad-dsde-dev"
+  # Defining leoServiceAccountJsonFile will run Leonardo in GCP mode; otherwise, it will run in Azure mode.
```
rtitle, line 9: Wondering if it's worth an explicit `application.mode = Azure` config 🤔

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/AzureSubscriber.scala
```diff
@@ -0,0 +1,13 @@
+package org.broadinstitute.dsde.workbench.leonardo.dao
+
+import fs2.Stream
+import org.broadinstitute.dsde.workbench.model.TraceId
+
+import java.time.Instant
+
+//TODO: Jesus will fill this out
+trait AzureSubscriber[F[_], A] {
```
rtitle: Probably needs to be converted to Scala, but example of using Azure Service Bus queue from Java: https://github.com/DataBiosphere/stairway/blob/develop/stairway-azure/src/main/java/bio/terra/stairway/azure/AzureServiceBusQueue.java

http/src/main/resources/reference.conf
Qi77Qi: maybe..the thing I want to avoid is having 
```
mode = Azure/GCP
leoServiceAccountJsonFile=<path>
```
So now we'll need to validate the 4 permutations, and reject the 2 invalid cases

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppService.scala
```diff
@@ -14,7 +15,11 @@ trait AppService[F[_]] {
     appName: AppName,
     req: CreateAppRequest,
     workspaceId: Option[WorkspaceId] = None
-  )(implicit as: Ask[F, AppContext]): F[Unit]
+  )(implicit
+    as: Ask[F, AppContext],
+    googleResourceService: GoogleResourceService[F],
```
rtitle, line 20: It would be nice to split `AppService` into `AppService` and `AppV2Service` or something (similar to `RuntimeService` and `RuntimeV2Service`). Then we could just not instantiate `AppService` for Azure.

rtitle (commented):
Need to spend more time in Boot.scala 😅 

Overall the approach looks good -- basically just abstracting cloud-specific functionality to areas which can be enabled/disabled.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppService.scala
Qi77Qi: I'm not sure...reason why we had one `AppService` is I'm hoping we can reuse code more easily this way...also, v2 will need to support GCP eventually...providing method level in service feels will still be needed probably...

Qi77Qi:
superceded by https://github.com/DataBiosphere/leonardo/pull/4162

PR #3971, opened 2023-11-28
feat(IA-4587): add grafana metrics for failed pubsub messages

cpate4:
__Jira ticket__: https://broadworkbench.atlassian.net/browse/IA-4587

### What

- refactored duplicate code that handled updateApp error messages, and
- refactored out handling of pubsub metric recording to facilitate failures along the same vectors

### Why

- originally the ticket was for only updateApp failures, but we actually have no visibility for any pubsub message failures

### What to test

- 

### Who tested and where

- [x] I validated this change 


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -209,6 +177,41 @@ class LeoPubsubMessageSubscriber[F[_]](
       _ <- F.delay(
         event.consumer.ack()
       )
+      _ <- recordMessageMetric(event)
+    } yield ()
+
+  private def processFailed(ctx: AppContext, event: Event[LeoPubsubMessage], e: Throwable)(implicit ev: Ask[F, AppContext]): F[Unit] = {
```
cpate4: @Qi77Qi - I wanted to bounce this refactor-ing off of you. I think it follows the cats-effect syntax. Trying to make the processing more testable ... working on this now.

LizBaldo (commented):
I really like it! I just have a couple of questions to dig into the potential consequences a bit (mostly me showing off my ignorance on the subject):
- Would there be any consequence in terms of metrics / logging cost in issuing a log for every pubsub message? Maybe Rob and/or Qi  would have some good context here
- Would it be possible to extend on the unit test to confirm that the metric is being issued for both the happy and sad path?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
Qi77Qi: this is great...for testing, u can do `private[<whatever package>]`, that way you'll be able to access this in unit tests...

Qi77Qi:
@LizBaldo I don't see additional logs introduced in this PR...can u link the line u're referring to? (in case I'm just missing it)

LizBaldo:
> @LizBaldo I don't see additional logs introduced in this PR...can u link the line u're referring to? (in case I'm just missing it)

🤦‍♀️ nvm, it is just a refactoring of existing code and no new feature was introduced, forget I said anything about a potential increase in the number of metrics issued 😬

Qi77Qi:
@LizBaldo we can have as many metrics as we want..logs are sent to GCP, and is the one we need to be a bit more careful about not generating too many..the existing logs are all for error path tho, which are legit use cases of logs

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -209,6 +177,46 @@ class LeoPubsubMessageSubscriber[F[_]](
...
+        logger.error(ctx.loggingCtx, e)(
+          "Fail to process retryable pubsub message due to Google API call failure"
+        ) >> F.delay(event.consumer.nack())
+      case _ =>
+        logger.error(ctx.loggingCtx, e)("Fail to process pubsub message due to unexpected error") >> ack(event)
+    }
+    recordMessageMetric(event, isFailure = true) >> handleErrorMessages
+  }
+
+  private def recordMessageMetric(event: Event[LeoPubsubMessage], isFailure: Boolean = false): F[Unit] =
```
Qi77Qi: nit, but `isFailure` equals true/false is a bit harder to reason abt than `isSuccess=true/false` :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -221,7 +229,12 @@ class LeoPubsubMessageSubscriber[F[_]](
                                 4 minutes,
                                 4.5 minutes
       )
-      metricsName = s"pubsub/ack/${event.msg.messageType.asString}"
+      messageType = event.msg.messageType.asString
+      metricsName = e match {
+        case Some(e) =>
+          s"pubsub/fail/${messageType}/${e.getClass.toString.split(".").last}"
+        case None => s"pubsub/ack/${messageType}"
+      }
```
cpate4, line 237: @lucymcnatt - here is the update to include / capture the error type. my thought here is that this will correlate to the log messaging that happens in `processMessageFailure(...)` and will give us a starting place to get more information. thoughts?

cpate4:
jenkins retest

cpate4:
jenkins retest

cpate4:
jenkins multi-test

cpate4:
jenkins retest

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
lucymcnatt: I think this is a good starting point, and we can iterate if we need to (if that's too granular/not granular enough)

lucymcnatt (approved):

cpate4:
jenkins retest

PR #3987, opened 2023-12-04
[IA-4679] Move some AKS helm chart pull out of leo dockerfile

LizBaldo:
__Jira ticket__: https://broadworkbench.atlassian.net/browse/IA-4679

### What

- Now that the helm scala sdk has the ability to update a helm repo and pull charts, we can remove it from the leonardo's dockerfile, and perform the pull directly before the helm install step. I am hoping that we can keep the local `leonardo` directory for caching.

### Why

- This is the first step in enabling k8s apps to release independently of Leo

### What to test

- 

### Who tested and where

- [ ] I validated this change 


http/src/main/resources/reference.conf
```diff
@@ -390,7 +390,7 @@ azure {
    }
 
    hail-batch-app-config {
-     chart-name = "/leonardo/hail-batch-terra-azure"
+     chart-name = "oci://terradevacrpublic.azurecr.io/hail/hail-batch-terra-azure"
```
rtitle: Looks like OCI is supported as of helm 3.8.0: https://helm.sh/docs/topics/registries/

I was wondering if we need to update helm but seems we're on 3.11.2 today: https://github.com/DataBiosphere/leonardo/blob/develop/Dockerfile#L52

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -205,6 +205,10 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
         s"Relay listener values for app ${params.appName.value} are ${values.asString}"
       )
 
+      // Pull Listener helm chart
+      _ = helmClient
```
rtitle: I think this would need to be 
```
_ <- helmClient.updateAndPullChart(config.listenerChartConfig.chartName, config.listenerChartConfig.chartVersion, "/leonardo").run(authContext)
```

Since it's an `F` effect, it probably is not being run if you just `_ = helmClient.updateAndPullChart(..)`.

local/helm-scala-sdk
rtitle, line 1: Did you intend to make this a new submodule?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
LizBaldo: Oh weird, but I am actually getting a type mismatch if I use `<-`

local/helm-scala-sdk
LizBaldo: It has always been a submodule, for running leo locally no?

project/Dependencies.scala
```diff
@@ -26,7 +26,7 @@ object Dependencies {
   val workbenchOauth2V = s"0.5-$workbenchLibsHash"
   val workbenchAzureV = s"0.6-$workbenchLibsHash"
 
-  val helmScalaSdkV = "0.0.20"
+  val helmScalaSdkV = "0.0.8.5"
```
rtitle, line 29: This seems like a downgrade?

http/src/main/resources/reference.conf
LizBaldo: We might very well be good, I have not yet tested hail batch on my BEE yet :)

project/Dependencies.scala
LizBaldo: I know, something fishy happened in helm scala sdk, but I triple checked on artifactory and this is the latest published version

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
rtitle: Oh hm - maybe this?
```
_ <- helmClient.updateAndPullChart(config.listenerChartConfig.chartName, config.listenerChartConfig.chartVersion, "/leonardo").run(())
```

Since it's defined as a `Kleisli[F, Unit, Unit]`:
https://github.com/broadinstitute/helm-scala-sdk/blob/master/src/main/scala/org/broadinstitute/dsp/HelmAlgebra.scala#L20

(It probably doesn't need to be a `Kleisli` at all, which is where the `.run` comes from..

local/helm-scala-sdk
rtitle: So it is. Ok nevermind 😅

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
LizBaldo: Ahhh, I had tried `.run()`, but not `run(())` 😓

http/src/main/resources/reference.conf
```diff
@@ -230,7 +230,7 @@ azure {
 
    coa-app-config {
      instrumentation-enabled = false
-     chart-name = "/leonardo/cromwell-on-azure"
+     chart-name = "cromwell-helm/cromwell-on-azure"
```
rtitle, line 233: Am confusing myself -- this is just referencing the remote chart (in the repo) right?

If this works then do we need to pull the charts to the filesystem `/leonardo` at all?

E.g. does this work if you just remove the `updateAndPullChart` calls?

LizBaldo: I guess that it probably does work, but my concern here is handling the caching so we don't pull the chart at every app creation 🤔

rtitle: Maybe `reference.conf` should still have `chart-name = "/leonardo/cromwell-on-azure"` so it uses the cached version.

LizBaldo: I could try to do the `updateAndPullChart` call using the remote chart, and then doing the install with the local chart? But I don't know if the `updateAndPUllChart` is clever enough to not pull every time if the chart already exist in the destination directory

rtitle: And more complexity about caching (sorry):
* What happens if multiple things concurrently try to pull the chart to the filesystem (e.g. do we need a lock or something)?
* What can trigger invalidating the cached chart and pulling a new chart version? That could be for a separate PR, but that will become important when we can dynamically update chart versions, right?

rtitle: >  I don't know if the updateAndPUllChart is clever enough to not pull every time if the chart already exist in the destination directory

Yeah, not sure either 🤔  -- I'm guessing it's not clever.

Because caching is hard -- what if we just don't do it for the first pass and use the remote chart every time? These charts are all in GCS I think, and hopefully pretty reliable..

Caching could be an optimization if we encounter issues.

LizBaldo: Yeah, that's why maybe we could try without caching and see if impacts app creation time too much?

LizBaldo: Haha looks like we are in agreement :) But that means not even using the new PullChart function 😭

rtitle: Yeah 😿

rtitle (approved):
The code with no caching looks good to me -- should make sure that all app types can launch.

http/src/main/resources/reference.conf
LizBaldo: Yeah I am having some trouble making this work. The helm install step is not happy with getting the remote chart directly
```unable to lookup chart "oci://terradevacrpublic.azurecr.io/hail/hail-batch-terra-azure", missing registry client```

project/Dependencies.scala
calypsomatic: Hmm I wonder if I did something wonky with the version when I was updating it...

http/src/main/resources/reference.conf
rtitle: Hm -- 👁️

calypsomatic (commented):
Just clarifying - as currently implemented, this has no caching and will pull the chart at every install?

http/src/main/resources/reference.conf
LizBaldo: The more I read on the subject, the more I suspect that we need to pull the chart before installing. I don't think that install alone is supported for OCI

LizBaldo:
@calypsomatic yeah, the caching feels very tricky to implement, and even the install from remote directly is tricky for a few charts (mainly the hail batch one, and the terra-app-setup ones) :(

calypsomatic (approved):

calypsomatic:
> @calypsomatic yeah, the caching feels very tricky to implement, and even the install from remote directly is tricky for a few charts (mainly the hail batch one, and the terra-app-setup ones) :(

@LizBaldo ok.  Do we have/want/need a ticket for implementing caching in case we decide we need it someday?

LizBaldo:
Oh for sure, I am just trying to even get this to work first, but I will make a follow up ticket to design how we want to cache these charts moving forward.

LizBaldo:
jenkins retest

LizBaldo:
jenkins retest

LizBaldo:
jenkins retest

LizBaldo:
FYI @rtitle and @calypsomatic, I do not think that our current GKE charts can be directly installed from remote as they need to access the local tls secrets. 
This is from [the helm documentation](https://helm.sh/docs/chart_template_guide/accessing_files/):
```There is no way to pass files external to the chart during helm install. So if you are asking users to supply data, it must be loaded using helm install -f or helm install --set.```

So in the end, the scope of this PR is only to install the cromwell and wds apps from remote directly. Is this worth merging in your opinion? 
I will make other tickets in the epic to investigate what is really going on with the GKE charts, but we have the potential to at least unblock the AJ and Workflow teams AKS charts, CC @cjllanwarne and @jgainerdewar.

LizBaldo:
jenkins retest

LizBaldo:
Automation tests actually passed a couple of times so force merging this

PR #4014, opened 2023-12-07
AJ-1436 Create azure_e2e_release_promotion_tests.yml

calypsomatic:
https://broadworkbench.atlassian.net/browse/AJ-1436

Adds a GHA that will run the azure e2e tests upon promotion to staging.  Once this PR merges, a [deploy hook](https://beehive.dsp-devops.broadinstitute.org/environments/staging/chart-releases/leonardo/deploy-hooks/new) will be added to call it.

### What

-

### Why

-

### What to test

- 

### Who tested and where

- [ ] I validated this change 


.github/workflows/azure_e2e_release_promotion_tests.yml
```diff
@@ -0,0 +1,190 @@
+name: Azure E2E Release Promotion Tests
+'on':
+  push:
+    branches:
+      - aj-1436-e2e-staging
```
calypsomatic: This is temporary for testing

.github/workflows/azure_e2e_release_promotion_tests.yml
```diff
@@ -0,0 +1,190 @@
...
+          access_token_scopes: 'profile, email, openid'
+          access_token_subject: 'hermione.owner@quality.firecloud.org'
+          export_environment_variables: false
+
+      - name: Call Sam to create pet
+        run: |
+          SAM_URL="https://sam.${{ env.BEE_NAME }}.bee.envs-terra.bio"
+          API_URL="${SAM_URL}/api/azure/v1/user/petManagedIdentity"
+          AZURE_TOKEN="${{ steps.obtain_token.outputs.access_token }}"
+          PAYLOAD='{
```
calypsomatic: These are the default values that I see used in the create bee/attach bp workflows.  I put them in directly here just to get this step working; should they be environment variables at the beginning of the yaml file?  Should there be an option to pass them in and then pass them along to the other workflows?

davidangb: In an ideal world, the tenantId, subscriptionId, and managedResourceGroupName would be outputs from the `attach-landing-zone-to-bee-workflow` job, so they match exactly the LZ that was actually used.

If that's a pain in the butt, could you make them inputs to this workflow, a la https://github.com/broadinstitute/terra-github-workflows/blob/31b20ac77ecae4eb8602db78287e5c5214449293/.github/workflows/attach-billing-project-to-landing-zone.yaml#L13-L27? If they're inputs to this workflow, you could even pass them over to `attach-landing-zone-to-bee-workflow` to ensure they're consistent.

davidangb (commented):
lgtm, I did not test myself. One comment inline about hardcoding the tenant/subscription/mrg.

.github/workflows/azure_e2e_release_promotion_tests.yml
```diff
@@ -0,0 +1,190 @@
...
+        uses: broadinstitute/workflow-dispatch@v4.0.0
+        with:
+          run-name: '${{ env.BEE_CREATE_RUN_NAME }}'
+          workflow: bee-create
+          repo: broadinstitute/terra-github-workflows
+          ref: refs/heads/main
+          token: '${{ env.TOKEN }}'
+          inputs: >-
+            { "run-name": "${{ env.BEE_CREATE_RUN_NAME }}", "bee-name": "${{
+            env.BEE_NAME }}", "bee-template-name": "rawls-e2e-azure-tests",
```
yuliadub, line 80: I know this has come up in another PR as a question I didnt know the answer too and I think you were OOO then - but whats the reason behind us using the "rawls" template name?

.github/workflows/azure_e2e_release_promotion_tests.yml
```diff
@@ -0,0 +1,190 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      id-token: write
+    steps:
+      - name: Echo Leo version
+        run: >
+          echo '${{ needs.leo-build-tag-publish-job.outputs.custom-version-json
+          }}'
+      - name: dispatch to terra-github-workflows
```
yuliadub, line 70: in WDS GAH for e2e tests, we have [retries ](https://github.com/DataBiosphere/terra-workspace-data-service/blob/main/.github/workflows/run-e2e-tests.yaml#L52)for BEE creation and attachment of billing project, I think it may be worth it to do this here as well. 

this gets me wondering if it would be better in the future to define the bee creation/lz attachment as actions in the repo where e2e tests live and then have WDS and Leo check those out. That way in theory we dont need to edit lots of different places if those actions ever change. (this would be a follow up ticket, dont think its urgent to do now)

yuliadub: but then again, both create bee and attach landing zone have another action that controls what they actually do....so not sure. maybe retries should happen in those actions that dev ops owns vs us implementing them?

calypsomatic:
presumably unrelated test failures.  jenkins retest

.github/workflows/azure_e2e_release_promotion_tests.yml
calypsomatic: Mostly copy-paste!  It's what all the e2e tests being called are using now so I figured I'd stick with that for this PR.  If any of the underlying tests change the template then we'll have to figure something out

.github/workflows/azure_e2e_release_promotion_tests.yml
calypsomatic: Yeah I think in theory there are plans to make create bee and attach landing zone more robust in general, making the retries a bit less necessary.

yuliadub:
another question - does this action have to live in Leo?

calypsomatic:
> another question - does this action have to live in Leo?

@yuliadub I think so - it needs to run on promotion to staging (that's the whole point), and for the time being at least, leo's promotion is what makes a difference to when the apps are released.

.github/workflows/azure_e2e_release_promotion_tests.yml
salonishah11: I think the rawls template was added so that Janitor can monitor resources used in these run and be able to clean them up (in the case workspace deletion fails)

.github/workflows/azure_e2e_release_promotion_tests.yml
```diff
@@ -0,0 +1,205 @@
...
+      subscription-id:
+        description: 'Azure subscription ID. This is a shared identifier used by all Apps deployed for Azure End-to-End (E2E) tests. By default, it is set to the current Azure subscription ID.'
+        required: false
+        type: string
+        default: 'f557c728-871d-408c-a28b-eb6b2141a087'
+      mrg-id:
+        description: 'Azure Managed Resource Group name. This is a shared identifier used by all Apps deployed for Azure End-to-End (E2E) tests. By default, it is set to the current Azure MRG name.'
+        required: false
+        type: string
+        default: 'e2e-8n6xqg'
```
salonishah11, line 23: Are these used anywhere? 🤔

calypsomatic: They were/will be!  I temporarily removed them the places they're used just to see if I could get some data on whether or not I broke things with [this commit](https://github.com/DataBiosphere/leonardo/pull/4014/commits/b60b450a65d1ef5c61ec0d2ef268f9139802ed4d).  Planning to un-revert those changes

yuliadub:
> > another question - does this action have to live in Leo?
> 
> @yuliadub I think so - it needs to run on promotion to staging (that's the whole point), and for the time being at least, leo's promotion is what makes a difference to when the apps are released.

I see -- so it is not desirable to say for us to run this twice a day on staging like we do with dev today? I am guessing the idea is to run as soon as bits are released to catch any possible errors asap. 

I wonder if we could configure the github action that lives in WDS repo to be configurable enough so that leo can just kick it off vs having another very similar action live here and now having to maintain both (long term thoughts) Although this action include cromwell e2e test as well, so maybe it wont quite work then.

.github/workflows/azure_e2e_release_promotion_tests.yml
```diff
@@ -0,0 +1,186 @@
...
+      contents: read
+      id-token: write
+    uses: >-
+      broadinstitute/dsp-reusable-workflows/.github/workflows/cromwell-az-e2e-test.yaml@main
+    with:
+      bee-name: '${{ needs.params-gen.outputs.bee-name }}'
+      billing-project-name: '${{ needs.params-gen.outputs.project-name }}'
+  sleep-job-2:
+    # One test needs to run before the others and have time to do one-time tasks
+    # Such as creating a k8s cluster DB entry in Leo to prevent race conditions
```
jgainerdewar: Kind of a nitpick, but can you update this comment to reflect the fact that this is a second, separate sleep step?

jgainerdewar (approved):

.github/workflows/azure_e2e_release_promotion_tests.yml
calypsomatic: Ah good point, sure.

calypsomatic:
> I see -- so it is not desirable to say for us to run this twice a day on staging like we do with dev today? I am guessing the idea is to run as soon as bits are released to catch any possible errors asap.
> 
> I wonder if we could configure the github action that lives in WDS repo to be configurable enough so that leo can just kick it off vs having another very similar action live here and now having to maintain both (long term thoughts) Although this action include cromwell e2e test as well, so maybe it wont quite work then.

@yuliadub Once apps are independent of Leo, we can probably get rid of this action altogether and kick off the tests directly upon app version release into staging, but in the meantime we'll have to have these both.

calypsomatic:
jenkins retest

davidangb (approved):
thank you!!

yuliadub (approved):
thank you!

calypsomatic:
jenkin retest

calypsomatic:
jenkins retest

PR #4020, opened 2023-12-07
RW-9480 add workspaceId in v1 createApp API

Qi77Qi:
This PR's scope is fairly small, which is to add an optional `workspaceId` field in createApp request. This will be good for a few different reasons as we're migrating to `workspaceId` centric APIs.

The direct motivation is [this doc](https://docs.google.com/document/d/13cfxxjTGXdn48m6kXSN5AL-D9usk5yFmBQ0nQaV0L5Y/edit)...This change is part of the current winning option. I'd like to talk with the team as well to go over options, but this change should be beneficial even if we end up taking a different option.

-

### What to test

- 

### Who tested and where

- [ ] I validated this change 

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoApiClient.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoApiClient.scala
@@ -125,6 +125,7 @@ object LeonardoApiClient {
     Map.empty,
     None,
     List.empty,
+    None,
     None
   )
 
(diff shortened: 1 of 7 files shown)


jdcanas (approved):
makes sense to me

LizBaldo (approved):
LGTM, and makes sense as the first step of the design doc shared :) Would this warrant a small unit test to make sure the leo front behaves as expected with a workspaceid provided?

Qi77Qi:
@LizBaldo I can add a unit test for the decoder if that's what you're asking

PR #4034, opened 2023-12-08
RW-10951 cgroup memory limit includes Spark Driver Memory

danking:
__Jira ticket__: https://precisionmedicineinitiative.atlassian.net/browse/RW-10951

## Summary of changes

This resolves a bug related to Hail. It previously manifested as "py4j.protocol.Py4JNetworkError: Answer from Java side is empty" but, due to changes in Hail 0.2.126, it now manifests as "RemoteDisconnectedError". This occurs when a Spark Driver JVM killed by the OOMKiller. When this happens, the Hail log contains no indication of a problem and there is no `hs_err_pid*.log` file.

#### How Hail Works

Hail comprises three key components:

1. Python client.
2. Spark Driver JVM.
3. Spark Worker JVMs.

When Hail initializes, it uses PySpark to create a Spark Driver JVM. Unless the [--deploy-mode](https://spark.apache.org/docs/latest/submitting-applications.html#launching-applications-with-spark-submit) of `spark-submit` is overridden (this can be done with [PYSPARK_SUBMIT_ARGS](https://github.com/apache/spark/blob/d9f0fccd967b5c8686353d524d2b31e27b7a473b/python/pyspark/java_gateway.py#L70)), this Spark Driver JVM is a sub-process of the Python client process.

Suppose we execute this Python code:

```python3
mt = hl.read_matrix_table("gs://bucket/foo.mt")
mt.count()
```

Hail constructs an abstract representation of this:

```
!1 = MatrixRead [Matrix{ ... }, ..., (MatrixNativeReader gs://bucket/foo.mt )]
MatrixCount(!1)
```

And sends that over a local HTTP/TCP socket to the Spark Driver JVM. The Spark Driver JVM determines the number of partitions and processes each partition on the Spark Worker JVMs. The results are either written to cloud storage or serialized back to the Python process over the socket.

#### The Bugs

There were two related bugs in how Leonardo sets the memory of Hail:

1. Leonardo (and, indeed, the current version of `hailctl`), for n1-highmem-8, sets the JVM maximum heap size to a number too close to the practically available memory.

2. Leonardo, for all machine types, sets (via docker, which uses cgroupv1) a memory limit on the Jupyter Notebook server's cgroup. Since the Notebook kernel and the Spark Driver JVM are sub-processes of the Notebook server, they are all collectively subjected to the memory limit.

Bug 1 is described in detail in this [comment](https://github.com/hail-is/hail/issues/13960#issuecomment-1836844790) on Hail issue https://github.com/hail-is/hail/issues/13960 . The gist of the issue is that the amount of RAM available used by system daemons on the Dataproc Driver VM has grown substantially since Hail introduced the "80% of total RAM" rule. It has already been resolved by https://github.com/danking/leonardo/commit/ef0879015a4dd16e5566b190c915328ce030d5fb .

Bug 2 is superficially similar to Bug 1. In both cases, the user sees RemoteDisconnectedError because the OOMKiller kills the Spark Driver JVM. However, in Bug 2, the OOMKiller is invoked because the *cgroup exceeded its memory limit* whereas in Bug 1 the OOMKiller is invoked because the *the entire system was nearly out of memory*.

Let's look at the syslog of an example of Bug 2. In this case, the user is using an n1-standard-4 with four cores and 15000 MiB of RAM [1]. At 2023-12-08 2045, there is 4280 MiB of memory available to the system; however, the cgroup `/docker/8be415e240b32e8545043ce136fc17a1206d847f24ed507646339b8798e50316` has exceeded its memory limit. The syslog shows the two Python processes and the Java process and the resident set size (rss), in pages. This trio of processes are using more than 4GB of memory. A moment later the OOMKiller kills the java process.

```
Dec  8 20:45:17 all-of-us-263-m earlyoom[440]: mem avail:  4280 of 15000 MiB (28.54%), swap free:    0 of    0 MiB ( 0.00%)
...
Dec  8 20:45:17 all-of-us-263-m kernel: [ 1166.445565] Memory cgroup stats for /docker/8be415e240b32e8545043ce136fc17a1206d847f24ed507646339b8798e50316:
...
Dec  8 20:45:17 all-of-us-263-m kernel: [ 1166.588599] [  pid  ]   uid  tgid total_vm      rss pgtables_bytes swapents oom_score_adj name
...
Dec  8 20:45:18 all-of-us-263-m kernel: [ 1166.637384] [   3240]  1000  3240    72396    26654   339968        0             0 python3
Dec  8 20:45:18 all-of-us-263-m kernel: [ 1166.647309] [   4378]  1000  4378   440607    63412  1114112        0             0 python3
Dec  8 20:45:18 all-of-us-263-m kernel: [ 1166.657204] [   5942]  1000  5942  2904213   952126  8302592        0             0 java
...
Dec  8 20:45:18 all-of-us-263-m kernel: [ 1166.718428] Memory cgroup out of memory: Killed process 5942 (java) total-vm:11616852kB, anon-rss:3808504kB, file-rss:0kB, shmem-rss:0kB, UID:1000 pgtables:8108kB oom_score_adj:0
```

#### The Fix

The fix is to set the memory limit to the sum of the memory we expect Python and Java to use.

### Who tested and where

- [ ] I validated this change 

[1] 15,000 MiB is lower than the advertised memory for this machine type. This machine type should have 15 GiB = 15,360 MiB of memory.

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreter.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreter.scala
@@ -274,10 +274,19 @@ class DataprocInterpreter[F[_]: Parallel](
               .getOrElse((None, None))
           } else (None, None)
 
+        sparkDriverMemory <- jupyterResourceConstraints.driverMemory match {
+          case Some(value) => F.pure(value)
+          case None =>
+            F.raiseError[MemorySize](
+              new RuntimeException(
+                s"Spark driver memory must be specified for DataprocInterpreter. This should never happen"
+              )
+            )
+        }
         softwareConfig = getSoftwareConfig(googleProject,
                                            params.runtimeProjectAndName.runtimeName,
                                            machineConfig,
-                                           jupyterResourceConstraints
+                                           sparkDriverMemory
         )
 
         // Enables Dataproc Component Gateway. Used for enabling cluster web UIs.
@@ -779,21 +788,30 @@ class DataprocInterpreter[F[_]: Parallel](
       )
       _ <- logger.debug(ctx.loggingCtx)(s"Resolved machine type: ${resolvedMachineType.toString}")
       total = MemorySize.fromMb(resolvedMachineType.getMemoryMb.toDouble)
+
+      sparkDriverMemory <- machineType match {
+        case MachineTypeName(n1standard) if n1standard.startsWith("n1-standard") =>
+          F.pure(MemorySize.fromGb((total.bytes / MemorySize.gbInBytes - 7) * 0.9))
+        case MachineTypeName(n1highmem) if n1highmem.startsWith("n1-highmem") =>
+          F.pure(MemorySize.fromGb((total.bytes / MemorySize.gbInBytes - 11) * 0.9))
+        case x =>
+          F.raiseError(
+            new RuntimeException(
+              s"Machine type (${x.value}) not supported by Hail. Consider use `n1-highmem` machine type"
+            )
+          )
+      }
     } yield {
       // For dataproc, we don't need much memory for Jupyter.
 
       // We still want a minimum to run Jupyter and other system processes.
       val minRuntimeMemoryGb = MemorySize.fromGb(config.dataprocConfig.minimumRuntimeMemoryInGb.getOrElse(4.0))
       // Note this algorithm is recommended by Hail team. See more info in https://broadworkbench.atlassian.net/browse/IA-4720
-      val sparkDriverMemory = machineType match {
-        case MachineTypeName(n1standard) if n1standard.startsWith("n1-standard") =>
-          Some(MemorySize.fromGb((total.bytes / MemorySize.gbInBytes - 7) * 0.9))
-        case MachineTypeName(n1highmem) if n1highmem.startsWith("n1-highmem") =>
-          Some(MemorySize.fromGb((total.bytes / MemorySize.gbInBytes - 11) * 0.9))
-        case _ => none[MemorySize]
-      }
-      val runtimeAllocatedMemory = MemorySize(sparkDriverMemory.map(_.bytes).getOrElse(0) + minRuntimeMemoryGb.bytes)
-      RuntimeResourceConstraints(runtimeAllocatedMemory, MemorySize(total.bytes), sparkDriverMemory)
+
+      val runtimeAllocatedMemory = MemorySize(
+        sparkDriverMemory.bytes + minRuntimeMemoryGb.bytes
+      )
+      RuntimeResourceConstraints(runtimeAllocatedMemory, MemorySize(total.bytes), Some(sparkDriverMemory))
     }
 
   /**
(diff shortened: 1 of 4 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/DataprocInterpreter.scala
```diff
@@ -779,28 +788,30 @@ class DataprocInterpreter[F[_]: Parallel](
...
+
+      sparkDriverMemory <- machineType match {
+        case MachineTypeName(n1standard) if n1standard.startsWith("n1-standard") =>
+          F.pure(MemorySize.fromGb((total.bytes / MemorySize.gbInBytes - 7) * 0.9))
+        case MachineTypeName(n1highmem) if n1highmem.startsWith("n1-highmem") =>
+          F.pure(MemorySize.fromGb((total.bytes / MemorySize.gbInBytes - 11) * 0.9))
+        case x =>
+          F.raiseError(
+            new RuntimeException(
+              s"Machine type (${x.value}) not supported by Hail. Consider use `n1-highmem` machine type"
```
Qi77Qi, line 800: Note, we're disallowing non `n1-standard` and `n1-highmem` machine types in this PR. The only other machine type family we allow in UI is `n1-highcpu`, which isn't going to work well for Dataproc anyway, hence nudge users to use `highmem` instead

LizBaldo: That makes sense to me, the error message is clear

LizBaldo (approved):
LGTM, restricting the machine type to what is actually supported simplifies the logic quite a bit 👍

cpate4 (commented):
Assuming this is the Linux OOM Killer process ... 
and the solution appears to be:
- reserve space for the kernel, and
- only allocate X% of the remaining space based on the machine size

Is there anything we need to do to ensure the cgroup has access to that much memory? e.g. Is k8s at all involved in this setup and do we need to tweak any pod memory settings?

Qi77Qi:
@cpate4 there's no k8s involved. This is for supporting users who wants to use Hail (dataproc clusters are provisioned in this case)

[this](https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/resources/init-resources/jupyter-docker-compose.yaml#L60) is where the memlimit is used in the end

mlilynolting:
Looks like there are some sbt format/build issues; otherwise looks good to me!

Qi77Qi (approved):

Qi77Qi:
CI from a fork is weird...we can keep this PR since it has lots of context..created [this](https://github.com/DataBiosphere/leonardo/pull/4036) just to trigger [jenkins run](https://fc-jenkins.dsp-techops.broadinstitute.org/job/swatomation-pipeline-leo/12823/)

Qi77Qi:
one PD test failed, which is unrelated to the change itself...

@cindy-broadinstitute  r u still reviewing the PR or you're ok with merging?

cindy-broadinstitute (approved):
LGTM

PR #3614, opened 2023-08-21
AJ-1270: look for pgbouncer-enabled tag on LZ Postgres server

davidangb:
__Jira ticket__: https://broadworkbench.atlassian.net/browse/AJ-1270

## Dependencies
enabled by https://github.com/DataBiosphere/terra-workspace-manager/pull/1443
will be consumed by https://github.com/broadinstitute/terra-helmfile/pull/4491

## Summary of changes

When creating a WDS/Cromwell app in Azure, look for a "pgbouncer-enabled" tag on the Landing Zone Postgres server. Send the boolean value of that tag, or false if the tag is missing, to the WDS/Cromwell app as a helm value.

### What

As well as the change in the summary above:
* rename `LandingZoneResources. postgresName` to `LandingZoneResources. postgresServer`
* change `PostgresName(value: String)` to `PostgresServer(name: String, pgBouncerEnabled: Boolean)`

These two changes cascaded a fair bit of syntax-compatibility changes elsewhere in the code; there's a lot of boilerplate change in this PR.

### Why

See https://docs.google.com/document/d/1_fRDJJn8Fgzg_MNzm5zKskRDcfYB5M-Xl6DoGEkhdho/edit for full context.

We are enabling PgBouncer in our LZs moving forward. This will allow many more apps to connect to the LZ Postgres without running out of Postgres connections. This Leo change is necessary to continue to support old LZs - which do not have PgBoucner - as well as the new LZs.

### What to test

- 

### Who tested and where

- [x] This change is covered by automated tests 
  - _NB: Rerun automation tests on this PR by commenting `jenkins retest` or `jenkins multi-test`._
- [x] I validated this change _in my bee. I manually changed the tags on the PG server and inspected the helm values sent to new apps; I see the helm values change appropriately._
- [ ] Primary reviewer validated this change 
- [ ] I validated this change in the dev environment

PR #4037, opened 2023-12-11
[AJ-1511] Pass AKS cluster tags to WDS

ashanhol:
__Jira ticket__: https://broadworkbench.atlassian.net/browse/AJ-1511

### What
This PR pipes Azure tags on the LZ AKS cluster to WDS's helm chart. Corresponding helm chart configuration [here](https://github.com/broadinstitute/terra-helmfile/pull/4874). 
This required changing the AKSClusterName argument in LandingZoneResources to a newly defined `AKSCluster`, which includes both a name and a map of relevant tags (in this case aks-cost-vpa-enabled)
### Why
This is a necessary step to pass the cost saving flag for VPA from Landing Zone to WDS on startup. 
Note: passing tag modeled after this PR https://github.com/DataBiosphere/leonardo/pull/3614

### What to test

- 

### Who tested and where

- [ ] I validated this change 

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodec.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodec.scala
@@ -687,6 +687,7 @@ object JsonCodec {
   implicit val allowedChartNameDecoder: Decoder[AllowedChartName] =
     Decoder.decodeString.emap(x => AllowedChartName.stringToObject.get(x).toRight("chart name not allowed"))
   implicit val aksClusterNameDecoder: Decoder[AKSClusterName] = Decoder.decodeString.map(AKSClusterName)
+  implicit val aksClusterDecoder: Decoder[AKSCluster] = Decoder.forProduct2("name", "tags")(AKSCluster)
   implicit val postgresServerDecoder: Decoder[PostgresServer] =
     Decoder.forProduct2("name", "pgBouncerEnabled")(PostgresServer)
 
@@ -759,6 +760,8 @@ object JsonCodec {
   implicit val azureDiskNameEncoder: Encoder[AzureDiskName] = Encoder.encodeString.contramap(_.value)
   implicit val relayNamespaceEncoder: Encoder[RelayNamespace] = Encoder.encodeString.contramap(_.value)
   implicit val aksClusterNameEncoder: Encoder[AKSClusterName] = Encoder.encodeString.contramap(_.value)
+  implicit val aksClusterEncoder: Encoder[AKSCluster] = 
+    Encoder.forProduct2("name", "tags")(x => (x.name, x.tags))
   implicit val postgresServerEncoder: Encoder[PostgresServer] =
     Encoder.forProduct2("name", "pgBouncerEnabled")(x => (x.name, x.pgBouncerEnabled))
 
@@ -800,7 +803,7 @@ object JsonCodec {
     "postgresName"
   )(x =>
     (x.landingZoneId,
-     x.clusterName,
+     x.aksCluster,
      x.batchAccountName,
      x.relayNamespace,
      x.storageAccountName,
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodec.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodec.scala
@@ -760,7 +760,7 @@ object JsonCodec {
   implicit val azureDiskNameEncoder: Encoder[AzureDiskName] = Encoder.encodeString.contramap(_.value)
   implicit val relayNamespaceEncoder: Encoder[RelayNamespace] = Encoder.encodeString.contramap(_.value)
   implicit val aksClusterNameEncoder: Encoder[AKSClusterName] = Encoder.encodeString.contramap(_.value)
-  implicit val aksClusterEncoder: Encoder[AKSCluster] = 
+  implicit val aksClusterEncoder: Encoder[AKSCluster] =
     Encoder.forProduct2("name", "tags")(x => (x.name, x.tags))
   implicit val postgresServerEncoder: Encoder[PostgresServer] =
     Encoder.forProduct2("name", "pgBouncerEnabled")(x => (x.name, x.pgBouncerEnabled))
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/app/WdsAppInstall.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/app/WdsAppInstall.scala
@@ -61,7 +61,7 @@ class WdsAppInstall[F[_]](config: WdsAppConfig,
       )
 
       // Get Vpa enabled tag 
-      vpaEnabled <- params.landingZoneResources.aksCluster.tags.getOrElse("aks-cost-vpa-enabled", false)
+      vpaEnabled <- java.lang.Boolean.parseBoolean(params.landingZoneResources.aksCluster.tags.getOrElse("aks-cost-vpa-enabled", "false"))
 
       valuesList =
         List(
@@ -98,7 +98,7 @@ class WdsAppInstall[F[_]](config: WdsAppConfig,
           // general configs
           raw"fullnameOverride=wds-${params.app.release.asString}",
           raw"instrumentationEnabled=${config.instrumentationEnabled}",
-          raw"vpaEnabled"=${vpaEnabled},
+          raw"vpaEnabled=${vpaEnabled}",
 
           // import configs
           raw"import.dataRepoUrl=${tdrConfig.url}",
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/app/WdsAppInstall.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/app/WdsAppInstall.scala
@@ -61,7 +61,7 @@ class WdsAppInstall[F[_]](config: WdsAppConfig,
       )
 
       // Get Vpa enabled tag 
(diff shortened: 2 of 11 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -644,14 +644,14 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
       ).interruptAfter(config.appMonitorConfig.updateApp.interruptAfter).compile.lastOrError
     } yield appOk.isDone
 
-  private[util] def getHelmAuthContext(clusterName: AKSClusterName,
+  private[util] def getHelmAuthContext(clusterName: String,
```
davidangb: nit: I suggest leaving this as `AKSClusterName` for type-safety and readability. I realize this means you'll need to construct `AKSClusterName` in a few places to satisfy the argument. You could add a convenience method to `AKSCluster` if it's helpful to do the translation, something like:
```scala
final case class AKSCluster(name: String, tags: Map[String, Boolean]) {
  def asClusterName: AKSClusterName = AKSClusterName(name)
}
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/app/WdsAppInstall.scala
```diff
@@ -95,6 +98,7 @@ class WdsAppInstall[F[_]](config: WdsAppConfig,
           // general configs
           raw"fullnameOverride=wds-${params.app.release.asString}",
           raw"instrumentationEnabled=${config.instrumentationEnabled}",
+          raw"vpaEnabled=${vpaEnabled}",
```
davidangb: nit: instead of a top-level "vpaEnabled" value, this should probably go up at line ~74 with the other "azure resources configs" and get renamed to "config.vpaEnabled" or better yet "config.aks.vpaEnabled"

davidangb (commented):
A couple cleanup comments inline, nice work on this, it is very close! There are a bunch of unit test failures that are real - they assert on the output of `buildHelmOverrideValues`, which you have changed.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/app/WdsAppInstall.scala
ashanhol: Thanks, I wasn't sure where to put it and wasn't sure whether to start a whole aks category

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpWsmDao.scala
```diff
@@ -153,12 +148,19 @@ class HttpWsmDao[F[_]](httpClient: Client[F], config: HttpWsmDaoConfig)(implicit
...
-                                                   false
+      aksResource <- getLandingZoneResource(groupedLzResources,
+                                            "Microsoft.ContainerService/managedClusters",
+                                            SHARED_RESOURCE
       )
+      aksCluster <- getLandingZoneResourceName(aksResource, useParent = false).map { aksName =>
+        val tagValue = getLandingZoneResourceTagValue(aksResource, "aks-cost-vpa-enabled")
+        val tags: Map[String, Boolean] =
+          Map("aks-cost-vpa-enabled" -> java.lang.Boolean.parseBoolean(tagValue.getOrElse("false")))
+        logger.info(
```
rtitle: Scala comment -- this log statement won't execute because it's wrapped in an `IO` -- it needs to be chained to other `IO`s. I'd probably do something like this if we want to log the AKS cluster and its tags:
```
aksCluster <- getLandingZoneResourceName(aksResource, useParent = false).map { aksName =>
  val tagValue = getLandingZoneResourceTagValue(aksResource, "aks-cost-vpa-enabled")
  val tags: Map[String, Boolean] =
    Map("aks-cost-vpa-enabled" -> java.lang.Boolean.parseBoolean(tagValue.getOrElse("false")))
  AKSCluster(aksName, tags)
}
_ <- logger.info(s"Retrieved Landing Zone AKS cluster: ${aksCluster}")
```

rtitle (commented):
1 comment on the log statement in HttpWsmDAO -- otherwise LGTM.

davidangb (approved):
modulo Rob's comment about logging, this looks good, and thank you!

I could see adding a test that verifies that if the LZ AKS has a "aks-cost-vpa-enabled"="true" tag, we pass the proper value to the WDS Helm chart. Right now, all the tests in WdsAppInstallSpec only verify "false".

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/app/BaseAppInstallSpec.scala
```diff
@@ -50,7 +51,7 @@ class BaseAppInstallSpec extends AnyFlatSpecLike with LeonardoTestSuite with Moc
 
   val lzResources = LandingZoneResources(
     UUID.fromString("5c12f64b-f4ac-4be1-ae4a-4cace5de807d"),
-    AKSClusterName("cluster"),
+    AKSCluster("cluster", Map.empty[String, Boolean]),
```
jladieu, line 54: Mind explaining the `Map.empty` param?  I don't know Scala well enough to understand what this is for.

jladieu: Nevermind I think I understand it after re-reading.

My next question for the Scala people: might this be an appropriate place to use a default parameter value and just set this to empty automatically, if not provided?  Then you could just declare `AKSCluster("cluster")` throughout?

jladieu (approved):
A couple questions, but seems good to me!   I'm approving with the caveat that I don't know how to check for Scala norms yet, but everything seems pretty reasonable.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpWsmDao.scala
ashanhol: fixed, thank you! I do think its worth logging the tags for easier debugging. :)

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/app/WdsAppInstallSpec.scala
```diff
@@ -38,6 +38,7 @@ class WdsAppInstallSpec extends BaseAppInstallSpec {
       "wds.environmentBase=live," +
       "config.resourceGroup=mrg," +
       "config.applicationInsightsConnectionString=applicationInsightsConnectionString," +
+      "config.aks.vpaEnabled=false," +
```
yuliadub, line 41: wouldn't this always make it be false? i.e. shouldn't we only make it false if there is no tags saying it would be true?

yuliadub: ha wait, this is a test, ignore my comment :)

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodec.scala
```diff
@@ -759,6 +760,8 @@ object JsonCodec {
   implicit val azureDiskNameEncoder: Encoder[AzureDiskName] = Encoder.encodeString.contramap(_.value)
   implicit val relayNamespaceEncoder: Encoder[RelayNamespace] = Encoder.encodeString.contramap(_.value)
   implicit val aksClusterNameEncoder: Encoder[AKSClusterName] = Encoder.encodeString.contramap(_.value)
+  implicit val aksClusterEncoder: Encoder[AKSCluster] =
+    Encoder.forProduct2("name", "tags")(x => (x.name, x.tags))
```
yuliadub, line 764: are we thinking that long term there may be more tags that are cluster specific that we may leverage to change behavior vs just adding a single VPA enabled vs not property?

I do like that cluster info (name and vpa status) is together, but just wondering if having the assumption we are handling more then one tag here may end up being confusing in the future?

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/app/BaseAppInstallSpec.scala
rtitle: You could certainly do that. However -- and this is just my personal preference -- I usually try to limit the use of default parameters. I find it makes the code harder to understand and easier to make mistakes, as opposed to being explicit. But 🤷

rtitle (approved):
👍

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/JsonCodec.scala
rtitle: It's a good point -- no real strong opinions either way, but I could see the general use of tags on AKS to be useful for more things in the future. Things like: different nodepool configurations, private endpoints, different AKS preview features, etc.

ashanhol: Yeah, after talking with David I think it makes more sense to leave room for other tags instead of just a bool since that's now the preferred way of info passing for WSM. I'm not sure what's confusing? tags is a Map of [string, bool] so the tag names will be explicit

yuliadub: confusing as in what are all the possible tag values that can be in there and how  each one of them may be used (since the code that generates the tags lives in a different repo). Maybe a helpful comment and link to where the tags are generated can be all that is needed :)

yuliadub (approved):

ashanhol:
jenkins retest

yuliadub:
jenkins retest

ashanhol:
jenkins retest

PR #4050, opened 2023-12-13
[IA-4639] Set HttpOnly flag

lucymcnatt:
__Jira ticket__: https://broadworkbench.atlassian.net/browse/IA-4639

### What

- Adding httpOnly flag to the leoToken cookie

### Why

> Because Leo is launching apps that allow for arbitrary compute, it is important to protect the token value from being stolen through XSS or other attacks. 

### What to test

- 

### Who tested and where

- [ ] I validated this change 

### DIFF
--- a/http/src/main/resources/init-resources/cluster-site-gce.conf
+++ b/http/src/main/resources/init-resources/cluster-site-gce.conf
@@ -34,7 +34,7 @@
     # Append SameSite=None to cookies set by RStudio. This is required by some browsers because we
     # render RStudio in an iframe. There does not appear to be a way within RStudio to do this, hence
     # doing it in the proxy.
-    Header edit Set-Cookie ^(.*)$ $1;Secure;SameSite=None; HttpOnly "expr=%{REQUEST_URI} =~ m#/proxy/[^/]*/[^/]*/rstudio/.*#"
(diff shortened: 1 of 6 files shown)


sarahgibs (approved):
Looks good. Thanks!

lucymcnatt:
jenkins retest

lucymcnatt:
jenkins retest

cpate4 (commented):
I did a grep in our `leonardo`-codebase for `Secure;SameSite=None` and also came up with:
- cluster-site.conf
- cluster-site-gce.conf

Do either of these need the `HttpOnly` to be appended also?

lucymcnatt:
> Do either of these need the HttpOnly to be appended also?

Since this ask was specifically for the Leo token, I think that's out of the scope (referencing the the comment above those lines)

`Append SameSite=None to cookies set by RStudio. This is required by some browsers because we render RStudio in an iframe. There does not appear to be a way within RStudio to do this, hence doing it in the proxy`

but @sarahgibs do you think we should add the httpOnly flag there as well?

sarahgibs:
> 
> `Append SameSite=None to cookies set by RStudio. This is required by some browsers because we render RStudio in an iframe. There does not appear to be a way within RStudio to do this, hence doing it in the proxy`
> 
> but @sarahgibs do you think we should add the httpOnly flag there as well?

All cookies should be HttpOnly if possible. The concern would be if there is JavaScript in RStudio that uses the cookie. Any cookies that do not have HttpOnly set on them could be stolen with malicious JavaScript.

cpate4 (approved):

PR #4083, opened 2024-01-03
RW-11593 adding -N so that the file isn't duplicated in gce-init script

Qi77Qi:
https://precisionmedicineinitiative.atlassian.net/browse/RW-11593

Notice that we're downloading the same file multiple times if users are reusing the same disk.

Something like
```
# ls
UKB_AOU        gitignore_global    gitignore_global.10	gitignore_global.2  gitignore_global.4	gitignore_global.6  gitignore_global.8	jupyter.log  packages	 workspaces_playground
cromwell.conf  gitignore_global.1  gitignore_global.11	gitignore_global.3  gitignore_global.5	gitignore_global.7  gitignore_global.9	lost+found   workspaces
```

This PR makes it so that the `wget` cmd will download and overwrite the file only if the server has a newer version

### What

-

### Why

-

### What to test

- 

### Who tested and where

- [ ] I validated this change 

### DIFF
--- a/http/src/main/resources/init-resources/gce-init.sh
+++ b/http/src/main/resources/init-resources/gce-init.sh
@@ -478,7 +478,7 @@ if [ ! -z "$JUPYTER_DOCKER_IMAGE" ] ; then
 
   # Copy gitignore into jupyter container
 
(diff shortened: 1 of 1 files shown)


yonghaoy (approved):

LizBaldo (approved):
Makes sense to me :)

Looks like the unit test failure might be handled by Justin's fix here https://github.com/DataBiosphere/leonardo/pull/4085

PR #4084, opened 2024-01-03
[IA-4768] Add semaphore for cluster creation

jdcanas:
This error message is appearing in logs sporadically during app testing

```
 HttpMethod(POST) http://leonardo.bm-aj-2023.bee.envs-terra.bio/api/apps/v2/824e9c20-a0e2-45d4-b2b5-6947a8633aa7/wds-824e9c20-a0e2-45d4-b2b5-6947a8633aa7: 500 Internal Server Error entity: {"source":"leonardo","message":"Duplicate entry 'AZURE-fad90753-2022-4456-9b0a-c7e5b934e408/f557c728-871d-408c-a2' for key 'IDX_KUBERNETES_CLUSTER_UNIQUE_V2'","statusCode":500,"exceptionClass":"class java.sql.SQLIntegrityConstraintViolationException","traceId":null}
```

I am not able to easily test this fix, but the error message is pretty descriptive to a point I think this is worth trying. My logic is as follows:
- java.sql.SQLIntegrityConstraintViolationException is occuring because we enforce uniqueness **in the db itself** on the `KUBERNETES_CLUSTER` table on `AzureCloudContext`, in other words, one cluster per cloud context. Note that this is **not** a request to WSM, this is leo attempting to save a record to the database that violates its own DB constraints
- Luckily, we only save clusters to the database in exactly two places. Furthermore, both places leverage the same function, which attempts to find a cluster in the database and saves one if it doesn't exist
- Therefore, it stands to reason a race condition is likely occurring where two `createApp` threads in quick succession could hit [this line](https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesServiceDbQueries.scala#L130) retrieving the db record, determine none exists, and then both threads continue past that check and attempt to save a cluster on [this line](https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesServiceDbQueries.scala#L145)
    - As an aside, a little confused since `saveOrGetClusterForApp` returns a `DBIO`, and I thought the contract of any flatMapped DBIO was an atomic transaction, but who knows 😅 . Not sure how else we would end up violating the SQL constraint though...

A more optimal approach would introduce a cache on the lines of `CloudContext, Semaphore`, like we do for nodepools, so that not all `createAppV2` requests block on the same semaphore, but I don't want to prematurely optimize before we can see if this actually solves the issue.


mlilynolting:
Would it make sense to try to address this by tweaking the transaction first? I understand semaphores to be a big hammer; it feels worrisome to introduce another concurrency construct here. I agree with you that this shouldn't be happening if database transactionality is being honored. I'd expect to see the two transactions start, read the database, attempt a write, and then when the first one persists, the second one fails on uniqueness, rolling back any other changes and cleanly aborting. It'd be worth exploring why that's not happening.

jdcanas:
> Would it make sense to try to address this by tweaking the transaction first? I understand semaphores to be a big hammer; it feels worrisome to introduce another concurrency construct here. I agree with you that this shouldn't be happening if database transactionality is being honored. I'd expect to see the two transactions start, read the database, attempt a write, and then when the first one persists, the second one fails on uniqueness, rolling back any other changes and cleanly aborting. It'd be worth exploring why that's not happening.

The database transaction is doing its job, no constraint violation is being introduced. The issue is that the e2e tests exercising this functionality call `createApp` in quick succession, and fail due to the race condition (aka leo spits back that error message to the createApp request). The workaround they are discussing is adding an arbitrary sleep in the test. It stands to reason that a user could potentially encounter this as well if they are trying to create apps in quick succession for some reason, and it seems like a leo bug that the endpoint doesn't handle this for you. I agree that the semaphore is a big hammer, but there are ways to reduce any bottlenecking by scoping a semaphore to each azure cloud context if this works. We actually do this for all clusters already for nodepool operations (a cache of `[KubernetesClusterId, Semaphore]`, since there is a kubernetes implementation limitation on concurrent nodepool operations in  cluster.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/Boot.scala
```diff
@@ -355,6 +356,7 @@ object Boot extends IOApp {
   ): Resource[F, AppDependencies[F]] =
     for {
       semaphore <- Resource.eval(Semaphore[F](applicationConfig.concurrency))
+      kubernetesClusterSemaphore <- Resource.eval(Semaphore[F](1))
```
LizBaldo: Had to read up on how Semaphore works a bit but that seems like a good option. One little ask though would be to add a little comment explaining why the concurrency is set to exactly 1, and not configurable in this case?

LizBaldo (approved):

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
```diff
@@ -164,7 +165,10 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
         getSavableCluster(originatingUserEmail, cloudContext, ctx.now)
       )
 
+      _ <- clusterSemaphore.acquire
```
rtitle: I'm not sure this is a good pattern -- if `KubernetesServiceDbQueries.saveOrGetClusterForApp(saveCluster).transaction` fails for whatever reason it may never release the lock. I think a better pattern is `clusterSemaphore.permit.use(…)` which wraps the acquire/release in a cats-effect `Resource`.

Also you might look into using [KeyLock](https://github.com/DataBiosphere/leonardo/blob/develop/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/KeyLock.scala) which already exists in Leo -- it basically does the same thing, but is able to lock on a per-key basis (since we probably don't need a global lock, but a lock per-cloudContext).

Last comment: maybe it's worth trying a `Serializable` transaction to see if it resolves the issue, and not using a semaphore at all? It should be as easy as:
```
saveClusterResult <- KubernetesServiceDbQueries.saveOrGetClusterForApp(saveCluster).transaction(isolationLevel = TransactionIsolation.Serializable)
```

rtitle (changes requested):
Suggested change around the use of Semaphore. Also if we think dirty reads are happening it might be worth trying a `Serializable` isolation level before introducing a new concurrency primitive.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
jdcanas: Ah, this is what I was probably looking for. For some reason, I thought anything in the context of a DBIO in a for comprehension (aka a flatmapped DBIO) was supposed to be atomic when I initially wrote this `saveOrGetClusterForApp` long long ago. I knew there had be a way to attempt to enforce a DBIO to be atomic...

rtitle (approved):
I think this makes sense -- I'm curious whether it resolves the concurrency issue from tests you noted above.

jdcanas:
jenkins retest

jdcanas:
> I think this makes sense -- I'm curious whether it resolves the concurrency issue from tests you noted above.

If it doesn't I can make a unit test that does something like the psuedo-code below to first flush out the issue and then verify its fix, but I'm more inclined to go a low effort route, since the developers are running the tests many times and will collect the data for me. Otherwise a fix like this is unlikely to get prioritized...

```
apiRequest1 = createAppV2()
apiRequest2 = createAppV2()
parSequence(apiRequest1, apiRequest2)
```

jdcanas:
jenkin retest

PR #4087, opened 2024-01-04
IA-4765: Leo use its own identity to resolve the workspace in the upgrade flow

rtitle:
https://broadworkbench.atlassian.net/browse/IA-4765

## Summary of changes

When upgrading apps, don't impersonate the user to resolve the WSM workspace. This allows apps to be upgraded administratively without dependence on the user's registration/TOS status.

### What

- Change Leo to use its own SA to resolve the WSM workspace when upgrading apps, instead of app owner's pet token.

### Why

- We encountered issues upgrading apps in production because some users hadn't accepted the new TOS yet. We need to be able to administratively upgrade apps across all users.

### What to test

- Tested creating and upgrading an app on a BEE.

### Who tested and where

- [ ] I validated this change 

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
@@ -319,24 +319,17 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
       _ <- logger.info(ctx.loggingCtx)(s"Updating app ${params.appName} in workspace ${params.workspaceId}")
 
       app = dbApp.app
-      referenceDatabaseNames = app.appType.databases.collect { case ReferenceDatabase(name) => name }.toSet
 
       // Resolve the workspace in WSM
-      tokenOpt <- samDao.getCachedArbitraryPetAccessToken(app.auditInfo.creator)
+      leoAuth <- samDao.getLeoAuthToken
       workspaceDescOpt <- childSpan("getWorkspace").use { implicit ev =>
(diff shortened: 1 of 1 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -319,24 +319,17 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
       _ <- logger.info(ctx.loggingCtx)(s"Updating app ${params.appName} in workspace ${params.workspaceId}")
 
       app = dbApp.app
-      referenceDatabaseNames = app.appType.databases.collect { case ReferenceDatabase(name) => name }.toSet
```
rtitle, line 322: This was an unused variable

LizBaldo (approved):
Thanks for addressing our comment and reducing the scope to updating apps only :)

Just wanted to triple check that this is not needed for GKE apps because we do not need to make a WSM call to get workspace information correct?

cpate4 (approved):
Is there a unit test we need to update? Or just the fact that our tests still pass is enough?

rtitle:
1. Correct, it's not needed for GKE apps because we are not calling WSM.

2. And yes, I think it is covered by the existing `AKSInterpreterSpec` `update and poll app` tests.

Unit tests _did_ fail due to the docker test, I can wait until https://github.com/DataBiosphere/leonardo/pull/4085 is merged and rebase this.

rtitle:
jenkins retest

rtitle:
jenkins retest

PR #4117, opened 2024-01-17
Add `APP_CHART_VERSION` table.

jladieu:
Create `APP_CHART_VERSION` table per: https://broadworkbench.atlassian.net/wiki/spaces/IA/pages/2906914878/Leonardo+App+Versioning+Management#App-vs-Chart-versions

__Jira ticket__: [IA-4736](https://broadworkbench.atlassian.net/browse/IA-4736)

## Summary of changes:

Adds `APP_CHART_VERSION` which can include desired target helm chart version for a given app.
```
mysql> show create table APP_CHART_VERSION;
+-------------------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Table             | Create Table                                                                                                                                                                                                                                                                                                                                                                                                                                                      |
+-------------------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| APP_CHART_VERSION | CREATE TABLE `APP_CHART_VERSION` (
  `chartName` varchar(254) NOT NULL,
  `chartVersion` varchar(254) NOT NULL,
  `appVersion` varchar(254) DEFAULT NULL,
  `timestamp` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
  PRIMARY KEY (`chartName`),
  CONSTRAINT `FK_APP_CHART_VERSION_APP_NAME` FOREIGN KEY (`chartName`) REFERENCES `APP` (`appName`) ON DELETE NO ACTION ON UPDATE NO ACTION
) ENGINE=InnoDB DEFAULT CHARSET=latin1 |
+-------------------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
1 row in set (0.00 sec)

mysql> describe APP_CHART_VERSION;
+--------------+--------------+------+-----+-------------------+-----------------------------+
| Field        | Type         | Null | Key | Default           | Extra                       |
+--------------+--------------+------+-----+-------------------+-----------------------------+
| chartName    | varchar(254) | NO   | PRI | NULL              |                             |
| chartVersion | varchar(254) | NO   |     | NULL              |                             |
| appVersion   | varchar(254) | YES  |     | NULL              |                             |
| timestamp    | timestamp    | NO   |     | CURRENT_TIMESTAMP | on update CURRENT_TIMESTAMP |
+--------------+--------------+------+-----+-------------------+-----------------------------+
4 rows in set (0.01 sec)

```
## Testing:

Empty `APP_CHART_VERSION` table should be present after liquibase migration.

- [x] verified locally after migration
- [x] verified in bee from this branch
- [x] Primary reviewer validated this change
- [ ] verified in dev after merge

[IA-4736]: https://broadworkbench.atlassian.net/browse/IA-4736?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changelog.xml
+++ b/http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changelog.xml
@@ -1,6 +1,8 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
 <!-- If you modify CLUSTER table, please make sure to notify @dspfieldeng on slack if there's updates to any of the fields in https://github.com/broadinstitute/workbench-analytics/blob/master/dbconfig.py#L136-L154 -->
-<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
     <include file="changesets/20170811_cluster.xml" relativeToChangelogFile="true"/>
     <include file="changesets/20170811_label.xml" relativeToChangelogFile="true"/>
     <include file="changesets/20170811_cluster-unique-key.xml" relativeToChangelogFile="true"/>
@@ -111,4 +113,5 @@
     <include file="changesets/20231002_drop_namespace_table.xml" relativeToChangelogFile="true" />
     <include file="changesets/20231030_add_replica_to_app_table.xml" relativeToChangelogFile="true" />
     <include file="changesets/20231214_update_chart_location.xml" relativeToChangelogFile="true" />
+    <include file="changesets/20240117_create_app_chart_version_table.xml" relativeToChangelogFile="true" />
 </databaseChangeLog>
--- a/http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changelog.xml
+++ b/http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changelog.xml
@@ -1,8 +1,6 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
 <!-- If you modify CLUSTER table, please make sure to notify @dspfieldeng on slack if there's updates to any of the fields in https://github.com/broadinstitute/workbench-analytics/blob/master/dbconfig.py#L136-L154 -->
-<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
-                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
     <include file="changesets/20170811_cluster.xml" relativeToChangelogFile="true"/>
     <include file="changesets/20170811_label.xml" relativeToChangelogFile="true"/>
     <include file="changesets/20170811_cluster-unique-key.xml" relativeToChangelogFile="true"/>
--- a/http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20240117_create_app_chart_version_table.xml
+++ b/http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20240117_create_app_chart_version_table.xml
@@ -1,0 +1,25 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="leonardo" xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+    <changeSet logicalFilePath="leonardo" author="ladieu" id="create_app_chart_version_table">
+        <createTable tableName="APP_CHART_VERSION">
+            <column name="chartName" type="VARCHAR(254)">
+                <constraints primaryKey="true"/>
+            </column>
+            <column name="chartVersion" type="VARCHAR(254)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="appTargetVersion" type="VARCHAR(254)">
+                <constraints nullable="true"/>
+            </column>
+            <column name="timestamp" type="TIMESTAMP">
+                <constraints nullable="false"/>
+            </column>
+        </createTable>
+        <addForeignKeyConstraint baseColumnNames="chartName" baseTableName="APP_CHART_VERSION"
+                                 constraintName="FK_APP_CHART_VERSION_APP_NAME" deferrable="false"
+                                 initiallyDeferred="false" onDelete="NO ACTION" onUpdate="NO ACTION"
+                                 referencedColumnNames="appName" referencedTableName="APP"/>
+    </changeSet>
+</databaseChangeLog>
--- a/http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20240117_create_app_chart_version_table.xml
+++ b/http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20240117_create_app_chart_version_table.xml
@@ -10,7 +10,7 @@
             <column name="chartVersion" type="VARCHAR(254)">
                 <constraints nullable="false"/>
             </column>
-            <column name="appTargetVersion" type="VARCHAR(254)">
+            <column name="appVersion" type="VARCHAR(254)">
                 <constraints nullable="true"/>
             </column>
             <column name="timestamp" type="TIMESTAMP">


http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20240117_create_app_chart_version_table.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="leonardo" xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+    <changeSet logicalFilePath="leonardo" author="ladieu" id="create_app_chart_version_table">
+        <createTable tableName="APP_CHART_VERSION">
```
jladieu, line 6: @LizBaldo Calling out `APP_CHART_VERSION` name slightly different than what was in original design doc, but it sounded from our conversation that this might be a better name anyway.  LMK if you have other ideas.

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changelog.xml
```diff
@@ -1,6 +1,8 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
 <!-- If you modify CLUSTER table, please make sure to notify @dspfieldeng on slack if there's updates to any of the fields in https://github.com/broadinstitute/workbench-analytics/blob/master/dbconfig.py#L136-L154 -->
-<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
```
LizBaldo: I usually do not touch these lines, was there a reason to change them?

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20240117_create_app_chart_version_table.xml
LizBaldo: 🤔 good point, this is less confusing indeed, and better reflects the fact that there is a row per chart, not per app :)

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20240117_create_app_chart_version_table.xml
```diff
@@ -0,0 +1,25 @@
...
+            <column name="chartName" type="VARCHAR(254)">
+                <constraints primaryKey="true"/>
+            </column>
+            <column name="chartVersion" type="VARCHAR(254)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="appTargetVersion" type="VARCHAR(254)">
```
LizBaldo: nit, `appVersion` instead of `appTargetVersion`?

LizBaldo (approved):
LGTM! I only have a few minor comments, if testing locally is giving you trouble you could point your BEE to this PR commit and see if the table pops up there?

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changelog.xml
jladieu: I think the formatter just took over at this point when I saved the file and wrapped them onto newlines.  I thought it looked a little bit easier to read so I didn't fight it.

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20240117_create_app_chart_version_table.xml
jladieu: I had doubts about `appTargetVersion`...thanks for the nudge.  Fixed in d31af9527337378aeecdcb19bc44a3685143de3f

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changelog.xml
jladieu: I put this back in d31af9527337378aeecdcb19bc44a3685143de3f to avoid potentially having dueling workspaces/IDEs.

jladieu:
@LizBaldo thanks for the tip about trying a bee...I was able to confirm that the table was generated in my bee

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20240117_create_app_chart_version_table.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="leonardo" xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+    <changeSet logicalFilePath="leonardo" author="ladieu" id="create_app_chart_version_table">
+        <createTable tableName="APP_CHART_VERSION">
+            <column name="chartName" type="VARCHAR(254)">
+                <constraints primaryKey="true"/>
```
nawatts, line 8: Will this table contain only one version per app? [This section in the design doc](https://broadworkbench.atlassian.net/wiki/spaces/IA/pages/2906914878/Leonardo+App+Versioning+Management#App-vs-Chart-versions) suggests it may contain multiple versions. If so, will `chartName` work as a primary key by itself? Or should `chartVersion` be included in the key?

calypsomatic (approved):

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20240117_create_app_chart_version_table.xml
```diff
@@ -0,0 +1,25 @@
...
+                <constraints nullable="true"/>
+            </column>
+            <column name="timestamp" type="TIMESTAMP">
+                <constraints nullable="false"/>
+            </column>
+        </createTable>
+        <addForeignKeyConstraint baseColumnNames="chartName" baseTableName="APP_CHART_VERSION"
+                                 constraintName="FK_APP_CHART_VERSION_APP_NAME" deferrable="false"
+                                 initiallyDeferred="false" onDelete="NO ACTION" onUpdate="NO ACTION"
+                                 referencedColumnNames="appName" referencedTableName="APP"/>
```
nawatts, line 23: Since the `chartName` column references `APP.appName`, would it be worth also calling this column "appName" instead of "chartName" to avoid having two terms for the same thing?

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20240117_create_app_chart_version_table.xml
jladieu: Thanks for reviewing the design carefully and calling this out.  I think this makes sense.

Fixed in 5eefc30f0a772cdd459746bb0218604049e15d2c

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20240117_create_app_chart_version_table.xml
jladieu: Makes sense, fixed in 5eefc30f0a772cdd459746bb0218604049e15d2c

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20240117_create_app_chart_version_table.xml
LizBaldo: Ah, I see where the confusion comes from, as the table in the doc is for sherlock and not leonardo. Leo's table should only have one chart version per app name only

nawatts (approved):

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20240117_create_app_chart_version_table.xml
LizBaldo: AKA chartName should be able to work as a PK by itself

LizBaldo (approved):
The change to the PK is not necessary as there is only one chart version per chart name in the table, but I guess it does not hurt? I would just use chart name as the PK for simplicity but up to you really it should not hurt

jladieu:
@LizBaldo If I understand @nawatts's suggestion correctly, I think you can't actually store the data depicted in this screenshot without making chart version part of the PK.  Otherwise you'll encounter a unique / primary key violation attempting to store the 2nd row with `appName = "wds"`, unless `chartVersion` is also part of its primary key.
image

nawatts:
Right. The doc sounded like the _only_ difference between Sherlock and Leo tables was the "Environment" column.

LizBaldo:
> Right. The doc sounded like the _only_ difference between Sherlock and Leo tables was the "Environment" column.

Yep, the presence of the environment column is why there are two different rows for wds here, let me rework this for clarity in the design doc

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20240117_create_app_chart_version_table.xml
Qi77Qi: curious why we're referencing on AppName instead of appId...not sure if appName is guranteed to be unique

jladieu:
Based on @LizBaldo's clarification, I'm going to undo the last commit, which will revert the column rename too.

I'm thinking the center of gravity for the table has tipped significantly toward being about the _chart_ so I think it makes sense to also undo the rename.

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20240117_create_app_chart_version_table.xml
jladieu: @Qi77Qi I had a similar question which I discussed with @LizBaldo over VC.  My takeaway was that this is really the name of the _chart_ not the App, and so its relationship with the `APP` model is at most a loose foreign key constraint.  The Chart must be tied to an App with the given name, but it doesn't need to be bound explicitly to a particular `appId`, (which would also unnecesasrily imply this chart was bound to a particular `nodepoolId`, which is part of the PK of `APP`).

Qi77Qi: oh yea...that makes sense 👍

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20240117_create_app_chart_version_table.xml
jladieu: Undid this (see https://github.com/DataBiosphere/leonardo/pull/4117#issuecomment-1900598800)

PR #4118, opened 2024-01-17
[IA-4776] Reduce the flakiness of runtime creations in swatomation tests

LizBaldo:
__Jira ticket__: https://broadworkbench.atlassian.net/browse/IA-4776

### What

- Make automation test creating runtimes retryable, as the VM provided by GCE does not always come with a working file system, causing the start up script to not run -> Did not work
- Move the `NewBillingProjectAndWorkspaceBeforeAndAfterAll` trait one level down to each suite, to increase the number of billing project created, reducing the load in each one -> Did not work
- Updating to the latest COS image -> Cannot hurt
- Remove the 5 minutes sleep before Runtime Fixture as it seemed to be making the tests longer without improving on the pass rate
- Fix the race condition on the PD formatting step in the init script did the trick 🎉 

### Why

- Summary of the bug investigation can be found here https://broadworkbench.atlassian.net/wiki/spaces/IA/pages/2967601179/2024-01-16+Integration+test+failure+mode+Jupyter+Welder+failed+to+start+up+in+10+minutes

### What to test

- 

### Who tested and where

- [ ] I validated this change 


jdcanas (commented):
Hmm,  did you ever determine if its tied to `RuntimeFixtureSpec` or inconclusive? If it is, we should include `NotebookHailSpec` and exclude `NotebookGceCustomizationSpec`. Probably pretty hard to sift through all the data and determine for sure, fine with getting this first pass in and tagging more if needed. 

Will almost definitely start failing the GHA again though since it is taking ~65mins currently, the timeout is 75mins and another try can take ~10mins. Maybe bump the number here to ~85 https://github.com/DataBiosphere/leonardo/pull/4111/files, and then remove the sleep in `RuntimeFixtureSpec`?

LizBaldo:
Sounds good, I will update the timeout. I could not find a pattern where the failures only happened with `RuntimeFixtureSpec` unfortunately which is why I cast a wider net

LizBaldo:
@jdcanas FYI I am trying to disable the runtime deletion when it fails during timeout just so I can provide google support team with an example VM where the failure occurs. If you could have a look and let me know if I am indeed disabling the correct flags, I would appreciate the sanity check :)

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/RuntimeFixtureSpec.scala
```diff
@@ -140,10 +134,10 @@ abstract class RuntimeFixtureSpec
   override def afterAll(): Unit = {
     logger.info("afterAll")
 
-    sys.props.get(googleProjectKey) match {
-      case Some(billingProject) => deleteRonRuntime(GoogleProject(billingProject))
```
jdcanas: yeah, this should be sufficient to prevent runtime cleanup

http/src/main/resources/init-resources/gce-init.sh
```diff
@@ -225,7 +225,13 @@ DISK_DEVICE_ID=${USER_DISK_DEVICE_ID:-sdb}
...
 ## Only format disk is it hasn't already been formatted
 if [ "$IS_GCE_FORMATTED" == "false" ] ; then
-  mkfs.ext4 -m 0 -E lazy_itable_init=0,lazy_journal_init=0,discard /dev/${DISK_DEVICE_ID}
+  # It's likely that the persistent disk was previously mounted on another VM and wasn't properly unmounted
+  # either because the VM was terminated or there is no unmount in the shutdown sequence and occasionally
+  # fs is getting marked as not clean.
+  # Passing -F -F to mkfs.ext4 should force the tool to ignore the state of the partition.
+  # Note that there should be two instances command-line switch (-F -F) to override this check
+
+  mkfs.ext4 -m 0 -E lazy_itable_init=0,lazy_journal_init=0,discard /dev/${DISK_DEVICE_ID} -F -F
```
jdcanas, line 234: 🤯

jdcanas (approved):

PR #4120, opened 2024-01-17
RW-11654 Added infrastructure necessary to support contract tests with runtimes.

evrii:
__Jira ticket__: [RW-11654](https://precisionmedicineinitiative.atlassian.net/browse/RW-11654)

### What

- Added a couple extra states to handle runtime functionality
- Separated states into area of concern

### Why

- With the potential of dozens of states being added, I believe that handling all of the state management in a single file would end up being cumbersome.

### What to test

- 

### Who tested and where

- [ ] I validated this change 


pact4s/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/provider/LeoProvider.scala
```diff
@@ -89,29 +67,7 @@ class LeoProvider extends AnyFlatSpec with BeforeAndAfterAll with PactVerifier {
     )
 
   private val providerStatesHandler: StateManagementFunction = StateManagementFunction {
```
evrii, line 73: Move state management functionality to separate files depending on the area of the app. This file only cares about combining the other state management functions.

ichengchang:
You can update the following code in your PR

```plaintext
# pact4s/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/provider/LeoProvider.scala: 92 in eric/RW-11654
```
to

```
case _                     => consumerVersionSelectors = consumerVersionSelectors.deployedOrReleased.mainBranch.branch("eric/RW-11654", "aou-rwb-api")
```

to test your consumer contract.

pact4s/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/provider/LeoProvider.scala
```diff
@@ -133,7 +89,7 @@ class LeoProvider extends AnyFlatSpec with BeforeAndAfterAll with PactVerifier {
   // 2. For normal Sam PR, verify all consumer pacts in Pact Broker labelled with a deployed environment (alpha, dev, prod, staging).
   consumerBranch match {
     case Some(s) if !s.isBlank => consumerVersionSelectors = consumerVersionSelectors.branch(s, consumerName)
-    case _                     => consumerVersionSelectors = consumerVersionSelectors.deployedOrReleased.mainBranch
+    case _                     => consumerVersionSelectors = consumerVersionSelectors.deployedOrReleased.mainBranch.branch("eric/RW-11654", Option("aou-rwb-api"))
```
ichengchang:
```suggestion
    case _                     => consumerVersionSelectors = consumerVersionSelectors.deployedOrReleased.branch("eric/RW-11654", Option("aou-rwb-api"))
```

pact4s/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/provider/LeoProvider.scala
```diff
@@ -126,24 +84,26 @@ class LeoProvider extends AnyFlatSpec with BeforeAndAfterAll with PactVerifier {
...
+  var consumerVersionSelectors: ConsumerVersionSelectors = ConsumerVersionSelectors().branch("whoops")
   // consumerVersionSelectors = consumerVersionSelectors.mainBranch
   // The following match condition basically says
   // 1. If verification is triggered by consumer pact change, verify only the changed pact.
-  // 2. For normal Sam PR, verify all consumer pacts in Pact Broker labelled with a deployed environment (alpha, dev, prod, staging).
-  consumerBranch match {
-    case Some(s) if !s.isBlank => consumerVersionSelectors = consumerVersionSelectors.branch(s, consumerName)
-    case _                     => consumerVersionSelectors = consumerVersionSelectors.deployedOrReleased.mainBranch
-  }
+  // 2. For normal Leo PR, verify all consumer pacts in Pact Broker labelled with a deployed environment (alpha, dev, prod, staging).
```
evrii, line 95: Do not merge this.

.github/workflows/verify_consumer_contracts.yml
```diff
@@ -172,7 +173,7 @@ jobs:
           echo "git checkout ${{ env.CURRENT_BRANCH }}"
           git checkout ${{ env.CURRENT_BRANCH }}
         fi
-        
+        git checkout "eric/RW-11654"
```
ichengchang: Removed this before merge

.github/workflows/verify_consumer_contracts.yml
```diff
@@ -22,6 +22,7 @@ on:
   push:
     branches:
     - develop
+    - eric/RW-11654
```
ichengchang: remove this before merge

pact4s/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/provider/LeoProvider.scala
ichengchang: Remember to revert before merge

pact4s/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/provider/LeoProvider.scala
```diff
@@ -126,24 +84,26 @@ class LeoProvider extends AnyFlatSpec with BeforeAndAfterAll with PactVerifier {
...
+//  }
 
   val provider: ProviderInfoBuilder =
     ProviderInfoBuilder(
       name = "leonardo",
       PactSource
         .PactBrokerWithSelectors(pactBrokerUrl)
         .withAuth(BasicAuth(pactBrokerUser, pactBrokerPass))
-        .withPendingPacts(true)
-        .withProviderTags(ProviderTags("develop"))
```
ichengchang, line 149: can try this

```
.withPendingPactsEnabled(ProviderTags(providerVer))
```

and

```
.withConsumerVersionSelectors(consumerVersionSelectors)
```

pact4s/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/provider/LeoProvider.scala
```diff
@@ -83,38 +62,16 @@ class LeoProvider extends AnyFlatSpec with BeforeAndAfterAll with PactVerifier {
...
-            ChartName(""),
-            None,
-            Map.empty[String, String],
-            false,
-            None
-          )
-        })
-    case _ =>
-      loggerIO.debug("other state")
+    RuntimeStateManager.handler(mockRuntimeService).orElse(AppStateManager.handler(mockAppService)).orElse { case _ =>
```
jdcanas, line 74: huh... a little confused why there are multiple `orElse` chained here, seems like a bit of a code smell but I'm not super familiar with what this code is doing enough to know for sure. I'd recommend trying to change this to a function that has a match statement if possible, psuedocode below
```
determiningThing match {
case 1 =>
case 2 =>
case _ => loggerIO.debug("Whoops: other state")
}
```

ichengchang: I think chaining partial functions encourages separation of duties and re-usability of code (like a runtime state manager and another state manager for the rest of the Leo app), the effect is going to be the same as using a `case match` block.

jdcanas: the downside is its not super clear why one is taking precedence over another (which is what my main point of confusion is). Don't feel super strong about switching the syntax though since semantically it would be the same, a comment could suffice

jdcanas (approved):
small scala related nit, otherwise looks good

pact4s/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/provider/LeoProvider.scala
ichengchang: If the states are mutually exclusive then I think there is no precedence one way over the other. @evrii would you add a comment to your code for clarity sake?

evrii: Thanks @jdcanas and @ichengchang .

I added a comment for clarification.

@jdcanas I would prefer a case statement here too, but I have not been able to figure out a way to handle that.

```
  // This function composes a large switch statement based on partial functions from
  // multiple state mangers (each with their own state that they handle).
  private val providerStatesHandler: StateManagementFunction = StateManagementFunction {
    RuntimeStateManager.handler(mockRuntimeService).orElse(AppStateManager.handler(mockAppService)).orElse { case _ =>
      loggerIO.debug("Whoops: other state")
    }
  }
```

The way this works is that a given state is passed in (say X in this case).

This function checks RuntimeStateManager and it finds that it has cases for A,B, and C but no X

It then proceeds to AppStateManager where is finds states X, Y, and Z. So it executes the code in the X case.

I could see this working like this, but I haven't been able to get it to work:
```
case (A|B|C):
   Call RuntimeStateManager
      case A:
           ...
      case B:
           ...
      case C:
           ...
case (X|Y|Z):
    Call AppStateManager
      case X:
           ...
      case Y:
           ...
      case Z:
           ...
```

PR #4121, opened 2024-01-18
[IA-4590] When deleting, check WSM status before sending pubsub

lucymcnatt:
__Jira ticket__: https://broadworkbench.atlassian.net/browse/IA-4590

### What

- Check the WSM status in front leo before sending the pubsub message delete an Azure resource

- Also check the status of sub-resources and don’t send delete on that specific resource if not found, error if undeletable
    - Apps:
        - database
        - namespace
        - identity
    - Runtimes
        - disk
        - *Note: can't check storage container yet...no get endpoint in wsm
    - Disks

### Why

- These resources are managed by WSM and the status should be the source of truth rather than Leo
- Fails faster when a resource cannot be deleted
- A 'parent' resource cannot be set as deleted in Leo until all its sub-resources are deleted --> their statuses must be verified as well

### What to test

- 

### Who tested and where

- [ ] I validated this change 


core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/models.scala
```diff
@@ -57,3 +57,25 @@ object StagingBucket {
...
+ * (already deleted in WSM, need to clean up leo resources)
+ */
+case class WsmState(value: Option[String]) {
+  def getValue: String = value.getOrElse("DELETED")
+
+  /** Any in-progress state cannot be deleted:
+CREATING, DELETING, UPDATING
+ */
+  def isDeletable: Boolean = value match {
+    case Some(s) => Set("BROKEN", "READY") contains s
```
cpate4: do we want to move this out into a const?

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/models.scala
```diff
@@ -57,3 +57,25 @@ object StagingBucket {
...
+    case _       => true
+  }
+  def isDeleted: Boolean = value match {
+    case Some(_) => false
```
cpate4: will this always be "false" because "getValue() will always return something?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
```diff
@@ -602,13 +602,25 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
...
         for {
-          _ <- deleteDiskResource(Left(runtime.id), msg.workspaceId, auth)
+          runtimeRecordOpt <- controlledResourceQuery
+            .getWsmRecordForRuntime(msg.runtimeId, WsmResourceType.AzureDisk)
+            .transaction
+          runtimeRecord <- F.fromOption(
+            runtimeRecordOpt,
+            AzureDiskResourceDeletionError(Left(msg.runtimeId),
+                                           msg.workspaceId,
+                                           "No disk resource found for delete azure disk. No-op for wsmDao.deleteDisk."
```
cpate4: thinking about in a production environment ...
this error will get raised. 

some questions I am asking myself:
- who will be handling the exception?
- if no one is catching it to handle the exception (it will be breaking out of the various processing and downstream execution of this method) - what is the intended result of this error?

which leads me to be more curious about:
- if we are really in a situation where (for some reason) we are being asked to delete a disk and the disk does not exist, should we instead clean up our own state and log (or metrics?)?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
```diff
@@ -762,7 +774,19 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
...
+            runtimeRecordOpt <- controlledResourceQuery
+              .getWsmRecordForRuntime(e.runtimeId, WsmResourceType.AzureDisk)
+              .transaction
+            runtimeRecord <- F.fromOption(
+              runtimeRecordOpt,
+              AzureDiskResourceDeletionError(
+                Left(e.runtimeId),
+                e.workspaceId,
+                "No disk resource found for delete azure disk. No-op for wsmDao.deleteDisk."
+              )
```
cpate4: ... same question here re: raising an error (which seems to be a No-op)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
```diff
@@ -861,60 +885,34 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
...
-    } yield ()
+          ),
+          auth
+        )
+        .void
+        .adaptError(e =>
+          AzureDiskDeletionError(
+            wsmResourceId,
+            workspaceId,
+            s"${ctx.traceId.asString} | WSM call to delete disk failed due to ${e.getMessage}. Please retry delete again"
```
cpate4, line 934: it looks like here the errors being thrown (if I am following this correctly) are raising an error, _but_ the ask is that we retry the delete (even though the resources don't exist). am I missing something else going on here? is this exception's usage being overloaded?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
```diff
@@ -923,69 +921,70 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
...
+                  } yield ()
+                case WsmJobStatus.Failed =>
+                  F.raiseError[Unit](
+                    DiskDeletionError(
+                      msg.diskId,
+                      msg.workspaceId,
+                      s"WSM delete disk job failed due to ${resp.errorReport.map(_.message).getOrElse("unknown")}"
+                    )
+                  )
+                case WsmJobStatus.Running =>
```
cpate4: this is interesting ... if we are "running", we consider the job status to be failed? 🤔

cpate4 (changes requested):
I have some questions about how we are using exceptions and if:
- that is our style, or
- if there is a better (more user friendly) way of handling this

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/models.scala
lucymcnatt: when a resource is deleted WSM has no record of it anymore, it gives a `None` response. I added this to translate that into a callable function, so when there is no record (value is `None`), it is deleted

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
lucymcnatt: > but the ask is that we retry the delete (even though the resources don't exist)

I'm not sure I fully understand this comment, this is when the deletion errors in WSM for some reason and it happens very rarely, the only time I've seen it recently was a `Received duplicate jobId` error.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
lucymcnatt: this is the same as it was before...but it should be an error about how the job timed out according to our maxAttempts instead, i'll fix that

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/models.scala
```diff
@@ -57,3 +57,19 @@ object StagingBucket {
...
+ * WSM state can be BROKEN, CREATING, DELETING, READY, UPDATING or None
+ * if None --> it is deletable and is deleted
+ * (already deleted in WSM, need to clean up leo resources)
+ */
+case class WsmState(state: Option[String]) extends AnyVal {
+
+  def value: String = state.getOrElse("DELETED").toUpperCase()
+
+  /** Any in-progress state cannot be deleted: CREATING, DELETING, UPDATING */
+  def isDeletable: Boolean = Set("BROKEN", "READY", "DELETED") contains this.value
```
cpate4: I think we are creating a new `Set("BROKEN", "REA...` each time this is called.
Does it make sense to create a constant? (e.g. DELETABLE_STATES or something like that?)

core/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/modelsSpec.scala
```diff
@@ -0,0 +1,35 @@
...
+  it should "determine if WSM state is deletable" in {
+    WsmState(None).isDeletable shouldBe true
+    WsmState(Some("BROKEN")).isDeletable shouldBe true
+    WsmState(Some("READY")).isDeletable shouldBe true
+    WsmState(Some("CREATING")).isDeletable shouldBe false
+    WsmState(Some("UPDATING")).isDeletable shouldBe false
+  }
+  it should "determine if WSM state is deleted" in {
+    WsmState(None).isDeletable shouldBe true
+    WsmState(Some("anything else")).isDeletable shouldBe false
```
cpate4: do we want to raise an exception (probably excessive) or some other notification (log a warning?) that we can respond to in the case of `WsmState(Some('unsupported type here'))`?

core/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/modelsSpec.scala
```diff
@@ -0,0 +1,35 @@
...
+
+//class kubernetesModelsSpec extends LeonardoTestSuite with Matchers with AnyFlatSpecLike {
+//  it should "convert chartName to AllowedChartName correctly" in {
+//    AllowedChartName.fromChartName(ChartName("/leonardo/cromwell")) shouldBe None
+//    AllowedChartName.fromChartName(ChartName("/leonardo/sas")) shouldBe Some(AllowedChartName.Sas)
+//    AllowedChartName.fromChartName(ChartName("/leonardo/rstudio")) shouldBe Some(
+//      AllowedChartName.RStudio
+//    )
+//  }
+//}
```
cpate4: Is this testing code?, and if so, should it be deleted? or is this going to be used in a subsequent PR / ticket?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/ClusterComponent.scala
```diff
@@ -712,7 +712,7 @@ object clusterQuery extends TableQuery(new ClusterTable(_)) {
         .result
         .headOption
       diskIdOpt <- runtimeConfigId.flatTraverse(rid => RuntimeConfigQueries.getDiskId(rid))
-      _ <- diskIdOpt.traverse(diskId => persistentDiskQuery.updateStatus(diskId, DiskStatus.Deleted, dateAccessed))
+      _ <- diskIdOpt.traverse(diskId => persistentDiskQuery.delete(diskId, dateAccessed))
```
cpate4, line 715: appreciating the method rename here - thanks!

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/PersistentDiskComponent.scala
```diff
@@ -254,7 +254,7 @@ object persistentDiskQuery {
 
   def nullifyDiskIds = persistentDiskQuery.tableQuery.map(x => x.lastUsedBy).update(None)
 
-  def delete(id: DiskId, destroyedDate: Instant) =
+  def delete(id: DiskId, destroyedDate: Instant): DBIO[Int] =
```
cpate4, line 257: confirming for my own knowledge and understanding. 
what we are doing here is making sure the return type is explicit and not leaving it up to the compiler, correct?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/Boot.scala
```diff
@@ -197,7 +198,8 @@ object Boot extends IOApp {
           appDependencies.googleDependencies.googleComputeService,
           googleDependencies.googleResourceService,
           gkeCustomAppConfig,
-          appDependencies.wsmDAO
+          appDependencies.wsmDAO,
+          appDependencies.wsmClientProvider
```
cpate4, line 202: thank you for making these explicit - 🤓

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskServiceInterp.scala
```diff
@@ -450,7 +450,14 @@ case class PersistentDiskAlreadyExistsException(googleProject: GoogleProject,
 
 case class DiskCannotBeDeletedException(id: DiskId, status: DiskStatus, cloudContext: CloudContext, traceId: TraceId)
     extends LeoException(
-      s"Persistent disk ${id.value} cannot be deleted in ${status} status. CloudContext: ${cloudContext.asStringWithProvider}",
+      s"Persistent disk ${id.value} cannot be deleted in $status status. CloudContext: ${cloudContext.asStringWithProvider}",
```
cpate4, line 453: was this you or the `scalafmt`?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskV2ServiceInterp.scala
```diff
@@ -115,6 +97,27 @@ class DiskV2ServiceInterp[F[_]: Parallel](config: PersistentDiskConfig,
...
       _ <- ctx.span.traverse(s => F.delay(s.addAnnotation("Done auth call for delete azure disk permission")))
 
+      // check that workspaceId is not null
+      workspaceId <- F.fromOption(disk.workspaceId, DiskWithoutWorkspaceException(diskId, ctx.traceId))
+
+      hasWorkspacePermission <- authProvider.isUserWorkspaceReader(
+        WorkspaceResourceSamResourceId(workspaceId),
+        userInfo
+      )
+      _ <- F.raiseUnless(hasWorkspacePermission)(ForbiddenError(userInfo.userEmail))
```
cpate4, line 109: Two questions:
- Does WSM own this resource?, and
- If so, should we defer to WSM to perform this authorization check versus possibly doing that check 2x?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskV2ServiceInterp.scala
```diff
@@ -115,6 +97,27 @@ class DiskV2ServiceInterp[F[_]: Parallel](config: PersistentDiskConfig,
...
+
+      // check if disk resource is deletable in WSM
+      wsmResourceId <- F.fromOption(disk.wsmResourceId, DiskWithoutWsmResourceIdException(disk.id, ctx.traceId))
+      wsmStatus <- wsmClientProvider.getDiskState(userInfo.accessToken.token, workspaceId, wsmResourceId)
+
+      _ <- F.raiseUnless(wsmStatus.isDeletable)(
+        DiskCannotBeDeletedWsmException(disk.id, wsmStatus, disk.cloudContext, ctx.traceId)
+      )
```
cpate4: ... as the caller asked to delete the disk, should we go through some level of either explanation back to the user or attempts on our end to delete the disk?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskV2ServiceInterp.scala
```diff
@@ -147,3 +150,10 @@ case class DiskWithoutWorkspaceException(id: DiskId, traceId: TraceId)
       StatusCodes.Conflict,
       traceId = Some(traceId)
     )
+
+case class DiskWithoutWsmResourceIdException(id: DiskId, traceId: TraceId)
```
cpate4: I am also kind of curious about this ... 
If there is no WsmResource for this, should we clean-up our own state? Mark it or ask the caller to confirm that the resource has been removed?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
```diff
@@ -806,6 +796,35 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
...
+    // check if databases, namespaces and managed identities associated with the app can be deleted
+    // (but only for Azure apps)
+    _ <-
+      if (cloudContext.cloudProvider == CloudProvider.Azure) {
+        for {
+          _ <- checkIfSubresourcesDeletable(app.id, WsmResourceType.AzureDatabase, userInfo, workspaceId)
+          _ <- checkIfSubresourcesDeletable(app.id, WsmResourceType.AzureManagedIdentity, userInfo, workspaceId)
+          _ <- checkIfSubresourcesDeletable(app.id, WsmResourceType.AzureKubernetesNamespace, userInfo, workspaceId)
+        } yield ()
+      } else F.unit
```
cpate4, line 813: I am curious why we are checking individually if each of these can be deleted before then deleting the disk?
Two questions:
- was this an issue with the pre-existing code?, or
- is this something we can delegate to WSM since they appear to be the owning entity for these resources?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
```diff
@@ -806,6 +796,35 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
...
+
+    // Get the disk and check if its deletable (if disk is being deleted)
+    diskIdOpt = if (deleteDisk) app.appResources.disk.map(_.id) else None
+    _ = (deleteDisk, diskIdOpt, cloudContext.cloudProvider) match {
+      // only check WSM state for Azure apps (Azure apps don't have disks currently, but they are coming...)
+      case (true, Some(diskId), CloudProvider.Azure) =>
+        for {
+          _ <- checkIfSubresourcesDeletable(app.id, WsmResourceType.AzureDisk, userInfo, workspaceId)
```
cpate4, line 821: see comment above re: `checkIfSubresourcesDeletable(...)` before making the call.
thinking through this some more ... is this because we don't have a way of updating our database state that the resource was not deleted and are trying to provide feedback to the user sync vs sending the message and having them check the state at a later point-in-time?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
```diff
@@ -815,14 +834,47 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
...
+            F.pure(WsmState(None)) // no get endpoint for a storage container in WSM yet
+        }
+        _ <- F
+          .raiseUnless(wsmState.isDeletable)(
+            AppResourceCannotBeDeletedException(resource.resourceId, appId, wsmState.value, resourceType, ctx.traceId)
+          )
+      } yield ()
+    }
+  } yield ()
+
```
cpate4, line 877: see comments above re: my trying to understand more deeply why we need to do this - thank you for helping me understand more!

cpate4 (commented):
I am through the first 11 files - wanted to get these questions out while I work through the next block of files

core/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/modelsSpec.scala
lucymcnatt: yeah I can delete that, sorry!

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/models.scala
lucymcnatt: oh yeah that makes sense

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskServiceInterp.scala
lucymcnatt: the git diff here is weird, I created the next exception below, but since the last 2 args are the same it reads as I added them to the above class....

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
lucymcnatt: These are resources specific to apps (not disks), and before this we weren't checking if they are deletable, just (blindly) sending a delete message and not polling on it. WSM has no concept of these resources being owned by specific apps, that's something Leo manages, so that's why they're checked here. This could really be before or after checking the disk state, they're all sub-resources, but since we don't always delete disks, there's more extra logic necessary, which is why checking the disk state is separated out.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
lucymcnatt: hmm maybe I'm not fully understanding this, but the idea behind checking if something is deletable in front leo is to fail faster so that the user can try again, rather than waiting until the message is processed by back leo, at which point it could be partially deleted and be in an unrecoverable state.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskV2ServiceInterp.scala
lucymcnatt: WSM does 'own' azure disks , but in back leo we send the message to WSM with the Leo auth, so WSM doesn't have knowledge of the user/their permissions

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskV2ServiceInterp.scala
lucymcnatt: I added a message to the error here for the user to wait and retry because the non-deletable messages are in-progress ones (CREATING, UPDATING, DELETING)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskV2ServiceInterp.scala
lucymcnatt: Hmm I think you're correct in this case, the wsmResourceId is set in back leo after the call to WSM to create the disk, so if there's no resourceId, then the resource was never created in WSM so we can just set it as deleted in Leo.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
lucymcnatt: also disks aren't _technically_ used for Azure apps yet, but with the Rstudio app they will be, so I thought it made sense to include with the rest of this work

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
cpate4: sorry ... I had to re-read a couple times to make sure I understood, and now I understand the comment better, and your code is fine. sorry to be confusing. 

I thought wrongly that this was erroring if we attempted to delete a disk that doesn't exist, but that is not the case here .

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
cpate4: yeah, I am seeing that now (getting more aligned with how things are currently designed). thank you for explaining this to me

cpate4 (approved):

cindy-broadinstitute (approved):
LGTM

PR #4122, opened 2024-01-18
RW-11643 Don't filter out apps with workspaceId

Qi77Qi:
https://precisionmedicineinitiative.atlassian.net/browse/RW-11643

There's recent change that we started passing workspaceId for AoU created apps. So the current query doesn't work anymore. 

This PR does the join to filter on `cloudContext`.

Actual prepared statement generated by Slick:
```
{"@timestamp":"2024-01-18T18:35:41.374558Z","@version":"1","logger_name":"slick.jdbc.JdbcBackend.statement","thread_name":"mysql.db-27","severity":"DEBUG","serviceContext":null,"message":"Preparing statement: \n            UPDATE APP\n            JOIN NODEPOOL ON APP.nodepoolId = NODEPOOL.id\n            JOIN KUBERNETES_CLUSTER ON KUBERNETES_CLUSTER.id = NODEPOOL.clusterId\n            SET APP.dateAccessed = ?\n            where APP.appName = ? AND cloudContext = ?\n         "}
{"@timestamp":"2024-01-18T18:35:41.381866Z","@version":"1","logger_name":"slick.jdbc.JdbcBackend.benchmark","thread_name":"mysql.db-27","severity":"DEBUG","serviceContext":null,"message":"Execution of prepared statement took 4ms"}
```

### What

-

### Why

-

### What to test

- 

### Who tested and where

- [ ] I validated this change 

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/AppComponent.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/AppComponent.scala
@@ -16,7 +16,7 @@ import org.broadinstitute.dsp.Release
 import org.http4s.Uri
 import slick.lifted.Tag
 
-import java.sql.SQLIntegrityConstraintViolationException
+import java.sql.{SQLIntegrityConstraintViolationException, Timestamp}
 import java.time.Instant
 import scala.concurrent.ExecutionContext
 
@@ -315,11 +315,13 @@ object appQuery extends TableQuery(new AppTable(_)) {
   def updateDateAccessed(appName: AppName, cloudContext: CloudContext, now: Instant): DBIO[Int] =
     cloudContext match {
       case CloudContext.Gcp(_) =>
-        // This query isn't efficient because appName isn't indexed. Future optimization is needed for findActiveByNameQuery method
-        findActiveByNameQuery(appName)
-          .filter(r => r.workspaceId.isEmpty)
-          .map(a => a.dateAccessed)
-          .update(now)
+        sql"""
+            UPDATE APP
+            JOIN NODEPOOL ON APP.nodepoolId = NODEPOOL.id
+            JOIN KUBERNETES_CLUSTER ON KUBERNETES_CLUSTER.id = NODEPOOL.clusterId
+            SET APP.dateAccessed = ${Timestamp.from(now)}
+            where APP.appName = ${appName.value} AND cloudContext = ${cloudContext.asCloudContextDb.value}
+         """.asUpdate
       case CloudContext.Azure(_) =>
         DBIO.failed(new RuntimeException("Please don't use this query for Azure apps"))
     }
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/DateAccessedUpdater.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/DateAccessedUpdater.scala
@@ -54,7 +54,10 @@ class DateAccessedUpdater[F[_]](
           .transaction
           .void
     case UpdateTarget.App(appName) =>
-      appQuery.updateDateAccessed(appName, msg.cloudContext, msg.dateAccessd).transaction.void
+      metrics.incrementCounter("appAccessCount") >> appQuery
+        .updateDateAccessed(appName, msg.cloudContext, msg.dateAccessd)
+        .transaction
+        .void
   }
 }
 
--- a/http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/db/AppComponentSpec.scala
+++ b/http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/db/AppComponentSpec.scala
@@ -11,6 +11,7 @@ import org.broadinstitute.dsde.workbench.leonardo.TestUtils._
 import scala.concurrent.ExecutionContext.Implicits.global
 import org.scalatest.flatspec.AnyFlatSpecLike
 
+import java.time.Instant
 import java.util.UUID
 
 class AppComponentSpec extends AnyFlatSpecLike with TestComponent {
@@ -72,6 +73,23 @@ class AppComponentSpec extends AnyFlatSpecLike with TestComponent {
       KubernetesServiceDbQueries.getActiveFullAppByName(savedCluster1.cloudContext, savedApp1.appName)
     }
     getApp.get.app.status shouldEqual AppStatus.Running
+  }
+
+  it should "update dateAccessed properly" in isolatedDbTest {
+    val savedCluster1 = makeKubeCluster(1).save()
+    val savedNodepool1 = makeNodepool(1, savedCluster1.id).save()
+
+    val app1 = makeApp(1, savedNodepool1.id)
+    val savedApp1 = app1.save()
+    val dateAccessed = Instant.ofEpochMilli(2000)
+
+    savedApp1.status shouldEqual app1.status
+    dbFutureValue(appQuery.updateDateAccessed(app1.appName, savedCluster1.cloudContext, dateAccessed)) shouldEqual 1
+
+    val getApp = dbFutureValue {
+      KubernetesServiceDbQueries.getActiveFullAppByName(savedCluster1.cloudContext, savedApp1.appName)
+    }
+    getApp.get.app.auditInfo.dateAccessed shouldEqual dateAccessed
   }
 
   it should "fail to save an app without a nodepool" in isolatedDbTest {


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/AppComponent.scala
```diff
@@ -315,11 +315,13 @@ object appQuery extends TableQuery(new AppTable(_)) {
   def updateDateAccessed(appName: AppName, cloudContext: CloudContext, now: Instant): DBIO[Int] =
     cloudContext match {
       case CloudContext.Gcp(_) =>
-        // This query isn't efficient because appName isn't indexed. Future optimization is needed for findActiveByNameQuery method
-        findActiveByNameQuery(appName)
-          .filter(r => r.workspaceId.isEmpty)
```
LizBaldo, line 320: Why do you need a new query? Could you not filter by empty workspaces anymore?

Qi77Qi: I could...it's just the query would be a bit less performant since `appName` is not indexed (might not even be unique I'm guessing)...and it's slightly less correct in that it's not filtering by `cloudContext` (altho think we're already doing this in a few other places, which is the reason why I didn't do the extra join in the original PR...figured I'll make things a bit more correct given this is relatively small PR itself)...

LizBaldo: That makes sense! For context this is the ticket that Lily worked on that made me pause https://broadworkbench.atlassian.net/browse/IA-4626

LizBaldo (approved):

yonghaoy (approved):

cpate4 (approved):
:LGTM:

cpate4:
... I also did a search for other instances of this and turned up nothing:
```shell
pate@wm9a4-6d5 leonardo % git grep 'workspaceId.isEmpty'
http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/AppComponent.scala:          .filter(r => r.workspaceId.isEmpty)
http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterpSpec.scala:        val service = if (workspaceId.isEmpty) testService else testServiceForOneWorkspace
pate@wm9a4-6d5 leonardo %
```

Qi77Qi:
@LizBaldo  https://github.com/DataBiosphere/leonardo/actions/runs/7574723382/job/20630643026?pr=4122 looks like it's running some workflow...I'm guessing it's irrelevant to the PR, but lmk if it does

Qi77Qi:
Looks like https://github.com/DataBiosphere/leonardo/actions/runs/7574723382/job/20630643026?pr=4122 is just duplicate of the jenkins job...

Since the jenkins job succeeded with one test that's unrelated to apps, I'm going to merge this to catch the release train tmr

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/AppComponent.scala
mlilynolting: 1. The query itself: Do we need to preserve the original parts of `findActiveByNameQuery`? It appears that that search also includes these terms:
-- status not deleted 
-- no destroyed date

2. The query language: Does this need to be a raw SQL query? It seems like it would be simple enough to write this as Slick.

3. Indexing: I think the APP table is indexed by appName:
```
+-------+------------+--------------------+--------------+---------------+-----------+-------------+----------+--------+------+------------+---------+---------------+
| Table | Non_unique | Key_name           | Seq_in_index | Column_name   | Collation | Cardinality | Sub_part | Packed | Null | Index_type | Comment | Index_comment |
+-------+------------+--------------------+--------------+---------------+-----------+-------------+----------+--------+------+------------+---------+---------------+
| APP   |          0 | PRIMARY            |            1 | id            | A         |           3 |     NULL | NULL   |      | BTREE      |         |               |
| APP   |          0 | IDX_APP_UNIQUE     |            1 | appName       | A         |           3 |     NULL | NULL   |      | BTREE      |         |               |
| APP   |          0 | IDX_APP_UNIQUE     |            2 | nodepoolId    | A         |           3 |     NULL | NULL   |      | BTREE      |         |               |
| APP   |          0 | IDX_APP_UNIQUE     |            3 | workspaceId   | A         |           3 |     NULL | NULL   | YES  | BTREE      |         |               |
| APP   |          0 | IDX_APP_UNIQUE     |            4 | destroyedDate | A         |           3 |     NULL | NULL   |      | BTREE      |         |               |
| APP   |          1 | FK_APP_NODEPOOL_ID |            1 | nodepoolId    | A         |           3 |     NULL | NULL   |      | BTREE      |         |               |
| APP   |          1 | FK_APP_DISK_ID     |            1 | diskId        | A         |           3 |     NULL | NULL   | YES  | BTREE      |         |               |
+-------+------------+--------------------+--------------+---------------+-----------+-------------+----------+--------+------+------------+---------+---------------+
```
This is from my BEE; is prod different? Here it seems that a search on appName first would be indexed as expected.

mlilynolting (commented):
Unless I am missing something, I think there are changes that need to happen to this to make it correct in regression.

Qi77Qi:
@mlilynolting 
> Why do we need to join nodepool or kubernetes_cluster? It seems like all the fields we need are on APP.

Becuz `cloudContext` only exists in kubernetes_cluster table

>Do we need to preserve the original parts of findActiveByNameQuery?

I don't think it hurts to have the extra filter (it's possible to have same `appName`, `cloudContext` for an active one and a deleted one)

`appName` alone is not indexed, nor it's guranteed to be unique in DB. The combination of (`appName`,`nodepoolId`,`workspaceId`,`destroyedDate`) is unique and indexed

mlilynolting:
The combination is unique; SQL indexes can be used for querying as long as the elements are present in the correct order: given an index `[1, 2, 3]`, a query on `WHERE 1` or `WHERE 1 AND 2` will use the index; a query on `WHERE 2 AND 3` will not.

mlilynolting:
I think this PR removed the filter on deleted/dateAccessed, right? Do we need to put it back in?

Qi77Qi:
the original code doesn't filter on `dateAccessed`, it's just the target to update...I'll add back the filter on nondeleted

mlilynolting:
@Qi77Qi thank you for adding back the filters on status and destroyed date! That's great.

I was referring to the filter on app `appName`, which as written does use the index. Just wanted to point out that it should use the index as written, and I don't think there is a performance issue there :]

mlilynolting:
It would also be really nice for you to change this back to Slick; we have just removed the last raw SQL strings from Leo and it'd be great to keep this consistent. I understand if you don't have time, though!

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/AppComponent.scala
```diff
@@ -315,11 +315,13 @@ object appQuery extends TableQuery(new AppTable(_)) {
...
-          .map(a => a.dateAccessed)
-          .update(now)
+        sql"""
+            UPDATE APP
+            JOIN NODEPOOL ON APP.nodepoolId = NODEPOOL.id
+            JOIN KUBERNETES_CLUSTER ON KUBERNETES_CLUSTER.id = NODEPOOL.clusterId
+            SET APP.dateAccessed = ${Timestamp.from(now)}
+            where APP.appName = ${appName.value} AND cloudContext = ${cloudContext.asCloudContextDb.value}
+         """.asUpdate
```
mlilynolting, line 324: @Qi77Qi Does this look like the right Slickification of your query?
```
someAppName = appName.value
someCloudContext = cloudContext.asCloudContextDb.value
appQuery
  .filter(
    appName === someAppName && 
    status =!= deleted && 
    destroyedDate === dummyDate
  )
  .join(nodepoolQuery)
  .on((app, nodepool) => app.nodepoolId === nodepool.id)
  .join(kubernetesClusterQuery)
  .on(((app, nodepool), kubernetesCluster) => 
    nodepool.clusterId === kubernetesCluster.id && 
    kubernetesCluster.cloudContext === someCloudContext
  )
  .map(((app, nodepool), kubernetesCluster) => app.dateAccessed)
  .update(Timestamp.from(now))
```

sjacobs146:
Is there a risk of SQL injection now that Slick is removed?

Qi77Qi:
@sjacobs146 I don't think it would. I think the problem with the list endpoint Lily worked on is because `#$` was used, which  was to support those label filters with dynamically generated SQL code. This code here does not have that.

PR #4152, opened 2024-01-26
[IA-4634] Add polling when deleting disk in WSM

lucymcnatt:
__Jira ticket__: https://broadworkbench.atlassian.net/browse/IA-4634

[WSM state reconciliation design](https://broadworkbench.atlassian.net/wiki/spaces/IA/pages/2849177613/2023-09-12+WSM+State+Reconciliation)

### What

- When deleting an Azure runtime, poll on disk deletion!

### Why

- If the disk deletion errors, we need to know so we can update the runtime state appropriately

### What to test

- 

### Who tested and where

- [ ] I validated this change 


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
```diff
@@ -896,14 +902,17 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
...
-                                  auth: Authorization
-  )(implicit
-    ev: Ask[F, AppContext]
-  ): F[WsmJobId] =
+  private def deleteDiskInWSM(diskId: DiskId,
+                              wsmResourceId: WsmControlledResourceId,
+                              workspaceId: WorkspaceId,
+                              auth: Authorization
+  )(implicit ev: Ask[F, AppContext]): F[Unit] = {
+    implicit val wsmDeleteDiskDoneCheckable: DoneCheckable[GetDeleteJobResult] = (v: GetDeleteJobResult) =>
```
LizBaldo: Is this implicit used by `wsmDao.getDeleteDiskJobResult`? Could you add a little comment explaining what it does?

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandlerSpec.scala
```diff
@@ -484,6 +484,118 @@ class AzurePubsubHandlerSpec
     res.unsafeRunSync()(cats.effect.unsafe.IORuntime.global)
   }
 
+  it should "poll on azure delete vm and disk " in isolatedDbTest {
```
LizBaldo, line 487: This  might not be easy to test (and perhaps outside of the scope of this PR), but wold it be possible to also test the case when the delete `WsmJobStatus` is failed?

LizBaldo (approved):
LGTM, the refactoring is super clear and follows the design doc 👍 
I just have a few nits:
- There is a curious implicit that could use a comment or two
- A potential sad path unit test
- Could you link the ticket / design doc in the PR as well?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
```diff
@@ -775,23 +774,24 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
       _ <- clusterQuery.updateClusterStatus(e.runtimeId, RuntimeStatus.Error, now).transaction
 
       auth <- samDAO.getLeoAuthToken
+      diskIdOpt <- clusterQuery.getDiskId(e.runtimeId).transaction
 
-      _ <- e.useExistingDisk match {
-        case false =>
+      _ <- (e.useExistingDisk, diskIdOpt) match {
+        // disk was supposed to be created and was
```
cpate4, line 779: this looks like we have a record in our database, and now we now need to verify that WSM is in a "correct state" (meaning that the record has been deleted) - does that sound correct?

cpate4: ... and if so, it feels like we are trying to contend with WSM's internal state management (e.g. if they are like this, then do that, but if they are like this other thing, then do that other thing).

Is this something where we can simply issue a request to WSM to say, "Our user has asked that we delete their disk. We have a record of the disk, so please delete any disks matching this criteria."

At this point it is WSM's responsibility to "do the right thing"?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
```diff
@@ -926,7 +935,60 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
...
+          },
+          ctx.now,
+          "deleteDiskV2"
+        )
+      )
+    } yield ()
+  }
 
   override def deleteDisk(msg: DeleteDiskV2Message)(implicit ev: Ask[F, AppContext]): F[Unit] = {
     implicit val wsmDeleteDiskDoneCheckable: DoneCheckable[GetDeleteJobResult] = (v: GetDeleteJobResult) =>
```
cpate4: this code is being shown as not being used anywhere by my IDE - can we delete it?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
cpate4: two things ... my IDE is showing:
- this is a duplicate code fragment to what is on line 548/994 in `#deleteDisk(...)`
- if I delete this line of code (or make it non-implicit, then line 950 will not compile (implicit is used there)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
```diff
@@ -926,7 +935,60 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
...
+        auth
+      )
+
+      // We need to wait until WSM deletion job to be done to update the database
+      taskToRun = for {
+        resp <- streamFUntilDone(
+          getDeleteJobResult,
+          config.deleteDiskPollConfig.maxAttempts,
+          config.deleteDiskPollConfig.interval
+        ).compile.lastOrError
```
cpate4, line 946: ... this is where the implicit is being used (before the `.compile(...)` on the `streamFUntilDone(...)` method

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
```diff
@@ -926,7 +935,60 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
...
+            } yield ()
+          case WsmJobStatus.Failed =>
+            F.raiseError[Unit](
+              DiskDeletionError(
+                diskId,
+                workspaceId,
+                s"WSM deleteDisk job failed due to ${resp.errorReport.map(_.message).getOrElse("unknown")}"
+              )
+            )
+          case WsmJobStatus.Running =>
```
cpate4, line 966: the job status will be left as "Running" if we exceed "maxAttempts"?

cpate4 (approved):
I think it looks good - just some general knowledge share comments

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandlerSpec.scala
lucymcnatt: I can definitely add that case!

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
lucymcnatt: So since this is within the handleAzureRuntimeCreationError function, this is only triggered when runtime creation fails for some reason and we need to clean up any extraneous resources in that case. `diskRecordOpt` will only be a `Some()` when our DB has been updated, which only happens after the disk creation message is sent to WSM --> we also need this wsmResourceId from the record to send the message to WSM.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
lucymcnatt: I'll pull this out to the top of the class so it can be used across functions

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
lucymcnatt: We're checking if the status is still 'Running' after our maxAttempts, if so, we error. Our maxAttempts is high enough that it happens very rarely (can't find any instances in the last 30 days)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
lucymcnatt: if you mean the implicit --> consider it done

PR #4155, opened 2024-01-29
Ia 4799 cloned workflows apps fail to upgrade

cindy-broadinstitute:
__Jira ticket__: https://broadworkbench.atlassian.net/browse/IA-4799

### What

- create records in APP_CONTROLLED_RESOURCE when apps are cloned
- do a migration to back populate these records

### Why

- updating apps fails due to not finding databases in APP_CONTROLLED_RESOURCE when a workspace has been cloned

### What to test

- 

### Who tested and where

- [ ] I validated this change 


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/AppControlledResourceComponent.scala
```diff
@@ -88,4 +88,14 @@ object appControlledResourceQuery extends TableQuery(new AppControlledResourceTa
       .filter(_.resourceType === resourceType)
       .result
       .map(_.toList)
+
+  def getAllForAppByResourceId(appId: Long, resourceId: WsmControlledResourceId)(implicit
```
LizBaldo: Nit, `getAllControlledResourcesForAppByResourceId`?

LizBaldo: i know it is getting very long, but at least it is explicit, not opinionated wither way  though

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -782,31 +787,56 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
...
         existingControlledDbsInWorkspace <- retrieveWsmDatabases(wsmResourceApi,
-                                                                 controlledDbsForApp.map(_.prefix).toSet,
+                                                                 controlledDbsRequiredForApp.map(_.prefix).toSet,
                                                                  workspaceId.value
         )
-        wsmControlledDBResources <- controlledDbsForApp
+        wsmControlledDBResources <- controlledDbsRequiredForApp
           .map { controlledDbForApp =>
+            val maybeExistingDb = existingControlledDbsInWorkspace
+              .find(existingDb => controlledDbForApp.prefix == existingDb.wsmDatabaseName)
```
LizBaldo: I assume (hope 😬 ) that the DB names are enforced to be unique within a workspace

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -782,31 +787,56 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
...
+                val maybeAppControlledResource = appControlledResourceQuery
+                  .getAllForAppByResourceId(app.id.id, WsmControlledResourceId(resourceId))
+                  .transaction
+
+                // if the appControlledResource already exists for this app, use
+                maybeAppControlledResource match {
+                  case Some(_) =>
+                    F.pure(existingDb)
+
+                  case Nil =>
```
LizBaldo: 🤔 this is effectively doing the backfilling / migration. Instead of having this branch of logic. Would it make sense to do a one time backfilling manually, and then remove this case from the code altogether?

cindy-broadinstitute: The migration will only fix up existing cloned workspaces/databases that are missing the app_controlled_resource link. But when a new clone happens, this code links the APP_CONTROLLED_RESOURCE to the newly cloned database.  So I may be not understand or have it backward, but I thought that with this we may not have to do the migration since this will catch and fix the missing app_controlled_resource record. But, will this guaranteed to catch it before it is ever used/needed elsewhere?

LizBaldo: Okay, so the use cases here are:
1- There is an existing DB and it has been previously linked (cloned or not cloned)
2- There is an existing DB, but it has not been linked to the app (cloned)
3- There is no DB and we need to create one (not cloned)

If that's the case, adding these comments in the code would help a lot I think

mlilynolting (approved):

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/azureModels.scala
```diff
@@ -33,7 +33,10 @@ final case class AKSCluster(name: String, tags: Map[String, Boolean]) {
 
 final case class WsmManagedAzureIdentity(wsmResourceName: String, managedIdentityName: String)
 
-final case class WsmControlledDatabaseResource(wsmDatabaseName: String, azureDatabaseName: String)
+final case class WsmControlledDatabaseResource(wsmDatabaseName: String,
+                                               azureDatabaseName: String,
+                                               controlledResourceId: Option[UUID] = None
```
rtitle: Why make this an `Option`? I would think the resource should always have an `id` defined?

(Flagged this because later on I see `controlledResourceId.get`, so it seems like we always expect this to be defined.)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
rtitle: They are -- WSM names are unique within a workspace.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -782,31 +787,60 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
...
-                  WsmControlledDatabaseResource(db.getAzureDatabase.getMetadata.getName,
-                                                db.getAzureDatabase.getAttributes.getDatabaseName
+            val maybeExistingDb = existingControlledDbsInWorkspace
+              .find(existingDb => controlledDbForApp.prefix == existingDb.wsmDatabaseName)
+
+            maybeExistingDb match {
+              // a database already exists (because of workspace cloning) verify or create a APP_CONTROLLED_RESOURCE for it
+              case Some(existingDb) =>
+                // there is an existing db, so check to see if we have previously mapped it to appControlledResource
+                val resourceId = existingDb.controlledResourceId.get
```
rtitle: This would crash if `controlledResourceId` is `None` -- what made me think this shouldn't be an Option :)

rtitle (commented):
One comment, but the change looks correct to me 👍 

Question -- how will we backfill the existing records? I was assuming this PR would have a liquibase migration?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
rtitle: Unless I'm mistaken, this code path (`createOrFetchWsmDatabaseResources`) only gets called for _new_ app creations (from `createAndPollApp`). So this should fix the issue going forward, I think. But we still have existing apps in the system who are missing the `APP_CONTROLLED_RESOURCE` records. We could either do a one-time migration script for those, or maybe add some code path from `updateAndPollApp` to fix up those records at upgrade time?

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/azureModels.scala
cindy-broadinstitute: I made it an Option when the unit tests were failing with a null value, then I realized it was the mocks were wrong so fixed them then forgot to remove the Option, thanks for pointing it out!

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
cindy-broadinstitute: Yeah, you are correct, this was only called for new app. I tested this on my BEE but obviously got confused somewhere 😸 

I added code to the `updateAndPollApp` and tested again, hopefully correctly this time. 

Could you take another look? @rtitle

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -354,17 +354,33 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
         F.blocking(wsmApi.getAzureManagedIdentity(workspaceId.value, wsmIdentity.resourceId.value))
       }
 
+      // Create or fetch WSM managed identity if shared app
```
rtitle: It looks like this code is enumerating all managed identities for the workspace in WSM, and choosing the right one. Are lines 350-355 needed in this case? Previously we would store the WSM IDs of an app's constituent resources in the `APP_CONTROLLED_RESOURCE` table, so Leo could retrieve the specific WSM records in the `updateApp` and `deleteApp` flows. Are we continuing to use `APP_CONTROLLED_RESOURCE`, or just use WSM as the only source of truth?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -354,17 +354,33 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
...
+      wsmResourceApi <- buildWsmResourceApiClient
       // Call WSM to get the list of databases for the app.
```
rtitle: I think this comment is misplaced now -- the database stuff is further down?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -354,17 +354,33 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
...
-      wsmDatabases <- wsmDatabases.traverse { wsmDatabase =>
-        F.blocking(wsmApi.getAzureDatabase(workspaceId.value, wsmDatabase.resourceId.value))
-          .map(db => WsmControlledDatabaseResource(db.getMetadata.getName, db.getAttributes.getDatabaseName))
+      namespacePrefix = app.appResources.namespace.value
+      wsmManagedIdentityOpt <- app.samResourceId.resourceType match {
+        case SamResourceType.SharedApp =>
+          // if a managed identity has already been created in the workspace use that otherwise create a new managed identity
+          retrieveWsmManagedIdentity(wsmResourceApi, app.appType, workspaceId.value).flatMap {
+            case Some(v) => F.pure(Option(v))
+            case None    => createAzureManagedIdentity(app, namespacePrefix, workspaceId)
```
rtitle: I don't think this is right -- the `updateApp` flow should never be creating new WSM resources. For app upgrades, the app + all constituent resources are already deployed. Leo just needs to retrieve them to populate the values for a `helm upgrade` command.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -354,17 +354,33 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
...
+          }
+        case _ => F.pure(None)
       }
 
+      // Create or fetch WSM databases
+      wsmDatabases <-
+        createOrFetchWsmDatabaseResources(
```
rtitle: A similar comment here -- I think we need `createOrFetch` semantics for app _creation_ (since it may be a fresh workspace or a cloned workspace). But app _updates_ should never create a database, we should assume it already exists.

rtitle (changes requested):
Sorry, some more comments above about the approach. Happy to talk with faces too.

I'm wondering if this approach would work:

1. In `createOrFetch[Identity|Database]` (which is called from `createApp`), make sure we populate `APP_CONTROLLED_RESOURCE` for the clone and non-clone cases. This should ensure correct records for new apps going forward.
2. To back-populate existing apps, add some logic in `updateAndPollApp`:
    ```
    If it's a shared app and wsmIdentityOpt == None, retrieve the managed identity in WSM and populate the APP_CONTROLLED_RESOURCE table.

    If it's an X app type and doesn't have X's required databases, retrieve the databases in WSM and populate the APP_CONTROLLED_RESOURCE table.
    ```

Then the next upgrade should "fix" the `APP_CONTROLLED_RESOURCE` table. Note that we also rely on that table for app deletion (to know which WSM resources to delete). So I think we're also at risk of leaking cloud resources today when we delete apps. Luckily those resources do get deleted at workspace deletion time.

cindy-broadinstitute:
> I'm wondering if this approach would work:
> 
> 1. In `createOrFetch[Identity|Database]` (which is called from `createApp`), make sure we populate `APP_CONTROLLED_RESOURCE` for the clone and non-clone cases. This should ensure correct records for new apps going forward.
> 2. To back-populate existing apps, add some logic in `updateAndPollApp`:
>    ```
>    If it's a shared app and wsmIdentityOpt == None, retrieve the managed identity in WSM and populate the APP_CONTROLLED_RESOURCE table.
>    
>    If it's an X app type and doesn't have X's required databases, retrieve the databases in WSM and populate the APP_CONTROLLED_RESOURCE table.
>    ```

That was what I was attempting to do. My confusion was with the ManagedIdentity needed to process the createOrFetchWsmDatabaseResources() which I thought would be fine to use knowing that in the update flow the tables would already be created so it would just fetch them. 

As I mentioned above, I will go back to considering the migration instead of trying to update old records on the fly. It does seem like the cleaner approach. I might like to discuss the optimal way to do that.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
cindy-broadinstitute: I actually think it would be cleaner and more resilient if we did use WSM as the source of truth instead of going into contortions to keep a local Leo table up to date with WSM tables.
So I will go back to looking at migration to fix up existing missing records instead of trying to patch them in code, especially since this approach won't fix for `delete app`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
cindy-broadinstitute: createOrFetch will create them on app creation, and fetch them on app updates. But as I mentioned, I can separate this call into fetch only for updates.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
rtitle: OK - yeah if we use WSM as the source of truth we should do it consistently for `updateApp` and `deleteApp`, and ideally drop the `APP_CONTROLLED_RESOURCE` tables.

(That might also align with the changes @aherbst-broad is thinking about to add an API alternative to WSM `enumerateResources`.)

rtitle:
> As I mentioned above, I will go back to considering the migration instead of trying to update old records on the fly. It does seem like the cleaner approach. I might like to discuss the optimal way to do that.

Sounds good - thank you! Happy to discuss.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -941,12 +1031,18 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
         .asScala
         .toList
     )
+
+    // TODO: this currently matches on the 'name' (actually type) of database so for example
```
rtitle, line 1035: Names are unique within a workspace, and there can only be 1 shared app of a given type per workspace, so this works. Agree about thinking about future options though.

Also would be good to move away from using WSM `enumerateResources` in favor of a `getByName` API or something. That has implications on auth domains I think (cc @aherbst-broad )

rtitle (approved):
Cool, seems like this approach works ✅

PR #4157, opened 2024-01-29
Cloud Pub/Sub Refactoring

giventocode:
## Summary of changes

Cloud-agnostic pub/sub support using  `workbench-libs`  messaging clients/wrappers. 

### What

-

### Why

-

### What to test

- 

### Who tested and where

- [ ] I validated this change 


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/AzurePubSubConfig.scala
```diff
@@ -0,0 +1,11 @@
+package org.broadinstitute.dsde.workbench.leonardo.config
+
+final case class AzurePubSubConfig (
+  topic:String,
+  subscription:String,
+  //required if using managed identity, but None if using connection string
+  namespace:Option[String],
+  //if not set, managed identity will be used
+  connectionString:Option[String],
+  queueSize: Int = 1000
```
cpate4, line 10: Is it possible to not limit the queue size?, or is this
- required?, and
- something we already do for Google Pub/Sub?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/Config.scala
```diff
@@ -902,4 +878,15 @@ object Config {
       gkeGalaxyDiskConfig
     )
 
-}
+  implicit private val azurePubSubConfigReader: ValueReader[AzurePubSubConfig] = ValueReader.relative { config =>
+    AzurePubSubConfig(
+      config.as[String]("topic"),
+      config.as[String]("subscription"),
+      config.as[Option[String]]("namespace"),
+      config.as[Option[String]]("connectionString")
```
cpate4, line 886: ... I am not seeing the `queueSize` here, which is maybe suggesting it isn't needed?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/LeoPublisher.scala
```diff
@@ -4,23 +4,24 @@ import cats.effect.Async
 import cats.effect.std.Queue
 import cats.syntax.all._
 import fs2.Stream
-import org.broadinstitute.dsde.workbench.leonardo.dao.CloudTopicPublisher
+import io.circe.syntax._
```
cpate4, line 7: curious what this is being used for?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -238,18 +222,17 @@ class LeoPubsubMessageSubscriber[F[_]](
...
-        case CloudPubsubEvent.Azure(_) => F.unit
-      }
+
+      _ <- F.delay(event.ackHandler.ack())
+
       end <- F.realTimeInstant
-      duration = (end.toEpochMilli - event.publishedTime.toEpochMilli).millis
+      // A zero duration means the publishing time is not available, this should never be case
+      // and it must be debugged if it happens.
+      duration = Duration.create( end.toEpochMilli - event.publishedTime.getOrElse(Instant.now()).toEpochMilli,
```
cpate4, line 234: this is interesting ... curiosity thought / questions.

were you experiencing an empty `.publishedTime`?, is this something that is specific to Azure Pub/Sub?

... and to your comment, I would like to lean towards at least logging this with a `.WARN`, or marking it as expected behavior because as is this will silently produce a duration of 0 (but, at the same time, this is actually an expected condition, so I am kind of wondering if anything really needs to be debugged)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/NonLeoMessageSubscriber.scala
```diff
@@ -5,48 +5,26 @@ import cats.effect.Async
...
 import org.broadinstitute.dsde.workbench.leonardo.util.{DeleteClusterParams, DeleteNodepoolParams, GKEAlgebra}
 import org.broadinstitute.dsde.workbench.model.TraceId
 import org.broadinstitute.dsde.workbench.model.google.GoogleProject
 import org.broadinstitute.dsde.workbench.openTelemetry.OpenTelemetryMetrics
-import org.typelevel.log4cats.StructuredLogger
-import org.broadinstitute.dsde.workbench.leonardo.http.ctxConversion
-import org.broadinstitute.dsde.workbench.leonardo.model.LeoAuthProvider
 import org.broadinstitute.dsde.workbench.util2.InstanceName
+import org.broadinstitute.dsde.workbench.util2.messaging.{CloudPublisher, CloudSubscriber, ReceivedMessage}
+import org.typelevel.log4cats.StructuredLogger
```
cpate4, line 27: there are a lot of import formatting with this PR - are your settings the same as everyone else's? or are these files unformatted for some time? 

... I am asking because explicit imports are being re-written with `_` and list styles are different (just trying to avoid thrashing in format styles)

project/Settings.scala
```diff
@@ -41,7 +41,7 @@ object Settings {
     "-language:implicitConversions",     // Allow definition of implicit functions called views
     "-unchecked",                        // Enable additional warnings where generated code depends on assumptions.
     "-Xcheckinit",                       // Wrap field accessors to throw an exception on uninitialized access.
-    "-Xfatal-warnings",                  // Fail the compilation if there are any warnings.
+    //"-Xfatal-warnings",                  // Fail the compilation if there are any warnings.
```
cpate4, line 44: 🫣 - curious what warnings you were seeing?

project/Dependencies.scala
```diff
@@ -218,6 +218,7 @@ object Dependencies {
     http4sBlazeServer % Test,
     scalaTestSelenium,
     scalaTestMockito
+   // "io.circe" %% "circe-generic" % "0.14.5"
```
cpate4, line 221: should we remove this?

cpate4 (commented):
biggest questions / comments:
- curious about `queueSize` requirement in config, and 
- formatting of all imports being different
- see comment about `.publishedTime` defaulting

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/AzurePubSubConfig.scala
Qi77Qi: this is sth we already...the queue size helps back pressure...basically when subscriber can't enqueue anymore, it means back leo can't keep up

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/Config.scala
```diff
@@ -30,18 +16,8 @@ import org.broadinstitute.dsde.workbench.leonardo.http.service.AppServiceConfig
...
-  InterruptablePollMonitorConfig,
-  LeoPubsubMessageSubscriberConfig,
-  PersistentDiskMonitorConfig,
-  PollMonitorConfig
-}
-import org.broadinstitute.dsde.workbench.leonardo.util.RuntimeInterpreterConfig.{
-  DataprocInterpreterConfig,
-  GceInterpreterConfig
-}
+import org.broadinstitute.dsde.workbench.leonardo.monitor._
```
Qi77Qi, line 19: this might be a bit dangerous...is this from IJ formatter? would be good to run `sbt scalafmtAll` in the branch

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/Boot.scala
```diff
@@ -749,25 +729,45 @@ object Boot extends IOApp {
               nonLeoMessageSubscriber
             )
           }
+          //This where Azure Dependencies are created when hosting Leonardo on Azure.
         case None =>
-          val publisher = new LeoPublisher(publisherQueue, new AzurePublisher[IO] {}) // TODO: Jesus will fix this
+          val pubConfig = AzureServiceBusPublisherConfig(
```
Qi77Qi, line 734: don't feel strongly, but we typically have this set up in config file to keep the `Boot` file a bit shorter...and we can reuse config value within config file...

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
Qi77Qi, line 234: event.publishedTime.getOrElse(end) is more correct...the `Instant.now()` may have drifted a bit from `end`

project/Settings.scala
Qi77Qi: we should bring this back...otherwise I can see the repo starts to get more and more unused imports :)

Qi77Qi:
@cpate4 I'm going to merge this PR into my branch...and try addressing some of the comments [there](https://github.com/DataBiosphere/leonardo/pull/3962) to help moving things along before I go on vacation... @giventocode  will take over [the PR](https://github.com/DataBiosphere/leonardo/pull/3962)

PR #4159, opened 2024-01-30
RW-11663 track all app usage

Qi77Qi:
https://precisionmedicineinitiative.atlassian.net/browse/RW-11663

People has requested to track more app usage by duration (AoU specifically cares about cromwell, SAS, RStudio)...This PR makes it so that it'll track all apps instead of just SAS previously

Will test things out in BEE as well.

### What

-

### Why

-

### What to test

- 

### Who tested and where

- [ ] I validated this change 

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
@@ -1547,7 +1547,6 @@ class GKEInterpreter[F[_]](
         config.monitorConfig.createApp.interval
       ).interruptAfter(config.monitorConfig.createApp.interruptAfter).compile.lastOrError
 
-      readyTime <- F.realTimeInstant
       _ <-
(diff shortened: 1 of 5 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -1553,7 +1554,7 @@ class GKEInterpreter[F[_]](
             s"AoU app installation has failed or timed out for app ${appName.value} in cluster ${cluster.getClusterId.toString}"
           logger.error(ctx.loggingCtx)(msg) >>
             F.raiseError[Unit](AppCreationException(msg))
-        } else appUsageQuery.recordStart(appId, readyTime).whenA(allowedChart.trackUsage)
```
Qi77Qi, line 1556: this is no longer needed becuz it's already checked in line 1177 for all apps

LizBaldo (approved):
Tracking for all app types makes sense, thanks!

PR #4170, opened 2024-01-31
[IA-4788] New deleteAllResources API

LizBaldo:
__Jira ticket__: https://broadworkbench.atlassian.net/browse/IA-4788
__Design doc__: https://broadworkbench.atlassian.net/wiki/spaces/IA/pages/2972942387/deleteWorkspaces+deleteRuntimes+deleteApps+Design

### What

- Implements two new Routes to manage the deletion of all Google Resources:
- - DeleteAllResources -> Delete everything, including sending deletion messages to google cloud (intended to be used by AOU to mitigate cost)
- - CleanupAllResources -> Only perform the clean up in SAM, and the leo databases, but no pubsub messages sent to the cloud (intended to be called by RAWLS to clean up after the google project deletion)

### Why

- This will allow AOU to mitigate cost by deleting resources within a workspace, and also cleanup child resources records in leo and sam once rawls has deleted the underlying google project

### What to test

TODO - testing on my BEE
- Create a new google workspace
- Create a couple of runtimes and apps with their attached persistent disks via swagger
- Create an orphaned persistent disk
- Call the deleteAllResources endpoint with `deleteInCloud` set to True and `deleteDisk` set to false
- - Check that the VMs and GKE resources have been deleted
- - Check that all of the persistent disks are still present in the project
- Call the deleteAllResources endpoint with `deleteInCloud` set to True and `deleteDisk` set to true
- - Check that all of the persistent disks are now deleted

- Create the same set of resources as above
- Go into the google console and delete the google project
- - Call the deleteAllResources endpoint with `deleteInCloud` set to False and `deleteDisk` set to True
- - Check that all resources are marked as deleted in the leo and sam db
- - Check that the workspace can be deleted from the UI

### Who tested and where

- [x] I validated this change 


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/Boot.scala
```diff
@@ -214,6 +214,10 @@ object Boot extends IOApp {
         appDependencies.publisherQueue
       )
 
+      val resourcesService = new ResourcesServiceInterp[IO](
```
LizBaldo, line 225: Because we are instantiating the class here, it cannot be abstract, hence the. mess I made below 🤔

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
```diff
@@ -1496,6 +1496,16 @@ case class DeleteAllAppsCannotBePerformed(workspaceId: WorkspaceId, apps: List[A
       traceId = Some(traceId)
     )
 
+case class DeleteAllAppsV1CannotBePerformed(googleProject: GoogleProject,
```
LizBaldo, line 1644: I put this one here to keep it where the existing V2 exception is, but happy to move them both in the `LeoException` file, I don't know why the App stuff are separate

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/ResourcesServiceInterp.scala
```diff
@@ -0,0 +1,358 @@
...
+import org.broadinstitute.dsde.workbench.model.UserInfo
+
+import scala.concurrent.ExecutionContext
+
+final class ResourcesServiceInterp[F[_]: Parallel](authProvider: LeoAuthProvider[F])(implicit
+  F: Async[F],
+  dbRef: DbReference[F],
+  ec: ExecutionContext
+) extends ResourcesService[F]
+    with RuntimeService[F]
```
LizBaldo: I don't want to reinvent the wheel, and make use of the existing services that already solved the listRuntimes, listApps, and listDisks problem. 
I cannot make this class abstract because we instantiate it in `boot.scala`, but having to override a gazillion of functions at the end of this file is clearly not the solution either.

I would appreciate any scala best practice here :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/ResourcesServiceInterp.scala
```diff
@@ -0,0 +1,358 @@
...
+
+  private def deleteAllResourcesInCloud(userInfo: UserInfo, googleProject: GoogleProject, deleteDisk: Boolean)(implicit
+    as: Ask[F, AppContext]
+  ): F[Unit] =
+    for {
+      _ <- deleteAllRuntimesInCloud(userInfo, googleProject, deleteDisk)
+      _ <- deleteAllAppsInCloud(userInfo, googleProject, deleteDisk)
+      // Delete any potential left over orphaned disk in the project - This may be a complete overkill as the deletion
+      // of most disks done as part of delete runtimes and delete apps will still be in progress at that point
+      _ <- if (deleteDisk) deleteAllDisksInCloud(userInfo, googleProject) else F.unit
```
LizBaldo: Not sure if we need to do this, it is more of a last minute effort to clean up potentially orphaned disks, but doing this async is weird

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/ResourcesServiceInterp.scala
rtitle: I'm not sure I would reach for inheritance here. What if you used composition instead -- e.g. have `ResourcesServiceInterp` take `RuntimeService`, `AppService`, and `DiskService` as constructor-args. Then you could call `runtimeService.deleteRuntime()`, etc.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/ResourcesServiceInterp.scala
```diff
@@ -0,0 +1,358 @@
...
+  override def deleteAllResources(userInfo: UserInfo,
+                                  googleProject: GoogleProject,
+                                  deleteInCloud: Boolean,
+                                  deleteDisk: Boolean
+  )(implicit
+    as: Ask[F, AppContext]
+  ): F[Unit] =
+    for {
+      ctx <- as.ask
+      _ <-
```
rtitle: Make sure you're checking appropriate permissions. I guess `deleteAllResourcesInCloud` will check each resource's permissions since we're reusing the existing `delete` code. But `deleteAllResourcesRecords` probably needs to do the same.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/ResourcesServiceInterp.scala
rtitle: I think we need to do this because there may be disks in the project that are not attached to a runtime/app. I'm curious if this causes any errors due to double-deletion of disks though (since attached disk deletion will still be in progress)..

rtitle: Or maybe we pass `deleteDisk=false` to `deleteAllRuntimesInCloud() / deleteAllAppsInCloud()` and just delete all the disks in one shot?

LizBaldo: 💡 That might be a lot cleaner, yes!

LizBaldo: Thinking about it, the issue with these async calls is that many of the delete disks calls will fail because they might still be attached to the VMs or Apps 🤔 
So we need some sort of polling for deletion before moving forward with the disks

rtitle: Oh yeah, I see. Or maybe:
1. Delete all the apps/runtimes with deleteDisk=true
2. Find all the _unattached_ disks via DB query, and just delete those

LizBaldo: Yep, sounds like a plan! I bet this won't be the most elegant query ever written but as long as it does the job we are good :)

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/ResourcesServiceInterpSpec.scala
```diff
@@ -0,0 +1,102 @@
...
+
+  it should "fail deleteAll if user doesn't have project level permission" in {
+    an[ForbiddenError] should be thrownBy {
+      resourcesService
+        .deleteAllResources(unauthorizedUserInfo, project, true, true)
+        .unsafeRunSync()(cats.effect.unsafe.IORuntime.global)
+    }
+  }
+
+  it should "call deleteAllAppsRecords, deleteAllRuntimesRecords and deleteAllDisksRecords when deleteInCloud flag is false and deleteDisk is true" in {
```
LizBaldo: @rtitle I might need some help figuring out if these tests are worth having. They are failing right now because I am not mocking enough methods I think, but I don't want to mock everything either otherwise it sort of defeats the point

LizBaldo:
@rtitle I am testing this on my BEE currently, but all unit tests have been written and passing, if you could give this another pass, thanks!

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -76,8 +78,53 @@ paths:
...
+      parameters:
+        - in: path
+          name: googleProject
+          description: googleProject
+          required: true
+          schema:
+            type: string
+        - in: query
+          name: deleteInCloud
+          description: Whether or not the runtimes and apps should be deleted in the cloud. If set to false, no call to the cloud will be made and the resources will be marked as deleted in SAM and Leonardo databases.
```
rtitle, line 98: Maybe worth specifying what the default is here? Based on reading the code it seems the default is `false`.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/ResourcesRoutes.scala
```diff
@@ -0,0 +1,84 @@
...
+
+  def deleteAllResourcesForGoogleProjectHandler(userInfo: UserInfo,
+                                                googleProject: GoogleProject,
+                                                params: Map[String, String]
+  )(implicit
+    ev: Ask[IO, AppContext]
+  ): IO[ToResponseMarshallable] =
+    for {
+      ctx <- ev.ask[AppContext]
+      deleteInCloud = params.get("deleteInCloud").exists(_ == "true")
```
rtitle: Right now these would both default to `false` if not specified in the API. Do we want to treat them as required, and throw an `BadRequestException` if they're not present?

(Just noticing they are marked required in OpenAPI, thinking the backend should valdiate that too)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/ResourcesRoutes.scala
```diff
@@ -0,0 +1,84 @@
...
+      deleteDisk = params.get("deleteDisk").exists(_ == "true")
+      // We do not support both deleteInCloud AND deleteDisk flags to be set to false
+      _ <-
+        if (deleteInCloud == false && deleteDisk == false) {
+          logger.info(
```
rtitle: Nit -- I don't know how consistent we are, but typically we pass a `loggingCtx` to log messages to more easily correlate logs with a trace ID: 
```
          logger.info(ctx.loggingCtx)(
            s"Non supported combination of deleteInCloud and deleteDisk flags: ${(deleteInCloud, deleteDisk)}"
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppService.scala
```diff
@@ -32,6 +33,18 @@ trait AppService[F[_]] {
     as: Ask[F, AppContext]
   ): F[Unit]
 
+  def deleteAllApps(userInfo: UserInfo, cloudContext: CloudContext.Gcp, deleteDisk: Boolean)(implicit
+    as: Ask[F, AppContext]
+  ): F[Vector[Option[DiskName]]]
```
rtitle: Haven't made it to the implementation yet, but `Vector[Option[DiskName]]` seems kind of unwieldy. What would a vector of None's mean? Could this just be a `Vector[DiskName]`?

rtitle: Also a comment might be helpful -- what does the return type represent? The disks attached to apps which have been deleted?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskServiceInterp.scala
```diff
@@ -318,6 +318,87 @@ class DiskServiceInterp[F[_]: Parallel](config: PersistentDiskConfig,
...
+      ctx <- as.ask
+      // Get a list of disks that are not attached to any runtime or app
+      disks <- listDisks(
+        userInfo,
+        Some(cloudContext),
+        Map.empty
+      )
+
+      orphanedDisks = disks.filterNot(disk =>
+        runtimeDiskIds.contains(disk.id) | appDisksNames.contains(Some(disk.name))
```
rtitle: Super nit -- was there a choice to use `|` instead of `||`? I think the difference is that `||` is short circuiting (I had to google it 😄 https://stackoverflow.com/questions/25067702/scala-booleans-vs-vs)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskServiceInterp.scala
```diff
@@ -318,6 +318,87 @@ class DiskServiceInterp[F[_]: Parallel](config: PersistentDiskConfig,
...
+                             cloudContext: CloudContext.Gcp,
+                             runtimeDiskIds: Vector[DiskId],
+                             appDisksNames: Vector[Option[DiskName]]
+  )(implicit
+    as: Ask[F, AppContext]
+  ): F[Unit] =
+    for {
+      ctx <- as.ask
+      // Get a list of disks that are not attached to any runtime or app
+      disks <- listDisks(
```
rtitle, line 331: I noticed we're passing the `userInfo` to the `list(Runtime/Disk/App)` calls. Does this mean that `deleteAllX` will only delete the records that the calling user has access to? If the user is a workspace owner then that should be everything in the workspace; but otherwise it might _not_ be everything in the workspace.

I just checked `deleteAllAppsV2` and it works the same way. I guess only workspace owners have permission to delete the workspace, and the `deleteAll` APIs are only called at workspace deletion time. So I think this is fine -- just calling out the behavior.

rtitle (approved):
I made a few nits but the overall behavior looks correct to me. 👍 I also didn't test -- so some BEE testing would be good. :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/ResourcesRoutes.scala
LizBaldo: Oh good catch, I will throw the BadRequest Exception if they are not present, thanks!

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/ResourcesRoutes.scala
LizBaldo: Not sure we are consistent either, but I can add that in

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppService.scala
LizBaldo: Yep, I will clean up the type. They represent the non-orphaned disks attached to the app indeed

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskServiceInterp.scala
LizBaldo: Oh no choice on my part, but short circuiting makes sense, I will add it :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskServiceInterp.scala
LizBaldo: Yep, this is by design. The goal is to impersonate the workspace owner when making these calls

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/ResourcesServiceInterp.scala
```diff
@@ -0,0 +1,66 @@
...
+                                  googleProject: GoogleProject,
+                                  deleteInCloud: Boolean,
+                                  deleteDisk: Boolean
+  )(implicit
+    as: Ask[F, AppContext]
+  ): F[Unit] =
+    for {
+      ctx <- as.ask
+      cloudContext = CloudContext.Gcp(googleProject)
+      // throw 403 if no project-level permission, fail fast if the user does not have access to the project
```
aherbst-broad, line 26: What do you think about excluding this check if `deleteInCloud` is false? If we delete the google project before invoking this endpoint it will 403. We can sequence the calls however we want in rawls, but it might be more obvious to non-rawls consumers.

LizBaldo: Oh of course, yeah that makes a lot of sense, I will exclude this check, thanks!

LizBaldo: I would rather have rawls first delete the project, then call this endpoint to clean up the records otherwise it feels kinda backwards (aka it would be bad to have deleted the records and then the project deletion fails)

aherbst-broad: Yeah, that makes sense, in which case the IAM check definitely needs to not happen if deleteInCloud = true.

LizBaldo: You mean `deleteInCloud=false` correct? If so I already pushed that change, lmk

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/ResourcesRoutes.scala
```diff
@@ -0,0 +1,94 @@
...
+          )
+        } else IO.unit
+      apiCall = resourcesService.deleteAllResources(userInfo, googleProject, deleteInCloud, deleteDisk)
+      tags = Map("deleteInCloud" -> deleteInCloud.toString, "deleteDisk" -> deleteDisk.toString)
+      _ <- metrics.incrementCounter("deleteAllResources", 1, tags)
+      _ <- ctx.span.fold(apiCall)(span =>
+        spanResource[IO](span, "deleteAllResources")
+          .use(_ => apiCall)
+      )
+    } yield StatusCodes.Accepted: ToResponseMarshallable
```
aherbst-broad, line 83: Can this be made to return 200 OK or be separated into a separate endpoint with clear sync/async semantics?

LizBaldo:
@aherbst-broad I took a stab at separating the routes so that the deleteAllResources returns a 202 because it is sending delete messages to back leo, while the new cleanupAllResources endpoint returns a 200 since there is no asynchronous processes in there. For your use case with Rawls, that would mean calling `cleanupAllResources`. I am going to triple check on my BEE tomorrow morning, but let me know if this works for you

yonghaoy (approved):

aherbst-broad (approved):

PR #4205, opened 2024-02-13
[IA-4527] Make app creation idempotent

lucymcnatt:
__Jira ticket__: https://broadworkbench.atlassian.net/browse/IA-4527

## Summary of changes

There's some changes in here to refactor how I structured the `wsmApiClientProvider` getWsmState calls, so not necessary for this PR (just a cleanup I initially thought was necessary for this PR, then realized it wasn't), I can split that out separately if that's helpful!

The changes that address the true problem are in the `AKSinterpreter`, mainly adding the `retrieveWsmNamespace` function and changing how similar `retrieveWsm...` functions are used in the createApp function
### What

Check if app resources (namespace, managed identity, databases) are already created in WSM before attempting to create them upon app creation

### Why

When Leo restarts mid-app creation, some of these resources may already exist and some may not, so we need to check for all of them before creating.

### What to test

- Start creating an app, restart leo (kick the pod on your BEE) in the middle of the process and the bootmonitor should pick it up and send a new createApp message --> this should succeed

### Who tested and where

- [ ] I validated this change 


core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/models.scala
```diff
@@ -60,23 +59,18 @@ object StagingBucket {
...
- * (already deleted in WSM, need to clean up leo resources)
+ * WSM state can be Some(BROKEN, CREATING, DELETING, READY, UPDATING) or None
+ * if None --> there is no record of this resource in WSM
+ * (already deleted or never existed in WSM, need to clean up leo resources)
  */
 case class WsmState(state: Option[String]) {
 
-  val deletableStatuses: Set[String] = Set("BROKEN", "READY", "DELETED")
-  val possibleStatuses: Array[String] = State.values().map(_.toString) :+ "DELETED"
+  val deletableStatuses: Set[String] = Set("BROKEN", "READY", "NONE")
```
lucymcnatt, line 68: switched this from "DELETED" to "NONE" to more accurately reflect the option that the resource could've never been created

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/WsmApiClientProvider.scala
```diff
@@ -23,32 +24,42 @@ import org.typelevel.log4cats.StructuredLogger
  */
 trait WsmApiClientProvider[F[_]] {
 
+  // WSM state can be BROKEN, CREATING, READY, UPDATING or None, (deleted or doesn't exist)
   val possibleStatuses: Array[WsmState] =
-    State.values().map(_.toString).map(Some(_)).map(WsmState(_)) :+ WsmState(Some("DELETED"))
+    State.values().map(_.toString).map(Some(_)).map(WsmState(_)) :+ WsmState(Some("NONE"))
 
   def getControlledAzureResourceApi(token: String)(implicit ev: Ask[F, AppContext]): F[ControlledAzureResourceApi]
   def getResourceApi(token: String)(implicit ev: Ask[F, AppContext]): F[ResourceApi]
```
lucymcnatt, line 32: did a bit of a refactor here to move the match case from the AppService here (in getWsmState) instead

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/WsmApiClientProvider.scala
```diff
@@ -23,32 +24,42 @@ import org.typelevel.log4cats.StructuredLogger
  */
 trait WsmApiClientProvider[F[_]] {
 
+  // WSM state can be BROKEN, CREATING, READY, UPDATING or None, (deleted or doesn't exist)
```
jdcanas: this casing doesn't match the casing on the line below?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -954,14 +960,43 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
...
+      resourceApi
+        .enumerateResources(workspaceId, 0, 100, ResourceType.AZURE_KUBERNETES_NAMESPACE, StewardshipType.CONTROLLED)
+        .getResources
+        .asScala
+        .toList
+    )
+    wsmNamespaces.map { namespaces =>
+      // there should be only 1 kubernetes namespace per app
+      namespaces
+        .find(ns =>
```
jdcanas: hmm... should we have some sort of exception and error logging here if for some reason we get back more than one namespace that fits this criteria? Seems like a contract against business logic that an external system enforces which would be difficult to diagnose if it changed based on how this is written. A bit different than previous cardinality enforcement since we could rely on internal database constraints and corresponding unit tests at that layer.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
```diff
@@ -855,18 +855,11 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
...
-          case WsmResourceType.AzureKubernetesNamespace =>
-            wsmClientProvider.getNamespaceState(userInfo.accessToken.token, workspaceId, resource.resourceId)
-          case WsmResourceType.AzureManagedIdentity =>
-            wsmClientProvider.getIdentityState(userInfo.accessToken.token, workspaceId, resource.resourceId)
-          case WsmResourceType.AzureDisk =>
-            wsmClientProvider.getDiskState(userInfo.accessToken.token, workspaceId, resource.resourceId)
-          case WsmResourceType.AzureStorageContainer =>
-            F.pure(WsmState(None)) // no get endpoint for a storage container in WSM yet
-        }
+        wsmState <- wsmClientProvider.getWsmState(userInfo.accessToken.token,
```
jdcanas, line 942: I'm really confused how this solves the issue as the ticket is written. First of all, the ticket is talking about createApp not delete? I bet the issue exists for both though.

Secondly, there shouldn't the main logic needed be backleo logic?

As I understand it, the flow is as follows:
1. Leo recieves API request to create/delete app
2. Leo queues message to backleo
3. backleo makes various API calls to WSM to create/delete resources

As I see it, the issue is in #3. Rob's comment on the ticket supports this assessment. Atomically before we submit a request to WSM, we need to check if the state of the object permits that request (or just not fail on 409/404 from the request itself instead of making an additional WSM call which seems better). 

The main issue being that pubsub guarantees re-delivery of any message if we do not fully process it, but our backend processing can be halfway done and then a resubmission causes a second WSM call request to do the same thing which errors out. 

Am I missing something?

jdcanas (changes requested):
I'm not sure this addresses the issue the ticket is identifying? Might be worth talking through, totally possible I'm missing something.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
lucymcnatt: this change is part of the refactor I did to how the wsmApiClientProvider structures the getWsmState calls, so not necessary for this PR (just a cleanup I initially thought was necessary for this PR, then realized it wasn't), I can split that out separately if that's helpful

lucymcnatt: the changes that address the true problem are in the AKSinterpreter, mainly adding the `retrieveWsmNamespace` function and changing how similar `retrieveWsm...` functions are used in the createApp function

jdcanas: Ohhh ok I totally missed that file, I must have collapsed it or something 😅

lucymcnatt: no worries! sorry the changeset got a bit bloated by the other fix I added in...

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
lucymcnatt: yeahhh I like that idea, I can add an exception if there's multiple

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -148,21 +149,23 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
...
-      wsmNamespace <- childSpan("createWsmKubernetesNamespaceResource").use { implicit ev =>
-        createWsmKubernetesNamespaceResource(
-          app,
-          params.workspaceId,
-          namespacePrefix,
-          wsmDatabases.map(_.wsmDatabaseName),
-          wsmManagedIdentityOpt.map(_.wsmResourceName)
-        )
+      // Create or fetch WSM kubernetes namespace
+      wsmNamespace <- retrieveWsmNamespace(wsmResourceApi, namespacePrefix, params.workspaceId.value).flatMap {
```
jdcanas: Thinking out loud, I think I like the approach of two API calls here instead of what we did for making google idempotent, which was catching 409s (in the case of create) and 404s (in the case of delete). Makes it a lot more explicit and resilient to external changes at the cost of an additional API call. 

Curious to get other thoughts here to make sure there's not other trade-offs I'm missing

jdcanas (commented):
Hmm... I wonder if this is an issue with createRuntime on the azure side too. I see this in the backlog for delete, but wonder if create needs the same treatment? 

https://broadworkbench.atlassian.net/browse/IA-4635

lucymcnatt:
I had the same thoughts about createRuntime @jdcanas, I just haven't investigated thoroughly enough to make that ticket yet

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
lucymcnatt: I like this approach better as well, I'm not as familiar with how to google side works, but it makes sense to me to follow this method in a future refactor

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/models.scala
LizBaldo: Also, `DELETED` does not really exist anyway since the record get's deleted from the table?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/WsmApiClientProvider.scala
```diff
@@ -86,67 +97,93 @@ class HttpWsmClientProvider[F[_]](baseWorkspaceManagerUrl: Uri)(implicit F: Asyn
...
+      case WsmResourceType.AzureDatabase =>
+        getDatabase(token, workspaceId, wsmResourceId)
+      case WsmResourceType.AzureKubernetesNamespace =>
+        getNamespace(token, workspaceId, wsmResourceId)
+      case WsmResourceType.AzureManagedIdentity =>
+        getIdentity(token, workspaceId, wsmResourceId)
+      case WsmResourceType.AzureDisk =>
+        getDisk(token, workspaceId, wsmResourceId)
+      case WsmResourceType.AzureStorageContainer =>
+        F.pure(None) // TODO: no get endpoint for a storage container in WSM yet
```
LizBaldo, line 180: Do you know if there is an existing ticket in the WSM backlog for that? If so it might be worth linking here to the comment

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/WsmApiClientProvider.scala
LizBaldo: Nice, it is much more extendable now, good one!

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/MonitorAtBoot.scala
```diff
@@ -58,13 +58,13 @@ class MonitorAtBoot[F[_]](publisherQueue: Queue[F, LeoPubsubMessage],
 
   private def processApps: Stream[F, Unit] =
     monitoredApps
-      .parEvalMapUnordered(10) { case (a, n, c) =>
+      .parEvalMapUnordered(10) { case (app, nodepool, cluster) =>
```
LizBaldo, line 61: Thank you ❤️

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/MonitorAtBoot.scala
```diff
@@ -416,6 +417,7 @@ class MonitorAtBoot[F[_]](publisherQueue: Queue[F, LeoPubsubMessage],
           traceId = Some(traceId)
         )
       case RuntimeStatus.Starting =>
+        // TODO: starting is supported now, update accordingly
```
LizBaldo: Do you have a ticket for this that you could link?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -148,21 +149,29 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
...
-      wsmNamespace <- childSpan("createWsmKubernetesNamespaceResource").use { implicit ev =>
-        createWsmKubernetesNamespaceResource(
-          app,
-          params.workspaceId,
-          namespacePrefix,
-          wsmDatabases.map(_.wsmDatabaseName),
-          wsmManagedIdentityOpt.map(_.wsmResourceName)
-        )
+      // Create or fetch WSM kubernetes namespace
+      wsmNamespaces <- retrieveWsmNamespace(wsmResourceApi, namespacePrefix, params.workspaceId.value)
```
LizBaldo: I think we should still do a childSpan, what do you think?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -185,7 +194,7 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
       // Authenticate helm client
       authContext <- getHelmAuthContext(landingZoneResources.aksCluster.asClusterName,
                                         params.cloudContext,
-                                        namespaceName
+                                        namespace.name
```
LizBaldo, line 180: Whoa, that is a lot neater, why were we reinventing the wheel like this before? 🤔

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -1182,23 +1233,24 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
       ev.ask.flatMap(ctx => logger.warn(ctx.loggingCtx)(s"Not updating relay listener for app ${app.appName.value}"))
     }
 
-  private def buildWsmControlledResourceApiClient(implicit ev: Ask[F, AppContext]): F[ControlledAzureResourceApi] =
+  private def getLeoAuthToken: F[String] =
```
LizBaldo: Is this a shared functionality that could go into an auth or utils file maybe?

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/dao/WsmApiClientProviderSpec.scala
```diff
@@ -0,0 +1,150 @@
...
+  def newWsmProvider() =
+    new HttpWsmClientProvider[IO](baseWorkspaceManagerUrl = Uri.unsafeFromString("test")) {
+      override def getControlledAzureResourceApi(token: String)(implicit
+        ev: Ask[IO, AppContext]
+      ): IO[ControlledAzureResourceApi] = IO.pure(setUpMockResourceApi)
+    }
+
+  val wsmProvider = newWsmProvider()
+
+  it should "return disk metadata" in {
```
LizBaldo, line 37: Nit, but maybe add the sad path test for the metadata (for the container?) that is not returned? That way when we do add it eventually it will fail the test and remoind the dev to fix it / add it

LizBaldo (approved):
LGTM, I only have a few minor comment / suggestions. I did not test on my BEE though, would you like me to do that or you already covered the testing portion?

Another thought though, would these changes also make the app upgrade and deletion idempotent as well, not just the creation?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
lucymcnatt: oh yeah I can add that back in, np! We don't use one for the managed identity either, so it probs makes sense to have that as well

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
lucymcnatt: good idea, will probs be needed by the AzurePubsubHandler down the line, I can break it out

lucymcnatt:
> I did not test on my BEE though, would you like me to do that or you already covered the testing portion?

I've tested this on my BEE, but if feel like doing some extra feel free to!

> would these changes also make the app upgrade and deletion idempotent as well?

On this question, I don't think so, but the retrieveXXX functions could definitely be leveraged to make app deletion idempotent as well (I wonder if I can sneak that in here). In terms of upgrades, I don't think UPGRADING is a currently monitored status by bootMonitor for apps...

cindy-broadinstitute (approved):
LGTM!

PR #4224, opened 2024-02-16
[TOAZ-341] Cloud-agnostic auth token provider for SAM calls from Leo

giventocode:
__Jira ticket__: https://broadworkbench.atlassian.net/browse/TOAZ-341

## Summary of changes

In this PR:
 -  A new trait that encapuslates getting a service auth token from different cloud providers `CloudAuthTokenProvider`.
 -  Implementations for GCP and Azure: `GcpCloudAuthTokenProvider` and `AzureCloudAuthTokenProvider`
 -  Token is now cached until it expires instead of a getting a new token every time. 
 - Refactoring of `HttpSamDAO` to use `CloudAuthTokenProvider`.
 - New config settings to indicate Leo is hosted on Azure, Azure environment, token scope and request timeouts. Defaults continue to be GCP. 

### Who tested and where

- [ ] I validated this change 

### DIFF
--- a/http/src/main/resources/leo.conf
+++ b/http/src/main/resources/leo.conf
@@ -156,6 +156,17 @@ auth {
 }
 
 azure {
+  hosting-mode-config{
+      # If true, it is assumed that Leo is hosted on Azure and will use Azure managed identity for authentication.
+      enabled = ${?AZURE_HOSTING_MODE_ENABLED}
+      # valid values are AZURE (Azure Commercial) and AZURE_GOV (Azure Government)
+      azure-environment = ${?AZURE_HOSTING_ENVIRONMENT}
+      managed-identity-auth-config{
+        token-scope = ${?AZURE_MI_TOKEN_SCOPE}
+        token-acquisition-timeout = ${?AZURE_MI_TOKEN_ACQUISITION_TIMEOUT}
+      }
+  }
+
   hail-batch-app-config {
     enabled = ${?HAIL_BATCH_APP_ENABLED}
   }
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/CloudAuthTokenProvider.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/CloudAuthTokenProvider.scala
@@ -1,0 +1,67 @@
+package org.broadinstitute.dsde.workbench.leonardo.auth
+
+import cats.effect.{Async, Ref}
+import cats.syntax.all._
+import org.broadinstitute.dsde.workbench.leonardo.config.AzureHostingModeConfig
+import org.broadinstitute.dsde.workbench.leonardo.model.ServiceAccountProviderConfig
+import org.http4s.headers.Authorization
+import org.http4s.{AuthScheme, Credentials}
+
+trait CloudAuthTokenProvider[F[_]] {
+  def getCloudProvider: String
+  def getAuthToken: F[Authorization]
+}
+
+/***
+ * Base class for cloud service auth token providers. Handles token caching and expiration.
+ * @param cloudProvider A string representing the cloud provider (e.g. "gcp", "azure")
+ */
+abstract class CloudServiceAuthTokenProvider[F[_]](cloudProvider: String)(implicit F: Async[F])
+    extends CloudAuthTokenProvider[F] {
+  // A reference to the token, so it can be updated when it expires.
+  private val tokenRef = Ref.unsafe(none[CloudToken])
+
+  override def getCloudProvider: String = cloudProvider
+
+  def getCloudProviderAuthToken: F[CloudToken]
+
+  override def getAuthToken: F[Authorization] =
+    for {
+      cloudTokenOpt <- tokenRef.get
+      now <- F.realTimeInstant
+      validAccessToken <- cloudTokenOpt match {
+        case None =>
+          for {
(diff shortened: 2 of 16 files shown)


dvoet (approved):

http/src/main/resources/leo.conf
```diff
@@ -156,6 +156,17 @@ auth {
 }
 
 azure {
+  hosting-mode-config{
```
LizBaldo, line 159: So these env variables will come from your own leonardo helmchart?

http/src/main/resources/leo.conf
```diff
@@ -156,6 +156,17 @@ auth {
...
+      # If true, it is assumed that Leo is hosted on Azure and will use Azure managed identity for authentication.
+      enabled = ${?AZURE_HOSTING_MODE_ENABLED}
+      # valid values are AZURE (Azure Commercial) and AZURE_GOV (Azure Government)
+      azure-environment = ${?AZURE_HOSTING_ENVIRONMENT}
+      managed-identity-auth-config{
+        token-scope = ${?AZURE_MI_TOKEN_SCOPE}
+        token-acquisition-timeout = ${?AZURE_MI_TOKEN_ACQUISITION_TIMEOUT}
```
LizBaldo, line 166: Nit, but what is the unit for the timeout? seconds?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/CloudAuthTokenProvider.scala
```diff
@@ -0,0 +1,67 @@
...
+import org.http4s.{AuthScheme, Credentials}
+
+trait CloudAuthTokenProvider[F[_]] {
+  def getCloudProvider: String
+  def getAuthToken: F[Authorization]
+}
+
+/***
+ * Base class for cloud service auth token providers. Handles token caching and expiration.
+ * @param cloudProvider A string representing the cloud provider (e.g. "gcp", "azure")
```
LizBaldo: Instead of using a String here, could you use the existing CloudProvider class defined in models.scala?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/auth/CloudAuthTokenProvider.scala
```diff
@@ -0,0 +1,67 @@
...
+    if (hostingModeConfig.enabled) {
+      new AzureCloudAuthTokenProvider[F](hostingModeConfig)
+    } else {
+      new GcpCloudAuthTokenProvider[F](serviceAccountProviderConfig)
+    }
+}
+
+final case class CloudToken(value: String, expiration: java.time.Instant)
+
+object CloudProviders {
```
LizBaldo: See my comment above about using the existing class

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/auth/CloudServiceAuthTokenProviderSpec.scala
```diff
@@ -0,0 +1,113 @@
...
+
+  it should "get a new token if the first token is expired" in {
+
+    val testProvider = new TestCloudServiceAuthTokenProvider[IO](provider, tokens)
+
+    val result = getTokenAndAssertResult(testProvider, firstTokenValue)
+
+    assertTokenInResultIsValid(result)
+
+    // wait for the first token to expire
```
LizBaldo, line 55: Nit, could you specify the unit again for the timeout?

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpSamDAOSpec.scala
```diff
@@ -31,12 +33,13 @@ import java.util.UUID
...
-class HttpSamDAOSpec extends AnyFlatSpec with LeonardoTestSuite with BeforeAndAfterAll {
+class HttpSamDAOSpec extends AnyFlatSpec with LeonardoTestSuite with BeforeAndAfterAll with MockitoSugar {
+  val cloudAuthProvider = mock[CloudAuthTokenProvider[IO]]
   val config = HttpSamDaoConfig(Uri.unsafeFromString("localhost"),
                                 false,
                                 1 seconds,
-                                10,
-                                ServiceAccountProviderConfig(Paths.get("test"), WorkbenchEmail("test"))
+                                10
+                                // ServiceAccountProviderConfig(Paths.get("test"), WorkbenchEmail("test"))
```
LizBaldo: Want to delete it for good now? 🎉

LizBaldo (approved):
This looks great! I only have a few minor comments / suggestions lmk if you have follow up questions

http/src/main/resources/leo.conf
giventocode: Yep, however, I could use some help to confirm my assumptions. A value in azure.hosting-mode-config is required. For GCP, I want to set the default to false to make it transparent. I did this by explicitly setting a default value in the config class (for clarity about the intent), as well as in the reference.conf. Is this sufficient, or do I need to update a deployment helm chart to include this value and ensure this change is transparent?

LizBaldo: I think it should be sufficient, I would not want to add azure deployment specific values to the leonardo helm chart that is designed to be deployed on GCP only 👍

PR #4233, opened 2024-02-20
[RW-11461] Return autodeleteThreshold and autodeleteEnable in Get/List APP APIs

yonghaoy:

### What

-

### Why

-

### What to test

- 

### Who tested and where

- [ ] I validated this change 


http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -4126,6 +4126,13 @@ components:
         labels:
           type: object
           description: The labels of each app in the response whose key is in includeLabels in the request. Of type Map[String,String]
+        autodeleteThreshold:
```
LizBaldo, line 4132: Nit, but could you put `autoDeleteEnabled` before `autoDeleteThreshold` to match the code (and also the logic)?

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -4126,6 +4126,13 @@ components:
...
+          type: integer
+          description: The number of minutes of idle time to elapse before the app is
+              deleted in minute. When autodeleteEnabled is true, a positive integer is required
```
LizBaldo: Nit, but the `in minute` at the end is redundant

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -4172,6 +4179,13 @@ components:
         labels:
           type: object
           description: The labels of each app in the response whose key is in includeLabels in the request. Of type Map[String,String]
+        autodeleteThreshold:
```
LizBaldo: Same comments as the two above for this endpoint as well :)

LizBaldo (approved):
LGTM! I only have a couple of suggestions on the wording in the API doc :)

http/src/main/resources/swagger/api-docs.yaml
yonghaoy: emmm, would that break Leo again? (last week there was an issue caused by swapping order of workspace id and workspace source id)

LizBaldo: It will break if the api doc and the code are inconsistent, which they were. The code was listing `autoDeleteEnabled` first and then `autoDeleteThreshold` which is why I was asking to swap them in the api doc to match the code. However I see that in the Create App request from your previous PR, it is `autoDeleteThreshold` first followed by `autoDeleteEnabled`. So really I will leave which one goes first up to you, as long as it is consistent throughout the code and the api doc

yonghaoy: I see. I also would like to swap order.  If I also swap order in `CreateAppRequest`, would that break Leo?

LizBaldo: It might for people using the old client. I would not worry too much about the createapp request, it is outside of the scope of this PR

PR #4238, opened 2024-02-21
[IA-4592] Create disk polling

jdcanas:
Initial draft, still need to test and write unit tests.

### DIFF
--- a/.github/workflows/leo-build-tag-publish-and-run-tests.yml
+++ b/.github/workflows/leo-build-tag-publish-and-run-tests.yml
@@ -20,7 +20,7 @@ on:
         required: true
         default: 'develop'
         type: string
-  
+
 env:
   BEE_NAME: '${{ github.event.repository.name }}-${{ github.run_id }}-${{ github.run_attempt}}-gcp-dev'
   TOKEN: '${{ secrets.BROADBOT_TOKEN }}' # github token for access to kick off a job in the private repo
@@ -263,7 +263,7 @@ jobs:
           workflow: leonardo-swat-tests
           repo: broadinstitute/terra-github-workflows
           ref: refs/heads/main
-          wait-for-completion-timeout: 75m
+          wait-for-completion-timeout: 85m
           token: ${{ secrets.BROADBOT_TOKEN }} # github token for access to kick off a job in the private repo
           inputs: '{
               "run-name": "${{ env.LEO_SWAT_TESTS_RUN_NAME }}",
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -200,6 +200,11 @@ azure {
       max-attempts = 240 # 10 seconds * 240 is 40 min
       interval = 10 seconds
     }
+    create-disk-poll-config {
+      initial-delay = 5 seconds
+      max-attempts = 20 # 5 seconds * 10 + 5 is 55 sec
+      interval = 5 seconds
+    }
 
     runtime-defaults {
       ip-name-prefix = "ip"
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/WsmApiClientProvider.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/WsmApiClientProvider.scala
@@ -1,13 +1,12 @@
 package org.broadinstitute.dsde.workbench.leonardo.dao
 
-import bio.terra.common.tracing.JerseyTracingFilter
+
 import bio.terra.workspace.api.{ControlledAzureResourceApi, ResourceApi, WorkspaceApi}
 import bio.terra.workspace.client.ApiClient
 import bio.terra.workspace.model.{ResourceMetadata, State}
 import cats.effect.Async
 import cats.mtl.Ask
 import cats.syntax.all._
-import io.opencensus.trace.Tracing
 import org.broadinstitute.dsde.workbench.leonardo.db.WsmResourceType
 import org.broadinstitute.dsde.workbench.leonardo.{AppContext, WorkspaceId, WsmControlledResourceId, WsmState}
 import org.broadinstitute.dsde.workbench.leonardo.util.WithSpanFilter
@@ -76,7 +75,6 @@ class HttpWsmClientProvider[F[_]](baseWorkspaceManagerUrl: Uri)(implicit F: Asyn
           super.performAdditionalClientConfiguration(clientConfig)
           ctx.span.foreach { span =>
             clientConfig.register(new WithSpanFilter(span))
-            clientConfig.register(new JerseyTracingFilter(Tracing.getTracer))
           }
         }
       }
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/WsmApiClientProvider.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/WsmApiClientProvider.scala
@@ -1,6 +1,5 @@
 package org.broadinstitute.dsde.workbench.leonardo.dao
 
-
 import bio.terra.workspace.api.{ControlledAzureResourceApi, ResourceApi, WorkspaceApi}
 import bio.terra.workspace.client.ApiClient
 import bio.terra.workspace.model.{ResourceMetadata, State}
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/WsmDao.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/WsmDao.scala
@@ -24,6 +24,7 @@ import org.broadinstitute.dsde.workbench.leonardo.JsonCodec.{
 }
 import org.broadinstitute.dsde.workbench.leonardo.dao.LandingZoneResourcePurpose.LandingZoneResourcePurpose
 import org.broadinstitute.dsde.workbench.leonardo.http.service.VMCredential
+import org.broadinstitute.dsde.workbench.leonardo.util.PollDiskParams
 import org.broadinstitute.dsde.workbench.model.google.GoogleProject
 import org.broadinstitute.dsde.workbench.model.{TraceId, WorkbenchEmail}
 import org.http4s.headers.Authorization
@@ -83,7 +84,7 @@ trait WsmDao[F[_]] {
 
 final case class StorageContainerRequest(storageContainerName: ContainerName)
 final case class CreateStorageContainerRequest(workspaceId: WorkspaceId,
-                                               commonFields: ControlledResourceCommonFields,
+                                               commonFields: InternalDaoControlledResourceCommonFields,
                                                storageContainerReq: StorageContainerRequest
 )
 final case class CreateStorageContainerResult(resourceId: WsmControlledResourceId)
@@ -126,7 +127,7 @@ final case class ListLandingZoneResourcesResult(id: UUID, resources: List[Landin
 
 //Azure Vm Models
 final case class CreateVmRequest(workspaceId: WorkspaceId,
-                                 common: ControlledResourceCommonFields,
+                                 common: InternalDaoControlledResourceCommonFields,
                                  vmData: CreateVmRequestData,
                                  jobControl: WsmJobControl
 )
@@ -157,7 +158,9 @@ final case class DeleteWsmResourceRequest(workspaceId: WorkspaceId,
                                           deleteRequest: DeleteControlledAzureResourceRequest
 )
 final case class CreateVmResult(jobReport: WsmJobReport, errorReport: Option[WsmErrorReport])
+
 final case class GetCreateVmJobResult(vm: Option[WsmVm], jobReport: WsmJobReport, errorReport: Option[WsmErrorReport])
+
 final case class GetDeleteJobResult(jobReport: WsmJobReport, errorReport: Option[WsmErrorReport])
 
 sealed trait ResourceAttributes extends Serializable with Product
@@ -172,7 +175,7 @@ final case class GetJobResultRequest(workspaceId: WorkspaceId, jobId: WsmJobId)
 
 // Azure Disk models
 final case class CreateDiskRequest(workspaceId: WorkspaceId,
-                                   common: ControlledResourceCommonFields,
+                                   common: InternalDaoControlledResourceCommonFields,
                                    diskData: CreateDiskRequestData
 )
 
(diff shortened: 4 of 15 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/WsmDao.scala
```diff
@@ -36,6 +37,10 @@ trait WsmDao[F[_]] {
     ev: Ask[F, AppContext]
   ): F[CreateDiskResponse]
 
+  def createDiskV2(request: CreateDiskRequestV2, authorization: Authorization)(implicit
```
lucymcnatt: general comment here, I'd rather use the WSM client than adding more things to the DAO. We use the client via the wsmApiClientProvider in the AKSInterpreter, but I think we'd need to add it to the AzurePubsubHandler cuz we don't use it there yet

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/WsmApiClientProvider.scala
```diff
@@ -74,7 +72,6 @@ class HttpWsmClientProvider[F[_]](baseWorkspaceManagerUrl: Uri)(implicit F: Asyn
           super.performAdditionalClientConfiguration(clientConfig)
           ctx.span.foreach { span =>
             clientConfig.register(new WithSpanFilter(span))
-            clientConfig.register(new JerseyTracingFilter(Tracing.getTracer))
```
cpate4, line 79: is there a reason why we removed this as a part of this review?

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriberSpec.scala
```diff
@@ -2161,4 +2221,54 @@ class LeoPubsubMessageSubscriberSpec
         ev: Ask[IO, TraceId]
       ): IO[Option[Disk]] = IO(Some(Disk.newBuilder().setLastDetachTimestamp(Random.nextInt().toString).build()))
     }
+
+  private def setUpMockWsmApiClientProvider(
```
cpate4: can this be shared the next spec?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
```diff
@@ -93,6 +110,11 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
         )
       )
       createVmJobId = WsmJobId(s"create-vm-${runtime.id.toString.take(10)}")
+
```
cpate4: I know this isn't technically a part of the review, but ...
seeing how as our service is making requests in a polling nature to a shared resource, does it make sense to add a `leo-` prefix to the jobId? or is there another way that we are keeping track of leo-based job requests?

cpate4 (approved):

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/WsmApiClientProvider.scala
jdcanas: I had to remove this to not enter dependency hell, and the `WithSpanFilter` is the only one that has docs about what it is actually providing us in terms of logging.

If we bump the wsm library it conflicts with the ridiculous amount of excludes we have and no longer works. 

We would need to bump terra-common-lib to include the new kind of `JakartaTracing`, but of course doing this would require excluding various Liquibase deprecated methods in that lib from overriding a pinned version we depend on for core Liquibase functionality. Additionally, we exclude Jakarta explicitly so we would need to also untangle that.

Tl;dr, we need to remove this or enter dependency hell that would involve changing our `Liquibase.update` to use a new non-deprecated method

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
jdcanas: Hmm.. it might be worth doing this in a larger PR to add this to all leo requests so we can retrieve all leo-related jobs.

We don't store job IDs at all, so if leo restarts mid job polling, we lose this ID entirely I believe. cc @lucymcnatt , we should perhaps think in a larger sense how we can handle this and not end up with "stuck" resources, unless I'm missing something.

cpate4: > ... if leo restarts mid job polling, we lose this ID entirely I believe

this kind of defeats the purpose and intention of the caller creating and submitting a job-id to an async endpoint. how hard would it be to take the job-id and storing that on our end to give a reference point to "check"

cc: @dvoet

jdcanas: Yeah, the issue comes back to leo and WSM are both storing this state...

Almost seems like we should nix leo internal state entirely on these objects, have it be a nullable field in our DB now for backwards compatibility, and if its null look it up in WSM 🤔   

Then we wouldn't have to poll at all... just fire and forget, and look up the resource in WSM whenever we need to report state to an end user request

A bit of an off-the-cuff thought but it seems a lot simpler than kicking the can down the line for restart resilience and idempotency

cpate4: > Yeah, the issue comes back to leo and WSM are both storing this state...

For me, this comes down to the fact we are aiming to leverage an async endpoint, but are consequently throwing away the job-id we provided to that async endpoint to track that job's overall progress (and therefore no longer able to track the async job's state).

That being said, are things idempotent in the currently designed solution? or do we stand to create orphaned resources?

jdcanas: Yes, idempotency is the goal. We should be able to resubmit the same message to backleo (i.e., create vm), and with the Ids we should be able to figure out the state from the DB  in conjunction with WSM calls. Might  be best to discuss this in standup, since Lucy is more aware of the current progress on this front and how it relates to job id (we might be able to make a WSM request from the resource ID and figure out the WSM state/subsequent polling needed from there?)

lucymcnatt: yeah there's definitely an opportunity for some work to be done to make the polling resources idempotent...I do like the idea of standardizing the job id names in some way, then we can always grab the id with the action + wsmResourceId or something. I don't know if its necessary for it to include a leo-specific prefix since leo is the only service contacting WSM to create/delete runtimes AFAIK

http/src/main/resources/reference.conf
```diff
@@ -200,6 +200,11 @@ azure {
       max-attempts = 240 # 10 seconds * 240 is 40 min
       interval = 10 seconds
     }
+    create-disk-poll-config {
+      initial-delay = 5 seconds
+      max-attempts = 10 # 5 seconds * 10 + 5 is 55 sec
+      interval = 5 seconds
+    }
```
lucymcnatt, line 207: Did you verify in testing that ~1min was enough time for a disk to create?

jdcanas: Yea, I'll be relying on the automation tests to run this a few times (pending).

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
lucymcnatt: on your thought @jdcanas on not polling altogether....it's definitely a tempting idea, but that would also mean with every get request from the UI we'd have to send an extra WSM get request for the state. Which would work, but maybe is kinda cumbersome and adds extra load on WSM? This problem is also why I like the idea Liz proposed in our walkthrough last week, of making the proposed new services handle the full lifecycle of a specific resource type --> ie we wouldn't have to call WSM anymore for AzureControlledResources.

jdcanas: Hmmmm.... maybe a way to get the best of both worlds is to pop createDisk out to a higher level and not to do at all as part of `AzurePubsubHhandler`'s `createAndPollRuntime` which is the interface exposed. Then, the caller of `AzurePubsub.createAndPollRuntime` can instead decide whether to create a disk first, and pass that to the VM creation if needed. 

We will still update the DB as appropriate but also set ourselves up to use other DAOs ( that could be separate from leo) down the line to create the disk independently of the VM

http/src/main/resources/reference.conf
jdcanas: Anecdotally, it took seconds in my manual testing

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
jdcanas: I made a ticket for chip's concerns though https://broadworkbench.atlassian.net/browse/IA-4850

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
```diff
@@ -346,9 +367,21 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
     } yield (stagingContainerName, resp.resourceId)
   }
 
-  private def createDisk(params: CreateAzureRuntimeParams, leoAuth: Authorization)(implicit
+  // Creates a disk and synchronously monitors its creation
+  // This function is written assuming that a disk is created as part of a runtime, and bubbles errors up to the runtime so they are visible to the user
```
lucymcnatt: this assumption is fine for now since we don't have a separate createDisk endpoint for Azure and no Azure apps use disks yet, might have to be adapted later on

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
```diff
@@ -1020,4 +1129,15 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
   def getWsmJobId(jobName: String, resourceId: WsmControlledResourceId): WsmJobId = WsmJobId(
     s"${jobName}-${resourceId.value.toString.take(10)}"
   )
+
+  private def buildWsmControlledResourceApiClient(implicit ev: Ask[F, AppContext]): F[ControlledAzureResourceApi] =
```
lucymcnatt, line 1128: I think there's a similar function in AKSInterpreter, is this something we could shift somewhere both classes could access?

project/Dependencies.scala
```diff
@@ -141,7 +141,7 @@ object Dependencies {
   val pact4sCirce =       "io.github.jbwheatley"  %% "pact4s-circe"     % pact4sV
   val okHttp =            "com.squareup.okhttp3"  % "okhttp"            % "4.12.0"
 
-  val workSpaceManagerV = "0.254.916-SNAPSHOT"
+  val workSpaceManagerV = "0.254.1044-SNAPSHOT"
```
lucymcnatt: thanks for doing the work to use the updated client!

lucymcnatt (approved):
A comment about consolidation, but nothing blocking, LGTM!

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
```diff
@@ -144,76 +168,77 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
...
-        )
-      }
-      _ <- wsmDao.createVm(createVmRequest, auth)
-    } yield ()
+          config.runtimeDefaults.vmCredential,
+          createDiskResult.resourceId
+        ),
+        jobControl
+      )
+    } yield CreateResourcesRuntimeResult(createVmRequest, createDiskResult)
```
jdcanas: RuntimeResources*

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
```diff
@@ -144,76 +168,77 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
                                                            cloudContext
       )
 
-      createDiskAction = createDisk(params, auth)
-
       (stagingContainerName, stagingContainerResourceId) <- createStorageContainer(params, auth)
```
jdcanas, line 168: TODO, run this in parallel in the async task that monitors VM WSM job success

.github/workflows/leo-build-tag-publish-and-run-tests.yml
```diff
@@ -263,7 +263,7 @@ jobs:
           workflow: leonardo-swat-tests
           repo: broadinstitute/terra-github-workflows
           ref: refs/heads/main
-          wait-for-completion-timeout: 75m
+          wait-for-completion-timeout: 85m
```
LizBaldo, line 266: Damn, polling for disk creation means bumping up tests timeouts by 10 minutes or is this unrelated to the change and just an attempts to increase integration tests stability?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
```diff
@@ -455,11 +536,28 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
       resourceId
     )
 
+  private def getCommonFieldsForWsmGeneratedClient(name: ControlledResourceName,
+                                                   resourceDesc: String,
+                                                   userEmail: WorkbenchEmail
+  ) =
+    new ControlledResourceCommonFields()
+      .accessScope(bio.terra.workspace.model.AccessScope.PRIVATE_ACCESS)
```
LizBaldo, line 544: Quick question, would the accessScope be different for creating apps using pre-existing databases in cloned workspace? Definitely outside of the scope of this PR, but something to be careful with once we get to polling for app resources creation @lucymcnatt

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
LizBaldo: +1 to that, would it make sense to have this in the `WsmApiClientProvider` file?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandlerAlgebra.scala
```diff
@@ -94,25 +95,42 @@ final case class CreateAzureRuntimeParams(workspaceId: WorkspaceId,
...
+
+final case class PollDiskParams(workspaceId: WorkspaceId,
+                                jobId: WsmJobId,
+                                diskId: DiskId,
+                                runtime: Runtime,
+                                wsmResourceId: WsmControlledResourceId
 )
 
+/**
+ * This case class represents the necessary information to poll all objects associated with the runtime, namely disk and vm
```
LizBaldo, line 121: and soon storage container :)

LizBaldo (approved):
LGTM! I have a couple of questions and minor suggestions but nothing blocking. Thanks for iterating over the different options this week :)

.github/workflows/leo-build-tag-publish-and-run-tests.yml
jdcanas: Its not the disk polling, this is the gcp suite timeout, and it has been timing out occasionally as its main current failure mode... we have tickets with low hanging fruit to speed it up, but it timed out once on this PR and I got tired of it.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
jdcanas: I think this makes sense here in terms of separation of responsibility, since it lets the consumer of the client decide how to get the token rather than the wsmClient wrapper knowing the internals of leo auth interfaces (the AKSInterpreter gets the token with the `authProvider` instead of the `samDao`)

PR #4272, opened 2024-02-28
[IA-4502] add workflow dispatch for leo gcp tests

jdcanas:
I can't run the tests  multiple times if it doesn't support workflow_dispatch 🙃 .

Fun part is there is no way to test this is correct without merging either!

See this for why this logic is currently incorrect and why I'm fairly confident this is the proper fix: https://github.com/DataBiosphere/leonardo/actions/runs/8084499439/job/22089889997

### DIFF
--- a/.github/workflows/leo-build-tag-publish-and-run-tests.yml
+++ b/.github/workflows/leo-build-tag-publish-and-run-tests.yml
@@ -21,8 +21,6 @@ on:
         default: 'develop'
         type: string
   
-      
-
 env:
   BEE_NAME: '${{ github.event.repository.name }}-${{ github.run_id }}-${{ github.run_attempt}}-gcp-dev'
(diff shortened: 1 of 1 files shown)


LizBaldo (approved):

cpate4:
for future ... a tool I have been using locally:
https://nektosact.com/introduction.html

... and specifically re: workflow dispatch testing:
https://github.com/nektos/act/issues/332

jdcanas:
> for future ... a tool I have been using locally: https://nektosact.com/introduction.html
> 
> ... and specifically re: workflow dispatch testing: [nektos/act#332](https://github.com/nektos/act/issues/332)

Yeah, a lot of our logic depends on repo secrets and custom in-house workflow plugins which complicates this 😬

PR #4278, opened 2024-02-29
[IA-4313] Delete App records if workspace no longer exists

lucymcnatt:

### What

When an MRG is deleted in Azure, the workspace is deleted as well, so when Leo asks for the workspaceDescription from WSM, a 404 is returned. Instead of erroring, Leo's app records should be cleaned up in this case because if the workspace has been deleted, none of the resources under it exist anymore.

### Why

A workspace should be able to be properly cleaned up in Terra, even if the MRG is removed from under it

### What to test

- 

### Who tested and where

- [ ] I validated this change 


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppService.scala
```diff
@@ -37,7 +37,7 @@ trait AppService[F[_]] {
     as: Ask[F, AppContext]
   ): F[Vector[DiskName]]
 
-  def deleteAppRecords(userInfo: UserInfo, cloudContext: CloudContext.Gcp, appName: AppName)(implicit
+  def deleteAppRecords(userInfo: UserInfo, cloudContext: CloudContext, appName: AppName)(implicit
```
LizBaldo, line 40: Oh nice! Glad that can be reused!

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
```diff
@@ -820,7 +822,22 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
...
+
+    workspaceApi <- wsmClientProvider.getWorkspaceApi(userInfo.accessToken.token)
+    attempt <- F.delay(workspaceApi.getWorkspace(workspaceId.value, IamRole.READER)).attempt
+
+    _ <- attempt match {
+      // if the workspace is found, delete the app normally
+      case Right(workspaceDesc) =>
+        deleteAppV2Base(appResult.app, appResult.cluster.cloudContext, userInfo, workspaceId, deleteDisk, workspaceDesc)
+      // if the workspace can't be found, delete the workspace records
+      case Left(_: NotFoundException) =>
```
LizBaldo: One thing to double check though is that the `NotFoundException` does not also corresponds with a Permission Denied. 

I know that internally within Leo we often return a RuntimeNotFound when a user does not have permission for instance

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
```diff
@@ -831,23 +848,53 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
...
+
+    _ <- attempt match {
+      // if the workspace is found, delete the app normally
+      case Right(workspaceDesc) =>
+        apps
+          .traverse { app =>
+            deleteAppV2Base(app._1, app._2, userInfo, workspaceId, deleteDisk, workspaceDesc)
+          }
+      // if the workspace can't be found, delete the workspace records
+      case Left(_: NotFoundException) =>
```
LizBaldo: Same comment as above re triple checking the NotFoundException only corresponds to a 404

project/Dependencies.scala
```diff
@@ -150,13 +150,13 @@ object Dependencies {
   def excludeSpringBoot = ExclusionRule("org.springframework.boot")
   def excludeSpringAop = ExclusionRule("org.springframework.spring-aop")
   def excludeSpringData = ExclusionRule("org.springframework.data")
-  def excludeSpringFramework = ExclusionRule("org.springframework")
+  // def excludeSpringFramework = ExclusionRule("org.springframework")
```
lucymcnatt: this change was necessary when I was writing the tests...apparently the org.springframework.HttpStatus was needed to throw a NotFoundException

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
lucymcnatt: we do already check workspace permission on line 848 (and in the other just deleteAppV2 function)
`_ <- F.raiseUnless(hasWorkspacePermission)(ForbiddenError(userInfo.userEmail))`

but yes on the workspace swagger there's a separate `403 Permission denied` response

LizBaldo (approved):

PR #4283, opened 2024-03-01
[IA-4788] Use userInfo not petSA when notifying Sam of resource cleanup

LizBaldo:
__Jira ticket__: https://broadworkbench.atlassian.net/browse/IA-4788

### What

- Small follow up to the delete Google workspace work done in https://github.com/DataBiosphere/leonardo/pull/4170 to use the user info instead of the pet SA to authenticate to Sam when notifying the service that the child resource was deleted

### Why

- The pet SA is being deleted with the google project by Rawls and might not be available by the time Leonardo notifies Sam and we should not use it to authenticate in this case

### What to test

- 

### Who tested and where

- [ ] I validated this change 


yonghaoy (approved):

aherbst-broad (approved):

LizBaldo:
I just tested on my BEE and the calls to SAM worked by using the user token instead of the pet SA 👍

PR #4292, opened 2024-03-05
[IA-4832] Add esm script to csp allowlist

LizBaldo:
__Jira ticket__: https://broadworkbench.atlassian.net/browse/IA-4832

### What

- Adding a list of esm scripts (e.g. `https://esm.sh/deck.gl@8.9.21?bundle`) to Leonardo's CSP allowlist

### Why

- This script is needed to run a notebook using anywidget on Terra

### What to test

- 

### Who tested and where

- [ ] I validated this change 


sarahgibs (approved):
This looks good to me, as it is only adding the approved files to the CSP. 👍

LizBaldo:
Adding the files to the CSP list directly does not seem to work. I keep getting the following types or errors when trying to run the user's notebook:

```
Refused to load the script 'https://esm.sh/deck.gl@8.9.21?bundle' because it violates the following Content Security Policy directive: "script-src 'self' data: 'unsafe-inline' 'unsafe-eval' blob: https://apis.google.com https://cdn.jsdelivr.net https://cdn.pydata.org". Note that 'script-src-elem' was not explicitly set, so 'script-src' is used as a fallback.
```

```
Refused to load the script 'https://esm.sh/d3@7' because it violates the following Content Security Policy directive: "script-src 'self' data: 'unsafe-inline' 'unsafe-eval' blob: https://apis.google.com https://cdn.jsdelivr.net https://cdn.pydata.org". Note that 'script-src-elem' was not explicitly set, so 'script-src' is used as a fallback.
```

I am experimenting with a bunch of stuff on my BEE, but we should do a full review to make sure this is safe before merging.

LizBaldo:
So adding these scripts to the CSP allowlist are somehow blocking Leonardo setCookies endpoint:

```
Access to fetch at 'https://leonardo.ebaldo.bee.envs-terra.bio/proxy/setCookie' from origin 'https://terra.ebaldo.bee.envs-terra.bio' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.
```

I am a bit puzzled as to why this is happening, @jdcanas would you be able to have a look?

PR #4294, opened 2024-03-06
[TOAZ-344]: Enable Azure hosting support and dependency management refactoring

cpate4:
**PR Goal**: Enable Azure hosting support for the application Leo.
- **Refactoring Dependencies**: The PR introduces a refactored process for building dependencies to support cloud environments.
  - **Baseline Dependencies**: First step, dependencies required across all cloud environments are constructed -  `BaselineDependencies` case class.
  - **Service Dependencies (Frontend)**: Second step, service-specific dependencies, including cloud-specific ones, are created.
  - **Back-end Process (Backend)**: Cloud-specific back-end processes are created.
- **Dependency Registry**: Incorporates a dependency registry akin to app-contexts in DI frameworks.
  - **Cloud-Specific Dependencies**: Cloud-specific dependencies are added for scenarios where lazy access to the dependencies is required for cloud-specific functionality (GCP in all current cases).
  - **Route Determination**: `HttpRoutes` creates gcp-only routes when the gcp-specific dependencies are created.  
- **Additional Refactoring**:
   - `GcpComputeService` and `GcpResourcesServices` are optional in `LepAppServiceInterp`:  These dependencies are only needed for GCP specific functionality `LepAppServiceInterp`. When running on Azure, these dependencies are not set (None)
    - `LeoPubsubMessageSubscriber`, `NonLeoMessageSubscriber` and `LeoPublisher`: Use the cloud-agnostic pub/sub from workbenchlibs.
    - Removed `GKEService[F]` dependency in ` KubernetesInterpreter` and `KubernetesAlgebra` as it was not being used.


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/AppDependenciesBuilder.scala
```diff
@@ -0,0 +1,319 @@
...
+                                cloudSpecificDependencies: ServicesRegistry,
+                                leoExecutionModeConfig: LeoExecutionModeConfig
+  )(implicit
+    logger: StructuredLogger[IO],
+    ec: ExecutionContext,
+    dbReference: DbReference[IO],
+    openTelemetry: OpenTelemetryMetrics[IO]
+  ): Resource[IO, LeoAppProcesses] = {
+
+    val cloudSpecificProcessList =
```
cpate4, line 153: this method was taken from Boot.scala to begin making affordances for the differences between GCP and Azure deployments

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/AppDependenciesBuilder.scala
```diff
@@ -0,0 +1,319 @@
...
+      baselineDependencies.cromwellDAO,
+      baselineDependencies.hailBatchDAO,
+      baselineDependencies.listenerDAO,
+      baselineDependencies.samDAO,
+      kubeAlg,
+      baselineDependencies.azureContainerService
+    )
+
+    val cromwellAppInstall = new CromwellAppInstall[IO](
+      ConfigReader.appConfig.azure.coaAppConfig,
```
cpate4, line 194: do we need azure in the general builder config?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/Boot.scala
```diff
@@ -130,132 +52,43 @@ object Boot extends IOApp {
 
     logger.info("Liveness server has been started").unsafeRunSync()(cats.effect.unsafe.IORuntime.global)
 
-    createDependencies[IO](applicationConfig.leoServiceAccountJsonFile.toString).use { appDependencies =>
-      val googleDependencies = appDependencies.googleDependencies
+    AppDependenciesBuilder().createAppDependencies().use { leoDependencies =>
```
cpate4, line 55: @giventocode - would it be possible to get a rundown for what the `AppDependenciesBuilder`, `AzureDependenciesBuilder`, `BaselineDependenciesBuilder`, `CloudDependenciesBuilder`, and `GcpDependenciesBuilder` are and how they interrelate with one another?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/KubernetesInterpreter.scala
```diff
@@ -43,22 +33,23 @@ class KubernetesInterpreter[F[_]](azureContainerService: AzureContainerService[F
...
+//    clusterOpt <- gkeService.getCluster(clusterId)
+//    cluster <- F.fromEither(
+//      clusterOpt.toRight(
+//        KubernetesClusterNotFoundException(
+//          s"Could not create client for cluster $clusterId because it does not exist in GCP. Trace ID: ${ctx.traceId.asString}"
+//        )
+//      )
+//    )
+//    _ <- F.blocking(credentials.refreshIfExpired())
+//    token = credentials.getAccessToken.getTokenValue
```
cpate4, line 50: is this something we want to take care of now? or later (and if so, should we create a JIRA ticket for it)?

cpate4: ... also curious where we are getting the client from now

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -54,30 +53,29 @@ import scala.jdk.CollectionConverters._
...
 class LeoPubsubMessageSubscriber[F[_]](
   config: LeoPubsubMessageSubscriberConfig,
-  subscriber: GoogleSubscriber[F, LeoPubsubMessage],
+  subscriber: CloudSubscriber[F, LeoPubsubMessage],
   asyncTasks: Queue[F, Task[F]],
-  googleDiskService: GoogleDiskService[F],
+  // googleDiskService: GoogleDiskService[F],
   authProvider: LeoAuthProvider[F],
-  gkeAlg: GKEAlgebra[F],
+  // gkeAlg: GKEAlgebra[F],
```
cpate4: would we be able to remove these commented out code items, please?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -54,30 +53,29 @@ import scala.jdk.CollectionConverters._
...
+  cloudSpecificDependenciesRegistry: ServicesRegistry
 )(implicit
   executionContext: ExecutionContext,
   F: Async[F],
   logger: StructuredLogger[F],
   dbRef: DbReference[F],
-  runtimeInstances: RuntimeInstances[F],
-  monitor: RuntimeMonitor[F, CloudService],
+  // runtimeInstances: RuntimeInstances[F],
+  // monitor: RuntimeMonitor[F, CloudService],
```
cpate4: ditto re: commented out code...

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/LeoPublisher.scala
```diff
@@ -43,42 +36,40 @@ final class LeoPublisher[F[_]](
...
                   )
-                )
+                }
             }
+            .covary[F]
       }
     Stream(publishingStream, recordMetrics).covary[F].parJoin(2)
   }
 
+  private def publishMessageAndSetAttributes(event: LeoPubsubMessage) =
```
jdcanas: Can we have a type for this function's return?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/BaselineDependenciesBuilder.scala
```diff
@@ -0,0 +1,468 @@
...
+
+    for {
+      httpClient <- client
+        .BlazeClientBuilder[F]
+        .withSslContext(sslContext)
+        // Note a custom resolver is needed for making requests through the Leo proxy
+        // (for example HttpJupyterDAO). Otherwise the proxyResolver falls back to default
+        // hostname resolution, so it's okay to use for all clients.
+        .withCustomDnsResolver(dnsResolver)
+        .withConnectTimeout(30 seconds)
```
jdcanas, line 394: is it worth having this live in `reference.conf`? Might be worth considering that or adding to be configurable based on env in `leo.conf`

cpate4 (commented):
Here is our general understanding of the changes:
- `Boot.scala` has been exploded out into Builders
- Routes have been hidden in `HttpRouts.scala`
- The pubsub has been updated to use the patter from `workbench-libs`
- could we get more information re: the `KubernetesInterpreter`?

Other than the clarification questions above, we want to make sure we are doing manual testing on this branch. We are doing some on our side as well. Would you be able to post your testing results (and what you tested) in the #dsp-interactive-analysis channel?

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreterSpec.scala
```diff
@@ -900,9 +900,9 @@ class AKSInterpreterSpec extends AnyFlatSpecLike with TestComponent with Leonard
       override def createAzureClient(cloudContext: AzureCloudContext, clusterName: AKSClusterName)(implicit
         ev: Ask[IO, AppContext]
       ): IO[CoreV1Api] = IO.pure(coreV1Api)
-      override def createGcpClient(clusterId: GKEModels.KubernetesClusterId)(implicit
-        ev: Ask[IO, AppContext]
-      ): IO[CoreV1Api] = IO.pure(coreV1Api)
+//      override def createGcpClient(clusterId: GKEModels.KubernetesClusterId)(implicit
```
jdcanas: is this needed or  can it be deleted?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/AzureDependenciesBuilder.scala
```diff
@@ -0,0 +1,94 @@
...
+import scala.concurrent.ExecutionContext
+
+class AzureDependenciesBuilder extends CloudDependenciesBuilder {
+
+  /**
+   * Registers the OT tracing for the cloud hosting provider.
+   *
+   * @return
+   */
+  // TODO: we must implement this for Azure.
```
jdcanas, line 21: is this TODO relevant? Should we have a ticket?

http/src/main/resources/leo.conf
```diff
@@ -165,6 +165,17 @@ azure {
         token-scope = ${?AZURE_MI_TOKEN_SCOPE}
         token-acquisition-timeout = ${?AZURE_MI_TOKEN_ACQUISITION_TIMEOUT}
       }
+      publisher-config {
+        topic-name = ${?AZURE_SB_TOPIC_NAME}
```
jdcanas, line 169: hmm... where do these values live? I would have expected to find them [here](https://github.com/broadinstitute/terra-helmfile/blob/master/values/app/leonardo/live.yaml.gotmpl#L215)

giventocode: These values are only relevant when deploying on Azure, for the current deployment on GCP these values should not be set. On Azure, Leo will use Azure Service Bus instead of GCP pub/sub as the messaging back end.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/AppDependenciesBuilder.scala
giventocode: yes, we do. As per my understanding this App, which exists on the Azure data plane, requires this configuration, regardless of where Leo is hosted.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/AzureDependenciesBuilder.scala
giventocode: Yes, it is relevant. It requires design work of where to send the metrics when hosting on Azure.  This is follow up we are planning to take as well.  The method should never be called when Leo is hosted in GCP.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/KubernetesInterpreter.scala
giventocode: > is this something we want to take care of now? or later (and if so, should we create a JIRA ticket for it)?

This is somethiing I took from @Qi77Qi previous work. https://github.com/DataBiosphere/leonardo/pull/4162/files#diff-e3fd74d11f3fb501750c9f01302095d8c843dd05cd8985e909f6e6ec34fbe828.  @Qi77Qi , any thoughts here?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/BaselineDependenciesBuilder.scala
giventocode: This is how it is in the current Boot file, and I agree that a conf value makes sense. However, it's out scope for this ticket, but a valid follow up item.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/Boot.scala
giventocode: I think the PR description covers this. Or are you suggesting to add it to the code as well?

LizBaldo (approved):
Thank you for addressing our comments, the clarification on the app dependencies builder is very helpful

LizBaldo:
@giventocode feel free to merge whenever :)

PR #4302, opened 2024-03-07
[WM-2499] Make Leo ECM uri aware

kpierre13:
__Jira ticket__: https://broadworkbench.atlassian.net/browse/[WM-2499]

### What

- Making Leonardo aware of the new ECM dependency in CBAS

### Why

- CBAS integration with ECM service

### What to test

- 

### Who tested and where

- [x] I validated this change in a BEE

[WM-2499]: https://broadworkbench.atlassian.net/browse/WM-2499?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


http/src/main/resources/reference.conf
```diff
@@ -337,6 +337,7 @@ azure {
     ]
     enabled = true
     chart-versions-to-exclude-from-updates = []
+    ecm-base-uri = "https://externalcreds.dsde-dev.broadinstitute.org"
```
salonishah11, line 340: Should this be called `ecm-base-url` since that's what it is called in leo.conf? Or maybe leo.conf one should be called `emc-base-uri`?

kpierre13: It looks like everywhere else was `uri` so I opted to change `leo.conf` instead

cjllanwarne (approved):
Looks correct to me. How will we validate that it actually got through to the apps themselves?

kpierre13:
> Looks correct to me. How will we validate that it actually got through to the apps themselves?

I tested this in my BEE and had Leo point to my terra-helm changes in conjunction with spinning up a new cbas pod containing my custom cbas image. I was able to see the ecm uri populated in the cbas config for my running app.

lucymcnatt (approved):

PR #4307, opened 2024-03-08
RW-12021 set transaction level to serializable

Qi77Qi:
https://precisionmedicineinitiative.atlassian.net/browse/RW-12021

See more details in [this doc](https://docs.google.com/document/d/1c6PBPNN65A-wJnDK26Ksv0l0_MKoOUXwsbBYKqMBtHA/edit#heading=h.r0s506xa7ph) in terms of why we're doing this.

This PR implements [Option 1](https://docs.google.com/document/d/1c6PBPNN65A-wJnDK26Ksv0l0_MKoOUXwsbBYKqMBtHA/edit#heading=h.dmiwwk7ozk2m)

My understanding is `Serielizable` will read-lock the table and does not allow new rows to be inserted until the transaction finishes, hence preventing 2 inserts for the same app.

I'm not exactly sure about performance penalty of this. But since createApp requests don't happen very often. This might be okay. But will observe metrics closely after this PR

### Who tested and where

- [ ] I validated this change 

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/AppUsageComponent.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/AppUsageComponent.scala
@@ -10,6 +10,7 @@ import org.broadinstitute.dsde.workbench.leonardo.db.LeoProfile.mappedColumnImpl
 import org.broadinstitute.dsde.workbench.leonardo.http.dbioToIO
 import org.broadinstitute.dsde.workbench.openTelemetry.OpenTelemetryMetrics
 import org.typelevel.log4cats.Logger
+import slick.jdbc.TransactionIsolation
 
 import java.sql.SQLDataException
 import java.time.Instant
@@ -51,7 +52,7 @@ object appUsageQuery extends TableQuery(new AppUsageTable(_)) {
     } yield id
 
     for {
-      res <- dbio.transaction.attempt
+      res <- dbio.transaction(TransactionIsolation.Serializable).attempt
       id <- res match {
         case Left(e) if e.getMessage.contains("usage startTime was recorded previously") =>
           // We should alert if this happens
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/AppUsageComponent.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/AppUsageComponent.scala
@@ -1,7 +1,7 @@
 package org.broadinstitute.dsde.workbench.leonardo
 package db
 
-import cats.effect.Sync
+import cats.effect.Async
 import cats.implicits._
 import org.broadinstitute.dsde.workbench.leonardo.AppId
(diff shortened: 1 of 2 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/AppUsageComponent.scala
```diff
@@ -51,7 +52,7 @@ object appUsageQuery extends TableQuery(new AppUsageTable(_)) {
     } yield id
 
     for {
-      res <- dbio.transaction.attempt
+      res <- dbio.transaction(TransactionIsolation.Serializable).attempt
```
mlilynolting, line 57: Would it make sense to do this for the other write action, `recordStop`, as well?

mlilynolting (approved):
👍, we could do both write actions but this is the fix

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/AppUsageComponent.scala
Qi77Qi: I don't think we need it in `recordStop` because even if it happens, it'll just mean the same row will be updated twice with very close value...

`TransactionIsolation.Serializable` can cause deadlock, so I think we should limit use of it too

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/AppUsageComponent.scala
```diff
@@ -51,22 +54,33 @@ object appUsageQuery extends TableQuery(new AppUsageTable(_)) {
...
       id <- res match {
         case Left(e) if e.getMessage.contains("usage startTime was recorded previously") =>
           // We should alert if this happens
           metrics.incrementCounter("appStartUsageTimeRecordingFailure") >> logger.error(e)(e.getMessage) >> F
             .raiseError(e)
-        case Left(e)      => F.raiseError(e)
-        case Right(appId) => F.pure(appId)
+        case Left(e: com.mysql.cj.jdbc.exceptions.MySQLTransactionRollbackException)
+            if e.getMessage contains "Deadlock" =>
```
Qi77Qi, line 64: Deadlock happens sometimes as revealed by the unit test I added...
```
{"@timestamp":"2024-03-08T20:09:01.787539Z","@version":"1","logger_name":"org.broadinstitute.dsde.workbench.leonardo.LeonardoTestSuite","thread_name":"io-compute-2","severity":"INFO","serviceContext":null,"message":"Deadlock detected, sleeping for 57 ms"}
```

This adds a bit jitter to recover from deadlock as the unit test demonstrates

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/AppUsageComponent.scala
mlilynolting: makes sense

PR #4310, opened 2024-03-11
[IA-4864] Use UUID for createDisk jobId

lucymcnatt:
__Jira ticket__: https://broadworkbench.atlassian.net/browse/IA-4864

### What

- Fixes bug in https://github.com/DataBiosphere/leonardo/pull/4238 that was cutting off the disk Id, causing a duplicate job id error from WSM

### Why

-

### What to test

- 

### Who tested and where

- [ ] I validated this change 

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
@@ -436,7 +436,7 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
                                                         params.runtime.auditInfo.creator
           )
 
-          createDiskJobId = WsmJobId(s"create-disk-${disk.id.toString.take(10)}")
+          createDiskJobId = WsmJobId(s"${UUID.randomUUID()}")
           jobControl = new JobControl()
             .id(createDiskJobId.value)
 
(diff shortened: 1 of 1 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
```diff
@@ -436,7 +436,7 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
                                                         params.runtime.auditInfo.creator
           )
 
-          createDiskJobId = WsmJobId(s"create-disk-${disk.id.toString.take(10)}")
+          createDiskJobId = WsmJobId(s"${UUID.randomUUID()}")
```
jdcanas: maybe make a follow-up to make this case class take a UUID?

jdcanas (approved):

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
lucymcnatt: yeah that makes sense, I can add it to the [this spike](https://broadworkbench.atlassian.net/browse/IA-4850)

mlilynolting (approved):

PR #4316, opened 2024-03-12
[IA-4858] Add new trigger to the build and publish workflow

LizBaldo:
__Jira ticket__: https://broadworkbench.atlassian.net/browse/IA-4858

### What

- Adds the successful automerge completion as a new trigger to the build and publish workflow 

### Why

- One [clear limitation](https://github.com/broadinstitute/automerge-action?tab=readme-ov-file#limitations) of the automerge action is that when a pull request is merged by this action, the merge will not trigger other GitHub workflows
- One [workaround](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run) is to use the successful completion of the automerge workflow as a trigger to the build and publish workflow so that any commit automerged to develop are pushed to the dev environment

### What to test

I am not quite sure how to test this without breaking the build and publish GHA 😬 I would appreciate a first pass review at the approach before I start breaking things.

### Who tested and where

- [ ] I validated this change 


.github/workflows/leo-build-tag-publish-and-run-tests.yml
```diff
@@ -38,6 +42,7 @@ jobs:
 
   init-github-context:
     runs-on: ubuntu-latest
+    if: ${{ github.event_name != 'workflow_run' || github.event.workflow_run.conclusion == 'success' }}
```
cpate4, line 45: nice ... you got the `success` criteria, too - 🤓

cpate4 (approved):

jdcanas (approved):
Hmmm.. wasn't super optimistic that [chips PR](https://github.com/DataBiosphere/leonardo/pull/4298) to bump the version would fix the automerge action, but there haven't been any failures since then so that may have done it!

I'm curious if we should have a strategy for what to do when the automerge doesn't run because the tests fail though... I don't think leaving them up indefinitely until human intervention takes place is a good option. 

Between re-running the tests until pass or closing the PRs on test fail, I think I'd prefer the latter since theoretically `eventually` the tests will pass and the version will be bumped sooner rather than later given the volume of these that come through. The assumption there being it doesn't seem like we care too much about every version bump as long as its updated periodically, with important versions being manually shepherded along by the relevant team? 

The alternative being leo gets spammed with PRs seems like a worse option at least.

LizBaldo:
I agree, let me bring these options to the teams having automated PRs on Leo and make sure they are onboard with the idea.

LizBaldo:
@jdcanas I asked the apps team about this and what I heard back was that:
- We should not spend too muchtime fixing this GHA as we are moving away from versioning apps in leo anyways
- If a PR failes due to tests, notify the teams via slack that manual intervention is needed

LizBaldo:
I am going to try this, aka a combination of deploying when the workflow succeeds, and notify the app teams slack channels when it fails. I will likely need a few rubber stamp thumbs in the near future if I made a typo or forgot something 😬

PR #4328, opened 2024-03-13
[IA-4749] Fix unrecoverable app upgrade errors

jdcanas:
This is a draft with the initial code to prevent apps getting stuck in `Updating` or inappropriately reporting an `Error` state.

It does this by delaying the transition to `Updating` until right before the helm calls, and ensures the app only gets into an `Error` state from fatal errors, while all others result in a transition back to `Running` with an error appended to the `AppError` table.

Note the `TODO` comment block that discusses a limitation of this approach. Mainly that if leo restarts after `Updating` transition before a terminal status, the app will be stuck in an unusable status. There will be a follow-on ticket to solve this edge case.

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/LeoPublisher.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/LeoPublisher.scala
@@ -122,7 +122,7 @@ final class LeoPublisher[F[_]](
           appQuery.updateStatus(m.appId, AppStatus.Stopping).transaction
         case m: LeoPubsubMessage.StartAppMessage =>
           appQuery.updateStatus(m.appId, AppStatus.Starting).transaction
-        case m: LeoPubsubMessage.UpdateAppMessage =>
+        case _: LeoPubsubMessage.UpdateAppMessage =>
           F.unit
         case _: LeoPubsubMessage.UpdateDiskMessage =>
           F.unit
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/LeoPublisher.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/LeoPublisher.scala
@@ -122,6 +122,7 @@ final class LeoPublisher[F[_]](
           appQuery.updateStatus(m.appId, AppStatus.Stopping).transaction
         case m: LeoPubsubMessage.StartAppMessage =>
           appQuery.updateStatus(m.appId, AppStatus.Starting).transaction
+        // We update this in backleo to prevent apps getting stuck in updating as much as possible, https://broadworkbench.atlassian.net/browse/IA-4749
         case _: LeoPubsubMessage.UpdateAppMessage =>
           F.unit
         case _: LeoPubsubMessage.UpdateDiskMessage =>
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/LeoPublisher.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/LeoPublisher.scala
@@ -123,7 +123,7 @@ final class LeoPublisher[F[_]](
         case m: LeoPubsubMessage.StartAppMessage =>
           appQuery.updateStatus(m.appId, AppStatus.Starting).transaction
         case m: LeoPubsubMessage.UpdateAppMessage =>
-          appQuery.updateStatus(m.appId, AppStatus.Updating).transaction
+          F.unit
         case _: LeoPubsubMessage.UpdateDiskMessage =>
           F.unit
         case _: LeoPubsubMessage.UpdateRuntimeMessage =>
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
@@ -1401,6 +1401,8 @@ class LeoPubsubMessageSubscriber[F[_]](
           )
       }
 
+      _ <- logger.info(s"here in handle in Leopubsubmessage, ${msg.cloudContext}")
+
       updateApp = msg.cloudContext match {
         case CloudContext.Gcp(_) =>
           gkeAlg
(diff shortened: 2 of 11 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -441,7 +441,14 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
...
 
-      // Update the relay listener deployment
+      // The app is blocked in updating now
+      _ <- appQuery.updateStatus(app.id, AppStatus.Updating).transaction
+
+      // Update the relay listener deployment if the app is not already updating from a message submission
+      // TODO: discuss this check. (if (dbApp.app.status != AppStatus.Updating))
+      // TODO: This might be sufficient, but really, we would need to store >>what version<< the app is being updated to determine if we should pass-through to polling or ignore
+      // TODO: the issue is that if we simply exit out if `Updating` status, then the app can remain stuck in updating if leo restarts during polling
+      // TODO: perhaps monitor at boot should just grab all apps in Updating and transition them to running + save an error to avoid this complexity?
```
jdcanas: this is a discussion for a separate PR, I just want to leave a large TODO section indicating not to merge this PR until the subsequent ticket is written

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
```diff
@@ -955,23 +955,7 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
     for {
       ctx <- ev.ask
       params = UpdateAKSAppParams(appId, appName, appChartVersion, workspaceId, cloudContext)
-      _ <- aksAlgebra.updateAndPollApp(params).adaptError { case e =>
```
jdcanas, line 958: there is no need to `adaptError` on this inner function call and on the outer function's call in the top-level `LeoPubsubMessageSubscriber`.  Consolidated error handling to the top-level call since the requirements are the same for both google and azure app updates, leaving room  here for divergence later if needed

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/LeoPublisher.scala
```diff
@@ -122,8 +122,8 @@ final class LeoPublisher[F[_]](
           appQuery.updateStatus(m.appId, AppStatus.Stopping).transaction
         case m: LeoPubsubMessage.StartAppMessage =>
           appQuery.updateStatus(m.appId, AppStatus.Starting).transaction
-        case m: LeoPubsubMessage.UpdateAppMessage =>
-          appQuery.updateStatus(m.appId, AppStatus.Updating).transaction
+        case _: LeoPubsubMessage.UpdateAppMessage =>
```
LizBaldo, line 126: Nit, but we should probably leave a comment here (maybe linking the Jira ticket) to explain why we differ from the other message types

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -1410,16 +1410,39 @@ class LeoPubsubMessageSubscriber[F[_]](
...
-          false,
-          None,
-          None,
-          None
-        )
-      }
+      _ <- updateApp
+        .handleErrorWith {
+          // Fatal case, polling app liveliness failed
+          // The two fatal cases are included separately, because later we may wish to fail fatally on `HelmException`, but roll back on `AppUpdatePollingException`
```
LizBaldo, line 1416: By fatal here, you mean that the app will transition from `Updating` to `Error` correct?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -442,6 +442,9 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
                                         namespaceName
       )
 
+      // The app is blocked in updating now
```
LizBaldo: What do you mean by blocked? This is just the new spot where the app status transitions from running to updating right?

LizBaldo: Ah, maybe just a nit to match the GKE interpreter `// Change app status to updating`

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreterSpec.scala
```diff
@@ -221,6 +225,62 @@ class AKSInterpreterSpec extends AnyFlatSpecLike with TestComponent with Leonard
       res.unsafeRunSync()(cats.effect.unsafe.IORuntime.global)
     }
 
+  it should s"exit with app in Updating status if error occurs in helm client" in isolatedDbTest {
```
LizBaldo, line 230: I would have expected the app to transition to Error in this case. The only time I see an app getting stuck in `updating` would be when leo restarts during the update, but handling this is outside the scope of the PR

LizBaldo (commented):
The new error handling is really nice. Thanks for tackling that refactoring :)

I only have a question around the app transitioning into error instead of staying in updating when a helm exception is caught.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
jdcanas: Yes, fatal in the context of the app being unusable. Will clarify

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreterSpec.scala
jdcanas: It does transition to error, but the handling (and test) for that is in the `LeoPubsubMessageSubcriber`, as it applies to both google and azure. This test essentially ensures that the backleo code does indeed transition an app to updating as front leo used to at the appropriate point in the flow, and will help ensure we don't break that when we want to layer in rolling back

LizBaldo: Ohhh gotcha, I missed that initially. Thanks! Adding a little comment in the test to that effect could be useful for future readers

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
jdcanas: I wanted to call attention to the fact that if leo restarts after this line, the app is 'unusable'. I'll make it clearer

LizBaldo (approved):
Thanks for addressing all of my comments :)

PR #4344, opened 2024-03-15
[IA-4866] Don't try to delete disk if already deleted

lucymcnatt:
__Jira ticket__: https://broadworkbench.atlassian.net/browse/IA-4866

### What

- When a disk is DELETED, there's no need to pass the diskId to back leo to delete

### Why

- Runtime deletion errored when it was trying to delete a disk that was already deleted

### What to test

- 

### Who tested and where

- [ ] I validated this change 


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala
```diff
@@ -332,10 +332,10 @@ class RuntimeV2ServiceInterp[F[_]: Parallel](
       diskIdToDeleteOpt <-
         if (deleteDisk) for {
           // check if disk is deletable in WSM if disk is being deleted
-          diskOpt <- persistentDiskQuery.getActiveById(diskId).transaction
```
lucymcnatt, line 335: this is the main change, so we shouldn't get the DiskNotFound error anymore

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala
```diff
@@ -347,11 +347,9 @@ class RuntimeV2ServiceInterp[F[_]: Parallel](
                     DiskCannotBeDeletedWsmException(disk.id, wsmState, disk.cloudContext, ctx.traceId)
                   )
                 _ <- persistentDiskQuery.markPendingDeletion(diskId, ctx.now).transaction
-              } yield if (wsmState.isDeleted) F.pure(none[DiskId]) else Some(diskId)
-            // if disk hasn't been created in WSM, don't pass id to back leo
-            case None => F.pure(none[DiskId])
-          }
-        } yield Some(diskId)
+              } yield if (wsmState.isDeleted) None else Some(diskId)
```
jdcanas, line 350: just make sure the if statement's return here isn't `any`, seems to have shifted from `if (wsmState.isDeleted) F.pure(none[DiskId]) else Some(diskId)` to `if (wsmState.isDeleted) None else Some(diskId)`

jdcanas:
While I agree that the front leo code should debatably be smart enough to realize if the disk is deleted and properly formulate the pubsub message, I also wonder if the backend code should be treating deleting a disk that is already deleted as a successful no-op for idempotency 🤔

jdcanas (approved):
Will thumb pending addressing the two comments I made so I don't hold it up if you're trying to get this in before the weekend

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala
lucymcnatt: yeah before the yield was returning any, but we didn't use the value so it didn't matter, but I fixed it so I could use it (ie its not `any` anymore)

lucymcnatt:
>I also wonder if the backend code should be treating deleting a disk that is already deleted as a successful no-op for idempotency

yeah it would be simple enough to add `Deleted` as a `deletableStatus`, but idk if that's in scope since I'm trying to get this out so I can delete all those errored workspaces from the staging tests

PR #4351, opened 2024-03-18
[IA-4868] Set runtime destroyedDate upon deletion

lucymcnatt:
__Jira ticket__: https://broadworkbench.atlassian.net/browse/IA-4868

### What

- Instead of just setting the status to deleted, we need to complete the deletion when a runtime is deleted (which updates the destroyedDate and runtimeConfig information)

### Why

-

### What to test

- 

### Who tested and where

- [ ] I validated this change 


LizBaldo (approved):
LGTM, so your previous PR was for google disks, and this one covers Azure?

lucymcnatt:
Basically I took a look through Leo to find where `updateStatus` was being used for `DELETED` and this was the only other case I could find --> we should never be using updateStatus when setting something to deleted because it doesn't update the destroyedDate

PR #4357, opened 2024-03-19
[IA-4859] Consolidate polling for Azure Runtime deletion

jdcanas:
Not tested and still need to dig into an  issue with wsm client, but would like a first pass.

Highly recommend looking at `deleteAndPollRuntime` in the current form alongside the diffed view,  otherwise it is a bit noisy


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
```diff
@@ -516,6 +523,150 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
...
 
+  private def monitorDeleteDisk(params: PollDeleteDiskParams)(implicit
+    ev: Ask[F, AppContext]
+  ): F[Unit] =
+    for {
+      ctx <- ev.ask
+      wsmApi <- buildWsmControlledResourceApiClient
+      getWsmJobResult = F.delay(wsmApi.getDeleteAzureDiskResult(params.workspaceId.value, params.jobId.value))
+
+      _ <- F.sleep(
```
LizBaldo, line 536: Do we need an initial delay here? My understanding is that the delete disk job is very quick

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
```diff
@@ -516,6 +523,150 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
...
+        config.deleteVmPollConfig.initialDelay
+      )
+
+      resp <- streamFUntilDone(
+        getWsmJobResult,
+        config.deleteVmPollConfig.maxAttempts,
+        config.deleteVmPollConfig.interval
+      ).compile.lastOrError
+
+      _ <- resp.getJobReport.getStatus match {
```
LizBaldo, line 598: This is the smallest nit ever, but could you keep the order of the statuses the same throughout? e.g. success, then failure then running?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
```diff
@@ -516,6 +523,150 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
...
+    ev: Ask[F, AppContext]
+  ): F[Unit] =
+    for {
+      ctx <- ev.ask
+      wsmApi <- buildWsmControlledResourceApiClient
+      getWsmJobResult = F.delay(
+        wsmApi.getDeleteAzureStorageContainerResult(params.workspaceId.value, params.jobId.value)
+      )
+
+      _ <- F.sleep(
```
LizBaldo, line 634: Same here, I do not think we need a delay because the deletion is so short they used to have a sync endpoint for it

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
```diff
@@ -697,140 +823,155 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
...
-        } yield ()
+          .map(_ => none[PollVmParams])
+      } { resourceId =>
+        for {
+          deleteJobControl <- F.delay(getDeleteControlledResourceRequest)
+          _ <- F.delay(wsmApi.deleteAzureVm(deleteJobControl, msg.workspaceId.value, resourceId.value))
+        } yield Some(PollVmParams(msg.workspaceId, WsmJobId(deleteJobControl.getJobControl.getId), runtime))
       }
+
+      // Delete and poll on all associated resources in WSM, and then mark the runtime as deleted
```
LizBaldo, line 919: I think it would be good to explicitly list out the resource types here:
- Hybrid connection (handled by leo directly not wsm)
- Storage container
- Disk
- VM
- other?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
```diff
@@ -697,140 +823,155 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
...
+        )
+        _ <- deleteHybridConnectionAction
+
+        _ <- logger.info(ctx.loggingCtx)(
+          s"beginning azure disk deletion and polling for runtime ${msg.runtimeId} in workspace ${msg.workspaceId}"
+        )
+
+        // if Some(leodisk) None(wsmDisk) -> mark deleted
+        // if Some(leodisk) Some(wsmDisk) -> poll then mark deleted
+        // if none(leodisk) Some(wsmDisk) -> do not poll, user wants to keep disk, noop
```
LizBaldo, line 945: 👍

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandlerSpec.scala
```diff
@@ -695,8 +597,11 @@ class AzurePubsubHandlerSpec
           diskStatus = diskStatusOpt.get
           isAttached <- persistentDiskQuery.isDiskAttached(disk.id).transaction
         } yield {
-          verify(mockWsmDao, times(1)).deleteStorageContainer(any[DeleteWsmResourceRequest], any[Authorization])(
-            any[Ask[IO, AppContext]]
+          verify(mockControlledResourceApi, times(1)).getDeleteAzureStorageContainerResult(any[UUID], any[String])
+          verify(mockControlledResourceApi, times(1)).getDeleteAzureVmResult(any[UUID], any[String])
+          verify(mockControlledResourceApi, times(0)).deleteAzureDisk(any[DeleteControlledAzureResourceRequest],
```
LizBaldo, line 602: 👍

LizBaldo (approved):
Looks great to me, I only have a couple of very minor comments, feel free to ignore if you disagree :)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
jdcanas: Its only 5 seconds, I think its fine to have since its async and keeps the configs and format for polling monitoring the same class.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
```diff
@@ -516,6 +523,152 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
...
+      )
+
+      resp <- streamFUntilDone(
+        getWsmJobResult,
+        config.deleteDiskPollConfig.maxAttempts,
+        config.deleteDiskPollConfig.interval
+      ).compile.lastOrError
+
+      _ <- resp.getJobReport.getStatus match {
+        case JobReport.StatusEnum.SUCCEEDED =>
```
lucymcnatt, line 547: do we want to do a DB update in this case, set the disk to deleted?

lucymcnatt (approved):
small comment, but otherwise looks great!

PR #4371, opened 2024-03-22
[IA-4861] Move storage container creation to async task

lucymcnatt:

### What

- Moves storage container and hybrid connection creation to the async task for runtime creation
- Also switched to use the WSM client

### Why

-

### What to test

- 

### Who tested and where

- [ ] I validated this change 


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
```diff
@@ -339,33 +343,47 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
...
-      resp <- wsmDao.createStorageContainer(
-        CreateStorageContainerRequest(
-          params.workspaceId,
-          storageContainerCommonFields,
-          StorageContainerRequest(stagingContainerName)
-        ),
-        auth
+      wsmApi <- buildWsmControlledResourceApiClient
+      common = getCommonFieldsForWsmGeneratedClient(ControlledResourceName(stagingContainerName.value),
+                                                    "leonardo staging bucket",
```
jdcanas, line 337: can we change this to storage container for consistency? staging bucket is a GCP term

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
```diff
@@ -339,33 +343,47 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
     ev: Ask[F, AppContext]
   ): F[(ContainerName, WsmControlledResourceId)] = {
     val stagingContainerName = ContainerName(s"ls-${params.runtime.runtimeName.asString}")
```
jdcanas: nit: storageContainer foor variable consistency

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
```diff
@@ -339,33 +343,47 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
...
       )
+      azureStorageContainer = new AzureStorageContainerCreationParameters()
+        .storageContainerName(stagingContainerName.value)
+
+      request = new CreateControlledAzureStorageContainerRequestBody()
+        .common(common)
+        .azureStorageContainer(azureStorageContainer)
+      storageContainer <- F
+        .delay(wsmApi.createAzureStorageContainer(request, params.workspaceId.value))
+        .handleErrorWith(err =>
```
jdcanas: I don't think this is correct for two reasons:
-  `handleErrorWith` in the `LeoPubsubMessageSubscriber` around `createAndPollRuntime` call will already handle any errors synchronous to the function call with the handler `handleAzureRuntimeCreationError`. If you really want to, you can `adaptError` here and make the error message `"Wsm createStorageContainer failed due to ${err.getMessage}"`, but that's not strictly necessary since the error message from WSM _should_ have information on what call failed. In other words, neither the `handleErrorWith`  or the `F.raiseError` are necessary here, but I could see an argument for an `adaptError`
- Shouldn't we be polling for the storage container result in the async task formed in `monitorCreateRuntime`, as this TODO states: https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala#L174

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
```diff
@@ -158,25 +161,26 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
...
       }
+
+      // These resources (relay connection, staging container) must be created before the VM
       hcName = RelayHybridConnectionName(params.runtime.runtimeName.asString)
       primaryKey <- azureRelay.createRelayHybridConnection(params.landingZoneResources.relayNamespace,
                                                            hcName,
                                                            cloudContext
       )
-
-      // TODO, run this in parallel to VM creation polling in the async task
```
jdcanas, line 174: see other comment, do we not need to poll anymore?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
```diff
@@ -339,33 +343,47 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
     ev: Ask[F, AppContext]
   ): F[(ContainerName, WsmControlledResourceId)] = {
```
jdcanas: nit: I prefer case classes  (i.e. `CreateStorageContainerResult`) over tuples wherever possible since it achieves the following, but not strictly necessary
- Makes the return type to functions stay the same if you want to change what is returned
- Makes accessing the values more clear as opposed to tuple deconstruction syntax

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzureTestUtils.scala
```diff
@@ -51,6 +57,21 @@ object AzureTestUtils extends MockitoSugar {
         .errorReport(new ErrorReport())
     }
 
+    // Create storage container
+    when {
+      api.createAzureStorageContainer(any, ArgumentMatchers.eq(workspaceId.value))
+    } thenAnswer { invocation =>
+      val requestBody = invocation.getArgument[CreateControlledAzureStorageContainerRequestBody](0)
+      new CreatedControlledAzureStorageContainer().resourceId(UUID.randomUUID())
```
jdcanas: unsure why you are retrieving the `requestBody` but then creating a new resourceId here instead of using the one possibly provided in the body? Seems like you can just remove line 64

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzureTestUtils.scala
```diff
@@ -51,6 +57,21 @@ object AzureTestUtils extends MockitoSugar {
...
+    }
+
+    // Create storage container
+    when {
+      api.createAzureStorageContainer(any, ArgumentMatchers.eq(workspaceId2.value))
+    } thenAnswer { _ =>
+      throw new Exception("test exception")
```
jdcanas: probably worth having an argument to the function to indicate whether we want this call to error with the default case being it not throwing an error. I doubt we want all tests that leverage these mocks to have an error in `createAzureStorageContainer`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
lucymcnatt: yeah I see where you're coming from here, this was an attempt to keep the storage container creation out of the async task because the output of `createStorageContainer` is necessary for the createVmRequest --> ie then I'd have to move the createVmRequest into the async as well, but I went ahead and moved all that + creating the hybrid connection into the async task so all the sub-resources could be created there

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzureTestUtils.scala
lucymcnatt: good point, removed

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzureTestUtils.scala
lucymcnatt: yeah I had it only trigger for workspace2, but that's a non-intuitive change, so I can alter that

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandlerAlgebra.scala
```diff
@@ -131,7 +136,16 @@ final case class PollStorageContainerParams(workspaceId: WorkspaceId, jobId: Wsm
 /**
  * This case class represents the necessary information to poll all objects associated with the runtime, namely disk and vm
  */
-final case class CreateRuntimeResourcesResult(vmRequest: CreateVmRequest, createDiskResult: CreateDiskForRuntimeResult)
+final case class CreateRuntimeResourcesResult(vmRequest: CreateVmRequest,
+                                              createDiskResult: CreateDiskForRuntimeResult,
+                                              hybridConnectionName: RelayHybridConnectionName,
+                                              primaryKey: PrimaryKey
```
jdcanas: consider naming this something other than the type, which doesn't tell me what it is the primary key for in this context

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
```diff
@@ -118,131 +120,97 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
...
-   * Disk needs to be created before this function is called
-   * If useExistingDisk is specified, disk is transferred to new runtime and no new disk is created
-   * */
-  private def createRuntime(params: CreateAzureRuntimeParams, jobControl: WsmJobControl)(implicit
-    ev: Ask[F, AppContext]
-  ): F[CreateRuntimeResourcesResult] =
-    for {
-      ctx <- ev.ask
-      auth <- samDAO.getLeoAuthToken
+  private def setupRuntimeCreateMessage(params: PollRuntimeParams,
```
jdcanas: Nice, I like splitting this out into its own method.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
```diff
@@ -118,131 +120,97 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
...
-      runtimeResourcesResult <- createRuntime(
-        CreateAzureRuntimeParams(
+      // send create disk message to WSM
+      createDiskResult <- createDiskForRuntime(
+        CreateAzureDiskParams(msg.workspaceId, runtime, msg.useExistingDisk, azureConfig)
+      )
+
+      // all other resources (hybrid connection, storage container, vm)
+      // are created within the async task
+      _ <- monitorCreateRuntime(
```
jdcanas, line 141: nit: might be worth looking into `PollRuntimeParams` case class and seeing if it can be simplified at all with this refactoring. For example, `createVmJobId` might only be used in the `monitorRuntimeCreate` function right now, and be extra noise to create here and pass along.

See comment near the new `setupRuntimeCreateMessage` first though

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
jdcanas: nit: might be worth adding the arguments below to the params object to keep the method signatures concise. Similar and a tad contrary to the prior comment about `PollRuntimeParams`, but with the common thread of simplifying. You may have already done this, just a call out.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
```diff
@@ -342,37 +310,40 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
...
-  ): F[(ContainerName, WsmControlledResourceId)] = {
-    val stagingContainerName = ContainerName(s"ls-${params.runtime.runtimeName.asString}")
-    val storageContainerCommonFields = getCommonFields(
-      ControlledResourceName(s"c-${stagingContainerName.value}"),
-      "leonardo staging bucket",
-      params.runtime.auditInfo.creator,
-      None
-    )
+  ): F[CreateStorageContainerResourcesResult] = {
+    val stagingContainerName = ContainerName(s"ls-${runtime.runtimeName.asString}")
```
jdcanas: nit: storage-> staging

jdcanas (approved):
few nits, overall looks good

PR #4373, opened 2024-03-22
[IA-4749] Fix upgrade app error handling

jdcanas:
The unit tests didn’t catch this sneaky issue because the F containing updateApp is run twice prior to this PR, once with error handling and once without, and the one without error handling causes the app to enter `Updating` and then never transitions it to error.

Might be worth brainstorming how to unit test this 🤔 ...


LizBaldo (approved):
Oh damned, I would have never caught this during review and got scalaed big time 😱

PR #4374, opened 2024-03-25
WOR-1519: Look up WSM resources by name instead of enumerating

blakery:
__Jira ticket__: https://broadworkbench.atlassian.net/browse/WOR-1519
Supporting changes in WSM: https://github.com/DataBiosphere/terra-workspace-manager/pull/1687

## Summary of changes
* Change instances of resource enumeration to retrieve by name instead

Switches to a new endpoint endpoint to look up resources by name, instead of enumerating all the resources in a workspace.

Using the new endpoint, I was able to get leo to create apps a workspace with an auth domain, in a BEE.
With the first attempt, the app creation still failed, because of a new call to retrieve the workspace description. After pushing a version that didn't have that call, leo was able to successfully create the apps.
Fixing that call will require a separate solution, tracked in this ticket: https://broadworkbench.atlassian.net/browse/WOR-1602

### What

-

### Why

-

### What to test

- 

### Who tested and where

- [ ] I validated this change 

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
@@ -1004,52 +1004,28 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
                                                appType: AppType,
                                                workspaceId: UUID
   ): F[Option[WsmManagedAzureIdentity]] = {
-    val wsmManagedIdentities = F.blocking(
-      resourceApi
-        .enumerateResources(workspaceId, 0, 100, ResourceType.AZURE_MANAGED_IDENTITY, StewardshipType.CONTROLLED)
-        .getResources
-        .asScala
-        .toList
-    )
     val wsmResourceName = generateWsmNameForIdentity(appType)
-    wsmManagedIdentities.map { identities =>
-      // there should be only 1 Azure managed identity per app
-      identities
-        .find(identity => wsmResourceName == identity.getMetadata.getName)
-        .map(identity =>
-          WsmManagedAzureIdentity(wsmResourceName,
-                                  identity.getResourceAttributes.getAzureManagedIdentity.getManagedIdentityName
-          )
-        )
-    }
+    F.blocking(Option(resourceApi.getResourceByName(workspaceId, wsmResourceName)))
+      .map(opt => opt.map(identity => WsmManagedAzureIdentity(wsmResourceName,
+      identity.getResourceAttributes.getAzureManagedIdentity.getManagedIdentityName
+    )))
   }
 
   private[util] def retrieveWsmDatabases(resourceApi: ResourceApi,
                                          databaseNames: Set[String],
                                          workspaceId: UUID
   ): F[List[WsmControlledDatabaseResource]] = {
-    val wsmResourceDatabases = F.blocking(
-      resourceApi
-        .enumerateResources(workspaceId, 0, 100, ResourceType.AZURE_DATABASE, StewardshipType.CONTROLLED)
-        .getResources
-        .asScala
-        .toList
-    )
-
     // TODO: this currently matches on the 'name' (actually type) of database so for example
     // it compares for a 'cbas' or 'cromwellmetadata' database. In the future, their maybe
     // multiple of those in a workspace so this approach will have to be re-considered
     // see https://broadworkbench.atlassian.net/browse/IA-4844
-    wsmResourceDatabases.map { dbs =>
-      dbs
-        .filter(r => databaseNames.contains(r.getMetadata().getName()))
-        .map(r =>
-          WsmControlledDatabaseResource(r.getMetadata().getName(),
-                                        r.getResourceAttributes().getAzureDatabase().getDatabaseName(),
-                                        r.getMetadata().getResourceId
-          )
-        )
-    }
+    F.blocking(
+      databaseNames.flatMap(dbName => Option(resourceApi.getResourceByName(workspaceId, dbName)).map(r =>
+        WsmControlledDatabaseResource(r.getMetadata.getName,
+          r.getResourceAttributes.getAzureDatabase.getDatabaseName,
+          r.getMetadata.getResourceId
+      ))
+    ))
   }
 
   private[util] def getWorkspaceDescription(workspaceApi: WorkspaceApi,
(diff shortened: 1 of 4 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -1098,57 +1083,42 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
...
-        )
-        .map(filtered_ns =>
-          WsmControlledKubernetesNamespaceResource(
-            NamespaceName(filtered_ns.getResourceAttributes.getAzureKubernetesNamespace.getKubernetesNamespace),
-            WsmControlledResourceId(filtered_ns.getMetadata.getResourceId),
-            ServiceAccountName(
-              filtered_ns.getResourceAttributes.getAzureKubernetesNamespace.getKubernetesServiceAccount
-            )
+  ): F[Option[WsmControlledKubernetesNamespaceResource]] = {
+    // The full namespace name will be {namespacePrefix}-{workspaceId},
```
marctalbott, line 1096: Is this a safe assumption to make? Who decides what the namespace name will be?

marctalbott (approved):
One question about how namespace name is determined, but otherwise looks good to me

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
LizBaldo: I think this logic is baked into WSM. It would be nice to at least link the logic where this is defined in WSM. But there is no real contract so if the logic changes this would break.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -1344,6 +1314,19 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
     }
   }
 
+  // This should probably be moved to WsmDao, if HttpWsmDao switches to using the WSM api clients
```
LizBaldo, line 1358: What do you mean by that?

LizBaldo (commented):
LGTM, I think my only ask would be to test the following in a BEE, because we do not have PR integration tests on data plane apps quite yet:
- Create a workspace in the bee ui and make sure the WDS app runs
- Create a Cromwell runner and a workflow shared app in that same workspace (you might need to use swagger for workflow shared)
- Clone the workspace
- Try to create the apps again

That should cover all test cases I think.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
blakery: At current, `HttpWsmDao` doesn't use the swagger-generated WSM clients. It's building paths directly with strings instead. This class already uses the `ResourceApi` though, so it seemed better to continue using the generated client directly, even if it wasn't in the dao. 

If the dao were to start using the published WSM clients, that would be the place where I'd expect this to live, since it is a thin wrapper around the api, with some common error handling.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
blakery: It is indeed baked into WSM, which I agree is a bit scary. 
In principal, this is exactly the sort of assumption contract tests would be useful to test. 

I'll add an addition to the comment to spell out where it's happening.

blakery:
> * Create a workspace in the bee ui and make sure the WDS app runs
> * Create a Cromwell runner and a workflow shared app in that same workspace (you might need to use swagger for workflow shared)
> * Clone the workspace
> * Try to create the apps again
> 
> That should cover all test cases I think.

All of those cases worked in my BEE.

LizBaldo (approved):
Amazing, thank you for addressing all of my comments :) :shipit:

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
cahrens: Is namespacePrefix constrained in length? I'm wondering if we need to worry about possible truncation (in WSM).

cahrens: @blakery or @LizBaldo, do you know about the possible length of namespacePrefix?

LizBaldo: That is a good question. Leo is not creating the namespacePrefix but getting it from WSM via the `AzureKubernetesNamespaceCreationParameters` so I am not sure if there are any constraints there.

blakery: `namespacePrefix` looks to be limited in WSM to 24 characters (`service/build/openapi/src/main/java/bio/terra/workspace/generated/model/ApiAzureKubernetesNamespaceCreationParameters.java`)

I _think_ it's typically the app name: at least that's what it looked like from what I saw in the test data, as well as during some debugging.

blakery:
I'm really confused. 
When the checks were running, merge was disabled. Now, even though `leo-build-tag-publish-and-run-tests / azure-automation-tests` failed, merging is enabled?
Also, `leo-build-tag-publish-and-run-tests / azure-automation-tests` seems to be running against dev (despite involving building a bee), and can be canceled by other runs (which appears to be [why this run failed](https://github.com/DataBiosphere/leonardo/actions/runs/8572259383)).

LizBaldo:
Okay we will look into it, we tried to implement the GHA job concurrency throttling that Jack demoed on Wednesday but it might have backfired a bit

lucymcnatt:
It looks like it did create a BEE with the correct version from your commit to run the tests ([see the output here from the BEE-create step])(https://github.com/broadinstitute/terra-github-workflows/actions/runs/8572168453/job/23494081562#step:3:20) 

but definitely not great that the action was cancelled yet still allows merging...

lucymcnatt:
We merged a revert of the GHA concurrency change, so with the branch updated the tests should run normally now

cahrens (approved):

PR #4376, opened 2024-03-25
IA 4808 update azure dsvm to 24.03.14

erdalcosgun:
__Jira ticket__: https://broadworkbench.atlassian.net/browse/IA-4808

### What

Test updates for DSVMs

### Why

-

### What to test

- 

### Who tested and where

- [ ] I validated this change 


http/src/main/resources/init-resources/azure_vm_init_script.sh
```diff
@@ -4,14 +4,23 @@ set -e
 
 # If you update this file, please update azure.custom-script-extension.file-uris in reference.conf so that Leonardo can adopt the new script
 
+# In order to avoid user id conflicts, change azureuser uid and gid before jupyter or welder users are created
```
LizBaldo: Technically, the welder-user is never created here, and only exist in the welder container, which is why this hack is working :)

LizBaldo: Could you add a line before the welder docker run to explain that the DSVM WORKING_DIR owned by jupyter-user 1001 gets mounted as the Welder container work directory owned by user 1001 (which luckily is welder-user in there, but this logic is brittle and we will fix it later)

LizBaldo (approved):
LGTM, glad this hack works! Feel free to ship it after adding a bit more clarification around why the hack works by working around the docker volume mounting rules

lucymcnatt (approved):

PR #4416, opened 2024-04-03
[IA-4889] debug tunnel

jdcanas:


automation/src/test/resources/startTunnel.sh
```diff
@@ -7,8 +7,14 @@
...
 
+echo "starting tunnel bastion ${BASTION_NAME} with resource id ${RESOURCE_ID} in resource group ${RESOURCE_GROUP} on port ${PORT}"
+
+nohup az network bastion tunnel --name "$BASTION_NAME" --resource-group "$RESOURCE_GROUP" --target-resource-id "${RESOURCE_ID}" --resource-port 22 --port $PORT > /tmp/nohup.out &
+
+echo "displaying nohup output"
+cat /tmp/nohup.out
+echo "done displaying nohup output"
 
-nohup az network bastion tunnel --name "$BASTION_NAME" --resource-group "$RESOURCE_GROUP" --target-resource-id "${RESOURCE_ID}" --resource-port 22 --port $PORT > /dev/null 2>&1 &
```
jdcanas, line 11: before, 
- this reads 'send the stdout of this command into /dev/null' (the ` > /dev/null`), 
- then `2>&1` means 'also send stderr where you sent stdout'.
- the final `&` runs it as a background process (i.e. allowing the script to continue execution)

It should be readable now and also not scarf stderr

LizBaldo (approved):
Nice so no more noise at the beginning of our Azure logs now? 

Also could you link the JIRA ticket before merging?

PR #4429, opened 2024-04-09
RW-12258 bump sas and rstudio chart version

Qi77Qi:
We should probably update welder version for in `reference.conf` as well so that we have welder logs in cloud logging for GCE VMs properly as well. But we can't update it yet because Azure VM uses the same welder hash, and latest welder image isn't in ACR yet because of https://broadworkbench.atlassian.net/browse/IA-4899 (cc @LizBaldo )

__Jira ticket__: https://precisionmedicineinitiative.atlassian.net/browse/RW-12258

### What

-

### Why

-

### What to test

- 

### Who tested and where

- [ ] I validated this change 

### DIFF
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -827,8 +827,8 @@ gke {
     # This is really just a prefix of the chart name. Full chart name for AOU app will be
     # this chartName in config file appended with allowedChartName from createAppRequest
     chartName = "/leonardo/"
-    rstudioChartVersion = "0.4.0"
-    sasChartVersion = "0.10.0"
+    rstudioChartVersion = "0.5.0"
+    sasChartVersion = "0.11.0"
 
     services = [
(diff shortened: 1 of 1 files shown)


yonghaoy (approved):

LizBaldo (commented):
You also need to update the dockerfile so the proper chart get's cached locally https://github.com/DataBiosphere/leonardo/blob/bfec50a1871d3542520799e0876fb8d97b7b7bac/Dockerfile#L35

LizBaldo:
Looks like the PR got merged, but the dockerfile is still pointing to the old chart version. Ping me if you need a quick 👍 in a follow up PR

Qi77Qi:
@LizBaldo yes! thought I pushed the commit...shouldn't try to multi task :) 
https://github.com/DataBiosphere/leonardo/pull/4430

PR #4432, opened 2024-04-09
WM-2596: Log upgrade errors

cjllanwarne:
__Jira ticket__: https://broadworkbench.atlassian.net/browse/WM-2596

## Summary of changes

Logs any errors resulting from failed app upgrades (rolled back before applying any changes)

### Why

The messages written to the Leo DB are insufficient for tracking what actually went wrong during the upgrade. In particular errors such as `{"message":"Resource not found.","statusCode":404,"causes":[]}` cannot be debugged without additional context. This change makes sure the full error and stack trace are in application logs, without changing the short version of the message in the database.

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
@@ -1436,7 +1436,7 @@ class LeoPubsubMessageSubscriber[F[_]](
             for {
               _ <- appQuery.updateStatus(msg.appId, AppStatus.Running).transaction
               errorContext = s"Error updating Azure app with id ${msg.appId.id} and cloudContext ${msg.cloudContext.asString}"
-              _ = logger.warn(ctx.loggingCtx, e)(errorContext)
+              _ <- logger.warn(ctx.loggingCtx, e)(errorContext)
               error = AppError(
                 s"${errorContext}: ${e.getMessage}",
                 ctx.now,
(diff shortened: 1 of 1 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -1435,8 +1435,10 @@ class LeoPubsubMessageSubscriber[F[_]](
           case e =>
             for {
               _ <- appQuery.updateStatus(msg.appId, AppStatus.Running).transaction
+              errorContext = s"Error updating Azure app with id ${msg.appId.id} and cloudContext ${msg.cloudContext.asString}"
+              _ = logger.warn(ctx.loggingCtx, e)(errorContext)
```
jdcanas: this won't log anything, it needs to be `_ <- logger.warn(ctx.loggingCtx, e)(errorContext)`.

cjllanwarne: Interesting, can you explain how this logger works? 

I thought the `warn` return is irrelevant (and presumably not a `F[]` so a flatMap might not even be valid?). I would have thought we're just using it for the side effect which will happen regardless?

jdcanas: since it returns an `F[]`, the log statement won't be executed unless its in the flatmap. It's similar to if you have a database query and are trying to  make it an atomic set of instructions, you need to flatmap over each database transaction `F`. Here an `F[]` is declared but it is never executed, just assigned and left dangling.

If the logger just returned `unit` it would be fine and the side effect would occur immediately

cjllanwarne: Fascinating, thanks! Apparently my preconception ("and presumably not a F[] ...") was completely wrong

cjllanwarne: Now changed

jdcanas (approved):

mspector (approved):

salonishah11 (approved):

PR #4436, opened 2024-04-10
RW-12276 Delete staging bucket when we delete app's disk

Qi77Qi:
https://precisionmedicineinitiative.atlassian.net/browse/RW-12276

We recently noticed that we're leaking staging buckets for SAS and RStudio (we're not cleaning them up when we delete apps).

### What

-

### Why

-

### What to test

- 

### Who tested and where

- [ ] I validated this change 

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
@@ -17,6 +17,7 @@ import org.broadinstitute.dsde.workbench.google2.{
   streamUntilDoneOrTimeout,
   DiskName,
   GoogleDiskService,
+  GoogleStorageService,
   MachineTypeName,
   ZoneName
 }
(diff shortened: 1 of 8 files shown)


LizBaldo (commented):
That makes sense to me, I think my only question is whether we want to persist the bucket containing the welder states when deleing the app without the PD?

Qi77Qi:
@LizBaldo yea, I think that would make sense as well..so basically the staging bucket's lifecycle will be the same as the user's PD, and the bucket will be deleted when the disk gets deleted?

LizBaldo:
I think that's how it should work yeah

Qi77Qi:
@LizBaldo yea I can make that change...it'll probably make sense to do sth similar for jupyter/rstudio too as minor improvements

Qi77Qi:
@LizBaldo  this is ready for a re-review...lmk if u have more comments

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriberSpec.scala
```diff
@@ -1147,7 +1148,7 @@ class LeoPubsubMessageSubscriberSpec
 
   // delete app and not delete disk when specified
   // update app status and disk id
-  it should "delete app and not delete disk when specified" in isolatedDbTest {
+  it should "delete app and not delete disk when diskId not specified" in isolatedDbTest {
```
LizBaldo, line 1151: Would it be possible to also verify that the call to delete the bucket is not made in this case?

LizBaldo (approved):
Thanks for addressing my comment :) LGTM, I only have one question about expanding an existing unit test to make sure we do not delete the bucket when we do not delete the PD

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -1261,7 +1262,41 @@ class LeoPubsubMessageSubscriber[F[_]](
                   }
               else F.unit
 
-            _ <- List(deleteDataDisk, deletePostgresDisk).parSequence_
+            deleteStagingBucket = dbApp.app.appType match {
+              case AppType.Allowed =>
```
jdcanas: hmm... would it be possible to detect whether a staging bucket exists via `getGoogleStorageServiceFromRegistry()` rather than decide based on the appType here?

Qi77Qi: we can just fire a delete, and ignore if it's 404..I don't think doing a GET first is necessary since it'll add one more GCP call for not much benefit.
Calling based on app type is a bit of optimization because we know welder is only used for GCP apps today. But I'm happy to remove this logic if you think we'll likely support welder for other apps

jdcanas: Yeah, if firing a delete if none exist doesn't have a downside as you say, I'd rather not optimize this logic and always try to clean it up to prevent us from accidentally leaking in the future

Qi77Qi:
@jdcanas @LizBaldo lmk if u have more comments

jdcanas (approved):

PR #4442, opened 2024-04-11
[IA-4627] App deletion should succeed when sub-resources dont exist

lucymcnatt:
__Jira ticket__: https://broadworkbench.atlassian.net/browse/IA-4627

### What

- When an app is deleted, many sub-resources (databases, managed identity, namespace) are deleted through WSM. If the underlying resource is deleted in WSM, the delete errors.
- This change handles the error and updates the database when this happens
- Also consolidates the WSM delete functions in the AKS interp so that all resource deletions are handled by the same function and can take advantage of the new error handling

### Why

- If a resource is deleted in WSM, then the deletion has already succeeded and Leo's database should reflect that

### What to test

- 

### Who tested and where

- [ ] I validated this change 

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
@@ -1322,17 +1322,18 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
       _ <- logger.info(ctx.loggingCtx)(
         s"Deleting WSM resource ${wsmResource.resourceId.value} for app ${app.appName.value} in workspace ${workspaceId.value}"
       )
+
+      jobId = UUID.randomUUID()
+      deleteResourceRequest = new DeleteControlledAzureResourceRequest().jobControl(
+        new JobControl().id(jobId.toString)
+      )
       _ <- wsmResource.resourceType match {
         case WsmResourceType.AzureManagedIdentity =>
           F.blocking(wsmApi.deleteAzureManagedIdentity(workspaceId.value, wsmResource.resourceId.value))
-        case WsmResourceType.AzureDatabase | WsmResourceType.AzureKubernetesNamespace =>
-          val jobId = UUID.randomUUID()
-          val deleteResourceRequest = new DeleteControlledAzureResourceRequest().jobControl(
-            new JobControl().id(jobId.toString)
-          )
-          if (wsmResource.resourceType == WsmResourceType.AzureDatabase)
-            deleteAndPollWsmDatabaseResource(workspaceId, app, wsmResource, jobId, deleteResourceRequest, wsmApi)
-          else deleteAndPollWsmNamespaceResource(workspaceId, app, wsmResource, jobId, deleteResourceRequest, wsmApi)
+        case WsmResourceType.AzureDatabase =>
+          deleteAndPollWsmDatabaseResource(workspaceId, app, wsmResource, jobId, deleteResourceRequest, wsmApi)
+        case WsmResourceType.AzureKubernetesNamespace =>
+          deleteAndPollWsmNamespaceResource(workspaceId, app, wsmResource, jobId, deleteResourceRequest, wsmApi)
         case _ =>
           F.raiseError(AppDeletionException(s"Unexpected WSM resource type ${wsmResource.resourceType}"))
       }
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
@@ -1213,42 +1213,22 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
 
   private[util] def deleteAndPollWsmNamespaceResource(workspaceId: WorkspaceId,
                                                       app: App,
-                                                      wsmResource: AppControlledResourceRecord
+                                                      wsmResource: AppControlledResourceRecord,
+                                                      jobId: UUID,
+                                                      deleteResourceRequest: DeleteControlledAzureResourceRequest,
+                                                      wsmApi: ControlledAzureResourceApi
   )(implicit
     ev: Ask[F, AppContext]
   ): F[Unit] =
     for {
       ctx <- ev.ask
 
-      // Build WSM client
-      wsmApi <- buildWsmControlledResourceApiClient
-
-      // Build delete namespace request
-      jobId = UUID.randomUUID()
-      deleteResourceRequest = new DeleteControlledAzureResourceRequest().jobControl(
-        new JobControl().id(jobId.toString)
-      )
-
-      _ <- logger.info(ctx.loggingCtx)(s"WSM delete resource request: ${deleteResourceRequest}")
+      _ <- logger.info(ctx.loggingCtx)(s"WSM delete namespace request: ${deleteResourceRequest}")
 
       // Execute WSM call
-      result <- wsmResource.resourceType match {
-        case WsmResourceType.AzureKubernetesNamespace =>
-          F.blocking(
-            wsmApi.deleteAzureKubernetesNamespace(deleteResourceRequest,
-                                                  workspaceId.value,
-                                                  wsmResource.resourceId.value
-            )
-          )
-        case WsmResourceType.AzureDatabase =>
-          F.blocking(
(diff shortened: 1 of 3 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -1211,9 +1211,9 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
       ServiceAccountName(namespaceAttributes.getKubernetesServiceAccount)
     )
 
-  private[util] def deleteWsmNamespaceResource(workspaceId: WorkspaceId,
-                                               app: App,
-                                               wsmResource: AppControlledResourceRecord
+  private[util] def deleteAndPollWsmNamespaceResource(workspaceId: WorkspaceId,
```
LizBaldo, line 1214: Nit, but this function deletes more than just the namespace, would it make sense to rename it to something like `deleteAndPollWsmAppResources`?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -1338,8 +1305,8 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
       _ <- wsmResource.resourceType match {
         case WsmResourceType.AzureManagedIdentity =>
           F.blocking(wsmApi.deleteAzureManagedIdentity(workspaceId.value, wsmResource.resourceId.value))
-        case WsmResourceType.AzureDatabase =>
-          deleteAzureDatabaseResource(workspaceId, app, wsmResource)
+        case WsmResourceType.AzureDatabase | WsmResourceType.AzureKubernetesNamespace =>
```
LizBaldo: Would it make sense to include the deletion of the managed identity in the `deleteAndPollWsmNamespaceResource` pending renaming of the function per my comment above?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
cpate4: ... this is aligning with the changes to naming in the method, too

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -1225,16 +1225,30 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
...
       )
 
-      _ <- logger.info(ctx.loggingCtx)(s"WSM delete namespace request: ${deleteNamespaceRequest}")
+      _ <- logger.info(ctx.loggingCtx)(s"WSM delete resource request: ${deleteResourceRequest}")
 
       // Execute WSM call
-      result <- F.blocking(
-        wsmApi.deleteAzureKubernetesNamespace(deleteNamespaceRequest, workspaceId.value, wsmResource.resourceId.value)
-      )
+      result <- wsmResource.resourceType match {
```
cpate4: are there other types of `resourceType`'s we should be accommodating here?
this will be SO nice in an event-driven world - 🤓

cpate4: ... dropping reference to [WsmResourceType](https://github.com/databiosphere/leonardo/blob/5409fb53e95f5ef69d88de4da69fca0d6704f597/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/RuntimeControlledResourceComponent.scala#L67-L95) for convenience

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -1225,16 +1225,30 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
...
+            )
+          )
+        case WsmResourceType.AzureDatabase =>
+          F.blocking(
+            wsmApi.deleteAzureDatabaseAsync(deleteResourceRequest, workspaceId.value, wsmResource.resourceId.value)
+          )
+        case _ =>
+          F.raiseError(
+            AppDeletionException(s"Unexpected WSM resource type for deletion polling ${wsmResource.resourceType}")
+          )
```
cpate4: I am curious the likelihood we are going to hit this as this is a new error condition

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -1247,83 +1261,36 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
...
+        case WsmResourceType.AzureDatabase =>
+          F.blocking(
+            wsmApi.getDeleteAzureDatabaseResult(workspaceId.value, jobId.toString)
+          )
+        case _ =>
+          F.raiseError[DeleteControlledAzureResourceResult](
+            AppDeletionException(s"Unexpected WSM resource type for deletion polling ${wsmResource.resourceType}")
+          )
+      }
```
cpate4: in theory, we will never get here because of the earlier default condition.
question ... 
does it make sense to collapse the case conditions and do the two operations side-by-side?
proposing order of operations being ...
- update our database state to DELETING
- then issue commands (make request, and poll) in single case context
- followed by cleanup code below

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
cpate4: +1

cpate4 (approved):
this is / was pretty gnarly, great work on it!
I had a suggestion re: simplifying the logic on the case statement, and also +1 to what @LizBaldo said.

given our current situation with this, code does what was asked and feel free to ignore suggestions (and hence the approval)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
lucymcnatt: ~~good point, I forgot to change the naming here!~~ Now its actually true

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
lucymcnatt: good question, these are the only 2 WSM resources for Apps that we poll on currently, we don't poll on `managedIdentity` so that deletion occurs in the call function, `deleteWsmResource`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
lucymcnatt: we really shouldn't hit this error case because of the case in `deleteWsmResource` where this is called, but the compiler yells at me if I don't include an 'else' case....

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
lucymcnatt: I thought about that, but it makes things a bit complicated because the `managedIdentity` is the only resource we don't poll on, and so we don't need the majority of that function. I condensed the polling WSM resources in the same function to avoid repeated code when generating the jobId, delete request etc, but that has the byproduct of some redundant case statements that you guys pointed out

lucymcnatt: I can try to experiment a bit more with it though

lucymcnatt: decided to pull some of the setup out of the shared function and have a different polling function for each resource, much cleaner this way

LizBaldo: Nit, but would it be a bit easier to read if we had case managed identity, case database and then case namespace and not do the |? It feels a bit redundant

LizBaldo (approved):
LGTM! I only have a small suggestion but that should not be blocking you from merging, thanks :)

PR #4448, opened 2024-04-11
RW-12371 Fix a bug when runtime fail to delete

Qi77Qi:
https://precisionmedicineinitiative.atlassian.net/browse/RW-12371

We had a support ticket when a user attempted to delete a dataproc cluster on 4/4. And leo transitioned the cluster to `ERROR` status because deletion timed out (there was an internal server error on dataproc side). User assumed the cluster is gone, and surprised by the unexpected extra spend when they looked at cloud cost.

This PR makes it so that if an error happens when a cluster is in `Deleting` status, `failedRuntime` function won't mark it as `ERROR`.

__Jira ticket__: https://precisionmedicineinitiative.atlassian.net/browse/RW-12371

### What

-

### Why

-

### What to test

- 

### Who tested and where

- [ ] I validated this change 

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BaseCloudServiceRuntimeMonitor.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BaseCloudServiceRuntimeMonitor.scala
@@ -193,7 +193,7 @@ abstract class BaseCloudServiceRuntimeMonitor[F[_]] {
         )
         _ <- clusterQuery.detachPersistentDisk(runtimeAndRuntimeConfig.runtime.id, ctx.now).transaction
         _ <- curStatus match {
-          case RuntimeStatus.Deleted =>
+          case RuntimeStatus.Deleted | RuntimeStatus.Deleting =>
             logger.info(ctx.loggingCtx)(
               s"failedRuntime: not moving runtime with id ${runtimeAndRuntimeConfig.runtime.id} because it is in ${curStatus} status."
             )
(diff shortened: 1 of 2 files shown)


yonghaoy (approved):
Thanks! I think this week we saw a similar issue in GKE app.

LizBaldo:
We unfortunately do not have access to the JIRA ticket so I am not sure if there is extra context in there. But my question is what advantage of leaving the runtime in `deleting` provides? If the runtime is in an `error` state then the user can try to delete it again, which won't be possible with a deleting runtime. And if the underlying resource has been deleted in the cloud, then the cron job should pick it up either way.

Also if I am slow to respond, feel free to tag the @DataBiosphere/broad-interactive-analysis group on Github so that any of us can see the PR

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/BaseCloudServiceRuntimeMonitor.scala
```diff
@@ -193,7 +193,7 @@ abstract class BaseCloudServiceRuntimeMonitor[F[_]] {
         )
         _ <- clusterQuery.detachPersistentDisk(runtimeAndRuntimeConfig.runtime.id, ctx.now).transaction
         _ <- curStatus match {
-          case RuntimeStatus.Deleted =>
+          case RuntimeStatus.Deleted | RuntimeStatus.Deleting =>
```
lucymcnatt, line 196: my worry with this is that then a runtime will be stuck in deleting, moving the runtime to error allows a user to then delete the runtime

Qi77Qi: that's a good point...maybe we can think about making the error msg a bit better

PR #4473, opened 2024-04-24
IA-4831: use originatingUserEmail instead of userInfo.userEmail

blakery:
__Jira ticket__: https://broadworkbench.atlassian.net/browse/IA-4831
Required for: https://broadworkbench.atlassian.net/browse/WOR-1545

## Summary of changes

Pass in `originatingUserEmail` instead of `userInfo.userEmail` for `AppInstall`, so it can get a pet with the email.
Most places the service was already doing the right thing, but there were a couple of places it wasn't.

### What

-

### Why

-

### What to test
WDS is created in a cloned workspace via async flows (which use a Pet to start WDS)
- 

### Who tested and where

I tested creating a workspace, creating apps, cloning a workspace, deleting a workspace, all in a bee.

- [ ] I validated this change 

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
@@ -780,7 +780,7 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
         getSavableApp(
           cloudContext,
           appName,
-          userInfo.userEmail,
+          originatingUserEmail,
           samResourceId,
           req,
           diskResultOpt.map(_.disk),
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
@@ -732,7 +732,7 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
 
       // Save or retrieve a KubernetesCluster record for the app
       saveCluster <- F.fromEither(
-        getSavableCluster(userInfo.userEmail, cloudContext, ctx.now)
+        getSavableCluster(originatingUserEmail, cloudContext, ctx.now)
       )
       saveClusterResult <- KubernetesServiceDbQueries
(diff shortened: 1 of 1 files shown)


dvoet (approved):

aherbst-broad (approved):

cpate4:
so, the team has been talking about this. there were concerns at first blush that we were discussing and trying to figure out next steps for:

1. this was a change to "get" a different identification for the user, 
2. there are still instances [where `userInfo`](https://github.com/DataBiosphere/leonardo/pull/4473/files#diff-6601f136c46d939b445dae3bcbe164de0c94da97bc87c450e3eea6dc3e9d66fbR763-R776) (in the same method body) is using the `userInfo` object (and if you look deeper, there is an ask to eventually get the `originatingUserEmail` in that method, but ...).
3. no new unit tests were introduced to support this work
4. OPEN QUESTION: Is there a need for data migration? Anywhere? Given that we were using different ids to do things ...

cc: @dvoet, @aherbst-broad, @blakery, @lucymcnatt, @jdcanas, @cindy-broadinstitute

blakery:
> so, the team has been talking about this. there were concerns at first blush that we were discussing and trying to figure out next steps for:
> 
> 1. this was a change to "get" a different identification for the user,
> 2. there are still instances [where `userInfo`](https://github.com/DataBiosphere/leonardo/pull/4473/files#diff-6601f136c46d939b445dae3bcbe164de0c94da97bc87c450e3eea6dc3e9d66fbR763-R776) (in the same method body) is using the `userInfo` object (and if you look deeper, there is an ask to eventually get the `originatingUserEmail` in that method, but ...).
> 3. no new unit tests were introduced to support this work
> 4. OPEN QUESTION: Is there a need for data migration? Anywhere? Given that we were using different ids to do things ...
> 
> cc: @dvoet, @aherbst-broad, @blakery, @lucymcnatt, @jdcanas, @cindy-broadinstitute

1. In the case of a non-pet user making the request, `userInfo.userEmail` will be the same as originatingUserEmail. 
2. The other places I see where `userInfo` is being used look fine to me. The issue isn't the usage of `userInfo` at all, it's that the email is being stored, and then it's used to retrieve a pet account later. If the user is already a pet, that won't work. The other places it's being used in that method, it's doing things like auth checks and the like. 
3. There are _existing_ checks that cover this part of the code (including some that validate the input referenced here). It's notable that this didn't break those, since the values are the same in most cases. 
4. There shouldn't be any data migrations needed. Anything where there's a change in the values being used would have just failed before this.

PR #4475, opened 2024-04-25
TOAZ-357 Open id connect api yaml

dvoet:
https://broadworkbench.atlassian.net/browse/TOAZ-357

use openIdConnect security scheme in open api instead of oauth2 so we can remove the custom implementations of authorize and token endpoints and work with a central swagger ui

### What to test

- 

### Who tested and where

- [ ] I validated this change 

### DIFF
--- a/http/src/main/resources/swagger/api-docs.yaml
+++ b/http/src/main/resources/swagger/api-docs.yaml
@@ -29,8 +29,6 @@ tags:
 security:
   - oidc:
       - openid
-      - email
-      - profile
 
 ##########################################################################################
 ## PATHS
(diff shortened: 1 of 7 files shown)


http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -29,8 +29,6 @@ tags:
 security:
   - oidc:
       - openid
-      - email
-      - profile
```
cpate4, line 33: do we want to remove any of these security providers?, meaning does this affect our ability to use / validate service accounts with google? (think admin apis, among maybe others?)

dvoet: This PR changes what the swagger ui uses for authentication, not what the service actually supports

jdcanas (approved):
azure automation test timeouts are expected currently, investigating

lucymcnatt (approved):

PR #4485, opened 2024-04-26
RW-12082 more metrics change

Qi77Qi:
1. Add `isAoU` label for createApp task
2. Add `cloudService` label for createRuntime task
3. Update existing `pubsub` metrics to use labels so that it's easier to create charts in grafana. I didn't find charts using these metrics, but if there is let me know and happy to update them after this PR...

### What

-

### Why

-

### What to test

- 

### Who tested and where

- [ ] I validated this change 

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
@@ -43,7 +43,7 @@ import org.broadinstitute.dsde.workbench.model.{ErrorReport, TraceId, WorkbenchE
 import org.broadinstitute.dsde.workbench.openTelemetry.OpenTelemetryMetrics
 import org.broadinstitute.dsde.workbench.util2.messaging.{CloudSubscriber, ReceivedMessage}
 import org.broadinstitute.dsp.{ChartVersion, HelmException}
-
+import org.broadinstitute.dsde.workbench.leonardo.db.LabelResourceType
 import java.time.Instant
 import java.util.concurrent.TimeUnit
 import scala.concurrent.ExecutionContext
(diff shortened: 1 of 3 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -227,12 +227,12 @@ class LeoPubsubMessageSubscriber[F[_]](
                                 4.5 minutes
       )
       messageType = event.msg.messageType.asString
-      metricsName = e match {
+      (metricsName, tags) = e match {
         case Some(e) =>
-          s"pubsub/fail/${messageType}/${e.getClass.toString.split('.').last}"
-        case None => s"pubsub/ack/${messageType}"
+          (s"pubsub/fail", Map("messageType" -> messageType, "exception" -> e.getClass.toString.split('.').last))
+        case None => (s"pubsub/success/${messageType}", Map("messageType" -> messageType))
```
Qi77Qi, line 233: @cpate4  I was looking to add this metrics in a chart, but think it'll be easier to do so if the exception info and the messageType are labels instead of having them in metrics name. In grafana, with the current metrics name, we'll need to add multiple queries in order to cover all message types and exceptions. If we use labels instead, we can use just less query depending on how you want to view the data

jdcanas (approved):

cpate4 (approved):

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -1127,13 +1127,15 @@ class LeoPubsubMessageSubscriber[F[_]](
...
+      labels <- labelQuery.getAllForResource(msg.appId.id, LabelResourceType.app).transaction
+      isAoU = labels.get(AOU_UI_LABEL).contains("true")
       _ <- asyncTasks.offer(
         Task(
           ctx.traceId,
           task,
           Some(handleKubernetesError),
           ctx.now,
-          TaskMetricsTags("createApp", Some(resolveAppType(msg.appType, appChart)), None, CloudProvider.Gcp)
+          TaskMetricsTags("createApp", Some(resolveAppType(msg.appType, appChart)), Some(isAoU), CloudProvider.Gcp)
```
cpate4, line 1138: How much do we need `AoU` to be referenced specifically in the codebase? Are these AoU specific features?

Qi77Qi: they're not AoU specific...but these tags make it possible to know whether these data points are AoU or non AoU

cpate4: not for now ... but is there another way of doing this?, besides hardcoding AoU in the codebase?

cpate4: it is something to think about for future our probably near-term future as we refactor this

Qi77Qi: Not sure I understand the question...
the way this is working is all AoU created apps, runtimes, there's a "label" (note this is not metrics label..but Leonardo's label table...there's an equivalent label for terra UI created runtimes/apps too) added for them (they're from UI). And we're using this leonardo label to set the metrics `isAoU` label...

Which part you're thinking of doing differently? (it's probably entirely possible to do it differently)

PR #4164, opened 2024-01-31
DDO-3414 Can I figure out how to deploy a non mainline version of wds to my bee?

mikeflinn:
__Jira ticket__: https://broadworkbench.atlassian.net/browse/[ticket_number]

## Summary of changes

This is my attempt at modifying the values.yaml file that gets passed to wds when leo spins it up in azure to also pass a `wds.image` value so that I can specify a custom version of wds to deploy in a workspace created in my bee without needing to merge a bunch of poc code to mainline.

I have no idea if this is the right way to do this but figured I'd try something before asking for help
### What to test

- 

### Who tested and where

- [ ] This change is covered by automated tests 
  - _NB: Rerun automation tests on this PR by commenting `jenkins retest` or `jenkins multi-test`._
- [ ] I validated this change 
- [ ] Primary reviewer validated this change 
- [ ] I validated this change in the dev environment

PR #4493, opened 2024-05-02
[DO NOT MERGE] Experiment with WDS App Insights metrics reduction

mikeflinn:
This is the same thing that I did in #4164. I have no intention of merging this. I'm just trying to test a custom build of wds in my bee and don't want to have to jump through the hoops of publishing a new helm chart etc.. so I'm doing this instead. I will close this once I'm done testing.

### DIFF
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -386,6 +386,7 @@ azure {
    wds-app-config {
      environment = "dev"
      environment-base = "live"
+     image: "us-central1-docker.pkg.dev/dsp-artifact-registry/mf-dataplane-metrics-wip/wds:insights"
      instrumentation-enabled = false
      chart-name = "terra-helm/wds"
      chart-version = "0.76.0"
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -386,7 +386,7 @@ azure {
    wds-app-config {
      environment = "dev"
      environment-base = "live"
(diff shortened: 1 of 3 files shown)


http/src/main/resources/reference.conf
```diff
@@ -386,6 +386,7 @@ azure {
    wds-app-config {
      environment = "dev"
      environment-base = "live"
+     image: "us-central1-docker.pkg.dev/dsp-artifact-registry/mf-dataplane-metrics-wip/wds:insights"
```
rtitle: Often to avoid the whole Leo dance I just `kubectl edit deploy` on a live WDS/CBAS/etc deployment to change the image and iterate.

mikeflinn: Thanks! yeah I know that's faster. wanted to test something end to end :) With a fresh landing zone and clean deploy of wds. I don't actually need to iterate on WDS itself that much. The change I'm making is a one liner. But it's going to be hard to tell if it did the right thing in pre-existing landing zone with a bunch of other WDSes in it.

mikeflinn: Also apparently I don't have kubectl access to landing zone aks clusters and I'm not sure how give that access to myself. I'm being blocked by deny assignments. I am in the `terra-workspace-dev` AD group

davidangb:
Whenever you return to this PR, note that while you were out we shipped https://github.com/broadinstitute/terra-helmfile/pull/5519, which reduces WDS frequency to 5 minutes

PR #4526, opened 2024-05-09
[IA-4935] Generate azure prod vm password randomly

LizBaldo:

### What

- Replaces the Azure VM password with a randomly generated one in prod

### Why

- Sharing a single password was a high security risk that needs to be mitigated

### What to test

I will test on my BEE, switching the logic to check for the random password there, but when it comes to testing if it works in prod... Well I am not quite sure how I can mock that 😅


http/src/main/resources/leo.conf
```diff
@@ -4,6 +4,7 @@ application {
   leoServiceAccountJsonFile = ${?LEO_SA_JSON_PATH}
   leoServiceAccountEmail = ${?LEO_SVC_ACCT_EMAIL}
   leoUrlBase = ${?LEO_URL_BASE}
+  environment = ${?APP_ENVIRONMENT}
```
LizBaldo, line 7: It was already in the leo helm values 🎉

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
```diff
@@ -164,6 +166,35 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
       )
     } yield ()
 
+  private def generateAzureVMSecurePassword(): String = {
```
LizBaldo: I kinda wanted to add a unit test for this, but not quite sure if this is the best location for this function in the first place. There does not seem to be a kind of `utils` module for leo that i could stick it into

jdcanas: I heard an interesting philosophy recently that `util` files are generally a smell and are an easy way to end up with dependency spaghetti (*points to UI dependency graph which has tons of utils*). Rather, it is better for util methods to live in the file they are consumed, especially if there is only one such file. I think its fine here, and fine to test directly (though you'll need `private[util]`

jdcanas: Also I just realized this is totally in a `util` folder 🤣

jdcanas (approved):
Looks good once unit tested (perhaps assert the azure password requirements and length?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
cpate4: (in the interest of adding to this conversation) - I agree with @jdcanas re: smell (the single file ends up being a dumping ground), and also, I've also read that an alternative is to having a single file is to break out the file into parts and have one "util" aptly named file that reflects the method it represents.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandlerAlgebra.scala
```diff
@@ -144,3 +147,33 @@ final case class AzurePubsubHandlerConfig(samUrl: Uri,
 ) {
   def welderImage: String = s"$welderAcrUri:$welderImageHash"
 }
+object AzurePubsubHandler {
+  private[util] def generateAzureVMSecurePassword(): String = {
+    // Azure is enforcing the following constraints for password generation
+    // Passwords must not include reserved words or unsupported characters.
+    // Password must have 3 of the following: 1 lower case character, 1 upper case character, 1 number, and 1 special character that is not '\'or '-'.
+    // The value must be between 12 and 123 characters long.
```
cpate4, line 155: just confirming that:
- the above is the requirements, and
- we have intentionally settled on 16char for our random password length (vs having random 12-to-123 characters long)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
```diff
@@ -228,6 +228,12 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
       .minorVersionAutoUpgrade(config.runtimeDefaults.customScriptExtension.minorVersionAutoUpgrade)
       .protectedSettings(protectedSettings.asJava)
 
+    // Generate random password for Azure VM in production, for the other lower level envs we can used the shared password
+    val vmPassword = applicationConfig.environment match {
+      case "prod" => AzurePubsubHandler.generateAzureVMSecurePassword()
+      case _      => config.runtimeDefaults.vmCredential.password
+    }
```
cpate4: my two cents ... I would put this logic in a method, and write a test to show `prod` behaves as we expect, and non-prod results in the expected password, and one showing the method is called where we expect it to be (here). 

those would be the only additional tests I would write, as opposed to figuring out integration tests to etc. etc. etc.

mlilynolting (approved):
Looks good to me, with the common-sense unit test :)

cpate4 (approved):
couple knowledge comments and just my two-cents re: testing. integration tests are good, but I am not entirely sure we need one here. happy to support either approach (#LLL - Long Live Leonardo - 🤓)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
LizBaldo: Good point, let me give this a try, thanks!

LizBaldo: Test added, and I was able to use the randomly generated password to create an Azure VM from the portal a couple of times, so we should be good to go

LizBaldo:
Going to force merge, this is an Azure change that has nothing to do with the GKE app creation failure that is a known flake

PR #4534, opened 2024-05-13
RW-12581 Remove user enabled check when redundant

Qi77Qi:
When `cloudContext` is defined, we're already checking if a user has project reader permission, which also makes sure the user is active, so we don't need to check whether user is enabled in these cases.

### What

-

### Why

-

### What to test

- 

### Who tested and where

- [ ] I validated this change 


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskServiceInterp.scala
```diff
@@ -202,13 +202,22 @@ class DiskServiceInterp[F[_]: Parallel](config: PersistentDiskConfig,
...
+            )
+            _ <- F.raiseWhen(!hasProjectPermission)(
+              ForbiddenError(userInfo.userEmail, Some(ctx.traceId))
+            )
+          } yield ()
+        case None =>
+          authProvider.checkUserEnabled(
+            userInfo
+          ) // when request doesn't have cloudContext defined, we check if user is enabled
+      }
```
Qi77Qi, line 220: Moving up the check if `cloudContext` is not defined; otherwise, remove this redundant check

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala
```diff
@@ -778,12 +778,11 @@ class RuntimeV2ServiceInterp[F[_]: Parallel](
 
 }
 
-final case class AuthorizedIds(
-  val ownerGoogleProjectIds: Set[ProjectSamResourceId],
-  val ownerWorkspaceIds: Set[WorkspaceResourceSamResourceId],
-  val readerGoogleProjectIds: Set[ProjectSamResourceId],
-  val readerRuntimeIds: Set[SamResourceId],
-  val readerWorkspaceIds: Set[WorkspaceResourceSamResourceId]
+final case class AuthorizedIds(ownerGoogleProjectIds: Set[ProjectSamResourceId],
```
Qi77Qi, line 781: these `val`s are redundant

mlilynolting (approved):
This makes sense to me!

yonghaoy (approved):

yonghaoy:
Change LG. But I throught all Terra services use the same proxy config, and the proxy will call SAM to check user ToS and enable status already. Why we need this change in Leo again ?

Qi77Qi:
@yonghaoy that's true...

@mlilynolting do you know why we added the explicit user enabled call in the first place?

Qi77Qi:
```
[info]   org.broadinstitute.dsde.workbench.StreamTimeoutError: AppCreationSpec: app terra-quality-93804962/automation-test-app-auymigcfz did not finish creating after 30 minutes
```

jenkins retest

LizBaldo:
@Qi77Qi just FYI. Our integration tests do not run on jenkins anymore. You can retrigger via GHA instead

PR #4539, opened 2024-05-14
[IA-4904] log wording

jdcanas:
I would have really liked to assert that the log message is as expected in a unit test, but unfortunately we cannot test that with mockito since our logger curries the arguements (i.e. `logger.error(arg1, arg2)(arg3)`)

I found some [auxiliary articles ](https://stackoverflow.com/questions/18294830/how-to-mock-a-call-by-name-argument-like-getorelse-using-scalamock) that seem to claim I can use some product format to verify it, but this is further complicated because I cannot simply mock the logger and pass it in since its an implicit declared globally for all tests. I decided it is not worth my time beyond the few hours spent with us trying to move away from leo for app upgrades anyways. 

https://github.com/mockito/mockito/issues/1134

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
@@ -1541,7 +1541,9 @@ class LeoPubsubMessageSubscriber[F[_]](
           case e =>
             for {
               _ <- appQuery.updateStatus(msg.appId, AppStatus.Running).transaction
-              // You cannot update this log wording without updating the corresponding alert here https://console.cloud.google.com/monitoring/alerting/policies/8184448493858086363?project=broad-dsde-prod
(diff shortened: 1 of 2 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -1541,6 +1541,7 @@ class LeoPubsubMessageSubscriber[F[_]](
           case e =>
             for {
               _ <- appQuery.updateStatus(msg.appId, AppStatus.Running).transaction
+              // You cannot update this log wording without updating the corresponding alert here https://console.cloud.google.com/monitoring/alerting/policies/8184448493858086363?project=broad-dsde-prod
```
LizBaldo: Thanks for giving the until testing a try, I agree this will suffice for now. 
One quit nit thought, could you update the comment to ask to update the log alert in staging as well?

LizBaldo (approved):
Thank you!

PR #4542, opened 2024-05-14
[IA-4932] Do not use random shuffle when generating passwords

LizBaldo:
__Jira ticket__: https://broadworkbench.atlassian.net/browse/IA-4932

### What

- Reinforces the logic to generate the Azure VM random passwords based on recommendation from @sarahgibs 

### Why

- Fast follow from #4526 that generated the password using Java's normal Random number generator, which might be predictable

### What to test

- 

### Who tested and where

- [x] I validated this change 

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandlerAlgebra.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandlerAlgebra.scala
@@ -23,7 +23,6 @@ import org.http4s.Uri
 
 import java.security.SecureRandom
 import java.time.Instant
-import scala.util.Random
 
 trait AzurePubsubHandlerAlgebra[F[_]] {
 
@@ -148,7 +147,7 @@ final case class AzurePubsubHandlerConfig(samUrl: Uri,
   def welderImage: String = s"$welderAcrUri:$welderImageHash"
 }
 object AzurePubsubHandler {
-  private[util] def generateAzureVMSecurePassword(): String = {
+  private[util] def generateAzureVMSecurePassword(passwordLength: Int): String = {
     // Azure is enforcing the following constraints for password generation
     // Passwords must not include reserved words or unsupported characters.
     // Password must have 3 of the following: 1 lower case character, 1 upper case character, 1 number, and 1 special character that is not '\'or '-'.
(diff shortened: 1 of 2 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandlerAlgebra.scala
```diff
@@ -168,19 +167,31 @@ object AzurePubsubHandler {
...
+          specialChars.contains
+        )
+
+    }
+    password
   }
 
   private[util] def getAzureVMSecurePassword(environment: String, sharedPassword: String): String =
     // Generate random password for Azure VM in production, for the other lower level envs we can used the shared password
+    // The password must be between 12 and 123 characters long. We are choosing 25 here
```
jdcanas, line 192: sounds like this should be a random range rather than a hardcoded constant as far as security is concerned....

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandlerAlgebra.scala
```diff
@@ -168,19 +167,31 @@ object AzurePubsubHandler {
...
+
+    while (isPasswordValid == false) {
+      password = pickRandomChars(fullCharset, passwordLength).mkString
+      isPasswordValid =
+        password.exists(_.isLower) && password.exists(_.isUpper) && password.exists(_.isDigit) && password.exists(
+          specialChars.contains
+        )
+
+    }
+    password
```
jdcanas, line 187: hmm... can we do some testing about how many loops on average this takes? Dislike while loops a lot 😬

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandlerAlgebra.scala
LizBaldo: I will let @sarahgibs comment on that one, my understanding was that the weakness was coming from how I was reshuffling the characters after selecting them, not how long the overall password was, but I am happy to randomly generate the length of the password too

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandlerAlgebra.scala
jdcanas: Perhaps an approach like this, where we divide the number of chars into randomly sized segments based on the required categories so no looping is required: https://stackoverflow.com/questions/19743124/java-password-generator?

LizBaldo: Agreed, this is why I wrote the unit test using a length of 4, for that it takes on average about 10 loops. With a password length of 25 I never managed to even trigger the while loop in the first place

jdcanas: thats fine then I guess, not worth spending too long as long as security team is happy...

LizBaldo: I will follow up with Sarah tomorrow. What you suggested is very close to what I initially implemented, but without the need for Java random.shuffle so it might be a good compromise, I really dislike while loops too!

sarahgibs: With a length of 25 I would expect it to not have to loop very often. This method is definitely more secure, and I would only want to consider less secure options if it had a performance impact.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandlerAlgebra.scala
sarahgibs: A set length can be fine if it's long enough. You're trying to avoid predictability of characters and brute forcing. Secure random helps avoid predictability, and the length prevents brute forcing.

LizBaldo:
Thanks both for your input! If the current approach works for you, could you approve? That way I can do a fast follow merge after #4526 gets to prod today

sarahgibs (approved):

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandlerAlgebra.scala
mlilynolting: yeah, i don't think we need to worry too much about the chance of flipping 10 heads in a row; like, it's possible from a probability standpoint but,

jdcanas (approved):

LizBaldo:
Going to admin merge this, as Azure automation tests are passing

PR #4575, opened 2024-05-20
[RW-11461] UpdateAppConfig to change autodelete settings

jmthibault79:
__Jira ticket__: https://precisionmedicineinitiative.atlassian.net/browse/RW-11461 (requires PMI-OPS Jira access)

Includes the work from the previous draft #4232 

## Summary of changes

Adds a new endpoint "Update App Config" `PATCH /api/google/v1/apps/{googleProject}/{appName}` which allows users to modify the two autodelete app config fields: `autodeleteEnabled` and `autodeleteThreshold`.  Requests can contain changes to one field or both (or neither) as long as (a) any specified threshold is a positive integer and (b) any requests which enable autodelete without specifying the threshold correspond to apps with pre-existing valid thresholds in the DB.

More details here: https://docs.google.com/document/d/1eWfY8PQoDKVOdIeREuEC9ObLhXlsTtG2ugLwuqov81k/edit#bookmark=id.n9sphs8lpm4l

### What

-

### Why

- AoU RWB users want the ability to modify and enable/disable the autodelete settings of running apps.  Currently they must delete the apps instead, creating them again with the desired settings

## Testing these changes

I deployed this change to a BEE and created a Galaxy app in the BEE's Terra UI.  I then observed that the app state (as seen via GET app) responded to my PATCH changes as expected. 

### What to test

- 

### Who tested and where

- [ ] I validated this change 

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/appRoutesModels.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/appRoutesModels.scala
@@ -39,8 +39,7 @@ final case class CreateAppRequest(kubernetesRuntimeConfig: Option[KubernetesRunt
                                   autodeleteEnabled: Option[Boolean]
 )
 
-final case class UpdateAppRequest(autodeleteThreshold: Option[Int],
-                                  autodeleteEnabled: Option[Boolean])
+final case class UpdateAppConfigRequest(autodeleteEnabled: Option[Boolean], autodeleteThreshold: Option[Int])
 
 final case class GetAppResponse(
   workspaceId: Option[WorkspaceId],
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/appRoutesModels.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/appRoutesModels.scala
@@ -39,7 +39,7 @@ final case class CreateAppRequest(kubernetesRuntimeConfig: Option[KubernetesRunt
                                   autodeleteEnabled: Option[Boolean]
 )
 
-final case class UpdateAppConfigRequest(autodeleteEnabled: Option[Boolean], autodeleteThreshold: Option[Int])
+final case class UpdateAppRequest(autodeleteEnabled: Option[Boolean], autodeleteThreshold: Option[Int])
 
 final case class GetAppResponse(
   workspaceId: Option[WorkspaceId],
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/appRoutesModels.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/appRoutesModels.scala
@@ -39,6 +39,9 @@ final case class CreateAppRequest(kubernetesRuntimeConfig: Option[KubernetesRunt
                                   autodeleteEnabled: Option[Boolean]
 )
 
+final case class UpdateAppRequest(autodeleteThreshold: Option[Int],
+                                  autodeleteEnabled: Option[Boolean])
+
 final case class GetAppResponse(
   workspaceId: Option[WorkspaceId],
   appName: AppName,
--- a/http/src/main/resources/swagger/api-docs.yaml
+++ b/http/src/main/resources/swagger/api-docs.yaml
@@ -1596,8 +1596,8 @@ paths:
                 $ref: "#/components/schemas/ErrorReport"
     patch:
         summary: Updates the configuration of an app
-        description: In order to update the configuration of an app, it must first be ready
-        operationId: updateApp
+        description: Updates the configuration of an app.  Currently limited to autodeletion config.
+        operationId: updateAppConfig
         tags:
             - apps
         parameters:
@@ -1614,10 +1614,10 @@ paths:
             schema:
                 type: string
         requestBody:
-            $ref: "#/components/requestBodies/UpdateAppRequest"
+            $ref: "#/components/requestBodies/UpdateAppConfigRequest"
         responses:
           "202":
-            description: App update request accepted
+            description: App configuration update request accepted
           "400":
             description: Bad Request
             content:
@@ -2950,14 +2950,14 @@ components:
             machineSize: "Standard_DS1_v2"
             disk: { labels: {}, name: "disk1", size: 50 }
             autopauseThreshold: 15
-    UpdateAppRequest:
+    UpdateAppConfigRequest:
       content:
         application/json:
           schema:
-            $ref: "#/components/schemas/UpdateAppRequest"
+            $ref: "#/components/schemas/UpdateAppConfigRequest"
           example:
-            authdeleteThrehold: 15
-            authdeleteThreholdEnabled: true
+            autodeleteEnabled: true
+            autodeleteThreshold: 60
     UpdateAppsRequest:
       content:
         application/json:
@@ -4154,17 +4154,17 @@ components:
           type: boolean
           description: Whether to turn on autodelete
 
-    UpdateAppRequest:
-      description: a request to update an app
+    UpdateAppConfigRequest:
+      description: a request to update the configuration of an app
       type: object
       properties:
+        autodeleteEnabled:
+          type: boolean
+          description: Whether to turn on autodelete
         autodeleteThreshold:
           type: integer
           description: The number of minutes of idle time to elapse before the app is
             deleted in minute. When autodeleteEnabled is true, a positive integer is required
-        autodeleteEnabled:
-          type: boolean
-          description: Whether to turn on autodelete
 
     UpdateAppsRequest:
       description: a request to update a specific set of apps (v1 or v2)
--- a/http/src/main/resources/swagger/api-docs.yaml
+++ b/http/src/main/resources/swagger/api-docs.yaml
@@ -1594,6 +1594,42 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/ErrorReport"
+    patch:
+        summary: Updates the configuration of an app
+        description: In order to update the configuration of an app, it must first be ready
+        operationId: updateApp
+        tags:
+            - apps
+        parameters:
+          - in: path
+            name: googleProject
+            description: googleProject
+            required: true
+            schema:
+                type: string
+          - in: path
+            name: appName
+            description: appName
+            required: true
+            schema:
+                type: string
+        requestBody:
+            $ref: "#/components/requestBodies/UpdateAppRequest"
+        responses:
+          "202":
+            description: App update request accepted
+          "400":
+            description: Bad Request
+            content:
+              application/json:
+                schema:
+                    $ref: "#/components/schemas/ErrorReport"
+          "500":
+            description: Internal Error
+            content:
+              application/json:
+                schema:
+                    $ref: "#/components/schemas/ErrorReport"
     delete:
       summary: Deletes an existing app in the given project
       description: deletes an App
@@ -2914,6 +2950,14 @@ components:
             machineSize: "Standard_DS1_v2"
             disk: { labels: {}, name: "disk1", size: 50 }
             autopauseThreshold: 15
+    UpdateAppRequest:
+      content:
+        application/json:
+          schema:
+            $ref: "#/components/schemas/UpdateAppRequest"
+          example:
+            authdeleteThrehold: 15
+            authdeleteThreholdEnabled: true
     UpdateAppsRequest:
       content:
         application/json:
@@ -4106,6 +4150,18 @@ components:
           type: integer
           description: The number of minutes of idle time to elapse before the app is
             deleted. When autodeleteEnabled is true, a positive integer is required
+        autodeleteEnabled:
+          type: boolean
+          description: Whether to turn on autodelete
+
+    UpdateAppRequest:
+      description: a request to update an app
+      type: object
+      properties:
+        autodeleteThreshold:
+          type: integer
+          description: The number of minutes of idle time to elapse before the app is
+            deleted in minute. When autodeleteEnabled is true, a positive integer is required
         autodeleteEnabled:
           type: boolean
           description: Whether to turn on autodelete
--- a/http/src/main/resources/swagger/api-docs.yaml
+++ b/http/src/main/resources/swagger/api-docs.yaml
@@ -4155,16 +4155,22 @@ components:
           description: Whether to turn on autodelete
 
     UpdateAppConfigRequest:
-      description: a request to update the configuration of an app
+      description: A request to update the configuration of an app
       type: object
       properties:
         autodeleteEnabled:
           type: boolean
-          description: Whether to turn on autodelete
+          description: > 
+            Optional: Whether to turn on autodelete.  When autodeleteEnabled is set to true but there is no
+            autodeleteThreshold supplied in this request, the existing app's configuration will be used.  If the 
+            existing configuration does not have an autodeleteThreshold with a positive integer, this would result in 
+            an invalid autodelete configuration, and therefore this update request would be rejected.
         autodeleteThreshold:
           type: integer
-          description: The number of minutes of idle time to elapse before the app is
-            deleted in minute. When autodeleteEnabled is true, a positive integer is required
+          description: >
+            Optional: The number of minutes of idle time to elapse before the app is deleted.  This must be a positive
+            integer.  Note: deletion timing is not exact.  The app will be deleted a short period of time after this
+            threshold time has elapsed.
 
     UpdateAppsRequest:
       description: a request to update a specific set of apps (v1 or v2)
--- a/http/src/main/resources/swagger/api-docs.yaml
+++ b/http/src/main/resources/swagger/api-docs.yaml
@@ -1597,7 +1597,7 @@ paths:
     patch:
         summary: Updates the configuration of an app
         description: Updates the configuration of an app.  Currently limited to autodeletion config.
-        operationId: updateAppConfig
+        operationId: updateApp
         tags:
             - apps
         parameters:
@@ -1614,7 +1614,7 @@ paths:
             schema:
                 type: string
         requestBody:
-            $ref: "#/components/requestBodies/UpdateAppConfigRequest"
+            $ref: "#/components/requestBodies/UpdateAppRequest"
         responses:
           "202":
             description: App configuration update request accepted
@@ -2950,11 +2950,11 @@ components:
             machineSize: "Standard_DS1_v2"
             disk: { labels: {}, name: "disk1", size: 50 }
             autopauseThreshold: 15
-    UpdateAppConfigRequest:
+    UpdateAppRequest:
       content:
         application/json:
           schema:
-            $ref: "#/components/schemas/UpdateAppConfigRequest"
+            $ref: "#/components/schemas/UpdateAppRequest"
           example:
             autodeleteEnabled: true
             autodeleteThreshold: 60
@@ -4154,7 +4154,7 @@ components:
           type: boolean
           description: Whether to turn on autodelete
 
-    UpdateAppConfigRequest:
+    UpdateAppRequest:
       description: A request to update the configuration of an app
       type: object
       properties:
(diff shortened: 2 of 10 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/AppRoutes.scala
```diff
@@ -167,6 +175,25 @@ class AppRoutes(kubernetesService: AppService[IO], userInfoDirectives: UserInfoD
...
+  )(implicit
+    ev: Ask[IO, AppContext]
+  ): IO[ToResponseMarshallable] =
+    for {
+      ctx <- ev.ask[AppContext]
+      apiCall = kubernetesService.updateApp(userInfo, CloudContext.Gcp(googleProject), appName, req)
+      _ <- foldSpan(ctx, apiCall, "updateApp")
+    } yield StatusCodes.Accepted
+
+  // TODO: this exact pattern is repeated over 20x, and I have no idea what it does.  My intuition is that it's not
```
jmthibault79, line 191: I'm not planning to commit `foldSpan()` as-is.  Instead I would like a discussion about your norms / code standards.  Based on that, I'll either remove this or apply it more generally (perhaps in a different PR)

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterpSpec.scala
```diff
@@ -3107,7 +3107,7 @@ class AppServiceInterpTest extends AnyFlatSpec with AppServiceInterpSpec with Le
     thrown.appType shouldBe AppType.HailBatch
   }
 
-  "checkIfAppCreationIsAllowedAndIsAoU" should "enable IntraNodeVisibility if customApp check is disabled" in {
+  "checkIfAppCreationIsAllowed" should "enable IntraNodeVisibility if customApp check is disabled" in {
```
jmthibault79, line 3110: not related to this PR, but I happened to notice that this did not match the method name

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
```diff
@@ -669,6 +669,56 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
...
+      appOpt <- KubernetesServiceDbQueries
+        .getActiveFullAppByName(cloudContext, appName)
+        .transaction
+      appResult <- F.fromOption(
+        appOpt,
+        AppNotFoundException(cloudContext, appName, ctx.traceId, "No active app found in DB")
+      )
+
+      tags = Map("appType" -> appResult.app.appType.toString)
+      _ <- metrics.incrementCounter("updateAppConfig", 1, tags)
```
jmthibault79: I noticed that the counter is sometimes incremented at the Service level and sometimes at the Routes level.  Unclear which to choose.

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/appRoutesModels.scala
```diff
@@ -39,6 +39,8 @@ final case class CreateAppRequest(kubernetesRuntimeConfig: Option[KubernetesRunt
                                   autodeleteEnabled: Option[Boolean]
 )
 
+final case class UpdateAppConfigRequest(autodeleteEnabled: Option[Boolean], autodeleteThreshold: Option[Int])
```
Qi77Qi: I think it'll be helpful to lay out all the permutations here. And what assumptions/defaults we're making in the code.
Like if both are none, and DB has `autodeleteEnabled` being `true`, does that mean we're disabling auto delete?

jmthibault79:
Moving back to Draft while I have a think about logic/semantics

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/appRoutesModels.scala
jmthibault79: this is now explicitly laid out in validateUpdateAppConfigRequest()

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/AppRoutes.scala
```diff
@@ -167,6 +176,30 @@ class AppRoutes(kubernetesService: AppService[IO], userInfoDirectives: UserInfoD
...
+                                          appName: AppName,
+                                          req: UpdateAppConfigRequest
+  )(implicit ev: Ask[IO, AppContext]): IO[ToResponseMarshallable] =
+    for {
+      _ <- foldSpan(kubernetesService.updateAppConfig(userInfo, CloudContext.Gcp(googleProject), appName, req),
+                    "updateAppConfig"
+      )
+    } yield StatusCodes.Accepted
+
+  // TODO: this pattern is repeated over 20x
```
jmthibault79: I'm not planning to commit foldSpan() as-is. Instead I would like a discussion about whether this sort of thing is desired.  Happy to remove if not.

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/appRoutesModels.scala
jmthibault79: This conversation inspired this section of the design doc https://docs.google.com/document/d/1eWfY8PQoDKVOdIeREuEC9ObLhXlsTtG2ugLwuqov81k/edit#bookmark=id.n9sphs8lpm4l

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -1594,6 +1594,42 @@ paths:
...
+        tags:
+            - apps
+        parameters:
+          - in: path
+            name: googleProject
+            description: googleProject
+            required: true
+            schema:
+                type: string
+          - in: path
```
LizBaldo, line 1610: Nit, but are appName unique within a project? Or should this be using AppID?

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -4110,6 +4154,18 @@ components:
           type: boolean
           description: Whether to turn on autodelete
 
+    UpdateAppConfigRequest:
```
LizBaldo: Should the valid combination be explained here as well since this is the user facing documentation? Aka users won't go into the leo code to look for it

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/AppRoutes.scala
LizBaldo: I personally like it, making the file DRYer while keeping the logic close enough for comprehension

http/src/main/resources/swagger/api-docs.yaml
LizBaldo: Ah nvm, looks like everything is using the name already 👍

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
```diff
@@ -669,6 +669,80 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
             .raiseError[Unit](AppNotFoundByWorkspaceIdException(workspaceId, appName, ctx.traceId, "permission denied"))
     } yield GetAppResponse.fromDbResult(app, Config.proxyConfig.proxyUrlBase)
 
+  // Autodeletion config logic requires a positive integer for threshold if enabled is true.
+  // For simplicity we can enforce this by rejecting the following scenarios:
+  // 1. The request includes a threshold which is invalid
```
LizBaldo: Can you define what an invalid threshold is here?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
```diff
@@ -669,6 +669,80 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
...
+      appOpt <- KubernetesServiceDbQueries.getActiveFullAppByName(cloudContext, appName).transaction
+      appResult <- F.fromOption(appOpt,
+                                AppNotFoundException(cloudContext, appName, ctx.traceId, "No active app found in DB")
+      )
+
+      _ <- metrics.incrementCounter("updateAppConfig", 1, Map("appType" -> appResult.app.appType.toString))
+
+      // throw 404 if no UpdateApp permission
+      listOfPermissions <- authProvider.getActions(appResult.app.samResourceId, userInfo)
+      hasPermission = listOfPermissions.toSet.contains(AppAction.UpdateApp)
```
LizBaldo, line 705: Is this pre-existing UpdateApp action in Sam doing what you would expect it to do in terms of permissions?

AKA, how is a user granted this action from Sam?

jmthibault79: That's a good question.  I don't have a good understanding of how this permission works, or what permission check would be appropriate here.  I will think about that.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
jmthibault79: updated

http/src/main/resources/swagger/api-docs.yaml
jmthibault79: updated

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
```diff
@@ -669,6 +669,80 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
...
+    _ <- F.raiseWhen(wantEnableWithoutThresholdButDbInvalid)(
+      BadRequestException(
+        "when enabling autodelete without passing an explicit autodeleteThreshold, the existing config must be valid",
+        Some(ctx.traceId)
+      )
+    )
+  } yield ()
+
+  private def updateAppConfigInternal(appId: AppId, validatedChanges: UpdateAppConfigRequest): F[Unit] = for {
+    _ <- validatedChanges.autodeleteEnabled.fold(F.unit)(enabled =>
```
Qi77Qi: can do 
```
validatedChanges.autodeleteEnabled.traverse{x => ???}
```
this will do the `???` only when the option is defined

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
```diff
@@ -669,6 +669,80 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
...
+      appQuery
+        .updateAutodeleteEnabled(appId, enabled)
+        .transaction
+        .void
+    )
+
+    // note: does not clear the threshold if None.  This only sets defined thresholds.
+    _ <- validatedChanges.autodeleteThreshold.fold(F.unit)(threshold =>
```
Qi77Qi: same thing, can use `traverse`

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/appRoutesModels.scala
Qi77Qi: Up to you if you want to do it, but if we use a custom type for `autodeleteThreshold` (say `AutodeleteThreshold`), we can then define a decoder for it and verify it has to be positive in the decoder. This way, anywhere this type is used, it's guaranteed to be postive(let's say we use this custom type in `CreateAppRequest`)

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -2914,6 +2950,14 @@ components:
             machineSize: "Standard_DS1_v2"
             disk: { labels: {}, name: "disk1", size: 50 }
             autopauseThreshold: 15
+    UpdateAppConfigRequest:
```
cpate4: This is probably my biggest nit.
Would it be possible to use `UpdateAppRequest`?

I know there is an `UpdateAppsRequest` which seems eerily close and markedly different (should really be `UpgradeApp...` - but that is for another time), but, this would also keep it inline with the `CreateAppRequest`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
```diff
@@ -669,6 +669,80 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
...
+    wantEnableWithoutThresholdButDbInvalid = req.autodeleteEnabled.getOrElse(false) &&
+      req.autodeleteThreshold.isEmpty &&
+      dbApp.autodeleteThreshold.getOrElse(0) <= 0
+
+    _ <- F.raiseWhen(wantEnableWithoutThresholdButDbInvalid)(
+      BadRequestException(
+        "when enabling autodelete without passing an explicit autodeleteThreshold, the existing config must be valid",
+        Some(ctx.traceId)
+      )
+    )
```
cpate4: I feel like this logic was coded for CreateApp - is there any possibility in being able to reuse it at all?

jmthibault79:
>  if we use a custom type for autodeleteThreshold

@Qi77Qi I made an attempt but I was unable to get it all the way working https://github.com/DataBiosphere/leonardo/pull/4583

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
jmthibault79: thanks, done

http/src/main/resources/swagger/api-docs.yaml
jmthibault79: I can do that, sure.  I can see both sides here.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
jmthibault79: That sounds sensible.

One possibility would be for this validation step to "pre-apply" the changes to the existing app and then use the same check that Create App does for validity: 
```
      autodeleteEnabled = req.autodeleteEnabled.getOrElse(false)
      autodeleteThreshold = req.autodeleteThreshold.getOrElse(0)

      _ <- Either.cond(!(autodeleteEnabled && autodeleteThreshold <= 0),
                       (),
                       BadRequestException("autodeleteThreshold should be a positive value", Some(ctx.traceId))
      )
 
```

However the logic we discussed in the design doc differs from this.  For one example, Create App won't reject [false, -1].

What do you think about unifying the logic but using the stricter standard as seen in this PR?  This would cause a small behavior change in Create App.

http/src/main/resources/swagger/api-docs.yaml
jmthibault79: done

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
cpate4: Personally ... I think it should be consistent (whatever it is). It seems / feels weird to have two different behaviors.

Talking it out here ...
If the autodelete is `enabled`, then we NEED to ensure that the `threshold` is `> 0`.

Given that the `threshold` is only valid if it is _greater than zero_, should we just restrict it to always being greater than zero?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/AppRoutes.scala
jmthibault79: Thanks!  I split it out into a new PR #4585 for further exploration, and removed it from here.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
jmthibault79: Agreed.  I found a nice way to do this.

Final logic for both create and update:
* reject threshold <=0 if present
* if enabled=true, reject if a threshold is not present (threshold value check is covered by previous step)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
```diff
@@ -1469,6 +1511,27 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
     )
   }
 
+  private[service] def validateAutodelete(autodeleteEnabled: Boolean,
```
jmthibault79, line 1511: not sure where this should be placed.  thoughts?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
jmthibault79: The list of App Actions seems to match that in Sam: https://github.com/broadinstitute/sam/blob/develop/src/main/resources/reference.conf

On L829 I see that a user with an app `creator` role has `update` as one of their actions.

This matches my expectation.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
```diff
@@ -1429,13 +1472,9 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
...
         else None
-      autodeleteEnabled = req.autodeleteEnabled.getOrElse(false)
-      autodeleteThreshold = req.autodeleteThreshold.getOrElse(0)
 
-      _ <- Either.cond(!(autodeleteEnabled && autodeleteThreshold <= 0),
-                       (),
-                       BadRequestException("autodeleteThreshold should be a positive value", Some(ctx.traceId))
-      )
+      autodeleteEnabled = req.autodeleteEnabled.getOrElse(false)
+      _ <- validateAutodelete(autodeleteEnabled, req.autodeleteThreshold, ctx.traceId)
```
jmthibault79, line 1477: note that this makes the **Create App** validation more strict, to match

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
```diff
@@ -669,6 +670,48 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
...
+      appQuery
+        .updateAutodeleteEnabled(appId, enabled)
+    )
+
+    // note: does not clear the threshold if None.  This only sets defined thresholds.
+    _ <- validatedChanges.autodeleteThreshold.traverse(threshold =>
+      appQuery
+        .updateAutodeleteThreshold(appId, Some(threshold))
+    )
+  } yield ()).transaction
```
jmthibault79, line 684: Don't want to commit one without the other!

Example: the app had a 1-hour threshold, but it was disabled.  Now the user wants to enable with a 10-hour threshold.  If only the first change goes through, the app has a surprise active 1-hour threshold.

cpate4 (approved):

Qi77Qi (approved):

PR #4577, opened 2024-05-21
[IA-4724] Reenable GPU automation test when merging to dev

LizBaldo:
__Jira ticket__: https://broadworkbench.atlassian.net/browse/IA-4724

### What

- Reenables the GPU automation test when merging to dev only

### Why

- We need to be warned when GPU functionality fails, but this test is too costly to run on every PR commit, and so hooking it as part of the merge to dev run that notifies failure to slack feels like a good compromise

### What to test

- 

### Who tested and where

- [ ] I validated this change 


.github/workflows/leo-build-tag-publish-and-run-tests.yml
```diff
@@ -76,8 +77,10 @@ jobs:
           if ${{ github.ref_name == 'develop' }}; then
             echo 'test-context=dev-merge'                              >> $GITHUB_OUTPUT
             echo 'relates-to-chart-releases=leonardo-dev'              >> $GITHUB_OUTPUT
+            echo 'leo-test-command=testOnly -- -l ExcludeFromJenkins'  >> $GITHUB_OUTPUT
           else
             echo 'test-context=pr-test'                                >> $GITHUB_OUTPUT
+            echo 'leo-test-command=testOnly -- -l "ExcludeFromJenkins ScheduledTest"'  >> $GITHUB_OUTPUT
```
jdcanas: shouldn't this be in the if branch and not the else?

core/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoTestTags.scala
```diff
@@ -9,4 +9,8 @@ object LeonardoTestTags {
   // intent: do not run in given execution context
   // this is also synonmous with 'azure test'
   object ExcludeFromJenkins extends Tag("ExcludeFromJenkins")
+
+  // intent: run in cron scheduled tests only, not at every PR commit
+  // example use case is for GPU testing that would otherwise run into resource allocation problems
+  object ScheduledTest extends Tag("ScheduledTest")
```
lucymcnatt: a nit, but since is not truly 'scheduled', could you update this comment to say on merges to dev? and maybe switch up the naming?

core/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoTestTags.scala
```diff
@@ -9,4 +9,8 @@ object LeonardoTestTags {
   // intent: do not run in given execution context
   // this is also synonmous with 'azure test'
   object ExcludeFromJenkins extends Tag("ExcludeFromJenkins")
+
+  // intent: run in cron scheduled tests only, not at every PR commit
```
mlilynolting: Is this still true now that we're running on merge to develop?

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/runtimes/RuntimeGceSpec.scala
```diff
@@ -76,58 +82,56 @@ class RuntimeGceSpec extends BillingProjectFixtureSpec with ParallelTestExecutio
...
-//    res.unsafeRunSync()(cats.effect.unsafe.IORuntime.global)
-//  }
+  "should be able to create a VM with GPU enabled" taggedAs ScheduledTest in { project =>
+    val runtimeName = randomClusterName
+    val diskName = genDiskName.sample.get
+
+    val toolImage = ContainerImage.fromImageUrl(
+      "us.gcr.io/broad-dsp-gcr-public/terra-jupyter-python:qi-gpu"
+    ) // Use the default base image once we migrate to use gpu images by default
+    // In a europe zone
```
mlilynolting: is this part of the TODO? isn't the zone us-west1-a?

.github/workflows/leo-build-tag-publish-and-run-tests.yml
LizBaldo: I don't think so, we want to exclude the scheduled test from the pr context right?

automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/runtimes/RuntimeGceSpec.scala
LizBaldo: Ohhhh, very good catch, thanks! Damn that test is old 😰

core/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoTestTags.scala
LizBaldo: Thanks for the catch, I will update the comment 👍

core/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoTestTags.scala
LizBaldo: Will do, thanks!

.github/workflows/leo-build-tag-publish-and-run-tests.yml
LizBaldo: I renamed the Tag to follow the Exclude example so hopefully it should feel clearer now

core/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoTestTags.scala
```diff
@@ -9,4 +9,8 @@ object LeonardoTestTags {
   // intent: do not run in given execution context
   // this is also synonmous with 'azure test'
   object ExcludeFromJenkins extends Tag("ExcludeFromJenkins")
+
+  // intent: do not run on every PR commit
+  // e.g. GPU tests should not run at every PR commit to avoid resource allocation problems
+  object ExcludeFromPRTest extends Tag("ExcludeFromPRTest")
```
mlilynolting: I'd name this ExcludeFromPrCommit to clarify that it's ignoring commits to PR branches. I think the -Test suffix isn't necessary.

mlilynolting (approved):

jdcanas (approved):
Ah yeah, these tags are excluded from this run. You probably need to modify the azure GHA to have this tag then, or it will get run in all azure automation tests.

core/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoTestTags.scala
LizBaldo: Good idea, will do :)

PR #4581, opened 2024-05-22
[IA-4926] Reenable ssh portion of Azure automation tests

LizBaldo:
__Jira ticket__: https://broadworkbench.atlassian.net/browse/IA-4926

### What

- Adds set -e -x to ssh scripts for easier debugging. Revert the changes made previously that was causing the startTunnel.sh script to hang

### Why

-

### What to test

- 

### Who tested and where

- [ ] I validated this change 

### DIFF
--- a/automation/src/test/resources/startTunnel.sh
+++ b/automation/src/test/resources/startTunnel.sh
@@ -34,5 +34,3 @@ done
 PID=$(lsof -i tcp:$PORT | tail -1 | awk '{print $2}')
 # The last line of this script needs to echo the pid for the sbt test suite to record and clean up
 echo "$PID"
-
-exit 0
--- a/automation/src/test/resources/startTunnel.sh
+++ b/automation/src/test/resources/startTunnel.sh
@@ -34,3 +34,5 @@ done
 PID=$(lsof -i tcp:$PORT | tail -1 | awk '{print $2}')
 # The last line of this script needs to echo the pid for the sbt test suite to record and clean up
 echo "$PID"
+
+exit 0
--- a/automation/src/test/resources/startTunnel.sh
+++ b/automation/src/test/resources/startTunnel.sh
@@ -14,13 +14,6 @@ echo "starting tunnel bastion ${BASTION_NAME} with resource id ${RESOURCE_ID} in
 
 nohup az network bastion tunnel --name "$BASTION_NAME" --resource-group "$RESOURCE_GROUP" --target-resource-id "${RESOURCE_ID}" --resource-port 22 --port $PORT > /dev/null 2>&1 &
 
-#nohup az network bastion tunnel --name "$BASTION_NAME" --resource-group "$RESOURCE_GROUP" --target-resource-id "${RESOURCE_ID}" --resource-port 22 --port $PORT > /tmp/nohup.out &
-#
-#
-#echo "displaying nohup output"
-#cat /tmp/nohup.out
-#echo "done displaying nohup output"
-
 NUM_LOOPS=0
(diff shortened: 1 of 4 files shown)


automation/src/test/resources/startTunnel.sh
```diff
@@ -1,3 +1,6 @@
+#!/usr/bin/env bash
+set -e -x
```
LizBaldo, line 2: @jdcanas were you saying that this is what is causing the GCP automation test to fail?

jdcanas: no, this line https://github.com/DataBiosphere/leonardo/pull/4581/files#diff-d176682250d61f7b1e0d9d9513e4a47c5ccfd02da8a27752c46e38d9b8825984R2

jdcanas: same premise though. Essentially it is expected that azure sign in won't work if the tests are being run for google. You can either
1) add an arg so the script is aware of gcp/google (may be a bit annoying since I believe it is triggered by something in `terra-github-workflows`
2) `| true` the lines that are now failing (that always failed for google) so the script doesn't exit. This could possibly hide a real issue though... (i.e. the creds rotating)

cpate4 (approved):
good luck!

automation/src/test/resources/startTunnel.sh
LizBaldo: How fun, yeah it is called here: https://github.com/broadinstitute/terra-github-workflows/blob/28726262ceb146ca26f44672f354aa00015f2464/.github/actions/run-leo-swat-tests/action.yaml#L425

LizBaldo: If you are cool with it, I might just leave it at -x only as part of the scope of this PR so we reenable the test ASAP, and I can create a tech debt ticket to run environment specific test setup scripts

jdcanas: Yeah I could see punting it... however not including `e` here is why the tests starting failing originally proceeding us disabling it  😞  (because the creds were rotated and the fact it wasn't signing in hidden in thousands of lines of output)

jdcanas (approved):
will leave the `-e` to your discretion, but it pains me to not fix the issue that originally resulted in a few days of time of debugging to figure out sign in wasn't working for azure :(

automation/src/test/resources/startTunnel.sh
LizBaldo: Oh, but i thought that the -x would still display the errors no?

LizBaldo: The errors are shown in the log, the script is just not tripping over it: https://github.com/broadinstitute/terra-github-workflows/actions/runs/9227610112/job/25389926510

jdcanas: Yeah, it displays them, but since it doesn't fail fatally it continues on to execute the tests and results in weird ssh errors. Could we just check if the variables are defined? That way we don't have to change the callsite in the other repo. Seems like a fair assumption to just echo a warning if the vars aren't defined and skip the login step? That way we can add in the `e`

LizBaldo: By variables you mean these ones?
LEO_AZURE_CLIENT_ID
LEO_AZURE_CLIENT_SECRET
LEO_AZURE_TENANT_ID
LEO_AZURE_SUBSCRIPTION_ID

jdcanas: yeah, you can probably just check one

PR #4583, opened 2024-05-24
[RW-11461] Use a value class for AutodeleteThreshold

jmthibault79:
__Jira ticket__: https://precisionmedicineinitiative.atlassian.net/browse/RW-11461

Follow-up to #4575 

### What

Changes the `autodeleteThreshold` field from Int to a new value class wrapping Int, in App structures and endpoints

### Why

Narrows the type to a valid range, instead of all integers

## Testing these changes

I confirmed that I can create and delete an app on a bee.  I exercised this field by hitting update endpoints with the same payloads that I used previous to this change.

### Who tested and where

- [x] I validated this change 


core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/appRoutesModels.scala
```diff
@@ -35,11 +36,11 @@ final case class CreateAppRequest(kubernetesRuntimeConfig: Option[KubernetesRunt
                                   extraArgs: List[String],
                                   workspaceId: Option[WorkspaceId],
                                   sourceWorkspaceId: Option[WorkspaceId],
-                                  autodeleteThreshold: Option[Int],
-                                  autodeleteEnabled: Option[Boolean]
+                                  autodeleteEnabled: Option[Boolean],
```
jmthibault79, line 40: while we are updating the threshold type, also change the order of these parameters so the more important one (enabled) is first

jmthibault79:
Azure test failure "failed to create billing project" is clearly unrelated.  Rerunning.

http/src/main/resources/swagger/api-docs.yaml
```diff
@@ -4146,13 +4146,15 @@ components:
...
-          description: The number of minutes of idle time to elapse before the app is
-            deleted. When autodeleteEnabled is true, a positive integer is required
         autodeleteEnabled:
           type: boolean
           description: Whether to turn on autodelete
+        autodeleteThreshold:
+          type: integer
+          description: >
+            Optional: The number of minutes of idle time to elapse before the app is deleted.  This must be a positive
+            integer.  Note: deletion timing is not exact.  The app will be deleted a short period of time after this
```
jmthibault79, line 4156: also updated this text

mlilynolting (approved):

Qi77Qi (approved):
thanks for working on this!

PR #4585, opened 2024-05-28
[RW-11461] Add withSpanResource helper function and use in AppRoutes and AppRoutesV2

jmthibault79:
__Jira ticket__: https://precisionmedicineinitiative.atlassian.net/browse/RW-11461

### What

No expected changes in functionality.  This is a code refactoring to simplify adding `spanResource` to API calls.

I updated AppRoutes and AppRoutesV2 to use this new function, but the change can be applied to all routes.  Should I do so here?

### Why

-

### What to test

- 

### Who tested and where

- [ ] I validated this change 

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/AppRoutes.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/AppRoutes.scala
@@ -119,17 +119,8 @@ class AppRoutes(kubernetesService: AppService[IO], userInfoDirectives: UserInfoD
                                     googleProject: GoogleProject,
                                     appName: AppName,
                                     req: CreateAppRequest
-  )(implicit
-    ev: Ask[IO, AppContext]
-  ): IO[ToResponseMarshallable] =
+  )(implicit ev: Ask[IO, AppContext]): IO[ToResponseMarshallable] =
     for {
-      ctx <- ev.ask[AppContext]
-      apiCall = kubernetesService.createApp(
-        userInfo,
-        CloudContext.Gcp(googleProject),
-        appName,
-        req
-      )
       _ <- req.allowedChartName match {
         case Some(cn) =>
           val tags = Map("appType" -> req.appType.toString) + ("chartName" -> cn.asString)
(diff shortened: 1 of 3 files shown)


Qi77Qi (approved):

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/AppRoutes.scala
```diff
@@ -141,92 +132,88 @@ class AppRoutes(kubernetesService: AppService[IO], userInfoDirectives: UserInfoD
...
+  ): IO[ToResponseMarshallable] = {
+    val apiCallName = "getApp"
     for {
-      ctx <- ev.ask[AppContext]
-      apiCall = kubernetesService.getApp(
-        userInfo,
-        CloudContext.Gcp(googleProject),
-        appName
+      _ <- metrics.incrementCounter(apiCallName)
+      resp <- withSpanResource(apiCallName,
```
Qi77Qi, line 151: can potentially put the metrics line into the `withSpanResource` too...

Qi77Qi (approved):

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/AppRoutes.scala
jmthibault79: I was thinking of that. However, not all routes have a metrics call like this.  Sometimes metrics are called from the service instead.

mlilynolting (approved):
Approved, but I like Qi's idea of folding the metrics call into `withSpanResource`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/AppRoutes.scala
jmthibault79: see also `createAppHandler()` where the metrics have some associated logic

Qi77Qi: ah right, I think for some, we needed additional info, so we had to move it to service layer...

jmthibault79:
Azure tests failed: https://github.com/broadinstitute/terra-github-workflows/actions/runs/9370192130/job/25796565555
```
[info] *** 1 TEST FAILED ***
Error:  Failed tests:
Error:  	org.broadinstitute.dsde.workbench.leonardo.LeonardoAzureSuite
```

Rerunning.

PR #4593, opened 2024-05-29
[IA-4974] add code for checking liveliness before upgrading

jdcanas:


http/src/main/resources/reference.conf
```diff
@@ -1171,6 +1171,11 @@ pubsub {
       interval = 3 seconds
       interruptAfter = 10 minutes
     }
+    appLivelinessCheck {
```
LizBaldo: Nit, but can the listener handle that frequency? Just wondering if that's something to play with but that can be a reasonable default to start with

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -1534,6 +1534,7 @@ class LeoPubsubMessageSubscriber[F[_]](
           // Fatal case (as in, the app is no longer usable), polling app liveliness failed
           // The two fatal cases are included separately, because later we may wish to fail fatally on `HelmException`, but roll back on `AppUpdatePollingException`
           // This would provide more cases in which an app is left in a usable state
+          // Not that an app can also emit this error if the livliness probe fails before an update is triggered, so rolling back may not have an effect
```
LizBaldo: nit `liveliness`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala
```diff
@@ -430,6 +430,21 @@ class AKSInterpreter[F[_]](config: AKSInterpreterConfig,
                                         namespaceName
       )
 
+      // Check if the app is alive before attempting to update. We transition the app to error status if this fails
```
LizBaldo, line 433: Works for me, could you add a similar thing to the GKEInterpreter though? We do app upgrades there too in theory

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
cpate4:
```suggestion
          // Note that an app can also emit this error if the liveliness probe fails before an update is triggered, so rolling back may not have an effect
```

http/src/main/resources/reference.conf
jdcanas: Hmm... my assumption is that in most cases this will return immediately, and if we are retrying we don't actually expect it to ever be available, so it doesn't matter much what this is configured to in my head

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreterSpec.scala
```diff
@@ -226,6 +226,67 @@ class AKSInterpreterSpec extends AnyFlatSpecLike with TestComponent with Leonard
       res.unsafeRunSync()(cats.effect.unsafe.IORuntime.global)
     }
 
+  it should s"emit an exception if the app is not alive prior to upgrading" in isolatedDbTest {
```
LizBaldo, line 229: Would you be able to add a similar test in the GKEInterpreterSpec?

LizBaldo (approved):
👍 to unblock you before I head out earlier this afternoon, but it would be nice to have a similar unit test for the GKE side of the upgrade

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/AppMonitorConfig.scala
```diff
@@ -13,5 +13,6 @@ case class AppMonitorConfig(nodepoolCreate: PollMonitorConfig,
                             scalingUpNodepool: PollMonitorConfig,
                             scalingDownNodepool: PollMonitorConfig,
                             startApp: InterruptablePollMonitorConfig,
-                            updateApp: InterruptablePollMonitorConfig
+                            updateApp: InterruptablePollMonitorConfig,
+                            appLivelinessCheck: PollMonitorConfig
```
mlilynolting: it seems like for most (but not all) of these verbs we do verb-subject, like `createApp`; maybe `checkAppLiveliness` is best here?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -764,12 +764,29 @@ class GKEInterpreter[F[_]](
...
 
           } yield (chartValues, last)
-        case _ => F.raiseError(AppUpdateException(s"App type ${app.appType} not supported on GCP", Some(ctx.traceId)))
+        case _ =>
+          F.raiseError[(String, Boolean)](
+            AppUpdateException(s"App type ${app.appType} not supported on GCP", Some(ctx.traceId))
+          )
       }
 
+      (chartOverrideValues, preUpdateAppOk) <- chartOverridesAndAppOkF
```
mlilynolting, line 773: Why break these into an `=` followed by a `<-`?

jdcanas: to re-use the ok check logic before and after update

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/config/AppMonitorConfig.scala
jdcanas: Hmmm.. there isn't a verb though its a single noun, the `appLivelinessCheck[Config]` . I could see an argument for a restructuring though, don't feel too strongly

mlilynolting: Why not a verb? all the rest of these are verbs

mlilynolting: well, except for the `scalingDown|Up` which for some reason are gerunds

jdcanas: They aren't though, a verb implies it is a method that performs an action. For example, `updateApp` could be interpreted in this ambiguous form as an action descriptor, but it really is the `updateAppConfig`, with the `config` part being omitted due to redundancy/variable name length.

mlilynolting: isn't "update app config" a verb with the object of the verb being the "app config"?

Nevertheless I think what you're saying is that this is a GET endpoint, not a mutative action, so it's not doing anything but retrieving a status. These usually get named things like `/status` or `/details` etc, and calling them `getDetails` is redundant. In that case I'd advocate for `appLiveliness` since the verb isn't important.

jdcanas: I'm fine with `appLiveness` here. Though for the sake of curiosity, I do believe `updateApp` is a compound adjective describing `config` in this case.

mlilynolting: oh, I see what you meant :) I was misreading when you talked about config here. Yeah, i get what you're saying, all these are the names of different config blocks tied to the various actions: updateApp, startApp etc.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
mlilynolting: Is it necessary to flatMap the chartOverridesAndAppOkF object twice to get two distinct ok check functions? Wouldn't it create a totally separate stream function if you did `if (preUpdateAppOk)` twice?

mlilynolting (commented):
No issues except for naming - it does seem like the terra ecosystem only uses `liveness` now, and since that's the k8s/helm term I'd change to use that instead of 'liveliness' for consistency here.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
jdcanas: No, using `=` establishes an `F[_]`, which is not executed until flatmapped (`isProxyAvailable` being the thing executed when it is flatmapped). Using `<-` in the original block actually prematurely runs the `appOk` check; this was a bug I am fixing here. You can find similar examples when we use the `AsyncTaskProcessor`, for which we do `task = ...`, where we pass the task `F[_]` to the task processor to execute.

mlilynolting: Gotcha, thanks!

mlilynolting (approved):
I still think the "check" in "appLivenessCheck" is incongruous, but that's not a blocker and all looks good!

PR #4601, opened 2024-05-31
RW-12676 Support Autopilot

Qi77Qi:
See more details in [this doc](https://docs.google.com/document/d/1nt4OCe7btcybAeOxI1ePlauU3nlOEgESekQOCPMLkfU/edit#heading=h.2qqbidhwai7s)

This PR implements the changes mentioned in [this part](https://docs.google.com/document/d/1nt4OCe7btcybAeOxI1ePlauU3nlOEgESekQOCPMLkfU/edit#heading=h.o3bg0iex6op5)

It supports autopilot in a way that only people in the `custom_app_users` group can try out the new feature. Apps without autopilot should continue to work the same way as before. This PR also only pass autopilot related settings to ALLOWED app as a start for evaluation purpose.

The goal of this PR is such that we can evaluate autopilot more thoroughly once it's available in leonardo prod. After this PR, we'll continue to evaluate in a few more aspects
1. Tune app resources to find a good default that balances performance and cost (it'll be more realistic once we can have it available on AoU's preprod)
2. We'll explore add workload separation for security concerns
3. Verify data exfil will continue to work as same before
4. Have a better understanding of time it takes to createApp from metrics

__Jira ticket__: https://precisionmedicineinitiative.atlassian.net/browse/RW-12676

### What

-

### Why

-

### What to test

- 

### Who tested and where

- [ ] I validated this change 


http/src/main/resources/reference.conf
```diff
@@ -716,6 +716,19 @@ gke {
     version = "1.28"
     nodepoolLockCacheExpiryTime = 1 hour
     nodepoolLockCacheMaxSize = 200
+
+    autopilot {
+      welder {
```
LizBaldo, line 721: So the idea is that only the autopilot config for the rstudio/sas are configurable via the API and you anticipate welder and wondershaper to stay constant?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/KubernetesClusterComponent.scala
```diff
@@ -303,7 +304,8 @@ case class SaveKubernetesCluster(cloudContext: CloudContext,
                                  status: KubernetesClusterStatus,
                                  ingressChart: Chart,
                                  auditInfo: AuditInfo,
-                                 defaultNodepool: DefaultNodepool // Question: does this have to be `DefaultNodepool`?
+                                 defaultNodepool: DefaultNodepool, // Question: does this have to be `DefaultNodepool`?
+                                 autopilot: Boolean
```
LizBaldo: Nit, rename to autopilotEnabled to avoid confusion with the Autopilot object?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
```diff
@@ -88,6 +89,14 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
       _ <- F.raiseWhen(!hasPermission)(ForbiddenError(userInfo.userEmail))
 
       enableIntraNodeVisibility = req.labels.get(AOU_UI_LABEL).exists(x => x == "true")
+      // We only allow autopilot mode for members of CUSTOM_APP_USERS group
```
LizBaldo, line 92: Would it make sense to remove this constraints so you can test on Terra BEES? Just a thought

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
```diff
@@ -1045,32 +1064,39 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
   private[service] def getSavableCluster(
     userEmail: WorkbenchEmail,
     cloudContext: CloudContext,
+    autopilot: Boolean,
```
LizBaldo: See my comment above, I think renaming to autopilotEnabled might be a bit clearer

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
```diff
@@ -1045,32 +1064,39 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
...
+      status = nodepoolStatus,
       auditInfo,
       machineType = config.leoKubernetesConfig.nodepoolConfig.defaultNodepoolConfig.machineType,
       numNodes = config.leoKubernetesConfig.nodepoolConfig.defaultNodepoolConfig.numNodes,
       autoscalingEnabled = config.leoKubernetesConfig.nodepoolConfig.defaultNodepoolConfig.autoscalingEnabled,
       autoscalingConfig = None
     )
 
+    // autopilot mode clusters are regional
+    val loc =
```
LizBaldo, line 1090: nit `location`

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -1120,7 +1072,9 @@ class LeoPubsubMessageSubscriber[F[_]](
 
         // create and monitor app
         _ <- getGkeAlgFromRegistry()
-          .createAndPollApp(CreateAppParams(msg.appId, msg.project, msg.appName, msg.machineType))
+          .createAndPollApp(
+            CreateAppParams(msg.appId, msg.project, msg.appName, msg.machineType)
+          ) // TODO: qi fix this
```
LizBaldo: What is this TODO about?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -139,17 +147,24 @@ class GKEInterpreter[F[_]](
 
       networkConfig = new com.google.api.services.container.model.NetworkConfig()
         .setEnableIntraNodeVisibility(params.enableIntraNodeVisibility)
+
+      networkPolicy =
+        if (params.autopilot) null else new com.google.api.services.container.model.NetworkPolicy().setEnabled(true)
```
LizBaldo, line 155: Why is the network policy null in case of autopilot?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -444,7 +458,7 @@ class GKEInterpreter[F[_]](
               app.release,
               app.chart,
               dbCluster,
-              dbApp.nodepool.nodepoolName,
+              dbApp.nodepool.nodepoolName, // TODO: support autopilot mode
```
LizBaldo: Do you mean this is the TODO to support autopilot mode for Galaxy? Is there a JIRA ticket that could be linked?

http/src/main/resources/reference.conf
Qi77Qi: yea, welder and wondershaper is not tunable by users since they user don't interact with them directly...right now, app (SAS and RStudio) resources are passed from API client, and I didn't set defaults for those (will be easy to add if we like later).

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
Qi77Qi: I've tested a bunch on BEE already...this is so that this PR can be merged and we don't need to worry about it getting to prod

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
Qi77Qi: this is not supported in autopilot mode

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
Qi77Qi: yea...I can create one

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
Qi77Qi: outdated...will remove

http/src/main/resources/reference.conf
yonghaoy: +1. Those are aou/app specific. Eventually it should be moved into app chart.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
```diff
@@ -139,17 +147,24 @@ class GKEInterpreter[F[_]](
...
+      networkPolicy =
+        if (params.autopilot) null else new com.google.api.services.container.model.NetworkPolicy().setEnabled(true)
+
       legacyCreateClusterRec = new com.google.api.services.container.model.Cluster()
         .setName(dbCluster.clusterName.value)
         .setInitialClusterVersion(config.clusterConfig.version.value)
         .setNodePools(nodepools.asJava)
+        .setAutopilot(
+          autopilot
+        )
```
yonghaoy, line 163: It's very important to set networkTags. That is how data exfil works in AoU. 
How to set network tag in autopilot? 
https://github.com/DataBiosphere/leonardo/blob/a2b32c2a2df35f841a4a8de90061bf1a8ab8394f/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala#L1796

http/src/main/resources/reference.conf
Qi77Qi: I didn't provide defaults for these in charts themselves since defining them in leo will be faster for us to iterate these values

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
Qi77Qi: not sure abt this one cuz it'll look like
```
SaveKubernetesCluster(
      cloudContext = cloudContext,
      clusterName = defaultClusterName,
      location = location,
```
a few lines below...I think Scala will allow us to use the same name, but I've had comments from others they don't like that

LizBaldo: Oh gotcha, I am used to that convention but that is me from Python talking, feel free to ignore then

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
```diff
@@ -197,27 +206,32 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
...
-              )
-              saveNodepool <- F.fromEither(
-                getUserNodepool(clusterId, cloudContext, originatingUserEmail, machineConfig, ctx.now)
-              )
-              savedNodepool <- nodepoolQuery.saveForCluster(saveNodepool).transaction
-            } yield savedNodepool
-        }
+        nodepool <-
+          if (req.autopilot.isDefined) {
+            F.pure(saveClusterResult.defaultNodepool.toNodepool())
```
lucymcnatt, line 216: just to be clear, this nodepool that's saved to the DB won't be truly accurate with what GCP is actually doing?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -1005,74 +1005,26 @@ class LeoPubsubMessageSubscriber[F[_]](
       // _initial_ GKE call synchronous to the pubsub processing so we can nack the message on
```
lucymcnatt, line 1005: does this comment need to be updated with the new autopilot logic?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
Qi77Qi: yes..it's the same as how we do things with Azure...it'll be a big overhaul with DB schema if we don't want to do this because APP table has `nodepoolId` as FK and it doesn't reference kubernetest_cluster table directly

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
Qi77Qi: I think it still applies since with autopilot, we're still creating the cluster when there's no existing clusters

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala
Qi77Qi: created https://broadworkbench.atlassian.net/browse/IA-4987

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/BuildHelmChartValues.scala
```diff
@@ -380,6 +382,26 @@ private[leonardo] object BuildHelmChartValues {
...
+          raw"""autopilot.welder.cpu=${config.clusterConfig.autopilotConfig.welder.cpuInMillicores}m""",
+          raw"""autopilot.welder.memory=${config.clusterConfig.autopilotConfig.welder.memoryInGb}Gi""",
+          raw"""autopilot.welder.ephemeral\-storage=${config.clusterConfig.autopilotConfig.welder.ephemeralStorageInGb}Gi""",
+          raw"""autopilot.wondershaper.cpu=${config.clusterConfig.autopilotConfig.wondershaper.cpuInMillicores}m""",
+          raw"""autopilot.wondershaper.memory=${config.clusterConfig.autopilotConfig.wondershaper.memoryInGb}Gi""",
+          raw"""autopilot.wondershaper.ephemeral\-storage=${config.clusterConfig.autopilotConfig.wondershaper.ephemeralStorageInGb}Gi"""
+        )
+        if (v.computeClass == ComputeClass.GeneralPurpose)
+          ls
+        else raw"""nodeSelector.cloud\.google\.com/compute-class=${v.computeClass.toString}""" :: ls
```
Qi77Qi, line 403: when it's general purpose, GCP doesn't allow us to pass the compute class value...the API behaves in a way that when the value isn't specified, general-purpose is used

http/src/main/resources/reference.conf
LizBaldo: Then maybe adding a small TODO to move these into the charts once the correct values have been found would be nice

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/BuildHelmChartValues.scala
LizBaldo: Can you add that as a comment in the code? That will help future dev understand better I think

http/src/main/resources/reference.conf
Qi77Qi: I don't know if having defaults in charts provide much value since they would almost never be used...they'll be overwriten by Leonardo

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -1150,6 +1104,55 @@ class LeoPubsubMessageSubscriber[F[_]](
       )
     } yield ()
 
+  private def createClusterAndNodepools(msg: CreateAppMessage,
+                                        clusterId: KubernetesClusterLeoId,
+                                        autopilot: Boolean,
```
LizBaldo: Could this one also be renamed to `autopilotEnabled` for clarity?

LizBaldo (approved):
LGTM, thanks for addressing my comments! I only only have one more little renaming suggestion to match the rest

yonghaoy (approved):
Change LG overrall, good enough for testing in dev.
Few questions about the API interface:
1: Shall we introduce a boolean for "autopilotEnabled" flag? If no, then we will use `Autopilot` field in the request? 
2: For fields inside `Autopilot`, we might want to add some good default value/minimum value? 
3: There are some overlaps between the new Autopilot with k8sRuntimeConfig. So that would be oneof in the future? What if I have both when calling Leo? 

We can polish this in the future, not a big concern.

Qi77Qi:
> 1: Shall we introduce a boolean for "autopilotEnabled" flag? If no, then we will use Autopilot field in the request?

I don't think we need to just so we dont' have to sanitize invalid combo of inputs...In evaluation phase, this field will precedence, meaning if this is defined, it'll use autopause. In the future, I think we would want to remove `k8sRuntimeConfig` if possible.

> 2: For fields inside Autopilot, we might want to add some good default value/minimum value?

I'm thinking of experimenting without defaults...in the past, we define defaults in both API and UI, whenever we update defaults in API, we also update UI..Figured maybe it makes sense to just have default in UI...Can add defaults in the future while we're actively evaluating things.

> 3: There are some overlaps between the new Autopilot with k8sRuntimeConfig. So that would be oneof in the future? What if I have both when calling Leo?

Yea, the hope is we'll be able to remove `k8sRuntimeConfig`...I'll add some clarity in swagger doc in a follow up PR

yonghaoy:
Why RStudio version in dockerfile is 0.9.0 while the version in `reference.conf` is 0.10.0? Same question for SAS

PR #4606, opened 2024-06-06
[IA-4877] Add table for app upgrade tracking

jdcanas:
Unit tests not written, want feedback on general approach/data model before writing (often verbose and type-specific) tests

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
@@ -597,3 +597,25 @@ object ComputeClass {
   val stringToObject = values.map(v => v.toString.toLowerCase -> v).toMap
 }
 final case class Autopilot(computeClass: ComputeClass, cpuInMillicores: Int, memoryInGb: Int, ephemeralStorageInGb: Int)
+
+final case class UpdateAppTableId(value: Long) extends AnyVal
+final case class UpdateAppJobId(value: UUID) extends AnyVal
+
+sealed abstract class UpdateAppJobStatus
+object UpdateAppJobStatus {
+
+  case object Running extends UpdateAppJobStatus {
+    override def toString: String = "RUNNING"
+  }
+
+  case object Error extends UpdateAppJobStatus {
+    override def toString: String = "ERROR"
+  }
+
+  final case object Success extends UpdateAppJobStatus {
+    override def toString: String = "SUCCESS"
+  }
+
+  def values: Set[UpdateAppJobStatus] = sealerate.values[UpdateAppJobStatus]
+  def stringToObject: Map[String, UpdateAppJobStatus] = values.map(v => v.toString -> v).toMap
+}
--- a/core/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/AdminRoutesTestJsonCodec.scala
+++ b/core/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/AdminRoutesTestJsonCodec.scala
@@ -1,7 +1,6 @@
 package org.broadinstitute.dsde.workbench.leonardo.http
 
 import io.circe.Encoder
-import io.circe.Encoder.AsArray.importedAsArrayEncoder
 import org.broadinstitute.dsde.workbench.leonardo.JsonCodec._
 import org.broadinstitute.dsde.workbench.leonardo.UpdateAppJobId
 import org.broadinstitute.dsp.ChartVersion
@@ -9,8 +8,6 @@ import org.broadinstitute.dsp.ChartVersion
 object AdminRoutesTestJsonCodec {
 
   implicit val chartVersionEncoder: Encoder[ChartVersion] = Encoder.encodeString.contramap(_.asString)
-
-  implicit val updateAppJobIdEncoder: Encoder[UpdateAppJobId] = Encoder.encodeUUID.contramap(_.value)
 
   implicit val getAppUpdateRequestEncoder: Encoder[UpdateAppsRequest] = Encoder.forProduct9(
     "jobId",
--- a/core/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/AdminRoutesTestJsonCodec.scala
+++ b/core/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/AdminRoutesTestJsonCodec.scala
@@ -8,7 +8,7 @@ object AdminRoutesTestJsonCodec {
 
   implicit val chartVersionEncoder: Encoder[ChartVersion] = Encoder.encodeString.contramap(_.asString)
 
-  implicit val getAppUpdateRequestEncoder: Encoder[UpdateAppsRequest] = Encoder.forProduct9(
+  implicit val updateAppRequestEncoder: Encoder[UpdateAppsRequest] = Encoder.forProduct9(
     "jobId",
     "appType",
     "cloudProvider",
--- a/core/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/AdminRoutesTestJsonCodec.scala
+++ b/core/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/AdminRoutesTestJsonCodec.scala
@@ -10,7 +10,7 @@ object AdminRoutesTestJsonCodec {
 
   implicit val chartVersionEncoder: Encoder[ChartVersion] = Encoder.encodeString.contramap(_.asString)
 
-  implicit val chartVersionEncoder: Encoder[UpdateAppJobId] = Encoder.encodeUUID.contramap(_.value)
+  implicit val updateAppJobIdEncoder: Encoder[UpdateAppJobId] = Encoder.encodeUUID.contramap(_.value)
 
   implicit val getAppUpdateRequestEncoder: Encoder[UpdateAppsRequest] = Encoder.forProduct9(
     "jobId",
--- a/core/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/AdminRoutesTestJsonCodec.scala
+++ b/core/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/AdminRoutesTestJsonCodec.scala
@@ -3,13 +3,17 @@ package org.broadinstitute.dsde.workbench.leonardo.http
 import io.circe.Encoder
 import io.circe.Encoder.AsArray.importedAsArrayEncoder
 import org.broadinstitute.dsde.workbench.leonardo.JsonCodec._
+import org.broadinstitute.dsde.workbench.leonardo.UpdateAppJobId
 import org.broadinstitute.dsp.ChartVersion
 
 object AdminRoutesTestJsonCodec {
 
   implicit val chartVersionEncoder: Encoder[ChartVersion] = Encoder.encodeString.contramap(_.asString)
 
-  implicit val getAppUpdateRequestEncoder: Encoder[UpdateAppsRequest] = Encoder.forProduct8(
+  implicit val chartVersionEncoder: Encoder[UpdateAppJobId] = Encoder.encodeUUID.contramap(_.value)
+
+  implicit val getAppUpdateRequestEncoder: Encoder[UpdateAppsRequest] = Encoder.forProduct9(
+    "jobId",
     "appType",
     "cloudProvider",
     "appVersionsInclude",
@@ -20,6 +24,7 @@ object AdminRoutesTestJsonCodec {
     "dryRun"
   )(x =>
     (
+      x.jobId,
       x.appType,
       x.cloudProvider,
       x.appVersionsInclude,
--- a/core/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/AdminRoutesTestJsonCodec.scala
+++ b/core/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/AdminRoutesTestJsonCodec.scala
@@ -2,7 +2,6 @@ package org.broadinstitute.dsde.workbench.leonardo.http
 
 import io.circe.Encoder
 import org.broadinstitute.dsde.workbench.leonardo.JsonCodec._
-import org.broadinstitute.dsde.workbench.leonardo.UpdateAppJobId
 import org.broadinstitute.dsp.ChartVersion
 
 object AdminRoutesTestJsonCodec {
--- a/http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20240605_add_update_app_log_table.xml
+++ b/http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20240605_add_update_app_log_table.xml
@@ -30,8 +30,10 @@
         <createIndex indexName="FK_UPDATE_APP_LOG_ID" tableName="UPDATE_APP_LOG">
             <column name="id"/>
         </createIndex>
+        <createIndex indexName="FK_UPDATE_APP_LOG_JOB_ID" tableName="UPDATE_APP_LOG">
+            <column name="jobId"/>
+        </createIndex>
         <addUniqueConstraint columnNames="jobId, appId" constraintName="IDX_APP_ID_PER_JOB_UNIQUE" tableName="UPDATE_APP_LOG"/>
-        <addUniqueConstraint columnNames="appErrorId" constraintName="IDX_APP_ERROR_UNIQUE" tableName="UPDATE_APP_LOG"/>
         <addForeignKeyConstraint baseColumnNames="appId" baseTableName="UPDATE_APP_LOG" constraintName="FK_UPDATE_APP_LOG_APP_ID" deferrable="false" initiallyDeferred="false" onDelete="NO ACTION" onUpdate="NO ACTION" referencedColumnNames="id" referencedTableName="APP"/>
         <addForeignKeyConstraint baseColumnNames="appErrorId" baseTableName="UPDATE_APP_LOG" constraintName="FK_UPDATE_APP_LOG_APP_ERROR_ID" deferrable="false" initiallyDeferred="false" onDelete="NO ACTION" onUpdate="NO ACTION" referencedColumnNames="id" referencedTableName="APP_ERROR"/>
     </changeSet>
--- a/http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20240605_add_update_app_log_table.xml
+++ b/http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20240605_add_update_app_log_table.xml
@@ -11,6 +11,9 @@
             <column name="appId" type="BIGINT">
                 <constraints nullable="false"/>
             </column>
+            <column name="appErrorId" type="BIGINT">
+                <constraints nullable="true"/>
+            </column>
             <column name="status" type="ENUM('ERROR', 'RUNNING', 'SUCCESS')" defaultValue="RUNNING">
                 <constraints nullable="false"/>
             </column>
@@ -28,6 +31,8 @@
             <column name="id"/>
         </createIndex>
         <addUniqueConstraint columnNames="jobId, appId" constraintName="IDX_APP_ID_PER_JOB_UNIQUE" tableName="UPDATE_APP_LOG"/>
+        <addUniqueConstraint columnNames="appErrorId" constraintName="IDX_APP_ERROR_UNIQUE" tableName="UPDATE_APP_LOG"/>
         <addForeignKeyConstraint baseColumnNames="appId" baseTableName="UPDATE_APP_LOG" constraintName="FK_UPDATE_APP_LOG_APP_ID" deferrable="false" initiallyDeferred="false" onDelete="NO ACTION" onUpdate="NO ACTION" referencedColumnNames="id" referencedTableName="APP"/>
+        <addForeignKeyConstraint baseColumnNames="appErrorId" baseTableName="UPDATE_APP_LOG" constraintName="FK_UPDATE_APP_LOG_APP_ERROR_ID" deferrable="false" initiallyDeferred="false" onDelete="NO ACTION" onUpdate="NO ACTION" referencedColumnNames="id" referencedTableName="APP_ERROR"/>
     </changeSet>
 </databaseChangeLog>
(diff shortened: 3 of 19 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AdminServiceInterp.scala
```diff
@@ -60,27 +61,31 @@ final class AdminServiceInterp[F[_]: Parallel](authProvider: LeoAuthProvider[F],
...
         .transaction
-      responseList = ListUpdateableAppResponse.fromClusters(matchingApps).toVector
 
-      // If not a dry run, enqueue messages requesting app update.
-      _ <- {
-        if (req.dryRun)
-          F.unit
-        else {
-          val appNames = responseList.map(_.appName.value).mkString(", ")
-          log.info(
```
jdcanas, line 71: this log wasn't ever ran

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -1520,44 +1520,53 @@ class LeoPubsubMessageSubscriber[F[_]](
...
-                ErrorSource.App,
-                None,
-                Some(ctx.traceId)
-              )
-              _ <- appErrorQuery.save(msg.appId, error).transaction
-            } yield ()
+        .handleErrorWith { throwable =>
+          updateAppLogQuery
+            .update(msg.appId, msg.jobId, UpdateAppJobStatus.Error, Some(ctx.now))
+            .transaction >> (throwable match {
```
jdcanas: I don't love this syntax (aka flatmapping), but I think it is cleaner when chaining two `F[_]` together, as opposed to a nested for comprehension (which is needed when we have >2 `F[_]`s in my opinion)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/LeoProfile.scala
```diff
@@ -313,6 +313,20 @@ private[leonardo] object LeoProfile extends MySQLProfile {
 
     implicit val instantSetParameter: SetParameter[Instant] =
       SetParameter.SetTimestamp.contramap(instant => java.sql.Timestamp.from(instant))
+
+    implicit val appUpgradeTableIdColumnType: BaseColumnType[UpdateAppTableId] =
+      MappedColumnType.base[UpdateAppTableId, Long](_.value, UpdateAppTableId.apply)
+
+    implicit val appUpgradeJobIdColumnType: BaseColumnType[UpdateAppJobId] =
```
mlilynolting: Upgrade? should this be update

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/LeoProfile.scala
```diff
@@ -313,6 +313,20 @@ private[leonardo] object LeoProfile extends MySQLProfile {
 
     implicit val instantSetParameter: SetParameter[Instant] =
       SetParameter.SetTimestamp.contramap(instant => java.sql.Timestamp.from(instant))
+
+    implicit val appUpgradeTableIdColumnType: BaseColumnType[UpdateAppTableId] =
+      MappedColumnType.base[UpdateAppTableId, Long](_.value, UpdateAppTableId.apply)
```
mlilynolting, line 331: nit, the table names (UpdateApp) are backward from the variable names (appUpdateJob); is there a reason?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/LeoProfile.scala
```diff
@@ -313,6 +313,20 @@ private[leonardo] object LeoProfile extends MySQLProfile {
 
     implicit val instantSetParameter: SetParameter[Instant] =
       SetParameter.SetTimestamp.contramap(instant => java.sql.Timestamp.from(instant))
+
+    implicit val appUpgradeTableIdColumnType: BaseColumnType[UpdateAppTableId] =
```
mlilynolting: upgrade/update

mlilynolting:
Is there a design approach for this ticket? Adding a new database table is one way to get to a persisted job report but persistence is only in the ticket as a "perhaps," and it seems like saving an output file to a folder or bucket, or even just sending a report to Slack consistently, would satisfy the ACs. I don't think we save information on job runs in the db this way elsewhere in Leo.

mlilynolting:
I might just have missed a conversation though, in which case my apologies 😓

core/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/AdminRoutesTestJsonCodec.scala
```diff
@@ -1,15 +1,15 @@
...
-import io.circe.Encoder.AsArray.importedAsArrayEncoder
 import org.broadinstitute.dsde.workbench.leonardo.JsonCodec._
 import org.broadinstitute.dsp.ChartVersion
 
 object AdminRoutesTestJsonCodec {
 
   implicit val chartVersionEncoder: Encoder[ChartVersion] = Encoder.encodeString.contramap(_.asString)
 
-  implicit val getAppUpdateRequestEncoder: Encoder[UpdateAppsRequest] = Encoder.forProduct8(
+  implicit val getAppUpdateRequestEncoder: Encoder[UpdateAppsRequest] = Encoder.forProduct9(
```
mlilynolting: nit: this is an updateAppsRequest, right, plural?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/LeoProfile.scala
LizBaldo: We should pick one and stick with it, I do not really care which one but it looks like `Update` was the one previously used?

LizBaldo:
@mlilynolting here is the design doc: https://broadworkbench.atlassian.net/wiki/spaces/IA/pages/2906914878/Leonardo+App+Versioning+Management#App-Upgrade-Job

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
```diff
@@ -571,3 +571,25 @@ final case class GalaxyOrchUrl(value: String) extends AnyVal
...
 final case class KsaName(value: String) extends AnyVal
+
+final case class UpdateAppTableId(value: Long) extends AnyVal
+final case class UpdateAppJobId(value: UUID) extends AnyVal
+
+sealed abstract class UpdateAppJobStatus
+object UpdateAppJobStatus {
+
+  case object Running extends UpdateAppJobStatus {
+    override def toString: String = "RUNNING"
```
LizBaldo, line 608: These are fine, but could you update the design doc to reflect that?

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20240605_add_update_app_log_table.xml
```diff
@@ -0,0 +1,33 @@
+<databaseChangeLog logicalFilePath="leonardo" xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+    <changeSet logicalFilePath="leonardo" author="jcanas" id="add_update_app_log_table">
+
+        <createTable tableName="UPDATE_APP_LOG">
```
LizBaldo, line 4: So in the design doc, there is also a column to store the error message, I think it would be useful to have, unless the error is already stored somewhere else

LizBaldo (commented):
The approach looks very sound, and close to the design doc. Apart from the update/upgrade language confusion (my bad), I think the only missing piece is logging the error message in the table as well, that will make the reporting super easy to do.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/UpdateAppLogComponent.scala
```diff
@@ -0,0 +1,85 @@
...
+      logRecords.toList
+    }
+
+  def getByJobId(jobId: UpdateAppJobId)(implicit ec: ExecutionContext): DBIO[List[UpdateAppLogRecord]] =
+    updateAppLogQuery.filter(_.jobId === jobId).result map { recs =>
+      val logRecords = recs map { rec => unmarshallAppUpdateLogRecord(rec) }
+      logRecords.toList
+    }
+
+  def unmarshallAppUpdateLogRecord(appErrorRecord: UpdateAppLogRecord): UpdateAppLogRecord =
```
mlilynolting: super small nit: marshal/unmarshal, one L

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AdminServiceInterp.scala
```diff
@@ -5,19 +5,19 @@ package service
 import cats.Parallel
 import cats.effect.Async
 import org.broadinstitute.dsde.workbench.leonardo.config.KubernetesAppConfig
-import org.broadinstitute.dsde.workbench.leonardo.db.KubernetesServiceDbQueries
+import org.broadinstitute.dsde.workbench.leonardo.db._
```
mlilynolting: Could this be explicit imports instead? Wildcards give me the willies

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -1520,44 +1520,54 @@ class LeoPubsubMessageSubscriber[F[_]](
...
-              errorContext =
-                s"Error updating Azure app with id ${msg.appId.id} and cloudContext ${msg.cloudContext.asString}"
-              _ <- logger.warn(ctx.loggingCtx, e)(errorContext)
-              error = AppError(
-                s"${errorContext}: ${e.getMessage}",
-                ctx.now,
-                ErrorAction.UpdateApp,
-                ErrorSource.App,
-                None,
-                Some(ctx.traceId)
```
mlilynolting, line 1560: It looks like this traceId would be the only way to connect this specific error to our app-update record; we have the app ID but I don't think we would be able to tell which errors went with which update events. Would it make sense to include the traceId on the app update log, or maybe the appErrorId?

mlilynolting: I guess we could use the timestamp but that feels a bit muddy.

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20240605_add_update_app_log_table.xml
mlilynolting: There's an AppError object but I just realized that there's not a direct link between the two records; we might want to add a foreign key to one of these tables

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/LeoProfile.scala
jdcanas: ah yeah, I started with `appUpgrade` and switched to `UpdateApp` for consistency with existing variables

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20240605_add_update_app_log_table.xml
jdcanas: yeah, we already have an `APP_ERROR` table to store this information with a reason already corresponding to updates. I think it is much better to do a join on this table rather than duplicate this information in the database

LizBaldo: Definitely, I just do not see how to easily do that join for now, because appID alone is not enough right?

LizBaldo: I am fine either way, if you can modify the design doc accordingly so users know what to do we should be good 👍

jdcanas: it should be, the appId in the new table matches the field of the same name in `APP_ERROR` table as FK references to the APP table

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
jdcanas: Hmm... I was originally planning to sort by timestamp, but I can see your point. I was trying to avoid duplication of including the entire message, but I like your idea of including the app error id.

LizBaldo: +1 to adding the app error id, it will make the join to the error table a lot easier

mlilynolting (commented):
It'd be good to standardize naming to either `updateApp` or `appUpdate`.

core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala
LizBaldo: Thanks for updating the doc!

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20240605_add_update_app_log_table.xml
```diff
@@ -0,0 +1,38 @@
...
+            <column name="id" type="BIGINT" autoIncrement="true">
+                <constraints primaryKey="true"/>
+            </column>
+            <column name="jobid" type="UUID">
+                <constraints nullable="false"/>
+            </column>
+            <column name="appId" type="BIGINT">
+                <constraints nullable="false"/>
+            </column>
+            <column name="appErrorId" type="BIGINT">
```
LizBaldo, line 14: nice 👍

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20240605_add_update_app_log_table.xml
```diff
@@ -0,0 +1,38 @@
...
+                <constraints nullable="false"/>
+            </column>
+            <column name="startTime" type="TIMESTAMP(6)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="endTime" type="TIMESTAMP(6)">
+                <constraints nullable="true"/>
+            </column>
+        </createTable>
+        <createIndex indexName="FK_UPDATE_APP_LOG_APP_ID" tableName="UPDATE_APP_LOG">
```
LizBaldo, line 27: Should the appErrorId also be a FK?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/UpdateAppLogComponent.scala
```diff
@@ -0,0 +1,86 @@
...
+  def get(appId: AppId, jobId: UpdateAppJobId)(implicit ec: ExecutionContext): DBIO[Option[UpdateAppLogRecord]] =
+    updateAppLogQuery
+      .filter(_.appId === appId)
+      .filter(_.jobId === jobId)
+      .result map { recs =>
+      val logRecords = recs map { rec => unmarshalAppUpdateLogRecord(rec) }
+      logRecords.toList.headOption
+    }
+
+  def unmarshalAppUpdateLogRecord(appErrorRecord: UpdateAppLogRecord): UpdateAppLogRecord =
```
LizBaldo: nit, 'unmarshalUpdateAppLogRecord'

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala
```diff
@@ -1523,44 +1523,53 @@ class LeoPubsubMessageSubscriber[F[_]](
...
+            error = AppError(
+              s"${s"Error updating app with id ${msg.appId.id} and cloudContext ${msg.cloudContext.asString}"}: ${throwable.getMessage}",
+              ctx.now,
+              ErrorAction.UpdateApp,
+              ErrorSource.App,
+              None,
+              Some(ctx.traceId)
+            )
+            errorId <- appErrorQuery.save(msg.appId, error).transaction
+            _ <- updateAppLogQuery
```
LizBaldo, line 1558: 👍

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/db/UpdateAppLogComponentSpec.scala
```diff
@@ -0,0 +1,179 @@
...
+    val test = for {
+      errorId <- appErrorQuery.save(savedApp.id, appError)
+      _ <- updateAppLogQuery.save(jobId1, savedApp.id, startTime)
+      endTime = Instant.now()
+      _ <- updateAppLogQuery.update(savedApp.id, jobId1, UpdateAppJobStatus.Error, Some(errorId), Some(endTime))
+      appUpdateLogRecordOpt <- updateAppLogQuery.get(savedApp.id, jobId1)
+    } yield {
+      appUpdateLogRecordOpt.isDefined shouldBe true
+      appUpdateLogRecordOpt.map(_.startTime.toEpochMilli) shouldBe Some(startTime.toEpochMilli)
+      appUpdateLogRecordOpt.map(_.errorId) shouldBe Some(Some(errorId))
```
LizBaldo, line 63: Just curious, but is there any way to collapse the Some(Some( ? That might be a scala thing but it looks fishy to me

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AdminServiceInterpSpec.scala
```diff
@@ -73,27 +74,110 @@ final class AdminServiceInterpSpec extends AnyFlatSpec with LeonardoTestSuite wi
...
 
     // Verify that the update message was sent
     val message = publisherQueue.take.unsafeRunSync()(cats.effect.unsafe.IORuntime.global)
     message.messageType shouldBe LeoPubsubMessageType.UpdateApp
     val updateMsg = message.asInstanceOf[UpdateAppMessage]
-    updateMsg.appName shouldBe AppName("app1")
+    updateMsg.appName shouldBe app1.appName
+    updateMsg.jobId shouldBe jobId
+
+    // Verify database has log records for update
```
LizBaldo, line 103: Nice, thanks for extending the existing tests

LizBaldo (approved):
Looks great to me, I only have a few minor comments but this is a great improvement to the UpgradeApp flow, thank you!

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20240605_add_update_app_log_table.xml
```diff
@@ -0,0 +1,38 @@
...
+            <column name="appId"/>
+        </createIndex>
+        <createIndex indexName="FK_UPDATE_APP_LOG_ID" tableName="UPDATE_APP_LOG">
+            <column name="id"/>
+        </createIndex>
+        <addUniqueConstraint columnNames="jobId, appId" constraintName="IDX_APP_ID_PER_JOB_UNIQUE" tableName="UPDATE_APP_LOG"/>
+        <addUniqueConstraint columnNames="appErrorId" constraintName="IDX_APP_ERROR_UNIQUE" tableName="UPDATE_APP_LOG"/>
```
mlilynolting: if appErrorId were a foreign key I think this could be removed

jdcanas: ah yeah, I did this when I was experimenting with a weird issue with the FK constraint that turned about to be needing to add dumping the new table to the test `isolatedDbTest` tag

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20240605_add_update_app_log_table.xml
jdcanas: ah yeah, I did this when I was experimenting with a weird issue with the FK constraint that turned about to be needing to add dumping the new table to the test `isolatedDbTest` tag, need to add it back for sure

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/db/UpdateAppLogComponentSpec.scala
jdcanas: this is the fact that the db query returns an optional record, and that field is itself optional. I could do a `.get` to remove the nesting rather than mapping the field inside the option if you prefer, but its generally better practice to avoid `.get` as it can cause runtime errors (mute in this case though tbh since we check if its defined on line 61)

mlilynolting (approved):

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20240605_add_update_app_log_table.xml
jdcanas: oh I already added in back!

http/src/main/resources/org/broadinstitute/dsde/workbench/leonardo/liquibase/changesets/20240605_add_update_app_log_table.xml
mlilynolting: looks like adding this explicitly isn't wrong tho, foreign keys apparently don't have to be unique. but mb since it's based on a unique id in the other table it's not needed here

PR #4618, opened 2024-06-11
[IA-4958] 3rd Party Cookie Support via CHIPS

LizBaldo:
__Jira ticket__: https://broadworkbench.atlassian.net/browse/IA-4813

### What

-

### Why

-

### What to test

- 

### Who tested and where

- [ ] I validated this change 


http/src/main/resources/init-resources/cluster-site-gce.conf
```diff
@@ -34,7 +34,7 @@
     # Append SameSite=None to cookies set by RStudio. This is required by some browsers because we
     # render RStudio in an iframe. There does not appear to be a way within RStudio to do this, hence
     # doing it in the proxy.
-    Header edit Set-Cookie ^(.*)$ $1;Secure;SameSite=None;HttpOnly "expr=%{REQUEST_URI} =~ m#/proxy/[^/]*/[^/]*/rstudio/.*#"
+    Header edit Set-Cookie ^(.*)$ $1;Secure;SameSite=None;HttpOnly;Partitioned "expr=%{REQUEST_URI} =~ m#/proxy/[^/]*/[^/]*/rstudio/.*#"
```
rtitle, line 37: I forgot we do this here. I wonder if it's still necessary, given that we also set `SameSite=None` in the Set-Cookie in the Leo proxy.

LizBaldo: Yeah I was not quite sure either, I can test without it, but thought it might be nice to keep it consistent

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/BuildHelmChartValues.scala
```diff
@@ -376,7 +376,7 @@ private[leonardo] object BuildHelmChartValues {
       raw"""ingress.annotations.nginx\.ingress\.kubernetes\.io/auth-tls-secret=${namespaceName.value}/ca-secret""",
       raw"""ingress.annotations.nginx\.ingress\.kubernetes\.io/proxy-redirect-to=${leoProxyhost}${ingressPath}""",
       raw"""ingress.annotations.nginx\.ingress\.kubernetes\.io/rewrite-target=/${rewriteTarget}""",
-      raw"""ingress.annotations.nginx\.ingress\.kubernetes\.io/proxy-cookie-path=/ "/; Secure; SameSite=None; HttpOnly"""",
+      raw"""ingress.annotations.nginx\.ingress\.kubernetes\.io/proxy-cookie-path=/ "/; Secure; SameSite=None; HttpOnly; Partitioned"""",
```
rtitle, line 379: Ah, I guess this is an nginx config to change the path of a set-cookie header: https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#proxy-cookie-path

Maybe it's needed to ensure that the AoU apps can set cookies (probably different cookies than our auth cookie).

rtitle (approved):
This looks correct to me! As discussed, we should probably try to verify our full suite of GCP apps.

(On the Azure side, cookie setting is done by the Relay Listener.)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/BuildHelmChartValues.scala
LizBaldo: @Qi77Qi would you know if the nginx proxy-cookie-path used by AOU apps would need to be modified like the leo proxy cookies?

LizBaldo:
@rtitle do you remember if Galaxy or Cromwell apps were working when you tested on Dev? I am running into authentication issues for both on the custom BEE with the different UI domain:

```{
  "source": "leonardo",
  "message": "Your account is not authenticated",
  "statusCode": 401,
  "exceptionClass": "class org.broadinstitute.dsde.workbench.leonardo.model.AuthenticationError",
  "traceId": null
}
```

I am trying to understand if it is a genuine issue with CHIPS, or if it an issue with the BEE setup

CC @cpate4 as testing on a BEE is honestly proving more challenging than doing it in dev at this point.

rtitle:
I don't think I tested Galaxy on dev :-/

Maybe there is some additional galaxy dependency on cookies -- I can take a quick look on the BEE.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/BuildHelmChartValues.scala
Qi77Qi: I think it's for setting cookie path that comes from SAS...so probably also needed?

Qi77Qi: lmk when this hits dev and we can test it out..

Qi77Qi (approved):

LizBaldo:
> I don't think I tested Galaxy on dev :-/
> 
> Maybe there is some additional galaxy dependency on cookies -- I can take a quick look on the BEE.

No worries, I think it is an issue with the BEE setup, because Galaxy and Cromwell are also not working on the BEE after re-enabling 3rd party cookie support on my browser

rtitle:
> No worries, I think it is an issue with the BEE setup, because Galaxy and Cromwell are also not working on the BEE after re-enabling 3rd party cookie support on my browser

That seems plausible. I can't remember the last time I tried to make a Galaxy from a BEE (maybe never?)

cpate4 (approved):

PR #4627, opened 2024-06-12
[IA-4933] Fix GPU test

lucymcnatt:

### What

- Changes how the GPU machine device names are checked in the GPU test

### Why

- This assertion was previously failing due to some out the output being cut off

### What to test

- 

### Who tested and where

- [ ] I validated this change 


automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/runtimes/RuntimeGceSpec.scala
```diff
@@ -114,15 +115,14 @@ class RuntimeGceSpec
         implicit0(authToken: AuthToken) <- Ron.authToken()
         _ <- IO(withWebDriver { implicit driver =>
           withNewNotebook(clusterCopy, Python3) { notebookPage =>
+            notebookPage.executeCell("import tensorflow as tf")
```
lucymcnatt, line 117: for some reason the test was previously trying to evaluate if the gpu string was in the output from this import step (instead of where we actually list the devices), so I pulled this step out

LizBaldo: Oh whoa this is the weirdest race condition, good find!

LizBaldo (approved):

mlilynolting (approved):
nice!

PR #4629, opened 2024-06-14
ID-1266: Assign azure_private_storage_account action managed identities to Azure VMs 

rtitle:
JIRA: https://broadworkbench.atlassian.net/browse/ID-1266

Design doc: https://docs.google.com/document/d/1VT__hmPH1cyYYm643cJzlsrVIyKeWPgEnjpC0fjV63E/edit

## Dependencies

Depends on WSM (https://github.com/DataBiosphere/terra-workspace-manager/pull/1754) and Welder (https://github.com/DataBiosphere/welder/pull/270/).

### What

- This PR updates Back Leo VM creation to retrieve an action managed identity from Sam for the `private_azure_storage_account/read` action. If one is present, pass it to the WSM VM creation API.
- Also updates welder hash for https://github.com/DataBiosphere/welder/pull/270

### Why

- See design doc above. This supports the Enterprise Data Access team-sized problem. These AAMIs will be used to grant access to external resources (storage accounts, ACRs).

### What to test

- Unit testing
- Automation tests cover VM creation with no AAMI - confirm no breakages
- Manual testing on a BEE:
   - Created an AAMI for a user. Created a notebook VM, verified it starts correctly and the user has access to the AAMI (and the external storage account).
   - Created a notebook VM without an AAMI, verified things still work with no change in behavior.

### Who tested and where

- [x] I validated this change in a BEE

### DIFF
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -885,7 +885,7 @@ gke {
 image {
   welderGcrUri = "us.gcr.io/broad-dsp-gcr-public/welder-server"
   welderDockerHubUri = "broadinstitute/welder-server"
-  welderHash = "2a3d22b"
+  welderHash = "fcb50d3"
   jupyterImage =  "us.gcr.io/broad-dsp-gcr-public/terra-jupyter-gatk:2.3.6"
   proxyImage = "broadinstitute/openidc-proxy:2.3.1_2"
   # Note: If you update this, please also update prepare_gce_image.sh and
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -885,7 +885,7 @@ gke {
 image {
   welderGcrUri = "us.gcr.io/broad-dsp-gcr-public/welder-server"
   welderDockerHubUri = "broadinstitute/welder-server"
-  welderHash = "2a3d22b"
+  welderHash = "51ef0f4"
   jupyterImage =  "us.gcr.io/broad-dsp-gcr-public/terra-jupyter-gatk:2.3.6"
   proxyImage = "broadinstitute/openidc-proxy:2.3.1_2"
   # Note: If you update this, please also update prepare_gce_image.sh and
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -885,7 +885,7 @@ gke {
 image {
   welderGcrUri = "us.gcr.io/broad-dsp-gcr-public/welder-server"
   welderDockerHubUri = "broadinstitute/welder-server"
-  welderHash = "2245db0"
+  welderHash = "2a3d22b"
   jupyterImage =  "us.gcr.io/broad-dsp-gcr-public/terra-jupyter-gatk:2.3.6"
   proxyImage = "broadinstitute/openidc-proxy:2.3.1_2"
   # Note: If you update this, please also update prepare_gce_image.sh and
--- a/http/src/main/resources/reference.conf
+++ b/http/src/main/resources/reference.conf
@@ -885,7 +885,7 @@ gke {
 image {
   welderGcrUri = "us.gcr.io/broad-dsp-gcr-public/welder-server"
   welderDockerHubUri = "broadinstitute/welder-server"
-  welderHash = "fcb50d3"
+  welderHash = "2a3d22b"
   jupyterImage =  "us.gcr.io/broad-dsp-gcr-public/terra-jupyter-gatk:2.3.6"
   proxyImage = "broadinstitute/openidc-proxy:2.3.1_2"
   # Note: If you update this, please also update prepare_gce_image.sh and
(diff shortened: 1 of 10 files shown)


project/Dependencies.scala
```diff
@@ -133,7 +133,7 @@ object Dependencies {
   val pact4sCirce =       "io.github.jbwheatley"  %% "pact4s-circe"     % pact4sV
   val okHttp =            "com.squareup.okhttp3"  % "okhttp"            % "4.12.0"
 
-  val workSpaceManagerV = "0.254.1093-SNAPSHOT"
+  val workSpaceManagerV = "0.254.1121-rt-aami-SNAPSHOT"
```
rtitle: TODO: update after https://github.com/DataBiosphere/terra-workspace-manager/pull/1754 is merged

http/src/main/resources/reference.conf
```diff
@@ -885,7 +885,7 @@ gke {
 image {
   welderGcrUri = "us.gcr.io/broad-dsp-gcr-public/welder-server"
   welderDockerHubUri = "broadinstitute/welder-server"
-  welderHash = "6648f5c"
+  welderHash = "2a3d22b"
```
rtitle: TODO: update after https://github.com/DataBiosphere/welder/pull/270 is merged

project/Dependencies.scala
```diff
@@ -18,7 +18,7 @@ object Dependencies {
   val pact4sV = "0.10.0"
 
   private val workbenchLibsHash = "3b2d0f4"
-  val serviceTestV = s"4.3-$workbenchLibsHash"
+  val serviceTestV = s"4.4-dbf75d26-SNAP"
```
rtitle: TODO: update after https://github.com/broadinstitute/workbench-libs/pull/1690 is merged

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
```diff
@@ -146,6 +148,17 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
         CreateAzureDiskParams(msg.workspaceId, runtime, msg.useExistingDisk, azureConfig)
       )
 
+      // Get optional action managed identity from Sam for the private_azure_storage_account/read action.
+      // Identities must be passed to WSM for application-managed resources.
+      tokenOpt <- samDAO.getCachedArbitraryPetAccessToken(runtime.auditInfo.creator)
```
LizBaldo, line 153: Where is this tokenOpt used?

LizBaldo (commented):
The logic to grab the private storage account MI and pass it to WSM at VM creation looks sound to me. My only question is how a user will then authenticate within their notebook between the two identities available to them, the original Pet and the optional private storage account one.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
rtitle: L154, `tokenOpt.flatTraverse`

Basically, if it's present, then using it to make the Sam call (which itself returns `IO[Option[..]]`)

rtitle:
> how a user will then authenticate within their notebook between the two identities available to them, the original Pet and the optional private storage account one.

Yeah good question -- so first of all, there is no change in behavior when there are no `azure_private_storage_account` resources present (which there won't be for a while, UX team is designing the UI that will let users manage these resources).

When there _are_ multiple identities present, the change in user experience is that `az login --identity` doesn't work anymore. You need to do `az login --identity -u <identity-resource-id>` to log in with either the pet or the action managed identity. It works as of this PR -- but there are probably ways to smooth out the experience in the future (e.g. via `tnu` utility and/or environment variables).

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
LizBaldo: Oh boy my bad, sorry about that 🤦‍♀️

LizBaldo (approved):
LGTM, thanks for the clarifications!

rtitle:
Update: I would still like to merge this, to enable the notebooks-private endpoint integration. This PR will be a no-op because the feature isn't enabled, and TBD when/if it will be enabled by default. But we would still like to tie this up so it can be demoed, etc.

project/Dependencies.scala
```diff
@@ -17,7 +17,7 @@ object Dependencies {
   val munitCatsEffectV = "1.0.7"
   val pact4sV = "0.10.0"
 
-  private val workbenchLibsHash = "a6ad7dc"
+  private val workbenchLibsHash = "b57a1ca"
```
rtitle, line 20: Needed for `jackson-module-scala` update, to be compatible with the new WSM client.

cpate4 (approved):

PR #4631, opened 2024-06-14
AJ-1894 pass concurrent job limit to CRA

calypsomatic:
__Jira ticket__: https://broadworkbench.atlassian.net/browse/AJ-1894

Depends on https://github.com/broadinstitute/terra-helmfile/pull/5693 to add the BPM url to leo and https://github.com/broadinstitute/terra-helmfile/pull/5641 for cromwell to actually read in the concurrent job limit parameter.

Am happy to hear input on additional tests I could add - the lines pointed out by sonar don't seem particularly in need of additional coverage but I welcome advice.

### What

- Adds BPM to Leo.  Calls BPM when launching a cromwell runner app to see if there is a concurrentjoblimit key in the limits map.  If so, adds it to CRA's helm values.

### Why

- For AnVIL Lite billing projects, we want to limit how fast users can spend.

### What to test

- 

### Who tested and where
I verified this works in my BEE.  I made sure BPM had a limits map with a "concurrent-job-limit" key for my billing project.  Then, with this version of Leo, I started a workflow in a new workspace, launching a new cromwell-runner-app.  Using `kubectl`, I checked the config map of the CRA deployment and saw that the `concurrent-job-limit` value was set.

- [ ] I validated this change 


Dockerfile
```diff
@@ -54,6 +54,9 @@ RUN helm repo add ingress-nginx https://kubernetes.github.io/ingress-nginx && \
     helm repo add terra-helm https://terra-helm.storage.googleapis.com && \
     helm repo update
 
+COPY ./cromwell-runner-app-0.159.4.tgz /leonardo
+RUN tar -xzf /leonardo/cromwell-runner-app-0.159.4.tgz -C /leonardo
+
```
davidangb: is this a temporary change?

cromwell-runner-app-0.159.4.tgz
davidangb, line 1: I assume this also will not be in the final PR?

http/src/main/resources/reference.conf
```diff
@@ -367,8 +371,8 @@ azure {
 
   cromwell-runner-app-config {
     instrumentation-enabled = false
-    chart-name = "terra-helm/cromwell-runner-app"
-    chart-version = "0.157.0"
+    chart-name = "leonardo/cromwell-runner-app"
+    chart-version = "0.159.4"
```
davidangb: also temp?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/app/CromwellRunnerAppInstall.scala
```diff
@@ -79,6 +84,26 @@ class CromwellRunnerAppInstall[F[_]](config: CromwellRunnerAppConfig,
...
+              val scalaLimits = limits.asScala
+              val value = scalaLimits.get("maxconcurrentworkflows")
+              value match {
+                case Some(v) => F.pure(Some(raw"config.maxConcurrentWorkflows=${v}"))
+                case None    => F.pure(None)
+              }
+            case _ => F.pure(None)
+          }
+        case _ => F.pure(None)
+      }
```
davidangb, line 100: you can use flatMap/map to unnest all the various options in this block. I think this can simplify to:
```scala
      maybeLimits = maybeProfile.flatMap(
        _.getOrganization.getLimits.asScala
          .get("maxconcurrentworkflows")
          .map(v => raw"config.maxConcurrentWorkflows=${v}")
      )
```

note my solution is slightly different - you are handling a case where `billingProfile.getOrganization.getLimits` returns null. Can this actually return null?

For `Option`s, `.map` means "if the current value is Some, do the next block of code, else return None" and the leading `flatMap` means "don't return Option[Option[Option[something]]], flatten them all out"

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/app/CromwellRunnerAppInstall.scala
```diff
@@ -133,7 +158,13 @@ class CromwellRunnerAppInstall[F[_]](config: CromwellRunnerAppConfig,
         raw"sam.baseUri=${samConfig.server}",
         raw"sam.acrPullActionIdentityResourceId=${params.billingProfileId.value}"
       )
-    } yield Values(values.mkString(","))
+
+      finalList = maybeLimits match {
+        case Some(str) => values ++ List(str)
```
davidangb: you shouldn't have to wrap in a `List` here, you can just append a single element:
```scala
case Some(str) =>  values :+ str
```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/app/CromwellRunnerAppInstall.scala
```diff
@@ -147,6 +178,9 @@ class CromwellRunnerAppInstall[F[_]](config: CromwellRunnerAppConfig,
      getAzureDatabaseName(dbResources, "tes"),
      getAzureDatabaseName(dbResources, "cromwellmetadata")
     ).mapN(CromwellRunnerAppDatabaseNames)
+
+  def stringToUUID(s: String): Try[UUID] =
+    Try(UUID.fromString(s))
```
davidangb: this is only used once; could you just inline it?

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/app/CromwellRunnerAppInstallSpec.scala
```diff
@@ -69,6 +77,49 @@ class CromwellRunnerAppInstallSpec extends BaseAppInstallSpec {
...
+      "postgres.user=ksa-1," +
+      s"postgres.dbnames.cromwell=$cromwellAzureDbName," +
+      s"postgres.dbnames.tes=$tesAzureDbName," +
+      s"postgres.dbnames.cromwellMetadata=$cromwellMetadataAzureDbName," +
+      s"ecm.baseUri=https://externalcreds.dsde-dev.broadinstitute.org," +
+      s"sam.baseUri=https://sam.test.org:443," +
+      s"sam.acrPullActionIdentityResourceId=$bpid," +
+      s"config.maxConcurrentWorkflows=100"
+  }
+
```
davidangb, line 91: looks like a lot of copy/paste with the previous test; could you extract an "expected" override string for use in both tests, then append "config.maxConcurrentWorkflows=100" for this test?

davidangb (commented):
comments inline!

Dockerfile
calypsomatic: Yes

cromwell-runner-app-0.159.4.tgz
calypsomatic: Correct

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/app/CromwellRunnerAppInstall.scala
```diff
@@ -79,6 +84,17 @@ class CromwellRunnerAppInstall[F[_]](config: CromwellRunnerAppConfig,
         AppCreationException(s"Pet not found for user ${params.app.auditInfo.creator}", Some(ctx.traceId))
       )
 
+      maybeProfile <- Try(UUID.fromString(params.billingProfileId.value)) match {
+        case Success(uuid) => bpmClient.getProfile(userToken, uuid)
+        case _             => F.pure(None)
+      }
```
davidangb, line 94: you could do this with `map` and `getOrElse` too:
```scala
maybeProfile <- Try(UUID.fromString(params.billingProfileId.value))
        .map(uuid => bpmClient.getProfile(userToken, uuid))
        .getOrElse(F.pure(None))
```
I have no preference, whichever one you find easier to read

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/BpmApiClientProvider.scala
```diff
@@ -0,0 +1,46 @@
...
+
+}
+
+class HttpBpmClientProvider[F[_]](baseBpmUrl: Uri)(implicit F: Async[F]) extends BpmApiClientProvider[F] {
+  private def getApiClient(token: String): F[ApiClient] = {
+    val client = new ApiClient()
+    client.setBasePath(baseBpmUrl.renderString)
+    client.setAccessToken(token)
+    F.pure(client)
+  }
```
davidangb: from our in-person discussion: this is missing the span/tracing configuration that currently exists in `HttpWsmClientProvider`. Do we need to figure out how to get that working?

davidangb (commented):
question about tracing inline, otherwise looking good and thank you!

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/BpmApiClientProvider.scala
calypsomatic: Ah yeah - I think I should be able to keep it like it was before now that I found what the problem was

calypsomatic: Ok I put it back in as I had originally copied it - verified that Leo can still fetch the limits map from BPM but don't know how to check that span/tracing is working correctly.  It at least should be the same as in Wsm.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/app/CromwellRunnerAppInstall.scala
jdcanas: nit: I'd recommend `Either.catchNonFatal` over `try` to be more consistent with the rest of the codebase to safely wrap `.fromString`. There should be numerous examples you can model the syntax after. `Either` plays nicely with for comprehensions too such that you should be able to avoid using a `match` to extract the success case .

Something like 
```
parsedUUID <- F.delay(Either.catchNotFatal(UUID.fromString(params.billingProfileId.value))))
profileAttempt <- parsedUUID.traverse { uuid =>
  bpmClient.getProfile(userToken, uuid)
}
limits <- profileAttempt.map(getOrganization.getLimits.asScala
          .get("concurrentjoblimit")
          .map(v => raw"config.concurrentJobLimit=${v}"))

```

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/BpmApiClientProvider.scala
```diff
@@ -0,0 +1,60 @@
...
+          }
+        }
+      }
+      _ = client.setBasePath(baseBpmUrl.renderString)
+      _ = client.setAccessToken(token)
+    } yield client
+//    val client = new ApiClient()
+//    client.setBasePath(baseBpmUrl.renderString)
+//    client.setAccessToken(token)
+//    F.pure(client)
```
jdcanas: can be deleted likely?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/app/CromwellRunnerAppInstall.scala
```diff
@@ -79,6 +84,17 @@ class CromwellRunnerAppInstall[F[_]](config: CromwellRunnerAppConfig,
         AppCreationException(s"Pet not found for user ${params.app.auditInfo.creator}", Some(ctx.traceId))
       )
 
+      maybeProfile <- Try(UUID.fromString(params.billingProfileId.value)) match {
+        case Success(uuid) => bpmClient.getProfile(userToken, uuid)
```
rtitle: Just calling out that this is using the pet token, which is probably not correct (the user might not have access to the billing project/profile). @dvoet has been working on resource-type-admin authorization for service-to-service calls. Should Leo use that instead for this use case?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/BpmApiClientProvider.scala
calypsomatic: Whoops!  Yep!

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/app/CromwellRunnerAppInstall.scala
calypsomatic: I was worried about that and tried it out with a user that didn't have access to the billing project and it worked - but I'm happy to switch if I knew how

rtitle: That’s odd, I wouldn’t expect it to work. :) How did you test exactly — I’m picturing something like this:
1. As user A, create billing project and workspace
2. As user A, make user B a writer on the workspace 
3. As user B, run a workflow in the workspace — Cromwell runner app creation should fail 

To use the resource-type-admin feature, I think you’d just use the Leo SA token instead of the pet (I think there is a method getLeoToken). There may be some more wiring needed to grant the Leo SA the right permissions on billing profiles.

calypsomatic: That's exactly what I did and it did for some reason, somehow, work.  I'll look into granting Leo the permissions.

calypsomatic: Turns out it was because BPM grants read permission to anyone with any access whatsoever to the billing profile.  https://broadworkbench.atlassian.net/browse/WOR-1769 fixed that but left read profile access to avoid changing current behavior.

http/src/main/resources/reference.conf
```diff
@@ -368,7 +372,7 @@ azure {
   cromwell-runner-app-config {
     instrumentation-enabled = false
     chart-name = "terra-helm/cromwell-runner-app"
-    chart-version = "0.164.0"
+    chart-version = "0.167.0"
```
davidangb: is this supposed to be part of the PR?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/app/CromwellRunnerAppInstall.scala
```diff
@@ -79,6 +86,21 @@ class CromwellRunnerAppInstall[F[_]](config: CromwellRunnerAppConfig,
...
+        bpmClient.getProfile(leoAuth, uuid)
+      }
+
+      maybeLimits = profileAttempt match {
+        case Right(Some(profile)) =>
+          profile.getOrganization.getLimits.asScala
+            .get("concurrentjoblimit")
+            .map(v => raw"config.concurrentJobLimit=${v}")
+        case _ => None
+      }
```
davidangb: after poking at this a bit, you could probably reduce this some:
```scala
      maybeLimits = profileAttempt.toOption.flatten.flatMap { profile =>
        profile.getOrganization.getLimits.asScala
          .get("concurrentjoblimit")
          .map(v => raw"config.concurrentJobLimit=${v}")
      }
```

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/ConfigReaderSpec.scala
```diff
@@ -148,7 +149,7 @@ class ConfigReaderSpec extends AnyFlatSpec with Matchers {
         ),
         CromwellRunnerAppConfig(
           ChartName("terra-helm/cromwell-runner-app"),
-          ChartVersion("0.164.0"),
+          ChartVersion("0.167.0"),
```
davidangb: wondering if this is also a necessary part of the PR

davidangb (approved):
a style comment inline, but I defer to IA on style inside Leo

http/src/main/resources/reference.conf
calypsomatic: Yes - I need the CRA chart with my changes, and previous automerges of CRA have all failed.

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/http/ConfigReaderSpec.scala
calypsomatic: Yep!  It goes with the other change - this test fails without it.  If the latest CRA chart gets merged in before this PR does, then I won't need it anymore, but until then, necessary.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/app/CromwellRunnerAppInstall.scala
calypsomatic: Done

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/app/CromwellRunnerAppInstall.scala
calypsomatic: Done

mlilynolting (approved):

PR #4643, opened 2024-06-17
[Prod-968] debug r version

jdcanas:


http/src/main/resources/init-resources/azure_vm_init_script.sh
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-set -e
+#set -e
```
LizBaldo: Could you uncomment this?

jdcanas: yeah, I have that locally, just waiting on this run of test

http/src/main/resources/init-resources/azure_vm_init_script.sh
```diff
@@ -176,15 +176,6 @@ while sudo fuser /var/lib/dpkg/lock-frontend > /dev/null 2>&1
     sleep 5
   done
 
-# Install updated R version
-echo "Installing R version ${R_VERSION}"
```
LizBaldo, line 180: Could you also remove the R_VERSION variable that is set above?

LizBaldo (approved):
I have a couple of comments to address, but 👍 to unblock you

lucymcnatt (approved):
LGTM!

PR #4666, opened 2024-06-25
RW-12822 add workload separation per AppSec request

Qi77Qi:
__Jira ticket__: https://precisionmedicineinitiative.atlassian.net/browse/RW-12822

Followed https://cloud.google.com/kubernetes-engine/docs/how-to/workload-separation#separate-workloads-autopilot

Verified in BEE that apps live on different nodes
```
➜  Github kubectl get pods --all-namespaces -o wide|grep "\-app\-ns"
ihzdh1-app-ns     qi-rstudio-6-24-2-74ffb9ddd5-sbfr5                         0/3     ContainerCreating   0              2m17s   <none>       gk3-k51b094f-2d16-4bbf-8-nap-1rqpofco-a3da0221-g9s4   <none>           <none>
ovvehz-app-ns     qi-rstudio-6-24-1-7866996477-ld82r                         0/3     ContainerCreating   0              5m14s   <none>       gk3-k51b094f-2d16-4bbf-8-nap-hcblhnre-06dceee2-mvcg   <none>           <none>
```

### What

-

### Why

-

### What to test

- 

### Who tested and where

- [ ] I validated this change 

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
@@ -89,14 +89,6 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
       _ <- F.raiseWhen(!hasPermission)(ForbiddenError(userInfo.userEmail))
 
       enableIntraNodeVisibility = req.labels.get(AOU_UI_LABEL).exists(x => x == "true")
-      // We only allow autopilot mode for members of CUSTOM_APP_USERS group
-      // This is only needed while we're evaluating autopilot (6/3/2024).
-      _ <- req.autopilot.traverse { _ =>
-        authProvider.isCustomAppAllowed(userInfo.userEmail) map { res =>
-          if (res) Right(())
-          else Left("You don't have permission to create APP in autopilot mode")
-        }
-      }
       _ <- req.appType match {
         case AppType.Galaxy | AppType.HailBatch | AppType.Wds | AppType.Cromwell | AppType.WorkflowsApp |
             AppType.CromwellRunnerApp =>
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/BuildHelmChartValues.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/BuildHelmChartValues.scala
@@ -395,7 +395,8 @@ private[leonardo] object BuildHelmChartValues {
     // spec.template.spec.tolerations[0].operator: Invalid value: "xxx": a valid label must be an empty string or
     // consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character
     // (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')], string=
-    val base64EncodedEmail = Base64.getEncoder.encodeToString(userEmail.value.getBytes(StandardCharsets.UTF_8))
+    val base64EncodedEmail =
+      Base64.getEncoder.encodeToString(userEmail.value.getBytes(StandardCharsets.UTF_8)).replace("=", "")
     val nodeSelectorGroupValue = s"leonardo_${base64EncodedEmail}"
     val autopilotParams = autopilot match {
       case Some(v) =>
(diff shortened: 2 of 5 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
```diff
@@ -89,14 +89,6 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
       _ <- F.raiseWhen(!hasPermission)(ForbiddenError(userInfo.userEmail))
 
       enableIntraNodeVisibility = req.labels.get(AOU_UI_LABEL).exists(x => x == "true")
-      // We only allow autopilot mode for members of CUSTOM_APP_USERS group
```
Qi77Qi, line 92: removing this becuz this was added per @dinvlad 's request without workload separation

yonghaoy (approved):

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/BuildHelmChartValues.scala
```diff
@@ -385,9 +385,18 @@ private[leonardo] object BuildHelmChartValues {
       raw"""ingress.tls[0].hosts[0]=${k8sProxyHostString}"""
     )
 
+    // a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character
```
LizBaldo, line 392: What does this comment correspond to in the code?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
LizBaldo: So from now, we can try to create other Terra GKE apps using autopilot (mainly Cromwell and Galaxy), but we would need to do similar work on the app charts like what you had to do for the SAS and RStudio ones, correct?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/BuildHelmChartValues.scala
```diff
@@ -385,9 +385,18 @@ private[leonardo] object BuildHelmChartValues {
...
+    // (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?'),
+    // spec.template.spec.tolerations[0].operator: Invalid value: "xxx": a valid label must be an empty string or
+    // consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character
+    // (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')], string=
+    val nodeSelectorGroupValue = s"a${userEmail.value.split("@")(0)}"
     val autopilotParams = autopilot match {
       case Some(v) =>
         val ls = List(
+          raw"""tolerations.enabled=true""",
```
LizBaldo, line 406: Adding a comment here explaining how this ensures workload separation would be useful

LizBaldo (approved):
I do not have access to the JIRA ticket and so if you can add a few more clarifying comments in the code that would be super useful.

The change looks straightforward though so 👍

Qi77Qi:
@LizBaldo the ticket doesn't have too much info..but see [this](https://docs.google.com/document/d/1nt4OCe7btcybAeOxI1ePlauU3nlOEgESekQOCPMLkfU/edit#heading=h.hbd0urfk0ufj)...lmk if this still doesn't clarify

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/BuildHelmChartValues.scala
Qi77Qi: I'll add a bit more clarification. But this is about naming convention for the `nodeSelectorGroupValue` below

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/BuildHelmChartValues.scala
Qi77Qi: will do

LizBaldo:
> @LizBaldo the ticket doesn't have too much info..but see [this](https://docs.google.com/document/d/1nt4OCe7btcybAeOxI1ePlauU3nlOEgESekQOCPMLkfU/edit#heading=h.hbd0urfk0ufj)...lmk if this still doesn't clarify

Thanks! I forgot to check there, this explains it https://cloud.google.com/kubernetes-engine/docs/how-to/workload-separation#separate-workloads-autopilot

Qi77Qi:
@LizBaldo some charts already set resources block, so no changes for these charts are needed (like the nginx one). I'll work on cromwell to add this resource block. I don't know the status for Galaxy. It's also possible that if the resource block isn't defined, there's defaults used, so potentially Galaxy will still work

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/BuildHelmChartValues.scala
dinvlad: Only concern if `userEmail.value.split("@")(0)` may have collisions (e.g. two users with the same username but different domains). Might be better to append a short hash of that email maybe?

PR #4672, opened 2024-06-25
[IA-5000] Make sure to transition DELETING Azure disks to ERROR

LizBaldo:
__Jira ticket__: https://broadworkbench.atlassian.net/browse/IA-5000

### What

- Ensures that when an Azure runtime deletion fails, both the runtime AND the persistent disk (if provided) transition their statuses to ERROR

### Why

- Even though persistent disks where being deleted properly in the Azure portal, they were still marked as DELETING in the Leonardo database (and therefore the cloud environment page in the UI), leading users to believe that they were leaked resources

### What to test

- 

### Who tested and where

- [ ] I validated this change 

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
@@ -728,7 +728,7 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
           logger.error(s"Wsm deleteVm job failed due to ${resp.getErrorReport.getMessage}") >> F.raiseError[Unit](
             AzureRuntimeDeletionError(
               params.runtime.id,
-              None,
+              params.diskId,
               params.workspaceId,
               s"WSM delete VM job failed due to ${resp.getErrorReport.getMessage}"
             )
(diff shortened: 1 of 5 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala
```diff
@@ -726,6 +728,7 @@ class AzurePubsubHandlerInterp[F[_]: Parallel](
           logger.error(s"Wsm deleteVm job failed due to ${resp.getErrorReport.getMessage}") >> F.raiseError[Unit](
             AzureRuntimeDeletionError(
               params.runtime.id,
+              None,
```
LizBaldo: For VM deletion polling, the diskID is not part of the parameters (same for storage container deletion below). If this is just for logging then it is okay, but if raising this error triggers database status changes then I will need to change that.

lucymcnatt: hmm yeah I think this is then picked up by the handleAzureRuntimeDeletionError func in the LeoPubsubMessageSubscriber, so we might need these to not be `None`...

LizBaldo: done

mlilynolting (approved):

mlilynolting:
Looks like test "should handle azure vm creation failure properly" is failing?

lucymcnatt (approved):
LGTM given that one unit test is fixed!

PR #4676, opened 2024-06-28
check delete actions when deleting runtimes

dvoet:

### What

-

### Why

-

### What to test

- 

### Who tested and where

- [ ] I validated this change 

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala
@@ -288,24 +288,16 @@ class RuntimeV2ServiceInterp[F[_]: Parallel](
     for {
       ctx <- as.ask
 
-      hasWorkspacePermission <- authProvider.isUserWorkspaceReader(
-        WorkspaceResourceSamResourceId(workspaceId),
-        userInfo
-      )
-      _ <- F.raiseUnless(hasWorkspacePermission)(ForbiddenError(userInfo.userEmail))
-
       runtime <- RuntimeServiceDbQueries.getActiveRuntimeRecord(workspaceId, runtimeName).transaction
-
-      hasPermission <-
-        if (runtime.auditInfo.creator == userInfo.userEmail) F.pure(true)
-        else
-          authProvider
-            .isUserWorkspaceOwner(WorkspaceResourceSamResourceId(workspaceId), userInfo)
-
-      _ <- ctx.span.traverse(s => F.delay(s.addAnnotation("Done auth call for delete azure runtime permission")))
-      _ <- F
-        .raiseError[Unit](RuntimeNotFoundException(runtime.cloudContext, runtimeName, "permission denied"))
-        .whenA(!hasPermission)
+      wsmResourceId = WsmControlledResourceId(UUID.fromString(runtime.internalId))
+      hasVmDeletePermission <- checkSamPermission(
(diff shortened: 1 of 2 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala
```diff
@@ -288,24 +288,16 @@ class RuntimeV2ServiceInterp[F[_]: Parallel](
...
-      hasWorkspacePermission <- authProvider.isUserWorkspaceReader(
-        WorkspaceResourceSamResourceId(workspaceId),
-        userInfo
-      )
-      _ <- F.raiseUnless(hasWorkspacePermission)(ForbiddenError(userInfo.userEmail))
-
       runtime <- RuntimeServiceDbQueries.getActiveRuntimeRecord(workspaceId, runtimeName).transaction
-
-      hasPermission <-
-        if (runtime.auditInfo.creator == userInfo.userEmail) F.pure(true)
```
mlilynolting, line 300: Without this "creator backdoor" we may need a solution for the Terra case of a runtime without a Sam representation, retained in Leonardo to represent an error state.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala
```diff
@@ -288,24 +288,16 @@ class RuntimeV2ServiceInterp[F[_]: Parallel](
...
-        else
-          authProvider
-            .isUserWorkspaceOwner(WorkspaceResourceSamResourceId(workspaceId), userInfo)
-
-      _ <- ctx.span.traverse(s => F.delay(s.addAnnotation("Done auth call for delete azure runtime permission")))
-      _ <- F
-        .raiseError[Unit](RuntimeNotFoundException(runtime.cloudContext, runtimeName, "permission denied"))
-        .whenA(!hasPermission)
+      wsmResourceId = WsmControlledResourceId(UUID.fromString(runtime.internalId))
+      hasVmDeletePermission <- checkSamPermission(
```
mlilynolting, line 293: This is great! I would want to verify that 
- a Terra user with the workspace Owner role has the Delete action on workspace resources
- Given a Terra user with a non-Owner workspace role (what Leonardo considers a "workspace reader"),
    - the user has the Delete action on a runtime they created
    - the user does not have the Delete action on a runtime they didn't create
    - the user can delete a runtime in Error state which has been purged from Sam
- a Terra user with no workspace roles does not have the Delete action on a runtime they created
    - when the user is added to the workspace again, they regain the Delete action on runtimes they created

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/RuntimeV2ServiceInterp.scala
```diff
@@ -337,9 +328,18 @@ class RuntimeV2ServiceInterp[F[_]: Parallel](
           diskIdToDelete <-
             if (disk.wsmResourceId.isDefined && disk.status.isDeletable) {
               for {
+                hasDiskDeletePermission <- checkSamPermission(
```
mlilynolting, line 331: Here I've got the same questions as above but for Disks

mlilynolting:
This is exactly what leonardo access control desperately needs. I'm sensitive to the need for regression testing given that the current behavior is pretty wild-wild-west.

PR #4683, opened 2024-07-01
RW-12956 Fixing a bug where PD is not cleaned up properly

Qi77Qi:
Last week, we had some bug in SAS chart that was preventing app creation to succeed. This revealed a different bug that if an app creation fails, the newly created PD may not be cleaned up, and this causes all future app creation in the same workspace to fail because the following error

```
[ERROR] [06/19/2024 20:37:56.262] [leonardo-akka.actor.default-dispatcher-9] [akka.actor.ActorSystemImpl(leonardo)] HttpMethod(POST) http://notebooks.firecloud.org/api/google/v1/apps/terra-vpc-sc-9614c68a/all-of-us-4275-cromwell-6348: 500 Internal Server Error entity: {"source":"leonardo","message":"Existing disk found, but no restore info found in DB","statusCode":500,"exceptionClass":"class org.broadinstitute.dsde.workbench.leonardo.model.LeoException","traceId":"df17b4103e731dfbac5abc2ab2595fb8/16266843891970221338"}
```

What this PR will does is pass the diskId to the `handleKubernetesError` function, where it'll transition the disk status to `Failed`. This way, UI won't attempt to reuse this disk and run into the above exception. We might want to consider adding a step to delete PD in GCP in the error handler for future improvements.

__Jira ticket__: https://precisionmedicineinitiative.atlassian.net/browse/RW-12956

### What

-

### Why

-

### What to test

- 

### Who tested and where

- [ ] I validated this change 

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/LeoAppServiceInterp.scala
@@ -1231,7 +1231,9 @@ final class LeoAppServiceInterp[F[_]: Parallel](config: AppServiceConfig,
             case (Some(FormattedBy.Galaxy), None) | (Some(FormattedBy.Cromwell), None) |
                 (Some(FormattedBy.Allowed), None) =>
               F.raiseError[Option[LastUsedApp]](
-                new LeoException("Existing disk found, but no restore info found in DB", traceId = Some(ctx.traceId))
+                new LeoException(s"Existing ${diskResult.disk.id} found, but no restore info found in DB",
+                                 traceId = Some(ctx.traceId)
+                )
               )
             case (Some(FormattedBy.Custom), _) =>
               F.raiseError[Option[LastUsedApp]](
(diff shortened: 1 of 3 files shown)


lucymcnatt (approved):

cindy-broadinstitute (commented):
Will approve after unit tests pass

Qi77Qi:
@lucymcnatt @cindy-broadinstitute I updated the PR a bit..the previous location where I added the diskId isn't correct given current implementation of the `handleKubernetesError` and `cleanUpAfterCreateAppError`. I added a comment to clarify.

I added the diskId when disk fails to detach instead because this is what I see in BEE after I launch an RStudio with bad helm chart
```
{"@timestamp":"2024-07-02T15:48:52.819519274Z","@version":"1","logger_name":"org.broadinstitute.dsde.workbench.leonardo.http.Boot","traceId":"90134d7c-631e-4dcc-ac8b-9d6044b5b2a6","version":"aee05698f","thread_name":"io-compute-blocker-100","stack_trace":"org.broadinstitute.dsde.workbench.leonardo.monitor.PubsubHandleMessageError$PubsubKubernetesError: An error occurred with a kubernetes operation from source disk during action deleteApp. \nOriginal message: The data disk failed to detach within the time limit, cannot proceed with delete disk\n","severity":"ERROR","serviceContext":{"service":"leonardo","version":"aee05698f"},"message":"An error occurred during resource clean up for app AppName(qi-rstudio-7-2-0) in project terra-quality-d7a5f18a"}
```

Basically, the disk wasn't cleaned up becuz it failed to detach. This PR will make is such that when disk fail to detach, the disk status will be marked as `FAILED` in `handleKubernetesError` instead of remaining in `Ready` status. There might be some issue in how we determine disk is detached today worth looking into as well.

cindy-broadinstitute (approved):

PR #4715, opened 2024-07-17
check runtime delete action instead of workspace owner role

dvoet:
__Jira ticket__: https://broadworkbench.atlassian.net/browse/IA-5025

## Summary of changes

allow users with workspace delete action instead of workspace owner role to delete runtimes

### Why

requiring owner role requires granting too much access to the rawls SA which needs to perform workspace delete

### What to test

- 

### Who tested and where

- [ ] I validated this change 

### DIFF
--- a/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/samModels.scala
+++ b/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/samModels.scala
@@ -239,6 +239,9 @@ object WsmResourceAction {
   final case object Read extends WsmResourceAction {
     val asString = "read"
   }
+  final case object Delete extends WsmResourceAction {
+    val asString = "delete"
+  }
   val allActions = sealerate.values[WsmResourceAction]
   val stringToAction: Map[String, WsmResourceAction] =
     sealerate.collect[WsmResourceAction].map(a => (a.asString, a)).toMap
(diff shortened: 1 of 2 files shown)


mlilynolting:
@dvoet this looks good! Have you confirmed that a user with Owner role on a workspace and no special access to runtimes under it can successfully delete any runtime? I'm not sure how access to the Delete action propagates.

dvoet:
> @dvoet this looks good! Have you confirmed that a user with Owner role on a workspace and no special access to runtimes under it can successfully delete any runtime? I'm not sure how access to the Delete action propagates.

Bummer, owners do not. I drilled into what the actual resource type is (the naming in Leo hides it) and it is `controlled-application-private-workspace-resource`. Only Leo has delete action on those. I am going to switch to checking delete on the workspace itself.

mlilynolting:
Conceptually I don't like checking the delete-workspace action as a proxy for delete-workspace-resource. I don't think it's much of an improvement on just checking for Owner role.

dvoet:
in my opinion, switching to an action check, even it if is not quite right, is vastly superior to a role check. Roles were never meant to gate access to features.

jdcanas (approved):

mlilynolting:
> in my opinion, switching to an action check, even it if is not quite right, is vastly superior to a role check. Roles were never meant to gate access to features.

That makes sense to me. Could you add a comment to the code indicating that a check on the runtime Delete action would be preferred but that it would require implementing a Sam hierarchy from Workspace to Runtimes?

mlilynolting (approved):

dvoet:
> Could you add a comment to the code indicating that a check on the runtime Delete action would be preferred but that it would require implementing a Sam hierarchy from Workspace to Runtimes?

@mlilynolting I will cover this in my doc, but hierarchy is only part of the answer. Right now, there is no sam resource that exists that Leo can check to see if the caller should be able to delete the requested runtime. The right resource needs to exist _and_ hierarchy can be used to make the correct permissions flow.

mlilynolting:
That's a good point.

PR #4716, opened 2024-07-18
Ia 4997 third party cookies

mlilynolting:
__Jira ticket__: https://broadworkbench.atlassian.net/browse/IA-4997

## Summary of changes

Can't test this locally or in a BEE, so here is a change which should be disabled everywhere by default.

NOTE: There are a bunch of nginx rewrites in the app helm charts here to add the Partitioned flag; for now I've commented them out and left the old logic in place. They ought not to be needed for me to test Galaxy under CHIPS.

Let's see what the Domain header does in Set-Cookie!

When the config flag IS_PROXY_COOKIE_PARTITIONED is set, Leo `setCookie` will respond with *three* Set-Cookie headers: one unscoped (will be blocked as a top-level third party cookie), one partitioned without a domain (will be scoped to the embedding domain, thus terra-ui), and one partitioned with a domain set to the proxy URL (might?? be scoped to the proxy domain and get past third-party protections on the new Galaxy tab!)

## Testing these changes

We'll need to check Galaxy, Jupyter, Rstudio, and AoU tools with the flag set. This should be possible by pointing a local terra-ui at dev, with the proxy cookie flag set in leo's `overrides.env` file.

This is not intended as a final fix, just an attempt to understand whether the Domain attribute could possibly allow us to use CHIPS for now to get past our third-party cookie woes.

### What to test

- Can you open and use each tool?

### Who tested and where

- [ ] I validated this change 

### DIFF
--- a/http/src/main/resources/init-resources/cluster-site-gce.conf
+++ b/http/src/main/resources/init-resources/cluster-site-gce.conf
@@ -34,7 +34,7 @@
     # Append SameSite=None to cookies set by RStudio. This is required by some browsers because we
     # render RStudio in an iframe. There does not appear to be a way within RStudio to do this, hence
     # doing it in the proxy.
-    Header edit Set-Cookie ^(.*)$ $1;Secure;SameSite=None;HttpOnly "expr=%{REQUEST_URI} =~ m#/proxy/[^/]*/[^/]*/rstudio/.*#"
+    Header edit Set-Cookie ^(.*)$ $1;Secure;SameSite=None;HttpOnly;Partitioned "expr=%{REQUEST_URI} =~ m#/proxy/[^/]*/[^/]*/rstudio/.*#"
 
     ####################
     # Welder
--- a/http/src/main/resources/init-resources/cluster-site-gce.conf
+++ b/http/src/main/resources/init-resources/cluster-site-gce.conf
@@ -34,7 +34,8 @@
     # Append SameSite=None to cookies set by RStudio. This is required by some browsers because we
     # render RStudio in an iframe. There does not appear to be a way within RStudio to do this, hence
     # doing it in the proxy.
-    Header edit Set-Cookie ^(.*)$ $1;Secure;SameSite=None;HttpOnly;Partitioned "expr=%{REQUEST_URI} =~ m#/proxy/[^/]*/[^/]*/rstudio/.*#"
+    # Header edit Set-Cookie ^(.*)$ $1;Secure;SameSite=None;HttpOnly;Partitioned "expr=%{REQUEST_URI} =~ m#/proxy/[^/]*/[^/]*/rstudio/.*#"
+    Header edit Set-Cookie ^(.*)$ $1;Secure;SameSite=None;HttpOnly "expr=%{REQUEST_URI} =~ m#/proxy/[^/]*/[^/]*/rstudio/.*#"
 
     ####################
     # Welder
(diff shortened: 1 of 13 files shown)


http/src/main/resources/init-resources/cluster-site-gce.conf
```diff
@@ -34,6 +34,7 @@
     # Append SameSite=None to cookies set by RStudio. This is required by some browsers because we
     # render RStudio in an iframe. There does not appear to be a way within RStudio to do this, hence
     # doing it in the proxy.
+    # Header edit Set-Cookie ^(.*)$ $1;Secure;SameSite=None;HttpOnly;Partitioned "expr=%{REQUEST_URI} =~ m#/proxy/[^/]*/[^/]*/rstudio/.*#"
```
jdcanas, line 38: is this needed? Its commented out

mlilynolting: I left the places we'd need to change to get full CHIPS support as comments so that it's easy to implement if this does work.

jdcanas: maybe just add a comment to indicate what this is in the relevant places

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/api/CookieSupport.scala
```diff
@@ -47,7 +44,34 @@ object CookieSupport {
   private def buildRawUnsetCookie(): RawHeader =
     RawHeader(
       name = "Set-Cookie",
-      value = s"$tokenCookieName=unset; expires=Thu, 01 Jan 1970 00:00:00 GMT; Path=/; Secure; SameSite=None; HttpOnly"
+      value = s"$tokenCookieName=unset; Expires=Thu, 01 Jan 1970 00:00:00 GMT; Path=/; Secure; SameSite=None; HttpOnly"
```
jdcanas, line 47: just confirming the `expires` case change here is intentional?

mlilynolting: Yeah, intentional!

http/src/main/resources/init-resources/cluster-site-gce.conf
mlilynolting: This isn't meant to stick around, I just can't test it anywhere but dev! If it works, I remove the comment and ship this line; if it doesn't, I revert the commit.

jdcanas (approved):

http/src/main/resources/leo.conf
```diff
@@ -65,6 +65,7 @@ proxy {
   # Should match the jupyter wildcard cert
   proxyDomain = ${?PROXY_DOMAIN}
   proxyUrlBase = ${?PROXY_URL_BASE}
+  isProxyCookiePartitioned = ${?IS_PROXY_COOKIE_PARTITIONED}
```
LizBaldo, line 68: Could you link the PR in Terra HelmFile that sets this env variable? I could not find it

LizBaldo: I think leo.conf overwrites reference.conf

LizBaldo (approved):
LGTM, you might just want to also set the flag to False in the leo helm files for the live and bee environments so that you don't just depend on reference.conf setting it to False

mlilynolting:
@LizBaldo doesn't an unset flag evaluate to False? I thought that was the ? part of `${?FLAG}`

LizBaldo:
> @LizBaldo doesn't an unset flag evaluate to False? I thought that was the ? part of `${?FLAG}`

I am not sure actually, that's why I am thinking it might be safer to set it, but if you know otherwise feel free to go ahead

mlilynolting:
No, the abundance of caution makes a ton of sense to me, thanks. I hadn't made a helmfile PR yet; just did: https://github.com/broadinstitute/terra-helmfile/pull/5741 (I had thought I'd be able to test it with the defaults but I was wrong)

PR #4725, opened 2024-07-24
[IA-5028] Increase default docker shared memory size

LizBaldo:
__Jira ticket__: https://broadworkbench.atlassian.net/browse/IA-5028

### What

- Increases the amount of shared memory allocated to the jupyter docker container. Bumping the shm size to 256m worked great, but we are attempting to dynamically adjust this value to about 50% of the total amount of memory allocated to the container

### Why

- The current default is 64mb, which is not sufficient for some users. Doing a dynamic allocation in the backend directly will avoid a lot of unnecessary UI and UX work, as exposing this might lead to a bad user experience (aka jupyter crashing)
### What to test

Tested on my BEE for both a GCE (jupyter and RStudio) runtime with and without GPU, as well as for a dataproc cluster, and can see the amount of memory matching what is requested in the docker composes:
Screenshot 2024-07-29 at 5 01 58 PM

### Who tested and where

- [ ] I validated this change 


http/src/main/resources/init-resources/startup.sh
```diff
@@ -127,6 +132,26 @@ then
     if [ ! -z "$JUPYTER_DOCKER_IMAGE" ] ; then
         echo "Restarting Jupyter Container $GOOGLE_PROJECT / $CLUSTER_NAME..."
 
+        # Make sure when runtimes restarts, they'll get a new version of jupyter docker compose file
```
LizBaldo, line 131: I don't know why, but by default the docker compose file was not being pulled when restarting the runtime. That can be problematic when users are trying to update their machine with different memory and CPU allocations.

http/src/main/resources/init-resources/gce-init.sh
```diff
@@ -103,6 +103,11 @@ export PROXY_DOCKER_IMAGE=$(proxyDockerImage)
 export CRYPTO_DETECTOR_SERVER_NAME=$(cryptoDetectorServerName)
 export CRYPTO_DETECTOR_DOCKER_IMAGE=$(cryptoDetectorDockerImage)
 export MEM_LIMIT=$(memLimit)
+# Setting the shared docker memory to 50% of the allocated memory limit, converting from byte to mb
```
LizBaldo: I know I could do it within the scala code in leo, but the list of runtime template values is already sooo long that it felt cleaner this way honestly, but happy to rework if you think otherwise, I just want to get this in front of users ASAP to help unblock them

lucymcnatt: totally fine with this approach 👍

lucymcnatt (approved):
Looks great!

http/src/main/resources/init-resources/gce-init.sh
jdcanas: hmm... I would honestly prefer to do it in leo. I find whenever I use `awk` it feels elegant to start with, but a year or so down the line its broken by bash's lack of type safety in conjunction with an assumption being changed somewhere else in the application (i.e. `memLimit` changing data types in this case).

jdcanas: also, our init scripts aren't tested very well 😬

jdcanas (approved):
would like to avoid logic in the init script like this since it isn't tested, but not going to hold the PR on it

http/src/main/resources/init-resources/gce-init.sh
LizBaldo: This is a very good point, I'll take a stab at moving the logic out of the init scripts and into Leonardo proper with unit testing tomorrow.

LizBaldo:
@jdcanas I made the changes and moved the logic so it an be unit testable. However I can't merge it because the integration tests can't run due to a path being deleted `ls: cannot access '/home/runner/work/terra-github-workflows/terra-github-workflows/leonardo/test-reports': No such file or directory`

I am not going to be around next week, neither will Lucy, so would you be able to:
- Rerun the tests on Monday and see if the underlying infra issue is still there?
- Ask Devops about it if it is still a problem
- Merge the PR and keep an eye on things if it needs to be reverted?


Many thanks 🙏

http/src/main/resources/init-resources/gce-init.sh
LizBaldo: @jdcanas I took a stab at it if you want to re-review and merge while I am away, thanks!

PR #4732, opened 2024-08-01
[IA-4681] Update terra-common-lib

lucymcnatt:
__Jira ticket__: https://broadworkbench.atlassian.net/browse/IA-4681

### What

- Updating terra-common-lib and fixing affected functions
- Also upgraded wblibs to pull in most recent version of openTelemetry to fix dep conflicts

### Why

-  In order to pull in a non-vulnerable version of Okio (3.4.0)

### What to test

- 

### Who tested and where

- [ ] I validated this change 


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/DbReference.scala
```diff
@@ -32,7 +33,9 @@ object DbReference extends LazyLogging {
 
       changelogParameters.foreach { case (key, value) => liquibase.setChangeLogParameter(key, value) }
 
-      liquibase.update(new Contexts(), new LabelExpression())
```
lucymcnatt, line 35: liquibase.update was deprecated, need to use a CommandScope instead (see below)

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoMetricsMonitor.scala
```diff
@@ -365,19 +365,7 @@ class LeoMetricsMonitor[F[_]](config: LeoMetricsMonitorConfig,
                 val labelSelector = s"leoAppName=${app.appName.value}"
                 for {
                   pods <- F.blocking(
-                    client.listNamespacedPod(namespace.value,
```
lucymcnatt: this method was changed so that the arguments are now setter functions

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/db/DbReference.scala
LizBaldo: Ohhh thanks for doing that

LizBaldo (approved):

lucymcnatt:
@LizBaldo I changed how the mocking is done in that one spec, so 🤞 it should be passing now

LizBaldo:
Thanks a bunch! I'll monitor the rest of the tests and will merge early next week

LizBaldo (approved):

mlilynolting (approved):

LizBaldo:
Note that updating to the new kubernetes java client was causing issues with GKE app creation and so given that the scope of this PR is to upgrade to a safe version of okio ( >= 3.4.0), pinning TCL to 1.1.4 lets us do that (okio is 3.6.0 now), while still using the legacy kubernetes java API.

We should upgrade to the latest kubernetes API eventually but this is outside of the scope of this PR

PR #4737, opened 2024-08-06
Enable v1 APIs for Disk, App, and Runtime for Azure hosted Leonardo instances

bennettn4:
__Jira ticket__: https://broadworkbench.atlassian.net/browse/TOAZ-366

## Summary of changes

Enable v1 APIs for Disk, App, and Runtime for Azure hosted Leonardo instances.

### What

-

### Why

- v1 APIs are still used in TerraUI for data retrieval even when hosted on Azure. This will be changed with ongoing development of platform agnostic v2 APIs 

- [ ] I validated this change 


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/DiskServiceInterp.scala
```diff
@@ -122,10 +122,10 @@ class DiskServiceInterp[F[_]: Parallel](config: PersistentDiskConfig,
   ): F[Unit] =
     for {
       ctx <- as.ask
-      sourceAncestry <- F.fromFuture(F.delay(googleProjectDAO.getAncestry(sourceGoogleProject.value)))
+      sourceAncestry <- F.fromFuture(F.delay(googleProjectDAO.get.getAncestry(sourceGoogleProject.value)))
```
rtitle, line 125: My concern with sprinkling `.get` in this code is that the API will fail with a HTTP 500 `scala.MatchError` if these APIs are called in a Leo deployment with no GCP dependencies.

A cleaner option might be doing something like this in the appropriate endpoints (e.g. create/delete runtime/disk):
```
projectDAO <- F.fromOption(googleProjectDAO, BadRequestException("GCP is not enabled"))
// then use projectDAO as needed
```

Then you would get a sensible error message and HTTP status code if a GCP-specific API is called.

(I realize that this only affects the VA deployment and v1 create/delete runtime/disk will never actually be called. So just a minor suggestion.)

rtitle (approved):
Approving, but with a minor suggestion above.

giventocode (approved):

PR #4741, opened 2024-08-09
PROD-981 Locking error in a non-shared notebook

rtitle:
https://broadworkbench.atlassian.net/browse/PROD-981

## Summary of changes

Users are getting "lock expired" errors while working in notebooks. Example:

image

I think this is because the `edit-mode.js` extension is not refreshing the object locks from welder, because it is not getting the right configs. The configs come from `/etc/jupyter/nbconfig/notebook.json`. This file is created in the `gce-init.sh` script, but not recreated after container startup in `startup.sh`. There was some logic to recreate `notebook.json` in [this PR](https://github.com/DataBiosphere/leonardo/pull/2559), but it was buried under a `NEEDS_MIGRATE` flag. This PR moves it out of the `NEEDS_MIGRATE` block.

### What to test

Deploy this PR to a BEE and run through the below repro steps:

Repro steps:

1. Create a workspace, runtime, and a notebook
2. Pause/resume the runtime manually
3. Open the notebook
4. Browser console should say `edit mode plugin initialized`
5. Browser console should NOT say `welder is disabled`
6. Network tab should show periodic calls to the welder `lock` API
7. Above error should not happen after several minutes

### Who tested and where

- [ ] I validated this change 

### DIFF
--- a/http/src/main/resources/init-resources/startup.sh
+++ b/http/src/main/resources/init-resources/startup.sh
@@ -165,13 +165,13 @@ END
 
         if [ "$NEED_MIGRATE" == "true" ] ; then
           docker exec $JUPYTER_SERVER_NAME /bin/bash -c "[ ! -d $JUPYTER_USER_HOME/notebooks/.jupyter ] && rsync -av --progress --exclude notebooks . $JUPYTER_USER_HOME/notebooks || true"
-
-          log 'Copy Jupyter frontend notebook config...'
-          $GSUTIL_CMD cp ${JUPYTER_NOTEBOOK_FRONTEND_CONFIG_URI} /var
-          JUPYTER_NOTEBOOK_FRONTEND_CONFIG=`basename ${JUPYTER_NOTEBOOK_FRONTEND_CONFIG_URI}`
-          retry 3 docker exec -u root ${JUPYTER_SERVER_NAME} /bin/bash -c "mkdir -p $JUPYTER_HOME/nbconfig"
-          docker cp /var/${JUPYTER_NOTEBOOK_FRONTEND_CONFIG} ${JUPYTER_SERVER_NAME}:${JUPYTER_HOME}/nbconfig/
         fi
+
+        log 'Copy Jupyter frontend notebook config...'
+        $GSUTIL_CMD cp ${JUPYTER_NOTEBOOK_FRONTEND_CONFIG_URI} /var
+        JUPYTER_NOTEBOOK_FRONTEND_CONFIG=`basename ${JUPYTER_NOTEBOOK_FRONTEND_CONFIG_URI}`
+        retry 3 docker exec -u root ${JUPYTER_SERVER_NAME} /bin/bash -c "mkdir -p $JUPYTER_HOME/nbconfig"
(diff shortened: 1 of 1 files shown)


http/src/main/resources/init-resources/startup.sh
```diff
@@ -165,13 +165,13 @@ END
...
           docker exec $JUPYTER_SERVER_NAME /bin/bash -c "[ ! -d $JUPYTER_USER_HOME/notebooks/.jupyter ] && rsync -av --progress --exclude notebooks . $JUPYTER_USER_HOME/notebooks || true"
-
-          log 'Copy Jupyter frontend notebook config...'
-          $GSUTIL_CMD cp ${JUPYTER_NOTEBOOK_FRONTEND_CONFIG_URI} /var
-          JUPYTER_NOTEBOOK_FRONTEND_CONFIG=`basename ${JUPYTER_NOTEBOOK_FRONTEND_CONFIG_URI}`
-          retry 3 docker exec -u root ${JUPYTER_SERVER_NAME} /bin/bash -c "mkdir -p $JUPYTER_HOME/nbconfig"
-          docker cp /var/${JUPYTER_NOTEBOOK_FRONTEND_CONFIG} ${JUPYTER_SERVER_NAME}:${JUPYTER_HOME}/nbconfig/
         fi
+
+        log 'Copy Jupyter frontend notebook config...'
```
rtitle, line 170: This is the same code but "lifted" out of the `NEED_MIGRATE` block. I think it needs to run regardless to recreate the `/etc/jupyter/nbconfig/notebook.json` file.

jdcanas (approved):

http/src/main/resources/init-resources/startup.sh
LizBaldo: My previous PR moved the docker compose up commands out of the block, which seemed to have caused this problem. 
Based on the `NEED_MIGRATE` comment above, it would make sense to move it out of the block indeed

http/src/main/resources/init-resources/startup.sh
```diff
@@ -228,13 +228,14 @@ else
...
           ${DOCKER_COMPOSE} -f ${JUPYTER_DOCKER_COMPOSE} up -d
-
-          log 'Copy Jupyter frontend notebook config...'
-          $GSUTIL_CMD cp ${JUPYTER_NOTEBOOK_FRONTEND_CONFIG_URI} /var
-          JUPYTER_NOTEBOOK_FRONTEND_CONFIG=`basename ${JUPYTER_NOTEBOOK_FRONTEND_CONFIG_URI}`
-          retry 3 docker exec -u root ${JUPYTER_SERVER_NAME} /bin/bash -c "mkdir -p $JUPYTER_HOME/nbconfig"
-          docker cp /var/${JUPYTER_NOTEBOOK_FRONTEND_CONFIG} ${JUPYTER_SERVER_NAME}:${JUPYTER_HOME}/nbconfig/
         fi
+
+        log 'Copy Jupyter frontend notebook config...'
```
LizBaldo, line 233: Why keep this here? It looks like it is exactly like the commands that were removed from the block above

LizBaldo (approved):
Tested on my BEE and it works, thank you!

I think we likely want a follow up PR (on IA, not you Rob 😅) to:
- Clean up the script and better document each step
- Add a step in the VM stop/start test to check that we can save the notebook

PR #4778, opened 2024-09-04
[IA-5048] init script documentation

LizBaldo:
__Jira ticket__: https://broadworkbench.atlassian.net/browse/IA-5048

### What

- Remove unused legacy code and document the many complexities of our init scripts

### Why

- We had a couple of PROD incident related to some legacy code in the init script, and this is an attempt to better document and understand what these scripts are doing

### What to test

- 

### Who tested and where

- [ ] I validated this change 


http/src/main/resources/init-resources/gce-init.sh
```diff
@@ -79,12 +141,16 @@ log "Running GCE VM init script..."
...
+
+## Used for profiling
 START_TIME=$(date +%s)
 STEP_TIMINGS=($(date +%s))
 
 # Set variables
 # Values like $(..) are populated by Leo when a cluster is created.
+# See https://github.com/DataBiosphere/leonardo/blob/e46acfcb409b11198b1f12533cefea3f6c7fdafb/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/RuntimeTemplateValues.scala#L192
 # Avoid exporting variables unless they are needed by external scripts or docker-compose files.
+export CLOUD_SERVICE='GCE'
```
LizBaldo: This is not ideal and won't backpopulate the existing runtimes, but this makes the script more legible

http/src/main/resources/init-resources/gce-init.sh
```diff
@@ -140,77 +207,28 @@ INIT_BUCKET_NAME=$(initBucketName)
...
 
+# Welder configuration, Rstudio files are saved every X seconds in the background but Jupyter notebooks are not
 if [ ! -z "$RSTUDIO_DOCKER_IMAGE" ] ; then
   export SHOULD_BACKGROUND_SYNC="true"
 else
   export SHOULD_BACKGROUND_SYNC="false"
 fi
 
+
+# Use specific docker compose command if the container is coming from GCR -> WHY?
```
LizBaldo: Seems like a GCR specific authentication step that might not translate perfectly to GAR FYI @lucymcnatt

http/src/main/resources/init-resources/gce-init.sh
```diff
@@ -140,77 +207,28 @@ INIT_BUCKET_NAME=$(initBucketName)
...
+# TODO - Also check for GAR see https://broadworkbench.atlassian.net/browse/IA-4518
 if grep -qF "gcr.io" <<< "${JUPYTER_DOCKER_IMAGE}${RSTUDIO_DOCKER_IMAGE}${PROXY_DOCKER_IMAGE}${WELDER_DOCKER_IMAGE}" ; then
   log 'Authorizing GCR...'
   DOCKER_COMPOSE="docker run --rm -v /var/run/docker.sock:/var/run/docker.sock -v /var:/var -w=/var cryptopants/docker-compose-gcr"
 else
   DOCKER_COMPOSE="docker run --rm -v /var/run/docker.sock:/var/run/docker.sock -v /var:/var docker/compose:1.29.2"
 fi
 
-function apply_user_script() {
```
LizBaldo, line 159: Just moved to the top of the file with the other functions

http/src/main/resources/init-resources/gce-init.sh
```diff
@@ -272,9 +291,10 @@ $GSUTIL_CMD cp ${SERVER_KEY} ${CERT_DIRECTORY}
 $GSUTIL_CMD cp ${ROOT_CA} ${CERT_DIRECTORY}
 $GSUTIL_CMD cp gs://${INIT_BUCKET_NAME}/* ${DOCKER_COMPOSE_FILES_DIRECTORY}
 
+# TODO: Figure out what this file is for. It is empty but used by the docker compose for both GCE and Dataproc
```
LizBaldo: @Qi77Qi or @rtitle if you know of the top of your head what this credential file is used for lmk if not I will dig a bit deeper and hopefully can delete eventually

http/src/main/resources/init-resources/gce-init.sh
```diff
@@ -395,7 +421,7 @@ if [ ! -z "$JUPYTER_DOCKER_IMAGE" ] ; then
     docker cp /var/${JUPYTER_NOTEBOOK_FRONTEND_CONFIG} ${JUPYTER_SERVER_NAME}:${JUPYTER_HOME}/nbconfig/
   fi
 
-  # Install NbExtensions
+  # Install NbExtensions (the ones that users might be providing likely AOU?)
```
LizBaldo: @Qi77Qi do you have more insights on these three types of jupyter extensions and what they are used for?

http/src/main/resources/init-resources/gce-init.sh
```diff
@@ -499,16 +525,15 @@ if [ ! -z "$JUPYTER_DOCKER_IMAGE" ] ; then
   # For older jupyter images, jupyter_delocalize.py is using 127.0.0.1 as welder's url, which won't work now that we're no longer using `network_mode: host` for GCE VMs
   docker exec $JUPYTER_SERVER_NAME /bin/bash -c "sed -i 's/127.0.0.1/welder/g' /etc/jupyter/custom/jupyter_delocalize.py"
 
-  # Copy gitignore into jupyter container
-
+  # Copy gitignore into jupyter container (ask AOU?)
```
LizBaldo, line 545: @Qi77Qi do you know why we have to copy the gitignore over? I just want to document the use case here

http/src/main/resources/init-resources/gce-init.sh
```diff
@@ -499,16 +525,15 @@ if [ ! -z "$JUPYTER_DOCKER_IMAGE" ] ; then
...
   docker exec $JUPYTER_SERVER_NAME /bin/bash -c "wget -N https://raw.githubusercontent.com/DataBiosphere/terra-docker/045a139dbac19fbf2b8c4080b8bc7fff7fc8b177/terra-jupyter-aou/gitignore_global"
 
-  # Install nbstripout and set gitignore in Git Config
-
+  # Install nbstripout and set gitignore in Git Config (ask AOU?)
   docker exec $JUPYTER_SERVER_NAME /bin/bash -c "pip install nbstripout \
         && nbstripout --install --global \
         && git config --global core.excludesfile $JUPYTER_USER_HOME/gitignore_global"
 
+  # Starts the locking logic (used for AOU). google_sign_in.js  is likely not used anymore
```
LizBaldo, line 553: @Qi77Qi if the google sign in plugin is not used anymore by AOU I might delete that part

http/src/main/resources/init-resources/gce-init.sh
```diff
@@ -583,13 +608,9 @@ RSTUDIO_USER_HOME=$RSTUDIO_USER_HOME" >> /usr/local/lib/R/etc/Renviron.site'
 fi
 
 # Resize persistent disk if needed.
-# This condition assumes Dataproc's cert directory is different from GCE's cert directory, a better condition would be
-# a dedicated flag that distinguishes gce and dataproc. But this will do for now
-# If it's GCE, we resize the PD. Dataproc doesn't have PD
-if [ -f "/var/certs/jupyter-server.crt" ]; then
```
LizBaldo, line 589: @rtitle do you know why we were checking if it is GCE? This is already in the context of the gce-init script so doing that check seemed redundant

http/src/main/resources/init-resources/init-actions.sh
```diff
@@ -344,38 +328,10 @@ END
 
     STEP_TIMINGS+=($(date +%s))
 
-    # If we have a service account JSON file, create an .env file to set GOOGLE_APPLICATION_CREDENTIALS
```
LizBaldo, line 347: This felt obsolete as we are identifying as the user pet. I am hoping this means that `/var/google_application_credentials.env` is also obsolete

http/src/main/resources/init-resources/init-actions.sh
```diff
@@ -344,38 +328,10 @@ END
...
-      fi
-    fi
-
-    STEP_TIMINGS+=($(date +%s))
-
     # Jupyter-specific setup, only do if Jupyter is installed
     if [ ! -z ${JUPYTER_DOCKER_IMAGE} ] ; then
       log 'Installing Jupydocker kernelspecs...'
 
-      # Install hail addition if the image is old leonardo jupyter image or it's a hail specific image
```
LizBaldo, line 371: We are not using this image anymore

http/src/main/resources/init-resources/startup.sh
```diff
@@ -116,18 +177,46 @@ if [ "${GPU_ENABLED}" == "true" ] ; then
   mount -o remount,exec /var/lib/nvidia
 fi
 
-# https://broadworkbench.atlassian.net/browse/IA-3186
-# This condition assumes Dataproc's cert directory is different from GCE's cert directory, a better condition would be
-# a dedicated flag that distinguishes gce and dataproc. But this will do for now
-if [ -f "$FILE" ]
+if [[ "${CLOUD_SERVICE}" == 'GCE' ]]; then
```
LizBaldo: It feels like we should be able to restart welder the same way for both GCE and Dataproc so I moved this at the top to make the script a bit more DRY

http/src/main/resources/init-resources/startup.sh
```diff
@@ -116,18 +177,46 @@ if [ "${GPU_ENABLED}" == "true" ] ; then
...
+    # GCE
 then
     CERT_DIRECTORY='/var/certs'
-    DOCKER_COMPOSE_FILES_DIRECTORY='/var/docker-compose-files'
     GSUTIL_CMD='docker run --rm -v /var:/var us.gcr.io/cos-cloud/toolbox:v20230714 gsutil'
     GCLOUD_CMD='docker run --rm -v /var:/var us.gcr.io/cos-cloud/toolbox:v20230714 gcloud'
     DOCKER_COMPOSE='docker run --rm -v /var/run/docker.sock:/var/run/docker.sock -v /var:/var docker/compose:1.29.2'
-    WELDER_DOCKER_COMPOSE=$(ls ${DOCKER_COMPOSE_FILES_DIRECTORY}/welder*)
     JUPYTER_DOCKER_COMPOSE=$(ls ${DOCKER_COMPOSE_FILES_DIRECTORY}/jupyter-docker*)
+    RSTUDIO_DOCKER_COMPOSE=$(ls ${DOCKER_COMPOSE_FILES_DIRECTORY}/rstudio-docker*)
```
LizBaldo: I am also making sure that we pull the latest docker compose for Rstudio so we match the behavior with Jupyter

http/src/main/resources/init-resources/startup.sh
```diff
@@ -164,18 +253,6 @@ END
         docker restart $JUPYTER_SERVER_NAME
         docker restart $WELDER_SERVER_NAME
 
-        # This line is only for migration (1/26/2022). Say you have an existing runtime where jupyter container's PD is mapped at $HOME/notebooks,
```
LizBaldo, line 167: I think we should simplify and not support these old runtimes anymore

http/src/main/resources/init-resources/gce-init.sh
rtitle: This might be a clue as to why, from the image documentation: https://hub.docker.com/r/cryptopants/docker-compose-gcr

```
This container helps resolve an issue with Container Optimized OS. The OS does not come with docker-compose, and instead expects users to use containerized docker-compose. However the official containerized docker-compose does not include docker-credential-gcr, which is needed to pull Docker images from a private Google Container Registry.

This container uses the official docker-compose as a base, and adds docker-credential-gcr. It can be used just like the normal containerized docker-compose, but can now access gcr.io.
```

So sounds like it just extends `docker/compose` to support GCR credentials. I am not sure if it will work with GAR…

http/src/main/resources/init-resources/gce-init.sh
rtitle: This seems good - not sure we need to worry about backpopulation, as the variable is just used in the scope of this script.

http/src/main/resources/init-resources/gce-init.sh
rtitle: I haven't verified, but my strong suspicion is it isn't used anymore. Before we implemented pet SAs we used to populate an application-default credentials file on Leo runtimes. But with pets that's no longer needed.

The file is referenced in the [docker-compose](https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/resources/init-resources/jupyter-docker-compose-gce.yaml#L39) so I'm guessing we need to make an empty file in this script to make the docker-compose not break. But perhaps we can remove it from docker-compose file and things will work.

http/src/main/resources/init-resources/gce-init.sh
```diff
@@ -272,9 +291,10 @@ $GSUTIL_CMD cp ${SERVER_KEY} ${CERT_DIRECTORY}
 $GSUTIL_CMD cp ${ROOT_CA} ${CERT_DIRECTORY}
 $GSUTIL_CMD cp gs://${INIT_BUCKET_NAME}/* ${DOCKER_COMPOSE_FILES_DIRECTORY}
 
+# TODO: Figure out what this file is for. It is empty but used by the docker compose for both GCE and Dataproc
 echo "" > /var/google_application_credentials.env
 
-# Install env var config
+# Install env var config (e.g. AOU uses it to inject workspace name)
```
rtitle: Terra UI specifies several custom env vars too. E.g. from a dev runtime creation:
```
{
  "WORKSPACE_NAME": "CARJune24",
  "WORKSPACE_NAMESPACE": "callisto-dev",
  "WORKSPACE_BUCKET": "gs://fc-09516ff0-136e-4874-8484-1be0afa267a6",
  "GOOGLE_PROJECT": "terra-dev-e67d9572",
  "CUSTOM_IMAGE": "false",
  "DRS_RESOLVER_ENDPOINT": "api/v4/drs/resolve",
  "TERRA_DEPLOYMENT_ENV": "dev"
}
```

http/src/main/resources/init-resources/gce-init.sh
```diff
@@ -299,6 +319,7 @@ fi
 if [ "${GPU_ENABLED}" == "true" ] ; then
   COMPOSE_FILES+=(-f ${DOCKER_COMPOSE_FILES_DIRECTORY}/`basename ${GPU_DOCKER_COMPOSE}`)
   if [ ! -z "$RSTUDIO_DOCKER_IMAGE" ] ; then
+    # Little bit of hack to switch the jupyter paths to the rstudio ones. Should be different env variables for Rstudio instead
```
rtitle: I guess it's replacing strings in [gpu-docker-compose.yaml](https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/resources/init-resources/gpu-docker-compose.yaml). Yeah really feels like we should have separate Jupyter and RStudio gpu-docker-compose files.

http/src/main/resources/init-resources/gce-init.sh
```diff
@@ -349,13 +370,16 @@ END
 # Create a network that allows containers to talk to each other via exposed ports
 docker network create -d bridge app_network
 
+# Docker compose config do dump the env variable to the yaml
```
rtitle: I think this is really just for logging/debugging, it dumps the rendered yaml to the init script log.

http/src/main/resources/init-resources/gce-init.sh
```diff
@@ -376,7 +400,9 @@ if [ ! -z "$JUPYTER_DOCKER_IMAGE" ] ; then
   mkdir -p ${WORK_DIRECTORY}/packages
   chmod a+rwx ${WORK_DIRECTORY}/packages
 
-  # TODO: update this if we upgrade python version
+  # Install everything after having mounted the empty PD
+  # This should not be needed anymore if the jupyter home is a directory of the PD mount point
```
rtitle, line 411: +1 this is a strong candidate for deletion, I can't imagine we still support runtimes where the home directory is not `/home/jupyter`.

http/src/main/resources/init-resources/gce-init.sh
rtitle: These are user-specified Jupyter extensions. I just checked and Terra UI is passing one:
```
{
  "nbExtensions": {
    "saturn-iframe-extension": "https://bvdp-saturn-dev.appspot.com/jupyter-iframe-extension.js"
  },
  "labExtensions": {},
  "serverExtensions": {},
  "combinedExtensions": {}
}
```

I think this is the extension which adds the little `X` in the top right of the jupyter iframe (maybe some other things as well):
image

http/src/main/resources/init-resources/gce-init.sh
rtitle: Umm good question, yeah this script should only execute for GCE anyway. 

Maybe an oversight, your change seems correct 🤷 😅

rtitle (commented):
Thanks for doing this! I got through `gce-init.sh`, will pick up the rest later.

http/src/main/resources/init-resources/gce-init.sh
```diff
@@ -386,7 +412,7 @@ if [ ! -z "$JUPYTER_DOCKER_IMAGE" ] ; then
     retry 3 docker exec -u root ${JUPYTER_SERVER_NAME} ${JUPYTER_SCRIPTS}/kernel/kernelspec.sh ${JUPYTER_SCRIPTS}/kernel ${KERNELSPEC_HOME}
   fi
 
-  # Install notebook.json
+  # Install notebook.json, which is related to the locking logic
```
rtitle: Specifically, I think `notebook.json` is used to populate `Jupyter.notebook.config` in JavaScript extensions. This is used in the [edit-mode.js](https://github.com/DataBiosphere/terra-docker/blob/master/terra-jupyter-base/custom/edit-mode.js#L11) extension that Terra/AoU use.

http/src/main/resources/init-resources/init-actions.sh
```diff
@@ -276,34 +275,19 @@ END
...
-      STAT_EXIT_CODE=0
-      gsutil -q stat ${RSTUDIO_LICENSE_FILE} || STAT_EXIT_CODE=$?
-      if [ $STAT_EXIT_CODE -eq 0 ] ; then
-        echo "Using RStudio license file $RSTUDIO_LICENSE_FILE"
-        gsutil cp ${RSTUDIO_LICENSE_FILE} /etc/rstudio-license-file.lic
-      else
-        echo "" > /etc/rstudio-license-file.lic
-      fi
-    fi
+    echo "" > /var/google_application_credentials.env
```
rtitle: Maybe we can remove the empty file as well if we remove it from [jupyter-docker-compose.yaml](https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/resources/init-resources/jupyter-docker-compose.yaml#L57)

http/src/main/resources/init-resources/shutdown.sh
```diff
@@ -3,10 +3,22 @@
 set -e -x
 
 ##
-# This is a shutdown script designed to run on Leo-created Dataproc clusters.
+# This is a shutdown script designed to run on Leo-created Google Dataproc clusters and Google Compute Engines (GCE).
 ##
 
-# Templated values
+# Runtimes created after September 2024 should have the CLOUD_SERVICE env exported, but for older runtimes, the CLOUD_SERVICE
```
rtitle: Exports in `gce-init.sh`/`init-actions.sh` won't carry over to `startup.sh` /`shutdown.sh` (bash exports are only in scope of that process or child processes). 

So I think the code is correct to always export it here, maybe remove the part in the comment about `Runtimes created after September 2024 should have the CLOUD_SERVICE env exported`

http/src/main/resources/init-resources/shutdown.sh
```diff
@@ -22,12 +34,8 @@ if [ -d '/mnt/disks/work/.jupyter' ] && [ "SHOULD_DELETE_JUPYTER_DIR" = "true" ]
     rm -rf /mnt/disks/work/.local || true
 fi
 
-# https://broadworkbench.atlassian.net/browse/IA-3186
-# This condition assumes Dataproc's cert directory is different from GCE's cert directory, a better condition would be
-# a dedicated flag that distinguishes gce and dataproc. But this will do for now
-# Attempt to fix https://broadworkbench.atlassian.net/browse/IA-3258
-if [ -f "/var/certs/jupyter-server.crt" ]
-then
+# TODO - Why do we need different docker compose commands?
```
rtitle: Probably because COS images need to run docker-compose as a container, whereas Dataproc has it natively installed.

It's interesting we're not using the `cryptopants/docker-compose-gcr` image here, probably because the startup/shutdown scripts never actually need to pull anything from GCR?

http/src/main/resources/init-resources/startup.sh
```diff
@@ -40,11 +40,70 @@ function retry {
...
+    fi
+    if [[ ! -z "$RSTUDIO_DOCKER_IMAGE" ]] ; then
+      IMAGES_TO_RESTART+=(-f /var/docker-compose-files/rstudio-docker-compose-gce.yaml)
+    fi
+    if [[ ! -z "$JUPYTER_DOCKER_IMAGE" ]] ; then
+      IMAGES_TO_RESTART+=(-f /var/docker-compose-files/jupyter-docker-compose-gce.yaml)
+      DATAPROC_IMAGES_TO_RESTART+=(-f /etc/jupyter-docker-compose.yaml )
+    fi
+
+    if [ "$certFileDirectory" = "/certs" ] #if its dataproc the cert directory is '/certs', and we can assume the docker images present
```
rtitle: Could we use the `$CLOUD_SERVICE` variable here?

rtitle:
Made it, overall looks great. We should make sure to do some testing, especially around pause/resume which has more bash restructuring. Looks like some `leo-swat` tests failed, not sure why.

LizBaldo:
Thanks a bunch for the review, let me do a second pass and address your comments and I'll dig into the testing part next.

http/src/main/resources/init-resources/startup.sh
LizBaldo: Oh yeah my bad, thanks for the catch :)

http/src/main/resources/init-resources/init-actions.sh
LizBaldo: Yep, I'll try that

http/src/main/resources/init-resources/shutdown.sh
LizBaldo: Gotcha, we should embed this in the runtime template then, but this will do for now. at least it is a bit clearer

http/src/main/resources/init-resources/gce-init.sh
LizBaldo: What I mean here is that if we do not mount the PD in the home.directory then we can install these in the container directly

http/src/main/resources/init-resources/gce-init.sh
LizBaldo: Thanks!

LizBaldo:
@rtitle and @lucymcnatt I just tested creating, pausing, resuming, deleting with and without the PD for both a GCE and dataproc instance on my BEE and everything worked as expected 👌

rtitle (approved):
Looks good!

lucymcnatt (approved):
Much cleaner now + clear comments, thanks for doing this work!

PR #4779, opened 2024-09-05
[IA-5053] Add retries to SamService, flesh out more authz methods

rtitle:
https://broadworkbench.atlassian.net/browse/IA-5053

## Summary of changes

Draft while I self-review and test. This PR fleshes out more authz logic in `SamService`/`SamServiceInterp`. It does NOT actually update Leo to use `SamService` for authorization yet (I think we will need to do some more testing and potential historical data migrations before that can happen).

### What

- Added pattern for retries in `SamServiceInterp`
- Added access control methods (`checkAuthz`, `create/deleteResource`, etc)
- Improved logging
- Added unit tests

### Why

- I'm trying to prepare for when we can cut over to `SamService` for real. Next step will be understanding/migrating historical data to the desired state. Docs [1](https://docs.google.com/document/d/1FAkxK25hqCdCJWXVdcmpaTE23IeTPHSm_lyR_MuoMf8/edit#heading=h.csta1rbbpzb3) and [2](https://broadworkbench.atlassian.net/wiki/spaces/IA/pages/3323592711/IA-4295+Use+Sam+client+in+leonardo) will help with this.

### What to test

- Unit tests
- Automation tests

### Who tested and where

- [ ] I validated this change 

### DIFF
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/sam/SamRetry.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/sam/SamRetry.scala
@@ -1,10 +1,14 @@
 package org.broadinstitute.dsde.workbench.leonardo.dao.sam
 
 import cats.effect.Async
+import cats.mtl.Ask
+import cats.syntax.all._
 import org.apache.http.HttpStatus
 import org.broadinstitute.dsde.workbench.RetryConfig
 import org.broadinstitute.dsde.workbench.client.sam.ApiException
+import org.broadinstitute.dsde.workbench.leonardo.AppContext
 import org.broadinstitute.dsde.workbench.util2.addJitter
+import org.typelevel.log4cats.StructuredLogger
 
 import java.net.SocketTimeoutException
 import scala.concurrent.duration._
@@ -38,31 +42,47 @@ object SamRetry {
 
   /**
    * Retries an effect with a given retry configuration. Delegates to
-   * fs2.Stream.retry under the hood.
+   * fs2.Stream.retry under the hood but adds logging on retries.
    * @param retryConfig the retry config to use
    * @param fa the effect to retry
+   * @param action String representing the retryable action for logging
    * @return the first successful effect or the last error after retrying
    */
-  def retry[F[_], A](retryConfig: RetryConfig)(fa: F[A])(implicit
-    F: Async[F]
-  ): F[A] =
+  def retry[F[_], A](retryConfig: RetryConfig)(fa: F[A], action: String)(implicit
+    F: Async[F],
+    logger: StructuredLogger[F],
+    ev: Ask[F, AppContext]
+  ): F[A] = {
+    val faWithLogging = fa.onError {
+      case ex if retryConfig.retryable(ex) =>
+        ev.ask.flatMap(ctx => logger.info(ctx.loggingCtx)(s"SamRetry: caught retry-able exception for $action: $ex"))
+    }
     fs2.Stream
-      .retry[F, A](fa,
-                   retryConfig.retryInitialDelay,
-                   retryConfig.retryNextDelay,
-                   retryConfig.maxAttempts,
-                   retryConfig.retryable
+      .retry[F, A](
+        faWithLogging,
+        retryConfig.retryInitialDelay,
+        retryConfig.retryNextDelay,
+        retryConfig.maxAttempts,
+        retryConfig.retryable
       )
       .compile
       .lastOrError
+      .onError(ex =>
+        ev.ask.flatMap(ctx =>
+          logger.error(ctx.loggingCtx, ex)(s"SamRetry: failed $action after ${retryConfig.maxAttempts} tries")
+        )
+      )
+  }
 
   /**
    * Convenience method which uses the default Sam retry policy and takes a
    * thunk instead of an effect. Wraps the thunk in F.blocking(), which is convenient
    * for working with the Java Sam client.
    */
-  def retry[F[_], A](thunk: => A)(implicit
-    F: Async[F]
+  def retry[F[_], A](thunk: => A, action: String)(implicit
+    F: Async[F],
+    logger: StructuredLogger[F],
+    ev: Ask[F, AppContext]
   ): F[A] =
-    retry(defaultSamRetryConfig)(F.blocking(thunk))
+    retry(defaultSamRetryConfig)(F.blocking(thunk), action)
 }
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/sam/SamRetry.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/sam/SamRetry.scala
@@ -1,17 +1,13 @@
 package org.broadinstitute.dsde.workbench.leonardo.dao.sam
 
 import cats.effect.Async
-import cats.mtl.Ask
+import org.apache.http.HttpStatus
 import org.broadinstitute.dsde.workbench.RetryConfig
 import org.broadinstitute.dsde.workbench.client.sam.ApiException
-import org.apache.http.HttpStatus
 import org.broadinstitute.dsde.workbench.util2.addJitter
 
 import java.net.SocketTimeoutException
 import scala.concurrent.duration._
-import org.broadinstitute.dsde.workbench.google2.tracedRetryF
-import org.broadinstitute.dsde.workbench.model.TraceId
-import org.typelevel.log4cats.StructuredLogger
 
 object SamRetry {
 
@@ -41,27 +37,32 @@ object SamRetry {
   }
 
   /**
-   * Retries an effect with a given retry configuration and logging.
-   * Uses `tracedRetryF` from workbench-libs under the hood.
+   * Retries an effect with a given retry configuration. Delegates to
+   * fs2.Stream.retry under the hood.
    * @param retryConfig the retry config to use
    * @param fa the effect to retry
-   * @param action a string representing the action for logging
    * @return the first successful effect or the last error after retrying
    */
-  def retry[F[_], A](retryConfig: RetryConfig)(fa: F[A], action: String)(implicit
-    F: Async[F],
-    logger: StructuredLogger[F],
-    ev: Ask[F, TraceId]
-  ): F[A] = tracedRetryF(retryConfig)(fa, action).compile.lastOrError
+  def retry[F[_], A](retryConfig: RetryConfig)(fa: F[A])(implicit
+    F: Async[F]
+  ): F[A] =
+    fs2.Stream
+      .retry[F, A](fa,
+                   retryConfig.retryInitialDelay,
+                   retryConfig.retryNextDelay,
+                   retryConfig.maxAttempts,
+                   retryConfig.retryable
+      )
+      .compile
+      .lastOrError
 
   /**
-   * Like #retry, but takes a thunk instead of an effect. Wraps the thunk in
-   * F.blocking(), which is convenient for working with the Java Sam client.
+   * Convenience method which uses the default Sam retry policy and takes a
+   * thunk instead of an effect. Wraps the thunk in F.blocking(), which is convenient
+   * for working with the Java Sam client.
    */
-  def retry[F[_], A](thunk: => A, action: String)(implicit
-    F: Async[F],
-    logger: StructuredLogger[F],
-    ev: Ask[F, TraceId]
+  def retry[F[_], A](thunk: => A)(implicit
+    F: Async[F]
   ): F[A] =
-    retry(defaultSamRetryConfig)(F.blocking(thunk), action)
+    retry(defaultSamRetryConfig)(F.blocking(thunk))
 }
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/sam/SamRetry.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/sam/SamRetry.scala
@@ -40,16 +40,28 @@ object SamRetry {
     case _ => false
   }
 
+  /**
+   * Retries an effect with a given retry configuration and logging.
+   * Uses `tracedRetryF` from workbench-libs under the hood.
+   * @param retryConfig the retry config to use
+   * @param fa the effect to retry
+   * @param action a string representing the action for logging
+   * @return the first successful effect or the last error after retrying
+   */
+  def retry[F[_], A](retryConfig: RetryConfig)(fa: F[A], action: String)(implicit
+    F: Async[F],
+    logger: StructuredLogger[F],
+    ev: Ask[F, TraceId]
+  ): F[A] = tracedRetryF(retryConfig)(fa, action).compile.lastOrError
+
+  /**
+   * Like #retry, but takes a thunk instead of an effect. Wraps the thunk in
+   * F.blocking(), which is convenient for working with the Java Sam client.
+   */
   def retry[F[_], A](thunk: => A, action: String)(implicit
     F: Async[F],
     logger: StructuredLogger[F],
     ev: Ask[F, TraceId]
   ): F[A] =
     retry(defaultSamRetryConfig)(F.blocking(thunk), action)
-
-  def retry[F[_], A](retryConfig: RetryConfig)(fa: F[A], action: String)(implicit
-    F: Async[F],
-    logger: StructuredLogger[F],
-    ev: Ask[F, TraceId]
-  ): F[A] = tracedRetryF(retryConfig)(fa, action).compile.lastOrError
 }
--- a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/sam/SamRetry.scala
+++ b/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/sam/SamRetry.scala
@@ -18,7 +18,7 @@ object SamRetry {
   /**
    * Initial delay of 1 second, exponential retry up to 5 times.
    */
-  private val retryConfig =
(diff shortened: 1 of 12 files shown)


http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/sam/SamRetry.scala
```diff
@@ -0,0 +1,67 @@
...
+      e.getCode == HttpStatus.SC_INTERNAL_SERVER_ERROR ||
+      e.getCode == HttpStatus.SC_BAD_GATEWAY ||
+      e.getCode == HttpStatus.SC_SERVICE_UNAVAILABLE ||
+      e.getCode == HttpStatus.SC_GATEWAY_TIMEOUT
+    case _ => false
+  }
+
+  /**
+   * Retries an effect with a given retry configuration and logging.
+   * Uses `tracedRetryF` from workbench-libs under the hood.
```
rtitle: One thing I want to evaluate is whether `tracedRetryF` suits our Sam needs here, especially the logging piece. It was originally used for GCP clients (but should be pretty general purpose).

TCL `SamRetry` basically implements retry logic from scratch, which I'd rather not do since `Stream.retry` exists in Scala.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/sam/SamService.scala
```diff
@@ -13,45 +19,45 @@ trait SamService[F[_]] {
 
   /**
    * Gets a user's pet GCP service account, using the user's token.
-   * @param userInfo the user info containing an access token
+   * @param bearerToken the user's access token
```
rtitle, line 23: I decided to have these methods take a token string instead of a `UserInfo` object. It was too tempting to use the `userInfo.email` field for logging/etc, but that field should not be relied upon. We should instead call `SamService.getUserEmail` when we need to resolve the email from a token.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/sam/SamServiceInterp.scala
```diff
@@ -34,38 +41,45 @@ class SamServiceInterp[F[_]](apiClientProvider: SamApiClientProvider[F],
...
-      googleApi <- apiClientProvider.googleApi(userInfo.accessToken.token)
-      pet <- F.blocking(googleApi.getPetServiceAccount(googleProject.value)).adaptError { case e: ApiException =>
-        SamException.create("Error getting pet service account from Sam", e, ctx.traceId)
+      googleApi <- apiClientProvider.googleApi(bearerToken)
+      userEmail <- getUserEmail(bearerToken)
+      pet <- SamRetry.retry(googleApi.getPetServiceAccount(googleProject.value), "getPetServiceAccount").adaptError {
+        case e: ApiException =>
+          SamException.create("Error getting pet service account from Sam", e, ctx.traceId)
       }
+      _ <- logger.info(ctx.loggingCtx)(
```
rtitle, line 61: Added some log statements to these methods.

Note: this class looks a lot like WSM [SamService](https://github.com/DataBiosphere/terra-workspace-manager/blob/main/service/src/main/java/bio/terra/workspace/service/iam/SamService.java), but in Scala. :)

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/dao/sam/SamServiceInterpSpec.scala
```diff
@@ -4,30 +4,32 @@ import akka.http.scaladsl.model.StatusCodes
...
+import org.mockito.Mockito.{doNothing, doThrow, verify, when}
+import org.mockito.{ArgumentCaptor, ArgumentMatchers}
 import org.scalatest.BeforeAndAfterAll
 import org.scalatest.funspec.AnyFunSpecLike
 import org.scalatestplus.mockito.MockitoSugar
 
 import java.util.UUID
+import scala.jdk.CollectionConverters._
 
 class SamServiceInterpSpec extends AnyFunSpecLike with LeonardoTestSuite with BeforeAndAfterAll with MockitoSugar {
```
rtitle, line 32: Lot of tests here but I think it has pretty complete coverage of `SamServiceInterp`. It mocks everything so runs very fast. I tried to organize test cases a bit with ScalaTest [FunSpec](https://www.scalatest.org/getting_started_with_fun_spec).

rtitle:
@LizBaldo this might be relevant to your auth domain work. I'm trying to get `SamService` in Leo "ready for prime time". This PR doesn't actually start using it yet for authz, but improves some of the code.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/sam/SamRetry.scala
```diff
@@ -0,0 +1,67 @@
...
+
+object SamRetry {
+
+  /**
+   * Initial delay of 1 second, exponential retry up to 5 times.
+   */
+  private val defaultSamRetryConfig =
+    RetryConfig(addJitter(1 seconds, 1 seconds), _ * 2, 5, isRetryable)
+
+  /**
```
rtitle, line 24: Borrowed this logic from TCL `SamRetry`: https://github.com/DataBiosphere/terra-common-lib/blob/develop/src/main/java/bio/terra/common/sam/SamRetry.java#L58

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/sam/SamServiceInterp.scala
```diff
@@ -135,4 +159,130 @@ class SamServiceInterp[F[_]](apiClientProvider: SamApiClientProvider[F],
...
+        case _ =>
+          F.raiseError(
+            LeoInternalServerError(
+              "Internal error creating Sam resource: google project or workspace parent is required",
+              Some(ctx.traceId)
+            )
+          )
+      }
+
+      // All Leo resources have a creator role
```
mlilynolting: Is this true for apps with the workspace-shared access policy? Leo's config has an "owner" role there. I'm assuming also that when you say "leo resource" you're referring only to v1 stuff? I'm not sure how WSM handles these roles.

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/mocks.scala
```diff
@@ -323,8 +323,30 @@ class BaseMockSamService extends SamService[IO] {
   ): IO[String] =
     IO.pure(tokenValue)
 
-  override def lookupWorkspaceParentForGoogleProject(userInfo: UserInfo, googleProject: GoogleProject)(implicit
+  override def lookupWorkspaceParentForGoogleProject(bearerToken: String, googleProject: GoogleProject)(implicit
     ev: Ask[IO, AppContext]
   ): IO[Option[WorkspaceId]] = IO.pure(Some(workspaceId))
+
+  override def checkAuthz(bearerToken: String, samResourceId: SamResourceId, action: String)(implicit
```
mlilynolting: Is there a reason for "authz" with the Z? I've seen it in WSM too. Is it a convention? Or is it a way to get around a reserved word, like "clazz" in Java?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/sam/SamServiceInterp.scala
mlilynolting: Also, are there cases where the Leonardo SA is creating resources?

rtitle: So I was trying to assume "desired state" here, where we migrate away from WSM-owned resources, and use dedicated resource types for Leo. That was a recommendation in [this doc](https://docs.google.com/document/d/1FAkxK25hqCdCJWXVdcmpaTE23IeTPHSm_lyR_MuoMf8/edit).

Hm you're right that `kubernetes-app-shared` has `owner` and `user` roles. And `kubernetes-app` has `creator` and `manager`.

Maybe I'll revisit the method signature here… I was trying to see if `createResource` could be generic to all types of resources Leo supports, or if we need a `createRuntimeResource`, `createDiskResource`, `createKubernetesAppResource`, etc.

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/mocks.scala
rtitle: Authz == authorization, as opposed to Authn which is authentication.

I actually just copied the method name from [Workspace Manager SamService](https://github.com/DataBiosphere/terra-workspace-manager/blob/5d0c98e78572d2e0fac39102d0eba5ddfa6649d1/service/src/main/java/bio/terra/workspace/service/iam/SamService.java#L596).

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/sam/SamServiceInterp.scala
rtitle: > Also, are there cases where the Leonardo SA is creating resources?

I don't think so -- all Sam resources should be created by the user. It's also a recommendation to not create Sam resources async. So I was intentionally giving the `createResource` method a `userToken` parameter.

mlilynolting: That makes sense. And I love the idea of a standardized Leo-managed resource with predictable roles. Maybe creating a resource gives you the Creator role but depending on the resource, the context, and the user, they could get other roles as well? This might just require a bit more design to nail down an approach.

mlilynolting: We could start with this method and add other methods for special cases if we need to.

rtitle: @mlilynolting what do you think of [this change](https://github.com/DataBiosphere/leonardo/pull/4779/commits/a74f5bfc48fca912b37913856d0cbc2ec467bb1e)? Instead of assuming a `creator` role, I updated `SamService.createResource` to just take the policies directly. We can maybe add specialized methods (`getRuntimePolicies`, `getAppPolicies`, etc) later on.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/sam/SamServiceInterp.scala
```diff
@@ -211,7 +217,7 @@ class SamServiceInterp[F[_]](apiClientProvider: SamApiClientProvider[F],
                               samResourceId: SamResourceId,
                               projectParent: Option[GoogleProject],
                               workspaceParent: Option[WorkspaceId],
-                              creator: Option[WorkbenchEmail]
+                              policies: Map[String, SamPolicyData]
```
mlilynolting, line 218: Does it make sense to type the key here as `SamPolicyName`?

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/mocks.scala
mlilynolting: It'd make me happier not to have the shortening; I'm easily confused by unfamiliar terms 😅 `checkAuthorization` or `checkAuthorized` would feel more readable

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/sam/SamServiceInterp.scala
mlilynolting: I think that's a solid change that reflects the Sam contract really faithfully - Leo calls Sam with a new policy it wants to create for this resource. The policy name doesn't need to be unique, right?

rtitle: Right, the policy name can be any string -- I think it just needs to be unique per resource (e.g. you can't have 2 `foo` policies on a resource). But the Sam request should fail if uniqueness is violated.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/sam/SamServiceInterp.scala
rtitle: I went with String because I don't love how `SamPolicyName` is an enum in Leo. It's not really an enumeration that maps to anything in Sam, this is just an arbitrary name for the policy. (Often role names are reused for policy names, see [WSM for example](https://github.com/DataBiosphere/terra-workspace-manager/blob/main/service/src/main/java/bio/terra/workspace/service/iam/SamService.java#L1173)).

`SamPolicyName` used all over the place in the _old_ auth code (and conflated with roles somewhat) so we can't remove/change it yet.

http/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/mocks.scala
rtitle: Fixed!

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/sam/SamRetry.scala
rtitle: Yeah the retries work OK but I don't like the logging behavior that `tracedRetryF` gives us. It seems to always log the response which is not always desired (Sam responses can include things like tokens and keys).

Will explore more here..

rtitle: Fixed in https://github.com/DataBiosphere/leonardo/pull/4779/commits/1bc0d4520fb4c6436c106a7134bffd0f2516001f, not using the wb-libs logging anymore.

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/sam/SamServiceInterp.scala
mlilynolting: That's a great reason not to do it, lol. Could you add a comment suggesting that we might want to move to a dedicated policyName type?

mlilynolting:
This all looks great to me. Are you planning anything else before you un-draft the PR?

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/sam/SamServiceInterp.scala
rtitle: Sure 👍

rtitle:
Thanks! I'm just tweaking some of the retry behavior, but not planning anything else substantial. I'll plan to un-draft it soon.

rtitle:
@mlilynolting @LizBaldo I tested this on a BEE and things look good. Moved out of draft, requesting reviews -- thanks.

mlilynolting (approved):

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/sam/SamServiceInterp.scala
```diff
@@ -135,4 +164,126 @@ class SamServiceInterp[F[_]](apiClientProvider: SamApiClientProvider[F],
...
+    ev: Ask[F, AppContext]
+  ): F[List[String]] = for {
+    ctx <- ev.ask
+    resourcesApi <- apiClientProvider.resourcesApi(bearerToken)
+    resources <- SamRetry
+      .retry(resourcesApi.listResourcesAndPoliciesV2(samResourceType.asString), "listResourcesAndPoliciesV2")
+      .adaptError { case e: ApiException =>
+        SamException.create("Error listing resources from Sam", e, ctx.traceId)
+      }
+  } yield resources.asScala.toList.map(_.getResourceId)
```
mlilynolting, line 212: "has access to" might be a bit limiting. I don't know that there is a use case for listing just the resource IDs here - there might be a specific access right that we want to check, like `read`. I don't think we necessarily need to change that in this PR, since this method isn't actually used yet. Just a headsup that this might need to change for Leo's use case

rtitle: Yeah that's a good point - we might want to additionally check specific actions here. I agree this can be changed in a subsequent PR.

LizBaldo: So y ou are saying adding a `scope` to this function that could be a role or an action? E.g. list all resources the user has read access to? That might be tricky since different resources can have completely different actions no?

LizBaldo (approved):

http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/sam/SamServiceInterp.scala
mlilynolting: I think ideally the current use cases would be covered by adding a list of actions. We don't want to check roles if we can avoid it (we might have to if we have to use parent role as a standin for child action). This method is only called in the context of a single resource type so that will scope us to only the actions for that type.

rtitle: +1 to @mlilynolting -- I'm imagining adding a new parameter `actions: List[SamAction]` to this method, which would further constrain the results.

PR #4782, opened 2024-09-10
Authorize to GAR in init scripts

lucymcnatt:
__Jira ticket__: https://broadworkbench.atlassian.net/browse/IA-4518

### What

Authorize to GAR using `auth configure-docker` rather than custom gcr-enabled docker compose

### Why

With switching to GAR instead of GCR, the init scripts that pull our docker images need to be pulling from GAR

### What to test

- 

### Who tested and where

- [ ] I validated this change 


LizBaldo (commented):
I understand that GCR is being redirected to GAR, but eventually the GCR repos will be fully deprecated, and so it would make sense to me to have Leo explicitly pull from GAR

 See https://github.com/DataBiosphere/leonardo/pull/4372/files

lucymcnatt:
We can discuss it more, but this is Leo pulling from GAR, on the cloud console it specifically says that:

> All "*gcr.io" traffic is routed to Artifact Registry.

The paths are all the same so maybe I'm missing something but I don't see how adding a different regex for GAR makes a difference?

LizBaldo (approved):
Gotcha, I did not realize the paths were similar since they contained `gcr` in the string 🤔  In that case I don't know why the Sfkit folks wanted to do that.

http/src/main/resources/init-resources/gce-init.sh
```diff
@@ -150,12 +150,13 @@ else
 fi
 
 if grep -qF "gcr.io" <<< "${JUPYTER_DOCKER_IMAGE}${RSTUDIO_DOCKER_IMAGE}${PROXY_DOCKER_IMAGE}${WELDER_DOCKER_IMAGE}" ; then
-  log 'Authorizing GCR...'
-  DOCKER_COMPOSE="docker run --rm -v /var/run/docker.sock:/var/run/docker.sock -v /var:/var -w=/var cryptopants/docker-compose-gcr"
-else
-  DOCKER_COMPOSE="docker run --rm -v /var/run/docker.sock:/var/run/docker.sock -v /var:/var docker/compose:1.29.2"
+  log 'Authorizing GAR...'
+  $GCLOUD_CMD auth configure-docker
```
rtitle: Not sure how important this is, but I think the reason `cryptopants/docker-compose-gcr` was used was to support private GCR images. The [cryptopants/docker-compose-gcr](https://hub.docker.com/r/cryptopants/docker-compose-gcr) doc says:
```
This container helps resolve an issue with Container Optimized OS. The OS does not come with docker-compose, and instead expects users to use containerized docker-compose. However the official containerized docker-compose does not include docker-credential-gcr, which is needed to pull Docker images from a private Google Container Registry.

This container uses the official docker-compose as a base, and adds docker-credential-gcr. It can be used just like the normal containerized docker-compose, but can now access gcr.io.
```

Do private images still work after this change?

(All our standard images are public, but technically we do support private GCR images. I'm not sure if there are any users currently using this feature though.)

lucymcnatt: ahh ok, I just tested that with an image not in the broad public repo and didn't work on this branch or on dev/prod, so it seems like something we may have supported in the past but now isn't working...

lucymcnatt: I can try to fix it though

http/src/main/resources/init-resources/gce-init.sh
```diff
@@ -220,9 +220,13 @@ START_TIME=$(date +%s)
 STEP_TIMINGS=($(date +%s))
 
 
-# Use specific docker compose command if the container is coming from GCR, see  https://hub.docker.com/r/cryptopants/docker-compose-gcr
-# TODO - Also check for GAR see https://broadworkbench.atlassian.net/browse/IA-4518
-if grep -qF "gcr.io" <<< "${JUPYTER_DOCKER_IMAGE}${RSTUDIO_DOCKER_IMAGE}${PROXY_DOCKER_IMAGE}${WELDER_DOCKER_IMAGE}" ; then
+# everything in the public broad us.gcr.io bucket has been migrated to GAR
+if grep -qF "us.gcr.io" <<< "${JUPYTER_DOCKER_IMAGE}${RSTUDIO_DOCKER_IMAGE}${PROXY_DOCKER_IMAGE}${WELDER_DOCKER_IMAGE}" ; then
```
LizBaldo: Might be a silly question sorry, but could there still be private repos out there that also start with this path? I do not thinks so since I am pretty sure the paths are unique but still wanted to double check

LizBaldo: Also you might not need this if, elif, else You could do:

if private and stored in GCR: Authorize GCR
else: Authorize GAR

What do you think?

jenkins/gce-custom-images/prepare_gce_image.sh
```diff
@@ -31,7 +31,6 @@ terra_jupyter_aou_old="us.gcr.io/broad-dsp-gcr-public/terra-jupyter-aou:2.2.13"
 cos_gpu_installer="gcr.io/cos-cloud/cos-gpu-installer:v2.1.9"
 google_cloud_toolbox="us.gcr.io/cos-cloud/toolbox:v20230714"
 docker_composer="docker/compose:1.29.2"
-docker_composer_with_auth="cryptopants/docker-compose-gcr"
```
LizBaldo: Wait, I think you still need it for GCE (COS) images no? It's dataproc that does not need it

http/src/main/resources/init-resources/gce-init.sh
lucymcnatt: ooh yeah good points...not entirely sure how to determine private vs. public at this level, but I'll think on that

jenkins/gce-custom-images/prepare_gce_image.sh
lucymcnatt: oops! yeah I thought I'd gotten rid of all these changes, but I missed one, thanks

LizBaldo (approved):

